PR #37, opened 2014-06-22
For discussion: compact field elements

peterdettman:
- secp256k1_fec_t: compact field element representation
- secp256k1_fec_* methods: import/export, mul/sqr/sqr_pow
- only field "64bit" at the moment (no asm)
- sqrt and constant-time inversion changed to use this

A compact representation allows faster multiplication/squaring, although the overheads of converting back and forth make it unwieldy in the point formulae. Nevertheless, sqrt and Fermat-based inversion require >256 consecutive mul/sqr operations, making the conversion worthwhile.

The measured performance improvement for isolated secp256k1_fe_sqrt testing (and by extension secp256k1_fe_inv) is around %18, including all overheads.

### DIFF
--- a/src/field.h
+++ b/src/field.h
@@ -109,4 +109,15 @@ void static secp256k1_fe_get_hex(char *r, int *rlen, const secp256k1_fe_t *a);
 /** Convert a hexadecimal string to a field element. */
 void static secp256k1_fe_set_hex(secp256k1_fe_t *r, const char *a, int alen);
 
+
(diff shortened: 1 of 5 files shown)


sipa:
Looks very outdated by now.

peterdettman:
@sipa Agreed, closing. mul/sqr are significantly faster now than when this was first mooted. Also, I was unable to get a _sqr_pow to show any significant improvement, against my expectations that a fully in-registers loop should have been faster. I never did investigate Montgomery in this context.

PR #64, opened 2014-09-21
enable libsecp256k1_jni shared lib

theuni:
This enables a libsecp256k1_jni shared lib to be built. It depends on libsecp256k1.

```./configure --enable-jni``` should be all it takes assuming your java environment is setup properly. For now, there's no way to specify where the includes should be found if they're not located in $JAVA_HOME/include, as I'm not sure if that's necessary or not.

Notice that the lib has been renamed in the .java to reflect the lib's new name. This was done to match standard distro packaging names.

This is untested in a real java/jni build, verification that it actually works will be needed before merging.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,6 +1,12 @@
 ACLOCAL_AMFLAGS = -I build-aux/m4
 
 lib_LTLIBRARIES = libsecp256k1.la
+if USE_JNI
+JNI_LIB = libsecp256k1_jni.la
+noinst_LTLIBRARIES = $(JNI_LIB)
+else
+JNI_LIB =
+endif
 include_HEADERS = include/secp256k1.h
 noinst_HEADERS =
 noinst_HEADERS += src/scalar.h
@@ -44,8 +50,10 @@ pkgconfig_DATA = libsecp256k1.pc
 
 libsecp256k1_la_SOURCES = src/secp256k1.c
 libsecp256k1_la_CPPFLAGS = -I$(top_srcdir)/include $(SECP_INCLUDES)
-libsecp256k1_la_LIBADD = $(SECP_LIBS)
+libsecp256k1_la_LIBADD = $(JNI_LIB) $(SECP_LIBS)
 
+libsecp256k1_jni_la_SOURCES  = src/java/org_bitcoin_NativeSecp256k1.c
+libsecp256k1_jni_la_CPPFLAGS = $(JNI_INCLUDES)
 
 noinst_PROGRAMS =
 if USE_BENCHMARK
--- a/build-aux/m4/ax_jni_include_dir.m4
+++ b/build-aux/m4/ax_jni_include_dir.m4
@@ -1,0 +1,126 @@
+# ===========================================================================
+#    http://www.gnu.org/software/autoconf-archive/ax_jni_include_dir.html
+# ===========================================================================
+#
+# SYNOPSIS
+#
+#   AX_JNI_INCLUDE_DIR
+#
+# DESCRIPTION
+#
+#   AX_JNI_INCLUDE_DIR finds include directories needed for compiling
+#   programs using the JNI interface.
+#
+#   JNI include directories are usually in the Java distribution. This is
+#   deduced from the value of $JAVA_HOME, $JAVAC, or the path to "javac", in
+#   that order. When this macro completes, a list of directories is left in
+#   the variable JNI_INCLUDE_DIRS.
+#
+#   Example usage follows:
+#
+#     AX_JNI_INCLUDE_DIR
+#
+#     for JNI_INCLUDE_DIR in $JNI_INCLUDE_DIRS
+#     do
+#             CPPFLAGS="$CPPFLAGS -I$JNI_INCLUDE_DIR"
+#     done
+#
+#   If you want to force a specific compiler:
+#
+#   - at the configure.in level, set JAVAC=yourcompiler before calling
+#   AX_JNI_INCLUDE_DIR
+#
+#   - at the configure level, setenv JAVAC
+#
+#   Note: This macro can work with the autoconf M4 macros for Java programs.
+#   This particular macro is not part of the original set of macros.
+#
+# LICENSE
+#
+#   Copyright (c) 2008 Don Anderson <dda@sleepycat.com>
+#
+#   Copying and distribution of this file, with or without modification, are
+#   permitted in any medium without royalty provided the copyright notice
+#   and this notice are preserved. This file is offered as-is, without any
+#   warranty.
+
+#serial 10
+
+AU_ALIAS([AC_JNI_INCLUDE_DIR], [AX_JNI_INCLUDE_DIR])
+AC_DEFUN([AX_JNI_INCLUDE_DIR],[
+
+JNI_INCLUDE_DIRS=""
+
+if test "x$JAVA_HOME" != x; then
+	_JTOPDIR="$JAVA_HOME"
+else
+	if test "x$JAVAC" = x; then
+		JAVAC=javac
+	fi
+	AC_PATH_PROG([_ACJNI_JAVAC], [$JAVAC], [no])
+	if test "x$_ACJNI_JAVAC" = xno; then
+		AC_MSG_ERROR([cannot find JDK; try setting \$JAVAC or \$JAVA_HOME])
+	fi
+	_ACJNI_FOLLOW_SYMLINKS("$_ACJNI_JAVAC")
+	_JTOPDIR=`echo "$_ACJNI_FOLLOWED" | sed -e 's://*:/:g' -e 's:/[[^/]]*$::'`
+fi
+
+case "$host_os" in
+        darwin*)        _JTOPDIR=`echo "$_JTOPDIR" | sed -e 's:/[[^/]]*$::'`
+                        _JINC="$_JTOPDIR/Headers";;
+        *)              _JINC="$_JTOPDIR/include";;
+esac
+_AS_ECHO_LOG([_JTOPDIR=$_JTOPDIR])
+_AS_ECHO_LOG([_JINC=$_JINC])
+
+# On Mac OS X 10.6.4, jni.h is a symlink:
+# /System/Library/Frameworks/JavaVM.framework/Versions/Current/Headers/jni.h
+# -> ../../CurrentJDK/Headers/jni.h.
+AC_CHECK_FILE([$_JINC/jni.h],
+	[JNI_INCLUDE_DIRS="$JNI_INCLUDE_DIRS $_JINC"],
+	[_JTOPDIR=`echo "$_JTOPDIR" | sed -e 's:/[[^/]]*$::'`
+	 AC_CHECK_FILE([$_JTOPDIR/include/jni.h],
+		[JNI_INCLUDE_DIRS="$JNI_INCLUDE_DIRS $_JTOPDIR/include"],
+                AC_MSG_ERROR([cannot find JDK header files]))
+	])
+
+# get the likely subdirectories for system specific java includes
+case "$host_os" in
+bsdi*)          _JNI_INC_SUBDIRS="bsdos";;
+freebsd*)       _JNI_INC_SUBDIRS="freebsd";;
+linux*)         _JNI_INC_SUBDIRS="linux genunix";;
+osf*)           _JNI_INC_SUBDIRS="alpha";;
+solaris*)       _JNI_INC_SUBDIRS="solaris";;
+mingw*)		_JNI_INC_SUBDIRS="win32";;
+cygwin*)	_JNI_INC_SUBDIRS="win32";;
+*)              _JNI_INC_SUBDIRS="genunix";;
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by sipa** [comment]
@theuni @TheBlueMatt Any opinions on how this should be done after #208? Reintroduce a global in the JNI C code, or have the Java side hold a pointer to an initialized context on startup?

**Comment by theuni** [comment]
@sipa holding long-lasting references via jni can be error-prone and problematic due to gc. I think I'd prefer the global in C unfortunately.

**Comment by sipa** [comment]
@theuni The Java GC shouldn't apply to C-allocated objects, I think?

In any case, if the JNI C side ends up keeping the context object, initialized at loading time, I feel unconfortable with it being the same .so file, as that would mean that ever C user of the library ends up allocating and initializing an extra context.

**Comment by theuni** [comment]
@sipa You're correct of course, brainfart on my part. I'm used to going the other way around with jni (calling java from c) where the gc gets in the way.

Looks like standard practice is to simply return a (jlong)ptr. Sounds good to me. I suppose we'll need to add a jni function to create/init the context?

**Comment by sipa** [comment]
It's a bit annoying that we don't have any tests for the Java code. Just a trivial Java program that verifies a valid and an invalid signature would already be nice.

**Comment by theuni** [comment]
@sipa Indeed. I'll see if I can throw something together. Would you like me to rebase this on top of #208 while I'm at it?

**Comment by sipa** [comment]
@theuni Sounds good; thanks a lot!

**Comment by iangfc** [comment]
On 13/02/2015 20:32 pm, Pieter Wuille wrote:
> It's a bit annoying that we don't have any tests for the Java code. Just
> a trivial Java program that verifies a valid and an invalid signature
> would already be nice.

Understanding you're pointing at me, I'm a bit snowed under at the 
moment and would be very happy if someone else could pick it up.

iang

**Comment by faizkhan00** [comment]
i realize this is lagging behind the current -master, but i was looking for a reason to explore this and ended up writing a simple testcase to maybe help things along- fwiw, this ran with no problems on my machine 

test: https://github.com/faizkhan00/secp256k1/blob/jni-addverifytest/src/java/org/bitcoin/NativeSecp256k1Test.java

**Comment by faizkhan00** [comment]
Not sure if this is going anywhere, but I did the rebase anyway : https://github.com/faizkhan00/secp256k1/tree/jni-rebase

Tests are passing , save 1 nit/warning from compiler 
```
src/java/org_bitcoin_NativeSecp256k1.c: In function ‘Java_org_bitcoin_NativeSecp256k1_secp256k1_1ecdsa_1verify’:
src/java/org_bitcoin_NativeSecp256k1.c:8:2: warning: ISO C90 forbids mixed declarations and code [-Wpedantic]
  secp256k1_context_t *vrfy = secp256k1_context_create(SECP256K1_CONTEXT_VERIFY);
  ^
```

Diff: https://github.com/bitcoin/secp256k1/compare/master...faizkhan00:jni-rebase

Happy to take care of that nit if this has a chance of going upstream.

**Comment by faizkhan00** [comment]
@theuni If you have a chance, do you mind looking over my rebase attempt and considering it for inclusion in this PR? I realize everybody is a bit snowed under ATM, but I believe the issues brought up prior in this thread have been addressed (and if something hasn't been, I can make changes - or if the PR is just not worth merging anymore I'd be interested to know this as well)

**Comment by sipa** [comment]
I will definitely merge it if it works and has tests.

**Comment by theuni** [comment]
@faizkhan00 Thanks for reviving this. Last time I rebased and was adding tests, I had a brief discussion about it on IRC and it sounded as though this wasn't wanted anymore. If @sipa says he still wants it, I'll have another look.

Thanks again!

**Comment by faizkhan00** [comment]
Sure thing @theuni. While working on this I noticed the JNI coverage of the library is lacking (only supporting ecdsa_verify); I didn't dig far enough to understand the intent of the original committer, but if keeping the JNI bindings upstream is the long-term intent, then I'd be interested in expanding the code coverage to include the rest of the secp256k1 API, including testcases, as well. 

I think there's quite a bit of value in that, but I don't want to get in the way of your impending changes, so if you pushed a WIP branch up sometime I could probably work with your changes in parallel.

**Comment by sipa** [comment]
@faizkhan00 Feel free to expand it.

**Comment by faizkhan00** [comment]
@sipa OK, should have something to share later this week.

**Comment by faizkhan00** [comment]
Hi all, i added support for several more secp256k1 functions (sign, pubkey_create, sec/pub_verify) as well as associated testcases, which are all looking good - - this is a preliminary commit as i have yet to go in and clean it up with the appropriate comments, but i wanted to share what i had so that it has some eyes

https://github.com/faizkhan00/secp256k1/commit/7838a3ea4aabaa372cb9f4422531502aa6e43e60
comments, review, critique, nits all appreciated - - ! @theuni @sipa

**Comment by faizkhan00** [comment]
Chatted with cfields/sipa about this on IRC, need to rebase with contexts handled once and only once in the JNI/Java wrapper, possibly using synch primitives

should have something mid-week or so with the update

**Comment by faizkhan00** [comment]
Updated branch with pass-by-reference context instead of reinit-per-operation refactor

branch: https://github.com/faizkhan00/secp256k1/tree/jni-add-sign-verify-test

file-that-needs-review (pls!): https://github.com/faizkhan00/secp256k1/blob/jni-add-sign-verify-test/src/java/org_bitcoin_NativeSecp256k1.c

Thanks again

**Comment by theuni** [comment]
I think this needs some way of destroying the context. Maybe add a finalize() method that calls into c similar to the context init function? I know nothing of java though, so I'm clueless as to how that works on static members.

**Comment by faizkhan00** [comment]
Yes, it seems I missed that method somehow, I've added a context_destroy() method that should now take care of that. About statics:

Static fields in Java aren't GC'ed until the class is unloaded or the classloader is unloaded, so the option for that is to move from static methods to a object-instantiation based approach, which would increase the complexity, for about 8 bytes of memory (java longs are 64 bit)- If its wanted I can look into doing this definitely, but I'm unsure if the added complexity would make the trade-off worthwhile.

**Comment by theuni** [comment]
@faizkhan00 The java long is just a pointer to malloc'd memory from the native side, which I believe is several hundred kb.

**Comment by faizkhan00** [comment]
Worked on this a bit tonight, basically there is a edge case that I'm not 100% sure with how to handle: 

We have builds on Travis for 32bit Linux , yet Travis doesn't A) support 32 bit JDKs  (not sure how to fix) and B) the JNI lib is relatively untested with 32bit (fixable)

For the moment, I've disabled Travis for 32bit builds, and added documentation to the runner, but I'm not sure if that is the correct strategy for this instance. (https://travis-ci.org/9fe95262548e/secp256k1/jobs/61240914#L395)

Side note: I've also disabled the tests for `make distcheck` due to its unique build routine, you can check this also by looking at the output of ./run_jni_tests.sh in Travis. (https://travis-ci.org/9fe95262548e/secp256k1/jobs/61240912#L570)

Interesting

**Comment by theuni** [comment]
Ah, I was working on this at the same time. Here's my take on it: https://github.com/theuni/secp256k1/tree/travis-jni .
This adds a ```make check-java``` target, which anyone can run locally. Travis runs that as well.

The jar is cached for the future, which should help avoid any website downtime issues.
Build results here: https://travis-ci.org/theuni/secp256k1/builds/61244462

**Comment by faizkhan00** [comment]
ah! nice work @theuni - i think i prefer your makefiles soln over hacky bash scripts ... so I think the only step is to include `make check-java` in .travis.yml, is that right?

**Comment by theuni** [comment]
@faizkhan00 it's already in there :)
See build 71.26 in the link above.

**Comment by faizkhan00** [comment]
Haha I see , yea that works far far better than what I had in mind :) Cool! Nice work @theuni

**Comment by theuni** [comment]
@faizkhan00 just taking one last look at this.

I think it would make sense to re-work the native functions to pass the size parameters as separate int's rather than through the bytebuffers. As-is, I believe big-endian is broken.

So that would make this:
```c
JNIEXPORT jint JNICALL Java_org_bitcoin_NativeSecp256k1_secp256k1_1ecdsa_1verify
  (JNIEnv* env, jclass classObject, jobject byteBufferObject, jlong ctx_l)
{
        secp256k1_context_t *ctx = (secp256k1_context_t*)ctx_l;

       unsigned char* data = (unsigned char*) (*env)->GetDirectBufferAddress(env, byteBufferObject);
       int sigLen = *((int*)(data + 32));
       int pubLen = *((int*)(data + 32 + 4));
```
look more like this:
```
JNIEXPORT jint JNICALL Java_org_bitcoin_NativeSecp256k1_secp256k1_1ecdsa_1verify
  (JNIEnv* env, jclass classObject, jobject byteBufferObject, jlong ctx_l, jint sigLen, jint pubLen)
{
       secp256k1_context_t *ctx = (secp256k1_context_t*)ctx_l;

       unsigned char* data = (unsigned char*) (*env)->GetDirectBufferAddress(env, byteBufferObject);
```
Obviously the function signature would have to be fixed up.

Thoughts?

faizkhan00:
@theuni Hey, I'm curious where you might be seeing platform-specific broken-ness... I would have thought this line in the Java code `byteBuff.order(ByteOrder.nativeOrder());` would have appropriately detected endian-ness, am I missing something? Although, if you just prefer to have a more explicit function sig, I definitely think it looks better from a API perspective

I'm also pretty interested in checking endianness on a VM or via qemu, but I totally can understand if you think thats a tangent here :) 

Interested to know more, I havent' dealt too heavily with platform-specific endianness issues in the past, and know how important they are to handle in this piece of code.

theuni:
@faizkhan00 pushed a quick change to https://github.com/theuni/secp256k1/commit/f69f4719e6f4337a3e5c2d929afad32fb131a90e.

theuni:
@faizkhan00 Yes, it looks like the nativeOrder() probably should've taken care of it. Still, I don't see why we should risk serializing/deserializing when we can pass by value more cleanly.

faizkhan00:
@theuni Ok, and yes, reviewing your code, the changes you made definitely make for a cleaner result, which is preferred, and also look good. 

One question about this `byteBuff.order(ByteOrder.LITTLE_ENDIAN);` would it matter specifying what endianness it's in, coming out on the JNI side before `GetDirectBufferAddress`? I think I might prefer to leave that to the JVM's auto-detection, actually

faizkhan00:
@theuni I suppose I just was curious what we get from that change, irrespective of what the JVM is doing when using nativeOrder()

theuni:
Whoops! Thanks for catching that!

I was playing around and forgot to revert that. Pushed a new change with that removed.

faizkhan00:
ok! the rest of it looks good to me :)

faizkhan00:
hey @theuni , looking over your branch a bit, i'm wondering - - is it worthwhile to add implementations for these functions? I think that with the code we have in place it should be pretty safe (and i feel fairly confident) in bringing these to JNI (perhaps with additional tests as well) - thoughts there? 
```
    secp256k1_ec_pubkey_decompress
    secp256k1_ec_privkey_export
    secp256k1_ec_privkey_import
    secp256k1_ecdsa_sign_compact
    secp256k1_ecdsa_recover_compact
```
Perhaps also `randomize()` but that would entail synchronization safety, which could be a good thing to have anyway

theuni:
@faizkhan00 Sure. Since you're doing the bulk of the work here now, how about adding the above (though you don't have to of course, it's fine as-is), then opening a new PR for fresh review?

I only have 2 complaints left, I believe. It'd be great if you could address them as you work on the other changes:
- tabs->spaces. Lots of tabs snuck in with your changes, please make sure that whitespace matches the rest of the codebase.
- On the native side of the JNI functions, there are several places where ```ret``` is ignored after library calls. Even if those should realistically never fail, there should still be a way to inform the Java side that they did.

If you're OK with it, at this point I'll close this PR and invite you re-open a fresh one when you're ready. You're welcome to squash down any of my stuff however it makes sense to you.

faizkhan00:
@theuni Ype yep, that sounds like a plan. I'll likely have time to work on this soon/tonight-ish, but I'll need to look into synch locking a bit more before committing anything- Sipa mentioned some really important failure modes that just got documented in the 'assumptions' thread, and I want to make sure we're good there. 

- Tabbing: Yup I noticed this a bit while working and its annoying, good to hear spaces is preferred as that is how my editor is setup so should be no problem
- Ret: Yea this is an interesting one, I think handling this will make the Java side more of a proper native library (rather than a easy-to-use libsecp256k1) so I definitely see where that is going and will find some way to pass that through to the Java side.

Lets hold on closing the PR for the moment, I'll get those nits and funcs above addressed, and then when I open the new one, lets say we close this one , ?

Squashing OK

apoelstra:
Hi @faizkhan00,

For the record, I think pulling in Rust devtool dependencies to libsecp256k1 is not appropriate for the sake of JNI bindings; it's a lot of extra complexity and it feels weird to have this layer of indirection. When we talked on IRC I did not realize you were working on code to be merged into libsecp256k1 itself.

Andrew

sipa:
Going to close this as outdated. Anyone feel free to take up the work of a testable JNI/whateverjava interface.

PR #97, opened 2014-11-08
Goto3

iangfc:
Removed 3 gotos.  Their style makes it hard for readers to trace, and maintainers can make mistakes too easily.

(Also included is fix to configure to allow homebrew, another MacOSX package manager, to provide the needed libraries.  Homebrew uses /usr/local rather than /opt/local which is the domain of macports.)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -38,8 +38,8 @@ esac
 
 case $host_os in
   darwin*)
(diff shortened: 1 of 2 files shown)


sipa:
Closing in favor of #213

PR #112, opened 2014-11-20
Portability fixups

luke-jr:


sipa:
@luke-jr Is this still needed?

luke-jr:
Looks like just the pkg-config stuff is useful now. Rebased

sipa:
@theuni Can you have a look?

theuni:
@luke-jr If we're going to fall back to AC_CHECK_HEADER/AC_CHECK_LIB after the pkg-config check for openssl has failed, we'll need to do a TRY_LINK on it before trusting it. Otherwise, a system lib could be found on a cross build where no pkg-config'd openssl exists, and the final link would fail.

Now that I think about it though, may as well just do a TRY_LINK after pkg-config _and_ the fallback, since pkg-config could feasibly find a system lib as well.

Other than that, looks good.

sipa:
@luke-jr care to have a look at @theuni's suggestion?

luke-jr:
If AC_* ever pick up the wrong library, that seems a bug in autotools?

theuni:
Yea, never mind that suggestion. Lgtm.

PR #164, opened 2014-12-14
Optimize _double_var

peterdettman:
- Save 1 _fe_negate
- Prefer operations on local variables
- Reorder code somewhat to group inline function calls together

I am definitely looking for others to test this one. I measure up to 1% improvement for bench_verify, but of the three things I listed above, the "reorder code" measures as the most significant, so it's possible I have only optimized for my machine/compiler somehow.

### DIFF
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -231,26 +231,33 @@ static void secp256k1_gej_double_var(secp256k1_gej_t *r, const secp256k1_gej_t *
         return;
     }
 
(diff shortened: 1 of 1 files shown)


sipa:
This results in a measurable 0.2% slowdown for me...

peterdettman:
@sipa I'll re-test when I get a chance.

peterdettman:
@sipa Well, I still see a difference (bench_verify), but it's smaller now, ~0.2us/55us. Since it's inconsistent at best... closed!

PR #167, opened 2014-12-17
Tighter verification bounds in mul/sqr

peterdettman:
I reviewed the VERIFY_BITS bounds-checking for 32bit and 64bit mul/sqr implementations, finding various places where stricter bounds can be specified (I found no cases of over-strictness).


real-or-random:
This is ages old, and I agree with @sipa that there's no point for this in the current code. (And 6 years later we're more confident in the current code.  :) There's #811 and #815 of course, but these are separate efforts.

Closing this. Feel free to object.

PR #177, opened 2014-12-31
Use secp256k1_fe_equal_var in secp256k1_fe_sqrt_var.

gmaxwell:
In theory this should be faster, since secp256k1_fe_equal_var is able to
 shortcut the normalization.  On x86_64 the improvement appears to be in
 the noise for me.  At least it makes the code cleaner.

### DIFF
--- a/src/field_10x26_impl.h
+++ b/src/field_10x26_impl.h
@@ -31,6 +31,7 @@ static void secp256k1_fe_verify(const secp256k1_fe_t *a) {
     r &= (d[8] <= 0x3FFFFFFUL * m);
     r &= (d[9] <= 0x03FFFFFUL * m);
     r &= (a->magnitude >= 0);
+    r &= (a->magnitude <= 32);
     if (a->normalized) {
         r &= (a->magnitude <= 1);
         if (r && (d[9] == 0x03FFFFFUL)) {
--- a/src/field_5x52_impl.h
+++ b/src/field_5x52_impl.h
@@ -43,6 +43,7 @@ static void secp256k1_fe_verify(const secp256k1_fe_t *a) {
     r &= (d[3] <= 0xFFFFFFFFFFFFFULL * m);
     r &= (d[4] <= 0x0FFFFFFFFFFFFULL * m);
     r &= (a->magnitude >= 0);
+    r &= (a->magnitude <= 2048);
     if (a->normalized) {
         r &= (a->magnitude <= 1);
         if (r && (d[4] == 0x0FFFFFFFFFFFFULL) && ((d[3] & d[2] & d[1]) == 0xFFFFFFFFFFFFFULL)) {
--- a/src/field_impl.h
+++ b/src/field_impl.h
@@ -135,10 +135,7 @@ static int secp256k1_fe_sqrt_var(secp256k1_fe_t *r, const secp256k1_fe_t *a) {
     /* Check that a square root was actually calculated */
 
     secp256k1_fe_sqr(&t1, r);
-    secp256k1_fe_negate(&t1, &t1, 1);
-    secp256k1_fe_add(&t1, a);
-    secp256k1_fe_normalize_var(&t1);
-    return secp256k1_fe_is_zero(&t1);
+    return secp256k1_fe_equal_var(&t1, a);
 }
 
 static void secp256k1_fe_inv(secp256k1_fe_t *r, const secp256k1_fe_t *a) {


peterdettman:
Actually it looks like 'a' needs to be weakly normalized; the existing code was wrong.

gmaxwell:
Can we construct an example where secp256k1_ge_set_xo_var would be incorrect? 

Doesn't the explicit magnitude tracking in VERIFY check that the sequence of operations executed in the tests could not have produced an internal overflow for any values? I thought it was conservative. What case are we missing?

sipa:
Presumable every actual code path leads to a weakly normalized a in that
place.

peterdettman:
As things stand, _equal_var contract is not enforced for second argument.

sipa:
So, the current implied contract is that field elements have at most magnitude 8, which is verified in all field element operations.

The fe_sqrt routine however accepts a field element a (which, according to the contract can have magnitude up to 8), and is added to t1 (which has magnitude 2), so the addition could result in a field element with magnitude 10, which is not allowed. In practice, all existing code paths are _guaranteed_ (because the veirifcation of magnitudes is static and conservative) to never result in a having magnitude 8.

So the 'bug' could be fixed by just declaring that fe_sqrt takes an argument with magnitude at most 6 (which is already the case in practice).

sipa:
In any case, ACK.

sipa:
Actually, no. fe_equal_var requires its arguments to be weakly normalized, but fe_sqrt does not. I don't understand why the tests don't fail on this.

gmaxwell:
I am really not following this.

The input to the sqrt, a, is not normalized in our code (it's the result of an addition with 7).

The result of the sqrt is squared and as the output of a squaring it is normalized.

secp256k1_fe_equal_var  negates the squared number, which is fine. Then adds a. Which is fine.

Then secp256k1_fe_normalizes_to_zero does the minimum amount of normalization needed to check the result. Whats violated?

sipa:
Nothing is semantically violated.

The only thing that is missing is a contract that stipulates that the
inputs to sqrt are at most magnitude 1 and magnitude 6.

All current code obeys that rule, and given sufficient code coverage there
is an absolute guarantee that no current code path will violate that
requirement.

Either sqrt does a normalization on the inputs, or it gains a comment
saying that the inputs have stronger than usual normalization requirements.
That's all that's missing.

gmaxwell:
ah! okay.  Thanks, I was at a loss in part because it isn't clear to me what the input requirements are in various functions (other that to the extent that they're checked), and I thought you were saying that the change would actually yield a wrong result. 

I assume by "inputs to sqrt" you mean inputs to "fe_equal_var".

ABISprotocol:
Hello, I wanted to leave this brief note here. I'm not sure if it will be helpful or not, but perhaps it may provide something of note or context in this discussion.
https://www.w3.org/Bugs/Public/show_bug.cgi?id=25618#c60
this could probably be updated...
http://ianix.com/pub/curve25519-deployment.html
ok, I'm going away now
thx

gmaxwell:
@peterdettman @sipa After reviewing all the code I do not see any error in the magnitude requirements, or a failure to enforce (except one, see below).  I did notice that fe_verify imposed no upper limit on the magnitude at all, which is incorrect, so I added a commit with a limit.

I concluded that the input to the square can be up to magnitude 8 (it's only ever an input to sqr(), to add() with another input of magnitude 1, and the equal). secp256k1_fe_equal_var()'s first input must be magnitude 1 and this is checked by the negate inside it, which always run, and the second input can be magnitude whatever (well, up to 31, which is where the add would overflow; which (with my second commit) is tested by the fe_verify inside the add, which is always executed).

The above is based on an understanding that fe_normalizes_to_zero_var works correctly for any magnitude that fits in the field. If that isn't true, someone should rescue me from a grave misunderstanding how how all this works, and we should add a VERIFY test to the function.

As far as enforcing contracts goes, what I think we generally want to do is perform the magnitude tests in verifies at the lowest possible level. E.g. nowhere but inside fe_verify or in special operations like the field specific fe_mul and fe_sqr that have their own internal magnitude requirements. Keeping the tests at the lowest levels make them more likely to be correct regardless of what field we're running on, and they're also where the magnitudes must be updated. HOWEVER, if an execution path of higher level code is non-deterministic, we may want to bubble up the worst case normalization path requirement and enforce it unconditionally. We have some code paths that execute with probability less than 2^-128, so we would never know if those paths overflow unless we manually bubble up the requirement.

Bubbling up presents a problem that the low level requirements differ based on the field in use, which means layering violations (e.g. explicit magnitude tests inside group logic), which it would be preferable to avoid.

sipa:
I'm not sure about the "only check normalization constraints at the lowest level except in cases of non-deterministism" policy; it makes it harder to reason about different implementations, and where leeway exists. Especially when non-determinism would depend on lower-level implementations, it becomes less clear.

I think we're going towards exposing more normalization constraints upwards anyway, so I would propose to make this explicit, have the field code expose macros to verify constraints on arguments, use those in higher levels that have non-trivial constraints, and more importantly, also enforce worst-case checking (so for example if fe_sqrt lists a constraint that its a argument has at most magnitude 13, its magnitude is at runtime also actually changed to 13, resulting in tests that effectively check whether the higher-level constraints implicitly satisfy the lower-level ones).

sipa:
ACK in any case, this is an improvement.

sipa:
I don't see any performance improvement.

gmaxwell:
Read the commit message.

sipa:
I saw it. I'm just confirming that I also don't see a performance improvement :)

PR #179, opened 2015-01-06
secp256k1 benchmarking programs can now accept parameters

midnightmagic:
-c will vary the loop counter
-i will vary the iterations per-loop
-w will change the WINDOW_G table size setting at runtime (primary reason for patch)

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -50,7 +50,8 @@ extern "C" {
  *  It cannot run in parallel with any other functions, but once
  *  secp256k1_start() returns, all other functions are thread-safe.
  */
-void secp256k1_start(unsigned int flags);
+void secp256k1_start(unsigned int flags, unsigned int tablesize);
+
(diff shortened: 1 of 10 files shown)


sipa:
I like the changes here, but am not too happy with the implementation.

There has been talk about getting rid of the globals, and explicitly initializating a context object. I think I'll go through with those changes first, and then get back to this.

midnightmagic:
Anything I can help with, let me know.

sipa:
@midnightmagic Will do; I'd like to get #208 in first.

sipa:
Needs pretty significant rewriting now, but should be cleaner :)

gmaxwell:
This would be good to see done, but I don't see a reason to keep it open when no one is currently working on it and the PR is so thoroughly out of date.  Closing without prejudice, anyone who wants to pick this up should feel free to do so.

Issue #186, opened 2015-01-11
Scalar / Point blinding


gmaxwell:
Constant time, constant memory access operations for secret data should make additional blinding unnecessary. But at least init-time static blinding can be done basically at no cost by baking it into the tables; and may afford some protection against EMI/power side-channels or timing in cases where a compiler has undermined the constant time operation.

We already do the novel unknown discrete log blinding. I'd held off previously adding more because we didn't have a hash function (e.g. to generate random unknown points for each table row). But we do now. Peter has also expressed some interest here.

PR #190, opened 2015-01-20
Add scalar blinding for ecmult_gen()

gmaxwell:
This computes (n-b)G + bG with random value b, in place of nG.

This is intended to reduce exposure to potential power/EMI sidechannels
 during signing and pubkey generation by blinding the secret value with
 another value which is hopefully unknown to the attacker.

It may not be very helpful if the attacker is able to observe the setup
 or if even the scalar addition has an unacceptable leak, but it has low
 overhead in any case and the security should be purely additive on top
 of the existing defenses against sidechannels.

For comments right now, there is no interface presented for users to pass in a seed. We should probably change the API to use a context pointer created by start, and could change start to take a random uchar[32] at that time.

### DIFF
--- a/src/ecmult_gen_impl.h
+++ b/src/ecmult_gen_impl.h
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (c) 2013, 2014 Pieter Wuille                             *
+ * Copyright (c) 2013, 2014, 2015 Pieter Wuille, Gregory Maxwell      *
  * Distributed under the MIT software license, see the accompanying   *
  * file COPYING or http://www.opensource.org/licenses/mit-license.php.*
  **********************************************************************/
@@ -10,6 +10,7 @@
 #include "scalar.h"
 #include "group.h"
 #include "ecmult_gen.h"
+#include "hash_impl.h"
 
 static void secp256k1_ecmult_gen_context_init(secp256k1_ecmult_gen_context_t *ctx) {
     ctx->prec = NULL;
@@ -74,6 +75,7 @@ static void secp256k1_ecmult_gen_context_build(secp256k1_ecmult_gen_context_t *c
             secp256k1_ge_to_storage(&(*ctx->prec)[j][i], &prec[j*16 + i]);
         }
     }
+    secp256k1_ecmult_gen_blind(ctx, NULL);
 }
 
 static int secp256k1_ecmult_gen_context_is_built(const secp256k1_ecmult_gen_context_t* ctx) {
@@ -87,24 +89,31 @@ static void secp256k1_ecmult_gen_context_clone(secp256k1_ecmult_gen_context_t *d
     } else {
         dst->prec = (secp256k1_ge_storage_t (*)[64][16])checked_malloc(sizeof(*dst->prec));
         memcpy(dst->prec, src->prec, sizeof(*dst->prec));
+        dst->initial = src->initial;
+        dst->blind = src->blind;
     }
 }
 
 static void secp256k1_ecmult_gen_context_clear(secp256k1_ecmult_gen_context_t *ctx) {
     free(ctx->prec);
+    secp256k1_scalar_clear(&ctx->blind);
+    secp256k1_gej_clear(&ctx->initial);
     ctx->prec = NULL;
 }
 
 static void secp256k1_ecmult_gen(const secp256k1_ecmult_gen_context_t *ctx, secp256k1_gej_t *r, const secp256k1_scalar_t *gn) {
     secp256k1_ge_t add;
     secp256k1_ge_storage_t adds;
+    secp256k1_scalar_t gnb;
     int bits;
     int i, j;
     memset(&adds, 0, sizeof(adds));
-    secp256k1_gej_set_infinity(r);
+    *r = ctx->initial;
+    /* Blind scalar/point multiplication by computing (n-b)G + bG instead of nG. */
+    secp256k1_scalar_add(&gnb, gn, &ctx->blind);
     add.infinity = 0;
     for (j = 0; j < 64; j++) {
-        bits = secp256k1_scalar_get_bits(gn, j * 4, 4);
+        bits = secp256k1_scalar_get_bits(&gnb, j * 4, 4);
         for (i = 0; i < 16; i++) {
             /** This uses a conditional move to avoid any secret data in array indexes.
              *   _Any_ use of secret indexes has been demonstrated to result in timing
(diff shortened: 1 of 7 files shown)


gmaxwell:
This addresses issue #186.

sipa:
I'd like to see this rebased on top of #208.

sipa:
Please rebase (I can do so, if you want to).

gmaxwell:
Updated.  If this looks good I'll add another commit that adds an additional re-projection with H(gnb) in ecmult_gen().

dcousens:
To clarify, the point of difference here is that: `n-b` is hopefully safer than `nG` because the point scalar-multiplication has known side channels, where scalar-subtraction does not?

And further, that the subsequent point scalar-multiplication `(n-b)G` is safer than `nG` because `b` is unknown?

gmaxwell:
Multipliers are much more likely to have sidechannels in general owing to their greater complexity and internal architecture. They have have, on various real systems, even have had data dependent timing, fully visible from software (though I am not aware of this being the case case for any current common desktop hardware; I've experienced this on ppc440, arm7tdmi, and some other processors).

More importantly, the blinding operation is tiny and very fast, compared to a huge number of multiply operations each of which could leak. In practice it would plausibly be harder to exploit any leak from the blinding because its so fast... but ultimately if additions leak, well, they leak and the library will thus leak regardless of what we do.

b is (hopefully) unknown, so n has uniform probability if you only learn (n-b).  If you can observe _all_ behavior flawlessly then the protection may not help, but each randomization is cumulative and if you miss only one (e.g. the first one) you're out of luck if your only leak is the multipliers.

It's not a guarantee, but we practice defense in depth, and this change does not produce a measurable slowdown. (The next commit will be, but it will be a tiny one).

dcousens:
Thanks for the explanation @gmaxwell.

dcousens:
> b is (hopefully) unknown, so n has uniform probability if you only learn (n-b). If you can observe all behavior flawlessly then the protection may not help, but each randomization is cumulative and if you miss only one (e.g. the first one) you're out of luck if your only leak is the multipliers.

But if you learn `(n - b)`, then use the same side channel you are trying to eliminate with `bG`, couldn't you then just subtract `(n - b)G + -bG` to discover `n`?

gmaxwell:
It doesn't compute bG every time (that would be slow and obviously have more leaks). It computes bG at start, and forward chains any time b is changed.

dcousens:
Ah!

Thanks for clearing that up :)

gmaxwell:
@sipa API question-- instead of a explicit reset, which is a pretty obscure option which I doubt wouldn't ever be used except for testing infrastructure, what would you think of me using a null seed pointer to reset?

sipa:
Oh yes, using NULL seems fine.

gmaxwell:
@sipa updated for the api change.

gmaxwell:
@sipa another behavior question:  'default seed`, zeros or the address of  one of the libraries functions?  (Advantage being that the state ends up somewhat unpredictable (esp on hosts with ASLR), even if the caller never call randomize; disadvantage is that behavior differs from host to host even if randomize isn't called which could complicate debugging)

sipa:
I don't think that using "attempted randomness" makes sense if no
guarantees can be made. It's not something to rely on, so let's just use
zeroes.

dcousens:
+1 to all zeros.  Uninitialized memory or externally 'sourced' memory may lead to unexpected consequences.

sipa:
Perhaps you can add a test to see whether the Z coordinate of the result
differs after signing with seeded context?

gmaxwell:
@dcousens My suggestion was not uninitialized memory. Use of uninitialized memory would be undefined behavior and would permit the compiler to instead substitute formatting the user's disk.  The what I would have been completely safe but also of very little value (or I would have just done it).  :)

@sipa: sounds good. I think I still need to do scalar equals functions in the tests.

src/ecmult_gen_impl.h
```diff
@@ -1,5 +1,5 @@
 /**********************************************************************
- * Copyright (c) 2013, 2014 Pieter Wuille                             *
+ * Copyright (c) 2013, 2014, 2015 Pieter Wuille, Gregory Maxwell      *
```
sipa: About time.

src/ecmult_gen_impl.h
```diff
@@ -123,6 +132,50 @@ static void secp256k1_ecmult_gen(const secp256k1_ecmult_gen_context_t *ctx, secp
...
+    unsigned char nonce32[32];
+    secp256k1_rfc6979_hmac_sha256_t rng;
+    int retry;
+    if (!seed32) {
+        secp256k1_gej_set_ge(&ctx->initial, &secp256k1_ge_const_g);
+        secp256k1_gej_neg(&ctx->initial, &ctx->initial);
+        secp256k1_scalar_set_int(&ctx->blind, 1);
+    }
+    secp256k1_scalar_get_b32(nonce32, &ctx->blind);
+    /** Using a CSPRNG allows a simpler-failure free interface, avoids needing large amounts of random data,
```
sipa: Simpler than what alternative? I know what you mean, but it may be unclear to someone just reading the code.

src/ecmult_gen_impl.h
```diff
@@ -123,6 +132,50 @@ static void secp256k1_ecmult_gen(const secp256k1_ecmult_gen_context_t *ctx, secp
...
+     *  and guards against weak or adversarial seeds.
+     */
+    secp256k1_rfc6979_hmac_sha256_initialize(&rng, seed32 ? seed32 : nonce32, 32, nonce32, 32, NULL, 0);
```
sipa: How does this cause a reset if seed32=NULL is passed? The previous ctx->blind is used as PRNG seed.

src/tests.c
```diff
@@ -901,6 +901,27 @@ void ge_equals_ge(const secp256k1_ge_t *a, const secp256k1_ge_t *b) {
     CHECK(secp256k1_fe_equal_var(&b->y, &b->y));
 }
 
+int gej_equals_gej(const secp256k1_gej_t *a, const secp256k1_gej_t *b) {
```
sipa: Perhaps add a comment that this compares the actual representation of the gej (including the Z) coordinate, and not just its geometric meaning?

sipa:
ACK apart from nits. I see no measurable slowdown (perhaps even a slight speedup, wtf?).

gmaxwell:
Updated.

src/ecmult_gen_impl.h
```diff
@@ -123,6 +132,53 @@ static void secp256k1_ecmult_gen(const secp256k1_ecmult_gen_context_t *ctx, secp
...
+    if (!seed32) {
+        /* When seed is NULL, reset the initial point and blinding value. */
+        secp256k1_gej_set_ge(&ctx->initial, &secp256k1_ge_const_g);
+        secp256k1_gej_neg(&ctx->initial, &ctx->initial);
+        secp256k1_scalar_set_int(&ctx->blind, 1);
+    }
+    /* The prior blinding value (if not reset) is chained forward by including it in the hash. */
+    secp256k1_scalar_get_b32(nonce32, &ctx->blind);
+    /** Using a CSPRNG allows a failure free interface, avoids needing large amounts of random data,
+     *   and guards against weak or adversarial seeds.  This is a simpler and safer interface then
```
sipa: s/then/than/

src/ecmult_gen_impl.h
```diff
@@ -123,6 +132,53 @@ static void secp256k1_ecmult_gen(const secp256k1_ecmult_gen_context_t *ctx, secp
...
+static void secp256k1_ecmult_gen_blind(secp256k1_ecmult_gen_context_t *ctx, const unsigned char *seed32) {
+    secp256k1_scalar_t b;
+    secp256k1_gej_t gb;
+    secp256k1_fe_t s;
+    unsigned char nonce32[32];
+    secp256k1_rfc6979_hmac_sha256_t rng;
+    int retry;
+    if (!seed32) {
+        /* When seed is NULL, reset the initial point and blinding value. */
+        secp256k1_gej_set_ge(&ctx->initial, &secp256k1_ge_const_g);
```
sipa: This, and the next line are actually redundant. (I'm fine with keeping them for clarity, though).

sipa:
ACK

PR #191, opened 2015-01-21
Reduce the amount of runtime-initialized constants

sipa:
Instead of having per-subsystem context that gets initialized at runtime, use some macros to write them directly as C static consts. Less code, faster, more readable.

### DIFF
--- a/src/bench_inv.c
+++ b/src/bench_inv.c
@@ -42,11 +42,7 @@ void bench_inv(void* arg) {
 }
 
 int main(void) {
-    secp256k1_ge_start();
-
     bench_inv_t data;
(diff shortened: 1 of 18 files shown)


gmaxwell:
Dogmatic macro hygiene nit: Macro variables should always be parenthesized lest someone change whats passed in and cause a really-surprising-parsing incident.

(I probably wouldn't comment for this but I know eventually I'll run some static analysis tool on the codebase that whines because of CERT PRE01-C or MIRSA-C rule 96)

sipa:
@gmaxwell Nit addressed.

gmaxwell:
ACK

PR #192, opened 2015-01-23
C89 nits and dead code removal.

gmaxwell:


mb300sd:
+1

Discovered these independently while trying to get it to compile on an ARM microcontroller.

PR #193, opened 2015-01-23
Convert field code to strict C89 (+ long long, +__int128)

gmaxwell:
Convert field code to strict C89 (+ long long, +__int128)

This makes the software more portable to embedded systems
 and static analysis tools.

Sadly, it can't result in identical binaries because C99 mixed
 declarations seem to make GCC emit superfluous stack-pointer
 updates. The compiler is also somewhat dependent on the
 declaration order.

( Builds on pull req #192 )

### DIFF
--- a/src/ecdsa.h
+++ b/src/ecdsa.h
@@ -10,9 +10,6 @@
 #include "scalar.h"
 #include "group.h"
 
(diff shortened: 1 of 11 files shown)


sipa:
ACK

PR #194, opened 2015-01-23
Shouldn't this be #ifdef?

mb300sd:
Right now I have to define HAVE_BUILTIN_EXPECT to get it to compile with IAR EWARM where __builtin_expect() dosen't exist, shouldn't it be the opposite?


gmaxwell:
ACK

PR #199, opened 2015-01-25
Introduce compact storage types

sipa:
Introduce packed data types for field and group elements, which are smaller, guaranteed to be normalized and non-infinite, and are comparable in memory.

This speeds up signing around 1-2%, and verification around 0.1-0.2%, and reduces memory footprint.


gmaxwell:
ACK.

PR #205, opened 2015-02-02
Include a comment on secp256k1_ecdsa_sign explaining low-s.

gmaxwell:
This was suggested by DavidEGrayson.

PR #206, opened 2015-02-02
Some comments explaining some of the constants in the code.

gmaxwell:


droark:
Thanks, Greg. I've been doing a review of libsecp256k1 in my free time. Comments like these and the one in pull request #205 are a huge help. :)

src/ecdsa_impl.h
```diff
@@ -15,11 +15,28 @@
 #include "ecmult_gen.h"
 #include "ecdsa.h"
 
+/** Group order for secp256k1 defined as 'n' in "Standards for Efficient Cryptography" (SEC2) 2.7.1
```
sipa: This prints p, not n.

sipa:
There are several copies of p and n in different places of the code. Maybe it's useful to only explain them once, and just put references in other places ("see the definition of secp256k1_scalar_const_order in file X").

sipa:
ACK

PR #207, opened 2015-02-02
Benchmarks for all internal operations

sipa:


gmaxwell:
Needs some endomorphism ifdefs.

gmaxwell:
The switching between ns and microseconds is potentially confusing. Maybe always print ns?  (e.g. makes it more complicated to make dumb graphing scripts, or a risk that you'll misread something)

gmaxwell:
secp256k1_ecmult_wnaf ?

gmaxwell:
secp256k1_ecmult_wnaf ?

sipa:
Updated. Everything is printed in us, in decimal notation, with at least 3 significant digits. Also, added a field_sqr and ecmult_wnaf benchmark.

gmaxwell:
ACK.

PR #174, opened 2014-12-29
Co-Z + effective affine precomputation + tests

sipa:
This builds on top of #171 but reworks the group element tests, making them much more extensive (and faster).

PR #210, opened 2015-02-11
Effective affine precomputation (by Peter Dettman)

sipa:
Significantly refactored version of Peter Dettman's effective affine precomputation tricks.

See #174 and #41 for old discussion.

### DIFF
--- a/src/group.h
+++ b/src/group.h
@@ -62,6 +62,17 @@ static void secp256k1_ge_set_gej(secp256k1_ge_t *r, secp256k1_gej_t *a);
 /** Set a batch of group elements equal to the inputs given in jacobian coordinates */
 static void secp256k1_ge_set_all_gej_var(size_t len, secp256k1_ge_t *r, const secp256k1_gej_t *a);
 
+/** Set a batch of group elements equal to the inputs given in jacobian
+ *  coordinates (with known z-ratios). zr must contain the known z-ratios such
+ *  that mul(a[i].z, zr[i+1]) == a[i+1].z. zr[0] is ignored. */
+static void secp256k1_ge_set_table_gej_var(size_t len, secp256k1_ge_t *r, const secp256k1_gej_t *a, const secp256k1_fe_t *zr);
+
+/** Bring a batch inputs given in jacobian coordinates (with known z-ratios) to
+ *  the same global z "denominator". zr must contain the known z-ratios such
+ *  that mul(a[i].z, zr[i+1]) == a[i+1].z. zr[0] is ignored. The x and y
+ *  coordinates of the result are stored in r, the common z coordinate is
+ *  stored in globalz. */
+static void secp256k1_ge_globalz_set_table_gej(size_t len, secp256k1_ge_t *r, secp256k1_fe_t *globalz, const secp256k1_gej_t *a, const secp256k1_fe_t *zr);
 
 /** Set a group element (jacobian) equal to the point at infinity. */
 static void secp256k1_gej_set_infinity(secp256k1_gej_t *r);
@@ -81,11 +92,11 @@ static void secp256k1_gej_neg(secp256k1_gej_t *r, const secp256k1_gej_t *a);
 /** Check whether a group element is the point at infinity. */
 static int secp256k1_gej_is_infinity(const secp256k1_gej_t *a);
 
-/** Set r equal to the double of a. */
-static void secp256k1_gej_double_var(secp256k1_gej_t *r, const secp256k1_gej_t *a);
+/** Set r equal to the double of a. If rzr is not-NULL, r->z = a->z * *rzr (where infinity means an implicit z = 0). */
+static void secp256k1_gej_double_var(secp256k1_gej_t *r, const secp256k1_gej_t *a, secp256k1_fe_t *rzr);
 
-/** Set r equal to the sum of a and b. */
-static void secp256k1_gej_add_var(secp256k1_gej_t *r, const secp256k1_gej_t *a, const secp256k1_gej_t *b);
+/** Set r equal to the sum of a and b. If rzr is non-NULL, r->z = a->z * *rzr (a cannot be infinity in that case). */
+static void secp256k1_gej_add_var(secp256k1_gej_t *r, const secp256k1_gej_t *a, const secp256k1_gej_t *b, secp256k1_fe_t *rzr);
 
 /** Set r equal to the sum of a and b (with b given in affine coordinates, and not infinity). */
 static void secp256k1_gej_add_ge(secp256k1_gej_t *r, const secp256k1_gej_t *a, const secp256k1_ge_t *b);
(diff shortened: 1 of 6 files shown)


sipa:
Rebased.

sipa:
This seems to give a 4.5-6.5% speedup for verification (without and with GLV).

sipa:
@gmaxwell @apoelstra: feel like reviewing this?

apoelstra:
Sure. I will need to do the prereading.

sipa:
Rebased.

gmaxwell:
I believe this needs a citation to https://eprint.iacr.org/2008/051  (you should check to see that the technique agrees).

sipa:
Rebased.

sipa:
Rebased.

sipa:
@apoelstra Feel free to ask any questions.

peterdettman:
@apoelstra Happy to answer questions also.

apoelstra:
Hi guys, sorry for the delay. I've been caught up in schoolwork as I transition departments and haven't had enough spare cycles at once. This weekend is hopeful.

Am I correct that there is an explicit description of the new code in https://eprint.iacr.org/2008/051 (which I haven't read)? Or should I use the algebra and motivations from the Co-Z paper (which I have read in some detail)?

sipa:
That paper is about using Co-Z for speeding up precomputations. It has
nothing to do with the effective affine trick (which can, however, be used
ad an alternative to co-z for the precomputation).

IIRC the effective affine trick is novel.

peterdettman:
@apoelstra Further to @sipa's comments, "Co-Z" and "Effective affine" are independent techniques. Possibly that was somewhat obscured during the initial development, but with hindsight and a subsequent refactoring of the code (thanks to @sipa) it is now clear.

I would like to clarify my position on claims of novelty. Since here we wish only to deal with "Effective affine", please also see https://github.com/bitcoin/secp256k1/pull/211 where I will shortly add a similar statement regarding "Co-Z".

*I claim independent discovery* of "the effective affine trick", the essence of which is to perform the main ladder of a scalar multiplication on an isomorphism of the original curve, which avoids the cost of an inversion in the field, whilst still admitting the use of mixed addition with precomputed points. There are some minor corollary techniques reflected in the implementation, to wit: i) applying the trick to the pre-computation itself, and ii) applying the trick in the context of a Strauss-Shamir ladder (with one point fixed).

*I think it would be premature to claim novelty at this time*, for (at least) the following reasons. Although I read quite widely in the relevant literature and know of no previous description or implementation of the technique, I have made only moderate efforts in a directed search. The technique is only a small step away from other commonly-understood techniques. I have not yet directly solicited the opinion of academic researchers, who are likely to have a more complete (e.g. paywalled) and up-to-date view of the literature. Although I have described the techniques on e.g. the IETF CFRG mailing list and in a few private communications, it has never been in the context of establishing novelty. Finally, I have not conducted or requested a prior art search with any patent office.

sipa:
Thanks for the summary!

sipa:
The isomorphism involved can be compactly written as:
* (in affine coordinates): (a,b) + (c,d) = (e,f) <=> (a*t^2,b*t^3) +
(c*t^2,d*t^3) = (e*t^2,f*t^3).
* (in Jacobian coordinates): (a,b,c) + (d,e,f) = (g,h,i) <=> (a,b,c*t) +
(d,e,f*t) = (g,h,i*t).

This is used in two different ways:
* When adding a number of Jacobian points together, which are all known to
have the same Z coordinate (called global Z in the source code), one can
use the formulae for affine addition (temporarily assuming that global Z
coordinate is 1), and then later multiply the result's Z coordinate with
the global Z.
* Addition with a Jacobian point of which the inverse of its Z coordinate
is known, can be done using a minor modification to the affine formulae.

apoelstra:
Awesome, thanks for the summary @peterdettman and for the algebra @sipa. I'll check over the code today and tomorrow.

sipa:
@apoelstra In particular, feel free to comment about explanations given in comments (not enough, confusing, ...).

src/group.h
```diff
@@ -82,10 +93,10 @@ static void secp256k1_gej_neg(secp256k1_gej_t *r, const secp256k1_gej_t *a);
 static int secp256k1_gej_is_infinity(const secp256k1_gej_t *a);
 
 /** Set r equal to the double of a. */
-static void secp256k1_gej_double_var(secp256k1_gej_t *r, const secp256k1_gej_t *a);
+static void secp256k1_gej_double_var(secp256k1_gej_t *r, const secp256k1_gej_t *a, secp256k1_fe_t *rzr);
```
apoelstra: I think the doccomment needs to be updated to reflect what happens to `rzr`. In particular the behaviour when `a` is infinity (`*rzr` being set to one) is not necessarily obvious.

apoelstra: I think the doccomment needs to be updated to reflect what happens to `rzr`. ("Is it `r.z/a.z` or its recipricol?" is not obvious just from the name. I also don't think the behaviour for infinity is obvious since then rzr ought be be "0/0".)

src/group.h
```diff
@@ -82,10 +93,10 @@ static void secp256k1_gej_neg(secp256k1_gej_t *r, const secp256k1_gej_t *a);
...
 
 /** Set r equal to the sum of a and b. */
-static void secp256k1_gej_add_var(secp256k1_gej_t *r, const secp256k1_gej_t *a, const secp256k1_gej_t *b);
+static void secp256k1_gej_add_var(secp256k1_gej_t *r, const secp256k1_gej_t *a, const secp256k1_gej_t *b, secp256k1_fe_t *rzr);
```
apoelstra: Ditto here.

src/group.h
```diff
@@ -93,11 +104,14 @@ static void secp256k1_gej_add_ge(secp256k1_gej_t *r, const secp256k1_gej_t *a, c
 /** Set r equal to the sum of a and b (with b given in affine coordinates). This is more efficient
     than secp256k1_gej_add_var. It is identical to secp256k1_gej_add_ge but without constant-time
     guarantee, and b is allowed to be infinity. */
-static void secp256k1_gej_add_ge_var(secp256k1_gej_t *r, const secp256k1_gej_t *a, const secp256k1_ge_t *b);
+static void secp256k1_gej_add_ge_var(secp256k1_gej_t *r, const secp256k1_gej_t *a, const secp256k1_ge_t *b, secp256k1_fe_t *rzr);
```
apoelstra: and here.

apoelstra:
Tested ACK. Confirmed that the algebra made sense, comments were sensible, that it was implemented clearly and correctly. Ran unit tests on both commits under valgrind (no leaks or errors) and kcov ("full coverage" as much as that tool can tell), also ran the unit tests for `rust-secp256k1` on them.

This is a really elegant and clever improvement. Thanks a ton for your contribution @peterdettman . Sorry all for the delays in my review.

sipa:
Updated the comments, as requested by @apoelstra.

sipa:
@apoelstra Thanks for the review!

peterdettman:
@apoelstra Thanks for your review, and for your kind words.

PR #211, opened 2015-02-11
Co-Z based precomputation (by Peter Dettman)

sipa:
Refactored version of Co-Z code from #41 and #174.

Builds on top of #210.

### DIFF
--- a/src/group.h
+++ b/src/group.h
@@ -99,7 +99,7 @@ static void secp256k1_gej_add_ge(secp256k1_gej_t *r, const secp256k1_gej_t *a, c
 /** Set r equal to the sum of a and b (with b given in affine coordinates). This is more efficient
     than secp256k1_gej_add_var. It is identical to secp256k1_gej_add_ge but without constant-time
     guarantee, and b is allowed to be infinity. */
-static void secp256k1_gej_add_ge_var(secp256k1_gej_t *r, const secp256k1_gej_t *a, const secp256k1_ge_t *b);
+static void secp256k1_gej_add_ge_var(secp256k1_gej_t *r, const secp256k1_gej_t *a, const secp256k1_ge_t *b, secp256k1_fe_t *rzr);
 
 /** Get a 131-character hex representation of a point. */
 static void secp256k1_gej_get_hex(char *r131, const secp256k1_gej_t *a);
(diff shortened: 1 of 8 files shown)


sipa:
Rebased.

sipa:
This gives a 2.0-2.3% speedup for verification on top of #210.

sipa:
Rebased.

peterdettman:
A brief recap regarding novelty of this idea. Short version: *this is not novel* (but it's still pretty cool).

Boring version: I learnt about the Co-Z formulae from several papers, of which http://joye.site88.net/papers/GJMRV11regpm.pdf may be taken as representative, and which in turn usually referenced papers of Meloni. At some point I realized it could be applied usefully to the pre-computation of a table of odd multiples, and duly implemented a rough version. Actually it was fairly fast, so I became curious what was the fastest known method for that pre-computation. A literature research turned up https://eprint.iacr.org/2008/051, which upon further investigation, turned out to describe essentially the same approach (explicitly mentioning Meloni's formulae), giving one scheme with the same cost as ours, and a second one that improved on that further. I subsequently implemented the second scheme, although that modification is not yet in any PR (and the effect is small).

src/group.h
```diff
@@ -31,6 +31,16 @@ typedef struct {
...
 #define SECP256K1_GEJ_CONST_INFINITY {SECP256K1_FE_CONST(0, 0, 0, 0, 0, 0, 0, 0), SECP256K1_FE_CONST(0, 0, 0, 0, 0, 0, 0, 0), SECP256K1_FE_CONST(0, 0, 0, 0, 0, 0, 0, 0), 1}
 
+#ifdef USE_COZ
+/** A group element of the secp256k1 curve, with an implicit z coordinate (and infinity flag).
+ *  An instance of secp256k1_coz_t is always "co-z" with some instance of secp256k1_gej_t, from
+ *  which it inherits its implied z coordinate and infinity flag. */
+typedef struct {
+    secp256k1_fe_t x; /* actual X: x/z^2 (z implied) */
+    secp256k1_fe_t y; /* actual Y: y/z^3 (z implied) */
+} secp256k1_coz_t;
```
apoelstra: Is there any benefit to adding an `#ifdef VERIFY` z coordinate, then `VERIFY_CHECK`ing in `secp256k1_coz_zaddu_var` that `ra` is actually co-Z with `b`?

sipa: I think that makes perfect sense.

sipa:
Added the consistency check that @apoelstra suggested.

gmaxwell:
Please also add an explicit citation to the Meloni publicaiton for the co-z formula (the cite is in Longa/Miri 2008)

sipa:
Rebased.

gmaxwell:
needs rebase

PR #212, opened 2015-02-13
Implement Schnorr signatures

sipa:


sipa:
Rebased.

sipa:
Rebased.

sipa:
TODO: use context to store the batch multipliers

dcousens:
This is awesome, as Schnorr signatures are dead easy to understand.
But, for what purpose have they been added?

sipa:
They have not been added yet :)

sipa:
Separated batch validation into a separate branch. This pull request should be usable. Question for reviewers:
* Enable Schnorr code by default, or hide behind a compile flag (and perhaps move to a separate include/.h file)?

sipa:
@gmaxwell Added multiparty signing API.

apoelstra:
I'm a bit concerned about using `secp256k1_nonce_function_default`. If you sign the same message in both ECDSA and Schnorr with the same nonce, it is possible to recover the secret key from the two signatures. So I think by default Schnorr should use a slightly different noncegen function.

src/schnorr_impl.h
```diff
@@ -0,0 +1,186 @@
...
+
+    secp256k1_ecmult_gen(ctx, &Rj, &n);
+    if (pubnonce) {
+        secp256k1_gej_add_ge(&Rj, &Rj, pubnonce);
+    }
+    secp256k1_ge_set_gej(&Ra, &Rj);
+    secp256k1_fe_normalize(&Ra.y);
+    if (secp256k1_fe_is_odd(&Ra.y)) {
+        secp256k1_scalar_negate(&n, &n);
+    }
```
apoelstra: Can you add a comment above this block explaining that forcing the parity of y allows batch verification (and is part of the definition of Schnorr signatures), at the cost of making inversion-free validation impossible?

sipa:
@apoelstra You're not being stupid, I think.

k is the secret nonce, r is R's x coordinate, m is the message, x is the private key, h = H(m || r). x and k are unknown to an attacker.

ECDSA: s_ecdsa * k = m + r * x
Schnorr: s_schnorr = k - h * x

Or:
ECDSA: k = (m + r * x) / s_ecdsa
Schnorr: k = s_schnorr + h * x

So:
(m + r * x) / s_ecdsa = s_schnorr + h * x

So:
x = (s_schnorr - m/s_ecdsa) / (r/s_ecdsa - h)

sipa:
Fixed the problem listed above by xoring the name of the signature algorithm into the message.

Added a rationale for the choice of full R validation and implicit even y.

sipa:
Rebased on top of #269 and passing the signature algorithm name into the PRNG, to avoid the issue @apoelstra found above.

sipa:
Rebased on top of new #269 and #282, and made compilation of Schnorr code optional (with API defined in a separate header).

@theuni Any comments on the module approach here?

apoelstra:
Valgrind shows a bunch of errors on the test binary with this code. You need `CFLAGS="-g -O0" ./configure` to see them.

apoelstra:
Almost there :) `test_schnorr_end_to_end` is never called (and as a consequence _sign and _recover aren't tested).

sipa:
Fixed.

apoelstra:
ACK

theuni:
@sipa Looks good to me. One nit though. IMO we should also add a define in the main header that lists which modules are available. Otherwise preprocessing would be necessary to determine if schnorr functionality could be #included.

Even easier than that would be a 3rd (pushing it, i know) header with those defines, to avoid hacking up the main interface header. As a bonus, we could also stick the current config in there so that features like USE_ENDOMORPHISM, USE_NUM_GMP, etc. could be known at the application level.

sipa:
@theuni Feel like writing a patch for that? I'm not sure I follow, but I
think it would be nice to have a clean extension mechanism (the two ECDH
implementations would also fit there).

sipa:
Hmm, I don't know about exposing implementation details in the headers. I
jist thought: if you enable the Schnorr module, the schnorr header file is
installed, otherwise it isn't... no preprocessing needed.

theuni:
@sipa There's no way to tell if it's installed without preprocessing though, you'd just "#include secp256k1_schnorr.h" and hope for the best.

I took a quick stab at it here: https://github.com/theuni/secp256k1/tree/212

If you decide to go this route, since the config details become part of the api, we'd need to namespace/cleanup their values.

Of course, you could also just take a subset of this and only expose the module details, leaving the rest as it is.

sipa:
Well, duh. But how is that different from doing an include #include <gmp.h>
and hoping that that libgmp is installed?

I would really like to avoid having the public API expose implementation
details. I'd rather just see them as two separate APIs that happen to be
implemented in the same library.

theuni:
@sipa Because when you 'apt-get install libgmp', you know you can "#include <gmp.h>". But if I apt-get install libsecp256k1, I have no idea what it contains. Using the existence of a file is shaky at best, as it'd be very easy to hit the system file in /usr/include when building against another or a cross build.

No problem if you don't want to expose implementation details, but I still think the user needs _some_ explicit way to know what optional apis are available.

sipa:
@theuni What would an application do with the knowledge that an extra API is available? If it needs it, the build will fail (at link time or at compile time), if it doesn't, it doesn't care...

theuni:
@sipa Just as libsecp256k1 falls back to internal implementations if gmp isn't usable, an application could do the same if libsecp256k1_foo feature isn't available.

But it's really not worth the argument, and certainly not worth holding up this PR.

sipa:
@theuni Right, there may be optional behaviour. But that's something you'd typically check at configure time, which can try to compile.

I guess the fundamental question I have is: why can you trust that the contents of a file on your disk corresponds with the library installed, when you can't trust the presence of the file corresponding to it.

sipa:
@theuni Ok, one other question: when you don't specify --enable-module-schnorr, the schnorr files don't get included in "make dist" I think. Is that something you'd typically want?

theuni:
@sipa all files needed to build all configs should be included in the dist tarball. From a quick check, that's currently working as expected:
```
./autogen.sh
./configure
make distdir # same as 'make dist' and extract tarball
cd libsecp256k1-0.1
./configure --enable-module-schnorr
make
```

apoelstra:
Before merge can you add bench_schnorr_verify to .gitignore?

apoelstra:
Also I think you need to change the last line of `travis.yml` so that the value of $SCHNORR actually affects configure.

sipa:
Rebased.

sipa:
@apoelstra Nits addressed.

apoelstra:
Thanks! Tested ACK.

jl777:
is there anyway to use schnorr combined pubkey in a standard p2sh script?
I assume the schnorr sigs are different so OP_CHECKSIG wont work, but if we can assume externally coordinated reveal on secret script, I am hoping there is a way to achieve this as a combined pubkey can be created, but I cant see how to get a script verifiable value before the signature. Maybe there is a way to cooperatively construct a combined privkey that can be used to sign?

voisine:
I believe schnor is planned for the first segwit upgrade. It will require a new checksig opcode and a soft fork.

jl777:
well a new schnorr_sig opcode would make this all too easy :)
that is what I am adding now but would be nice to get this on the bitcoin side too

sipa:
@jl777 Unsure what you mean. Schnorr signatures are not compatible with ECDSA at all, and needs different signing and verification code. Further, Bitcoin scripts that do not use any digital signature are inherently insecure (if there is no secret information being verified, miners can steal all the coins once they see a spend). Thus, indeed, the only way is using a new opcode.

jl777:
it would be combined with a normal signature: OP_HASH160 <hash(sharedsecret)> OP_EQUALVERIFY <normal_pubkey> OP_CHECKSIG

granted the protection from just the normal_pubkey OP_CHECKSIG isnt as much as full verification, but if there was a way to offchain calculate a schnorr sig based sharedsecret equivalent, then I think it would work.

problem is I dont see a way to know ahead of time any value that a group signature would create. The combined pubkey appears to be a standard pubkey, but without any normal privkey. Maybe there is a way for a set of nodes to create onetime use keypairs and somehow create a combined privkey that can be used with the combined pubkey.

sipa:
You seem to be describing a normal signature with a hashlocked secret. Where does Schnorr come in?

jl777:
the schnorr would come in offchain, so a group can do larger scale MofN between themselves
was hoping there was a way to generate some sort of sharedsecret equivalent among the participants so that if each member of the group validates the schnorr combined sig, they release their portion of the sharedsecret.

using shamir's shared secret that each of the participants have and release only if they are satisfied, comes really close to being the hashlocked secret to use. The problem is we cant have any single node to know the secret that is being shamir'ed.

maybe some sort of oblivious transfer method could be used, but this is getting to where I am not comfortable as I dont know of any C libraries for that.

So the constraints are that each member of the group can directly run the schnorr lib, they can coordinate and precalculate any sort of magic value, but ultimately a hashlocked bitcoin tx needs to be used

PR #213, opened 2015-02-13
Removed gotos, which are hard to trace and maintain.

sipa:
Rebased version of #97.

### DIFF
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -37,7 +37,6 @@ int secp256k1_ecdsa_verify(const unsigned char *msg32, const unsigned char *sig,
     secp256k1_ge_t q;
     secp256k1_ecdsa_sig_t s;
     secp256k1_scalar_t m;
-    int ret = -3;
     DEBUG_CHECK(secp256k1_ecmult_consts != NULL);
     DEBUG_CHECK(msg32 != NULL);
     DEBUG_CHECK(sig != NULL);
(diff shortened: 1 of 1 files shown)


gmaxwell:
utACK.

src/secp256k1.c
```diff
@@ -46,20 +45,16 @@ int secp256k1_ecdsa_verify(const unsigned char *msg32, const unsigned char *sig,
     secp256k1_scalar_set_b32(&m, msg32, NULL);
 
     if (!secp256k1_eckey_pubkey_parse(&q, pubkey, pubkeylen)) {
-        ret = -1;
-        goto end;
+        return -1;
```
droark: (Caveat: I know this code has already been committed. I didn't see the pull request until just now.)

If maintenance is a concern, wouldn't a slightly better solution be to return to the original code, ditch the gotos, and switch the if statements to if/else if/else if? That way, the code doesn't become littered with return statements. I was taught growing up that, whenever possible, it's best to write code that has only one return statement in order to make it easier to read and debug.

Thanks.

gmaxwell: to quote myself before:

03:21 <gmaxwell> Would you consider this a good alternative to Ian's goto elimination patch? https://0bin.zertrin.org/paste/757a01c8686d7258013d707101734d4121dbee5c#Lk85xk9ip5I9LXVPBoOLOGiJH25ZKTVXLTWB1HDVUJg=  it also avoids the multiple returns (which are really just goto in disguise, and can be problematic for the same reasons)

sipa: This paste has expired.

sipa: Anyway, sure; patches welcome

PR #215, opened 2015-02-14
Use RFC6979 for test PRNGs

sipa:
Around 10% slower.


sipa:
Added a commit that changes the seed to a 16-byte hex, avoiding the need for a 64-bit format specifier.

gmaxwell:
ACK.

PR #216, opened 2015-02-15
Use faster byteswapping and avoid alignment-increasing casts

sipa:


sipa:
9% faster SHA256 and RFC6979.

gmaxwell:
ACK.  (remind me to test on BE if I forget to remove this request.)

sipa:
clang seems to warn about changed alignment when casting. I don't see why that is a problem, so perhaps we should try to avoid that warning?

sipa:
@theuni opinion?

theuni:
@sipa I took a quick stab at avoiding changing alignment requirements: https://github.com/theuni/secp256k1/commit/78df7673f9d31627ae6425face109a0df6ac08ac just to see how much trouble it'd be. Assuming I didn't muck it up too badly, overhead seems minimal. It's hard to imagine this causing any real issues, though.

sipa:
Hmm, a 1% slowdown. This shouldn't be necessary at all...

sipa:
I've rewritten this based on @theuni's suggestion, taking it a step further. Oddly enough, it's even faster than the previous version.

Also, I've just switched to a different byteswapping macro, which recent GCC at -O2 and above automatically convert to a bswap instruction.

sipa:
@theuni clang-3.5 also optimizes the construct to a bswap; I haven't tried any other implementation.

sipa:
10.5% faster hashing now.

theuni:
Nice! Tested ACK. Roughly an 8% speedup here from a very quick test.

PR #217, opened 2015-02-17
Add built-in self-tests and test vectors.

gmaxwell:
This has a collection of static test vectors which were both randomly
and hand created and then machine reduced.  All of the vectors were
tested against OpenSSL for agreement.

The vectors are checked at _start(), unless specifically built with
the them disabled (e.g. to reduce binary size.).

LCOV claims 91% branch coverage with the tests now.


gmaxwell:
I guess for discuss right now, I'm not super happy that the branch coverage is only 91%.

gmaxwell:
Here is a recent lcov run with this patch in: http://people.xiph.org/~greg/secp256k1.bist1/

There are some unhit branches which are cryptographically unreachable, e.g. ecdsa_impl.h:206, and others are just serialization handling which would be easy enough to hit if thats what I was currently attempting.  I'm most irritated right now with the uncovered carries in the scalar code, but I haven't tried to specifically craft something to hit them yet.

src/bist.h
```diff
@@ -0,0 +1,17 @@
+/**********************************************************************
+ * Copyright (c) 2014                                                 *
```
sipa: You can put your name here :)

sipa:
Doesn't compile with --enable-benchmark.

gmaxwell:
Hmph. Pretty sure it did at one point, I must have hosed it on a rebase. Will fix!

sipa:
It only fails with non-standard -O options. I suppose it's something that's expected to be inlined but becomes an external symbol otherwise.

sipa:
@gmaxwell See my https://github.com/sipa/secp256k1/commits/bist branch for a fix.

sipa:
For me, this increases the time for a secp256k1_start(SECP256K1_START_VERIFY | SECP256K1_START_SIGN) from 15ms to 53ms.

jtimon:
Fast review ACK. Needs rebase.

PR #223, opened 2015-02-23
Add DLL exports.

evoskuil:


sipa:
@theuni Comments?

include/secp256k1.h
```diff
@@ -40,6 +40,26 @@ extern "C" {
...
+#   define SECP256K1_HELPER_DLL_EXPORT __attribute__ ((visibility ("default")))
+#  else
+#   define SECP256K1_HELPER_DLL_IMPORT
+#   define SECP256K1_HELPER_DLL_EXPORT
+#  endif
+# endif
+
+# if defined SECP256K1_STATIC
+#  define SECP256K1_API
+# elif defined SECP256K1_DLL
```
chfast: CMake has a convention to add macro `MyLibrary_EXPORTS` when building a DLL. Is it possible to rename `SECP256K1_DLL` to SECP256K1_EXPORTS`?

evoskuil: Certainly, the symbol is otherwise arbitrary.

theuni: Sounds good to me.

I think this should be ```__attribute__ ((visibility ("hidden")))``` in the ```SECP256K1_STATIC``` case though, since it wouldn't need to be externally visible

include/secp256k1.h
```diff
@@ -40,6 +40,26 @@ extern "C" {
 #  define SECP256K1_ARG_NONNULL(_x)
 # endif
 
+# if defined _MSC_VER || defined __CYGWIN__
+#  define SECP256K1_HELPER_DLL_IMPORT __declspec(dllimport)
```
chfast: dllimport is only useful in case of global variables and C++ classes, not in case of functions. Removing dllimport support would simplify the patch.

include/secp256k1.h
evoskuil: This would be fine but it needs to be isolated from the Windows build. I'd do that by adding another helper:
```c
    define SECP256K1_HELPER_DLL_EXPORT __declspec(dllexport)
else
    if __GNUC__ >= 4
        define SECP256K1_HELPER_DLL_IMPORT __attribute__ ((visibility ("default")))
        define SECP256K1_HELPER_DLL_EXPORT __attribute__ ((visibility ("default")))
        define SECP256K1_HELPER_STATIC __attribute__ ((visibility ("hidden")))
    else
        define SECP256K1_HELPER_DLL_IMPORT
        define SECP256K1_HELPER_DLL_EXPORT
        define SECP256K1_HELPER_STATIC
    endif
endif

if defined SECP256K1_STATIC
    define SECP256K1_API SECP256K1_HELPER_STATIC
elif defined SECP256K1_DLL
```

include/secp256k1.h
evoskuil: I'm not sure this is the case. This is an old reference, but I have to assume it still holds:

> You do not need to use __declspec(dllimport) for your code to compile correctly, but doing so allows the compiler to generate better code.

https://msdn.microsoft.com/en-us/library/aa271769(v=vs.60).aspx

include/secp256k1.h
```diff
@@ -131,6 +131,30 @@ typedef int (*secp256k1_nonce_function_t)(
...
+# endif
+
+# if defined SECP256K1_STATIC
+#  define SECP256K1_API SECP256K1_HELPER_LOCAL
+# elif defined SECP256K1_EXPORTS
+#  define SECP256K1_API SECP256K1_HELPER_DLL_EXPORT
+# else
+#  define SECP256K1_API SECP256K1_HELPER_DLL_IMPORT
+# endif
+
```
evoskuil: The above matches closely the [GCC example](https://gcc.gnu.org/wiki/Visibility) (bottom of page). There is deviation in the public symbols `SECP256K1_STATIC` and `SECP256K1_EXPORTS`. The latter is modified for better CMake integration (see [comment](https://github.com/bitcoin/secp256k1/pull/223#discussion_r39398451) by @chfast). WRT the former I consider "static" more explicit than "local" and therefore better for publication.

src/secp256k1.c
```diff
@@ -254,7 +254,10 @@ static int nonce_function_rfc6979(unsigned char *nonce32, const unsigned char *m
    return 1;
 }
 
+SECP256K1_API
 const secp256k1_nonce_function_t secp256k1_nonce_function_rfc6979 = nonce_function_rfc6979;
+
+SECP256K1_API
 const secp256k1_nonce_function_t secp256k1_nonce_function_default = nonce_function_rfc6979;
```
evoskuil: I have not had a chance to verify these two lines, but according to documentation it appears correct for both VC and GCC.

include/secp256k1.h
chfast: @evoskuil, you are right. I've checked the Visual Studio 2013 and it generates better code when dllimport is used (as described in the article).

gmaxwell:
Please take a look at #314

evoskuil:
I prefer the readability of the gcc example layout, and the use of `dllimport`, but not a big deal in either case.

chfast:
@gmaxwell #314 looks like a subset of this one. With the except GCC visibility is not set to hidden by default here.

sipa:
This defines SECP256K1_STATIC and SECP256K1_EXPORT, but they aren't being set anywhere?

If in #314, in the non-SECP256K1_BUILD branch for windows, declspec(dllimport) is set rather than nothing, do we get the best of both worlds?

sipa:
I also think that this fails to set dllexport/import when compiling with MinGW?

sipa:
Superseded by #314.

PR #227, opened 2015-03-08
BUG: array bounds overflow

fsb4000:
A value of type "const char [33]" cannot be used to initialize an entity of type "const unsigned char [32]"
(C string ends \0 symbol)

### DIFF
--- a/src/ecmult_gen_impl.h
+++ b/src/ecmult_gen_impl.h
@@ -46,7 +46,7 @@ static void secp256k1_ecmult_gen_start(void) {
 
     /* Construct a group element with no known corresponding scalar (nothing up my sleeve). */
     {
-        static const unsigned char nums_b32[32] = "The scalar for this x is unknown";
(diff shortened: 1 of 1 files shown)


gmaxwell:
This is valid C.  

"(3.5.7) ... An array of character type may be initialized by a character string
literal, optionally enclosed in braces.  Successive characters of the
character string literal (including the terminating null character if
there is room or if the array is of unknown size) initialize the
members of the array."

If you want to change it (e.g. to use this code in C++ where the behavior is different) the correct change would be to change the array length.  I believe the value you've changed that to is _not_ an element of the group.

fsb4000:
MSVC
https://msdn.microsoft.com/en-us/library/5awwxh25.aspx
Although it is possible not a bug in gcc. I do not know...

gmaxwell:
Your software is incorrect.

fsb4000:
Ok. Thank you.

PR #228, opened 2015-03-08
Silence some warnings from pedantic static analysis tools, improve compatibility with C++

gmaxwell:
C doesn't include the null in an array initilized from a
string literal if it doesn't fit, in C++ this is invalid.

The vararray style prototypes and init+calc also changed in
 this commit are not C89 enough for some tools.


gmaxwell:
(Sorry, wrote this in PR and seemingly never got around to PRing it; presumably because I expected to have more tool-nits like this.)

sipa:
ACK

PR #229, opened 2015-03-20
Add ability to pass extra entropy to rfc6979

sipa:
Suggested by Greg Maxwell.

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -97,7 +97,10 @@ typedef int (*secp256k1_nonce_function_t)(
   const void *data
 );
 
-/** An implementation of RFC6979 (using HMAC-SHA256) as nonce generation function. */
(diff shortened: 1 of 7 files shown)


gmaxwell:
I've sent email to Trevor Perrin on the question of exactly how the 3.6 "additional data" was supposed to be applied with respect to steps 3.2.d vs 3.2.f in the draft.

I believe all solutions are likely to be adequate but for consistency I thought it would be good to clarify whats intended in the draft.

sipa:
Updated to apply it both 3.2.d and 3.2.f.

gmaxwell:
ACK with https://people.xiph.org/~greg/0001-Add-a-simple-testcase-for-signing-with-rfc6979-extra.patch  or something substantially similar added. :)

sipa:
Included your commit.

sipa:
ACK extra commit.

PR #230, opened 2015-03-27
Brace all the if/for/while.

gmaxwell:
Unbraced statements spanning multiple lines has been shown in many
 projects to contribute to the introduction of bugs and a failure
 to catch them in review, especially for maintenance on infrequently
 modified code.

Most, but not all, of the existing practice in the codebase were not
 cases that I would have expected to eventually result in bugs but
 applying it as a rule makes it easier for other people to safely
 contribute.

I'm not aware of any such evidence for the case with the statement
 on a single line, but some people strongly prefer to never do that
 and the opposite rule of "_always_ use a single line for single
 statement blocks" isn't a reasonable rule for formatting reasons.
 Might as well brace all these too, since that's more universally
 acceptable.

[In any case, I seem to have introduced the vast majority of the
 single-line form (as they're my preference where they fit).]

This also removes a broken test which is no longer needed.

### DIFF
--- a/src/ecdsa_impl.h
+++ b/src/ecdsa_impl.h
@@ -53,35 +53,59 @@ static int secp256k1_ecdsa_sig_parse(secp256k1_ecdsa_sig_t *r, const unsigned ch
     int lenr;
     int lens;
     int overflow;
-    if (sig[0] != 0x30) return 0;
+    if (sig[0] != 0x30) {
+        return 0;
+    }
     lenr = sig[3];
-    if (5+lenr >= size) return 0;
+    if (5+lenr >= size) {
+        return 0;
+    }
     lens = sig[lenr+5];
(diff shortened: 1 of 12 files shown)


sipa:
ACK

dcousens:
ACK

dcousens:
@gmaxwell I understand the sentiment, but I don't think the added brace helps in the 1-liner case.
It usually only seems to help in the

``` javascript
if (x == 3)
  func()
```

Case, because the indentation is the only differentiation between that and other code.

Compare that to

``` javascript
if (x == 3) func()
```

Still not great, which is why typically these statements are limited to early exits only, as they are clearly identifiable by the `return` statement.

``` javascript
if (x == 3) return 0
```

Just my 2c.

PS.  Why not just remove the test in a separate commit?  Atomic commits! :)

PR #232, opened 2015-03-28
Avoid some implicit type conversions to make C++ compilers happy.

gmaxwell:


dcousens:
@gmaxwell any reason for C style casts?

dcousens:
@gmaxwell any reason for C style casts?

gmaxwell:
@dcousens Because this software is written in strict C89 (plus a couple of narrow additions like long long).

The reason for this change is that someone may still compile it with a C++ compiler because that is all they have available or some other reason (e.g. they have some analysis or instrumentation code that is C++ only), and there is no reason to be gratuitously incompatible.

dcousens:
@gmaxwell ah!  Sorry, I completely forgot about that,  just context switched and read C++ in the title.  Cheers for clearing that up :)

PR #233, opened 2015-04-03
initialize variable

theuni:
Without this, clang's static analyzer finds:
```
./src/field_5x52_impl.h:425:24: warning: The left operand of '&' is a garbage value
    r->n[0] = (r->n[0] & mask0) | (a->n[0] & mask1);
```


gmaxwell:
I was kind of lamenting if we should do something about that. GCC -O3 complains too, it's just that it's not smart enough to know that the conditional move will blow it away. 

I measure a 2% slowdown from this, but it could be measurement noise.  (seems kinda high!)

ACK for whatever it's worth from my side.

sipa:
I can't measure any difference when locking my CPU to a single frequency.

PR #236, opened 2015-04-11
Expose ability to deep-copy a context

apoelstra:
For the Rust bindings (and likely for binding to other languages which hide memory management from the user) I need to be able to deep-copy a secp256k1 context. This exposes a `secp256k1_context_clone` function which serves this purpose.

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -59,6 +59,14 @@ secp256k1_context_t* secp256k1_context_create(
   int flags
 ) SECP256K1_WARN_UNUSED_RESULT;
 
+/** Copies a secp256k1 context object.
+ *  Returns: a newly created context object.
+ *  In:      ctx: an existing context to copy
+ */
+secp256k1_context_t* secp256k1_context_clone(
+  const secp256k1_context_t* ctx
+) SECP256K1_WARN_UNUSED_RESULT;
+
 /** Destroy a secp256k1 context object.
  *  The context pointer may not be used afterwards.
  */
--- a/src/ecmult.h
+++ b/src/ecmult.h
@@ -20,6 +20,8 @@ typedef struct {
 
 static void secp256k1_ecmult_context_init(secp256k1_ecmult_context_t *ctx);
 static void secp256k1_ecmult_context_build(secp256k1_ecmult_context_t *ctx);
+static void secp256k1_ecmult_context_clone(secp256k1_ecmult_context_t *dst,
+                                           const secp256k1_ecmult_context_t *src);
 static void secp256k1_ecmult_context_clear(secp256k1_ecmult_context_t *ctx);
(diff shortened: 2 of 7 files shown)


gmaxwell:
ACK.

sipa:
I don't think this code will work for anything but a VERIFY + SIGN
initialized context.

apoelstra:
Eek, you're right @sipa.

apoelstra:
I've pushed a new version which behaves properly for all context types, and tests that they work correctly.

sipa:
ACK

DavidEGrayson:
> For the Rust bindings (and likely for binding to other languages which hide memory management from the user) I need to be able to deep-copy a secp256k1 context.

A few months ago, before pull request #208 was merged, I made a [Ruby binding](https://github.com/DavidEGrayson/secp256k1-rb) for secp256k1 which uses contexts.  I store the context pointer in an `FFI::Autopointer` object.  When that object gets garbage collected, it calls `secp256k1_context_destroy` on the underlying pointer.  (If I had written the binding in C instead of using FFI, I could have accomplished the same thing.)

So does Rust really require you to be able to deep copy objects like this?  [Libusb](http://libusb.info)'s contexts do not have a clone feature but it looks like someone made a [Rust binding for Libusb](https://github.com/kevinmehall/rust-usb/).

apoelstra:
@DavidEGrayson The short answer is no; I could avoid this either by hurting usability for users of my bindings, or by introducing synchronization primitives to share data rather than cloning it.

The long answer:

No, I could just keep it in an atomic refcount wrapper for example, which would give similar behaviour to Ruby. But then I'm introducing a dependency on Rust's runtime; `rust-secp256k1` makes a point of never using Rust's allocator or synchronization primitives to ensure it can be use in very limited environments. (I haven't measured but I'd also expect that in many contexts copying a couple hundred bytes will be faster than using the refcount, both because of the cost of atomic operations and because you'd need a pointer which is accessible from multiple threads; in any case it's easy for a user to add refcounting if the bindings don't, and not easy to remove it if the bindings do.)

The Libusb bindings provide a Rust `Context` object to users of its API, but this cannot be copied in any way, which forces users who need copies to create new contexts either to create new ones or to use some other synchronization wrapper. (And for libusb, atomic refcounts are not sufficient because libusb contexts are mutated in use, so a more expensive synchronization primitive like a mutex would be needed.)

Finally, secp256k1 contexts are *really* expensive to create from scratch, and are needed for things like validating that keys are well-formed, which isn't naturally tied to any long-lived object that could hold a context. My guess is that libusb's contexts are fairly cheap to create and that in practice they tend to be tied to some sort of object representing a USB device, so they don't need to be created that often anyway.

DavidEGrayson:
That makes sense, thanks!

PR #237, opened 2015-04-12
Fix typo: avg -> max

sipa:


gmaxwell:
"The keys are right next to each other!"

ACK.

PR #241, opened 2015-04-16
Fix memory leak in context unit test

apoelstra:
Before fix, `valgrind ./tests 1` outputs
  ==21959==     in use at exit: 2,228,288 bytes in 8 blocks

after:
  ==23974==     in use at exit: 0 bytes in 0 blocks


gmaxwell:
tested ACK. Thanks for fixing this so quickly.

sipa:
Tested ACK

PR #244, opened 2015-04-22
travis: update to new build infrastructure

theuni:
See here for the reasoning:
http://blog.travis-ci.com/2014-12-17-faster-builds-with-container-based-infrastructure/
    
Quick version: these changes remove our use of sudo so that we can move to the container-based
builds. This yields quicker builds and less reliance on the old infrastructure. It looks to be ~30%-50% quicker, though it's hard to tell because there's so much variance.

Since we can't use sudo, we now have to handle packages in the build-matrix rather than as part of the build step.

The second commit is pretty much unrelated, I just noticed that there was a gap in testing here.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ matrix:
         apt:
           packages:
             - gcc-multilib
+            - libgmp-dev:i386
     - compiler: clang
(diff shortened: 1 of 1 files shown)


sipa:
ACK

PR #249, opened 2015-05-10
bugfix

faizkhan00:
bad memory access, should not be ptr

### DIFF
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -278,7 +278,7 @@ int secp256k1_ec_pubkey_decompress(const secp256k1_context_t* ctx, unsigned char
     DEBUG_CHECK(pubkeylen != NULL);
     (void)ctx;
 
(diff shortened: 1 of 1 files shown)


faizkhan00:
Hmm, maybe the fix is wrong, but the function signature expects an int, and the call is passing a pointer

gmaxwell:
The code is correct.  The decompress function takes a pointer to an int; this pointer is derefernced before passing to the internal parse function.  The reason the decompress input is a pointer is because the length is an in/out field-- it gets updated with the new length after decompression.

(As you can see from the travis results, this function is extensively tested and the change breaks the tests).

I am guessing that the application code you are using is incorrectly passing in an integer (with type coercion to make it compile).

faizkhan00:
Yea, I was passing in a integer and it was being coerced, completely my fault.

PR #252, opened 2015-05-15
Add constant-time point multplication for ECDH

apoelstra:
Add constant-time `secp256k1_point_multiply` for ECDH
    
Designed with clear separation of the wNAF conversion, precomputation
and exponentiation (since the precomp at least we will probably want
to separate in the API for users who reuse points a lot).
    
Future work:
  - actually separate precomp in the API 
  - do multiexp rather than single exponentiation


sipa:
I'm on vacation until june 1st. I'll review afterwards.

gmaxwell:
"/** Do an ellitic curve scalar multiplication in constant time."

spelling.

apoelstra:
Added endomorphism optimization

apoelstra:
Pushed; had to do some algebraic changes to get the endomorphism stuff to work. The pre-endomorphism commits have been changed in some minor ways (the unit tests are a bit cleaner, `secp256k1_scalar_wnaf_force_odd` is replaced by `secp256k1_scalar_wnaf_cond_negate` and `secp256k1_scalar_is_even`, and the return type of `secp256k1_ecdh_wnaf` is now `void` since it used to return a fixed value, which was confusing. This value is now computed by a macro.)

As for the endomorphism stuff, the wNAF conversion is actually a bit different. We require all wNAF digits to be odd, so to represent even 256-bit numbers, we (a) negate the number, making it odd since our modulus is odd; (b) convert this to wNAF; (c) negate the individual digits, un-negating the number while keeping all digits odd. In this way all 256-bit numbers can be represented. **However**, with the endomorphism optimization, we find ourselves converting 128-bit numbers to wNAF, which cannot be negated this way. (They can, but their negations are 256 bits, so we lose the optimization.) We therefore use a different strategy for handling even numbers when endomorphism optimization is enabled. Specifically, we add 1 to even numbers, add 2 to odd ones, and compensate after the multiplication by subtracting either `P` or `2P` from the result. This is suggested in the Okeya/Tagaki paper.

sipa:
I have a few nits regarding naming and a question (see above), but the code looks well-written, well-documented and well-tested. Nice!

apoelstra:
When I add a test against `secp256k1_ec_pubkey_create` that runs 100 times with a `random_scalar_order` scalar it fails usually when the endomorphism optimization is enabled. Investigating.

apoelstra:
Rebased on https://github.com/bitcoin/secp256k1/pull/261, which fixes the problems I was having earlier with the endomorphism. Also add the additional test @sipa suggested (which didn't pass until #261 :))

src/secp256k1.c
```diff
@@ -223,6 +225,32 @@ int secp256k1_ecdsa_recover_compact(const secp256k1_context_t* ctx, const unsign
...
+    DEBUG_CHECK(scalar != NULL);
+
+    if (secp256k1_eckey_pubkey_parse(&pt, point, *pointlen)) {
+        secp256k1_scalar_set_b32(&s, scalar, &overflow);
+        if (overflow) {
+            ret = -2;
+        } else {
+            secp256k1_ecdh_point_multiply(&res, &pt, &s);
+            secp256k1_ge_set_gej(&pt, &res);
+            ret = secp256k1_eckey_pubkey_serialize(&pt, point, pointlen, *pointlen <= 33);
```
peterdettman: The output point is supposed to be secret (ECDH agreement), but _pubkey_serialize uses _normalize_var internally because it's assuming a public key point.

peterdettman:
A couple of issues with the documentation for secp256k1_point_multiply:

- The return code -1 is listed for "scalar was zero", but it looks to me like it would actually return 0 in that case (via secp256k1_eckey_pubkey_serialize).
- Is the scalar actually restricted to a normalized field value (no "overflow")? In which case, "a 32-byte scalar" is a little misleading.

include/secp256k1.h
```diff
@@ -217,6 +217,24 @@ SECP256K1_WARN_UNUSED_RESULT int secp256k1_ecdsa_recover_compact(
   int recid
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
 
+
+/** Do an elliptic curve scalar multiplication in constant time.
+ *  Returns: 1: exponentiation was successful
+ *          -1: scalar was zero (cannot serialize output point)
+ *          -2: scalar overflow
+ *          -3: invalid input point
+ *  In:      scalar:   a 32-byte scalar with which to multiple the point
```
peterdettman: multiple -> multiply

src/secp256k1.c
apoelstra: Yikes! Thanks for catching this.

apoelstra:
Rebase and fix minor comments from Peter.

apoelstra:
I've pushed a new version which fixes the timing leak. I've changed the API to output the hash of the computed point rather than the point itself. @gmaxwell had suggested early on that giving the user a curvepoint was "too low level" and not consistent with our policy of not exposing raw cryptographic objects.

The concern was that a user who has access to a "ECDH secret" which is a curvepoint might try to build his or her own cryptosystem in a fragile or broken way. I think this timing leak is an example of me getting burned by this -- I didn't think to look for this timing leak because I was thinking in terms of "points" rather than "secrets".

apoelstra:
This means that #262 is going to need to be changed to use `secp256k1_point_multiply` rather than `secp256k1_ecdh_point_multiply` since the latter now outputs a hash (which is algebraically useless).

src/bench_ecdh.c
```diff
@@ -0,0 +1,49 @@
...
+    data->pointlen = sizeof(point);
+    memcpy(data->point, point, data->pointlen);
+}
+
+static void bench_multiply(void* arg) {
+    int i;
+    bench_multiply_t *data = (bench_multiply_t*)arg;
+
+    for (i = 0; i < 20000; i++) {
+        CHECK(secp256k1_ecdh_point_multiply(data->point, &data->pointlen, data->scalar) == 1);
```
peterdettman: Seems to have been missed in the API change?

apoelstra: Yeah, oops. Should we put `-Werror` in Travis to prevent this oversight in future?

include/secp256k1.h
```diff
@@ -217,6 +217,24 @@ SECP256K1_WARN_UNUSED_RESULT int secp256k1_ecdsa_recover_compact(
   int recid
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
 
+/** Compute an EC Diffie-Hellman secret in constant time
+ *  Returns: 1: exponentiation was successful
+ *           0: scalar was zero (cannot serialize output point)
+ *          -1: scalar overflow
+ *          -2: invalid input point
```
peterdettman: The error codes seem out-of-sync with the method body.

peterdettman:
@apoelstra @gmaxwell Is there somewhere I can review the discussion about exposing low-level cryptographic operations (alluded to above)?

src/tests.c
```diff
@@ -1235,6 +1253,148 @@ void run_ge(void) {
...
+void run_ecdh_tests(void) {
+    ecdh_mult_zero_one();
+    ecdh_random_mult();
+    ecdh_commutativity();
+    ecdh_chain_multiply();
+}
+
+void run_ecdh_api_tests(void) {
+    ecdh_generator_basepoint();
+}
```
peterdettman: Method run_ecdh_api_tests not actually called?

apoelstra: Ugh, thanks, this was a rebasing error.

apoelstra:
@peterdettman, @gmaxwell and I had a private conversation on May 15 with the lines
```
13:18:19 <gmaxwell> One thing I wonder about is that we've tried to avoid offering users APIs that were raw EC operations... both becaues the users won't use them safely, and because they'll construct less efficient implementations of crypto algorithims (e.g. if we expose add and double they can go write a non-constant time, slow as heck ecdh.).  I wonder if there are any api affordances we can make to make that multiply s
13:23:47 <andytoshi> there thave been a lot of comments on -dev and -wizards from people who are pretty jacked to have ECDH but who IMHO are probably gonna do something unsafe with it
```

This was right before I started working on this PR. Then a few days ago in #secp256k1 on IRC:
```
10:51:09 <andytoshi> in https://github.com/bitcoin/secp256k1/pull/252#discussion_r33412268 peter dettman notes that pubkey_serialize uses fe_normalize_var which is a timing leak for ECDH
10:51:23 <andytoshi> what do we do about this? do we need to split serialize into serialize and serialize_var?
11:02:55 <sipa> i suppose
11:03:00 <sipa> wait
11:03:08 <sipa> right
11:03:28 <sipa> ecdh is weird because it has an EC point which is secret information
13:12:02 <gmaxwell> well I slightly protested before that returning a point was maybe too low level.
13:31:22 <andytoshi> i recall that ... this might be a concrete reason to -not- do it
13:31:50 <andytoshi> i mean, this feels like we got cut by our own sharp corner here..
14:36:30 <sipa> we could hash it before returning
16:31:30 <gmaxwell> that was my suggestion.
```

To the best of my knowledge this was the entirety of our discussion.

peterdettman:
Thanks for chasing that up. Not sure how I feel about the hashing yet, but I would like to try and get agreement that the result be just the x-coordinate (or the hash of it) and not include the y-bit. Of course, this is somewhat with a view to integrating #262 cleanly (as an optimization for compressed input points), but possibly x-only output could be considered sufficient impediment to building adhoc schemes on top of ECDH-as-scalar-mult. In any case, I think taking just the x-coordinate as the shared secret is fairly standard, or at least noncontroversial.

gmaxwell:
It depends on the application, I'm sure I can construct a contrived protocol where the 'malleability' of someone flipping the point but still getting the same secret would be bad;  e.g. where it causes you to reuse a key in a stream cipher where you otherwise wouldn't.

peterdettman:
Yes, I understand this to be "benign malleability" per Shoup e.g. http://www.shoup.net/papers/iso-2_1.pdf section 2.3. I would like to simply say that it can be addressed in the outer scheme, but I gather we're trying to foolproof things. I suppose one alternative here is to include the respective public keys under the result hash.

src/secp256k1.c
```diff
@@ -223,6 +225,50 @@ int secp256k1_ecdsa_recover_compact(const secp256k1_context_t* ctx, const unsign
...
 
+int secp256k1_ecdh(unsigned char *result, unsigned char *point, int *pointlen, const unsigned char *scalar) {
+    int ret = 0;
+    int overflow = 0;
+    secp256k1_gej_t res;
+    secp256k1_ge_t pt;
+    secp256k1_scalar_t s;
+    DEBUG_CHECK(point != NULL);
+    DEBUG_CHECK(pointlen != NULL);
+    DEBUG_CHECK(scalar != NULL);
```
peterdettman: DEBUG_CHECK(result != NULL). Also, point can now be const, and pointlen by-value.

src/ecdh_impl.h
```diff
@@ -0,0 +1,219 @@
...
+        secp256k1_gej_add_ge(r, r, &tmpa);
+
+        n = wnaf_lam[i];
+        ECMULT_TABLE_GET_GE(&tmpa, pre_a_lam, n, WINDOW_A);
+        VERIFY_CHECK(n != 0);
+        secp256k1_gej_add_ge(r, r, &tmpa);
+#else
+        n = wnaf[i];
+        VERIFY_CHECK(n != 0);
+        ECMULT_TABLE_GET_GE(&tmpa, pre_a, n, WINDOW_A);
```
peterdettman: _GET_GE is a simple table lookup, with the index derived from bits of the scalar, so we'll need a _cmov-based alternative.

src/ecdh_impl.h
```diff
@@ -0,0 +1,219 @@
...
+        secp256k1_gej_add_ge(r, r, &tmpa);
+#endif
+    }
+
+    if (!r->infinity) {
+        secp256k1_fe_mul(&r->z, &r->z, &Z);
+    }
```
peterdettman: Just perform the _fe_mul unconditionally?

src/ecdh_impl.h
sipa: Ugh, nice catch!

apoelstra:
Fixed all of @peterdettman's minor comments. Added a new commit to fix the timing leak in `GET_GE`. The perf hit I measure, using bench_ecdh, is 10%, both with and without endomorphism. :/

sipa:
Rebase?

apoelstra:
Rebased.

peterdettman:
@apoelstra See https://github.com/peterdettman/secp256k1/commit/de5436efcc37a0b051058ba60647284d7bef4337 for some ideas on how to recover another 1% or so from the cmov table-lookup.

apoelstra:
Thanks Peter. I've cherry-picked that commit. The total perf hit from doing constant-time lookups now looks like 1.7% to me.

apoelstra:
Rebase; change API to take a pubkey_t rather than point/pointlen, and to take a context_t.

apoelstra:
I've added a commit which moves the ECDH code into its own module but leaves the constant-time multiply code in the main library. It also renames the const-multiply functions, so the result is kinda a mess of renaming and moving.

I'll go back and rebase the original changeset as though this had been the layout from the start, but first I'd like an ACK that this is how we want to organize things.

src/bench_ecdh.c
```diff
@@ -0,0 +1,51 @@
...
+
+    for (i = 0; i < 20000; i++) {
+        CHECK(secp256k1_ecdh(data->ctx, res, &data->point, data->scalar) == 1);
+    }
+}
+
+int main(void) {
+    bench_multiply_t data;
+
+    run_benchmark("ecdh_mult", bench_multiply, bench_multiply_setup, NULL, &data, 10, 20000);
```
sipa: Technically, ECDH is no longer just a multiplication.

src/ecmult_const_impl.h
```diff
@@ -0,0 +1,242 @@
...
+    (r)->infinity = 0; \
+    secp256k1_fe_negate(&neg_y, &(r)->y, 1); \
+    secp256k1_fe_cmov(&(r)->y, &neg_y, (n) != abs_n); \
+} while(0)
+
+/** Convert a number to WNAF notation. The number becomes represented by sum(2^{wi} * wnaf[i], i=0..return_val)
+ *  with the following guarantees:
+ *  - each wnaf[i] an odd integer between -(1 << w) and (1 << w)
+ *  - each wnaf[i] is nonzero
+ *  - the number of words set is returned; this is always (256 + w - 1) / w
```
sipa: Is this true, even with USE_ENDOMORPHISM?

src/ecmult_const_impl.h
```diff
@@ -0,0 +1,242 @@
...
+
+    for (i = WNAF_SIZE(WINDOW_A - 1); i >= 0; i--) {
+        int n;
+        int j;
+        for (j = 0; j < WINDOW_A - 1; ++j) {
+            /* This is a variable-time doubling, but it is actually constant-time for
+             * nonzero points. We know on the first iteration that `r` will be zero
+             * and know (by uniqueness of wNAF) that `r` will never be zero after
+             * that iteration, so this does not result in a timing leak. */
+            secp256k1_gej_double_var(r, r, NULL);
```
sipa: Would you mind adding a secp256k1_gej_double_nonzero, which just VERIFY_CHECK's for nonzeroness, and dispatches to secp256k1_gej_double_var? Placing the assumption about algorithm internals in a higher level is a bit ugly.

sipa:
ACK design, with a few more nits.

Please squash and I'll merge.

apoelstra:
Thanks for the comments @sipa. All fixed, and rebased for a cleaner history.

sipa:
Thanks! Can you also make some changes to .travis so this gets tested?

Kagami:
It's a bit unfortunate that `secp256k1` always hash the derived secret, because it means I can't use it for existing protocols like [Bitmessage](https://bitmessage.org/wiki/Encryption).

Any ideas how can I get `Px` with `secp256k1`? Only by implementing my own version of `secp256k1_ecdh`? (Note that Bitmessage uses `SHA512(Px)`, so it's not unsafe.)

apoelstra:
@Kagami As you say, ECIES actually only needs a hash of Px, though it's not our hash. Can you open an issue requesting we extend the ECDH API to support a custom hash function (like the signing functions support custom nonce functions)?

This is something we've considered but haven't had any concrete usecases and I guess it forgot to get written down..

PR #258, opened 2015-06-18
Fix compilation with C++

vinniefalco:
This is needed to compile the library as C++


apoelstra:
How do you compile as C++? Is this something we can get Travis to do?

sipa:
Judging by earlier comments by Vinnie, I assume he is directly #include'ing
the .c source inside another project's C++ source code.

vinniefalco:
@sipa Right

apoelstra:
ACK

gmaxwell:
For testing you can override CC=g++   and bypass the compile test in configure.   I don't have a way to do the test without patching right now but perhaps its possible. A CI build that does this would be nice, otherwise it'll just be periodic fixes.

PR #260, opened 2015-06-22
Correct secp256k1_gej_add_ge

### LINKED ISSUE #257: `secp256k1_gej_add_ge` does the wrong thing with points P, Q with P_y = -Q_y but P≠-Q
It is possible, by using the value β from the endomorphism optimization, to obtain two points P = x : y : z and Q = βx : -y : z which are both on the curve. If you attempt to add these points using `secp256k1_gej_add_ge`, it incorrectly returns infinity, since it sees that the z-coordinates match and the y-coordinates are negative.

Specifically, it computes the value M = Y_1 * Z_2^3 + Y_2 * Z_1^3, and sets the returned z-coordinate to 2 * M * z_1z_2, and of course M will be zero for these two points. So it is not a simple matter of fixing how the infinity flag is set, the algebra actually does not compute the complete group law. This is derived in the paper "Eric Brier and Marc Joye, Weierstrass Elliptic Curves and Side-Channel Attacks" in Proposition 1, which explicitly assumes that the points P and Q don't happen.

The current secp256k1 API does not use this function except in one place, `secp256k1_ecmult_gen`, where it is impossible to get such an input to the function, so right now this bug is harmless. However https://github.com/bitcoin/secp256k1/pull/252 exposes it more directly, since when the endomorphism optimization is turned on roughly half of all scalars produce a bad pair P, Q in the wNAF ladder.

sipa:
Fixes #257 by computing both double and add, and cmov'ing the correct result.

This is a simple and complete solution, but it has a significant performance impact (18% slowdown for signing). More efficient solutions can come later.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -958,11 +958,17 @@ void ge_equals_gej(const secp256k1_ge_t *a, const secp256k1_gej_t *b) {
 
 void test_ge(void) {
     int i, i1;
+#ifdef USE_ENDOMORPHISM
+    int runs = 6;
+#else
     int runs = 4;
+#endif
     /* Points: (infinity, p1, p1, -p1, -p1, p2, p2, -p2, -p2, p3, p3, -p3, -p3, p4, p4, -p4, -p4).
      * The second in each pair of identical points uses a random Z coordinate in the Jacobian form.
      * All magnitudes are randomized.
      * All 17*17 combinations of points are added to eachother, using all applicable methods.
+     *
+     * When the endomorphism code is compiled in, p5 = lambda*p1 and p6 = lambda^2*p1 are added as well.
      */
     secp256k1_ge_t *ge = (secp256k1_ge_t *)malloc(sizeof(secp256k1_ge_t) * (1 + 4 * runs));
     secp256k1_gej_t *gej = (secp256k1_gej_t *)malloc(sizeof(secp256k1_gej_t) * (1 + 4 * runs));
@@ -977,6 +983,14 @@ void test_ge(void) {
         int j;
         secp256k1_ge_t g;
         random_group_element_test(&g);
+#ifdef USE_ENDOMORPHISM
+        if (i >= runs - 2) {
+            secp256k1_ge_mul_lambda(&g, &ge[1]);
+        }
+        if (i >= runs - 1) {
+            secp256k1_ge_mul_lambda(&g, &g);
+        }
+#endif
         ge[1 + 4 * i] = g;
         ge[2 + 4 * i] = g;
         secp256k1_ge_neg(&ge[3 + 4 * i], &g);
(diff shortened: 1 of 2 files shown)


gmaxwell:
Can this come with some a more comprehensive test?

sipa:
@gmaxwell Added a test that fails due to #257, and the first commit here fixes it.

sipa:
@gmaxwell Also added @apoelstra's test which does not depend on having the endomorphism compiled in.

src/tests.c
```diff
@@ -1146,11 +1160,79 @@ void test_ge(void) {
...
+     * lam  = (1 - x^3).roots()[1][0]
+     *
+     * # random "bad pair"
+     * P = C.random_element()
+     * Q = -int(lam) * P
+     * print "    P: %x %x" % P.xy()
+     * print "    Q: %x %x" % Q.xy()
+     * print "P + Q: %x %x" % (P + Q).xy()
+     */
+    secp256k1_gej_t aj = SECP256K1_GEJ_CONST(
```
dcousens: `const`?

apoelstra: I tried making these `const`, but `secp256k1_ge_set_gej` scales its
Jacobian input, blocking it. (IIRC one of the three constants can
be `const` without trouble, but I didn't to make the code more
symmetric.)

On Tue, Jun 23, 2015 at 10:05:53PM -0700, Daniel Cousens wrote:
> > +     * C = EllipticCurve ([F (0), F (7)])
> > +     * G = C.lift_x(0x79BE667EF9DCBBAC55A06295CE870B07029BFCDB2DCE28D959F2815B16F81798)
> > +     * N = FiniteField(G.order())
> > +     *
> > +     * # endomorphism values (lambda is 1^{1/3} in N, beta is 1^{1/3} in F)
> > +     * x = polygen(N)
> > +     * lam  = (1 - x^3).roots()[1][0]
> > +     *
> > +     * # random "bad pair"
> > +     * P = C.random_element()
> > +     * Q = -int(lam) * P
> > +     * print "    P: %x %x" % P.xy()
> > +     * print "    Q: %x %x" % Q.xy()
> > +     * print "P + Q: %x %x" % (P + Q).xy()
> > +     */
> > +    secp256k1_gej_t aj = SECP256K1_GEJ_CONST(
> 
> `const`?
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/bitcoin/secp256k1/pull/260/files#r33117648

-- 
Andrew Poelstra
Mathematics Department, University of Texas at Austin
Email: apoelstra at wpsoftware.net
Web:   https://www.wpsoftware.net/andrew

"When I came into my land, I did not understand:
 neither dry rot, nor the burn pile, nor the bark-beetle,
 nor the dry well, nor the black bear."
    -- Joanna Newsom

sipa:
Closing; @apoelstra has a better implementation which I'll turn into a PR soon.

PR #261, opened 2015-06-24
Correct `secp256k1_gej_add_ge` with 3% performance hit on signing

apoelstra:
Fixes https://github.com/bitcoin/secp256k1/issues/257 by catching the degenerate case, which appears as an intermediate variable being 0/0, and using an alternate expression for that variable in this case.

This results in only a ~3% perf hit on signing. There are no changes in the multiplication or squaring count; this is entirely due to additions and cmovs. "I doubt we can do better" but I said that about an earlier revision with a 6% hit and Peter Dettman found a way to beat it, so my doubts aren't worth much :)

### DIFF
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -464,7 +464,8 @@ static void secp256k1_gej_add_ge(secp256k1_gej_t *r, const secp256k1_gej_t *a, c
     /* Operations: 7 mul, 5 sqr, 5 normalize, 17 mul_int/add/negate/cmov */
     static const secp256k1_fe_t fe_1 = SECP256K1_FE_CONST(0, 0, 0, 0, 0, 0, 0, 1);
     secp256k1_fe_t zz, u1, u2, s1, s2, z, t, tt, m, n, q, rr;
-    int infinity;
+    secp256k1_fe_t m_alt, rr_alt;
+    int infinity, degenerate;
     VERIFY_CHECK(!b->infinity);
     VERIFY_CHECK(a->infinity == 0 || a->infinity == 1);
 
@@ -488,6 +489,34 @@ static void secp256k1_gej_add_ge(secp256k1_gej_t *r, const secp256k1_gej_t *a, c
      *    Y3 = 4*(R*(3*Q-2*R^2)-M^4)
      *    Z3 = 2*M*Z
      *  (Note that the paper uses xi = Xi / Zi and yi = Yi / Zi instead.)
+     *
+     *  This formula has the benefit of being the same for both addition
+     *  of distinct points and doubling. However, it breaks down in the
+     *  case that either point is infinity, or that y1 = -y2. We handle
+     *  these cases in the following ways:
+     *
+     *    - If b is infinity we simply bail by means of a VERIFY_CHECK.
+     *
+     *    - If a is infinity, we detect this, and at the end of the
+     *      computation replace the result (which will be meaningless,
+     *      but we compute to be constant-time) with b.x : b.y : 1.
+     *
+     *    - If a = -b, we have y1 = -y2, which is a degenerate case.
+     *      But here the answer is infinity, so we simply set the
+     *      infinity flag of the result, overriding the computed values
+     *      without even needing to cmov.
+     *
+     *    - If y1 = -y2 but x1 != x2, which does occur thanks to certain
+     *      properties of our curve (specifically, 1 has nontrivial cube
+     *      roots in our field, and the curve equation has no x coefficient)
+     *      then the answer is not infinity but also not given by the above
+     *      equation. In this case, we cmov in place an alternate expression
+     *      for lambda. Specifically (y1 - y2)/(x1 - x2). Where both these
+     *      expressions for lambda are defined, they are equal, and can be
+     *      obtained from each other by multiplication by (y1 + y2)/(y1 + y2)
+     *      then substitution of x^3 + 7 for y^2 (using the curve equation).
+     *      For all pairs of nonzero points (a, b) at least one is defined,
+     *      so this covers everything.
      */
 
     secp256k1_fe_sqr(&zz, &a->z);                       /* z = Z1^2 */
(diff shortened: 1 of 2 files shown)


src/group_impl.h
```diff
@@ -499,24 +500,43 @@ static void secp256k1_gej_add_ge(secp256k1_gej_t *r, const secp256k1_gej_t *a, c
...
+    degenerate = secp256k1_fe_normalizes_to_zero(&m) *
+                 secp256k1_fe_normalizes_to_zero(&rr) *
+                 (1 - a->infinity);
+    /* This only occurs when x1 = beta*x2 or x1 = beta^2*x2. In
+     * this case the correct value of lambda is (y1 - y2)/(x1 - x2),
+     * so set R/M equal to this. */
+    secp256k1_fe_negate(&rr_alt, &s2, 1);   /* rr = -Y2*Z1^3  */
+    secp256k1_fe_add(&rr_alt, &s1);         /* rr = Y1*Z2^3 - Y2*Z1^3 */
+    secp256k1_fe_negate(&m_alt, &u2, 1);    /* m = -X2*Z1^2 */
+    secp256k1_fe_add(&m_alt, &u1);          /* m = X1*Z2^2 - Y1*Z2^2 */
```
peterdettman: /* m = X1*Z2^2 - X2*Z1^2 */

src/group_impl.h
```diff
@@ -499,24 +500,43 @@ static void secp256k1_gej_add_ge(secp256k1_gej_t *r, const secp256k1_gej_t *a, c
...
-    secp256k1_fe_sqr(&t, &rr);                                      /* t = R^2 (1) */
-    secp256k1_fe_mul(&r->z, &m, &z);                                /* r->z = M*Z (1) */
+    secp256k1_fe_mul(&tt, &u1, &u2); secp256k1_fe_negate(&tt, &tt, 1); /* t = -U1*U2 (2) */
+    secp256k1_fe_add(&rr, &tt);                                        /* rr = R = T^2-U1*U2 (3) */
+    /** If lambda = R/M = 0/0 we have a problem (except in the "trivial"
+     *  case that Z = z1z2 = 0, and this is special-cased later on). */
+    degenerate = secp256k1_fe_normalizes_to_zero(&m) *
+                 secp256k1_fe_normalizes_to_zero(&rr) *
+                 (1 - a->infinity);
+    /* This only occurs when x1 = beta*x2 or x1 = beta^2*x2. In
```
peterdettman: x1 = beta*x2 or beta*x1 = x2
might make it clearer there are just the two possibilities.

src/group_impl.h
```diff
@@ -499,24 +500,43 @@ static void secp256k1_gej_add_ge(secp256k1_gej_t *r, const secp256k1_gej_t *a, c
...
+     * this case the correct value of lambda is (y1 - y2)/(x1 - x2),
```
apoelstra: Oh, this is the reciprocal of what it should be.

src/group_impl.h
apoelstra: Yeah, agreed. And I should probably add that these are exactly the nontrivial cases for which `x1^3 == x2^3`, which explains why there are only the two of them.

src/group_impl.h
```diff
@@ -499,24 +500,43 @@ static void secp256k1_gej_add_ge(secp256k1_gej_t *r, const secp256k1_gej_t *a, c
...
     secp256k1_fe_sqr(&rr, &t);                          /* rr = T^2 (1) */
-    secp256k1_fe_mul(&t, &u1, &u2); secp256k1_fe_negate(&t, &t, 1); /* t = -U1*U2 (2) */
-    secp256k1_fe_add(&rr, &t);                                      /* rr = R = T^2-U1*U2 (3) */
-    secp256k1_fe_sqr(&t, &rr);                                      /* t = R^2 (1) */
-    secp256k1_fe_mul(&r->z, &m, &z);                                /* r->z = M*Z (1) */
+    secp256k1_fe_mul(&tt, &u1, &u2); secp256k1_fe_negate(&tt, &tt, 1); /* t = -U1*U2 (2) */
+    secp256k1_fe_add(&rr, &tt);                                        /* rr = R = T^2-U1*U2 (3) */
+    /** If lambda = R/M = 0/0 we have a problem (except in the "trivial"
+     *  case that Z = z1z2 = 0, and this is special-cased later on). */
+    degenerate = secp256k1_fe_normalizes_to_zero(&m) *
```
sipa: Using & instead of * is probably slightly faster.

apoelstra: Oh, I can also probably drop the `* (1 - a->infinity)` since the cmov code at the bottom of the function will overwrite everything I do anyway.

sipa: Indeed!

src/group_impl.h
```diff
@@ -499,24 +500,43 @@ static void secp256k1_gej_add_ge(secp256k1_gej_t *r, const secp256k1_gej_t *a, c
...
+     * this case the correct value of lambda is (y1 - y2)/(x1 - x2),
+     * so set R/M equal to this. */
+    secp256k1_fe_negate(&rr_alt, &s2, 1);   /* rr = -Y2*Z1^3  */
+    secp256k1_fe_add(&rr_alt, &s1);         /* rr = Y1*Z2^3 - Y2*Z1^3 */
+    secp256k1_fe_negate(&m_alt, &u2, 1);    /* m = -X2*Z1^2 */
+    secp256k1_fe_add(&m_alt, &u1);          /* m = X1*Z2^2 - Y1*Z2^2 */
+
+    secp256k1_fe_cmov(&rr_alt, &rr, !degenerate);
+    secp256k1_fe_cmov(&m_alt, &m, !degenerate);
+    /* Now Ralt / Malt = lambda and is guaranteed not to be 0/0.
```
sipa: Would be nice if you could also update the algebra in the comments above to reflect this case.

src/group_impl.h
```diff
@@ -499,24 +500,43 @@ static void secp256k1_gej_add_ge(secp256k1_gej_t *r, const secp256k1_gej_t *a, c
...
+     * We keep the old values in R, M because we will need them
+     * for other reasons. */
+    secp256k1_fe_sqr(&n, &m_alt);                       /* n = M^2 (1) */
+    secp256k1_fe_mul(&q, &n, &t);                       /* q = Q = T*M^2 (1) */
+    secp256k1_fe_mul(&n, &n, &m);                       /* n = M^4 (1) */
+    secp256k1_fe_mul(&n, &n, &m_alt);                   /* n = M^4 (1) */
```
peterdettman: I can't test code right now, but it seems to me that 526,527:

    n = n * m * m_alt

could be replaced with:

    n = n * m^2

because either m == m_alt or m == 0.

Or even:

    n = n^2; cmov n,m,degenerate

apoelstra: I can't test for another hour, but if this cmov thing works I think it erases the perf hit!

apoelstra: I tried it (using vim on phone keyboard, ugh, but I couldn't wait..). It works, and drops the perf hit I measure from 5.7% to 3.8%! :D

apoelstra:
Push to fix everyone's comments (comment fixes and Peter's awesome speedup).

src/group_impl.h
```diff
@@ -499,29 +500,55 @@ static void secp256k1_gej_add_ge(secp256k1_gej_t *r, const secp256k1_gej_t *a, c
...
+    secp256k1_fe_add(&rr, &tt);                                        /* rr = R = T^2-U1*U2 (3) */
+    /** If lambda = R/M = 0/0 we have a problem (except in the "trivial"
+     *  case that Z = z1z2 = 0, and this is special-cased later on). */
+    degenerate = secp256k1_fe_normalizes_to_zero(&m) &
+                 secp256k1_fe_normalizes_to_zero(&rr);
+    /* This only occurs when y1 == -y2 and x1^3 == x2^3, but x1 != x2.
+     * This means either x1 == beta*x2 or beta*x1 == x2, where beta is
+     * a nontrivial cube root of one. In either case, an alternate
+     * non-indeterminate expression for lambda is (x1 - x2)/(y1 - y2),
+     * so we set R/M equal to this. */
```
apoelstra: Oh, I had the expression for lambda right the first time :) it *should* be `(y1 - y2)/(x1 - x2)`.

gmaxwell:
We might want to update the citation above to point out that the additive law given in the citation is actually not complete, lest some unwary traveler see the citation and go there rather than reading the code.

src/group_impl.h
```diff
@@ -499,29 +500,55 @@ static void secp256k1_gej_add_ge(secp256k1_gej_t *r, const secp256k1_gej_t *a, c
...
+    /* Now Ralt / Malt = lambda and is guaranteed not to be 0/0.
+     * From here on out Ralt and Malt represent the numerator
+     * and denominator of lambda; R and M represent the explicit
+     * expressions x1^2 + x2^2 + x1x2 and y1 + y2. */
+    secp256k1_fe_sqr(&n, &m_alt);                       /* n = Malt^2 (1) */
+    secp256k1_fe_mul(&q, &n, &t);                       /* q = Q = T*Malt^2 (1) */
+    /* These two lines use a clever observation of Peter Dettman's,
+     * that either M == Malt or M == 0. So M^3 * Malt is either Malt^4
+     * (which is computed by squaring), or zero (which is computed by
+     * the cmov). So the cost is one squaring versus two multiplications. */
```
peterdettman: It's of course nice to be appreciated, but I think code comments would be better kept to the immediate technical details.

gmaxwell: Direct experience shows that people do find attribution in commit messages just fine. E.g. I had people contact me on the inversion free R == r trick.

src/group_impl.h
```diff
@@ -499,29 +500,55 @@ static void secp256k1_gej_add_ge(secp256k1_gej_t *r, const secp256k1_gej_t *a, c
...
     m = s1; secp256k1_fe_add(&m, &s2);                  /* m = M = S1+S2 (2) */
-    secp256k1_fe_sqr(&n, &m);                           /* n = M^2 (1) */
-    secp256k1_fe_mul(&q, &n, &t);                       /* q = Q = T*M^2 (1) */
-    secp256k1_fe_sqr(&n, &n);                           /* n = M^4 (1) */
     secp256k1_fe_sqr(&rr, &t);                          /* rr = T^2 (1) */
-    secp256k1_fe_mul(&t, &u1, &u2); secp256k1_fe_negate(&t, &t, 1); /* t = -U1*U2 (2) */
-    secp256k1_fe_add(&rr, &t);                                      /* rr = R = T^2-U1*U2 (3) */
-    secp256k1_fe_sqr(&t, &rr);                                      /* t = R^2 (1) */
-    secp256k1_fe_mul(&r->z, &m, &z);                                /* r->z = M*Z (1) */
+    secp256k1_fe_mul(&tt, &u1, &u2); secp256k1_fe_negate(&tt, &tt, 1); /* t = -U1*U2 (2) */
```
sipa: comment: tt instead of t

apoelstra:
For what it's worth, the citation does say that it's assuming the y1 = -y2 case does not occur.

src/group_impl.h
sipa: Is it (y1 - y2)/(x1 - x2) or the inverse? The comment still says the inverse.

apoelstra:
Push with comment fixes and a much-expanded comment about our high-level strategy for obtaining constant-time multiplication for all pairs of points (well, except those with a = infinity).

We might be the first to have produced a fast, constant-time algorithm which works for all pairs of points. When developing this fix I checked both Trezor and OpenSSL, and their constant-time additions both have degenerate cases (specifically both don't work when asked to double) which they avoid by being careful about how constadds are used in their larger algorithms.

src/group_impl.h
apoelstra: If you'd like me to remove your name, I will. But I agree with @gmaxwell that it's probably good for readers that it be there.

gmaxwell: Actually I was saying that people will find it if it's just in the commit message rather than a comment in the code.

apoelstra: Oh :) well if you're both against me I'll take it out of the source comment.

sipa: I think we eentually want some AUTHORS document that explains who contributed what - that's easier to access than digging through commit messages, and less invasive than putting it in the code.

Some documentation in the source tree that explains the various tricks and optimizations implemented would be even nicer.

sipa:
ACK

PR #263, opened 2015-06-26
Add secp256k1_ec_pubkey_compress()

afk11:
Added general function to reserialize a public key, and added secp256k1_ec_pubkey_compress() after seeing https://github.com/bitcoin/secp256k1/issues/259

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -260,10 +260,27 @@ SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_create(
   int compressed
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
 
+/** Compress a public key.
+ * In:     ctx:       pointer to a context object (cannot be NULL)
+ *         pubkeyin:  pointer to a 33-byte or 65-byte public key (cannot be NULL)
+ * Out:    pubkeyout: pointer to a 33-byte array to put the compressed public key (cannot be NULL)
+ *                    May alias pubkeyin.
+ *         pubkeylen: pointer to the size of the public key pointed to by pubkeyin (cannot be NULL)
+ *                    It will be updated to reflect the size of the public key in pubkeyout.
+ * Returns: 0: pubkeyin was invalid
+ *          1: pubkeyin was valid, and pubkeyout is its compressed version
+ */
+SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_compress(
+  const secp256k1_context_t* ctx,
+  const unsigned char *pubkeyin,
+  unsigned char *pubkeyout,
+  int *pubkeylen
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
+
 /** Decompress a public key.
  * In:     ctx:       pointer to a context object (cannot be NULL)
- * In:     pubkeyin:  pointer to a 33-byte or 65-byte public key (cannot be NULL)
- * In/Out: pubkeyout: pointer to a 65-byte array to put the decompressed public key (cannot be NULL)
+ *         pubkeyin:  pointer to a 33-byte or 65-byte public key (cannot be NULL)
+ * Out:    pubkeyout: pointer to a 65-byte array to put the decompressed public key (cannot be NULL)
  *                    May alias pubkeyin.
  *         pubkeylen: pointer to the size of the public key pointed to by pubkeyin (cannot be NULL)
  *                    It will be updated to reflect the size of the public key in pubkeyout.
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -285,6 +285,21 @@ int secp256k1_ec_pubkey_decompress(const secp256k1_context_t* ctx, const unsigne
     return ret;
 }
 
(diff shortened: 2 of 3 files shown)


afk11:
I'm not yet sure why travis failed, it was a seemingly unrelated check. Will take another look.

afk11:
Ok - nothing major here, just added a function that does the opposite of secp256k1_ec_pubkey_decompress(). 

Ended up removing the new function in eckey_impl.h - wasn't really worth it since the code is just a few lines. 

Anyways, hope this proves useful enough to merge. I wouldn't push C code too often - any feedback/criticism would be welcome.

I added a test also, similar to how decompress is tested here: https://github.com/bitcoin/secp256k1/pull/263/files#diff-4655d106bf03045a3a50beefc800db21R1521

src/eckey.h
```diff
@@ -14,6 +14,7 @@
 
 static int secp256k1_eckey_pubkey_parse(secp256k1_ge_t *elem, const unsigned char *pub, int size);
 static int secp256k1_eckey_pubkey_serialize(secp256k1_ge_t *elem, unsigned char *pub, int *size, int compressed);
+static int secp256k1_eckey_pubkey_reserialize( const unsigned char *pub, unsigned char *pubkeyout, int *size, int compressed);
```
sipa: This function does not seem to exist.

sipa:
Please squash your last commit into the previous ones. All commits should result in a tree that passes tests, or we may break the ability to bisect.

src/eckey.h
afk11: Thanks, removed it now.

afk11:
Ok, just pushed them once more.

src/eckey.h
```diff
@@ -24,3 +24,4 @@ static int secp256k1_eckey_privkey_tweak_mul(secp256k1_scalar_t *key, const secp
 static int secp256k1_eckey_pubkey_tweak_mul(const secp256k1_ecmult_context_t *ctx, secp256k1_ge_t *key, const secp256k1_scalar_t *tweak);
 
 #endif
+
```
apoelstra: Can you rebase without this newline?

Normally I wouldn't complain but this is the only place you touch this file.

src/secp256k1.c
```diff
@@ -418,3 +433,4 @@ int secp256k1_context_randomize(secp256k1_context_t* ctx, const unsigned char *s
     secp256k1_ecmult_gen_blind(&ctx->ecmult_gen_ctx, seed32);
     return 1;
 }
+
```
apoelstra: And this one, while you're rebasing anyway.

include/secp256k1.h
```diff
@@ -260,6 +260,23 @@ SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_create(
   int compressed
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
 
+/** Compress a public key.
+ * In:     ctx:       pointer to a context object (cannot be NULL)
+ * In:     pubkeyin:  pointer to a 33-byte or 65-byte public key (cannot be NULL)
```
apoelstra: Typo: drop second `In:`

src/secp256k1.c
```diff
@@ -285,6 +285,21 @@ int secp256k1_ec_pubkey_decompress(const secp256k1_context_t* ctx, const unsigne
...
+int secp256k1_ec_pubkey_compress(const secp256k1_context_t* ctx, const unsigned char *pubkeyin, unsigned char *pubkeyout, int *pubkeylen) {
+    secp256k1_ge_t p;
+    int ret = 0;
+    DEBUG_CHECK(pubkeyin != NULL);
+    DEBUG_CHECK(pubkeyout != NULL);
+    DEBUG_CHECK(pubkeylen != NULL);
+    (void)ctx;
+
+    if (secp256k1_eckey_pubkey_parse(&p, pubkeyin, *pubkeylen)) {
+        ret = secp256k1_eckey_pubkey_serialize(&p, pubkeyout, pubkeylen, 1);
```
apoelstra: I wonder if we should `VERIFY_CHECK` this instead of returning the return value. Is there a way this can fail if the parsing was successful?

sipa: Nope.

sipa:
ACK, though addressing @apoelstra's nits would be fine.

src/eckey.h
afk11: Woops - hadn't spotted that, thanks.

afk11:
Cool, rebased and addressed the above!

apoelstra:
Tested ACK from me. (The `VERIFY_CHECK` thing is no big deal.)

include/secp256k1.h
```diff
@@ -260,9 +260,26 @@ SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_create(
   int compressed
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
 
+/** Compress a public key.
+ * In:     ctx:       pointer to a context object (cannot be NULL)
+ *         pubkeyin:  pointer to a 33-byte or 65-byte public key (cannot be NULL)
+ * In/Out: pubkeyout: pointer to a 33-byte array to put the compressed public key (cannot be NULL)
```
sipa: pubkeyout is Out only no?

afk11: You're right. Could still be the same for decompress as well, I'll take a look now.

Issue #265, opened 2015-07-04
Are _fe_cmov methods handling VERIFY fields correctly?


peterdettman:
The current implementation just masks the `magnitude` and `normalized` fields from one of the inputs, but I would have expected both to get "worst-case" values, i.e. `magnitude` becomes the maximum of both inputs', and `normalized` becomes the AND of both inputs'.

PR #266, opened 2015-07-04
Fix VERIFY calculations in _fe_cmov methods

peterdettman:
Proposed changes to address #265.

### DIFF
--- a/src/field_10x26_impl.h
+++ b/src/field_10x26_impl.h
@@ -1083,8 +1083,10 @@ static SECP256K1_INLINE void secp256k1_fe_cmov(secp256k1_fe_t *r, const secp256k
     r->n[8] = (r->n[8] & mask0) | (a->n[8] & mask1);
     r->n[9] = (r->n[9] & mask0) | (a->n[9] & mask1);
 #ifdef VERIFY
-    r->magnitude = (r->magnitude & mask0) | (a->magnitude & mask1);
-    r->normalized = (r->normalized & mask0) | (a->normalized & mask1);
+    if (a->magnitude > r->magnitude) {
+        r->magnitude = a->magnitude;
+    }
+    r->normalized &= a->normalized;
 #endif
 }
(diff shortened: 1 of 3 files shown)


sipa:
Untested ACK

apoelstra:
tested ACK

src/field_10x26_impl.h
```diff
@@ -1083,8 +1083,10 @@ static SECP256K1_INLINE void secp256k1_fe_cmov(secp256k1_fe_t *r, const secp256k
     r->n[8] = (r->n[8] & mask0) | (a->n[8] & mask1);
     r->n[9] = (r->n[9] & mask0) | (a->n[9] & mask1);
 #ifdef VERIFY
-    r->magnitude = (r->magnitude & mask0) | (a->magnitude & mask1);
-    r->normalized = (r->normalized & mask0) | (a->normalized & mask1);
+    if (a->magnitude > r->magnitude) {
+        r->magnitude = a->magnitude;
+    }
```
dcousens: `r->magnitude = MAX(a->magnitude, r->magnitude)`

Is clearer IMHO, given the short as hell variable names.
The compiler probably already makes the branchless optimization though,  but conceptually it could be faster too.

peterdettman:
@dcousens I couldn't see any existing usage of min/max in the code, so I just kept it simple here. Not sure it's important for VERIFY checks anyway, but I think it's pretty common for the compiler to eliminate any branch here (at least when optimizations are enabled).

sipa:
MAX is not standard C afaik? we can define it ourselves, of course.

I don't care about performance inside VERIFY. It should be correct, and readable. Usin MAX/MIN helps there, but I think we should do perhaps introduce those independently in another PR if we want it.

dcousens:
@sipa sure,  I see no reason to hold up the PR for it.  Just wanted to flag that is all.

PR #269, opened 2015-07-08
Change RFC6979 implementation to be a generic PRNG

sipa:
Instead of making it take key, message, and extra data separate, just pass in a byte array of seed data, turning it into just a non-EC specific PRNG.

The usage inside the blinding code and the tests is also adapted to make use of this in a more natural way.


sipa:
@gmaxwell Want to review?

dcousens:
@sipa maybe its taken care by `secp256k1_rfc6979_hmac_sha256_initialize`,  but `RFC6979` specifies that you

> repeat until T is within the interval [1, n - 1] and is suitable for ECDSA

Where suitable for `ECDSA` means the `r` and `s` values are not `=== 0`.
Is this handled externally as part of the libraries signing process?
In which case, I assume you feed `T` back in as the seed or?

sipa:
@dcousens Right, secp256k1_rfc6979_hmac_sha256 only implements the PRNG side of the construction. The specific use as defined by RFC6979 is done inside the signing code.

apoelstra:
With benchmarks on
```
src/bench_internal.c: In function ‘bench_rfc6979_hmac_sha256’:
src/bench_internal.c:268:9: error: too many arguments to function ‘secp256k1_rfc6979_hmac_sha256_initialize’
         secp256k1_rfc6979_hmac_sha256_initialize(&rng, data->data, 32, data->data, 32, NULL, 0);
         ^
In file included from src/bench_internal.c:11:0:
src/hash_impl.h:205:13: note: declared here
 static void secp256k1_rfc6979_hmac_sha256_initialize(secp256k1_rfc6979_hmac_sha256_t *rng, const unsigned char *key, size_t keylen) {
             ^
make: *** [src/bench_internal-bench_internal.o] Error 1
```

sipa:
Rebased and fixed bug found by @apoelstra.

gmaxwell:
ACK. (but see nit)

apoelstra:
tested ACK

sipa:
Addressed nit.

PR #270, opened 2015-07-09
Verify the result of GMP modular inverse using non-GMP code

sipa:
This hides potential GMP bugs from the verification.


gmaxwell:
Oh no! A ~1/10th of 1% verify slowdown measured here. :)

Looks good to me, and I verified that if I break GMP that it detects it for each of the cases. Spun it up in some more testsing; just for good measure.

gmaxwell:
ACK.

We had some discussion in IRC about a fancier way to do fatal error handling which might be useful in the future.

PR #271, opened 2015-07-11
Improve performance of _ecmult_wnaf

peterdettman:
- Track carry explicitly instead of adding to scalar
- Branch-free code for carry calculations

Gives ~0.6% improvement for bench_verify (64bit, endo=yes)

### DIFF
--- a/src/bench_internal.c
+++ b/src/bench_internal.c
@@ -229,7 +229,7 @@ void bench_ecmult_wnaf(void* arg) {
     bench_inv_t *data = (bench_inv_t*)arg;
 
     for (i = 0; i < 20000; i++) {
-        secp256k1_ecmult_wnaf(data->wnaf, &data->scalar_x, WINDOW_A);
+        secp256k1_ecmult_wnaf(data->wnaf, 256, &data->scalar_x, WINDOW_A);
         secp256k1_scalar_add(&data->scalar_x, &data->scalar_x, &data->scalar_y);
     }
 }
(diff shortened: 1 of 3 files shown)


sipa:
ACK, nice!

peterdettman:
With latest changes, perf. increased by ~2% total for bench_verify (64bit, endo=yes), maybe half that for endo=no, but...

Note that the tests are currently failing for endo=no, because of the VERIFY_CHECK for w <= 15; when endo=no, _ecmult_wnaf is called for WINDOW_G==16. There is a bug lurking here because 'int' is only required to be 16 bits AFAIK, and if it is in fact 16, _ecmult_wnaf wouldn't work for w > 15 (thus the check). Note that the bug is independent of this PR.

I guess the simplest fix is to change the element type of 'wnaf' to int32_t. Alternately, we could make it explicitly int16_t and just set WINDOW_G to 15 in all cases. Thoughts?

gmaxwell:
The code is not 16 bit safe in a number of places. I've fixed things that I've touched, but an effort should be made at some point to fix that.   The current requirement for the availability of a 64-bit long long though keeps me from simply just trying to build it for a 16 bit platform. :)

peterdettman:
OK, I just changed the check to w <= 31 for now, to reflect the effective assumption of 32-bit int.

apoelstra:
Tested ACK, though if others feel we should react more seriously to the 16-bit incompatibility I'll withdraw it :)

Very good stuff, thanks Peter!

peterdettman:
I reconsidered the "external initialisation of wnaf" change (rebased out that version) and changed to a memset inside _ecmult_wnaf. Still ~2% improvement.

sipa:
ACK

PR #273, opened 2015-07-13
Add missing casts to SECP256K1_FE_CONST_INNER

apoelstra:
This corrects for compilers where `int` is too small to allow constants to be shifted as much as they are in the `SECP256K1_FE_CONST_INNER` macro.

On my x86_64 system this change results in a bit-for-bit identical test binary, with both 32- and 64-bit field and scalars.


sipa:
ACK

PR #275, opened 2015-07-15
Fix build with static ecmult altroot and make dist.

gmaxwell:
Gen_context was unable to find the required headers without some
 autotools fixups.  Make dist was also broken without the extra
 sources for the host side table builder utility.


gmaxwell:
Fixes build problem for #274.

PR #278, opened 2015-07-18
Introduce callback functions for dealing with errors

sipa:


sipa:
Made some changes.

apoelstra:
Maybe when `VERIFY` is defined, the callback should be called even when the condition passes (with some sort of "success" flag, or a different callback altogether). What I'd like is for the test binary to be able to detect if an `ARG_CHECK` is defined but no never gets tripped.

sipa:
Will rebase/update this after #282.

sipa:
Rebased.

I haven't incorporated @apoelstra's change. I like the idea (I think an extra callback entry in the context is best), but I'm not exactly sure about the implementation. How would you identify a particular ARG_CHECK? Would they need a number you can test for?

apoelstra:
I'm not sure. I don't want to require a number for each one because that'd be awful to maintain and result in a lot of boilerplate code.

I think using the stringification of the ARG_CHECK should be sufficient. This is also a maintenance burden if specific checks are ever referred to, but I don't think this will happen. To detect an untriggered check it suffices to maintain an array (hash table, I guess :/, if the checks are identified by string) of number-of-times-triggered counts for each check.

sipa:
@apoelstra I think all of that can wait. Being able to catch errors (rather than their absence) from the tests is much easier and useful on itself.

include/secp256k1.h
```diff
@@ -74,6 +74,37 @@ void secp256k1_context_destroy(
...
+ *  an API call. The philosophy is that these shouldn't be dealt with through a
+ *  specific return value, as calling code should not have branches to deal with
+ *  the case that this code itself is broken.
+ *  On the other hand, during debug stage, one would want to be informed about
+ *  such mistakes, and the default (crashing) may be inadvisable.
+ *  When this callback is triggered, the API function called is guaranteed not
+ *  to cause a crash, though its return value and output arguments are
+ *  undefined.
+ */
+void secp256k1_context_set_illegal_callback(
```
apoelstra: I think you should comment that these two functions are not threadsafe, since everything else in the library is (assuming you lock your contexts appropriately).

Alternately, we could gain threadsafety by storing these in the secp256k1 contexts. I think this is a bad idea since it'd clutter up the API and almost nobody will be using these functions, just throwing it out there.

sipa: Adding all properties of a context to their constructor is not really feasible...

apoelstra: My comment was based on a misunderstanding anyway. Sorry.

apoelstra:
Tested ACK, bearing in mind that the new functionality is not actually (yet) tested.

PR #280, opened 2015-07-19
Update configure.ac to make it build on OpenBSD

mlmikael:
This update is to make libsecp256k1 build on OpenBSD (more specifically OpenBSD 5.7 with Autotools 2.69).

Without the "AM_PROG_CC_C_O" line in configure.ac, ./autogen.sh crashes with "Makefile.am: C objects in subdir but `AM_PROG_CC_C_O' not in `configure.ac'\nautoreconf-2.69: automake failed with exit status: 1".

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -23,6 +23,8 @@ if test "x$CFLAGS" = "x"; then
   CFLAGS="-O3 -g"
 fi
 
+AM_PROG_CC_C_O
+
 AC_PROG_CC_C89
 if test x"$ac_cv_prog_cc_c89" = x"no"; then
   AC_MSG_ERROR([c89 compiler support required])


sipa:
@theuni comments?

mlmikael:
Further research shows that this was not an OpenBSD version problem (and not an Autotools version problem, as I understand it, also), but an Automake version problem: "AM_PROG_CC_C_O" is supported at first in Automake version 1.14, and that's a fairly recent version, and a lot of people may have older Automake versions.

In particular, it's extremely confusing to get this error as generally which Automake version is required is not stated overly clearly in a project.

(This is so much of an accessory library, that I'd say it'd be kind of odd to have particular Automake version requirements though.)

mlmikael:
(edited the comment above, please see it by visiting the github page.)

theuni:
That macro is deprecated, but I don't see the harm in adding it if it fixes a build. Looks like worst-cast it's a nop.

mlmikael:
Great - you can just write it as:
>  
>  # Legacy, for compatibility with Automake versions prior to 1.14
>  AM_PROG_CC_C_O
>

PR #283, opened 2015-07-26
Actually test static precomputation in Travis

sipa:


theuni:
ACK. Not sure what happened there.

PR #284, opened 2015-07-31
Gen context make

theuni:
BUILT_SOURCES simply forces the header to be built first when no target is specified.

Use real dependencies instead. This fixes builds with dependency tracking disabled where a real target is specified.

Also fixup the build so that "gen_context.exe" is used for windows.


sipa:
Looks good, thanks!

gmaxwell:
Looks good to me. utACK.

PR #289, opened 2015-08-20
API documentation improvements

sipa:


dcousens:
Reviewed with [w=0?](https://github.com/bitcoin/secp256k1/pull/289/files?w=0).

utACK

include/secp256k1.h
```diff
@@ -361,60 +419,113 @@ SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_seckey_verify(
...
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
 
-/** Import a private key in DER format. */
+/** Import a private key in DER format.
+ * Returns: 1 if a private key was extracted.
+ * In:  ctx:        pointer to a context object (cannot be NULL).
+ *      privkey:    pointer to a private key in DER format (cannot be NULL).
+ *      privkeylen: length of the DER private key pointed to be privkey.
+ * Out: seckey:     pointer to a 32-byte array for storing the private key.
+ *                  (cannot be NULL).
```
dcousens: Any reason this functions in/out parameters are not in the same order as the documentation?

include/secp256k1.h
```diff
@@ -336,24 +392,26 @@ int secp256k1_ecdsa_sign(
  *  Out:     pubkey:     pointer to the recoved public key (cannot be NULL)
  */
 SECP256K1_WARN_UNUSED_RESULT int secp256k1_ecdsa_recover(
-  const secp256k1_context_t* ctx,
-  const unsigned char *msg32,
-  const secp256k1_ecdsa_signature_t *sig,
-  secp256k1_pubkey_t *pubkey
+    const secp256k1_context_t* ctx,
+    const unsigned char *msg32,
+    const secp256k1_ecdsa_signature_t *sig,
```
dcousens: Documentation calls this `sig64`,  its `sig` here.

src/bench_internal.c
```diff
@@ -282,6 +283,22 @@ void bench_rfc6979_hmac_sha256(void* arg) {
...
+    }
+}
+
+void bench_context_sign(void* arg) {
+    (void)arg;
+    int i;
+    for (i = 0; i < 200; i++) {
+        secp256k1_context_destroy(secp256k1_context_create(SECP256K1_CONTEXT_SIGN));
+    }
+}
```
dcousens: For posterity, it should probably be mentioned this PR also adds 2 new benchmarks.

include/secp256k1.h
```diff
@@ -5,6 +5,83 @@
...
 
+/** Opaque data structure that holds context information (precomputed tables etc.).
+ *
+ *  The purpose of context structures is to cache large precomputed data tables
+ *  that are expensive to construct, and also to maintain the randomization data
+ *  for blinding.
+ *
+ *  Do not create a new context object for each operation, as construction is
+ *  far slower than all other API calls (~100 times slower than an ECDSA
+ *  verification), and then use the context for all other operations.
```
apoelstra: I think part of this sentence is missing; you say "do not do this .. and then" rather than "do not do this .. instead do this, and then".

include/secp256k1.h
```diff
@@ -5,6 +5,83 @@
...
+ *
+ *  A constructed context can safely be used from multiple threads
+ *  simultaneously, but some API calls need exclusive access, in particular
+ *  secp256k1_context_destroy and secp256k1_context_randomize.
```
apoelstra: Do we want to drop the comment that `const` makes the difference between requiring exclusive access and not?

include/secp256k1.h
sipa: I'd to have a convention of having all functions AND all documentation to have order: ctx/out/inout/in or ctx/in/inout/out.

sipa:
Addressed nits.

src/bench_internal.c
sipa: Done, split into a separate commit.

include/secp256k1.h
dcousens: I'm personally in favour of ctx/in/inout/out

include/secp256k1.h
dcousens: ping @sipa,  missed nit or ?

include/secp256k1.h
```diff
@@ -336,24 +393,26 @@ int secp256k1_ecdsa_sign(
  *  Out:     pubkey:     pointer to the recoved public key (cannot be NULL)
```
dcousens: ^`sig64`

include/secp256k1.h
sipa: Fixed.

include/secp256k1.h
sipa: There is a convention within C APIs to use out/in order, to mimic assignment.

dcousens: I guess that makes sense,  but,  I wouldn't be able to change the habit of reading things in terms of INPUT / OUTPUT (IO).

It'd be similar to seeing an anonymous function callback being used with the callback passed before arguments to the function.

apoelstra: I prefer out/in order, it is extremely common in C API's.

src/bench_internal.c
```diff
@@ -282,6 +283,22 @@ void bench_rfc6979_hmac_sha256(void* arg) {
...
+    (void)arg;
+    int i;
+    for (i = 0; i < 20; i++) {
+        secp256k1_context_destroy(secp256k1_context_create(SECP256K1_CONTEXT_VERIFY));
+    }
+}
+
+void bench_context_sign(void* arg) {
+    (void)arg;
+    int i;
```
apoelstra: I'm getting a "mixed declarations and code" warning for having `int i` after the `(void)arg`. Ditto in the above function.

sipa: Fixed.

apoelstra:
ACK

gmaxwell:
ACK.

include/secp256k1.h
```diff
@@ -5,6 +5,84 @@
...
+} secp256k1_pubkey_t;
+
+/** Opaque data structured that holds a parsed ECDSA signature, optionally
+ *  supporting pubkey recovery.
+ *
+ *  The exact representation of data inside is implementation defined and not
+ *  guaranteed to be portable between different platforms or versions. It is
+ *  however guaranteed to be 65 bytes in size, and can be safely copied/moved.
+ *  If you need to convert to a format suitable for storage or transmission, use
+ *  the secp256k1_ecdsa_signature_serialize_* and
```
droark: After-the-fact nit: "secp256k1_ecdsa_signature_serialize_*" appears twice. I'm guessing this is a mistake.

PR #293, opened 2015-08-27
[API BREAK] Change argument order to out/outin/in

sipa:
This introduces a strict argument order for all API functions (including modules), explained in include/secp256k1.h.

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -5,6 +5,21 @@
 extern "C" {
 # endif
 
+/* These rules specify the order of arguments in API calls:
+ *
+ * 1. Context pointers go first, followed by output arguments, combined
+ *    output/input arguments, and finally input-only arguments.
+ * 2. Array lengths always immediately the follow the argument whose length
+ *    they describe, even if this violates rule 1.
+ * 3. Within the OUT/OUTIN/IN groups, pointers to data that is typically generated
+ *    later go first. This means: signatures, public nonces, private nonces,
+ *    messages, public keys, secret keys, tweaks.
+ * 4. Arguments that are not data pointers go last, from more complex to less
+ *    complex: function pointers, algorithm names, messages, void pointers,
+ *    counts, flags, booleans.
+ * 5. Opaque data pointers follow the function pointer they are to be passed to.
+ */
+
 /** Opaque data structure that holds context information (precomputed tables etc.).
  *
  *  The purpose of context structures is to cache large precomputed data tables
@@ -59,15 +74,15 @@ typedef struct {
 /** A pointer to a function to deterministically generate a nonce.
  *
  * Returns: 1 if a nonce was successfully generated. 0 will cause signing to fail.
+ * Out:     nonce32:   pointer to a 32-byte array to be filled by the function.
  * In:      msg32:     the 32-byte message hash being verified (will not be NULL)
  *          key32:     pointer to a 32-byte secret key (will not be NULL)
  *          algo16:    pointer to a 16-byte array describing the signature
  *                     algorithm (will be NULL for ECDSA for compatibility).
+ *          data:      Arbitrary data pointer that is passed through.
  *          attempt:   how many iterations we have tried to find a nonce.
  *                     This will almost always be 0, but different attempt values
  *                     are required to result in a different nonce.
- *          data:      Arbitrary data pointer that is passed through.
- * Out:     nonce32:   pointer to a 32-byte array to be filled by the function.
  *
  * Except for test cases, this function should compute some cryptographic hash of
  * the message, the algorithm, the key and the attempt.
@@ -77,8 +92,8 @@ typedef int (*secp256k1_nonce_function_t)(
     const unsigned char *msg32,
     const unsigned char *key32,
     const unsigned char *algo16,
-    unsigned int attempt,
-    const void *data
+    const void *data,
+    unsigned int attempt
 );
 
 # if !defined(SECP256K1_GNUC_PREREQ)
@@ -132,7 +147,7 @@ secp256k1_context_t* secp256k1_context_create(
 /** Copies a secp256k1 context object.
  *
  *  Returns: a newly created context object.
- *  In:      ctx: an existing context to copy (cannot be NULL)
+ *  Args:    ctx: an existing context to copy (cannot be NULL)
  */
 secp256k1_context_t* secp256k1_context_clone(
     const secp256k1_context_t* ctx
@@ -141,7 +156,7 @@ secp256k1_context_t* secp256k1_context_clone(
 /** Destroy a secp256k1 context object.
  *
  *  The context pointer may not be used afterwards.
- *  In:     ctx: an existing context to destroy (cannot be NULL)
+ *  Args:   ctx: an existing context to destroy (cannot be NULL)
  */
 void secp256k1_context_destroy(
     secp256k1_context_t* ctx
@@ -161,10 +176,10 @@ void secp256k1_context_destroy(
  *  to cause a crash, though its return value and output arguments are
  *  undefined.
  *
- *  In:   ctx: an existing context object (cannot be NULL)
- *        fun: a pointer to a function to call when an illegal argument is
- *             passed to the API, taking a message and an opaque pointer
- *             (cannot be NULL).
+ *  Args: ctx:  an existing context object (cannot be NULL)
+ *  In:   fun:  a pointer to a function to call when an illegal argument is
+ *              passed to the API, taking a message and an opaque pointer
+ *              (cannot be NULL).
  *        data: the opaque pointer to pass to fun above.
  */
 void secp256k1_context_set_illegal_callback(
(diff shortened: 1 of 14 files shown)


dcousens:
concept ACK

sipa:
Rebased.

DavidEGrayson:
Here is an excerpt from the secp256k1.h in this pull request:

~~~~
General ordering of arguments in API calls:

- context pointer.
- output-only arguments.
- output/input arguments.
- input-only arguments.

Within each group, argument types are ordered based on their type:
- Counts of other arguments always go immdiately before those arguments.
- Serializations
- Serialization lengths go immediately after the serialization, even if
  that violates the output/input order.
- Signatures
- Public keys
- Secret keys
- Tweaks
- Messages
- Public nonces
- Private nonces
- Nonce functions
- Other function pointers
- Opaque data pointers follow the function pointer they are to be passed to.
- Algorithm names
- Attempt number
- Flags
~~~~

With the long list of argument types that does not come with any justification, this scheme sounds pretty arbitrary and hard to remember.  I would propose a simpler scheme based on chronology:

~~~~
These rules specify the ordering of arguments in API calls:

1. Context pointers come first, and are exempt from all other rules.
2. Buffer/array lengths are immediately after the argument they describe,
   and are exempt from all other rules.
3. Opaque data pointers are immediately after the function pointer they
   are passed to, and are exempt from all other rules.
4. Arguments are ordered chronologically: data that was generated earlier
   comes first.
5. When there is no clear chronological ordering, use these additional
   arbitrary rules to break ties:
    1. Flags, algorithm names, and function pointers come last.
    2. For IN/OUT parameters, place the parameter as if it were just an
       input.

~~~~

I think it is natural for humans to assign chronological orders to things, and it is natural (in English documents like this API) to order things from left to right according to that chronology.  This means outputs will always be last, which is roughly the opposite of what is proposed in this pull request.

Not every user story has the same chronology though.  For example, with secp256k1_ec_pubkey_serialize, maybe the user planned ahead of time to always set `compressed` to 0 when serializing a public key, in which case they would want flags to come before the public key.  Alternatively, maybe they always generate both compressed and uncompressed versions of every public key, so the decision of which value to use for `compressed` in the current function call really was made after the public key existed.  That's why we need some arbitrary rules like 5.1 to clear up ambiguity.

So here is what the API functions and their argument order would look like if we applied the rules from this chronological scheme:

~~~~
secp256k1_nonce_function_type(IN key32, IN msg32, IN data, IN algo16, IN attempt, OUT nonce32)
secp256k1_context_set_illegal_callback(ctx, IN fun, IN data)
secp256k1_context_set_error_callback(ctx, IN fun, IN data)
secp256k1_ec_pubkey_parse(ctx, IN input+length, OUT pubkey)
secp256k1_ec_pubkey_serialize(ctx, IN pubkey, IN compressed, OUT output+length)
secp256k1_ecdsa_signature_parse_der(ctx, IN input+length, OUT sig)
secp256k1_ecdsa_signature_serialize_der(ctx, IN sig, OUT output+length)
secp256k1_ecdsa_verify(ctx, IN pubkey, IN msg, IN sig)
secp256k1_ecdsa_sign(ctx, IN seckey, IN msg32, IN noncefp+data, OUT sig)
secp256k1_ec_seckey_verify(ctx, IN seckey)
secp256k1_ec_pubkey_create(ctx, IN seckey, OUT pubkey)
secp256k1_ec_privkey_export(ctx, IN seckey, IN compressed, OUT privkey+length)
secp256k1_ec_privkey_import(ctx, IN privkey+length, OUT seckey)
secp256k1_ec_privkey_tweak_add(ctx, INOUT seckey, IN tweak)
secp256k1_ec_pubkey_tweak_add(ctx, INOUT pubkey, IN tweak)
secp256k1_ec_privkey_tweak_mul(ctx, INOUT seckey, IN tweak)
secp256k1_ec_pubkey_tweak_mul(ctx, INOUT pubkey, IN tweak)
secp256k1_context_randomize(ctx, IN seed32)
secp256k1_ec_pubkey_combine(ctx, IN pubkeyins+length, OUT pubkey)
~~~~

dcousens:
@DavidEGrayson that is exactly what I would have expected from a modern API.  C conventions aside.

DavidEGrayson:
Thanks @dcousens!  I had to make a few arbitrary decisions that weren't really backed up by any principles, so I'm glad it exactly matches what you had in mind.

sipa:
@DavidEGrayson That's pretty close to what I was aiming for, only the other way around. I'm aiming for following "out = f(in)" as much as possible, and retain that argument order when earlier outs become ins (for example: signing will do sig = f(key,msg), then later verification takes f(sig,[pub]key,msg). Perhaps I should have given the principles behind that ordering.

dcousens:
IMHO, it also lends it self to a more formal type notation too.  eg, using Haskell type grammar

```
secp256k1_ec_privkey_import(ctx, IN privkey+length, OUT seckey)
secp256k1_ec_privkey_import :: ctx -> d -> secret 
```

DavidEGrayson:
@sipa I thought that usually the keys (or at least an algorithm for generating keys) is created before the message, so the "sig, key, msg" ordering looks off to me even if you are going for a reverse ordering.  With GPG you make a key pair before signing things with it.  With Bitcoin, you use the public key to receive funds before you generate transactions (messages) to use those funds.

dcousens:
@DavidEGrayson

> I'm aiming for following "out = f(in)" as much as possible, and retain that argument order when earlier outs become ins

That is,  composability of arguments comes from the front,  which makes sense if you look at how you would do it if you were following the: `secp256k1_ec_privkey_import :: secret <- (context -> privateKey)` grammar.

dcousens:
@sipa though,  given this reasoning,  why is the context at the front?

sipa:
Because every function has a context it seems reasonable to put it first.
It's also not really an input or an output. Compare it with the object in
OO on which you invoke methods.

Having the key before the message makes sense.

dcousens:
> Because every function has a context it seems reasonable to put it first.

Well,  I guess it depends on how you deep you want to take that rabbit hole,  by no means do functions actually need a context.  We just choose that for optimization purposes because instantiating it on every call would be ridiculous.

sipa:
Sure, it's just a rationalization. Any order is fine, IMHO, and I think we
should just aim for a consistent and easily understandable one. I like the
"order of generation" one for that reason.

I mostly pick "first out, then in" because that's what most are already
doing (and it's not reasonable, as it is common in C APIs).

dcousens:
>  Any order is fine, IMHO, and I think we should just aim for a consistent and easily understandable one

Of course,  this was just bike shedding,  what is most important is consistency.
The devil isn't in the API,  its in the implementation details [for this library].

It would be nice if you could even put it under test,  though I'm not sure how what C tooling is like in that regard.

dcousens:
> easily understandable one

That is always going to be subjective,  as noted above,  myself and @DavidEGrayson read things (chronologically) in the reverse order to your proposal.

dcousens:
What might be an important consideration, is what background do you think most programmers reviewing / using this library will have?
If it is a C background, then the `out/in` will cater to the crowd;  however, if you think most people are going to use a modern language,  it might be more likely an `in/out` will be less disruptive.

sipa:
No reason a wrapper can't use another order.

apoelstra:
@dcousens most people developing the software have a C background ;)

I don't really want to participate in a bikeshedding thread but for what it's worth my intuition matches @sipa's. Also I do strongly prefer out-then-in; none of my other preferences are strong.

voisine:
Out-then-in is nice because it matches the ordering of assignments. a = b

gmaxwell:
This is the most common convention than C; indeed, because it matches assignment.

Probably consistency is more important than external convention, but the conventional ordering is good.

sipa:
Updated to a "out-then-in", and within each group a "later-then-earlier" for data arguments, and "complex-to-simple" for the rest policy.

sipa:
Also changed the order of the documentation for each function to match the actual argument order.

dcousens:
> Updated to a "out-then-in", and within each group a ("later-then-earlier" for data arguments, and "complex-then-simple" for the rest) policy.

ACK,  thanks for simplifying it :+1:

dcousens:
> @dcousens most people developing the software have a C background ;)

@apoelstra and assuredly so will those writing the wrappers,  but I get the feeling this library will see a lot of usage across language boundaries,  and I'd be surprised if I didn't see these particular remnants of the API translating into other languages.
Food for thought, but LGTM.

DavidEGrayson:
The new, simpler rules are looking good.  But I think it's a bit odd that rules 2 and 3 are separate:

~~~~
2. Buffer array lengths always immediately the follow the argument whose
    length they describe, even if this violates rule 1.
3. Counts of batches of arguments always go before that batch.
~~~~

Rule 3 seems to affect the ordering of arguments to `secp256k1_ec_pubkey_combine`:

~~~~
SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_combine(
    const secp256k1_context_t* ctx,
    secp256k1_pubkey_t *out,
    int n,
    const secp256k1_pubkey_t * const * ins
) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(4);
~~~~

However, the documentation says `ins` is a "pointer to array", so I would think we should apply Rule 2.  This means `n` would come last.

Unless you are going to use C's variadic arguments, I don't see how Rule 3 would ever really apply, and it can probably be removed.

sipa:
I wasn't considering ins as a buffer (rule 2) but as a list of "data
pointer" arguments (rule 4), but I guess they can be unified.

sipa:
@DavidEGrayson Unified the batch size and buffer size rules, and updated the code accordingly.

PR #294, opened 2015-08-28
[API BREAK] Remove _t in types due to being POSIX-reserved

### LINKED ISSUE #198: _t is reserved

sipa:
Fixes #198.

Builds on top of #293.

Mechanical application of

`
for A in include/*.h src/*.h src/*.c src/modules/*/*.h; do sed -e 's/secp256k1_\([a-z_]*\)_t\([^a-zA-Z]\)/secp256k1_\1\2/g' <$A >/tmp/bla; mv /tmp/bla $A; sed -e 's/callback_t/secp256k1_callback/g' <$A >/tmp/bla; mv /tmp/bla $A; done
`

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -40,7 +40,7 @@ extern "C" {
  *  Regarding randomization, either do it once at creation time (in which case
  *  you do not need any locking for the other calls), or use a read-write lock.
  */
-typedef struct secp256k1_context_struct secp256k1_context_t;
+typedef struct secp256k1_context_struct secp256k1_context;
 
 /** Opaque data structure that holds a parsed and valid public key.
  *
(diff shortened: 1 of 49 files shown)


dcousens:
> for A in include/*.h src/*.h src/*.c src/modules/*/*.h; do sed -e  <$A >/tmp/bla; mv /tmp/bla $A; sed -e  <$A >/tmp/bla; mv /tmp/bla $A; done

vs

`sed -i 's/secp256k1_\([a-z_]*\)_t\([^a-zA-Z]\)/secp256k1_\1_type\2/g' `find include/ src/ | grep "\.\(c\|h\)$"`

sipa:
@dcousens I'll use that if I need to redo it :)

sipa:
Rebased.

apoelstra:
ACK without reading the code ("only-tested ACK?"); but I ran the tests binary and benchmarks on 512 different ./configure setups with no problems.

sipa:
Rebased.

sipa:
As suggested by @rustyrussell: just drop the _t instead of replacing by _type.

gmaxwell:
ACK

Issue #300, opened 2015-09-01
Bad callback logic


theuni:
Given that ARG_CHECK is defined as:
```c
#define ARG_CHECK(cond) do { \
    if (EXPECT(!(cond), 0)) { \
        ctx->illegal_callback.fn(#cond, ctx->illegal_callback.data); \
        return 0; \
    } \
} while(0)
```
These checks (17 of 'em from a quick grep) make no sense and are doomed to explode:
```c
ARG_CHECK(ctx != NULL);
```

gmaxwell:
Haha.

sipa:
Good catch!

sipa:
Fixed by #301

PR #301, opened 2015-09-01
bugfix: don't blow up on "ARG_CHECK(ctx != NULL)"

theuni:
The obvious fix for #300. I'm not sure if this should be handled in the same macro though.

### DIFF
--- a/src/modules/recovery/main_impl.h
+++ b/src/modules/recovery/main_impl.h
@@ -90,7 +90,7 @@ int secp256k1_ecdsa_sign_recoverable(const secp256k1_context_t* ctx, const unsig
     int ret = 0;
     int overflow = 0;
     unsigned int count = 0;
-    ARG_CHECK(ctx != NULL);
+    VERIFY_CHECK(ctx != NULL);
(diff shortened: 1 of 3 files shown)


apoelstra:
I think a better fix is to replace the affected `ARG_CHECK`s with `VERIFY_CHECK`. Or even to put a `VERIFY_CHECK` for `ctx != NULL` in the `ARG_CHECK` macro itself.

I think we'll get burned if we don't terminate the program on null context, since it's used inside of so many macros.

sipa:
Agree with @apoelstra. This will just make the accidental passing of
ctx==NULL harder to discover.

theuni:
Ok, fine by me. This was just the "fix it as it lies" change. I'd much prefer to handle the ctx logic separately from the other args.

apoelstra:
Can you replace this with a patch that replaces any `ARG_CHECK`s on `ctx` with `VERIFY_CHECK`? (If not I can do it.)

theuni:
Updated as suggested. Also moved one case where the context was checked after other arguments.

theuni:
Updated again. I accidentally left the original change to the ARG_CHECK macro in, gone now.

PR #302, opened 2015-09-03
Add sage verification script for the group laws

sipa:
This adds a SAGE script for verifying the implemented group laws.

The output should be:
```
Formula secp256k1_gej_add_var:
  add_infinite_ab:
    branch 0: OK
  double:
    branch 2: OK
  add:
    branch 4: OK
  add_infinite_b:
    branch 1: OK
  add_infinite_a:
    branch 0: OK
  add_opposite:
    branch 3: OK

Formula secp256k1_gej_add_ge_var:
  add_infinite_ab:
    branch 0: OK (assuming Bz - 1 = 0 [b.z=1])
  double:
    branch 2: OK (assuming Bz - 1 = 0 [b.z=1])
  add:
    branch 4: OK (assuming Bz - 1 = 0 [b.z=1])
  add_infinite_b:
    branch 1: OK (assuming Bz - 1 = 0 [b.z=1])
  add_infinite_a:
    branch 0: OK (assuming Bz - 1 = 0 [b.z=1])
  add_opposite:
    branch 3: OK (assuming Bz - 1 = 0 [b.z=1])

Formula secp256k1_gej_add_zinv_var:
  add_infinite_ab:
    branch 0: OK
  double:
    branch 2: OK
  add:
    branch 4: OK
  add_infinite_b:
    branch 0: OK
  add_infinite_a:
    branch 1: OK
  add_opposite:
    branch 3: OK

Formula secp256k1_gej_add_ge:
  add_infinite_ab:
  double:
    branch 0: OK (assuming Bz - 1 = 0 [b.z=1], Bi = 0 [b_finite])
  add:
    branch 0: OK (assuming Bz - 1 = 0 [b.z=1], Bi = 0 [b_finite])
    branch 2: OK (assuming Bz - 1 = 0 [b.z=1], Bi = 0 [b_finite])
    branch 3: OK (assuming Bz - 1 = 0 [b.z=1], Bi = 0 [b_finite])
  add_infinite_b:
  add_infinite_a:
    branch 4: OK (assuming Bz - 1 = 0 [b.z=1], Bi = 0 [b_finite])
    branch 6: OK (assuming Bz - 1 = 0 [b.z=1], Bi = 0 [b_finite])
    branch 7: OK (assuming Bz - 1 = 0 [b.z=1], Bi = 0 [b_finite])
    branch 13: OK (assuming Bz - 1 = 0 [b.z=1], Bi = 0 [b_finite])
  add_opposite:
    branch 9: OK (assuming Bz - 1 = 0 [b.z=1], Bi = 0 [b_finite])

Formula secp256k1_gej_add_ge_old [should fail]:
  add_infinite_ab:
  double:
    branch 0: OK (assuming Bz - 1 = 0 [b.z=1], Bi = 0 [b_finite])
  add:
    branch 0: OK (assuming Bz - 1 = 0 [b.z=1], Bi = 0 [b_finite])
    branch 2: FAIL, ['on_curve', 'finite_point', 'colinear_2', 'colinear_1', 'colinear_3'] fails (assuming Bz - 1 = 0 [b.z=1], Bi = 0 [b_finite])
  add_infinite_b:
  add_infinite_a:
    branch 1: OK (assuming Bz - 1 = 0 [b.z=1], Bi = 0 [b_finite])
  add_opposite:
    branch 2: OK (assuming Bz - 1 = 0 [b.z=1], Bi = 0 [b_finite])
```

### DIFF
--- a/sage/group_prover.sage
+++ b/sage/group_prover.sage
@@ -1,0 +1,322 @@
+# This code supports verifying group implementations which have branches
+# or conditional statements (like cmovs), by allowing each execution path
+# to independently set assumptions on input or intermediary variables.
+#
+# The general approach is:
+# * A constraint is a tuple of two sets of of symbolic expressions:
+#   the first of which are required to evaluate to zero, the second of which
+#   are required to evaluate to nonzero.
+#   - A constraint is said to be conflicting if any of its nonzero expressions
+#     is in the ideal with basis the zero expressions (in other words: when the
+#     zero expressions imply that one of the nonzero expressions are zero).
+# * There is a list of laws that describe the intended behaviour, including
+#   laws for addition and doubling. Each law is called with the symbolic point
+#   coordinates as arguments, and returns:
+#   - A constraint describing the assumptions under which it is applicable,
+#     called "assumeLaw"
+#   - A constraint describing the requirements of the law, called "require"
+# * Implementations are transliterated into functions that operate as well on
+#   algebraic input points, and are called once per combination of branches
+#   exectured. Each execution returns:
+#   - A constraint describing the assumptions this implementation requires
+#     (such as Z1=1), called "assumeFormula"
+#   - A constraint describing the assumptions this specific branch requires,
+#     but which is by construction guaranteed to cover the entire space by
+#     merging the results from all branches, called "assumeBranch"
+#   - The result of the computation
+# * All combinations of laws with implementation branches are tried, and:
+#   - If the combination of assumeLaw, assumeFormula, and assumeBranch results
+#     in a conflict, it means this law does not apply to this branch, and it is
+#     skipped.
+#   - For others, we try to prove the require constraints hold, assuming the
+#     information in assumeLaw + assumeFormula + assumeBranch, and if this does
+#     not succeed, we fail.
+#     + To prove an expression is zero, we check whether it belongs to the
+#       ideal with the assumed zero expressions as basis. This test is exact.
+#     + To prove an expression is nonzero, we check whether each of its
+#       factors is contained in the set of nonzero assumptions' factors.
+#       This test is not exact, so various combinations of original and
+#       reduced expressions' factors are tried.
+#   - If we succeed, we print out the assumptions from assumeFormula that
+#     weren't implied by assumeLaw already. Those from assumeBranch are skipped,
+#     as we assume that all constraints in it are complementary with each other.
+#
+# Based on the sage verification scripts used in the Explicit-Formulas Database
+# by Tanja Lange and others, see http://hyperelliptic.org/EFD
+
+class fastfrac:
+  """Fractions over rings."""
+
+  def __init__(self,R,top,bot=1):
+    """Construct a fractional, given a ring, a numerator, and denominator."""
+    self.R = R
+    if parent(top) == ZZ or parent(top) == R:
+      self.top = R(top)
+      self.bot = R(bot)
+    elif top.__class__ == fastfrac:
+      self.top = top.top
+      self.bot = top.bot * bot
+    else:
+      self.top = R(numerator(top))
+      self.bot = R(denominator(top)) * bot
+
+  def iszero(self,I):
+    """Return whether this fraction is zero given an ideal."""
+    return self.top in I and self.bot not in I
+
+  def reduce(self,assumeZero):
+    zero = self.R.ideal(map(numerator, assumeZero))
+    return fastfrac(self.R, zero.reduce(self.top)) / fastfrac(self.R, zero.reduce(self.bot))
+
+  def __add__(self,other):
+    """Add two fractions."""
+    if parent(other) == ZZ:
+      return fastfrac(self.R,self.top + self.bot * other,self.bot)
+    if other.__class__ == fastfrac:
+      return fastfrac(self.R,self.top * other.bot + self.bot * other.top,self.bot * other.bot)
+    return NotImplemented
+
+  def __sub__(self,other):
+    """Subtract two fractions."""
+    if parent(other) == ZZ:
+      return fastfrac(self.R,self.top - self.bot * other,self.bot)
+    if other.__class__ == fastfrac:
+      return fastfrac(self.R,self.top * other.bot - self.bot * other.top,self.bot * other.bot)
+    return NotImplemented
+
+  def __neg__(self):
+    """Return the negation of a fraction."""
+    return fastfrac(self.R,-self.top,self.bot)
+
+  def __mul__(self,other):
+    """Multiply two fractions."""
+    if parent(other) == ZZ:
+      return fastfrac(self.R,self.top * other,self.bot)
+    if other.__class__ == fastfrac:
+      return fastfrac(self.R,self.top * other.top,self.bot * other.bot)
+    return NotImplemented
+
+  def __rmul__(self,other):
+    """Multiply something else with a fraction."""
+    return self.__mul__(other)
+
+  def __div__(self,other):
+    """Divide two fractions."""
+    if parent(other) == ZZ:
+      return fastfrac(self.R,self.top,self.bot * other)
+    if other.__class__ == fastfrac:
+      return fastfrac(self.R,self.top * other.bot,self.bot * other.top)
+    return NotImplemented
+
+  def __pow__(self,other):
+    """Compute a power of a fraction."""
+    if parent(other) == ZZ:
+      if other < 0:
+        # Negative powers require flipping top and bottom
+        return fastfrac(self.R,self.bot ^ (-other),self.top ^ (-other))
+      else:
+        return fastfrac(self.R,self.top ^ other,self.bot ^ other)
+    return NotImplemented
+
+  def __str__(self):
+    return "fastfrac((" + str(self.top) + ") / (" + str(self.bot) + "))"
+  def __repr__(self):
+    return "%s" % self
+
+  def numerator(self):
+    return self.top
+
+class constraints:
+  """A set of constraints, consisting of zero and nonzero expressions.
+
+  Constraints can either be used to express knowledge or a requirement.
+
+  Both the fields zero and nonzero are maps from expressions to description
+  strings. The expressions that are the keys in zero are required to be zero,
+  and the expressions that are the keys in nonzero are required to be nonzero.
+
+  Note that (a != 0) and (b != 0) is the same as (a*b != 0), so all keys in
+  nonzero could be multiplied into a single key. This is often much less
+  efficient to work with though, so we keep them separate inside the
+  constraints. This allows higher-level code to do fast checks on the individual
+  nonzero elements, or combine them if needed for stronger checks.
+
+  We can't multiply the different zero elements, as it would suffice for one of
+  the factors to be zero, instead of all of them. Instead, the zero elements are
+  typically combined into an ideal first.
+  """
+
+  def __init__(self, **kwargs):
+    if 'zero' in kwargs:
+      self.zero = dict(kwargs['zero'])
+    else:
+      self.zero = dict()
+    if 'nonzero' in kwargs:
+      self.nonzero = dict(kwargs['nonzero'])
+    else:
+      self.nonzero = dict()
+
+  def negate(self):
+    return constraints(zero=self.nonzero, nonzero=self.zero)
+
+  def __add__(self, other):
+    zero = self.zero.copy()
+    zero.update(other.zero)
+    nonzero = self.nonzero.copy()
+    nonzero.update(other.nonzero)
+    return constraints(zero=zero, nonzero=nonzero)
+
+  def __str__(self):
+    return "constraints(zero=%s,nonzero=%s)" % (self.zero, self.nonzero)
+
+  def __repr__(self):
+    return "%s" % self
+
+
+def conflicts(R, con):
+  """Check whether any of the passed non-zero assumptions is implied by the zero assumptions"""
+  zero = R.ideal(map(numerator, con.zero))
+  if 1 in zero:
+    return True
+  # First a cheap check whether any of the individual nonzero terms conflict on
+  # their own.
+  for nonzero in con.nonzero:
+    if nonzero.iszero(zero):
+      return True
+  # It can be the case that entries in the nonzero set do not individually
+  # conflict with the zero set, but their combination does. For example, knowing
+  # that either x or y is zero is equivalent to having x*y in the zero set.
+  # Having x or y individually in the nonzero set is not a conflict, but both
+  # simultaneously is, so that is the right thing to check for.
+  if reduce(lambda a,b: a * b, con.nonzero, fastfrac(R, 1)).iszero(zero):
+    return True
+  return False
+
+
+def get_nonzero_set(R, assume):
+  """Calculate a simple set of nonzero expressions"""
+  zero = R.ideal(map(numerator, assume.zero))
+  nonzero = set()
+  for nz in map(numerator, assume.nonzero):
+    for (f,n) in nz.factor():
+      nonzero.add(f)
+    rnz = zero.reduce(nz)
+    for (f,n) in rnz.factor():
+      nonzero.add(f)
+  return nonzero
+
+
+def prove_nonzero(R, exprs, assume):
+  """Check whether an expression is provably nonzero, given assumptions"""
+  zero = R.ideal(map(numerator, assume.zero))
+  nonzero = get_nonzero_set(R, assume)
+  expl = set()
+  ok = True
+  for expr in exprs:
+    if numerator(expr) in zero:
+      return (False, [exprs[expr]])
+  allexprs = reduce(lambda a,b: numerator(a)*numerator(b), exprs, 1)
+  for (f, n) in allexprs.factor():
+    if f not in nonzero:
+      ok = False
+  if ok:
+    return (True, None)
+  ok = True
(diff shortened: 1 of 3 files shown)


sipa:
Updated with more comments.

sipa:
Added exhaustive verification in Z43.

sipa:
Significantly rewrote the code, it now prints assumptions more explicitly, and can deal with cases involving infinity. On the downside, the exhaustive check is too slow now, so it's disabled.

sipa:
Update: make prove_zero call prove_nonzero on the denominators, to prevent 0/0. This shouldn't ever occur in reality, as the formulas should never result in 0/0, and the implementations don't return fractions... but better be safe.

sipa:
Refactored the code into multiple files, in their own directory, and updated the comments.

sage/group_prover.sage
```diff
@@ -0,0 +1,303 @@
...
+  def __str__(self):
+    return "constraints(zero=%s,nonzero=%s)" % (self.zero, self.nonzero)
+
+  def __repr__(self):
+    return "%s" % self
+
+
+def conflicts(R, con):
+  """Check whether any of the passed non-zero assumptions is implied by the zero assumptions"""
+  zero = R.ideal(map((lambda x: mynumerator(R, x)), con.zero))
```
apoelstra: These double-parens are unnecessary and a bit confusing because you don't always use them (eg the `reduce` call a few lines down does not), and also because the eye tends to match `((` with `))` which is wrong. Recommend removing them everywhere (several calls to `map` are like this; I don't think anything else is).

sage/group_prover.sage
```diff
@@ -0,0 +1,303 @@
...
+  if len(con.nonzero) > 0 and 1 in zero:
+    return True
```
apoelstra: I think `len(con.nonzero)` shouldn't be checked here. If one is implied to be zero that's a contradiction by itself.

(I also checked that the result of tests were unchanged by removing the check -- had they not been I'd have been very surprised!)

sage/group_prover.sage
```diff
@@ -0,0 +1,303 @@
...
+  for nonzero in con.nonzero:
+    if nonzero.iszero(zero):
+      return True
+  if reduce(lambda a,b: a * b, con.nonzero, fastfrac(R, 1)).iszero(zero):
+    return True
```
apoelstra: Should this check be here? If so, the previous check implies it and can be removed.

apoelstra: This check actually weirds me out because it implies that `conflicts(zero, nonzero1)` and `conflicts(zero, nonzero2)` might be OK while `conflicts(zero, nonzero1 union nonzero2)` fails, which is counterintuitive.

sage/group_prover.sage
```diff
@@ -0,0 +1,303 @@
...
+    else:
+      self.top = R(numerator(top))
+      self.bot = R(denominator(top)) * bot
+
+  def iszero(self,I):
+    """Return whether this fraction is zero given an ideal."""
+    return self.top in I and self.bot not in I
+
+  def reduce(self,assumeZero):
+    zero = self.R.ideal(map((lambda x: mynumerator(R, x)), assumeZero))
```
apoelstra: By `mynumerator(R, x)` do you mean `mynumerator(self.R, x)`? (This is more of a "how does Python interpret this?" question than a code nit.)

apoelstra: Ah, I think this code is invalid but is never called.

sage/group_prover.sage
```diff
@@ -0,0 +1,303 @@
...
+  def __repr__(self):
+    return "%s" % self
+
+  def numerator(self):
+    return self.top
+
+def mynumerator(R,x):
+  """Compute the numerator of a fraction."""
+  if parent(x) == R:
+    return x
```
apoelstra: This branch is also never called, but I think if `R` is a ring then `x` will equal `numerator(x)` in that case, so I think the branch is unnecessary.

apoelstra: Oh, and without this branch you can `s/mynumerator/numerator/` the whole code :)

sage/group_prover.sage
```diff
@@ -0,0 +1,303 @@
...
+    return True
+  return False
+
+
+def get_nonzero_set(R, assume):
+  """Calculate a simple set of nonzero expressions"""
+  zero = R.ideal(map((lambda x: mynumerator(R, x)), assume.zero))
+  nonzero = set()
+  for nz in assume.nonzero:
+    if mynumerator(R, nz) != 0:
```
apoelstra: I'd prefer this check not be here. If a literal zero is in `assume.nonzero` your constraints are inherently inconsistent and should get flagged. If you remove the check then `factor` will throw an exception on the next line, which is I think is fine.

I also think the check is confusingly written, since `R` appears but as I pointed out in the definition of `mynumerator` it is not actually used.

sage/group_prover.sage
```diff
@@ -0,0 +1,303 @@
...
+  if reduce(lambda a,b: a * b, con.nonzero, fastfrac(R, 1)).iszero(zero):
+    return True
+  return False
+
+
+def get_nonzero_set(R, assume):
+  """Calculate a simple set of nonzero expressions"""
+  zero = R.ideal(map((lambda x: mynumerator(R, x)), assume.zero))
+  nonzero = set()
+  for nz in assume.nonzero:
```
apoelstra: We only ever use `nz` as `mynumerator(nz)`. Prefer iterating over `assume.nonzero.map(numerator)` (or whatever the correct Python for that is)

sage/group_prover.sage
```diff
@@ -0,0 +1,303 @@
...
+    if mynumerator(R, nz) != 0:
+      for (f,n) in mynumerator(R, nz).factor():
+        nonzero.add(f)
+    rnz = zero.reduce(mynumerator(R, nz))
+    if rnz != 0:
```
apoelstra: If you move the definition of `describe` below `if conflicts(R, assume)` in `check_symbolic`, this function will only ever be called after `conflicts()` has been run, so this check should never trigger.

I'll add a note below to suggest that move. For here, I think it suffices to remove this check and add a comment at the top of the function saying we assume we have consistent assumptions.

sage/group_prover.sage
```diff
@@ -0,0 +1,303 @@
...
+      for (f,n) in rnz.factor():
+        nonzero.add(f)
+  return nonzero
+
+
+def prove_nonzero(R, exprs, assume):
+  """Check whether an expression is provably nonzero, given assumptions"""
+  zero = R.ideal(map((lambda x: mynumerator(R, x)), assume.zero))
```
apoelstra: change to `map(numerator, assume.zero)` as per above comments about `mynumerator`

apoelstra: Also all the loops over `expr` should be over `expr.map(numerator)`

sage/group_prover.sage
```diff
@@ -0,0 +1,303 @@
...
+  nonzero = get_nonzero_set(R, assume)
+  expl = set()
+  ok = True
+  for expr in exprs:
+    if mynumerator(R, expr) in zero:
+      return (False, [exprs[expr]])
+  allexprs = reduce(lambda a,b: a*b, exprs, fastfrac(R, 1))
```
apoelstra: If you did `lambda a, b: numerator(a) * numerator(b)`, you could start with 1, and the function could take a list of numbers (rather than just fastfracs) as `expr`

sage/group_prover.sage
```diff
@@ -0,0 +1,303 @@
...
+        expl.add(exprs[expr])
+  if expl:
+    return (False, list(expl))
+  else:
+    return (True, None)
+
+
+def prove_zero(R, exprs, assume):
+  """Check whether all of the passed expressions are provably zero, given assumptions"""
+  r, e = prove_nonzero(R, dict(map(lambda x: (fastfrac(R, x.bot, 1), exprs[x]), exprs)), assume)
```
apoelstra: If `prove_nonzero` could take a dict of numbers as `expr` (which it almost can, see above) then `fastfrac(R, x.bot, 1)` can be replaced with just `x.bot`.

sage/group_prover.sage
```diff
@@ -0,0 +1,303 @@
...
+  if not r:
+    return (False, map(lambda x: "Possibly zero denominator: %s" % x, e))
+  zero = R.ideal(map((lambda x: mynumerator(R, x)), assume.zero))
```
apoelstra: `map(numerator, assume.zero)`

sage/group_prover.sage
```diff
@@ -0,0 +1,303 @@
...
+  nonzero = reduce((lambda x, y: x * y), assume.nonzero, fastfrac(R, 1))
```
apoelstra: Can this `reduce` (and all other instances of it) be replaced with the `prod` function?

sage/group_prover.sage
```diff
@@ -0,0 +1,303 @@
...
+  expl = []
+  for expr in exprs:
+    if not (expr * nonzero).iszero(zero):
```
apoelstra: I'm a little confused by this use of `nonzero` as a witness. I guess the idea is that your ring of fractions is localized at all nonzero expressions (or something algebraic like that). Can you add a comment clarifying this?

sage/group_prover.sage
sipa: Fixed in next push.

sage/group_prover.sage
sipa: Fixed in next push.

sage/group_prover.sage
sipa: Fixed in next push.

sage/group_prover.sage
sipa: Fixed in next push.

sage/group_prover.sage
sipa: It's true, however.

Say that you know that either x=0 or y=0. That's represented as putting x*y in the zero set.

Knowing that x is nonzero or y is nonzero individually does not conflict with that. Both at the same time do conflict, however.

sipa: Added a comment in the next push.

sage/group_prover.sage
sipa: Done in next push.

sage/group_prover.sage
sipa: Done in next push.

sage/group_prover.sage
sipa: Done in next push.

sage/group_prover.sage
sipa: Not possible, as that way you can't find the name in exprs anymore.

sage/group_prover.sage
sipa: Done in next push.

sage/group_prover.sage
sipa: It seems it can't, will look later.

sage/group_prover.sage
sipa: Fixed in next push.

sage/group_prover.sage
sipa: Fixed in next push.

sage/group_prover.sage
sipa: There was a reason for this, but it seems it's no longer needed. Gone in next push.

sipa:
Pushed new version to address @apoelstra's nits.

sage/group_prover.sage
```diff
@@ -0,0 +1,301 @@
...
+    else:
+      self.top = R(numerator(top))
+      self.bot = R(denominator(top)) * bot
+
+  def iszero(self,I):
+    """Return whether this fraction is zero given an ideal."""
+    return self.top in I and self.bot not in I
+
+  def reduce(self,assumeZero):
+    zero = self.R.ideal(map(lambda x: numerator(x), assumeZero))
```
apoelstra: `lambda x: numerator(x)` can just be `numerator` :)

sage/group_prover.sage
```diff
@@ -0,0 +1,301 @@
...
+  def __str__(self):
+    return "constraints(zero=%s,nonzero=%s)" % (self.zero, self.nonzero)
+
+  def __repr__(self):
+    return "%s" % self
+
+
+def conflicts(R, con):
+  """Check whether any of the passed non-zero assumptions is implied by the zero assumptions"""
+  zero = R.ideal(map(lambda x: numerator(x), con.zero))
```
apoelstra: ditto

sage/group_prover.sage
```diff
@@ -0,0 +1,301 @@
...
+  if 1 in zero:
+    return True
+  for nonzero in con.nonzero:
+    if nonzero.iszero(zero):
+      return True
+  # It can be the case that entries in the nonzero set do not individually
+  # conflict with the zero set, but their combination does. For example, if we
+  # know that either x or y is zero, x*y would be in the zero set. Having x or y
+  # individually in the nonzero set is not a conflict, but both simultaneously
+  # is.
```
apoelstra: Suppose that our ring is Z, and our zero set is the ideal (6). Then neither 3 nor 2 is zero, but the product is. So I don't get why we can't simultaneously have 3 and 2 in the nonzero set? It seems like you're testing for something stronger than "for all nonzero things x, x is not zero".

sage/group_prover.sage
```diff
@@ -0,0 +1,301 @@
...
+  else:
+    return (True, None)
+
+
+def prove_zero(R, exprs, assume):
+  """Check whether all of the passed expressions are provably zero, given assumptions"""
+  r, e = prove_nonzero(R, dict(map(lambda x: (fastfrac(R, x.bot, 1), exprs[x]), exprs)), assume)
+  if not r:
+    return (False, map(lambda x: "Possibly zero denominator: %s" % x, e))
+  zero = R.ideal(map(lambda x: numerator(x), assume.zero))
```
apoelstra: `lambda x: numerator(x)` -> `numerator`

sage/group_prover.sage
apoelstra: As @sipa explained to me on IRC, if the zero set is the ideal (6), it will be so because the tester wanted to encode "2 OR 3 is zero".

apoelstra: And since we have a polynomial ring over a finite field, it will have no zero divisors, so this will *always* be the case. That is, `f(x)*g(x)` being zero is equivalent to either `f(x)` or `g(x)` being zero.

Also, by design "being zero" and "being in the zero ideal" mean the same thing in this function. The zero ideal is not some arbitrary ideal; my example with (6) being a zero ideal would mean that somebody literally encoded "6 = 0" as a condition on a formula which of course is nonsense.

sage/secp256k1.sage
```diff
@@ -0,0 +1,306 @@
...
+  t2 = -t4
+  ry = ry + t2
+  return jacobianpoint(rx, ry, rz)
+
+def formula_secp256k1_gej_add_var(branch, a, b):
+  """libsecp256k1's secp256k1_gej_add_var"""
+  if branch == 0:
+    return (constraints(), constraints(nonzero={a.Infinity : 'a_finite'}), b)
+  if branch == 1:
+    return (constraints(), constraints(zero={a.Infinity : 'a_infinite'}, nonzero={b.Infinity : 'b_infinite'}), a)
```
apoelstra: The description string for `a.Infinity` should be `finite`; two lines above `finite` should be `infinite`

sage/secp256k1.sage
```diff
@@ -0,0 +1,306 @@
...
+  s2 = s2 * a.Z
+  h = -u1
+  h = h + u2
+  i = -s1
+  i = i + s2
+  if branch == 2:
+    r = formula_secp256k1_gej_double_var(a)
+    return (constraints(), constraints(zero={h : 'h=0', i : 'i=0', a.Infinity : 'a_finite', b.Infinity : 'b_finite'}), r)
+  if branch == 3:
+    return (constraints(), constraints(zero={h : 'h=0', a.Infinity : 'a_infinite', b.Infinity : 'b_infinite'}, nonzero={i : 'i!=0'}), point_at_infinity())
```
apoelstra: both `_infinite` here should be `_finite`

sage/secp256k1.sage
```diff
@@ -0,0 +1,306 @@
...
+  h3 = -h3
+  ry = ry + h3
+  return (constraints(), constraints(zero={a.Infinity : 'a_finite', b.Infinity : 'b_finite'}, nonzero={h : 'h!=0'}), jacobianpoint(rx, ry, rz))
+
+def formula_secp256k1_gej_add_ge_var(branch, a, b):
+  """libsecp256k1's secp256k1_gej_add_ge_var, which assume bz==1"""
+  if branch == 0:
+    return (constraints(zero={b.Z - 1 : 'b.z=1'}), constraints(nonzero={a.Infinity : 'a_infinite'}), b)
+  if branch == 1:
+    return (constraints(zero={b.Z - 1 : 'b.z=1'}), constraints(zero={a.Infinity : 'a_infinite'}, nonzero={b.Infinity : 'b_infinite'}), a)
```
apoelstra: a_infinite -> a_finite

apoelstra:
Except for the finite/infinite typos in description strings (I think I noted them all), ACK.

Thanks for clarifying the comments!

sipa:
Fixed the incorrect finite/infinite descriptions.

apoelstra:
ACK

PR #303, opened 2015-09-03
remove unused assignment

theuni:
This is insanely pedantic, but it was found along with #300 using clang-tidy. Since this clears up one of the remaining few things that clang-tidy complains about, it makes sense to address it.

t0 is unused after this point, so don't bother assigning it.


dcousens:
utACK

gmaxwell:
Seems reasonable, the dead code was introduced when this function was created by simplifying out the general normalization code.

PR #304, opened 2015-09-03
Fixes a bug where bench_sign would fail due to passing in too small a…

gmaxwell:
Fixes a bug where bench_sign would fail due to passing in too small a buffer.
This was introduced by the recent API change when the signature types were split.


dcousens:
utACK

Issue #308, opened 2015-09-17
Small subgroup alternative curve verification of group law


gmaxwell:
04:22 < gmaxwell> Which I think also suggests another test we can add. Our group law should also hold for all B even when B results in a curve that has low order points.
04:23 < gmaxwell> We can exhaustively test low order subgroups.
04:23 < gmaxwell> e.g. this is something tests.c could do.
04:28 < gmaxwell> this has the benefit of allowing an 'exhaustive' test of the group law without changing the field.

E.g.  on y^2 = x^3 + 4  there is a subgroup with a generator of order 199, and we can 'exhaustively' verify the group law on this subgroup by generating all the points and trying all pairs of adds and multiplies. (Though this doesn't cover all possible projective points).

sipa:
Fixed by #310.

PR #310, opened 2015-09-18
Add exhaustive test for group functions on a low-order subgroup

apoelstra:
We observe that when changing the b-value in the elliptic curve formula
`y^2 = x^3 + ax + b`, the group law is unchanged. Therefore our functions
for secp256k1 will be correct if and only if they are correct when applied
to the curve defined by `y^2 = x^3 + 4` defined over the same field. This
curve has a point P of order 199.

This commit adds a test which computes the subgroup generated by P and
exhaustively checks that addition of every pair of points gives the correct
result.

Unfortunately we cannot test scalar multiplication, const-time or otherwise,
by the same mechanism. The reason is that our ecmult functions both compute
a wNAF representation of the scalar, and this representation is tied to the
order of the group.

Testing with the incomplete version of gej_add_ge (found in 5de4c5dff^)
shows that this detects the incompleteness when adding P - 106P, which
is exactly what we expected since 106 is a cube root of 1 mod 199.

Exhaustive tests added for the following internal functions:
- `secp256k1_ge_is_infinity`, `secp256k1_gej_is_infinity`
- `secp256k1_gej_add_var`, `secp256k1_gej_add_ge`, `secp256k1_gej_add_ge_var`, `secp256k1_gej_add_zinv_var`
- `secp256k1_gej_double_nonzero`, `secp256k1_gej_double_var`
- `secp256k1_ge_neg`, `secp256k1_gej_neg`
- `secp256k1_ecmult`, `secp256k1_ecmult_const`

And also (unfortunately with a bit of surgery; the original functions don't work with small group orders)
- `secp256k1_ecdsa_sign`
- `secp256k1_ecdsa_verify`

### DIFF
--- a/src/ecdsa_impl.h
+++ b/src/ecdsa_impl.h
@@ -203,7 +203,9 @@ static int secp256k1_ecdsa_sig_serialize(unsigned char *sig, size_t *size, const
 static int secp256k1_ecdsa_sig_verify(const secp256k1_ecmult_context *ctx, const secp256k1_scalar *sigr, const secp256k1_scalar *sigs, const secp256k1_ge *pubkey, const secp256k1_scalar *message) {
     unsigned char c[32];
     secp256k1_scalar sn, u1, u2;
+#if !defined(EXHAUSTIVE_TEST_ORDER)
     secp256k1_fe xr;
+#endif
     secp256k1_gej pubkeyj;
     secp256k1_gej pr;
 
@@ -219,6 +221,21 @@ static int secp256k1_ecdsa_sig_verify(const secp256k1_ecmult_context *ctx, const
     if (secp256k1_gej_is_infinity(&pr)) {
         return 0;
     }
+
+#if defined(EXHAUSTIVE_TEST_ORDER)
+{
+    secp256k1_scalar computed_r;
+    int overflow = 0;
+    secp256k1_ge pr_ge;
+    secp256k1_ge_set_gej(&pr_ge, &pr);
+    secp256k1_fe_normalize(&pr_ge.x);
+
+    secp256k1_fe_get_b32(c, &pr_ge.x);
+    secp256k1_scalar_set_b32(&computed_r, c, &overflow);
+    /* we fully expect overflow */
+    return secp256k1_scalar_eq(sigr, &computed_r);
+}
+#else
     secp256k1_scalar_get_b32(c, sigr);
     secp256k1_fe_set_b32(&xr, c);
 
@@ -252,6 +269,7 @@ static int secp256k1_ecdsa_sig_verify(const secp256k1_ecmult_context *ctx, const
         return 1;
     }
     return 0;
+#endif
 }
 
 static int secp256k1_ecdsa_sig_sign(const secp256k1_ecmult_gen_context *ctx, secp256k1_scalar *sigr, secp256k1_scalar *sigs, const secp256k1_scalar *seckey, const secp256k1_scalar *message, const secp256k1_scalar *nonce, int *recid) {
--- a/src/ecmult_const_impl.h
+++ b/src/ecmult_const_impl.h
@@ -78,7 +78,7 @@ static int secp256k1_wnaf_const(int *wnaf, secp256k1_scalar s, int w) {
     /* Negative numbers will be negated to keep their bit representation below the maximum width */
     flip = secp256k1_scalar_is_high(&s);
     /* We add 1 to even numbers, 2 to odd ones, noting that negation flips parity */
-    bit = flip ^ (s.d[0] & 1);
+    bit = flip ^ !secp256k1_scalar_is_even(&s);
     /* We check for negative one, since adding 2 to it will cause an overflow */
     secp256k1_scalar_negate(&neg_s, &s);
     not_neg_one = !secp256k1_scalar_is_one(&neg_s);
--- a/src/scalar_impl.h
+++ b/src/scalar_impl.h
@@ -14,7 +14,9 @@
 #include "libsecp256k1-config.h"
 #endif
 
-#if defined(USE_SCALAR_4X64)
+#if defined(EXHAUSTIVE_TEST_ORDER)
+#include "scalar_low_impl.h"
+#elif defined(USE_SCALAR_4X64)
 #include "scalar_4x64_impl.h"
 #elif defined(USE_SCALAR_8X32)
 #include "scalar_8x32_impl.h"
@@ -31,17 +33,37 @@ static void secp256k1_scalar_get_num(secp256k1_num *r, const secp256k1_scalar *a
 
 /** secp256k1 curve order, see secp256k1_ecdsa_const_order_as_fe in ecdsa_impl.h */
 static void secp256k1_scalar_order_get_num(secp256k1_num *r) {
+#if defined(EXHAUSTIVE_TEST_ORDER)
+    static const unsigned char order[32] = {
+        0,0,0,0,0,0,0,0,
+        0,0,0,0,0,0,0,0,
+        0,0,0,0,0,0,0,0,
+        0,0,0,0,0,0,0,EXHAUSTIVE_TEST_ORDER
+    };
+#else
     static const unsigned char order[32] = {
         0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,
         0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFF,0xFE,
         0xBA,0xAE,0xDC,0xE6,0xAF,0x48,0xA0,0x3B,
         0xBF,0xD2,0x5E,0x8C,0xD0,0x36,0x41,0x41
     };
+#endif
     secp256k1_num_set_bin(r, order, 32);
 }
 #endif
 
 static void secp256k1_scalar_inverse(secp256k1_scalar *r, const secp256k1_scalar *x) {
+#if defined(EXHAUSTIVE_TEST_ORDER)
+    int i;
+    *r = 0;
+    for (i = 0; i < EXHAUSTIVE_TEST_ORDER; i++)
+        if ((i * *x) % EXHAUSTIVE_TEST_ORDER == 1)
+            *r = i;
+    /* If this VERIFY_CHECK triggers we were given a noninvertible scalar (and thus
+     * have a composite group order; fix it in exhaustive_tests.c). */
+    VERIFY_CHECK(*r != 0);
+}
+#else
     secp256k1_scalar *t;
     int i;
     /* First compute x ^ (2^N - 1) for some values of N. */
@@ -233,9 +255,9 @@ static void secp256k1_scalar_inverse(secp256k1_scalar *r, const secp256k1_scalar
 }
 
 SECP256K1_INLINE static int secp256k1_scalar_is_even(const secp256k1_scalar *a) {
-    /* d[0] is present and is the lowest word for all representations */
     return !(a->d[0] & 1);
 }
+#endif
 
 static void secp256k1_scalar_inverse_var(secp256k1_scalar *r, const secp256k1_scalar *x) {
 #if defined(USE_SCALAR_INV_BUILTIN)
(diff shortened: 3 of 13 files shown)


apoelstra:
Fixes #308.

apoelstra:
Pushed a bounds-check fix that @gmaxwell pointed out (so we no longer test negation of infinity). Also added a test for negating infinity elsewhere in the code so that this edge case is tested somewhere.

The problem with testing -infinity along with the other loworder stuff was that setting an unitialized point to infinity leaves its y-coordinate uninitialized, and when calling fe_negate on this, FE_VERIFY branches on it. The result is a valgrind error, and the result of some discussion on IRC was that it'd be best to explicitly test negating fully-initialized infinities elsewhere in tests.c.

src/tests.c
```diff
@@ -1271,6 +1283,95 @@ void run_ge(void) {
...
+    /* Generate the entire group */
+    secp256k1_ge_set_infinity(&group[0]);
+    secp256k1_gej_set_infinity(&groupj[0]);
+    for (i = 1; i < order; i++) {
+        secp256k1_fe_t z;
+        random_fe(&z);
+
+        secp256k1_gej_add_ge(&groupj[i], &groupj[i - 1], &gen);
+        secp256k1_ge_set_gej(&group[i], &groupj[i]);
+        secp256k1_fe_mul(&groupj[i].z, &groupj[i].z, &z);
```
sipa: Use secp256k1_gej_rescale ?

apoelstra:
Push to move exhaustive tests to their own binary; add testing for non-constant ecmult code.

sipa:
Needs rebase (even if github doesn't realize that).

apoelstra:
Rebased and cleaned up. I think this is ready for review.

There aren't any exhaustive tests for modules yet.

sipa:
Travis's on fire, yo.

apoelstra:
Rebased and completed verification

apoelstra:
Updated to just disable static precomputation for `tests_exhaustive.c` instead of whining to the user about her ./configure options.

src/ecdsa_impl.h
```diff
@@ -203,7 +203,9 @@ static int secp256k1_ecdsa_sig_serialize(unsigned char *sig, size_t *size, const
 static int secp256k1_ecdsa_sig_verify(const secp256k1_ecmult_context *ctx, const secp256k1_scalar *sigr, const secp256k1_scalar *sigs, const secp256k1_ge *pubkey, const secp256k1_scalar *message) {
     unsigned char c[32];
     secp256k1_scalar sn, u1, u2;
+#if !defined(EXHAUSTIVE_TEST_ORDER)
```
sipa: ifndef ?

src/scalar_low.h
```diff
@@ -0,0 +1,17 @@
+/**********************************************************************
+ * Copyright (c) 2014 Pieter Wuille                                   *
```
sipa: I can't remember writing this code!

src/scalar_low.h
```diff
@@ -0,0 +1,17 @@
...
+
+#ifndef _SECP256K1_SCALAR_REPR_
+#define _SECP256K1_SCALAR_REPR_
+
+#include <stdint.h>
+
+/** A scalar modulo the group order of the secp256k1 curve. */
+typedef uint32_t secp256k1_scalar;
+
+#define SECP256K1_SCALAR_CONST(d7, d6, d5, d4, d3, d2, d1, d0) ((uint32_t)(d0))
```
sipa: Is this called anywhere? If it's needed, maybe it's better to have a SECP256K1_SCALAR_CONST_SMALL instead that just takes a single uint32_t, and in the non-low case call the normal one?

src/ecdsa_impl.h
apoelstra: We're inconsistent about this throughout the code. Unless you feel strongly about this I'm not going to change it .. we can do a cleanup pass at some point in the future.

sipa: OKAY

src/ecmult_const_impl.h
```diff
@@ -78,7 +78,7 @@ static int secp256k1_wnaf_const(int *wnaf, secp256k1_scalar s, int w) {
     /* Negative numbers will be negated to keep their bit representation below the maximum width */
     flip = secp256k1_scalar_is_high(&s);
     /* We add 1 to even numbers, 2 to odd ones, noting that negation flips parity */
-    bit = flip ^ (s.d[0] & 1);
+    bit = flip ^ !secp256k1_scalar_is_even(&s);
```
sipa: Good. Layer violation prevented.

src/scalar_low_impl.h
```diff
@@ -0,0 +1,121 @@
...
+
+SECP256K1_INLINE static int secp256k1_scalar_is_even(const secp256k1_scalar *a) {
+    return !(*a & 1);
+}
+
+SECP256K1_INLINE static void secp256k1_scalar_clear(secp256k1_scalar *r) { *r = 0; }
+SECP256K1_INLINE static void secp256k1_scalar_set_int(secp256k1_scalar *r, unsigned int v) { *r = v; }
+
+SECP256K1_INLINE static unsigned int secp256k1_scalar_get_bits(const secp256k1_scalar *a, unsigned int offset, unsigned int count) {
+    if (offset < 32)
```
sipa: Either the if/else is unnecessary, or the `(offset < 32) * ` is.

src/scalar_low_impl.h
```diff
@@ -0,0 +1,121 @@
...
+        return (offset < 32) * ((*a >> offset) & ((((uint32_t)1) << count) - 1));
+    else
+        return 0;
+}
+
+SECP256K1_INLINE static unsigned int secp256k1_scalar_get_bits_var(const secp256k1_scalar *a, unsigned int offset, unsigned int count) {
+    if (offset < 32)
+        return (offset < 32) * ((*a >> offset) & ((((uint32_t)1) << count) - 1));
```
sipa: Same here.

apoelstra:
Updated to force-disable the endomorphism optimization for the exhaustive tests, since the `scalar_low` impl it uses does not support it.

apoelstra:
Updated again to put endomorphism support back and just implement `SECP256K1_SCALAR_CONST`, which was the point of incompatibility. Turns out the Schnorr module also needs this macro.

sipa:
Hmm, how can the endomorphism work without changing the lambda constant? (a cubic root of 1 mod the curve order)

apoelstra:
@sipa it observably does .. I have not checked into why but my belief is that it always splits `n` into `n + 0*lambda` for such small `n`.

src/scalar_low_impl.h
```diff
@@ -0,0 +1,114 @@
...
+SECP256K1_INLINE static void secp256k1_scalar_set_int(secp256k1_scalar *r, unsigned int v) { *r = v; }
+
+SECP256K1_INLINE static unsigned int secp256k1_scalar_get_bits(const secp256k1_scalar *a, unsigned int offset, unsigned int count) {
+    if (offset < 32)
+        return ((*a >> offset) & ((((uint32_t)1) << count) - 1));
+    else
+        return 0;
+}
+
+SECP256K1_INLINE static unsigned int secp256k1_scalar_get_bits_var(const secp256k1_scalar *a, unsigned int offset, unsigned int count) {
```
sipa: Does this need a separate implementation from secp256k1_scalar_get_bits? It could just call the other function, or even be #defined to be identical.

src/scalar_impl.h
```diff
@@ -259,6 +281,18 @@ static void secp256k1_scalar_inverse_var(secp256k1_scalar *r, const secp256k1_sc
...
+#if defined(EXHAUSTIVE_TEST_ORDER)
+/**
+ * Find k1 and k2 given k, such that k1 + k2 * lambda == k mod n; unlike in the
+ * full case we don't bother making k1 and k2 be small, we just want them to be
+ * nontrivial to get full test coverage for the exhaustive tests. We therefore
+ * (arbitrarily) set k2 = k + 5 and k1 = k - k2 * lambda.
+ */
+static void secp256k1_scalar_split_lambda(secp256k1_scalar *r1, secp256k1_scalar *r2, const secp256k1_scalar *a) {
+    *r2 = (*a + 5) % EXHAUSTIVE_TEST_ORDER;
+    *r1 = (10000 * EXHAUSTIVE_TEST_ORDER + *a - *r2 * EXHAUSTIVE_TEST_LAMBDA) % EXHAUSTIVE_TEST_ORDER;
```
sipa: I think this can be written as `r1 = (*a + (EXHAUSTIVE_TEST_ORDER - *r2) * EXHAUSTIVE_TEST_LAMBDA) & EXHAUSTIVE_TEST_ORDER;`, avoiding the weird constant.

src/scalar_low_impl.h
```diff
@@ -0,0 +1,114 @@
...
+    if (offset < 32)
+        return ((*a >> offset) & ((((uint32_t)1) << count) - 1));
+    else
+        return 0;
+}
+
+SECP256K1_INLINE static int secp256k1_scalar_check_overflow(const secp256k1_scalar *a) { return *a > EXHAUSTIVE_TEST_ORDER; }
```
sipa: `>=` ?

src/scalar_low_impl.h
```diff
@@ -0,0 +1,114 @@
...
+    bin[28] = *a >> 24; bin[29] = *a >> 16; bin[30] = *a >> 8; bin[31] = *a;
+}
+
+SECP256K1_INLINE static int secp256k1_scalar_is_zero(const secp256k1_scalar *a) {
+    return *a == 0;
+}
+
+static void secp256k1_scalar_negate(secp256k1_scalar *r, const secp256k1_scalar *a) {
+    *r = EXHAUSTIVE_TEST_ORDER - *a;
+    *r %= EXHAUSTIVE_TEST_ORDER;
```
sipa: Faster: `if (*a) { *r = EXHAUSTIVE_TEST_ORDER - *a; } else { *r = 0; }`.

src/tests_exhaustive.c
```diff
@@ -120,26 +155,90 @@ void test_exhaustive_addition(const secp256k1_ge *group, const secp256k1_gej *gr
...
+    secp256k1_fe x;
+    unsigned char x_bin[32];
+    k %= EXHAUSTIVE_TEST_ORDER;
+    x = group[k].x;
+    secp256k1_fe_normalize(&x);
+    secp256k1_fe_get_b32(x_bin, &x);
+    secp256k1_scalar_set_b32(r, x_bin, NULL);
+}
+
+/* hee hee hee */
```
sipa: LOL

src/tests_exhaustive.c
```diff
@@ -120,26 +155,90 @@ void test_exhaustive_addition(const secp256k1_ge *group, const secp256k1_gej *gr
...
+        secp256k1_scalar check_x;
+        r_from_k(&check_x, group, i);
+        if (*x_coord == check_x) {
+           return i;
+        }
+    }
+    return -1;
+}
+
+void test_exhaustive_sign(const secp256k1_context *ctx, const secp256k1_ge *group, int order) {
```
sipa: Is there a use in passing order as a parameter? Almost all of this code already depends on knowing it is equal to EXHAUSTIVE_TEST_ORDER anyway.

src/scalar_impl.h
```diff
@@ -259,6 +281,18 @@ static void secp256k1_scalar_inverse_var(secp256k1_scalar *r, const secp256k1_sc
...
+#if defined(EXHAUSTIVE_TEST_ORDER)
+/**
+ * Find k1 and k2 given k, such that k1 + k2 * lambda == k mod n; unlike in the
+ * full case we don't bother making k1 and k2 be small, we just want them to be
+ * nontrivial to get full test coverage for the exhaustive tests. We therefore
+ * (arbitrarily) set k2 = k + 5 and k1 = k - k2 * lambda.
+ */
+static void secp256k1_scalar_split_lambda(secp256k1_scalar *r1, secp256k1_scalar *r2, const secp256k1_scalar *a) {
+    *r2 = (*a + 5) % EXHAUSTIVE_TEST_ORDER;
+    *r1 = (*a + (EXHAUSTIVE_TEST_ORDER - *r2) * EXHAUSTIVE_TEST_LAMBDA) % EXHAUSTIVE_TEST_ORDER;;;;
```
sipa: ;;;;?

src/scalar_low_impl.h
sipa: Fixed.

PR #313, opened 2015-09-21
Eliminate a few unbraced statements that crept into the code.

gmaxwell:
Also avoids some easily avoided multiple-returns.


sipa:
ACK (after reviewing with url?w=1 on github)

PR #315, opened 2015-09-23
Simple cmake build script

chfast:
Greatings,

I'm working on building your library with cmake. In the end I plan to build it with Visual Studio.

If you were interested in that I would like to spend more time on improving build configs in general. E.g. we can query the compiler about 32/64 bit target architecture instead of checking that during autoconf step.

Share your opinions. Thanks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,22 +14,23 @@ env:
     - FIELD=auto  BIGNUM=auto  SCALAR=auto  ENDOMORPHISM=no  STATICPRECOMPUTATION=yes  ASM=no  BUILD=check  EXTRAFLAGS=  HOST=  ECDH=no  RECOVERY=no  EXPERIMENTAL=no  JNI=no
     - GUAVA_URL=https://search.maven.org/remotecontent?filepath=com/google/guava/guava/18.0/guava-18.0.jar GUAVA_JAR=src/java/guava/guava-18.0.jar
   matrix:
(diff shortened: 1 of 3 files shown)


chfast:
Ping.

sipa:
Is there any means of testing this automatically?

chfast:
automatic builds are probably doable by Travis. However, I'm not familiar with Travis very much.

But this PR is not production ready. I would like to know if you are interested. If yes we can discuss how to proceed.

chfast:
Anyone interested in this?

sipa:
The cmake script looks simple enough, but doesn't provide much flexibility.

I'm fine with merging it as long as there is automated testing that makes sure it doesn't go stale over time, for each of its supported configurations.

chfast:
Ok, I will work on this.

chfast:
Sorry it takes so long...

Do you want me to extend the Travis CI matrix with another dimension of "build system"? As an alternative I can propose using Circle CI for CMake builds.

gmaxwell:
Unfortunately, we lack the resources or interest to adequately maintain another parallel build system on an ongoing basis. To the extent that build-system concerns would be a good use of development time, we believe that they'd better be spent making it easier to build the library without any build system and documenting how that's done.  I believe we're likely to close any further cmake or vcproject pull requests unless something changes-- nothing wrong with your efforts and thank you for the attempt, they just don't reflect this libraries focus at this time.

PR #316, opened 2015-09-23
Use the explicit NULL macro for pointer comparisons.

gmaxwell:
This makes it more clear that a null check is intended. Avoiding the
 use of a pointer as a test condition alse increases the type-safety
 of the comparisons.

(This is also MISRA C 2012 rules 14.4 and 11.9)


gmaxwell:
FWIW, resulting share library is bit identical for me.

sipa:
utACK

PR #317, opened 2015-09-24
Fix miscellaneous style nits that irritate overactive static analysis.

gmaxwell:
Also increase consistency with how overflow && zero is tested, and
 avoid some mixed declarations and code that GCC wasn't detecting.


sipa:
utACK

PR #318, opened 2015-09-24
[Schnorr API BREAK] Improve Schnorr multisigning API + refactoring

sipa:
Instead of asking users to manually generate nonces separately and combine them, there are now specific stage1, stage2 and combine functions which generate 1) signed nonces 2) partial signatures 3) full signatures. This is easier, faster, and more secure (by baking signing of nonces into the API).


sipa:
Added a document explaining Schnorr single and multisigning.

sipa:
Superseded by #322.

PR #319, opened 2015-09-27
Extensive interface and operations tests for secp256k1_ec_pubkey_parse. (and some API paranoia improvements)

gmaxwell:
This also makes use of optional valgrind instrumentation if -DVALGRIND
     is set.
    
    This also moves secp256k1.c above secp256k1.h in tests.c or otherwise
     we get non-null macros on the public functions which may defeat some
     of the VERIFY checks.


gmaxwell:
Perhaps foolishly I made the flags handling in secp256k1_ec_pubkey_serialize absolutely strict. Is this what we want to do? ... would we really ever want to add any more flags there? (hybrid? 0_o).

src/secp256k1.c
```diff
@@ -159,6 +162,11 @@ int secp256k1_ec_pubkey_serialize(const secp256k1_context* ctx, unsigned char *o
     secp256k1_ge Q;
 
     (void)ctx;
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(output != NULL);
+    ARG_CHECK(outputlen != NULL);
+    ARG_CHECK(pubkey != NULL);
+    ARG_CHECK((flags | SECP256K1_EC_COMPRESSED) == SECP256K1_EC_COMPRESSED);
```
sipa: There is more advanced flag checking in #324. Can you leave it for now?

gmaxwell: Sure.  Do you mean leave all this checking out or just the flag part?

sipa: Just the flag part.

gmaxwell:
Okay, flags test gone.

src/tests.c
```diff
@@ -1752,6 +1763,395 @@ void run_endomorphism_tests(void) {
...
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x03
+        }
+    };
+#define SECP256K1_EC_PARSE_TEST_NINVALID (7)
+    const unsigned char invalid[SECP256K1_EC_PARSE_TEST_NINVALID][64] = {
+        {
+            /* x is third root of -8, y is abs(x^3+7); also on the curve for y^2 = x^3 + 9. */
```
sipa: I don't think 'abs' has much meaning in a finite field...

sipa: Do you mean sqrt?

src/tests.c
```diff
@@ -1752,6 +1763,395 @@ void run_endomorphism_tests(void) {
...
+            /* Try all type bytes. */
+            int xpass;
+            int ypass;
+            int ysign;
+            pubkeyc[0] = i;
+            /* What sign does this point have? */
+            ysign = (input[63] & 1) + 2;
+            /* For the current type (i) do we expect parsing to work? Handled all of compressed/uncompressed/hybrid. */
+            xpass = xvalid && (pubkeyclen == 33) && ((i & 254) == 2);
+            ypass = xvalid && yvalid && ((i & 4) == ((pubkeyclen == 65) << 2)) &&
```
sipa: I'm confused about what ypass means, if it can also be true for 33-byte pubkeys?

sipa: Got it. It means "when taking the current prefix byte + subsequence of input, we can parse and reserialize the result as uncompressed, and it will be identical to the input". Took me 10 minutes :)

src/tests.c
```diff
@@ -1752,6 +1763,395 @@ void run_endomorphism_tests(void) {
...
+        shortkey[1] = i >> 8;
+        VG_UNDEF(&pubkey, sizeof(pubkey));
+        CHECK(secp256k1_ec_pubkey_parse(ctx, &pubkey, shortkey, 2) == 0);
+        VG_CHECK(&pubkey, sizeof(pubkey));
+        CHECK(ecount == 0);
+        CHECK(secp256k1_pubkey_load(ctx, &ge, &pubkey) == 0);
+        CHECK(ecount == 1);
+    }
+    /* 33 bytes claimed on otherwise valid input starting with 0x04, fail, zeroize output, no illegal arg error. */
+    memset(&pubkey, 0xfe, sizeof(pubkey));
```
sipa: Seems it starts with 0xfe, not with 0x04?

sipa: Nevermind, pubkeyc vs pubkey...

src/tests.c
```diff
@@ -1752,6 +1763,395 @@ void run_endomorphism_tests(void) {
...
+    CHECK(secp256k1_ec_pubkey_parse(ctx, &pubkey, pubkeyc, 64) == 0);
+    VG_CHECK(&pubkey, sizeof(pubkey));
+    CHECK(ecount == 0);
+    CHECK(secp256k1_pubkey_load(ctx, &ge, &pubkey) == 0);
+    CHECK(ecount == 1);
+    /* 66 bytes claimed, fail, zeroize output, no illegal arg error. */
+    memset(&pubkey, 0xfe, sizeof(pubkey));
+    ecount = 0;
+    VG_UNDEF(&pubkey, sizeof(pubkey));
+    CHECK(secp256k1_ec_pubkey_parse(ctx, &pubkey, pubkeyc, 66) == 0);
```
sipa: That looks like an error. We know ec_pubkey_parse won't go look beyond the 65th byte, but it's a bit ugly that the correctness of the test code relies on that. Can you add a padding byte to pubkeyc?

sipa:
ACK with tiny nits.

gmaxwell:
Nits should all be corrected now.

BIAgents:
$ git push updates

PR #322, opened 2015-09-28
[Schnorr API BREAK] Improve Schnorr multisigning API + fix vulnerability

sipa:
This reworks the Schnorr multisigning API to 4 self-contained functions:
* secp256k1_multischnorr_stage1 computes public nonces
* secp256k1_multischnorr_stage2 computes partial signatures using those public nonces
* secp256k1_multischnorr_combine_sigs combines stage2 sigs into a full signature
* secp256k1_multischnorr_combine_keys combines signers' public keys into a combined public key.

This changes the combined public key to be `A*H(A) + B*H(B) + C*H(C) + ...`, in an attempt to prevent a pubkey cancellation vulnerability.

The documentation is also moved out to a separate schnorr.md document.

### DIFF
--- a/include/secp256k1_schnorr.h
+++ b/include/secp256k1_schnorr.h
@@ -9,7 +9,8 @@ extern "C" {
 
 /** This header file defines an API for a custom EC-Schnorr-SHA256 constructions.
  *  It supports non-malleable 64-byte signatures which support public key
- *  recovery, batch validation, and multiparty signing.
+ *  recovery, batch validation, and multiparty signing. See schnorr.md for more
+ *  details.
  */
 
 /** Create a single party Schnorr signature.
--- a/include/secp256k1_schnorr.h
+++ b/include/secp256k1_schnorr.h
@@ -7,9 +7,12 @@
 extern "C" {
 # endif
 
-/** Create a signature using a custom EC-Schnorr-SHA256 construction. It
- *  produces non-malleable 64-byte signatures which support public key recovery
- *  batch validation, and multiparty signing.
+/** This header file defines an API for a custom EC-Schnorr-SHA256 constructions.
+ *  It supports non-malleable 64-byte signatures which support public key
+ *  recovery, batch validation, and multiparty signing.
+ */
+
+/** Create a single party Schnorr signature.
  *  Returns: 1: signature created
  *           0: the nonce generation function failed, or the private key was
  *              invalid.
@@ -33,13 +36,16 @@ SECP256K1_API int secp256k1_schnorr_sign(
   const void *ndata
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
 
-/** Verify a signature created by secp256k1_schnorr_sign.
+/** Verify a Schnorr signature.
  *  Returns: 1: correct signature
  *           0: incorrect signature
  *  Args:    ctx:       a secp256k1 context object, initialized for verification.
  *  In:      sig64:     the 64-byte signature being verified (cannot be NULL)
  *           msg32:     the 32-byte message hash being verified (cannot be NULL)
  *           pubkey:    the public key to verify with (cannot be NULL)
+ *
+ *  Signatures verifiable by this function can be created using
+ *  secp256k1_schnorr_sign, or secp256k1_multischnorr_combine_sigs.
  */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_schnorr_verify(
   const secp256k1_context* ctx,
@@ -48,8 +54,7 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_schnorr_verify(
   const secp256k1_pubkey *pubkey
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
 
-/** Recover an EC public key from a Schnorr signature created using
- *  secp256k1_schnorr_sign.
+/** Recover an EC public key from a Schnorr signature.
  *  Returns: 1: public key successfully recovered (which guarantees a correct
  *           signature).
  *           0: otherwise.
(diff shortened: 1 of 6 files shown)


sipa:
Rebased, and added several guarantee-zeroed-output cases. Ping @gmaxwell

include/secp256k1_schnorr.h
```diff
@@ -68,104 +77,123 @@ SECP256K1_API int secp256k1_schnorr_recover(
...
+ *  result is a full signature (though different than the one produced by
+ *  secp256k1_schnorr_sign, given the same msg32, sec32, noncefp, ndata).
  *
- *  Note that secp256k1_schnorr_partial_combine and secp256k1_ec_pubkey_combine
- *  function take their arguments in any order, and it is possible to
- *  pre-combine several inputs already with one call, and add more inputs later
- *  by calling the function again (they are commutative and associative).
+ *  All cosigners must use the same msg32, and the same as in stage1. You must
+ *  also use the same noncefp/ndata for your own stage1 and stage2. Other
+ *  participants may use different nonce generation, though.
```
apoelstra: I think you should be more explicit: "Different participants may use different nonce generating functions and data, as long as they are each consistent between stage 1 and stage 2."

Up to you.

src/modules/schnorr/schnorr_impl.h
```diff
@@ -16,89 +16,84 @@
...
+
+static int secp256k1_schnorr_ge_set_b32(secp256k1_ge* p, const unsigned char *b32) {
+    secp256k1_fe x;
+    if (!secp256k1_fe_set_b32(&x, b32)) {
+        return 0;
+    }
+    return secp256k1_ge_set_xo_var(p, &x, 0);
+}
+
+/** Computes {priv = +/- (scalar)b32; pub = priv * G } with pub.y even. */
```
apoelstra: `pub = priv * G + pubothers`

src/modules/schnorr/schnorr_impl.h
```diff
@@ -16,89 +16,84 @@
...
+    secp256k1_gej gej;
+
+    secp256k1_scalar_set_b32(priv, b32, &overflow);
+    if (overflow || secp256k1_scalar_is_zero(priv)) {
+        secp256k1_scalar_clear(priv);
+        return 0;
+    }
+    secp256k1_ecmult_gen(ctx, &gej, priv);
+    if (secp256k1_gej_is_infinity(&gej)) {
+        return 0;
```
apoelstra: Can this branch be hit? If `priv` is zero the above check gets it; if `priv` is n (or higher) then `overflow` should be set and the above check gets it.

sipa: Replaced with VERIFY_CHECK.

src/modules/schnorr/schnorr_impl.h
sipa: It's more complicated even. Fixed.

include/secp256k1_schnorr.h
sipa: Fixed. Used your language.

sipa:
Addressed nits.

apoelstra:
Hi, sorry for the delays, this fell off my radar. Will review today while in flight.

sipa:
@apoelstra Subtle ping

fanatid:
@sipa can you explain `Verification (method 1)`?

```
x - private key, k - private nonce
m - 32 byte message, Q - public key, signature pair - (R.x, s)

signing:
R = G^k
e = SHA256(R.x || m)
s = k - e * x
signature is (R.x, s)

verify:
e = SHA256(R.x || m)
R' = Q^e + G^s = G^(x * e) + G^(k - e * x) = ???
```
How `G^(x * e) + G^(k - e * x)` can be equal to `R = G^k`?
if instead `+` will be `*` all be ok:
```
R' = G^(x * e) * G^(k - e * x) = G^(x * e + k - e * x) = G^k
```

sipa:
@fanatid I think you're confusing additive notation with multiplicative notation.

We call the EC group operation '+', and its repeated application a number of times '*', so when a + b = c, then c*G = (a+b)*G = a*G + b*G.

fanatid:
@sipa thank you! I really forgot that EC multiplication is repeated addition.

src/modules/schnorr/schnorr.md
```diff
@@ -0,0 +1,195 @@
...
+
+Inputs:
+* Partial stage 2 signatures `(R_all(j).x, s(j))` from all cosigners, for all
+  `j`
+
+Steps:
+* Check whether all `R_all(j).x` values in each of the stage 2 signature are
+  identical. If not, fail.
+* Compute the sum `s_all` of all `s(j)` values.
+* The full combined signature is `(R_all.x, s(i))`.
```
fanatid: `(R_all.x, s_all)` ?

src/modules/schnorr/schnorr.md
```diff
@@ -0,0 +1,195 @@
...
+* Compute (or reuse from stage 1) public nonce `R(i) = k(i) * G`.
+* If `R(i).y` is odd, negate `k(i)` and `R(i)`.
+* Convert each `R(j).x` coordinate into an `R(j)` point, with even `y`
+  coordinate.
+* Compute the sum `R_all(i)` of all the `R(j)` points, including your own
+  `R(i)`.
+* If `R_all.y` is odd, negate `k(i)` and `R_all(i)`.
+* Compute scalar `e = SHA256(R_all(i).x || m)`. If `e == 0` or `e >= order`,
+  fail.
+* Compute scalar `s(i) = k(i) - e * x`.
```
fanatid: `s(i) = k(i) - e * x(i)` ?

sipa:
Going to do this differently.

prusnak:
> Going to do this differently.

Is there a source where I can learn how this is being done today?

Also not sure if helpful, but I found a working cosigning implementation for ed25519 in Go:

* https://godoc.org/github.com/bford/golang-x-crypto/ed25519/cosi
* https://github.com/bford/golang-x-crypto/tree/master/ed25519/cosi

sipa:
@prusnak In review we discovered the A*H(A) + B*H(B) + ... scheme was vulnerable to a generalized birthday attack. We have a new scheme, but no strong proof for security yet. I didn't know about CoSi - I'll have a look at it.

ysangkok:
@prusnak more details here: https://github.com/kanzure/diyhpluswiki/blob/master/transcripts/bitcoin-core-dev-tech/2017-09-06-signature-aggregation.mdwn

businessintegrator:
What about collision period of the H function?

PR #324, opened 2015-09-29
[API BREAK] Make flags more explicit, add runtime checks.

rustyrussell:
Fewer zeroes, plus check bit to make sure it value is valid.

Signed-off-by: Rusty Russell <rusty@rustcorp.com.au>

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -148,12 +148,21 @@ typedef int (*secp256k1_nonce_function)(
 # endif
 
 /** Flags to pass to secp256k1_context_create. */
-# define SECP256K1_CONTEXT_VERIFY (1 << 0)
-# define SECP256K1_CONTEXT_SIGN   (1 << 1)
+# define SECP256K1_CONTEXT_VERIFY (SECP256K1_CONTEXT_VERIFY_BIT|SECP256K1_CONTEXT_CHECK_BIT)
+# define SECP256K1_CONTEXT_SIGN   (SECP256K1_CONTEXT_SIGN_BIT|SECP256K1_CONTEXT_CHECK_BIT)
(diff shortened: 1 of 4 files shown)


jgarzik:
ut ACK

include/secp256k1.h
```diff
@@ -148,12 +148,21 @@ typedef int (*secp256k1_nonce_function)(
...
-# define SECP256K1_CONTEXT_VERIFY (1 << 0)
-# define SECP256K1_CONTEXT_SIGN   (1 << 1)
-
+# define SECP256K1_CONTEXT_VERIFY (SECP256K1_CONTEXT_VERIFY_BIT|SECP256K1_CONTEXT_CHECK_BIT)
+# define SECP256K1_CONTEXT_SIGN   (SECP256K1_CONTEXT_SIGN_BIT|SECP256K1_CONTEXT_CHECK_BIT)
+# define SECP256K1_CONTEXT_NONE   (SECP256K1_CONTEXT_CHECK_BIT)
+
+#  define SECP256K1_CONTEXT_VERIFY_BIT     (1 << 2)    
+#  define SECP256K1_CONTEXT_SIGN_BIT       (2 << 2)    
+#  define SECP256K1_CONTEXT_CHECK_BIT      (1)
```
apoelstra: I'd prefer VERIFY and SIGN to be (1 << 0) and (1 << 1); COMPRESSED should also be (1 << 0), and in general the actual flags should be written as (1 << n) for increasing n.

The CHECK_BITs would then need to be something high, maybe (1 << 10), (1 << 11), etc, which would give us room for ten flags in each of five different flag categories, while still fitting into a 16-bit int.

apoelstra:
Can you add [API BREAK] to the title of this PR?

Also concept ACK.

sipa:
Superseded by #339.

PR #326, opened 2015-09-29
Do not force benchmarks to be statically linked.

gmaxwell:
Libtool will do the right thing and use whatever is available
 based on --enable-shared/--enable-static.

This also means that some of the things we build actually
 test the dynamic library.


gmaxwell:
As an aside, this means that to valgrind the things this builds by default you have to do the libtool --mode=execute thing on them. :)

sipa:
ACK

Issue #311, opened 2015-09-18
check that travis builds what we think it builds


gmaxwell:
right now "32bit scalar + recovery" is actually building:

configure: Using assembly optimizations: x86_64
configure: Using field implementation: 64bit
configure: Using bignum implementation: gmp
configure: Using scalar implementation: 32bit
configure: Using endomorphism optimizations: no
configure: Building ECDH module: no
configure: Building Schnorr signatures module: 
configure: Building ECDSA pubkey recovery module: no

sipa:
ouch

PR #328, opened 2015-09-30
travis: fixup module configs

theuni:
Looking over all of the vars, it appears as though RECOVERY was the only one improperly represented. I went ahead and changed schnorr to make it consistent with the others.

Assuming configure works as intended, this addresses #311.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ compiler:
   - gcc
 env:
   global:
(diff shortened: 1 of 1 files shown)


theuni:
Added one more change. A while back there was some odd breakage in the -DDETERMINISTIC build that turned out to be caused by the fact that it replaced the default CFLAGS with its own, meaning that instead of being built with "-O3 -g" by default, everything was built with the compiler's default (-O0).

Since that's somewhat useful to have in the matrix, I've separated the two conditions by adding separate builds for the -DDETERMINISTIC and -O0 cases.

apoelstra:
ACK. Checked six of the Travis builds (yes/no for each of the three modules) to check that the right things are being built.

PR #334, opened 2015-10-12
Overhaul ECDSA signature parsing: strict DER, compact sigs, tests, lower-S

sipa:
There are now 2 encoding formats supported: 64-byte "compact" and DER. The latter is strict: the data has to be exact DER, though the values inside don't need to be valid.

This means that applications that need more than strict DER as input, it needs to have a parser itself. I wish there was a better solution, but there really isn't. Any subset of BER to implement would be arbitrary, incompatible with many other applications anyway, or be very complex and nearly untestable. Despite that, a code snippet to do so (with extensive testing) is included in the contrib/ directory now.

Finally, by default, we now require non-malleable signatures (by requiring the lower-S form), but a separate function is provided to normalize a signature before verification if needed.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -40,6 +40,7 @@ noinst_HEADERS += src/hash_impl.h
 noinst_HEADERS += src/field.h
 noinst_HEADERS += src/field_impl.h
 noinst_HEADERS += src/bench.h
+noinst_HEADERS += contrib/lax_der_parsing.h
 
 pkgconfigdir = $(libdir)/pkgconfig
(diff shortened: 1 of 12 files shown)


sipa:
Added a commit that introduces a snippet for lax DER parsing (tested!).

sipa:
Addressed many comments.

sipa:
Added a certainly_not_der as output from the BER generator, and tests using it.

gmaxwell:
ACK

Issue #251, opened 2015-05-15
secp256k1_ecmult_gen_context_build allocates 90KB on the stack


tdaede:
secp256k1_ge_t prec[1024];
    secp256k1_gej_t precj[1024]; /* Jacobian versions of prec. */

Microcontrollers have expensive SRAM and cheap Flash. These should be turned into precomputed tables, at least as an option.

DavidEGrayson:
There is already a discussion happening in issue #189 about how to make this library work well on microcontrollers.  I suspect it would be best to just have a precomputed context that is neither created nor destroyed, and gets stored in flash, but I haven't looked into it.

laanwj:
Precomputing the tables and storing them in the executable could also help in the libbitcoinconsensus case, where we don't really want a dynamically allocated context.

gmaxwell:
Precomputing verify would be very undesirable-- it's 1.3MB with current settings (which are optimal for performance, when enough memory is available).

Libbitcoinconsensus really should be really using a context in order to _avoid_ the constant dynamic memory allocations all through its internals which it currently has.

laanwj:
The context brings a lot of issues for users, and risks to use more memory in total.
Every client of secp256k1 makes its own context, all lugging along 1.3MB of the same read-only data.
Say a program uses both secp256k1 directly and through libbitcoin_consensus. This means (at least) two contexts. Passing an existing secp256k1 context to libbitcoin_consensus would leak implementation details.
Also it is not clear whether the libbitcoin_consensus contexts may be shared between threads. If not, this is 1.3MB per thread. If they can, this needs to be documented well as it is bound to cause confusion.
Hence I much prefer stateless, context-less functions.

sipa:
Yes, contexts are sharable between threads.Only function that modify them
need exclusive access (randomize being the only one currently).

tdaede:
There are actually two problems here:
One is that signing context initialization currently uses the size of the context, plus over 90K of temporary data. In my case, the temporary data causes me to exceed my RAM budget.
The second is that, in some use cases (including mine), I have memory-mapped Flash that can store the contexts instead. This totally avoids the problem of context initialization.

I am not sure it is worth trying to reduce the temporary RAM usage, because the devices where that is the problem generally also have memory-mapped Flash.

DavidEGrayson:
[...]

PR #337, opened 2015-10-18
variable sized precomputed table for signing

douglasbakkum:
This pull request gives an option to reduce the precomputed table size for the signing context (`ctx`) by setting `#define ECMULT_GEN_PREC_BITS [N_BITS]`.

Motivation: Per #251 and #254, the static table can be reduced to 64kB. However, this is still too big for some of my embedded applications. Setting `#define ECMULT_GEN_PREC_BITS 2` produces a 32kB table at a tradeoff of about 75% of the signing speed. Not defining this value will default to the existing implementation of 4 bits. Statistics:

```
ECMULT_GEN_PREC_BITS = 1
Precomputed table size: 32kB
./bench_sign 
ecdsa_sign: min 195us / avg 200us / max 212us

ECMULT_GEN_PREC_BITS = 2
Precomputed table size: 32kB
./bench_sign 
ecdsa_sign: min 119us / avg 126us / max 134us

ECMULT_GEN_PREC_BITS = 4 (default)
Precomputed table size: 64kB
./bench_sign
ecdsa_sign: min 83.5us / avg 89.6us / max 95.3us

ECMULT_GEN_PREC_BITS = 8
Precomputed table size: 512kB
./bench_sign 
ecdsa_sign: min 96.4us / avg 99.4us / max 104us
```

Only values of 2 and 4 make sense. 8 bits causes a larger table size with no increase in speed. 1 bit runs, actually, but does not reduce table size and is slower than 2 bits.


dbitbox:
See also comments about alternatively using wNAF:
https://github.com/bitcoin/secp256k1/issues/189#issuecomment-149000292

gmaxwell:
Reasonable enough.

Can you help me understand your memory trade-off requirements?  There is a way to halve the table size without largely changing the performance though it requires getting rid of the nums blinding.  I'd been under the impression that with the table completely static most of the reasonable target embedded devices (e.g. arm m3s with 192k ram) would end up with the table in their relatively copious flash and there was no real need to reduce below 64k on them.

dbitbox:
Sure. The current device has 256kB of flash and 64kB of RAM on a Cortex M4. The secp256k1 library (including the 64kB table) takes up about 140kB in total.

jonasschnelli:
Concept ACK.
I think hardware wallets and other use-cases on embedded devices don't really seek for super-performance. But they like to use a reliable and well-testes library that fits on a chip with tiny ram and tiny flash. Removing precomputed tables sounds after a logical tradeoff in that territory.

ddustin:
Is there a reason this was never merged in?

gmaxwell:
@real-or-random  This complements the WINDOW_G work, would you mind reviewing it?

real-or-random:
I'll have a look. :)

src/ecmult_gen_impl.h
```diff
@@ -49,39 +49,39 @@ static void secp256k1_ecmult_gen_context_build(secp256k1_ecmult_gen_context *ctx
 
     /* compute prec. */
     {
-        secp256k1_gej precj[1024]; /* Jacobian versions of prec. */
+        secp256k1_gej precj[ECMULT_GEN_PREC_N * ECMULT_GEN_PREC_G]; /* Jacobian versions of prec. */
         secp256k1_gej gbase;
         secp256k1_gej numsbase;
         gbase = gj; /* 16^j * G */
```
real-or-random: there are some comments that should be changed tere and up to line 67

real-or-random:
@douglasbakkum I know this is very old. Are you still interested in this PR? If yes, it needs rebase. If not, we could adopt it.

douglasbakkum:
Awesome. Yes definitely interested. We have been using this in production BitBox's since the PR was submitted.

I will rebase.

douglasbakkum:
@real-or-random The PR is now rebased. Let me know if you would like further changes. 

(I am not sure how you would prefer the comments to be changed and so did not change them in the rebase.)

src/ecmult_gen_impl.h
douglasbakkum: I updated the comments, assuming you meant to change, for example, `16` to `PREC_G`. Let me know if there is something in addition.

real-or-random:
Thanks, looks good to me!

What we need to do is make this and #596 consistent (for naming of constants, ./configure, etc.) I think the easiest way forward is to get #596 merged and then adapt this PR here. (I know this is also the easiest way for me as the author of #596 but I'm of course open to other suggestions too. :P) 

@gmaxwell What do you think? I'm not convinced that "window size" would be an appropriate term here, so maybe PREC_BITS makes more sense in general?

gmaxwell:
Sounds fine to me (both on the sequence and the naming)

gmaxwell:
#596 is now merged and I don't believe there is anything at risk of imminent merger which will cause difficult conflicts with this PR.

gmaxwell:
@real-or-random ^

real-or-random:
Great, so if ECMULT_GEN_PREC_BITS is fine, then this constant should be `./configure`-able as in #596. See this commit for reference: https://github.com/bitcoin-core/secp256k1/pull/596/commits/2842dc523e88ff8f01c16486a1601ee49f4e60ff

@douglasbakkum  Do you want to give it a try?

douglasbakkum:
@real-or-random Yea, agree. If it is quick for you, would you mind doing so?

vhnatyk:
Hi! any progress with this?

douglasbakkum:
Sorry for the delays. My colleague @benma made a commit to make the value configurable.

benma:
@real-or-random did you have a chance to test the configuration?

real-or-random:
No, sorry, not yet. It's on my long "GitHub saved notifications" list, I'll have a look next week. :)

src/gen_context.c
```diff
@@ -45,23 +46,23 @@ int main(int argc, char **argv) {
     fprintf(fp, "#define _SECP256K1_ECMULT_STATIC_CONTEXT_\n");
     fprintf(fp, "#include \"src/group.h\"\n");
     fprintf(fp, "#define SC SECP256K1_GE_STORAGE_CONST\n");
-    fprintf(fp, "static const secp256k1_ge_storage secp256k1_ecmult_static_context[64][16] = {\n");
+    fprintf(fp, "static const secp256k1_ge_storage secp256k1_ecmult_static_context[ECMULT_GEN_PREC_N][ECMULT_GEN_PREC_G] = {\n");
```
real-or-random: This is somewhat weird if you change the config in between but for some reason `gen_context` is not run again.

So I *think* the generated `ecmult_static_context.h` should have the `gen_context.c` compile-time values of `ECMULT_GEN_PREC_N`  and `ECMULT_GEN_PREC_G` hardcoded and contain a precompiler assertion (`#error`) that verifies they're identical to the `ecmult_static_context.h`-compile-time values. This should make sure that we never compile with a `ecmult_static_context.h` generated with a different configuration. (Alternatively, we could make the file name of `ecmult_static_context.h` make dependent on the configured values, e.g., `ecmult_static_context_64_16.h` or something.)

But maybe I make a mistake here, it's easy to get these things wrong. So correct me if you think I'm wrong.

benma: That's a great catch. Normally gen_context on `make` is run if the configuration changed, but it might not be absolutely failsafe (e.g. by manually overwriting the file with a different version etc.).

Change incoming.

vhnatyk:
Seems we are so close 😄, sorry for spamming - what else needs to be done?

real-or-random:
Ah I haven't seen the changes. I'll have a look next week!

configure.ac
```diff
@@ -165,6 +165,14 @@ AC_ARG_WITH([ecmult-window], [AS_HELP_STRING([--with-ecmult-window=SIZE|auto],
 )],
 [req_ecmult_window=$withval], [req_ecmult_window=auto])
 
+AC_ARG_WITH([ecmult-gen-precision], [AS_HELP_STRING([--with-ecmult-gen-precision=2|4|8|auto],
+[Precision bits to tune the precomputed table size for signing.]
+[The size of the table is 32kB for 2 bits, 64kB for 4 bits, 512kB for 8 bits of precision.]
+[A smaller table size usually results in slower signing.]
```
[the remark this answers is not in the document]
benma: > With `./configure --disable-openssl-tests --with-ecmult-gen-precision=8` (but not `2` and `4`) and `valgrind ./exhaustive_tests` I get millions of warnings like

There is no issue if you call it like this:

```valgrind --max-stackframe=2097912 ./exhaustive_tests```

[Reference](http://valgrind.org/docs/manual/manual-core.html):

> -max-stackframe=<number> [default: 2000000]
>    The maximum size of a stack frame. If the stack pointer moves by more than this amount then Valgrind will assume that the program is switching to a different stack.
>    You may need to use this option if your program has large stack-allocated arrays. 

Seems like the big option just needs more stack space. It's a valgrind heuristic only and the tests run fine otherwise.

File: Makefile.am [hunk 4]
```diff
@@ -172,7 +172,7 @@ src/ecmult_static_context.h: $(gen_context_BIN)
 CLEANFILES = $(gen_context_BIN) src/ecmult_static_context.h $(JAVAROOT)/$(JAVAORG)/*.class .stamp-java
 endif
 
-EXTRA_DIST = autogen.sh src/gen_context.c src/basic-config.h $(JAVA_FILES)
+EXTRA_DIST = autogen.sh src/gen_context.c src/libsecp256k1-config.h src/basic-config.h $(JAVA_FILES)
```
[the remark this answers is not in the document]
benma: > Can you add a sentence like "Try deleting ecmult_static_context.h before the build." or similar?

Done.

> I don't think we want to distribute src/libsecp256k1-config.h. It will be generated by configure.

That was a misguided attempt at having gen_context rebuild the static context upon configuration change. It seems to work now anyway, so I removed it again.

configure.ac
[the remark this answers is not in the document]
real-or-random: Oh indeed, thanks!

README.md
```diff
@@ -59,3 +59,12 @@ libsecp256k1 is built using autotools:
...
     $ sudo make install  # optional
+
+Exhaustive tests
+-----------
+
+    $ ./exhaustive_tests
+
+With valgrind, you might need to increase the max stack size:
+
+    $ valgrind --max-stackframe=2097912 ./exhaustive_tests
```
real-or-random: I need slightly larger values.... Maybe recommend `2500000`?

src/gen_context.c
```diff
@@ -4,9 +4,10 @@
  * file COPYING or http://www.opensource.org/licenses/mit-license.php.*
  **********************************************************************/
 
+#include "libsecp256k1-config.h" // for ECMULT_GEN_PREC_BITS
```
real-or-random: It's not crucial but we should make sure that this warning disappears:
```/configure --disable-openssl-tests --with-ecmult-gen-precision=8 --enable-experimental --enable-module-recovery --enable-module-ecdh --with-ecmult-window=22  CC=clang```

```
In file included from src/gen_context.c:9:
src/basic-config.h:33: warning: "ECMULT_WINDOW_SIZE" redefined
   33 | #define ECMULT_WINDOW_SIZE 15
      |
In file included from src/gen_context.c:7:
src/libsecp256k1-config.h:18: note: this is the location of the previous definition
   18 | #define ECMULT_WINDOW_SIZE 22
```

I guess other compilers warn, too.

In the end it does not matter what `ECMULT_WINDOW_SIZE` is for use in `gen_context.c`. I think sticking to 15 is reasonable (but the fact that we ignore the config setting here should be more prominent in the source).

Makefile.am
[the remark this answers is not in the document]
real-or-random: Oh hm, apparently `make distcheck` fails here:
https://travis-ci.org/bitcoin-core/secp256k1/jobs/558864214#L847

On a second thought, I'm not sure that including `src/libsecp256k1-config.h` is a good idea. There are downstream users of this library (e.g., https://github.com/rust-bitcoin/rust-secp256k1) that don't use the autotools. But I'm not sure what the best thing is. Maybe including it only if `ECMULT_GEN_PREC_BITS` is not defined. Sounds pragmatic to me.

src/gen_context.c
real-or-random: > In the end it does not matter what `ECMULT_WINDOW_SIZE` is for use in `gen_context.c`. I think sticking to 15 is reasonable (but the fact that we ignore the config setting here should be more prominent in the source).

Actually it does not matter too much because `gen_context` does not use `ECMULT_WINDOW_SIZE` at all.

README.md
benma: Done

src/gen_context.c
benma: Fixed via `#undef ECMULT_WINDOW_SIZE` in basic-config.h, like with the other definitions.

Makefile.am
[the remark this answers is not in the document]
benma: fixed

src/gen_context.c
vhnatyk: @real-or-random  sorry to bug, seems ok now? 😊

benma: Ping @real-or-random - seems we are really close now :P

real-or-random:
Yes, we're close. :)

ACK 1d26b27ac90092306bfbc9cdd5123e8a5035202a (mod squashing) I read the changes, and tested them both with and without static precomputation, with different precision values and also with valgrind.

Can you squash some of the commits?

real-or-random:
@sipa Can you have look at this? I think you know the algorithm best.

sipa:
Code review ACK 1d26b27ac90092306bfbc9cdd5123e8a5035202a. Looks great; all comments I wnated to make were addressed in later commits.

Please squash the fixup commits.

elichai:
Maybe better docs to differentiate the different between `ecmult_static_precomputation`,`ecmult-window`,`ecmult-gen-precision`?

benma:
@elichai okay to do this in a different PR? Not strictly related.

real-or-random:
I think this can happen in a different PR.

elichai:
Sure, just getting to a lot of tables that it's getting confusing which table will affect what.

douglasbakkum:
Thanks all.
The commits are now squashed and rebased.

vhnatyk:
So happy this got merged finally!) Hopefully I can start getting rid of black magik in my code now))

PR #339, opened 2015-10-21
[API BREAK] Explicit flag handling (rebase + modification of #324)

sipa:


sipa:
This is an update of #324 with the following changes:
* Keep the lower 8 bits of flags to indicate their type
* Revert the change to pass-through the compression flag to the eckey module (which was a layer violation, the flag was defined at a higher level but used on a lower one).

sipa:
Comments, @rustyrussell @luke-jr ?

luke-jr:
I do not understand the *reason* for pulling flags out of src/eckey_impl.h.

Besides that, fully-reviewed-and-seems-good ACK.

sipa:
Luke: the lower-level eckey code should not depend on the top-level wrapper
whose interface is exposed.

sipa:
Rebased, and renamed unsigned int to int for bool flag.

sipa:
Rebased on top of #340.

PR #340, opened 2015-10-21
Test improvements and some small API fixes that they turned up.

gmaxwell:
Included in here,

One commit changes the 6979 nonce behavior when algo16 is provided so that non-provided data and zero data do not provide the same nonce. I think equivalence between NULL and zero-bytes there is less safe.

Another commit makes the usage of VERIFY_CHECK() side-effect free. I think this makes the code more reviewable (I personally have a tendency to assume CHECK macros aren't hit at runtime) but mostly it simplifies coverage analysis where dead code from VERIFY otherwise counts against coverage. To avoid that I need to make VERIFY_CHECK a no-op during coverage analysis. This PR doesn't change VERIFY_CHECK's behavior itself.

This PR also makes the failure cases for the tweaks functions and create behave more consistently with respect to zeroization of the output on failure.

Other commits move recovery into the recovery module, and the privkey parser into contrib.

There are more details in individual commit messages.


gmaxwell:
This is on top of (current) #334.  I still have static test vectors outstanding, but I have a lot of changes here and wanted to get what I had done in the pipeline rather than suffer rebase hell just waiting for the static vectors to finish optimizing, since they're fairly free-standing.

sipa:
Needs rebase.

gmaxwell:
rebased.

sipa:
ACK

PR #342, opened 2015-10-26
Make lax_der_privatekey_parsing.h not depend on internal code

sipa:


gmaxwell:
ACK. As an aside... Should we consider changing the interface so privkeylen  is in/out so we can check the output size?

apoelstra:
utACK

PR #343, opened 2015-10-26
Change contrib/laxder from headers-only to files compilable as standalone C

apoelstra:
Verified that both programs compile with

    gcc -I. -I../include -lsecp256k1 -c -W -Wextra -Wall -Werror -ansi -pedantic lax_der_privatekey_parsing.c
    gcc -I. -I../include -lsecp256k1 -c -W -Wextra -Wall -Werror -ansi -pedantic lax_der_parsing.c


sipa:
Eh, I don't like this. These things are not compilable modules with a neat exposed API, they're just pieces of example code...

gmaxwell:
I don't see where the modules comment comes from?  They're now just ordinary C code, not weird header stuff.  They're still just code fragments, not libraries, not modules, not added to the build system, etc. 

Oh you're referring the linking into libsecp256k1?  Yea, I don't agree with that, but this patch doesn't do that.

sipa:
Turning it into full .c files with corresponding header files gives a
pretty strong impression that this is code that is to be compiled and
linked into your application as-is.

I really want to discourage that. You shouldn't be compiling code from
another project's source tree. If that was the intention, we should just
make it a compile time flag and expose it through the library.

This is an example code snippet you can copy-paste into your application or
translate to another language as needed, making changes as needed too.

apoelstra:
Updated to fix tests.c, add warning for users not to link directly from our codebase, and remove the `secp256k1_` prefixes from the functions to emphasize that they are not part of libsecp256k1 proper.

sipa:
You need to add the .c files to the makefiles.

sipa:
@apoelstra Ping?

apoelstra:
I added the .c files to `noinst_HEADERS`. Is this right?

sipa:
It seems the almighty Travis is happy with it.

PR #344, opened 2015-10-30
Adds 32 static test vectors for scalar mul, sqr, inv.

gmaxwell:
These were generated by testing more than 10^12 random test vectors
 for coverage on instrumented (comparison operator outcomes) 32-bit
 and 64-bit code, plus additional edge condition requirements (e.g.
 inputs of 0, 1, -1) and then solving a minimum set cover problem.

The required responses were generated with Sage.

This significantly improves the lcov branch coverage report and
 makes the tests much more sensitive to mutation testing of the
 scalar code.


sipa:
Any reason for not using SECP256K1_SCALAR_CONST?

gmaxwell:
No-- didn't think to, but on consideration: this way tests the set operation.  Would you hate me terribly if I made one of the two tables use the macros and continued to use the function for the other?

sipa:
@gmaxwell Well specifically for the all 0xFF value; using SECP256K1_CONST_SCALAR likely keeps working longer even if the scalar implementation/representation changes.

gmaxwell:
@sipa Changed the all 0xFF case.

PR #345, opened 2015-10-30
Better error case handling for pubkey_create & pubkey_serialize, more tests.

gmaxwell:
Makes secp256k1_ec_pubkey_serialize set the length to zero on failure,
 also makes secp256k1_ec_pubkey_create set the pubkey to zeros when
 the key argument is NULL.

Also adds many additional ARGCHECK tests.


sipa:
ACK

PR #348, opened 2015-11-02
Improvements for coordinate decompression

sipa:
Adds:
* Separate secp256k1_ge_set_xquad_var which constructs a point with a given X coordinate, guaranteeing a Y coordinate that is a quadratic residue.
* Implement secp256k1_ge_set_xo_var in function of the above.
* Documentation to explain this.
* Unit tests for all of the above.


gmaxwell:
utACK.   If you wanted to elaborate the comments a bit you could point out why the _sqr() power ladder always results in the result which is itself a quadratic residue.

sipa:
Done.

gmaxwell:
ACK.

PR #350, opened 2015-11-03
Fix use of uninitialized memory in bench_verify

apoelstra:


gmaxwell:
This actually may be my error.  The API does not claim this is in/out.

So, either the API doc should be updated (sigh), OR we should stop zeroizing (sigh) OR we should only zeroize 33 bytes (sigh).  The first has the problem that its still slightly less consistent with the calls that have fixed size output (like the compact seralize) in that they do not have an in or out size.

apoelstra:
Closing in favor of #350.

PR #351, opened 2015-11-05
Turn secp256k1_ec_pubkey_serialize outlen to in/out

sipa:
An alternative to #350.


apoelstra:
utACK. Prefer this to #350.

apoelstra:
Actual ACK

PR #356, opened 2015-11-15
Benchmark against OpenSSL verification

sipa:


gmaxwell:
Can we make any use of openssl ./configure conditional, make it clear in the configure help that its for comparison tests and benchmarks?  I don't care if it defaults to off or on.

sipa:
I fail to parse your sentence.

sipa:
@gmaxwell I was assuming we already had a configure flag for OpenSSL, and I didn't understand what you thought had to be changed about it, but I assume you mean we just need to make using OpenSSL configurable? That seems reasonable... @theuni feel like having a look?

gmaxwell:
ACK.

PR #359, opened 2015-11-25
[SCHNORR BREAK] Use quadratic-residue-y as symmetry breaker

sipa:
(Builds on top of #322)

This changes the implied y coordinate in Schnorr signatures to be the one which is a quadratic residue (the older logic required it to be even). This should be faster, as it doesn't need a modular inverse.

### DIFF
--- a/include/secp256k1_schnorr.h
+++ b/include/secp256k1_schnorr.h
@@ -9,7 +9,8 @@ extern "C" {
 
 /** This header file defines an API for a custom EC-Schnorr-SHA256 constructions.
  *  It supports non-malleable 64-byte signatures which support public key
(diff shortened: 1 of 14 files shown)


peterdettman:
I guess it's worth reminding that our sqrt returns the particular root (if any exist) that is a quadratic residue already, which seems relevant to "Verification (method 2)".

sipa:
@peterdettman Yes, I implemented it that way. It could use some explanation
indeed.

sipa:
@peterdettman By the way, do you have any intuition for a fast (and hopefully not too complex) algorithm for computing a Jacobi symbol (a|p) (where p is the size of our field)?

sipa:
Going to do this differently.

PR #360, opened 2015-11-25
Add experimental features to configure

sipa:


apoelstra:
Concept ACK. Without looking at why the build failed, I expect it's because you forgot to update `.travis.yml` to include the experimental flag :)

gmaxwell:
The "error: Schnorr signature module is experimental"  should tell you what you need to do to fix it.  "enable experimental configure options" sounds like it will enable them, perhaps it should say "allow".
I would actually soften "are not considered safe for production use" to may not be safe for production use.

sipa:
Done.

PR #363, opened 2015-12-07
ECDH from compressed points with "free" sqrt

peterdettman:
Builds on #362 as an alternative to #262.

Compressed points are moved onto an isomorphism (maybe the twist) instead of calling sqrt. After the scalar mult, parallel rsqrt/inv is used to reconstruct the correct output y-coordinate in parallel with the inversion of the z-coordinate. Total overhead vs uncompressed input is ~1%.

This implementation foregoes the jacobi check, meaning that the actual scalar multiplication could take place on a quadratic twist. This admits points with x == 0 as a possibility, but still not y == 0; the twist has a 37-bit, odd, cofactor. If this occurs, the parallel rsqrt/inv at the end will catch it and reject it. I think ECDH peers can already waste each other's time at will, so I don't think the delayed catch is a problem.

Do we know whether our formulae work "correctly" for the twist? Here we actually just need them not to crash (e.g. if an x==0 or P==INF turns up) in the scalar mult. Of course, we could always put the jacobi test back in, which has ~2% additional overhead.

Note:
The initial "decompression" produces (x,y) == (K.X, K^2) on the curve y^2 = x^3 + 7.K^3, with K != 0. If K is not a quadratic residue then this is a twist of the original curve.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 bench_inv
 bench_ecdh
+bench_ecdh_opt
 bench_sign
 bench_verify
 bench_schnorr_verify
--- a/include/secp256k1_ecdh.h
+++ b/include/secp256k1_ecdh.h
@@ -24,6 +24,14 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ecdh(
   const unsigned char *privkey
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
 
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ecdh_opt(
+  const secp256k1_context* ctx,
+  unsigned char *result,
+  const unsigned char *pub,
+  size_t publen,
+  const unsigned char *privkey
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(5);
+
 # ifdef __cplusplus
 }
 # endif
--- a/src/bench_ecdh_opt.c
+++ b/src/bench_ecdh_opt.c
@@ -13,7 +13,7 @@
 
 typedef struct {
     secp256k1_context *ctx;
-    secp256k1_pubkey pubkey;
+    unsigned char point[33];
     unsigned char scalar[32];
 } bench_ecdh_data;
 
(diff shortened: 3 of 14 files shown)


apoelstra:
I think our formulae work on the twist, but we have never checked this. In particular Pieter's sage verification code assumes `b = 7`. (For fun, I checked with a bunch of other `b` values, but only ones that differed by a factor of a sixth power, which give isomorphic curves.)

Can you get points on arbitrary twists or is there some fixed set we can check with sage?

Or maybe just checking one member of all the isomorphism classes is sufficient?

peterdettman:
"The twist" is an isomorphism class of the curves given by all the possible non-square K values. It should suffice to check y^2 = x^3 - 7. Generally, group formulae don't make use of B (==7.K^3) and so should calculate correctly on the twist. Point validation (using B obviously) doesn't appear to crop up *during* a scalar mult. The derivation of secp256k1_gej_add_ge apparently involves substitution of the curve equation, so I think it bears another look (B's probably cancel out).

The x==0 possibility deserves attention; I haven't seen any problems yet. 3 divides the cofactor of the twist, and indeed {P3:P3.x == 0} is a point of order 3, with 2(P3) == -P3 (i.e. doubling negates the y coordinate).

I think though that the endomorphism breaks correctness. The twist *does* have an endomorphism, with of course the same 2 `beta` value possibilities (which depend only on the field characteristic), but the corresponding `lambda` values are different. In particular, multiplying a twist-point by the "normal" `lambda` does not preserve the y-coordinate and does not map x --> beta.x. I've generated the full endomorphism parameters for y^2 = x^3 - 7 if anyone wants them for an endomorphism-compatible test of the twisted curve.

Nevertheless, in this case we don't need a right answer (it will be discarded). Scaling a point's x-coordinate by `beta` still leaves a valid point (x^3 == (beta.x)^3), so group operations will all still be on valid (twist) points. The scalar decomposition is incorrect, but I can't see how that breaks anything.

apoelstra:
The `b`s do cancel out in the derivation of `gej_add_ge`. It's used to convert between "y1^2 - y2^2" to "x1^3 - x2^3", i.e. a difference of squares in y is equivalent to a difference of cubes in x. This obviously depends on `a` being zero, but `b` is free.

As for the rest of your post I need to digest it. It's not obvious why this is the only twist we care about (or the only twist that exists?) but maybe I am just being dumb.

peterdettman:
y^2 = x^3 + 7 mod 43, generator e.g. (2,12), order 31, cofactor 1, beta 6, lambda 5.
y^2 = x^3 - 7 mod 43, generator e.g. (4,10), order 19, cofactor 3, beta 6, lambda 7. (0, 6) is a point of order 3.
Should be "suitably interesting" enough for starters. The same equation with either 67 or 79 as the field characteristic also gives cofactor 1 curves with cofactor 3 twists (the 2 points of order 3 will have x==0 of course, and the odd cofactor means no points with y==0).

apoelstra:
Nice, thanks Peter! I'll submit a PR to add exhaustive checks for the twist using the sum of (4, 10) and (0, 6) as generator.

We've been discussing on IRC the fact that some of our internal functions are missing `_var` annotations, which should be fixed before 1.0 ... and that it would be nice if we could mechanically check that non-`_var` things only ever call non-`_var` things. Perhaps we want something similar for "depends on the value of `b`" so we could know what functions are twist-correct?

peterdettman:
Perhaps we could flag group elements as having an implicit z factor (under VERIFY flags, similar to the way field element magnitudes are tracked) - just a flag, not the actual value. Then any functions that wouldn't work (or wouldn't work _right_) for such inputs can catch it. That flag can then be set wherever we use any of the isomorphism tricks, and carried through by group operations that support it.

PR #364, opened 2015-12-09
JNI rebased

greenaddress:
This is another continuation of #64 and #248 implementing JNI library for secp256k1, rebased to master.

Also it includes a fix for a bug we noticed while including it in GreenBits - buffers were overflowing if two different functions were called consecutively with the second one needing more buffer space.


faizkhan00:
Thanks for pulling this forward and testing. It looks like Travis is still having a few problems in picking up the jni object, but I'm going to close my original PR in favor of this, as its rebased and tests are passing (running make check-java) locally at least

greenaddress:
@faizkhan00 thank you and @theuni (and anyone else involved I missed) for the work done thus far.

tests are passing locally but there seems to be a race condition in make where by the shared library is not fully linked by the time the test starts (but strangely only in travis, we can't reproduce locally) - looks like we need to fix the dependency

we tried to fix that and failed (we are not too strong on makefiles) so we asked @theuni and he suggested he may have a look at this at some point but anyone with any input is welcome

theuni:
@greenaddress Sorry for not getting to this at the meetup as promised, I kept getting pulled away.

Anyway, that's probably for the best. This turned out to be non-trivial to track down, and not build-system related at all.

See the top few commits here for the fix: https://github.com/theuni/secp256k1/commits/jni-test

tl;dr: JNIEXPORT on some Java versions (the one Travis uses by default, at least) is busted wrt symbol visibility.

greenaddress:
@theuni no worries! Cheers for finding the actual source of the problem!

sipa:
Concept ACK, but can you rebase and squash the fixups?

src/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -12,49 +15,737 @@
...
+
+        byte[][] retByteArray = null;//secp256k1_ecdsa_recover_compact(byteBuff, Secp256k1Context, compressed, recID);
+
+        byte[] pubArr = retByteArray[0];
+        int pubLen = new BigInteger(new byte[] { retByteArray[1][0] }).intValue();
+        int retVal = new BigInteger(new byte[] { retByteArray[1][1] }).intValue();
+
+        assertEquals(pubArr.length, pubLen, "Got bad signature length.");
+
+        assertEquals(retVal, 1, "Failed return value check.");
```
sipa: This function can legitimately return 0, if the passed signature's R coordinate does not correspond to a point on the curve (around 50% chance for arbitrary byte arrays).

src/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -12,49 +15,737 @@
...
+          r.unlock();
+        }
+    }
+
+
+    /**
+     * libsecp256k1 Compute Pubkey - computes public key from secret key
+     *
+     * @param seckey ECDSA Secret key, 32 bytes
+     * @param compressed 1 to return compressed key, 0 for uncompressed
```
sipa: Seems unimplemented.

src/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -12,49 +15,737 @@
...
+            nativeECDSABuffer.set(byteBuff);
+        }
+        byteBuff.rewind();
+        byteBuff.put(seckey);
+
+        byte[][] retByteArray;
+
+        r.lock();
+        try {
+          retByteArray = secp256k1_ec_privkey_import(byteBuff,Secp256k1Context.getContext(), seckey.length);
```
sipa: I don't think that function still exists.

src/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -12,49 +15,737 @@
...
+        }
+
+        byte[] privArr = retByteArray[0];
+
+        int privLen = (byte) new BigInteger(new byte[] { retByteArray[1][0] }).intValue() & 0xFF;
+        int retVal = new BigInteger(new byte[] { retByteArray[1][1] }).intValue();
+
+        assertEquals(privArr.length, privLen, "Got bad pubkey length.");
+
+        assertEquals(retVal, 1, "Failed return value check.");
```
sipa: This function can legitimately fail (though only in adverserial or very unlikely conditions).

src/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -12,49 +15,737 @@
...
+
+        assertEquals(pubArr.length, pubLen, "Got bad pubkey length.");
+
+        assertEquals(retVal, 1, "Failed return value check.");
+
+        return pubArr;
+    }
+
+    /**
+     * libsecp256k1 create ECDH secret - constant time ECDH calculation
```
sipa: Does not match the name of the function below.

src/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -12,49 +15,737 @@
...
+     *
+     * @param seckey byte array of secret key used in exponentiaion
+     * @param pubkey byte array of public key used in exponentiaion
+     */
+    //TODO schnorrSign, schnoorVerify, schnorrRecover()
+    public static byte[] schnoorOps(byte[] pubkey, byte[] msg32) throws AssertFailException{
```
sipa: Name is wrong (Schnorr, nor Schnoor).

src/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -12,49 +15,737 @@
...
+            byteBuff.order(ByteOrder.nativeOrder());
+            nativeECDSABuffer.set(byteBuff);
+        }
+        byteBuff.rewind();
+        byteBuff.put(pubkey);
+
+        byte[][] retByteArray;
+        r.lock();
+        try {
+          retByteArray = secp256k1_ecdsa_recover(byteBuff,Secp256k1Context.getContext(), pubkey.length);
```
sipa: Why all this commented-out code?

faizkhan00: When @greenaddress opened the PR, I definitely can say it was unfinished (not covering 100% of the secp256k1 headers) but I had intended to cover that. Depending on if @greenaddress was planning on completeing the coverage, I'd be happy to contribute the rest of the code as well (as a patch if desired)

greenaddress: @sipa as you say the work is not complete, it doesn't cover all functions yet but the aim is to have all of them.

we plan to do more towards completing coverage but would be great to work together with @faizkhan00 (and @theuni) to finish it.

we currently have pedersen_blind_sum, pedersen_commit and rangeproof_sign in the works

sipa: @greenaddress I'm very interested in that work, but it shouldn't go into the main library's commit history. Please clean things up.

src/java/org/bitcoin/NativeSecp256k1.java
```diff
@@ -12,49 +15,737 @@
...
+
+    /**
+     * recover the given secp256k1 pubkey in native code.
+     *
+     * @param data The data which was signed, must be exactly 32 bytes
+     * @param signature The signature
+     * @param compressed whether to recover a compressed pubkey
+     * @param pub The public key which did the signing
+     */
+    //TODO recoverCompact()
```
sipa: What's TODO about it?

greenaddress:
`secp256k1_ecdsa_recover` was not implemented in org_bitcoin_NativeSecp256k1.c, hence the TODO and commented out code.

I've now removed the unimplemented and commented parts and cleaned it up in general. I think it's reasonable to keep these two commits for future reference, in case someone wants to look up the incomplete implementations in git history.

PR #368, opened 2015-12-21
Create secp256k1.spec

AliceWonderMiscreations:
RPM spec file for Fedora / RHEL / CentOS - builds on CentOS 7

### DIFF
--- a/secp256k1.spec
+++ b/secp256k1.spec
@@ -1,0 +1,69 @@
+%define		gitdate	20015.12.20
+
+Name:		secp256k1
+Version:	0.0.%{gitdate}
+Release:	1%{?dist}
+Summary:	Optimized C library for EC operations on curve secp256k1
+
+Group:		System Environment/Libraries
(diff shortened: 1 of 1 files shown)


sipa:
I don't have much of an opinion on including distribution-specific build files in the repository.

@laanwj @theuni Opinions?

AliceWonderMiscreations:
My current opinion is don't. At the time I wrote the spec file I didn't realize the library was going to be bundled inside bitcoin and assumed it was going to be an external build dependency.

AliceWonderMiscreations:
Also at the time I wrote this spec file, there was an experimental php module that used the library. I was playing with the php module for a shopping cart system I was designing, but have since re-designed the system so that everything key generation is done completely separately from the php part and can be done with python.

So I don't need the ability to have this library separate from bitcoin and I doubt very many people do.

sipa:
@AliceWonderMiscreations Ok, thank you!

PR #376, opened 2016-01-31
rename COPYING to LICENSE

prusnak:
It's an unwritten rule to use COPYING for copyleft licenses (such as GPL) and LICENSE for non-copyleft licenses (such as MIT or BSD).

Also I changed the formatting slightly to match the template used at Github as this formatting seems to be what most projects are using nowadays.


prusnak:
Bitcoin Core people don't want this change. Retracting PR.

PR #377, opened 2016-02-03
Put precomputed ecmult table in build dir

kr:
Include ecmult_static_context.h using the include path
so that it can be generated with a VPATH-style build in
a separate directory, when the original source directory
is not writable.

I'm not 100% sure this is the right way to accomplish
this. I'm down to make any changes needed to do it
properly.


sipa:
@theuni Opinion?

theuni:
This already works fine, I just verified with 'make distcheck' which checks for that scenario. The naming of "src" is arbitrary.

Does this fail for you with the current code?

kr:
Never mind, it works fine for me too.

I apologize for the noise.

PR #378, opened 2016-02-16
.gitignore build-aux cleanup

upgradeadvice:
Don't ignore everything under build-aux. 
"build-aux/" in .gitignore will lead to missing macros in some cases
ignore generated macros by name


laanwj:
Makes sense; ignoring a whole directory is only appropriate when the entire directory and its contents are generated.

Issue #255, opened 2015-05-24
Bindings for PHP


afk11:
I've been working on an extension for PHP which wraps libsecp256k1, and think it's just about complete. I noticed the PR's for java bindings, and I wondered if it might be worth moving the code here? 

I'm not sure if anyone else is developing these, but mainly asking since I'm experimenting with packaging the extension as .deb package, but I need to ensure that libsecp256k1 is installed on the system as well.

I know it's quite early for a secp256k1-dev package in apt, but since I'm now considering gitian builds to produce a .deb for this extension, the same could be done for the library itself. If we keep the sources here we can build each binding/release in the one gitian build. 

The bindings are here: https://github.com/Bit-Wasp/secp256k1-php, and there are tests for pretty much everything besides the privkey_import/export functions(). Yet to include the edge cases (r/s = 0, etc), but it's standing up pretty well. It is my first foray into C in a while, so I'd appreciate all the feedback I can get!

Anyway, interested to hear what people think, I can follow up with a PR if people are happy enough!

faizkhan00:
hey @afk11 , I noticed your bindings before I got started on the java bindings, and had a discussion on IRC w/ @gmaxwell about this awhile back as well; it looks like upstream may not be 100% ready to start taking bindings because of the evolving API (eg. adding contexts just a few mo. ago) and I only figured that out after I was halfway through the implementation process.

I originally started the work to try to close out #64, and since that has (basically) been deprecated now, I was going to ask (somewhere, here) if it made more sense to just take that PR and maintain them in a fork until the API is more mature... possibly w/ a link in the Readme to direct new users, if that makes sense, too.

DavidEGrayson:
Hello, I wrote some [Ruby bindings for the secp256k1](https://github.com/DavidEGrayson/secp256k1-rb).  In case it's useful to you, the main features I wanted to achieve for that project were:

* Provides access to all features of the secp256k1 library.
* Does not add any new features.
* Avoids making arbitrary decisions when possible.
* Provides a safe interface: there should be no way for a user of the wrapper to cause undefined behavior or memory leaks.  (One exception is when multiple threads are used in an unsafe way.)
* Exception messages produced by this wrapper are less verbose than they could be to help avoid leaking secret information.

I noticed your PHP binding is creating some kind of global context instead of letting the user create their own.  That only seems like a good idea to me if that global context can somehow be saved so it doesn't have to be recreated for every HTTP request that uses it.  Is that the case?

[...]

PR #381, opened 2016-03-06
[readme] Add section Bindings

fanatid:
I think section in readme about bindings is better than folding bindings code in this repository (like with Java).

Related issue: #255 

Bindings authors: @toxeus @xenog @wanderer @afk11 @ludbb @DavidEGrayson @wink @dpc

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,3 +59,14 @@ libsecp256k1 is built using autotools:
     $ make
     $ ./tests
     $ sudo make install  # optional
+
+Bindings
+--------
+
+- Go: [toxeus/go-secp256k1](https://github.com/toxeus/go-secp256k1)
+- Haskell: [haskoin/secp256k1-haskell](https://github.com/haskoin/secp256k1-haskell)
+- JavaScript: [cryptocoinjs/secp256k1-node](https://github.com/cryptocoinjs/secp256k1-node)
+- PHP: [Bit-Wasp/secp256k1-php](https://github.com/Bit-Wasp/secp256k1-php)
+- Python: [ludbb/secp256k1-py](https://github.com/ludbb/secp256k1-py)
+- Ruby: [DavidEGrayson/secp256k1-rb](https://github.com/DavidEGrayson/secp256k1-rb) [GemHQ/secp256k1-rb](https://github.com/GemHQ/secp256k1-rb)
+- Rust: [dpc/bitcoin-secp256k1-rs](https://github.com/dpc/bitcoin-secp256k1-rs)


gmaxwell:
I disagree that we should be recommending bindings that we have not reviewed.  Previously some bindings have compromised the security of the library.  (e.g. the go bindings silently using insecure nonce generation).

dpc:
@apoelstra : What's the story with https://github.com/dpc/bitcoin-secp256k1-rs ? Did you supersede it with some other repository?

fanatid:
@gmaxwell there no words about *recommended*, but maybe it make sense add note that bindings wasn't reviewed?
OOI, whom you had in mind under *we*?

gmaxwell:
Directing people to things which may be insecure like that is just 'laundering' when we know the note would just be ignored-- it might be necessary but it's not sufficient: Assuming one of the bindings is incorrect (or even malicious) listing it with a note would be a boobytrap and incompatible with our design principle of being hard to misuse.

Bindings should be subject to peer review by cryptographers; same as the extern C interface of libsecp256k1, since it's functionally the same when used in different environments.  The review will be easier when we finish documenting our design assumptions and objectives; since then the bindings could be checked against satisfying those properties.

dpc:
@gmaxwell : People will just google "bitcoin bindings in X" or use search function in the package manager of the language of their choice etc. 

IMO having a "big red warning" on the list here, and requiring projects listed in the list to have one in the README, is a win, as it has some chance of informing the potential users.

DavidEGrayson:
Oh, I see you found my Ruby bindings.  I haven't updated them since April 2015 so I suspect that they don't work with the latest version of the library.  I was going wait for you guys to release version 1.0.0 of the library before I work on that again.  Once everything is polished up, I'd be happy to help cryptographers review it.

gmaxwell:
We can include a warning without recommending any bindings:  "Note that bindings to other languages may undermine the security of the library and have done so in the past. Great care should be taken when using bindings found on the internet as they may not be suitable for your purposes."

toxeus:
>  (e.g. the go bindings silently using insecure nonce generation)

Are you referring to [this](https://github.com/toxeus/go-secp256k1/blob/ac9288b1243e520faadc66a33eb41fcc18ec1e97/secp256k1_test.go#L28)?

apoelstra:
@dpc Sorry I missed this message. Yes, the rust libsecp bindings are now at https://github.com/apoelstra/rust-secp256k1

laanwj:
I think if you decide to add a list of 'recommended' bindings then there should be a clear list of review criteria the projects have to adhere to, and a list of tests that it has to pass in the bound language.

But to take the task of 'vetting' projects is quite a responsibility and I can understand if @gmaxwell and al aren't willing to include that in their scope.

This may be better for a wiki article or such, with the appropriate warnings.

On the other hand, good bindings are important for adoption of the library. Putting the bar too high may result in people just ignoring advice, or rolling their own without cooperation and review of others.

fanatid:
I'll glad if somebody will add this list into wiki.

PR #386, opened 2016-04-26
Add some missing `VERIFY_CHECK(ctx != NULL)`

apoelstra:
The `ARG_CHECK` macro requires that a variable called `ctx` exist and be
non-NULL. However, in several functions that do not use the context variable,
we simply ignore it with `(void)ctx`. Replace these with explicit checks for
non-NULLness to avoid invalid memory accesses.


sipa:
utACK

PR #388, opened 2016-04-27
bench_ecdh: fix call to secp256k1_context_create

apoelstra:


apoelstra:
Confirmed that this is the only call to `context_create` in the codebase that does not use one of the new constants.

FelixWeis:
tested, works

peterdettman:
Oops, I had noted this as long ago as #363 and forgot to send a separate PR.

apoelstra:
@sipa @gmaxwell This is a one-line patch -- can I get a quick ACK?

sipa:
Quick ACK

PR #389, opened 2016-04-28
Cast pointers through uintptr_t under JNI

jgriffiths:
Fixes warnings of the form "warning: cast to pointer from integer of
different size" when building on 32 bit platforms. This is the same
approach used for pointer conversions in the openjdk sources.


sipa:
utACK

PR #391, opened 2016-05-02
JNI: Only call ecdsa_verify if its inputs parsed correctly

jgriffiths:
Return 0 otherwise instead of calling it with bad parameters.


sipa:
ACK

PR #392, opened 2016-05-03
Add testcase which hits additional branch in secp256k1_scalar_sqr

jonasnick:
This commit adds an testcase that increases coverage by executing [this muladd2](https://github.com/bitcoin-core/secp256k1/blob/7b549b1abc06fe1c640014603346b85c8bc83e0b/src/scalar_4x64_impl.h#L865) such that `c1 < th2`.


sipa:
utACK

PR #397, opened 2016-05-21
Small fixes for sha256

fanatid:


gmaxwell:
utACK.   sipa: having these very short non-namespaced defines and then undefing them is kind of asking to break something. Should we also make them namespaced? (with adding secp256k1 as a prefix?)

sipa:
utACK

PR #398, opened 2016-05-25
Test whether ECDH and Schnorr are enabled for JNI

sipa:


afk11:
+1. I ran into side effects from compiling the jni bindings without all dependencies enabled (unknowingly, I must have installed java at some point). Using libsecp256k1 failed at runtime, the root cause seemed to be the jni library was built without the dependencies.

PR #400, opened 2016-06-06
A couple minor cleanups

apoelstra:


gmaxwell:
on the non-null checks, something is wrong-- the build setup should not be compiling libsecp256k1 itself with the non-null prototype annotation, as that _will_ make GCC optimize out any form of nullness check.

apoelstra:
Hmm, I started seeing this warning since upgrading my system to a recent Arch build. `gcc (GCC) 6.1.1 20160501`.

apoelstra:
As per IRC conversation, changed the `-Wno-nonnull` commit to instead add `-DSECP256K1_BUILD` to the benchmark builds, since that was the actual problem.

sipa:
I don't think -DSECP256K1_BUILD should be used for the verify and sign benchmarks, as those don't build the internal code, but link against the library. Only the internal benchmarks build the internal code.

sipa:
ACK

PR #401, opened 2016-06-30
ecmult_const: unify endomorphism and non-endomorphism skew cases

apoelstra:
We now do a skew correction even without the endomorphism optimization,
which costs one additional group addition but unifies a lot of code.

### DIFF
--- a/src/ecmult_const_impl.h
+++ b/src/ecmult_const_impl.h
@@ -58,22 +58,24 @@ static int secp256k1_wnaf_const(int *wnaf, secp256k1_scalar s, int w) {
     int global_sign;
     int skew = 0;
     int word = 0;
+
     /* 1 2 3 */
     int u_last;
     int u;
 
-#ifdef USE_ENDOMORPHISM
     int flip;
     int bit;
     secp256k1_scalar neg_s;
     int not_neg_one;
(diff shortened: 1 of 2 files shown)


src/ecmult_const_impl.h
```diff
@@ -202,59 +181,57 @@ static void secp256k1_ecmult_const(secp256k1_gej *r, const secp256k1_ge *a, cons
...
         secp256k1_gej_double_var(&tmpj, &tmpj, NULL);
         secp256k1_ge_set_gej(&correction, &tmpj);
         secp256k1_ge_to_storage(&correction_1_stor, a);
+#ifdef USE_ENDOMORPHISM
         secp256k1_ge_to_storage(&correction_lam_stor, a);
+#endif
         secp256k1_ge_to_storage(&a2_stor, &correction);
 
         /* For odd numbers this is 2a (so replace it), for even ones a (so no-op) */
         secp256k1_ge_storage_cmov(&correction_1_stor, &a2_stor, skew_1 == 2);
```
sipa: I think skew_1 == 2 can never be true in the non-USE_ENDOMORPHISM case.

apoelstra:
I think it can, e.g. if the scalar is low and odd.

sipa:
ACK

PR #404, opened 2016-07-06
Replace 3M + 4S doubling formula with 2M + 5S one

apoelstra:


apoelstra:
Apparently I wrote this a year ago and completely forgot about it ... was cleaning up the old branches on my local repo and found it. It's complete and works and is a bit faster.

sipa:
Please add the sage translation of this code to the algebraic tests.

sipa:
With master:
```
group_double_var: min 0.224us / avg 0.224us / max 0.225us
```

With this PR:
```
group_double_var: min 0.238us / avg 0.239us / max 0.240us
```

sipa:
Please update with just a note that an alternative formula was tried, but found to be slower in practice.

sipa:
ACK

PR #405, opened 2016-07-09
Make secp256k1_fe_sqrt constant time

sipa:
The current field element square root algorithm is entirely constant time already, apart from a variable time field element comparison at end (whose time is negligible compared to the exponentiation ladder).

Furthermore, no faster variable time algorithm is known, so we don't lose anything by actually making the algorithm fully constant time. This simplifies some potential future changes in the Schnorr code.


apoelstra:
ACK pending travis

PR #408, opened 2016-07-28
Add `secp256k1_ec_pubkey_negate` and `secp256k1_ec_privkey_negate`

apoelstra:
I have an application which uses summed secret keys as a keysplitting mechanism. To use this with BIP32, I would like to negate the public parent key, which would give the following result:
- one of the key shards is the negative parent key, the other the child key
- the "summed key" is then the BIP32 tweak

In my case I can change my application to use a difference of keys rather than sum of keys, but this adds another way for the user to get things wrong, especially when used with a more "symmetric" key split where it's unclear which of the two shards ought to be negated.

### DIFF
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -424,6 +424,33 @@ int secp256k1_ec_pubkey_create(const secp256k1_context* ctx, secp256k1_pubkey *p
     return ret;
 }
 
+int secp256k1_ec_privkey_negate(const secp256k1_context* ctx, unsigned char *seckey) {
+    secp256k1_scalar sec;
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(seckey != NULL);
+
+    secp256k1_scalar_set_b32(&sec, seckey, NULL);
+    secp256k1_scalar_negate(&sec, &sec);
+    secp256k1_scalar_get_b32(seckey, &sec);
+
+    return 1;
+}
+
+int secp256k1_ec_pubkey_negate(const secp256k1_context* ctx, secp256k1_pubkey *pubkey) {
+    int ret = 0;
+    secp256k1_ge p;
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(pubkey != NULL);
+
+    ret = secp256k1_pubkey_load(ctx, &p, pubkey);
+    memset(pubkey, 0, sizeof(*pubkey));
+    if (ret) {
+        secp256k1_ge_neg(&p, &p);
+        secp256k1_pubkey_save(pubkey, &p);
+    }
+    return ret;
+}
+
 int secp256k1_ec_privkey_tweak_add(const secp256k1_context* ctx, unsigned char *seckey, const unsigned char *tweak) {
     secp256k1_scalar term;
     secp256k1_scalar sec;
--- a/src/tests.c
+++ b/src/tests.c
@@ -10,6 +10,7 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <string.h>
 
 #include <time.h>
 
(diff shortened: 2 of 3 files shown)


sipa:
Concept ACK if you also add a `secp256k1_ec_privkey_negate`. Also, what about using the `secp256k1_ec_*_tweak_negate` name, for consistency with `_tweak_add` and `_tweak_mul` ?

apoelstra:
Sure, I can add a privkey negate.

As for the `tweak_` names, I had thought that the "tweak" was the extra parameter, not the original element -- you are multiplying by a tweak, or adding a tweak.

apoelstra:
Added negate_privkey.

That test failure is a randomness test failure, I think it's just a fluke.

instagibbs:
change PR title to match new functionality?

instagibbs:
This is failing a test.

apoelstra:
kicked travis

src/secp256k1.c
```diff
@@ -424,6 +424,30 @@ int secp256k1_ec_pubkey_create(const secp256k1_context* ctx, secp256k1_pubkey *p
...
+    return 1;
+}
+
+int secp256k1_ec_pubkey_negate(const secp256k1_context* ctx, secp256k1_pubkey *pubkey) {
+    int ret = 0;
+    secp256k1_ge p;
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(pubkey != NULL);
+
+    ret = secp256k1_pubkey_load(ctx, &p, pubkey);
```
gmaxwell: The pubkey tweak functions zeroize the output on load failure,  this code would operate on uninitialized memory.

src/tests.c
```diff
@@ -3457,6 +3459,13 @@ void test_ecdsa_end_to_end(void) {
     memset(&pubkey, 0, sizeof(pubkey));
     CHECK(secp256k1_ec_pubkey_parse(ctx, &pubkey, pubkeyc, pubkeyclen) == 1);
 
+    /* Verify negation changes the key and changes it back */
+    memcpy(&pubkey_tmp, &pubkey, sizeof(pubkey));
+    CHECK(secp256k1_ec_pubkey_negate(ctx, &pubkey_tmp) == 1);
+    CHECK(memcmp(&pubkey_tmp, &pubkey, sizeof(pubkey)) != 0);
+    CHECK(secp256k1_ec_pubkey_negate(ctx, &pubkey_tmp) == 1);
+    CHECK(memcmp(&pubkey_tmp, &pubkey, sizeof(pubkey)) == 0);
+
```
gmaxwell: Test a null argument and test a bad pubkey object.

gmaxwell:
No strong view, but these are multiplicative tweaks with a fixed tweak value of -1.

Concept ACK, will ACK with nits fixed.

apoelstra:
Fixed nits.

gmaxwell:
@apoelstra ping see above: "for consistency, memset(pubkey, 0, sizeof(*pubkey)); after the load, unless you've got an argument that its bad to clear it completely (in which case we should change it everywhere)"

apoelstra:
Sorry, I forgot to address this. My argument was that the pubkey going into this function is a public key, isn't secret, and there is no need to zeroize it. But this is true even for `pubkey_tweak_add`, so I guess the purpose of the zeroing is to turn any invalid pubkey into a "standard invalid pubkey", which seems like a sane functional behaviour.

Added the `memset`.

voisine:
in the case of BIP32, a single private key in combination with a master public key exposes all private keys. So, it's possible that a "public" key may indeed be secret.

gmaxwell:
> so I guess the purpose of the zeroing is to turn any invalid pubkey into a "standard invalid pubkey",

exactly.

@voisine Nope, unrelated. The issue there is the _extended_ public key. The EC point being described here does not have the properties you're referring to-- it cannot be used to convert extend another private key.

gmaxwell:
ACK.

sipa:
utACK

PR #411, opened 2016-08-26
Remove guarantees about memcmp-ability

apoelstra:
We had a discussion in person some months ago about the guarantees on the representation of the publicly exposed types, and I think guaranteeing memcmp is too strong. There are lots of internal things that can't be memcmp'd and it stands to reason that we might want the freedom to use similar representations for public types in future.

Note that removing this guarantee makes comparison of points that might be infinity quite annoying.


sipa:
ACK.

If there is demand we may need to add `_compare` functions to avoid a roundtrip to the serializer to test equality, but I don't think that's needed yet.

PR #413, opened 2016-09-07
fix auto-enabled static precompuatation

theuni:
This was broken in aa0b1fd14979145d54ef85485cb497a9cdbc22c7

The configure output even lied about it :(


sipa:
utACK

PR #416, opened 2016-09-13
Fix infinity loop in secp256k1_schnorr_sign

fanatid:
[secp256k1_schnorr_sign](https://github.com/bitcoin-core/secp256k1/blob/7d15cd78599bdaeec72259e22a77bbad70c0428a/src/modules/schnorr/main_impl.h#L23) stop the loop only if [secp256k1_schnorr_sig_sign](https://github.com/bitcoin-core/secp256k1/blob/7d15cd78599bdaeec72259e22a77bbad70c0428a/src/modules/schnorr/schnorr_impl.h#L62)  return non-zero. But if secret key is zero `secp256k1_schnorr_sig_sign` will always [return zero](https://github.com/bitcoin-core/secp256k1/blob/7d15cd78599bdaeec72259e22a77bbad70c0428a/src/modules/schnorr/schnorr_impl.h#L70) and `secp256k1_schnorr_sign` never stops.

### DIFF
--- a/src/modules/schnorr/main_impl.h
+++ b/src/modules/schnorr/main_impl.h
@@ -35,19 +35,21 @@ int secp256k1_schnorr_sign(const secp256k1_context* ctx, unsigned char *sig64, c
         noncefp = secp256k1_nonce_function_default;
     }
 
-    secp256k1_scalar_set_b32(&sec, seckey, NULL);
-    while (1) {
-        ret = noncefp(nonce32, msg32, seckey, secp256k1_schnorr_algo16, (void*)noncedata, count);
-        if (!ret) {
-            break;
-        }
-        secp256k1_scalar_set_b32(&non, nonce32, &overflow);
-        if (!secp256k1_scalar_is_zero(&non) && !overflow) {
-            if (secp256k1_schnorr_sig_sign(&ctx->ecmult_gen_ctx, sig64, &sec, &non, NULL, secp256k1_schnorr_msghash_sha256, msg32)) {
+    secp256k1_scalar_set_b32(&sec, seckey, &overflow);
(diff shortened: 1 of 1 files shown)


apoelstra:
utACK. Note this also catches the case when the user passes the field order as secret key (which will also be interpreted as zero). I think this is a reasonably likely footgun to run into, especially because in many cases applications can accept secret keys from untrusted sources and not worry that anything will break.

Also note that `secp256k1_ecdsa_sign` has the same behaviour.

fanatid:
@apoelstra I don't sure, but shouldn't we also check that secret key is less than field order?

fanatid:
Updated.
I also changed [conditions order for nonce](https://github.com/bitcoin-core/secp256k1/blob/7d15cd78599bdaeec72259e22a77bbad70c0428a/src/modules/schnorr/main_impl.h#L46), because with `overflow` first makes check faster (`overflow` already exists, while for zero we should call function)

fanatid:
ping :)

sipa:
Needs rebase. Note that I plan to remove the Schnorr code in this repository, as it shouldn't be used for production use in its current form (and both API and signature compatibility will break before it's brough back).

fanatid:
Rebased.
@sipa does it means that secp256k1 bindings (like [secp256k1-node](https://github.com/cryptocoinjs/secp256k1-node)) should not use Schnorr right now?

sipa:
Nobody should use Schnorr as implemented now. It is an experiment, and
subject to change at any time (as the configure script will warn you about
if you try to enable it). The reason I want to remove it from the codebase
is because I keep hearing about people who want to use it for production,
so this is apparently not clear.

fanatid:
Thank you for explanation.

sipa:
No longer needed.

PR #417, opened 2016-09-14
Remove useless code in secp256k1_schnorr_sig_recover

fanatid:
assign zero value to `overflow` before `secp256k1_scalar_set_b32` doesn't make sense


apoelstra:
I think some lint tool complained about passing the address of uninitialized variables. Unfortunately C doesn't have any way to signal out-parameters.

I'd prefer the useless code stay, just as a matter of "defense in depth" coding practice of not passing addresses of unitiliazed variables.

sipa:
No longer needed.

PR #422, opened 2016-10-21
Restructure nonce clearing

bgorlick:
Make sure we clear the nonce data even if the nonce function fails (it may have written partial data), and call memset only once in the case we iterate to produce a valid signature.


apoelstra:
utACK

sipa:
utACK

PR #425, opened 2016-11-18
Remove Schnorr experiment

sipa:
This module implemented a naive custom Schnorr signature scheme, though several flaws have been discovered with its approach since (lack of commitment to public keys, ability for 3rd parties to adapt signatures to related keys, and a multisigning API that is vulnerable to cancellation attacks).

I have been working on a better scheme, but I don't believe this should be upstream until it's been vetted in more thorough ways. People assumed that it being included in the repository was a sign that it was final, so I'm removing it to avoid any confusion.


gmaxwell:
ACK.

apoelstra:
ACK.

PR #426, opened 2016-11-26
Increase test coverage and add a configure flag to ease coverage analysis

apoelstra:


configure.ac
```diff
@@ -154,6 +159,14 @@ AC_COMPILE_IFELSE([AC_LANG_SOURCE([[void myfunc() {__builtin_expect(0,0);}]])],
     [ AC_MSG_RESULT([no])
     ])
 
+if test x"$enable_coverage" = x"yes"; then
+    AC_DEFINE(COVERAGE, 1, [Define this symbol to compile out all VERIFY code])
+    CFLAGS="$CFLAGS -O0 --coverage"
+    LDFLAGS="--coverage"
+else
+    CFLAGS="$CFLAGS -O3"
```
sipa: Is this needed?

apoelstra:
Reduced iterations of recovery API tests from `64 * count` to just `count`; stopped leaking four contexts' worth of memory on every one of these iterations.

Travis should be happier now :)

gmaxwell:
ACK.

Issue #432, opened 2016-12-06
configure should check for opacity of ECDSA_SIG and not use the openssl tests if it's opaque


gmaxwell:
http://bitcoin.stackexchange.com/questions/49937/cannot-build-secp256k1-c-library-on-debian-stretch

laanwj:
Same issue as #429

gmaxwell:
Resolved by #433

PR #433, opened 2016-12-12
Make the libcrypto detection fail the newer API.

gmaxwell:
OpenSSL 1.1 makes ECDSA_SIG opaque and our tests need access
inside this object.

The comparison tests against OpenSSL aren't important for most
users, but the build failing is...


gmaxwell:
Fixes #432

sipa:
utACK 12de86387f302140ab87d73c555ff1b3ffae1c20

PR #437, opened 2017-01-10
Unroll secp256k1_fe_(get|set)_b32 to make them much faster.

gmaxwell:
fe_get_b32 was taking 7 times longer than a fe_sqr due to working two bits at a time.

Rico666 on Bitcointalk noticed this and provided a patch for 5x52, I formatted it and wrote the rest.

This is perhaps a 0.5% speedup for ecdsa_verify using 32bit field code.

### DIFF
--- a/src/bench.h
+++ b/src/bench.h
@@ -23,7 +23,7 @@ void print_number(double x) {
     if (y < 0.0) {
         y = -y;
     }
-    while (y < 100.0) {
+    while (y > 0 && y < 100.0) {
         y *= 10.0;
         c++;
     }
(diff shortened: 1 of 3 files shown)


sipa:
I believe that the benchmark results are due to the entire loop being optimized out (reducing the iteration count increases the timings proportionally).

sipa:
Can you adapt the benchmark so it's not optimized away?

gmaxwell:
I stripped the benchmarks out.

sipa:
Benchmarked `bench_verify` with GMP disabled on a i7-6820HQ CPU, pegged to 2.6GHz.
* master: 97.7us (64-bit), 366us (32-bit).
* this PR: 97.5us (64-bit), 365us (32-bit).
So a 0.2% (64-bit) and 0.3% improvement (32-bit).

sipa:
utACK a2b6b19

PR #438, opened 2017-01-26
[trivial] clear stack variable by function

isle2983:
The bare `bits = 0;` raises a clang static analysis issue.

```
--------------------------------------------------------------------------------
An issue has been found in ./src/ecmult_gen_impl.h:153:5
Type:         Dead assignment
Description:  Value stored to 'bits' is never read

0: ./src/ecmult_gen_impl.h:153:5 - Value stored to 'bits' is never read
--------------------------------------------------------------------------------
```

The line was introduced in 2f6c8019114743e46a741fc6a01f9c6600ccdb5d (when `secp256k1_ecmult_gen()` lived in `src/ecmult_impl.h`) with the purpose of clearing stack variables that might have secret info.

However, without that context, the purpose for clearing isn't obvious from reading the code. Aside from making clang happy, the function name should help it be understood.

### DIFF
--- a/src/ecmult_gen_impl.h
+++ b/src/ecmult_gen_impl.h
@@ -150,7 +150,7 @@ static void secp256k1_ecmult_gen(const secp256k1_ecmult_gen_context *ctx, secp25
         secp256k1_ge_from_storage(&add, &adds);
         secp256k1_gej_add_ge(r, r, &add);
     }
-    bits = 0;
+    secp256k1_int_clear(&bits);
     secp256k1_ge_clear(&add);
     secp256k1_scalar_clear(&gnb);
 }
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -214,8 +214,13 @@ static void secp256k1_ge_clear(secp256k1_ge *r) {
     secp256k1_fe_clear(&r->y);
 }
 
-static void __attribute__((optimize("O0"))) secp256k1_secure_clear(void *s, size_t n) {
-    (void) memset(s, 0, n);
+typedef void *(*memset_t)(void *, int, size_t);
+
+static volatile memset_t secure_memset = memset;
+
+static void secp256k1_secure_clear(void *s, size_t n) {
+    (void) secure_memset(s, 0, n);
+
 }
 
 static void secp256k1_int_clear(int *r) {
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -7,6 +7,7 @@
 #ifndef _SECP256K1_GROUP_IMPL_H_
 #define _SECP256K1_GROUP_IMPL_H_
 
+#include <string.h>
 #include "num.h"
 #include "field.h"
 #include "group.h"
@@ -211,6 +212,14 @@ static void secp256k1_ge_clear(secp256k1_ge *r) {
     r->infinity = 0;
     secp256k1_fe_clear(&r->x);
     secp256k1_fe_clear(&r->y);
+}
+
+static void __attribute__((optimize("O0"))) secp256k1_secure_clear(void *s, size_t n) {
+    (void) memset(s, 0, n);
+}
+
+static void secp256k1_int_clear(int *r) {
+    secp256k1_secure_clear(r, sizeof(*r));
 }
 
 static int secp256k1_ge_set_xquad(secp256k1_ge *r, const secp256k1_fe *x) {


gmaxwell:
It's a fine structural change, but if you're going to break this out, can you perhaps make it a generic unsigned char clear that takes a point and length (which you'd use sizeof() on here)? --- and try to get the compiler to not optimize it out.

isle2983:
The new branch has two commits with two tentative examples for comment. They focus just on the `bits` variable for the purpose of discussion even though the problem extends to the other `*_clear()` functions.


In the first example commit 72c0ebbea413a3d05847a849d629082e5408232e,  `__attribute__((optimize("O0")))` achieves what we want and the call to `secp256k1_clear()` can be seen at the bottom of the `-03` assembly for `secp256k1_ecmult_gen()`:


```
.LBE2707:
	.loc 6 135 0
	cmpq	$64, %rdi	#, ivtmp.241
	je	.L451	#,
	salq	$34, %rdi	#, D.9544
	movl	%r11d, 104(%rsp)	# D.9544, %sfp
	movl	%ebp, %r14d	# D.9544, D.9552
	shrq	$37, %rdi	#, D.9544
	movl	%r8d, 112(%rsp)	# D.9544, %sfp
	movl	%r12d, %r15d	# D.9544, D.9552
	movl	416(%rsp,%rdi,4), %edi	# gnb.d, D.9552
	movl	%r10d, 96(%rsp)	# D.9544, %sfp
	movl	%ecx, %r11d	# D.9544, D.9552
	movl	%esi, 88(%rsp)	# D.9544, %sfp
	movl	%edx, 80(%rsp)	# D.9544, %sfp
	movl	%r9d, %r13d	# D.9544, D.9552
	movl	%eax, 48(%rsp)	# D.9544, %sfp
	jmp	.L452	#
.L451:
.LVL2579:
.LBB2708:
.LBB2709:
	.loc 2 222 0
	leaq	412(%rsp), %rdi	#, tmp5277
.LVL2580:
	movl	$4, %esi	#,
	call	secp256k1_secure_clear	#
.LVL2581:
.LBE2709:
.LBE2708:
	.loc 6 156 0
	addq	$1144, %rsp	#,
```

(As was assumed, it was indeed being optimized out before, but I didn't quote it here to save space)


However, compiling the same with `clang`, it throws a warning and doesn't do what we want:

```
$ clang -I. -g -O3 -Wall -Wextra -Wno-unused-function -c src/gen_context.c -fverbose-asm -S
In file included from src/gen_context.c:13:
src/group_impl.h:217:28: warning: unknown attribute 'optimize' ignored
      [-Wunknown-attributes]
static void __attribute__((optimize("O0"))) secp256k1_secure_clear(void ...
                           ^
1 warning generated.
```

So, this approach looks like it is not very universal. ugh.


Taking a step back and looking at OpenSSL, they clear stack variables with a function named `BN_clear_free()` in `(openssl)/crypto/bn/bn_lib.c` which calls `OPENSSL_cleanse()` in `(openssl)/crypto/mem_clr.c` which is implemented with the `volatile` keyword:

```
/*
 * Pointer to memset is volatile so that compiler must de-reference
 * the pointer and can't assume that it points to any function in
 * particular (such as memset, which it then might further "optimize")
 */
typedef void *(*memset_t)(void *,int,size_t);

static volatile memset_t memset_func = memset;

void OPENSSL_cleanse(void *ptr, size_t len)
{
    memset_func(ptr, 0, len);
}
```

So, this approach is also tried in cdd46814c4abcd7bc9ce73d616347964ac7ac9be. The assembly in `-03` turns out a bit different, but still achieves the aim:

```
.LBE2372:
	.loc 6 135 0
	cmpq	$64, %rax	#, ivtmp.263
	je	.L424	#,
	salq	$34, %rax	#, D.9409
	shrq	$37, %rax	#, D.9409
	movl	400(%rsp,%rax,4), %r9d	# gnb.d, D.9415
	jmp	.L425	#
.L424:
.LVL2528:
.LBB2373:
.LBB2374:
.LBB2375:
	.loc 2 222 0
	leaq	396(%rsp), %rdi	#, tmp5505
.LVL2529:
	movq	secure_memset(%rip), %rax	# secure_memset, D.9410
	movl	$4, %edx	#,
	xorl	%esi, %esi	#
	call	*%rax	# D.9410
.LVL2530:
.LBE2375:
.LBE2374:
.LBE2373:
	.loc 6 156 0
	addq	$1176, %rsp	#,
```

This works perfectly with `clang` too, so this is definitely more viable.

Also, for C++ there is discussion on the internet around a new `memset_s()` call designed for this purpose but it is only introduced in C++11. In the long run, if `memset_s()` makes it to C, that will probably be the preferred way to do this.


Questions:

1. What kind of benchmarking needs to be done to quantify the performance impacts?
2. Should we look at applying this to all stack vars? or should the focus be a subset of functions that touch private keys?
3. Any cosmetic suggestions for function names/interface/presentation?

sipa:
memset_s is an optional function in C11, not C++11.

gmaxwell:
10:59 < gmaxwell> glibc 2.25 has explicit_bzero
11:01 < gmaxwell> void
11:01 < gmaxwell> explicit_bzero (void *s, size_t len)
11:01 < gmaxwell> {
11:01 < gmaxwell>   memset (s, '\0', len);
11:01 < gmaxwell>   /* Compiler barrier.  */
11:01 < gmaxwell>   asm volatile ("" ::: "memory");
11:01 < gmaxwell> }

isle2983:
closing in favor of #448

PR #440, opened 2017-01-29
Fix typos

practicalswift:


practicalswift:
How is `Tag_ABI_optimization_goals` used? Will fixing the typo have any consequence on the actual binary produced?

sipa:
>  is everything after the @ comment text intended for humans only?

I believe so.

@laanwj Can you comment?

laanwj:
> is everything after the @ comment text intended for humans only?

Yep

sipa:
utACK 4c0f32ed51409b340da84dd3d3f91f66b4cb4643

PR #441, opened 2017-02-14
secp256k1_context_randomize: document.

rustyrussell:
I prefer the idea of adding a randseed32 arg to create() and clone() (as well as leaving this here for reseeding), but if we're API stable that is a bigger change...

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -163,6 +163,8 @@ typedef int (*secp256k1_nonce_function)(
  *
  *  Returns: a newly created context object.
  *  In:      flags: which parts of the context to initialize.
+ *
+ *  See also secp256k1_context_randomize.
  */
 SECP256K1_API secp256k1_context* secp256k1_context_create(
     unsigned int flags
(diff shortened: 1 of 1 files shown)


include/secp256k1.h
```diff
@@ -543,11 +545,24 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_tweak_mul(
...
  *  Args:    ctx:       pointer to a context object (cannot be NULL)
  *  In:      seed32:    pointer to a 32-byte random seed (NULL resets to initial state)
+ *
+ * While secp256k1 code is written to be constant-time no matter what secret
+ * values are, it's possible that a future compiler may output code which isn't,
+ * and also that the CPU may not emit the same radio frequencies or draw the same
+ * amount power for all values.
+ *
+ * This function provides a seed which is combined into the blinding value: that
+ * blinding value added before each multiplication (and removed afterwards) so
```
sipa: missing `is` after `value`?

sipa:
ACK

PR #354, opened 2015-11-12
[ECDH API change] Support custom hash function

fanatid:
Solve #352

Issue #352, opened 2015-11-05
Support custom hash functions in ECDH API


Kagami:
Current ECDH implementation always hashes both `Px` and `Py` of point multiplication result with SHA256. It makes impossible to use `secp256k1` with existing protocols like [Bitmessage](https://bitmessage.org/wiki/Encryption) which uses `SHA512(Px)` as shared secret. As mentioned [here](https://github.com/bitcoin/secp256k1/pull/252#issuecomment-154052769), possible solution for this is to accept custom hash function in API similar to custom nonce function.

cc @apoelstra

peterdettman:
It seems to me that a major reason for handling the hash internally was to avoid "revealing" the raw point coordinate(s). Supporting a (callback-based) custom hash function lets them "escape" again (even if the callback is isolated somehow, it could still calculate a no-op of the input).

So it seems to me that internal hashing with a customisable hash function is a needless complication. Of course, if you're restricted to SHA256, you can't use this even for what should be an obvious target - TLS key exchange.

I propose we abandon the internal hashing approach altogether.

gmaxwell:
@peterdettman We are not trying to avoid revealing anything, but rather not provide interfaces that by that we know will primarily be used in less safe ways because using it that way is the simplest thing possible. 

[Because of things like http://lists.linuxfoundation.org/pipermail/bitcoin-dev/2014-March/004720.html (see also later posts where I gave some more attacks) I wish we didn't have a reason to expose ECDH at all, in favor of safe higher level constructs-- e.g. ECIES instead of "here is ecdh then you can makeup your own crypto scheme", because we should assume that the caller is not a subject matter expert in cryptography whos work is being reviewed by other subject matter experts-- but thats not always the best trade-off.. and sometimes we have to expose more to get functionality.  But the interface can still encourage more prudent behavior.]

An example of this is how we handle DSA nonces.  By default the library uses RFC6979 internally. But if you need to do something special, you can send in a callback that implements whatever you need-- upto and including "emits a constant nonce". :)   So the caller can do whatever they like, but if they do something dangerous or inadvisable at least it wasn't because it was the easiest thing to do (or worse, the only obvious thing to do). 

for this PR, I think it's a good approach; we'll probably want to change the api;  I've just been too slammed this week to give it a proper review yet. :)

peterdettman:
@gmaxwell OK, that's reasonable, although I think the argument is a bit weaker than for the ECDSA/RFC6979 case.

[...]

PR #446, opened 2017-03-09
ECDH: output compressed point

chfast:


fanatid:
Related: #352 #354

chfast:
Oh, thank. Good to know it is not only my issue. Any final conclusions how the API should look like?

gmaxwell:
It is generally unsafe to use the raw point for almost any application. Following our design principle of being hard to abuse, we should not provide a function that directly does so.

Instead, I think we can do is take a hasher function point, similar to how the signing takes a nonce function. Callers that need some other scheme for handling the nonce can pass a function pointer that accomplishes whatever they need, without creating an appearance that it's generally okay to use the raw point for ECDH.

There is a lot of reckless cryptography out there. We're not going to change that, but we also don't need to be a part of it.

fanatid:
@gmaxwell IIRC I implemented hash function as argument in #354, but not receive any feedback... should I reopen that PR?

chfast:
It doesn't matter for me which version it will be. Is the one with the callback is to be accepted? Is the callback parameters as in the #354 fine?

sipa:
Closing in favor of #354.

PR #447, opened 2017-03-13
Export SHA256 functions

chfast:


laanwj:
Should SHA256 functions be part of the library's API? The library is secp256k1, not sha256lib, there are plenty of libraries available that export SHA256.

I think we should avoid exporting internal implementation details. This reduces the flexibility to change things in the future.

chfast:
Why would you like to have 2 implementations of SHA256 in your codebase if you need both secp256k1 and SHA256? This increases code size and decreases performance.

sipa:
Sorry, NACK. libsecp256k1 is a library for elliptic curve crypto, not a generic cryptography library. It has a copy of SHA256 for randomization and a few related algorithms, but it is not optimized in any way. I fear that exposing the function will lead to people demanding that it gets optimized for various purposes and/or platforms, which I don't think should be the library's focus.


If code size matters that much to you (SHA256 is only a few kB), perhaps we should provide a way for libsecp256k1 to be compiled using an external SHA256 implementation instead.

dcousens:
> If code size matters that much to you (SHA256 is only a few kB), perhaps we should provide a way for libsecp256k1 to be compiled using an external SHA256 implementation instead.

For me,  supporting exporting SHA256 here was purely for the purposes of ensuring a consistent implementation being used. [not to doubt the quality of any implementation]

PR #452, opened 2017-04-18
Minor optimizations to _scalar_inverse to save 4M

peterdettman:
The existing secp256k1_scalar_inverse uses a straight-forward "blocks-of-1" approach, but the exponent here admits of a few tweaks.

Firstly, the pattern 01010 occurs 3 times, so if we have (u5=)x^5 available we can save 3M during accumulation. Actually u5 can be precalculated cheaply, replacing 1S with 1M. 

Secondly, the addition chain invites optimization since it ends with 127 (a critical threshold) and also contains unnecessary (to the chain)x4 and un-accumulated x7. The addition chain 2, 3, 6, 8, 14, 28, 56, 112, 126 looks inviting... and fortuitously, armed with u5, we can replace [x127 0 x] with [x126 u5] and [x4  0 x] with [x3 u5](two occurrences) - for free. The new shorter addition chain saves 2M.

There should be ~256 squares and I counted 44 multiplies in the existing version, and a saving of 1S + 4M with this PR. Since performance of scalar mul/sqr are very similar in my config, I'd expect about 5/300 ~= 1.6% performance improvement, and measure roughly 1.5%.

### DIFF
--- a/src/scalar_impl.h
+++ b/src/scalar_impl.h
@@ -66,76 +66,70 @@ static void secp256k1_scalar_inverse(secp256k1_scalar *r, const secp256k1_scalar
 #else
     secp256k1_scalar *t;
     int i;
-    /* First compute x ^ (2^N - 1) for some values of N. */
-    secp256k1_scalar x2, x3, x4, x6, x7, x8, x15, x30, x60, x120, x127;
+    /* First compute xN as x ^ (2^N - 1) for some values of N,
+     * and uM as x ^ M for some values of M. */
+    secp256k1_scalar x2, x3, x6, x8, x14, x28, x56, x112, x126;
+    secp256k1_scalar u2, u5;
 
-    secp256k1_scalar_sqr(&x2,  x);
-    secp256k1_scalar_mul(&x2, &x2,  x);
+    secp256k1_scalar_sqr(&u2, x);
+    secp256k1_scalar_mul(&x2, &u2,  x);
+    secp256k1_scalar_mul(&u5, &u2, &x2);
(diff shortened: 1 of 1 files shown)


briansmith:
ICYMI: I submitted a PR to improve upon this by 1 square and 3 multiplications (total savings 2 squares + 7 multiplications), but because it was based on Peter's unmerged PR, the PR ended up being a PR against his repo instead of this one: https://github.com/peterdettman/secp256k1/pull/1.

gmaxwell:
These changes are awesome. Thanks. Will test and review carefully soon.

sipa:
Benchmarked `bench_verify` with GMP disabled on a i7-6820HQ CPU, pegged to 2.6GHz.
* master: 97.7us
* this PR: 97.3us
* https://github.com/peterdettman/secp256k1/pull/1: 97.0us

ACK on both. The result is just a sequence of mul/sqr calls, which by definition is an exponentiation ladder. If the result were incorrect, it would implement an incorrect exponent, which nearly every test run should catch.

briansmith:
FWIW, I also reviewed Peter's PR and it is correct. I also agree with @sipa's assessment that many things would fail if the sequence of multiplications and squarings were incorrect. Nonetheless, I also manually verified that Peter's addition chain is correct when improving it.

gmaxwell:
@peterdettman Can you pull briansmith's commit into this PR?

peterdettman:
@gmaxwell Done. Cumulative 255S+44M -> 253S+37M, measured speed of scalar_inverse at +3% as expected.

PR #456, opened 2017-05-05
Fix secp256k1.h includes for contrib/lax_*

sergey-chernikov:


gmaxwell:
I believe? this is incorrect.  The correct headers should be in the include path, not with 'include' stuck on the front.  Consider the behavior with the library installed.

sergey-chernikov:
Besides bench_*.c files containing "#include <include/secp256k1.h>" there are 2 other .h files having the same inclusion style in src/java directory. Other 2 .h files having plain "#include "secp256k1.h" are already in the include directory.
So it's not trivial to distinguish what inclusion style is one-way correct.

sipa:
That's because the JNI bindings are part of the library itself. They're in the `src/` directory.

The `contrib/` directory is not part of the library. It contains snippets that you can copy and modify into your own projects should they be useful. They *use* libsecp256k1.

gmaxwell:
The headers include secp256k1.h, the inclusions in the C files are potentially redundant but harmless. I don't currently see a need for this PR. If I'm confused please feel free to ping me (or open a new one).

PR #459, opened 2017-06-08
Add pubkey prefix constants to include/secp256k1.h

apoelstra:
In future multisig implementations we will need to pass nonces around, which are algebraically pubkeys but should not be decodable as pubkeys. The way to do this is to change the prefix byte from the ordinary 0x02/0x03 to something else. However, some forks (notably `secp256k1-zkp`) have started using some bytes for their own encodings, and if we continue to use hardcoded constants the risk of conflict is increased.

This commit puts the prefixes used by the main library into the `include/secp256k1.h` so that the constants we're using will at least be in a standard easy-to-reference place.

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -159,6 +159,13 @@ typedef int (*secp256k1_nonce_function)(
 #define SECP256K1_EC_COMPRESSED (SECP256K1_FLAGS_TYPE_COMPRESSION | SECP256K1_FLAGS_BIT_COMPRESSION)
 #define SECP256K1_EC_UNCOMPRESSED (SECP256K1_FLAGS_TYPE_COMPRESSION)
 
+/** Prefix byte used to tag various encoded curvepoints for specific purposes */
+#define SECP256K1_TAG_PUBKEY_EVEN 0x02
+#define SECP256K1_TAG_PUBKEY_ODD 0x03
+#define SECP256K1_TAG_PUBKEY_UNCOMPRESSED 0x04
+#define SECP256K1_TAG_PUBKEY_HYBRID_EVEN 0x06
+#define SECP256K1_TAG_PUBKEY_HYBRID_ODD 0x07
+
 /** Create a secp256k1 context object.
  *
  *  Returns: a newly created context object.
--- a/src/eckey_impl.h
+++ b/src/eckey_impl.h
@@ -15,16 +15,17 @@
 #include "ecmult_gen.h"
 
 static int secp256k1_eckey_pubkey_parse(secp256k1_ge *elem, const unsigned char *pub, size_t size) {
-    if (size == 33 && (pub[0] == 0x02 || pub[0] == 0x03)) {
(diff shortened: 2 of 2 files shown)


include/secp256k1.h
```diff
@@ -159,6 +159,13 @@ typedef int (*secp256k1_nonce_function)(
 #define SECP256K1_EC_COMPRESSED (SECP256K1_FLAGS_TYPE_COMPRESSION | SECP256K1_FLAGS_BIT_COMPRESSION)
 #define SECP256K1_EC_UNCOMPRESSED (SECP256K1_FLAGS_TYPE_COMPRESSION)
 
+/** Prefix byte used to tag various encoded curvepoints for specific purposes */
+#define SECP256K1_TAG_PUBKEY_EVEN	0x02
+#define SECP256K1_TAG_PUBKEY_ODD	0x03
+#define SECP256K1_TAG_PUBKEY_UNCOMPRESSED	0x04
+#define SECP256K1_TAG_PUBKEY_UNCOMPRESSED_EVEN	0x06
```
sipa: This should probably be called `SECP256K1_TAG_HYBRID_EVEN`.

gmaxwell:
Somewhere I figured out what standard defined terrible things like the hybrid values. Anyone remember?  This should get a comment.

sipa:
ANSI X9.62 Sections 4.3.6 and 4.3.7.

peterdettman:
Actually, I would argue that the relevant standard is "Standards for Efficient Cryptography". The domain parameters for secp256k1 are given in "SEC 2: Recommended Elliptic Curve Domain Parameters", and the octet string conversions are in "SEC 1: Elliptic Curve Cryptography" (2.3).

Of course it is basically the same overall scheme as ANSI X9.62, but notably there is no mention of hybrid encodings; only 00, 02, 03, 04 may appear as the first octet. The conversion routines explicitly require an output of "invalid" for anything else.

apoelstra:
Bitcoin supports the hybrid encodings, so I guess we have to support them too.

include/secp256k1.h
```diff
@@ -159,6 +159,13 @@ typedef int (*secp256k1_nonce_function)(
 #define SECP256K1_EC_COMPRESSED (SECP256K1_FLAGS_TYPE_COMPRESSION | SECP256K1_FLAGS_BIT_COMPRESSION)
 #define SECP256K1_EC_UNCOMPRESSED (SECP256K1_FLAGS_TYPE_COMPRESSION)
 
+/** Prefix byte used to tag various encoded curvepoints for specific purposes */
+#define SECP256K1_TAG_PUBKEY_EVEN	0x02
```
sipa: Weird spacing...

apoelstra: They're just tabs.

I guess the other #defines use single spaces, I'll change this.

sipa:
utACK bc61b91aca21a103514c5d47a20f9c7b84d5feb5

PR #463, opened 2017-06-23
Reduce usage of hardcoded size constants

tjps:
In particular the usage of keylen in nonce_function_rfc6979 seemed precarious - in one conditional it was unconditionally set, then in the next it was added to.  While it was clearly correct as written, I think this change makes it easier to reason about for new eyes and more resistant to breakage if there is any future change to what gets fed into the PRNG.

### DIFF
--- a/src/hash_impl.h
+++ b/src/hash_impl.h
@@ -133,9 +133,10 @@ static void secp256k1_sha256_write(secp256k1_sha256 *hash, const unsigned char *
     hash->bytes += len;
     while (bufsize + len >= 64) {
         /* Fill the buffer, and process it. */
-        memcpy(((unsigned char*)hash->buf) + bufsize, data, 64 - bufsize);
-        data += 64 - bufsize;
-        len -= 64 - bufsize;
+        size_t chunk_len = 64 - bufsize;
+        memcpy(((unsigned char*)hash->buf) + bufsize, data, chunk_len);
+        data += chunk_len;
+        len -= chunk_len;
         secp256k1_sha256_transform(hash->s, hash->buf);
         bufsize = 0;
     }
@@ -162,11 +163,11 @@ static void secp256k1_sha256_finalize(secp256k1_sha256 *hash, unsigned char *out
 }
 
 static void secp256k1_hmac_sha256_initialize(secp256k1_hmac_sha256 *hash, const unsigned char *key, size_t keylen) {
-    int n;
+    size_t n;
     unsigned char rkey[64];
-    if (keylen <= 64) {
+    if (keylen <= sizeof(rkey)) {
         memcpy(rkey, key, keylen);
-        memset(rkey + keylen, 0, 64 - keylen);
+        memset(rkey + keylen, 0, sizeof(rkey) - keylen);
     } else {
         secp256k1_sha256 sha256;
         secp256k1_sha256_initialize(&sha256);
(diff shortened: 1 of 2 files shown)


apoelstra:
I'm ambivalent about the `buffer_append`, I think it feels weird to use that in one place but nowhere else (and I'm not convinced it would aid readability/reviewability to do this throughout the codebase).

For everything else, ut/concept ACK.

src/hash_impl.h
```diff
@@ -162,11 +163,11 @@ static void secp256k1_sha256_finalize(secp256k1_sha256_t *hash, unsigned char *o
 }
 
 static void secp256k1_hmac_sha256_initialize(secp256k1_hmac_sha256_t *hash, const unsigned char *key, size_t keylen) {
-    int n;
+    unsigned int n;
```
chfast: Can you make it `size_t` to match the result type of `sizeof()`?

src/secp256k1.c
```diff
@@ -120,7 +120,7 @@ static int secp256k1_pubkey_load(const secp256k1_context* ctx, secp256k1_ge* ge,
          * representation inside secp256k1_pubkey, as conversion is very fast.
          * Note that secp256k1_pubkey_save must use the same representation. */
         secp256k1_ge_storage s;
-        memcpy(&s, &pubkey->data[0], 64);
+        memcpy(&s, &pubkey->data[0], sizeof(secp256k1_ge_storage));
```
chfast: I'm not sure it is better, but you can do `sizeof(s)`.

tjps: I actually do have a preference ordering of:

memcpy(&s, ..., <numeric constant>)    <   memcpy(&s, ..., sizeof(type))   <  memcpy(&s, ..., sizeof(s))

Will update to sizeof(s), was an oversight on my part.

tjps:
@apoelstra I agree that having buffer_append for just the one function feels odd but I do think it makes the function it is used in both clearer and safer for future updates.  I did a brief perusal of the rest of the code to see if there were any other places it could be used (move it into a general utility spot) but came up empty.  Do any others spring to mind?

src/hash_impl.h
tjps: Done.

apoelstra:
Well, RFC 6979 has not changed in the four years since it became a RFC, so I'm doubtful that there will be updates. But regardless, this spreads the buffer-setting logic across 20 LOC and a function-call boundary as opposed to the original 2 LOC.

It's easy for reviewers to keep a couple numbers in their heads and do simple arithmetic, and it is obvious what these values should be and what they are for. It's not as easy to verify multple function calls which share outptrs, so whatever additional assurance this gives is cancelled out by the increased code complexity.

tjps:
Hard for me to argue that - I wish there were more usage points to justify it.

Would you be amenable to me changing 'keylen = 96;' in the original code to 'keylen += 32;' for parity across the conditional blocks?

sipa:
I agree that the `buffer_append` seems a bit unrelated, but I do think it makes things easier to follow. As RFC6979 is just an instantiation of a NIST DRBG with a concatenation of key and message as input, I don't think it hurts to write it in a way that makes that concatenation clearer (rather than following the RFC6979 language exactly).

ACK

apoelstra:
tACK `2906340`

tjps:
I like to periodically clean up my outstanding PRs. It seems there was consensus on this one, but no action taken.  Should I leave it?

sipa:
Hi, can you rebase?

tjps:
@sipa - rebased!

apoelstra:
tACK `https://github.com/bitcoin-core/secp256k1/pull/463/commits/b7593cdb44dd0bdc98e4abe705833552efce6de3`

tjps:
Just to note, my last change was to fix an incorrectly set committer line.  No code diff from b7593cd.

apoelstra:
Confirmed no code diff. tACK https://github.com/bitcoin-core/secp256k1/pull/463/commits/c7680e570f6086d9ce4ceb9077359552ce79e8b0 :P

PR #464, opened 2017-07-07
Strauss-wNAF multimul

sipa:
This extends `secp256k1_multi` to support arbitrary numbers of input points.

Features:
* Any point can be infinity (resulting in it being ignored).
* The G factor can be NULL.
* Will split into batches of a maximum size and combine them.
* Uses effective affine addition across all inputs, and to build the odd-multiples tables.
* Supports endomorphism (TODO: figure out up to what size it is worth it).
* Uses Strauss' algorithm to only do 255 doublings (129 with endomorphism) in each batch.

### DIFF
--- a/src/ecmult_impl.h
+++ b/src/ecmult_impl.h
@@ -283,50 +283,80 @@ static int secp256k1_ecmult_wnaf(int *wnaf, int len, const secp256k1_scalar *a, 
     return last_set_bit + 1;
 }
 
-static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej *r, const secp256k1_gej *a, const secp256k1_scalar *na, const secp256k1_scalar *ng) {
-    secp256k1_ge pre_a[ECMULT_TABLE_SIZE(WINDOW_A)];
-    secp256k1_ge tmpa;
-    secp256k1_fe Z;
+struct secp256k1_ecmult_point_state {
 #ifdef USE_ENDOMORPHISM
-    secp256k1_ge pre_a_lam[ECMULT_TABLE_SIZE(WINDOW_A)];
     secp256k1_scalar na_1, na_lam;
-    /* Splitted G factors. */
-    secp256k1_scalar ng_1, ng_128;
     int wnaf_na_1[130];
     int wnaf_na_lam[130];
     int bits_na_1;
     int bits_na_lam;
-    int wnaf_ng_1[129];
-    int bits_ng_1;
-    int wnaf_ng_128[129];
-    int bits_ng_128;
 #else
     int wnaf_na[256];
     int bits_na;
-    int wnaf_ng[256];
-    int bits_ng;
 #endif
-    int i;
-    int bits;
+    size_t input_pos;
+};
 
 #ifdef USE_ENDOMORPHISM
-    /* split na into na_1 and na_lam (where na = na_1 + na_lam*lambda, and na_1 and na_lam are ~128 bit) */
-    secp256k1_scalar_split_lambda(&na_1, &na_lam, na);
-
-    /* build wnaf representation for na_1 and na_lam. */
-    bits_na_1   = secp256k1_ecmult_wnaf(wnaf_na_1,   130, &na_1,   WINDOW_A);
-    bits_na_lam = secp256k1_ecmult_wnaf(wnaf_na_lam, 130, &na_lam, WINDOW_A);
-    VERIFY_CHECK(bits_na_1 <= 130);
-    VERIFY_CHECK(bits_na_lam <= 130);
-    bits = bits_na_1;
-    if (bits_na_lam > bits) {
-        bits = bits_na_lam;
-    }
+#define STRAUSS_WNAF_MAX_POINTS 24
 #else
-    /* build wnaf representation for na. */
-    bits_na     = secp256k1_ecmult_wnaf(wnaf_na,     256, na,      WINDOW_A);
-    bits = bits_na;
+#define STRAUSS_WNAF_MAX_POINTS 32
 #endif
+
+static void secp256k1_ecmult_strauss_wnaf(const secp256k1_ecmult_context *ctx, secp256k1_gej *r, int num, const secp256k1_gej *a, const secp256k1_scalar *na, const secp256k1_scalar *ng) {
+    secp256k1_ge pre_a[STRAUSS_WNAF_MAX_POINTS][ECMULT_TABLE_SIZE(WINDOW_A)];
+#ifdef USE_ENDOMORPHISM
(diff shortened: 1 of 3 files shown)


sipa:
@bbuenz This implements a faster multi-multiplication, though there is no intention of exposing it in the public API (which we keep restricted to higher-level safe-by-default functionality). It is useful for building other things on top of in derived projects, though.

bbuenz:
@sipa Thanks, I'll consider forking it.

gmaxwell:
I suggested to Pieter today that we switch to using the split-G table at all times, w/ endo or without.  I believe(tm) that given equal WNAF dimensions for a 2-point multiexp split-G will be faster even when used without the endomorphism by about 1 doubling, because half the time the G have the highest non-zero value.   Making this change would make it easier for a endomorphism enabled codebase to drop the endomorphism when it isn't a help. 

The endo will stop being helpful very quickly... (a fact that will probably radically lower the crossover point to bos-coster when endo is in use, since endo helps bos-coster out to a very high count).

peterdettman:
@sipa Nice to see the cross-table effective-affine working, since previously I had only proved it would!

So IIUC, this PR precomputes an odd-multiples table for each point in the batch, and uses wNAF scalar recoding, so that there's one addition per input point per (w+1) bits (or something). I don't know if you recall the FourQ paper Patrick Longa presented at RWC this year (https://eprint.iacr.org/2015/565.pdf). There they are using 2 endos, so for a standard scalar multiplication they end up with 4 points in a multi-exp. Instead of precomputing a table for each of those points, they make one table with 8 values, each combination of P1 with +/-P2, +/-P3, +/-P4 (negation is used if you need the -P1 combinations). The scalars are recoded ("signed-digit recoding") so that each bit is +/-1. The main ladder then performs 1 addition per bit, assembling the +/-1 from each scalar into a table lookup.

An idea then is to take batches and group them by 4s (or 5s?). On paper it looks like it would save quite a bit of precomputation time and memory usage, while possibly costing a few extra additions in the ladder (wNAF is sliding window?).  Unfortunately this is not as easy to add/test as the current scheme since we lack various pieces (the table(s) can still be made effective-affine, at a slightly higher cost). It also might still need the current wNAF approach for "leftovers" that don't make a full 4-or-5 (and maybe G too).

gmaxwell:
Though that kind of signed digit recoding and joint table entries moves the endomorphism ahead of the precomputation, instead of after. 

IIRC... Current code, without endo does  ~42 additions point non-G point plus one double and 7 adds for the precomp, plus the multiplies for the EA conversion. G does ~15 adds.  So unless I'm off on the numbers, if you would have to 256 adds, it'll need to be adding in more than 5 points; and the precomp cost grows.  I don't see how to get it particularly close.

peterdettman:
> Though that kind of signed digit recoding and joint table entries moves the endomorphism ahead of the precomputation, instead of after.

I think you'd just "chunk" the points first, then build the precomp table for each chunk, then if you want to use the endomorphism, you take the (lambda*P[1..n])s as a new chunk, and get its precomp table directly from the original precomp (in the same way secp256k1_ecmult does).

> IIRC... Current code, without endo does ~42 additions point non-G point plus one double and 7 adds for the precomp, plus the multiplies for the EA conversion. G does ~15 adds. So unless I'm off on the numbers, if you would have to 256 adds, it'll need to be adding in more than 5 points; and the precomp cost grows. I don't see how to get it particularly close.

I think I over-estimated the size of the precomp tables in the current code. Still, if we chunked by 6s (to get the ladder additions to about the same number), then the precomp table would be 32 points, instead of 48 (8*6 separate tables). The form of the table suggests you need 31 calculations of the form P +/- Q (producing both points); it happens that each of those is only a little more expensive than a single addition (and they're mostly mixed additions if the input points are affine). I haven't worked out the details, but I would guess the cost of precomputation is about the same as current code.

I find that interesting because the sliding wNAF relies on "skipping zeroes" in the scalar (an average of 1 extra bit per window), whereas this alternative is more compatible with a constant-time multiexp. Now I'm wondering if there are multi-point versions of the Joint Sparse Form, or equivalents with wider window sizes.

peterdettman:
I should add that the signed recoding scheme only needs one "windows" array per chunk instead of one wNAF array per point, which is a further significant memory saving (possibly it could  even be calculated left-to-right on-the-fly).

gmaxwell:
I don't think it's possible to do the 6 point precomp in 32 adds.

peterdettman:
Depends on how you're counting. I'm supposing the use of an "extended" addition that calculates XA(P,Q) = (P+Q,P-Q). This needs 1M+1S more than a regular addition, but gives 2 output points.

If C(n) is the number of XA needed to precompute the table T(n) for P[1..n], then:
C(1) = 0
C(n+1) = C(n) + 2^(n-1)

i.e. T(n+1) is built from T(n) using 1XA per entry in T(n).

So, C(n) = 0 + 1 + ... + 2^(n-2) = 2^(n-1) - 1. T(6) has 32 entries, needing 31XA, which is ~ 37A in terms of cost. Or you could say it's 62 additions, half of which are very cheap. This is better than 6D+42A for 6 wNAF, but I expect the effective-affine transformation to be a little more clunky, and the wNAF tables could eventually use coZ, so it might be a wash on costs.

Chunks of 7 might actually be optimal speed-wise, and taking into account the reduced window array usage, total memory still comes in less than for the current scheme.

peterdettman:
It appears there are also width-3 variants of the Joint Sparse Form, which is something like chunking by 2s, but giving a sparser recoding (joint hamming weight ~0.36), with a 12 entry lookup table and single window array (or even none) per 2 input points. Probably the same performance as current scheme, with less memory needed.

gmaxwell:
Ah, the extended addition is what I was missing.

src/ecmult_impl.h
```diff
@@ -403,4 +462,31 @@ static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej
     }
 }
 
+static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej *r, const secp256k1_gej *a, const secp256k1_scalar *na, const secp256k1_scalar *ng) {
+    secp256k1_ecmult_strauss_wnaf(ctx, r, 1, a, na, ng);
+}
+
+static void secp256k1_ecmult_multi(const secp256k1_ecmult_context *ctx, secp256k1_gej *r, size_t n, const secp256k1_gej *a, const secp256k1_scalar *na, const secp256k1_scalar *ng) {
```
apoelstra: Can you split this function out into a separate commit so that it can easily be replaced with a dispatch function?

sipa:
Superseded by #473.

PR #466, opened 2017-07-10
Don't fail JNI dependency checks if we can build a JNI program

jgriffiths:
Not all setups will result in JNI_INCLUDE_DIRS having a value, in
particular Android builds from standalone toolchain directories have
the JNI headers installed in the standard include directories.

Instead of failing if the configure macro didn't find a directory
based on the installed Java runtime, only fail if we cannot build a
JNI-including program.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -401,6 +401,9 @@ if test x"$use_jni" != x"no"; then
     have_jni_dependencies=no
   fi
   if test "x$have_jni_dependencies" = "xno"; then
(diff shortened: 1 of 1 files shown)


sipa:
@theuni Care to have a look?

theuni:
This can re-enable jni if it's been turned off because ecdh is disabled. I think the quick fix is just to move the ```enable_module_ecdh``` check down a bit.

real-or-random:
Closing this because we removed the JNI bindings in #682.

Issue #467, opened 2017-07-12
Redundant conditional in ecdsa_impl.h


bytting:
The following code block can be seen in src/secp256k1/src/ecdsa_impl.h starting at line 83

```
while (lenleft > 0) {
    if ((ret >> ((sizeof(size_t) - 1) * 8)) != 0) {
    }
    ret = (ret << 8) | **sigp;
    if (ret + lenleft > (size_t)(sigend - *sigp)) {
        /* Result exceeds the length of the passed array. */
        return -1;
    }
    (*sigp)++;
    lenleft--;
}
```

The conditional _if_ statement at line 84 doesn't appear to have a purpose other than spending cycles

sipa:
Eh, agree...

corebob:
Yeah, probably. I was wondering if the line right after it should actually have been inside, but I can't really tell. Seems unlikely though.

Is it worth making a PR?

apoelstra:
Looks like this should've been closed when #468 was merged.

PR #468, opened 2017-07-13
Remove redundant conditional expression

corebob:


apoelstra:
Fixes https://github.com/bitcoin-core/secp256k1/issues/467

utACK

PR #471, opened 2017-08-11
add secp256k1_recovery.h to include_HEADERS

dwightguth:
If you intend to use this as a library in order to do public key recovery, it's necessary to have access to secp256k1_recovery.h, but make install wasn't installing it. This commit modifies Makefile.am in order to add this file to the list of headers installed by make install.


sipa:
If you configure with `--enable-module-recovery`, it should be part of `install`.

dwightguth:
Thanks! Sorry, I didn't see that.

PR #472, opened 2017-08-16
[build] Set --enable-jni to no by default instead of auto.

kallewoof:
Having `--enable-jni` be `auto` doesn't make a lot of sense, and results in things like https://github.com/bitcoin/bitcoin/pull/11056.


kallewoof:
Squashed.

gmaxwell:
ACK

PR #474, opened 2017-08-26
Fix header guards using reserved identifiers

danra:
Identifiers starting with an underscore and followed immediately by a capital letter are reserved by the C++ standard.

The only header guards not fixed are those in the headers auto-generated from java.


sipa:
ACK abe2d3e84baf5a3a54514cd7abd60dce9b544387

PR #475, opened 2017-09-06
README: Require uint128_t support, not just uint64_t

ysangkok:
On trying to run the test suite in the TIS Interpreter, I realized that I can't, because of the uint128_t requirement.


sipa:
If you configure for a 32-bit system, only uint64_t is needed.

ysangkok:
Thanks for the clarification!

PR #477, opened 2017-09-19
Implements multiset module

tomasvdw:
This adds support for Elliptic Curve Multiset operations

Currently benches at about 800k/sec

### DIFF
--- a/src/bench_multiset.c
+++ b/src/bench_multiset.c
@@ -1,0 +1,52 @@
+/**********************************************************************
+ * Copyright (c) 2014-2015 Pieter Wuille                              *
+ * Distributed under the MIT software license, see the accompanying   *
+ * file COPYING or http://www.opensource.org/licenses/mit-license.php.*
+ **********************************************************************/
+
+#include "include/secp256k1.h"
+#include "include/secp256k1_multiset.h"
+#include "util.h"
+#include "bench.h"
+
+secp256k1_context *ctx;
+
+#define UNUSED(x) (void)(x)
+
+void bench_multiset(void* arg) {
+    static int it=0;
+    unsigned n,m;
+    unsigned char result[32];
+    secp256k1_roller roller;
+
+    UNUSED(arg);
+    secp256k1_multiset_create_roller(ctx, &roller, (unsigned char*)"", 0);
+
+    for (m=0; m < 100000; m++)
+    {
+        unsigned char buf[32*3];
+        secp256k1_roller x;
+
+        for(n = 0; n < sizeof(buf); n++)
+            buf[n] = it++;
+
+        secp256k1_multiset_create_roller(ctx, &x, buf, sizeof(buf));
+        secp256k1_multiset_add_roller(ctx, &roller, &x);
+    }
+
+    secp256k1_multiset_finalize_roller(ctx, result, &roller);
+}
+
+void bench_multiset_setup(void* arg) {
+    UNUSED(arg);
+}
+
+int main(void) {
+
+    ctx = secp256k1_context_create(SECP256K1_CONTEXT_VERIFY);
+
+    run_benchmark("ecdsa_multiset", bench_multiset, bench_multiset_setup, NULL, NULL, 10, 20);
+
+    secp256k1_context_destroy(ctx);
+    return 0;
+}
--- a/Makefile.am
+++ b/Makefile.am
@@ -175,3 +175,8 @@ endif
 if ENABLE_MODULE_RECOVERY
 include src/modules/recovery/Makefile.am.include
 endif
+
+if ENABLE_MODULE_MULTISET
+include src/modules/multiset/Makefile.am.include
+endif
+
--- a/configure.ac
+++ b/configure.ac
@@ -129,6 +129,11 @@ AC_ARG_ENABLE(module_ecdh,
     [enable_module_ecdh=$enableval],
     [enable_module_ecdh=no])
 
+AC_ARG_ENABLE(module_multiset,
+    AS_HELP_STRING([--enable-module-multiset],[enable multiset operations (experimental)]),
+    [enable_module_multiset=$enableval],
+    [enable_module_multiset=no])
+
 AC_ARG_ENABLE(module_recovery,
     AS_HELP_STRING([--enable-module-recovery],[enable ECDSA pubkey recovery module (default is no)]),
     [enable_module_recovery=$enableval],
@@ -431,6 +436,10 @@ if test x"$enable_module_ecdh" = x"yes"; then
   AC_DEFINE(ENABLE_MODULE_ECDH, 1, [Define this symbol to enable the ECDH module])
 fi
 
+if test x"$enable_module_multiset" = x"yes"; then
+  AC_DEFINE(ENABLE_MODULE_MULTISET, 1, [Define this symbol to enable the multiset module])
+fi
+
 if test x"$enable_module_recovery" = x"yes"; then
   AC_DEFINE(ENABLE_MODULE_RECOVERY, 1, [Define this symbol to enable the ECDSA pubkey recovery module])
 fi
@@ -480,6 +489,7 @@ AM_CONDITIONAL([USE_EXHAUSTIVE_TESTS], [test x"$use_exhaustive_tests" != x"no"])
 AM_CONDITIONAL([USE_BENCHMARK], [test x"$use_benchmark" = x"yes"])
 AM_CONDITIONAL([USE_ECMULT_STATIC_PRECOMPUTATION], [test x"$set_precomp" = x"yes"])
 AM_CONDITIONAL([ENABLE_MODULE_ECDH], [test x"$enable_module_ecdh" = x"yes"])
+AM_CONDITIONAL([ENABLE_MODULE_MULTISET], [test x"$enable_module_multiset" = x"yes"])
 AM_CONDITIONAL([ENABLE_MODULE_RECOVERY], [test x"$enable_module_recovery" = x"yes"])
 AM_CONDITIONAL([USE_JNI], [test x"$use_jni" == x"yes"])
 AM_CONDITIONAL([USE_EXTERNAL_ASM], [test x"$use_external_asm" = x"yes"])
--- a/include/secp256k1_multiset.h
+++ b/include/secp256k1_multiset.h
@@ -1,0 +1,99 @@
+#ifndef _SECP256K1_HOMOMORPHIC_
+# define _SECP256K1_HOMOMORPHIC_
+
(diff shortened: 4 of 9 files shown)


apoelstra:
1. Can you change the copyright headers on all the new files to use your name(s) rather than those on the files you templated from?
2. Can you write a markdown file like [this one](https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/src/modules/whitelist/whitelist.md) or [this one](https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/src/modules/surjection/surjection.md) that explains what you're doing and a bit of motivation?
3. I have no idea what our policy on accepting upstreamed modules is.

tomasvdw:
1 & 2 will do.

Note that this implements what Pieter Wiulle proposed on the Bitcoin mailing list. It isn't really upstream ATM.

apoelstra:
Ah, that's right, I knew this looked familiar. Even a .md in `modules/multiset` that was nothing but a link to the mailing list post would be great.

tomasvdw:
I've changed the attribution and added a [readme](https://github.com/tomasvdw/secp256k1/blob/993da60e45ca754ff3e279b001c95610851d48a9/src/modules/multiset/README.md).

I understand if this isn't merged before a BIP is embraced, although you could argue that optional modules can be experimental.

Notes for review:
* I believe I may be too eagerly normalizing field elements in secp256k1_multiset_create(). Someone with more knowledge of the normalization behaviour of the field operations may want to optimize some away.
* I am not using context, but added it to the API calls for consistency. I am not sure whether this is preferred.

gmaxwell:
> * We use trial-and-increment which is fast but non-constant time.

I believe that this is slower than shallue van de woestijne at least in batches when a fast legendre symbol is available. It also results in non-unform points, which probably would kill hope of creating a strong security proof (though without creating a pratical weakness).

> Currently benches at about 800k/sec

In Bitcoin I am a little doubtful this approach is acceptable because it would add  CPU hours to synchronization, multiplicative ( https://github.com/bitcoin/bitcoin/pull/10434 ) has much better performance and a more robust security argument. But is less cachable so it has worse performance at the tip.  It's a difficult trade-off.

>  Inits the multiset with the constant for empty data  

This initialization is non-portable; data encoded in a ge_storage_t (or gej) is potentially platform and or compile option specific.  What is that value in any case?  Looks like a backdoor as is, Since if there is any possible collection of set inputs that sum to that value which can be computed, then the initial state is prefilled with that entry as a member; so the initial state should be a value which is unreachable by any known input. The norm in the codebase is to include sage code to generate and verify all non-trivial constants. Please explain precisely where this constant originates.

I'm somewhat dubious about the point at infinity handling.  I guess the idea behind the unexplained 'random' constant is to make it unreachable?

> +    secp256k1_ge_set_gej(&ge_result, &gej_result);

Reprojecting to affine is devastating for performance and is entirely unneeded with the API setup here.

sipa:
> > We use trial-and-increment which is fast but non-constant time.
> I believe that this is slower than shallue van de woestijne at least in batches when a fast legendre symbol is available.

I think that's only true for characteristic-2 curves. Constant-time Shallue-Van de Woestijne here needs 1 inverse and 3 square roots. Even variable-time SVdW would need 1 inverse, 1.5 Jacobi symbols, and 1 square root.

The increment-and-test approach is indeed biased, and I'd prefer to avoid it, but using a hash operation rather than an increment would be fine (variable time doesn't matter here), and should be pretty fast when a fast Jacobi symbol is available (almost 2x the speed of the current PR, as it relies on using sqrt to test quadratic residues).

tomasvdw:
Thanks for the input! I've created an improved version (~0.76us/hash) based on these comments.

> In Bitcoin I am a little doubtful this approach is acceptable because it would add CPU hours to synchronization, multiplicative ( bitcoin/bitcoin#10434 ) has much better performance and a more robust security argument.

I am not sure where these CPU hours come from. With 800,000/sec (now improved to 1.3million/sec) it should take less than a minute for the current UTXO-set? I think this is hard to beat with multiplicative hashing. Also, I've written this as I am experimenting with hybrid rolling hash/prefix tree approaches for which maintaining 768 byte per rolling hash is not an option. I believe such approaches cannot only aid in UTXO proofs, but also simplify chunked syncing.

> This initialization is non-portable; data encoded in a ge_storage_t (or gej) is potentially platform and or compile option specific. What is that value in any case? Looks like a backdoor as is, Since if there is any possible collection of set inputs that sum to that value which can be computed, then the initial state is prefilled with that entry as a member; so the initial state should be a value which is unreachable by any known input. The norm in the codebase is to include sage code to generate and verify all non-trivial constants. Please explain precisely where this constant originates.

Good point. Using ge_storage_t was a bit hacky. 

The idea of the constant  was indeed to avoid infinite, but it was actually rather silly, as you lose empty+empty=empty and more important derived properties. I've changed this to a (custom) portable multiset format which is infinite aware, and properly using infinite as empty set.

> Reprojecting to affine is devastating for performance and is entirely unneeded with the API setup here.

Good point. My reasoning was to ensure a multiset fits in 64 bytes, but the costs of that are indeed devastating  (~25%). I've changed this to use a jacobian as multiset (96 bytes). We could always add a "multiset_compact" if we need a 64 byte version.

> but using a hash operation rather than an increment would be fine 

That sounds better indeed. I've changed it from trial-and-increment to trial-and-rehash.

tomasvdw:
Thanks for the input! I've made some improvements.

> In Bitcoin I am a little doubtful this approach is acceptable because it would add CPU hours to synchronization, multiplicative ( bitcoin/bitcoin#10434 ) has much better performance and a more robust security argument. But is less cachable so it has worse performance at the tip. It's a difficult trade-off.

It is currently running at ~14us/hash, which won't add CPU hours, but only a few minutes. Also, I am writing this as I am experimenting with hybrid prefix tree/rolling hash approaches for which 768 bytes/rolling hash isn't really an option.

 > This initialization is non-portable; data encoded in a ge_storage_t (or gej) is potentially platform and or compile option specific. What is that value in any case? Looks like a backdoor as is, Since if there is any possible collection of set inputs that sum to that value which can be computed, then the initial state is prefilled with that entry as a member; so the initial state should be a value which is unreachable by any known input. The norm in the codebase is to include sage code to generate and verify all non-trivial constants. Please explain precisely where this constant originates.

Good point. Using ge_storage_t as multiset was a bit hacky.

I was indeed using the constant to avoid infinite, but this was a rather silly idea. This breaks empty+empty=empty and more important derived properties. I now switched to a portable multiset type which handles infinity and properly using infinity as empty set.

> Reprojecting to affine is devastating for performance and is entirely unneeded with the API setup here.

Good point. My reasoning was, to use a 64-byte multiset, but the performance effect is indeed devastating (~25%). I now switched to a 96-byte (jacobian) multiset to avoid this. We could always add a 64 byte multiset_compact format if needed.

> using a hash operation rather than an increment would be fine

Using a hash seems better indeed. I've switched to trial-and-hash instead of trial-and-increment.

sipa:
> > because it would add CPU hours to synchronization

> It is currently running at ~14us/hash, which won't add CPU hours, but only a few minutes.

At 14us per operation, synchronization from scratch would gain 3.5 hours of CPU time (there are about 900M inputs+outputs in the chain).

You could argue that in a post-UTXO-sync world that isn't relevant, but we're not there, and I don't think we should disregard the cost for doing a full validation.

tomasvdw:
> You could argue that in a post-UTXO-sync world that isn't relevant, but we're not there, and I don't think we should disregard the cost for doing a full validation.

But even for full validation, why would you hash spend TXOs? Surely, you only need to hash the current UTXO set, at ~50 million or ~12 min (with a single thread). And only once.

> I think less than 14us/op is doable though; I'll have a look at the code soon.

I would very much appreciate that.

sipa:
> But even for full validation, why would you hash spend TXOs? Surely, you only need to hash the current UTXO set, at ~50 million or ~12 min (with a single thread).

Perhaps you're thinking of a different use case than I am. My hope is that every full node can at every point in time know what the hash of their UTXO set is. It could be used to compare the state of nodes, do sanity checking, or as part of a future UTXO commitment in blocks.

All of those things are not possible if you don't know the UTXO set hash during (even initial) synchronization.

tomasvdw:
I have the similar use cases in mind. In the future adding UTXO commitment to blocks, fast syncing, and possibly (with a hybrid prefix tree/rolling hash approach) UTXO proofs.

But I am afraid I do not understand the benefit of processing each historic UTXO change that way. What is the difference between the initial sync implementation:

A. Download all blocks, and for each block update the UTXO set and the commitment.
B. Download all blocks, and for each block update the UTXO set. Then update the commitment from the UTXO set.

I thought the primary advantage of a commutative hash is that A and B yield the same result. So what would be the drawback of B? Regardless of further optimizations, A seems wasteful.

sipa:
If we'd have UTXO commitments (as a consensus rule) in blocks, you wouldn't be able to validate the correctness of those commitments in the historical chain.

tomasvdw:
Agreed. I don't think the 900M existing outputs are relevant in this respect (as they don't have a commitment), but certainly the per block update time does for future full validation.

src/bench_multiset.c
```diff
@@ -0,0 +1,52 @@
...
+
+void bench_multiset_setup(void* arg) {
+    UNUSED(arg);
+}
+
+int main(void) {
+
+    ctx = secp256k1_context_create(SECP256K1_CONTEXT_VERIFY);
+
+    run_benchmark("ecdsa_multiset", bench_multiset, bench_multiset_setup, NULL, NULL, 10, 20);
```
jonasnick: This will report wrong results. The last parameter is supposed to acount for the number of iterations in the loop above. So it should be 10,000 but  still it won't properly account for `multiset_finalize`.

My suggestion is remove `multiset_finalize` above and change the line to
```
run_benchmark("ecdsa_multiset_add", bench_multiset, bench_multiset_setup, NULL, NULL, 10, 100000);
```
which is `12.5us` on my machine whereas minimum should be should be about `6us`: two hashes (=0.46us), three mul and sqrt (5.2us) and `gej_add_ge_var` (0.228us).

elichai:
For future reference, I also think this lacks a multiset serialization/parse methods for persistent storage (unless the author expected people to write the 96 byte struct to persistent storage, which would be really inconsistent with this library's API)

real-or-random:
Our parent project Bitcoin Core has MuHash3072 now (https://doxygen.bitcoincore.org/class_mu_hash3072.html). I feel that obsoletes this PR. (But please don't hesitate to disagree.)

PR #485, opened 2017-11-09
Remove dead store in secp256k1_ecmult_gen.

BitonicEelis:
This fixes the following clang-analyzer warning:

    src/ecmult_gen_impl.h:153:5: warning: Value stored to 'bits' is never read


sipa:
utACK 0279786adc8132874a2e62e3026584248090f47e

gmaxwell:
These assignments of variables holding secret data should be replaced with explicit zeroization, not just removed. Existing code is a placeholder for that.

practicalswift:
utACK 0279786adc8132874a2e62e3026584248090f47e

sipa:
Agree with @gmaxwell; we should instead replace it with explicit wiping.

BitonicEelis:
Roger that, closing this PR then. Thanks for the review!

PR #473, opened 2017-08-17
Multi-point multiplication support

sipa:
This uses the algorithm (Strauss with wNAF) from #464 to implement multi-multiplication, together with the API, tests, and scratch space support from #461.

PR #486, opened 2017-11-13
Add pippenger_wnaf for multi-multiplication

jonasnick:
This PR is based on #473 and adds a variant of "Pippengers algorithm" (see [Bernstein et al., Faster batch forgery identification](https://eprint.iacr.org/2012/549.pdf), page 15 and https://github.com/scipr-lab/libff/pull/10) for point multi-multiplication that performs better with a large number of points than Strauss' algorithm.

aggsig

Thanks to @sipa for providing `wnaf_fixed`, benchmarking, and the crucial suggestion to use affine addition.

The PR also makes `ecmult_multi` decide which algorithm to use, based on the number of points and the available scratch space.
For restricted scratch spaces this can be further optimized in the future (f.e. a 35kB scratch space allows batches of 11 points with strauss or 95 points with pippenger; choosing pippenger would be 5% faster).

As soon as this PR has received some feedback I'll repeat the benchmarks to determine the optimal `pippenger_bucket_window` with the new benchmarking code in #473.


apoelstra:
Hooray!

Please rebase -- should be straightforward, we merged only some minor things last week.

jonasnick:
Rebased

src/ecmult_impl.h
```diff
@@ -403,4 +477,538 @@ static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej
...
+    } else if(bucket_window == 8) {
+        return 2560;
+    } else if(bucket_window == 9) {
+        return 9200;
+    } else if(bucket_window == 10) {
+        return 17400;
+    } else if(bucket_window == 11) {
+        return 28600;
+    } else if(bucket_window == PIPPENGER_MAX_BUCKET_WINDOW) {
+        return INT_MAX;
```
sipa: INT_MAX is not defined.

apoelstra: `#include <limits.h>`  ?

apoelstra:
In `secp256k1_ecmult_multi_split_strauss_wnaf` we immediately convert a `secp256k1_ge` received from the callback to a `secp256k1_gej`. Would be better to change the callback API to just return a gej in the first place, otherwise the caller may be forced to do an expensive and pointless gej->ge conversion.

sipa:
@apoelstra Yes, I believe it was a conscious choice to making the interface pass affine points because no callers need more.

include/secp256k1.h
```diff
@@ -243,6 +256,28 @@ SECP256K1_API void secp256k1_context_set_error_callback(
     const void* data
 ) SECP256K1_ARG_NONNULL(1);
 
+/** Create a secp256k1 scratch space object.
+ *
+ *  Returns: a newly created scratch space.
+ *  Args: ctx:  an existing context object (cannot be NULL)
+ *  In:  init_size: initial amount of memory to allocate
+ *        max_size: maximum amount of memory to allocat
```
sipa: Nit: allocat

src/scratch_impl.h
```diff
@@ -0,0 +1,69 @@
...
+ * Distributed under the MIT software license, see the accompanying   *
+ * file COPYING or http://www.opensource.org/licenses/mit-license.php.*
+ **********************************************************************/
+
+#ifndef _SECP256K1_SCRATCH_IMPL_H_
+#define _SECP256K1_SCRATCH_IMPL_H_
+
+#include "scratch.h"
+
+#define ALIGNMENT 16
```
sipa: Add a comment to explain that we assume that common architectures never have alignment requirements above 8 for any of the types we care about (and make it 16 because who cares about a few byes)?

src/ecmult_impl.h
```diff
@@ -403,4 +457,94 @@ static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej
...
+    size_t max_points = secp256k1_scratch_max_allocation(scratch, 6) / point_size;
+    size_t n_batches, points_per_batch;
+    struct secp256k1_strauss_state state;
+
+    if (max_points == 0) return 0;
+    if (max_points > 160) max_points = 160; /* At this point, gains are not longer compensating for locality degradation */
+    n_batches = (n + max_points - 1) / max_points;
+    points_per_batch = (n + n_batches - 1) / n_batches;
+
+    /* Attempt to allocate sufficient space for Strauss */
```
sipa: @andytoshi Is this halving loop still necessary? The max allocation call above should be perfect, so I think we can just assert that this never fails.

src/ecmult_impl.h
```diff
@@ -403,4 +484,540 @@ static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej
     }
 }
 
+static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej *r, const secp256k1_gej *a, const secp256k1_scalar *na, const secp256k1_scalar *ng) {
```
sipa: At some point this should be renamed to `_var` as well, but let's leave that for later.

src/ecmult_impl.h
```diff
@@ -403,4 +484,540 @@ static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej
...
+}
+
+static size_t secp256k1_strauss_max_points(secp256k1_scratch *scratch) {
+    return secp256k1_scratch_max_allocation(scratch, STRAUSS_SCRATCH_OBJECTS) / secp256k1_strauss_scratch_size(1);
+}
+
+/** Convert a number to WNAF notation.
+ *  The number becomes represented by sum(2^{wi} * wnaf[i], i=0..WNAF_SIZE(w)+1) - return_val.
+ *  It has the following guarantees:
+ *  - each wnaf[i] an odd integer between -(1 << w) and (1 << w)
```
sipa: These two lines are more accurately and correctly written as `each wnaf[i] is either zero or an odd integer between -(1 << w) and (1 << w)`.

src/ecmult_impl.h
```diff
@@ -403,4 +484,540 @@ static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej
...
+    return 1;
+}
+
+/**
+ * Returns optimal bucket_window (number of bits of a scalar represented by a
+ * set of buckets) for a given number of points.
+ */
+static int secp256k1_pippenger_bucket_window(size_t n) {
+#ifdef USE_ENDOMORPHISM
+    if (n <= 4) {
```
sipa: Maybe write this in a tree structure rather than linear (as in: top level is is a bisection for first half/second half) etc.

src/ecmult_impl.h
```diff
@@ -403,4 +484,540 @@ static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej
...
+        return 11;
+    } else {
+        return PIPPENGER_MAX_BUCKET_WINDOW;
+    }
+#endif
+}
+
+static size_t secp256k1_pippenger_bucket_window_inv(int bucket_window) {
+#ifdef USE_ENDOMORPHISM
+    if(bucket_window == 1) {
```
sipa: Nit: space after `if`.

src/ecmult_impl.h
```diff
@@ -403,4 +484,540 @@ static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej
...
+                idx = (n - 1)/2;
+                secp256k1_gej_add_ge_var(&buckets[idx], &buckets[idx], &pt[point_state.input_pos], NULL);
+            } else if (n < 0) {
+                idx = -(n + 1)/2;
+                secp256k1_ge_neg(&tmp, &pt[point_state.input_pos]);
+                secp256k1_gej_add_ge_var(&buckets[idx], &buckets[idx], &tmp, NULL);
+            }
+        }
+        secp256k1_gej_set_infinity(&running_sum);
+        secp256k1_gej_set_infinity(&walking_sum);
```
sipa: Do you need a separate variable for `walking_sum`? Can't you reuse `r` for this?

src/ecmult_impl.h
```diff
@@ -403,4 +484,540 @@ static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej
...
+        space_for_points = max_alloc - ((1<<bucket_window) * sizeof(secp256k1_gej) + entry_size + sizeof(struct secp256k1_pippenger_state));
+        if (space_for_points < 0) {
+            break;
+        }
+        n_points = space_for_points/entry_size;
+        n_points = n_points > max_points ? max_points : n_points;
+        if (n_points > res) {
+            res = n_points;
+        }
+        if(n_points < max_points) {
```
sipa: Nit: space after `if`.

src/ecmult_impl.h
```diff
@@ -403,4 +484,540 @@ static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej
...
+    max_points = secp256k1_pippenger_max_points(scratch);
+    if (max_points == 0) {
+        return 0;
+    } else if (max_points > ECMULT_MAX_POINTS_PER_BATCH) {
+        max_points = ECMULT_MAX_POINTS_PER_BATCH;
+    }
+    n_batches = (n+max_points-1)/max_points;
+    n_batch_points = (n+n_batches-1)/n_batches;
+
+    if(n_batch_points >= ECMULT_PIPPENGER_THRESHOLD) {
```
sipa: Nit: space after `if`.

src/ecmult_impl.h
```diff
@@ -403,4 +484,540 @@ static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej
...
+        }
+        n_batches = (n+max_points-1)/max_points;
+        n_batch_points = (n+n_batches-1)/n_batches;
+        f = &secp256k1_ecmult_strauss_batch;
+    }
+    for(i = 0; i < n_batches; i++) {
+        size_t nbp = n < n_batch_points ? n : n_batch_points;
+        size_t offset = n_batch_points*i;
+        secp256k1_gej tmp;
+        if(!(*f)(ctx, scratch, error_callback, &tmp, i == 0 ? inp_g_sc : NULL, cb, cbdata, nbp, offset)) {
```
sipa: Nit: you can just use `f(...)` instead of `(*f)(...)`.

src/ecmult_impl.h
```diff
@@ -403,4 +484,540 @@ static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej
...
+    if(n_batch_points >= ECMULT_PIPPENGER_THRESHOLD) {
+        f = &secp256k1_ecmult_pippenger_batch;
+    } else {
+        max_points = secp256k1_strauss_max_points(scratch);
+        if (max_points == 0) {
+            return 0;
+        }
+        n_batches = (n+max_points-1)/max_points;
+        n_batch_points = (n+n_batches-1)/n_batches;
+        f = &secp256k1_ecmult_strauss_batch;
```
sipa: Nit: ampersand is optional here.

src/ecmult_impl.h
sipa: I think this can be written more efficiently using a switch case.

src/scratch.h
```diff
@@ -0,0 +1,35 @@
...
+ * Distributed under the MIT software license, see the accompanying   *
+ * file COPYING or http://www.opensource.org/licenses/mit-license.php.*
+ **********************************************************************/
+
+#ifndef _SECP256K1_SCRATCH_
+#define _SECP256K1_SCRATCH_
+#include <sys/types.h>
+
+/* The typedef is used internally; the struct name is used in the public API
+ * (where it is exposed as a different typedef) */
```
sipa: The typedef and struct name are the same publicly; the struct is just not defined there.

src/scratch_impl.h
sipa: Perhaps add a TODO that we should try to determine this at configure time (using sizeof(max_align_t), if available, for example).

src/ecmult_impl.h
jonasnick: Specifically, do you suggest to replace
```
 *  - each wnaf[i] is an odd integer between -(1 << w) and (1 << w)
 *  - each wnaf[i] is nonzero, unless the given number is 0
```
with 
```
* wnaf[i] is either zero or an odd integer between -(1 << w) and (1 << w)
```
?
If so, that sounds just as accurate but less helpful.

src/ecmult_impl.h
jonasnick: I don't see how. If `r` stores the result of previous iterations it can't be used in the addition ladder.

src/ecmult_impl.h
jonasnick: What's the goal of this?

src/scratch.h
apoelstra: I'm not sure what you mean, the typedef here is `secp256k1_scratch`, the public one is `secp256k1_scratch_space`.

src/ecmult_impl.h
apoelstra: To be clear: is it possible for a nonzero number to have 0 wnaf digits in our representation? I believe the answer is yes, though for wnaf_const (for example) it is not, so this is an important distinction.

src/ecmult_impl.h
sipa: Only having a logarithmic number of conditionals on average.

src/ecmult_impl.h
sipa: @jonasnick The first statement "each wnaf[i] is an odd integer between -(1 << w) and (1 << w)" is just incorrect. 0 is not an odd integer, and 0 is possible in case the input is 0.

src/scratch.h
sipa: Oops, ignore me!

src/ecmult_impl.h
jonasnick: Oh, of course 0 is even. But I believe the second statement should stay because at least empirically it doesn't happen that a nonzero number has 0 wnaf digits (we test for this, copied from wnaf_const). @sipa can you confirm?

sipa: I don't care that it doesn't happen often. The first sentence is stated as a condition which is always true, and it is not. You could say "There result satisfies one of the two following conditions:" and leave the two in. But leaving them as-is is just wrong.

src/ecmult_impl.h
jonasnick: While being better in the average case I think this is worse for small batches (where this code has more impact compared to larger batches) and looks ugly. Maybe that's something for another PR.

jonasnick:
Pushed fixes

peterdettman:
@jonasnick I opened a PR here: https://github.com/jonasnick/secp256k1/pull/1 for suggested changes to save some additions in each _pippenger_wnaf window.

jonasnick:
As usual, @peterdettman delivers. I see about 2% to 4% speedup. Merged the commit into this PR.

src/ecmult_impl.h
sipa: Fair enough.

jonasnick:
Squashed fixes and updated bucket windows. Previous bucket windows did not not include peterdettman's optimization, did not use bench_ecmult and had a bug where the number of points for a bucket window were twice of what they should have been. The result is that the performance without endomorphism is very similar, with endomorphism it's up to 5% faster for some number of points. The performance graph looks very similar to the one above.

sipa:
ACK 50f0779be9c8d67fc5c8ca3a7a21b17da74cfa33

src/scratch.h
```diff
@@ -0,0 +1,35 @@
+/**********************************************************************
+ * Copyright (c) 2017 Andrew Poelstra	                              *
+ * Distributed under the MIT software license, see the accompanying   *
+ * file COPYING or http://www.opensource.org/licenses/mit-license.php.*
+ **********************************************************************/
+
+#ifndef _SECP256K1_SCRATCH_
+#define _SECP256K1_SCRATCH_
+#include <sys/types.h>
```
apoelstra: We can remove this #include. There are two places that `ssize_t` are used in `ecmult_impl.h` which need to be removed. The first, `ssize_t max_alloc =` can be changed to `size_t` because it is assigned a `size_t` anyway (we changed `secp256k1_scratch_max_allocation` to not return a sign value).

The second use is `ssize_t space_for_points`. This can also be made into a `size_t`. The check on line 938, `if (space_for_points < 0)`, will need to be rearranged to compare the two things that are subtracted on line 937 to get `space_for_points`.

src/scratch.h
```diff
@@ -0,0 +1,35 @@
...
+
+/* The typedef is used internally; the struct name is used in the public API
+ * (where it is exposed as a different typedef) */
+typedef struct secp256k1_scratch_space_struct {
+    void *data;
+    size_t offset;
+    size_t init_size;
+    size_t max_size;
```
apoelstra: Can we add an `error_callback` pointer to this structure and remove it from the arguments of `secp256k1_scratch_resize`? It would avoid the need to pass error callbacks into a bunch of higher level functions.

src/ecmult.h
```diff
@@ -28,4 +30,9 @@ static int secp256k1_ecmult_context_is_built(const secp256k1_ecmult_context *ctx
 /** Double multiply: R = na*A + ng*G */
 static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej *r, const secp256k1_gej *a, const secp256k1_scalar *na, const secp256k1_scalar *ng);
 
+typedef int (secp256k1_ecmult_multi_callback)(secp256k1_scalar *sc, secp256k1_ge *pt, size_t idx, void *data);
+
+/** Multi-multiply: R = inp_g_sc * G + sum_i ni * Ai. */
```
apoelstra: Update comment to say that this function will clobber the scratch space that gets passed in? I've nearly been burned by that in the Bulletproofs code a couple times.

src/ecmult_impl.h
apoelstra: Agreed, please change this to a `CHECK` on the return value of `secp256k1_scratch_resize`.

src/tests.c
```diff
@@ -2522,6 +2522,243 @@ void run_ecmult_const_tests(void) {
...
+        CHECK(secp256k1_gej_is_infinity(&r));
+    }
+
+    /* Sanity check that zero scalars don't cause problems */
+    memset(&sc[0], 0, sizeof(sc[0]));
+    CHECK(secp256k1_ecmult_multi(&ctx->ecmult_ctx, scratch, &ctx->error_callback, &r, &szero, ecmult_multi_callback, &data, 20));
+    memset(&sc[1], 0, sizeof(sc[0]));
+    memset(&sc[2], 0, sizeof(sc[0]));
+    memset(&sc[3], 0, sizeof(sc[0]));
+    memset(&sc[4], 0, sizeof(sc[0]));
```
apoelstra: Can we use `secp256k1_scalar_clear` instead of memsetting?

jonasnick:
Pushed fixes

src/ecmult_impl.h
apoelstra: Oh, derp, Jonas points out that this is fixed in a later commit. Never mind.

jonasnick: This is outdated by "Add pippenger_wnaf ecmult_multi". If resize fails then return 0.

src/ecmult_impl.h
```diff
@@ -477,74 +495,414 @@ static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej
...
+    return 1;
+}
+
+/* Wrapper for secp256k1_ecmult_multi_func interface */
+static int secp256k1_ecmult_pippenger_batch_single(const secp256k1_ecmult_context *actx, secp256k1_scratch *scratch, const secp256k1_callback* error_callback, secp256k1_gej *r, const secp256k1_scalar *inp_g_sc, secp256k1_ecmult_multi_callback cb, void *cbdata, size_t n) {
+    return secp256k1_ecmult_pippenger_batch(actx, scratch, error_callback, r, inp_g_sc, cb, cbdata, n, 0);
+}
+
+#define MAX_BATCH_SIZE 1024
+typedef int (*secp256k1_ecmult_multi_func)(const secp256k1_ecmult_context*, secp256k1_scratch*, const secp256k1_callback*, secp256k1_gej*, const secp256k1_scalar*, secp256k1_ecmult_multi_callback cb, void*, size_t);
```
apoelstra: I'd like this definition and the two wrapper functions to be moved to `src/ecmult.h` rather than the `_impl` file. @sipa do you have thoughts on that? These are used (only) in bench_ecmult.c and tests.c.

apoelstra:
ACK https://github.com/bitcoin-core/secp256k1/pull/486/commits/5979a573829b166f05ea8e03747bad9225451919 .. please squash. The only nit is the position of the `secp256k1_ecmult_multi_func` typedef, which I don't care about _that_ much.

jonasnick:
Squashed

PR #490, opened 2017-12-19
Disambiguate bench functions and types

sipa:
This fixes a Travis failure.


etienneroudeix:
👍 
It also fix `make` issues when configuring with `./configure --enable-experimental --enable-module-ecdh --enable-module-recovery`

PR #491, opened 2017-12-19
Fixing a build failure (see below)

pirapira:
With gcc (Ubuntu 7.2.0-8ubuntu3) 7.2.0:
```
src/bench_recover.c:18:6: error: ‘bench_recover’ redeclared as different kind of symbol
 void bench_recover(void* arg) {
      ^~~~~~~~~~~~~
src/bench_recover.c:16:3: note: previous declaration of ‘bench_recover’ was here
 } bench_recover;
   ^~~~~~~~~~~~~
```


apoelstra:
Dupe of #490. Thanks for the quick catch. You can workaround for now by adding `--enable-benchmark=no`.

PR #494, opened 2017-12-28
Support OpenSSL versions >= 1.1 for ENABLE_OPENSSL_TESTS

codablock:
The only reason OpenSSL 1.1 was not supported was the removal of direct
access to r and s in ECDSA_SIG. This commit adds a simplified version of
ECDSA_SIG_get0 for < 1.1 that can be used like ECDSA_SIG_get0 in >= 1.1


gmaxwell:
Concept ACK!  Thanks!

PR #499, opened 2018-01-12
tests: Make sure we get the requested number of bytes from /dev/urandom

practicalswift:
Make sure we get the requested number of bytes from `/dev/urandom`.


gmaxwell:
ACK, we might want to add a warning in the case urandom isn't used, but I'll make a separate PR.

PR #504, opened 2018-01-29
Update URLs to MIT license.

### LINKED ISSUE #503: Update links to MIT license (use https)
In line of bitcoin/bitcoin#12190 and bitcoin/bitcoin#12192 the URLs to the MIT license should be updated to use https protocol. I will draft a PR.

fwolfst:
This closes #503 (where further discussion is linked).
Command(s) used to create this changeset was/were:

'''
find . -type f -print0 | xargs -0 sed -i -e
's,http://www.opensource.org/licenses/mit-license.php\.\*,https://opensource.org/licenses/mit-license.php\.
\*,g'
'''

### DIFF
--- a/contrib/lax_der_parsing.c
+++ b/contrib/lax_der_parsing.c
@@ -1,7 +1,7 @@
 /**********************************************************************
  * Copyright (c) 2015 Pieter Wuille                                   *
  * Distributed under the MIT software license, see the accompanying   *
- * file COPYING or http://www.opensource.org/licenses/mit-license.php.*
+ * file COPYING or https://opensource.org/licenses/mit-license.php.   *
  **********************************************************************/
 
 #include <string.h>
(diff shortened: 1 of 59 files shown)


gmaxwell:
Might it not just be better to eliminate the URL from the header.  I don't see what value it gains, and in 5 years that domain might well be a porn site. :)

fwolfst:
Quite unlikely that the domain will be used for that but sure its possible
(some goes for github).

I agree that the added benefit is not small, but from that page you can
access "Further resources on MIT License" (top right) and it motivates to
look around what other Open Source licenses and ideas are there.

Also it shows the motivation to let you know your rights when you find the
source file in the wild, without an accompanying COPYING file (or one that
contains a different license). Not that this happens to me on a daily basis
...

gmaxwell:
Sorry, thanks for your attention, but I don't think this change is worth taking at this time.

PR #505, opened 2018-01-29
WIP: Add aggsig example code

jonasnick:
Based on https://github.com/bitcoin-core/secp256k1/pull/461 + some things I suggested for that PR.

I think it'd be a good idea to have some example code to show how to use the aggsig module. For example, the seed of the nonce RNG could be misused and it's not all that intuitive how a scratch space is used. This PR addresses this by adding an `aggsig.c` file  into an `examples/` directory and building it as `example_aggsig` (it's also run as part of `make check`).

Since https://github.com/bitcoin-core/secp256k1/pull/461 does not allow for multi party signing at the moment, `aggsig.c` demonstrates single user signing and verification for now.


examples/aggsig.c
```diff
@@ -0,0 +1,142 @@
...
+    /* The probability that this not a valid secret key is approximately 2^-128 */
+    } while (!secp256k1_ec_seckey_verify(ctx, seckey));
+
+    ret = secp256k1_ec_pubkey_create(ctx, pubkey, seckey);
+    assert(ret);
+    fclose(frand);
+    return 1;
+}
+
+/* Sign a message hash with the given key pairs and and store the result in sig */
```
kallewoof: `and and`

examples/aggsig.c
```diff
@@ -0,0 +1,142 @@
...
+    if (!secp256k1_aggsig_combine_signatures(ctx, aggctx, sig, partial_signatures)) {
+        return 0;
+    }
+
+    secp256k1_aggsig_context_destroy(aggctx);
+    return 1;
+}
+
+/* Verify an aggregated signature of n_pubkeys pubkeys over msghash */
+int verify(const secp256k1_context* ctx, const unsigned char *sig, const unsigned char *msghash, const secp256k1_pubkey *pubkeys, size_t n_pubkeys) {
```
kallewoof: You use `N_PUBKEYS` in other functions but here you add a `size_t n_pubkeys`. Particular reason or just ended up that way? (You only call this with `N_PUBKEYS` as argument I think.)

examples/aggsig.c
```diff
@@ -0,0 +1,142 @@
...
+    size_t scratch_size = secp256k1_aggsig_verify_scratch_size(n_pubkeys);
+    secp256k1_scratch_space* scratch;
+    /* The scratch space can be limited or set to a fixed size. If it's smaller
+     * than the optimum, then the verification algorithm will run slower. But
+     * the scratch space must fit at least one public key. */
+    if (scratch_size > 9000) {
+        scratch_size = 9000;
+    }
+    scratch = secp256k1_scratch_space_create(ctx, 0, scratch_size);
+    if(scratch == NULL) {
```
kallewoof: Nit: You're mixing `if()` and `if ()` style, and the same for `for()` and `for ()`. Personally prefer the latter, but probably shouldn't mix.

examples/aggsig.c
```diff
@@ -0,0 +1,142 @@
...
+    secp256k1_scratch_space_destroy(scratch);
+    return 1;
+}
+
+int main(void) {
+    secp256k1_context* ctx;
+    int i;
+    unsigned char seckeys[N_PUBKEYS][32];
+    secp256k1_pubkey pubkeys[N_PUBKEYS];
+    unsigned char msghash[32] = "this_should_actually_be_msg_hash";
```
kallewoof: This will actually overbuffer, writing a `\0` into the 33rd byte of `msghash`. Surprised compilers don't warn about that.

examples/aggsig.c
jonasnick: good catch, thanks!

examples/aggsig.c
jonasnick: Hm, http://port70.net/~nsz/c/c89/c89-draft.html#3.5.6 says terminating null characters are only added "if there is room".

kallewoof: That would explain the lack of compiler warnings! First thing I ever heard of such an arbitrary rule. C can be freaky.

jonasnick:
Closing as this will be probably superseded by MuSig

PR #507, opened 2018-02-09
Add secp256k1_ec_grind to grind keypairs

sipa:
Unsure if this should be in upstream, but it may be useful for things like shaving off a few bytes from signatures.


kallewoof:
Tested ACK 885ea63

This PR is used as is in https://github.com/kallewoof/btcdeb/tree/vanitygen -- I verified that the keypair is valid.

PR #443, opened 2017-02-21
Adding compressed public key flag to java api

Christewart:
Implements the ability to generate a compressed public key from the java API.

PR #508, opened 2018-02-21
Implement `decompress` function in JNI

pm47:
This function was already exported but not implemented.

Being able to efficiently decompress a public key is useful for implementing LN on JVM-based languages, mainly because [network announcements](https://github.com/lightningnetwork/lightning-rfc/blob/master/07-routing-gossip.md) use the 33-bytes representation.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,9 @@ addons:
 compiler:
   - clang
   - gcc
-cache:
-  directories:
-  - src/java/guava/
 env:
   global:
     - FIELD=auto  BIGNUM=auto  SCALAR=auto  ENDOMORPHISM=no  STATICPRECOMPUTATION=yes  ASM=no  BUILD=check  EXTRAFLAGS=  HOST=  ECDH=no  RECOVERY=no  EXPERIMENTAL=no  JNI=no
-    - GUAVA_URL=https://search.maven.org/remotecontent?filepath=com/google/guava/guava/18.0/guava-18.0.jar GUAVA_JAR=src/java/guava/guava-18.0.jar
   matrix:
     - SCALAR=32bit    RECOVERY=yes
     - SCALAR=32bit    FIELD=32bit       ECDH=yes  EXPERIMENTAL=yes
@@ -59,8 +55,6 @@ matrix:
           packages:
             - gcc-multilib
             - libgmp-dev:i386
-before_install: mkdir -p `dirname $GUAVA_JAR`
-install: if [ ! -f $GUAVA_JAR ]; then wget $GUAVA_URL -O $GUAVA_JAR; fi
 before_script: ./autogen.sh
 script:
  - if [ -n "$HOST" ]; then export USE_HOST="--host=$HOST"; fi
--- a/Makefile.am
+++ b/Makefile.am
@@ -121,8 +121,6 @@ endif
 
 JAVAROOT=src/java
 JAVAORG=org/bitcoin
-JAVA_GUAVA=$(srcdir)/$(JAVAROOT)/guava/guava-18.0.jar
-CLASSPATH_ENV=CLASSPATH=$(JAVA_GUAVA)
 JAVA_FILES= \
   $(JAVAROOT)/$(JAVAORG)/NativeSecp256k1.java \
   $(JAVAROOT)/$(JAVAORG)/NativeSecp256k1Test.java \
@@ -131,20 +129,15 @@ JAVA_FILES= \
 
 if USE_JNI
 
-$(JAVA_GUAVA):
-	@echo Guava is missing. Fetch it via: \
-	wget https://search.maven.org/remotecontent?filepath=com/google/guava/guava/18.0/guava-18.0.jar -O $(@)
-	@false
-
 .stamp-java: $(JAVA_FILES)
 	@echo   Compiling $^
-	$(AM_V_at)$(CLASSPATH_ENV) javac $^
+	$(AM_V_at) javac $^
 	@touch $@
 
 if USE_TESTS
 
-check-java: libsecp256k1.la $(JAVA_GUAVA) .stamp-java
-	$(AM_V_at)java -Djava.library.path="./:./src:./src/.libs:.libs/" -cp "$(JAVA_GUAVA):$(JAVAROOT)" $(JAVAORG)/NativeSecp256k1Test
+check-java: libsecp256k1.la .stamp-java
+	$(AM_V_at)java -Djava.library.path="./:./src:./src/.libs:.libs/" -cp "$(JAVAROOT)" $(JAVAORG)/NativeSecp256k1Test
 
 endif
 endif
--- a/src/java/org/bitcoin/NativeSecp256k1.java
+++ b/src/java/org/bitcoin/NativeSecp256k1.java
@@ -141,10 +141,10 @@ public class NativeSecp256k1 {
      * libsecp256k1 Compute Pubkey - computes public key from secret key
      *
      * @param seckey ECDSA Secret key, 32 bytes
+     * @param compressed Should the generated public key be compressed
      * @return pubkey ECDSA Public key, 33 or 65 bytes
      */
-    //TODO add a 'compressed' arg
-    public static byte[] computePubkey(byte[] seckey) throws AssertFailException{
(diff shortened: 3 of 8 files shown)


pm47:
Last 2 commits are trivial but not related to the topic, please let me know if I should put them in a separate PR.

Christewart:
@pm47 Does this supersede #443?

pm47:
@Christewart Only partly, it doesn't include the `fCompressed` flag.

Apart from that, attempting to decompress an invalid public key will result in an exception been thrown, so that's functionally equivalent to `isValidPubKey`. And we both removed the guava dependency ;-)

Also I somehow missed the file `NativeSecp256k1Test.java`, I'm going to add a test on `decompress`.

Christewart:
@pm47 Maybe you want to rip my code from #443 to have `isValidPubKey` that doesn't throw and then add a `fCompressed` flag? Then I'll close #443 and we can go with this one.

pm47:
@Christewart Sounds good! I'll work on that this weekend

pm47:
@Christewart done, I have taken your other changes and added more tests.

Guava dependency has been completely removed, the JNI part can be built and tested by simply doing:
```shell
$ ./autogen.sh
$ ./configure --enable-jni --enable-experimental --enable-module-ecdh
$ make
$ javac src/java/org/bitcoin/*
$ java -Djava.library.path=.libs -cp src/java org.bitcoin.NativeSecp256k1Test
```
If this gets approved, I'd also like to reformat java files before merging, they are quite messy currently.

Christewart:
Great, I'll test with bitcoin-s and make sure everything is working smoothly. Getting rid of the guava dependency is nice. Thanks for integrating! I'll close the other PR now

sipa:
@greenaddress Feel like reviewing this?

Christewart:
Confirming these changes work with bitcoin-s. They are implemented in 

https://github.com/bitcoin-s/bitcoin-s-core/pull/120

Christewart:
ping @greenaddress @sipa Thoughts on merging?

greenaddress:
@pm47 @Christewart 

nice changes!

sorry this took long.

(1) GUAVA_URL=https://search.maven.org/remotecontent?filepath=com/google/guava/guava/18.0/guava-18.0.jar GUAVA_JAR=src/java/guava/guava-18.0.jar

in .travis.yml

This should be removed since you removed [GUAVA](https://github.com/bitcoin-core/secp256k1/pull/508#issuecomment-372135347)

(2) AssertFailException should also be removed from 465 (randomize)

(3) isValidPubKey is missing some checks that are in its sister method decompress. Maybe would be better if isValidPubKey was implemented in terms of calling decompress and returning false on failure.

pm47:
Good catch re: guava in .travis.yml!

I'm not too sure about (3), but it is indeed simpler.

If that's ok with you I'd like to reformat before merging

floreslorca:
bump this. we are also interested in getting this merged @greenaddress

Christewart:
It's pretty ridiculous this hasn't been merged yet. @sipa @gmaxwell @greenaddress

gmaxwell:
Do not ping me. I no longer work on this project.

sipa:
Sorry for the slow answers here; there hasn't been that much activity in this project lately, and even less from people familiar with Java. I'll try to improve upon that, but it may be better if the JNI bindings move to a different project eventually, as it's hard to guarantee the same review standards as the rest of the project by the current maintainers. However, as long as that hasn't happened, it shouldn't stand in the way of merging useful improvements to the JNI code.

Overall, concept ACK to the code changes, but a few comments:

* This will need (probably trivial) rebase on top of #651. 
* This PR is doing a lot more than adding a JNI decompress function, perhaps it should be renamed.
* I can't run the tests; I get `src/java/org/bitcoin/NativeSecp256k1Test.java:206: cannot find symbol` ... `DataTypeConverter`. What am I missing? (given that it works in Travis, I probably am).

real-or-random:
@sipa The `DataTypeConverter` error should be gone after the rebase on #651; usage of this class has been removed therein.

ysangkok:
Just FYI, the DataTypeConverter problem is due to [JEP 320](http://openjdk.java.net/jeps/320#Java-EE-modules). I guess @sipa must be on OpenJDK 11+.

sstone:
We started this PR and eventually decided to fork secp256k1 and modified the JNI bindings quite a bit, see https://github.com/ACINQ/secp256k1/tree/jni-embed

We did this because we wanted some changes, including this one, which just impacted the JNI bindings (we did not change anything in the core library), but also because it made things a lot simpler.

For example, we wanted to use a proper java build/dependency management tool (yes, maven, I can hear you all scream from where I am :)), organise source files in a more standard way, package the native library into the jar which means changing how it is loaded, and publish the jar to maven central which is just impossible (for us) if we keep it under org.bitcoin...  

This is probably the reason why keeping JNI bindings here is a bit awkward: using java sources as they are now is clumsy because of how they are packaged (that's simple to fix though: move bindings to src/main/java and unit tests to src/test/java), and more importantly most JVM users will want something that integrates with their build tool, is deployment friendly which means embedding the native library into the JAR (this is what surebits and us have done independently for example) and available on maven central (or another widely used repo) and all that is really hard to do now.

There are also build issues: to cross-build you need JNI headers for the * target * OS which a standard JDK install won't give you, on Android you need to use a specific toolchain and cannot embed native libs into jars, ...

A first step towards this would be to re-organise java sources to match the maven convention (which is also used by sbt/gradle/...), which would easily let people add their own build files. But I think that the biggest issues are embedding native libs into the jar, and publishing the jar, is it something that secp256k1 maintainers would be willing to do ?

elichai:
@sstone Maybe you can start a dedicated jni bindings repository, and the support will be removed from the core library? (just like https://github.com/rust-bitcoin/rust-secp256k1 for rust)

apoelstra:
For my part I can say no, I am not willing or able to reorganize the Java sources. I don't know Java, don't know Maven, don't understand the org structure, and have never looked at the JNI parts of libsecp.

I'm curious if the story is different for any of the libsecp maintainers. If not, we should probably do what we can to separate the JNI bindings from the lib.

real-or-random:
(Not a maintainer but) I think separating the bindings is a good idea. There is no reason to believe that they will get more love in the future.

Who else are large users? Certainly bitcoinj but I don't know any other project. /cc @schildbach

schildbach:
I'm not aware of any projects using it. Theoretically the bindings are a nice thing to have, but in reality signing is an extremely rare operation.

schildbach:
Maybe ask on the bitcoinj mailing list?

real-or-random:
@schildbach 
Just for clarification: You're saying you're not aware of any projects besides bitcoinj? Or doesn't even bitcoinj use it? I'm asking because it seems to me that bitcoinj uses it:
https://github.com/bitcoinj/bitcoinj/blob/master/core/src/main/java/org/bitcoin/NativeSecp256k1.java

> Theoretically the bindings are a nice thing to have, but in reality signing is an extremely rare operation.

Yes, but we provide verification, too.

sstone:
@elichai @real-or-random users can have a look at https://github.com/ACINQ/secp256k1/tree/jni-embed/src/java, and I believe that @Christewart has a similar project. Of course if we're the only users of the JNI bindings I guess the problem is solved :)

schildbach:
@real-or-random Bitcoinj provides the bindings, but they're optional to use. By default, BouncyCastle is used for signing and verification.

sipa:
I think it's become clear that building and maintaining proper Java integration, including the requirements for build systems used in that community, will need work by people familiar with the language. I've personally not used Java since 2005, and I believe the other maintainers also don't have much experience.

If there is a minimal set of improvements we can make to the JNI bindings to keep them usable, that would be great. But if the result is going to involve other projects having their own packages/modification on top, I believe the best action is probably to remove it entirely. As the JNI bindings internally only use the public C API, there is no actual technical reason why it needs to be part of the same library anyway.

greenaddress:
I'm no longer using the Java secp JNI bindings so I am unlikely to do more development or testing around it - I am supportive of moving things out, that being said I reviewed the above changes, they seem fine but it would need some rebase and some squashing I think

real-or-random:
Closing this because we removed the JNI bindings in #682.

PR #513, opened 2018-03-01
Increase sparsity of pippenger fixed window naf representation

### LINKED ISSUE #506: Increase sparsity of pippenger fixed window naf representation
The current fixed window naf used in the pippenger never produces zeros in its representation, even though the code handles them fine and they're faster.   The existing representation can be converted into one where there are more zeros by setting a coefficient to zero if it is 1 or -1 and the proceeding digit is strictly negative or strictly positive respectively.  

For small windows (where pippenger is already slower than strauss) it's a pretty big speedup.  For realistic window sizes it probably isn't a big speedup, but free performance is good.

jonasnick:
Fixes #506

### DIFF
--- a/src/ecmult_impl.h
+++ b/src/ecmult_impl.h
@@ -571,9 +571,8 @@ static int secp256k1_wnaf_fixed(int *wnaf, const secp256k1_scalar *s, int w) {
     const secp256k1_scalar *work = s;
 
     if (secp256k1_scalar_is_zero(s)) {
-        while (pos * w < WNAF_BITS) {
+        for (pos = 0; pos < WNAF_SIZE(w); pos++) {
             wnaf[pos] = 0;
-            ++pos;
         }
         return 0;
     }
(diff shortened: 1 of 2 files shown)


jonasnick:
ecmult_multi pippenger speedup
```
without endo:    2 points 11%,     pippenger threshold (160 points) 0.74%
with endo:       2 points 10%,     pippenger threshold (88 points)  2.1%    
```

jonasnick:
Added wnaf_fixed bugfix and two commits that further increase sparsity. Now it's possible to use precomputation with pippenger similar to using the endomorphism: split a scalar, point pair `(s, G)` into `(s1, ..., sn, G, ..., Gn)` where all `si` have `256 -256/n` leading zeros and `s1*G + ... + sn*Gn = s*G`.

With 128 points and `n=8` this gives a speedup of 6% over endomorphism (13% over non-endo). See https://github.com/jonasnick/secp256k1/blob/bench_ecmult_precomp/src/bench_ecmult_precomp.c#L138 for the benchmarking and splitting code.

sipa:
Lightly tested ACK ec0a7b3ae3a09bd2ffbbeec83abe6c5042ce4a56

apoelstra:
Let me test it with the BP code before merging.

apoelstra:
ACK

PR #516, opened 2018-03-06
improvements to random seed in src/tests.c

hdon:
I've made two small changes to `src/tests.c` circa random seed generation.

Added a warning when `/dev/urandom` fails, mostly to defend against the case that someone should use the code verbatim, but also to enhance its illustrative power.

Also I fixed a bug with how the return value of `fread()` was being evaluated. In fact, `/dev/urandom` was never being applied before as the check on the return value of `fread()` always failed!

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -4919,8 +4919,8 @@ int main(int argc, char **argv) {
     } else {
         FILE *frand = fopen("/dev/urandom", "r");
         if ((frand == NULL) || fread(&seed16, 1, sizeof(seed16), frand) != sizeof(seed16)) {
-            fprintf(stderr, "WARNING: could not read 16 bytes from /dev/urandom; falling back to insecure PRNG\n");
             uint64_t t = time(NULL) * (uint64_t)1337;
+            fprintf(stderr, "WARNING: could not read 16 bytes from /dev/urandom; falling back to insecure PRNG\n");
             seed16[0] ^= t;
             seed16[1] ^= t >> 8;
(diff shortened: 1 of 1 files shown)


sipa:
utACK 8b3841c91daf7b75ed1bf81fa07db69d1688e8b6

src/tests.c
```diff
@@ -4918,7 +4918,8 @@ int main(int argc, char **argv) {
         }
     } else {
         FILE *frand = fopen("/dev/urandom", "r");
-        if ((frand == NULL) || fread(&seed16, sizeof(seed16), 1, frand) != sizeof(seed16)) {
+        if ((frand == NULL) || fread(&seed16, 1, sizeof(seed16), frand) != sizeof(seed16)) {
+            fprintf(stderr, "WARNING: could not read 16 bytes from /dev/urandom; falling back to insecure PRNG\n");
```
jonasnick: src/tests.c: In function ‘main’:
src/tests.c:4923:13: warning: ISO C90 forbids mixed declarations and code [-Wdeclaration-after-statement]
             uint64_t t = time(NULL) * (uint64_t)1337;

gmaxwell: @hdon to fix this, move the fprint down a line. Then we should be good to merge.

gmaxwell:
@sipa @jonasnick @real-or-random @hdon  please at least utACK my trivial fixup commit so I can merge. :P

real-or-random:
utACK

PR #518, opened 2018-03-10
Summarize build options after running configure

eklitzke:
This is a trivial build system change to summarize the build options after running configure.

Example output:
```
$ ./configure
....
<many lines omitted>
...
config.status: src/libsecp256k1-config.h is unchanged
config.status: executing depfiles commands
config.status: executing libtool commands

Build Options:
  with endomorphism   = no
  with ecmult precomp = yes
  with jni            = no
  module ecdh         = no
  module recovery     = no

  asm                 = x86_64
  bignum              = gmp
  field               = 64bit
  scalar              = 64bit

  CC                  = gcc
  CFLAGS              = -g -O2 -W -std=c89 -pedantic -Wall -Wextra -Wcast-align -Wnested-externs -Wshadow -Wstrict-prototypes -Wno-unused-function -Wno-long-long -Wno-overlength-strings -fvisibility=hidden -O3
  CPPFLAGS            = 
  LDFLAGS             = 
```

I tried to just include the configure options that looked interesting; let me know if there are any I didn't include that I should have.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -441,18 +441,6 @@ if test x"$use_external_asm" = x"yes"; then
   AC_DEFINE(USE_EXTERNAL_ASM, 1, [Define this symbol if an external (non-inline) assembly implementation is used])
 fi
 
-AC_MSG_NOTICE([Using static precomputation: $set_precomp])
-AC_MSG_NOTICE([Using assembly optimizations: $set_asm])
-AC_MSG_NOTICE([Using field implementation: $set_field])
-AC_MSG_NOTICE([Using bignum implementation: $set_bignum])
-AC_MSG_NOTICE([Using scalar implementation: $set_scalar])
-AC_MSG_NOTICE([Using endomorphism optimizations: $use_endomorphism])
-AC_MSG_NOTICE([Building benchmarks: $use_benchmark])
-AC_MSG_NOTICE([Building for coverage analysis: $enable_coverage])
-AC_MSG_NOTICE([Building ECDH module: $enable_module_ecdh])
-AC_MSG_NOTICE([Building ECDSA pubkey recovery module: $enable_module_recovery])
-AC_MSG_NOTICE([Using jni: $use_jni])
-
 if test x"$enable_experimental" = x"yes"; then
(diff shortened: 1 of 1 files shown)


sipa:
Looks good, but some of these things are already reported on lines 444-454. Can you incorporate/merge them?

eklitzke:
Fair point, I didn't notice because of the other messages printed by `AC_OUTPUT`. In that case the only thing new here is the display of `CFLAGS` (which is less interesting than the other information in my opinion). I kind of like the tabular display of output at the end but it's less compelling now that you mentioned the existing block of notices.

sipa:
@eklitzke I like having this kind of summary at the end, but there's no need to duplicate the information. Moving everything to the summary would be fine.

eklitzke:
Updated to move all the options to the summary at the end. The summary for me with no options to configure looks like this:

```
checking gmp.h presence... yes
checking for gmp.h... yes
checking for __gmpz_init in -lgmp... yes
checking for CRYPTO... yes
checking for EC functions in libcrypto... yes
checking whether byte ordering is bigendian... no
checking that generated files are newer than configure... done
configure: creating ./config.status
config.status: creating Makefile
config.status: creating libsecp256k1.pc
config.status: creating src/libsecp256k1-config.h
config.status: executing depfiles commands
config.status: executing libtool commands

Build Options:
  with endomorphism   = no
  with ecmult precomp = yes
  with jni            = no
  with benchmarks     = yes
  with coverage       = no
  module ecdh         = no
  module recovery     = no

  asm                 = x86_64
  bignum              = gmp
  field               = 64bit
  scalar              = 64bit

  CC                  = gcc
  CFLAGS              = -g -O2 -W -std=c89 -pedantic -Wall -Wextra -Wcast-align -Wnested-externs -Wshadow -Wstrict-prototypes -Wno-unused-function -Wno-long-long -Wno-overlength-strings -fvisibility=hidden -O3
  CPPFLAGS            = 
  LDFLAGS             = 

```

gmaxwell:
ACK

PR #521, opened 2018-04-01
Add libtool dependency to build instructions

IIIIllllIIIIllllIIIIllllIIIIllllIIIIll:


gmaxwell:
There is no runtime dependency on libtool, it's a build time dependency if using autotools. Autotools will detect the absence of libtool and tell you that you need it, if you use autotools.

PR #523, opened 2018-04-05
scratch: add stack frame support

apoelstra:
Replaces the single-blob stack space ith one that internally manages multiple blobs, which are exposed to the user as "frames". Users allocate new blobs with `secp256k1_scratch_allocate_frame` and deallocate them with `secp256k1_scratch_deallocate_frame`. Then any calls to `secp256k1_scratch_alloc` use the frame at the top of the stack. This is guaranteed to succeed, assuming that the frame allocation succeeded and that the user is not requesting more memory than the frame was allocated with.


intizarabe:
Thank you, really need this help

sipa:
utACK 6fe50439ae54abee820d2c1a9c686e5c0b2730f3

PR #529, opened 2018-04-10
fix tests.c in the count == 0 case

### LINKED ISSUE #528: Test with count == 0 and optimization disabled trip a verify
The repro process is a bit involved, but it seems like there is a possible problem somewhere in the ecmult_multi logic (or in gcc itself).

Repro:
Edit ./configure.ac to remove `CFLAGS="$CFLAGS -O3"`
Then:
```
$ CFLAGS='-g -O0' ../configure --with-asm=no
$ make tests && ./tests 0
[compiling...]
test count = 0
random seed = 6a777122d5fd053929192734d4913476
../src/field_5x52_impl.h:421: test condition failed: b->magnitude <= 8
Aborted
```

It's unclear why this problem doesn't show up when the optimization are turned on. The compiler used is gcc 7.3.0 .

apoelstra:
Fixes #528

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -2725,6 +2725,11 @@ void test_ecmult_multi(secp256k1_scratch *scratch, secp256k1_ecmult_multi_func e
     }
 
     /* Sanity check that zero scalars don't cause problems */
+    for (ncount = 0; ncount < 20; ncount++) {
+        random_scalar_order(&sc[ncount]);
+        random_group_element_test(&pt[ncount]);
+    }
+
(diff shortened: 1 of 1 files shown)


deadalnix:
I can confirm this fixes the problem. Thanks @apoelstra .

sipa:
utACK 95e99f196fd08a8b2c236ab99d7e7fec8f6dc78f

deadalnix:
Is that possible to merge this ? That sounds trivial enough.

PR #532, opened 2018-05-06
tests: Make sure we're not passing an uninitialized array to memcmp(...)

practicalswift:
Make sure we're not passing an uninitialized array to `memcmp(...)`.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -4194,13 +4194,13 @@ int test_ecdsa_der_parse(const unsigned char *sig, size_t siglen, int certainly_
     int ret = 0;
 
     secp256k1_ecdsa_signature sig_der;
-    unsigned char roundtrip_der[2048];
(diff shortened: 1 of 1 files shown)


gmaxwell:
I think this change would reduce the sensitivity of valgrind.  By initializing it ourselves we'll suppress errors where the underlying code, which should have initialized it at least up to the length we're using failed to do so.

gmaxwell:
I still don't see a reason to do this and there hasn't been recent activity. I believe I understand the motivation, I think the disadvantage of masking errors that valgring/mem-santizer would find makes it not worth it.   The valgrind component of the concern could be avoided by using valgrind macros to mark the array uninitilized but I'm aware of no comparable change for the sanitizers.

practicalswift:
Fix in #627.

PR #533, opened 2018-05-06
Make sure we're not using an uninitialized variable in secp256k1_wnaf_const(...)

practicalswift:
Make sure we're not using an uninitialized variable in `secp256k1_wnaf_const(...)`:

```
In file included from src/secp256k1.c:15:0,
                 from src/tests.c:17:
src/ecmult_const_impl.h: In function ‘secp256k1_wnaf_const’:
src/ecmult_const_impl.h:117:20: warning: ‘u’ may be used uninitialized in this function [-Wmaybe-uninitialized]
     wnaf[word] = u * global_sign;
                    ^
```

**Note to reviewers:** Perhaps an `assert(…);` is a bit drastic. What would be a more graceful way to handle this? :-)

### DIFF
--- a/src/ecmult_const_impl.h
+++ b/src/ecmult_const_impl.h
@@ -61,6 +61,10 @@ static int secp256k1_wnaf_const(int *wnaf, secp256k1_scalar s, int w, int size) 
     int bit;
     secp256k1_scalar neg_s;
     int not_neg_one;
+
+    VERIFY_CHECK(w > 0);
+    VERIFY_CHECK(size > 0);
+
     /* Note that we cannot handle even numbers by negating them to be odd, as is
      * done in other implementations, since if our scalars were specified to have
      * width < 256 for performance reasons, their negations would have width 256
@@ -93,7 +97,7 @@ static int secp256k1_wnaf_const(int *wnaf, secp256k1_scalar s, int w, int size) 
 
     /* 4 */
     u_last = secp256k1_scalar_shr_int(&s, w);
-    while (word * w < size) {
+    do {
         int sign;
         int even;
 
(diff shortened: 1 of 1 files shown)


sipa:
We don't use `assert` in the secp256k1 codebase. Instead I suggest you:
* Turn the `while` loop into a `do` loop.
* Add a `VERIFY_CHECK(size > 0);` (which only affects the code in unit tests).

practicalswift:
@sipa Makes perfect sense. Thanks for the quick feedback!

I've now pushed an updated version. Please re-review :-)

Empact:
utACK 5f10203

gmaxwell:
ACK

ymmah:
; ) 確認

jonasnick:
utACK

gmaxwell:
On re-review, I believe we need to VERIFY_CHECK w>0 for the loop transformation to be correct. (obviously w>0 or the code is already broken... so.) @sipa opinion?

real-or-random:
I think it's best to just `VERIFY_CHECK(w > 0)` too. If `w == 0`, we even divide by 0 in the `VERIFY_CHECK` here: 
https://github.com/bitcoin-core/secp256k1/blob/5f10203cfe3c40e7295aced2f36e090bb3f50e33/src/ecmult_const_impl.h#L119

real-or-random:
@practicalswift: Could you update this?

practicalswift:
@real-or-random Sure!

Just to clarify (better safe than sorry!):

Is the suggestion to change from:

```
VERIFY_CHECK(size > 0);
```

To:

```
VERIFY_CHECK(w > 0);
VERIFY_CHECK(size > 0);
```

And otherwise leave the PR unchanged?

real-or-random:
Yes that's my suggestion. :)

practicalswift:
@real-or-random Updated. Please re-review :-)

real-or-random:
utACK 248f0466112c96b9851c662fa829f20d28d16344

gmaxwell:
ACK

PR #543, opened 2018-07-31
Version4

fypm123:
Need Developing in Visual Studio~  Please help me,Thanks!


chfast:
It would be probably much easier to use CMake.

PR #545, opened 2018-08-01
Add Fedora spec file

suvayu:
This PR adds a spec file for Fedora.  You can find the corresponding build on [copr]( https://copr.fedorainfracloud.org/coprs/fatka/secp256k1/build/792466/).

Since there haven't been any releases, the spec file builds snapshots.  One needs to manually update the `git` tag in the spec file with the commit hash (`git show-ref --hash=8 HEAD`) and update the `checkout` tag with the commit date.  Once there is a release, this kludge can be cleaned up.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -7,7 +7,10 @@ noinst_LTLIBRARIES = $(JNI_LIB)
 else
 JNI_LIB =
 endif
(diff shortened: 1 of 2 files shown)


gmaxwell:
We'll revisit this when we make a formal release.  I've noted it in issue #622

PR #546, opened 2018-08-04
Signed-digit multi-comb for ecmult_gen

peterdettman:
See section 3.3 of https://eprint.iacr.org/2012/309 for a description of the algorithm. Briefly, the scalar is recoded into signed-binary form, then divided into several blocks. A separate precomp. table is prepared for each block, and performing a multiplication is done using one comb per block, interleaved.

This implementation is constant-time, preserves the existing scalar blinding, but the NUMS group element is not yet used, perhaps not really useful (no zeroes in the signed-digit recoding). ~~Static precomputation is not yet implemented.~~ Settings are overridden to let the exhaustive tests work.

You can play with the comb parameters in ecmult_gen.h .

Compared to the existing approach, this gives improved performance/memory tradeoffs, and allows considerable flexibility in the parameters depending on platform details.

The following table gives an idea of the sort of tradeoffs available (bench_sign results - best "min" of 3, asm=no, 64bit field and scalar, -O3, Haswell):

Blocks | Teeth | Spacing | Memory (KiB) | Time (us)
:----: | :---: | :-----: | :-----------: | :-------:
43|6|1|86|39.2
22|6|2|44|39.7
11|6|4|22|40.1
4|6|11|8|41.0
4|5|13|4|42.6
2|5|26|2|44.6
2|4|32|1|48.4
1|4|64|0.5|53.3
1|3|86|0.25|63.1
1|2|128|0.125|82.3
1|1|256|0.0625|140

For existing approach: 44.6us (64KiB precomp. data)

### DIFF
--- a/src/ecmult_gen.h
+++ b/src/ecmult_gen.h
@@ -10,9 +10,7 @@
 #include "scalar.h"
 #include "group.h"
 
-#ifndef USE_ECMULT_STATIC_PRECOMPUTATION
 #define USE_COMB 1
-#endif
 
 #if USE_COMB
 
--- a/src/ecmult_gen.h
+++ b/src/ecmult_gen.h
@@ -67,6 +67,11 @@ typedef struct {
      */
     secp256k1_ge_storage (*prec)[COMB_BLOCKS][COMB_POINTS];
 #if COMB_OFFSET
+    /* Signed recoding of a 256-bit scalar must be at least 257 bits, with the top bit always 1. We
+     * support a 256-bit comb over a 257-bit recoding by pre-adding an 'offset' value to the context's
+     * 'initial' value, to account for the high 1 bit. Note that the 'offset' is calculated to allow
+     * for the (COMB_SPACING - 1) doublings in the _ecmult_gen ladder.
+     */
     secp256k1_ge offset;
 #endif
 #else
--- a/src/ecmult_gen.h
+++ b/src/ecmult_gen.h
@@ -49,6 +49,16 @@
 
 #endif
 
+#if !(1 <= COMB_BLOCKS && COMB_BLOCKS <= 256)
+#  error "COMB_BLOCKS must be in the range [1, 256]"
+#endif
+#if !(1 <= COMB_TEETH && COMB_TEETH <= 8)
+#  error "COMB_TEETH must be in the range [1, 8]"
+#endif
+#if !(1 <= COMB_SPACING && COMB_SPACING <= 256)
+#  error "COMB_SPACING must be in the range [1, 256]"
+#endif
+
 /* The remaining COMB_* parameters are derived values, don't modify these. */
 #define COMB_BITS (COMB_BLOCKS * COMB_TEETH * COMB_SPACING)
 #define COMB_GROUPED ((COMB_SPACING == 1) && ((32 % COMB_TEETH) == 0))
@@ -56,6 +66,10 @@
 #define COMB_POINTS (1 << (COMB_TEETH - 1))
 #define COMB_POINTS_TOTAL (COMB_BLOCKS * COMB_POINTS)
 #define COMB_MASK (COMB_POINTS - 1)
+
+#if !(256 <= COMB_BITS && COMB_BITS <= 288)
+#  error "COMB_BITS must be in the range [256, 288]"
+#endif
 
 #endif
 
--- a/src/ecmult_gen.h
+++ b/src/ecmult_gen.h
@@ -10,7 +10,68 @@
 #include "scalar.h"
 #include "group.h"
 
+#ifndef USE_ECMULT_STATIC_PRECOMPUTATION
+#define USE_COMB 1
+#endif
+
+#if USE_COMB
+
+#if defined(EXHAUSTIVE_TEST_ORDER)
+
+  /* We need to control these values for exhaustive tests because
+   * the tables cannot have infinities in them (secp256k1_ge_storage
+   * doesn't support infinities) */
+#  if EXHAUSTIVE_TEST_ORDER > 32
+#    define COMB_BLOCKS 52
+#    define COMB_TEETH 5
+#  elif EXHAUSTIVE_TEST_ORDER > 16
+#    define COMB_BLOCKS 64
+#    define COMB_TEETH 4
+#  elif EXHAUSTIVE_TEST_ORDER > 8
+#    define COMB_BLOCKS 86
+#    define COMB_TEETH 3
+#  elif EXHAUSTIVE_TEST_ORDER > 4
+#    define COMB_BLOCKS 128
+#    define COMB_TEETH 2
+#  else
+#    define COMB_BLOCKS 256
+#    define COMB_TEETH 1
+#  endif
+
+#  define COMB_SPACING 1
+
+#else
+
+  /* COMB_BLOCKS, COMB_TEETH, COMB_SPACING must all be positive and the product of the three (COMB_BITS)
+   * must evaluate to a value in the range [256, 288]. The resulting memory usage for precomputation
+   * will be COMB_POINTS_TOTAL * sizeof(secp256k1_ge_storage). */
+  #define COMB_BLOCKS 4
+  #define COMB_TEETH 5
+  #define COMB_SPACING 13
+
+#endif
+
+/* The remaining COMB_* parameters are derived values, don't modify these. */
+#define COMB_BITS (COMB_BLOCKS * COMB_TEETH * COMB_SPACING)
+#define COMB_GROUPED ((COMB_SPACING == 1) && ((32 % COMB_TEETH) == 0))
+#define COMB_OFFSET (COMB_BITS == 256)
+#define COMB_POINTS (1 << (COMB_TEETH - 1))
+#define COMB_POINTS_TOTAL (COMB_BLOCKS * COMB_POINTS)
+#define COMB_MASK (COMB_POINTS - 1)
+
+#endif
+
 typedef struct {
+#if USE_COMB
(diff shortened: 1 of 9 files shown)


src/ecmult_gen_impl.h
```diff
@@ -141,7 +141,14 @@ static void secp256k1_ecmult_gen_context_build(secp256k1_ecmult_gen_context *ctx
 #endif
 #else
     (void)cb;
-    ctx->prec = (secp256k1_ge_storage (*)[64][16])secp256k1_ecmult_static_context;
+#if USE_COMB
+    ctx->prec = (secp256k1_ge_storage (*)[COMB_BLOCKS][COMB_POINTS])secp256k1_ecmult_gen_ctx_prec;
+#if COMB_OFFSET
+    secp256k1_ge_from_storage(&ctx->offset, &secp256k1_ecmult_gen_ctx_offset);
```
apoelstra: `secp256k1_ecmult_gen_ctx_offset` is not declared except in gen_context.c, so this line doesn't compile for me when I set the parameters to 4/4/16.

src/ecmult_gen.h
```diff
@@ -10,7 +10,68 @@
...
+  #define COMB_BLOCKS 4
+  #define COMB_TEETH 5
+  #define COMB_SPACING 13
+
+#endif
+
+/* The remaining COMB_* parameters are derived values, don't modify these. */
+#define COMB_BITS (COMB_BLOCKS * COMB_TEETH * COMB_SPACING)
+#define COMB_GROUPED ((COMB_SPACING == 1) && ((32 % COMB_TEETH) == 0))
+#define COMB_OFFSET (COMB_BITS == 256)
```
apoelstra: Can you add some documentation somewhere about what the offset does?

src/ecmult_gen_impl.h
peterdettman: This is a build system issue; there's no dependency of gen_context on ecmult_gen.h (and presumably others). At the moment, after changing comb parameters in ecmult_gen.h, you'd need to touch gen_context.c (or just make clean).

src/ecmult_gen.h
peterdettman: Done: https://github.com/bitcoin-core/secp256k1/pull/546/commits/e8beef9ab2476da9a912a7a12982b1e86d2f166c .

peterdettman:
The test failure (1407.7) seems spurious, might need investigation.

apoelstra:
Agreed it seems spurious - I had the same issue on #557. I kicked travis to see if it works this time.

src/ecmult_gen.h
apoelstra: Great, thanks!

src/ecmult_gen.h
```diff
@@ -10,7 +10,68 @@
...
+#    define COMB_TEETH 1
+#  endif
+
+#  define COMB_SPACING 1
+
+#else
+
+  /* COMB_BLOCKS, COMB_TEETH, COMB_SPACING must all be positive and the product of the three (COMB_BITS)
+   * must evaluate to a value in the range [256, 288]. The resulting memory usage for precomputation
+   * will be COMB_POINTS_TOTAL * sizeof(secp256k1_ge_storage). */
```
apoelstra: Should comment that `COMB_SPACING` should not exceed 32 or else the `bit_pos` logic in `secp256k1_ecmult_gen` stops working.

src/ecmult_gen.h
```diff
@@ -10,7 +10,68 @@
...
+  #define COMB_BLOCKS 4
+  #define COMB_TEETH 5
+  #define COMB_SPACING 13
+
+#endif
+
+/* The remaining COMB_* parameters are derived values, don't modify these. */
+#define COMB_BITS (COMB_BLOCKS * COMB_TEETH * COMB_SPACING)
+#define COMB_GROUPED ((COMB_SPACING == 1) && ((32 % COMB_TEETH) == 0))
```
apoelstra: I think we should drop `COMB_GROUPED` because it's impossible for `32 % COMB_TEETH` to be 0. (If it's zero, things obviously won't work...but if it's ≥ 32, then several other things break: `COMB_MASK` overflows; the `bits =` expression in ecmult_gen_impl.h:228 left-shifts too far; I also get this bizarre error about the size of `prec` being negative in ecmult_gen.h:71.)

apoelstra: In fact, once `COMB_TEETH` gets much past 10 we start running into trouble in `secp256k1_ecmult_gen_context_build` because we put `COMB_POINTS_TOTAL`-many gejs on the stack.

Given that this is a ridiculous thing to do I don't think we should try to support it, e.g. by using heap allocations rather than stack allocations in gen_context_build. We should just assume it doesn't happen and drop `COMB_GROUPED`.

apoelstra:
Since `USE_COMB` is now always set, I think it's worth adding another commit that removes the old precomp code.

Other than that, and my nits about extreme parameter settings, ACK.

src/ecmult_gen.h
peterdettman: I don’t see why, and all cases in the table above passed the tests before benchmarking.

src/ecmult_gen.h
apoelstra: You're right, I misread `32 % COMB_TEETH` (which is 0 for 1,2,4,8) as `COMB_TEETH % 32` (which can't be 0). Let me reassess COMB_GROUPED in that light.

src/ecmult_gen.h
apoelstra: Oh, I think I forgot to `make clean` after recompiling again. The tests do pass now.

But my reasoning was that you're only looking at one word of `recoded` at once, so if the `bit_pos` index jumps forward by more than one word, eventually `bit_pos >> 5` will increment by more than one per iteration and you'll have skipped an entire word. But I'm re-reading the code with fresher eyes and now I see that this is exactly the intended behaviour.

src/ecmult_gen_impl.h
apoelstra: This has burned me multiple times while testing -- @sipa can you advise how to fix this?

src/ecmult_gen.h
peterdettman: https://github.com/bitcoin-core/secp256k1/pull/546/commits/4c8ff9b648a7062133922d2d0c945712e77dd753 adds precompiler guards on the comb constants.

src/ecmult_gen_impl.h
```diff
@@ -225,13 +225,14 @@ static void secp256k1_ecmult_gen(const secp256k1_ecmult_gen_context *ctx, secp25
...
-            bit_pos += COMB_SPACING * COMB_TEETH;
+            bits = recoded[bit_pos >> 5] >> (bit_pos & 0x1F);
+            bit_pos += COMB_TEETH;
 #else
             bits = 0;
             for (tooth = 0; tooth < COMB_TEETH; ++tooth) {
-                bit = (recoded[bit_pos >> 5] >> (bit_pos & 0x1F)) & 1;
-                bits |= bit << tooth;
+                bit = recoded[bit_pos >> 5] >> (bit_pos & 0x1F);
+                bits &= ~(1 << tooth);
```
apoelstra: I think this bit-clearing line isn't actually needed, because `bits` starts at 0 and `1 << tooth` is unique on each iteration.

peterdettman: Note that ‘bit’ may now contain junk (or noise) bits beyond bit 0. Iterations after the first therefore need to clear the target bit (and we try to leave noise in the high bits). The change is in response to [1], which I think @gmaxwell mentioned on IRC a while back.

[1] https://www.usenix.org/system/files/conference/usenixsecurity18/sec18-alam.pdf

apoelstra: Ah! I understand, I was misreading `bit << tooth` as `(bit & 1) << tooth`.

apoelstra: Because the variable is named `bit` :P

apoelstra:
ACK except that I'd really like to fix the `make clean` issue.

src/ecmult_gen_impl.h
```diff
@@ -126,6 +205,81 @@ static void secp256k1_ecmult_gen(const secp256k1_ecmult_gen_context *ctx, secp25
...
+    /* Blind scalar/point multiplication by computing (n-b)G + bG instead of nG. */
+    secp256k1_scalar_add(&gnb, gn, &ctx->blind);
+    secp256k1_scalar_signed_recoding(recoded, &gnb, COMB_BITS + COMB_OFFSET);
+
+    comb_off = COMB_SPACING - 1;
+    for (;;) {
+        bit_pos = comb_off;
+        for (block = 0; block < COMB_BLOCKS; ++block) {
+#if COMB_GROUPED
+            bits = recoded[bit_pos >> 5] >> (bit_pos & 0x1F);
```
apoelstra: Should this be done as part of a cmov ladder, like the old algorithm, to avoid cache-timing attacks?

peterdettman: This line is reading a window from the scalar. IIUC, the cmov ladder you mean is the _ge_storage_cmov loop at lines 255-257.

apoelstra: Oh, yes! I missed that - you do indeed have a cmov ladder at the point where the secret data is actually used. Thanks.

sipa:
I've opened a rebased PR with somewhat better integration added in #693.

real-or-random:
> I've opened a rebased PR with somewhat better integration added in #693.

Closing this in favor of #693.

sipa:
Another iteration: #1058.

PR #553, opened 2018-08-15
add static context object which has no capabilities

apoelstra:


gmaxwell:
Should we document that you can't call set_illegal_callback on this and perhaps add some checks on some of the disallowed calls to fail if you try?

sipa:
@gmaxwell I believe the set of permitted functions on this object is exactly the functions which take a `const` context pointer.

apoelstra:
Sure, we can do a `VERIFY_CHECK` on those functions that the `ctx` is not the static one, much like how we compare to NULL now.

apoelstra:
Added a commit which `VERIFY_CHECK`s on all functions that modify the context object.

I considered using `ARG_CHECK` but I didn't see the point, since you can't change the callback function from its default CHECK-like behaviour. Also, since it is actually UB to call these functions on static data (without synchronization primitives anyway), I wanted to prevent it more emphatically.

include/secp256k1.h
```diff
@@ -179,6 +179,13 @@ typedef int (*secp256k1_nonce_function)(
 #define SECP256K1_TAG_PUBKEY_HYBRID_EVEN 0x06
 #define SECP256K1_TAG_PUBKEY_HYBRID_ODD 0x07
 
+/** A simple secp256k1 context object with no precomputed tables. These are useful for
+ *  type serialization/parsing functions which require a context object to maintain
+ *  API consistency, but currently do not require expensive precomputations or dynamic
+ *  allocations.
+ */
+SECP256K1_API extern secp256k1_context *secp256k1_context_no_precomp;
```
sipa: Can you mark this object `const`?

apoelstra: Done

src/secp256k1.c
```diff
@@ -91,6 +99,7 @@ secp256k1_context* secp256k1_context_clone(const secp256k1_context* ctx) {
 }
 
 void secp256k1_context_destroy(secp256k1_context* ctx) {
+    CHECK(ctx != secp256k1_context_no_precomp);
```
sipa: Are these necessary? A non-`const` ctx pointer is used, which can't be `secp256k1_context_no_precomp` (as that one is const).

sipa:
utACK 40fde611bd145bdafb78a52af6d197489101bafb

real-or-random:
ACK 40fde611bd145bdafb78a52af6d197489101bafb

PR #554, opened 2018-08-30
simple pow implementation

kallewoof:
This adds a `secp256k1_scalar_pow` which takes two scalars and returns pow(first, second) modulo the group order. It also adds wrappers in eckey and secp256k1 as `secp256k1_ec_privkey_tweak_pow`.

It should be helpful when using secp256k1 to calculate the jacobi symbol. Right now I believe you need GMP to do so otherwise.

### DIFF
--- a/src/scalar.h
+++ b/src/scalar.h
@@ -50,6 +50,9 @@ static void secp256k1_scalar_cadd_bit(secp256k1_scalar *r, unsigned int bit, int
 /** Multiply two scalars (modulo the group order). */
 static void secp256k1_scalar_mul(secp256k1_scalar *r, const secp256k1_scalar *a, const secp256k1_scalar *b);
 
+/** Calculate a^b modulo the group order. */
+static void secp256k1_scalar_pow(secp256k1_scalar *r, const secp256k1_scalar *a, const secp256k1_scalar *b);
+
 /** Shift a scalar right by some amount strictly between 0 and 16, returning
  *  the low bits that were shifted off */
 static int secp256k1_scalar_shr_int(secp256k1_scalar *r, int n);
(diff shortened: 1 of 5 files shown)


sipa:
That seems a bit too low level; we try to expose high level API features that are by default safe to use, not expose the internal mathematical operations.

kallewoof:
@sipa Having a way to calculate power-of seems like a natural addition to the priv/pubkey_tweak_add/muls that are there already, IMO. I don't have a strong opinion on the matter though, so I'll remove the secp256k1.h change.

sipa:
What use is there for raising a private key to some power? There isn't even a corresponding public key operation.

Maybe first motivate what application this is useful for before we discuss exposing a primitive.

kallewoof:
@sipa I was initially writing this in order to natively calculate the Jacobi symbol. I then realized I'm still missing some components (the biggest issue being that the scalars are modulo n while the jacobi symbol calculation is modulo p).

Perhaps what I'm ultimately aiming for is something like a `secp256k1_ec_pubkey_jacobi`, but does that implicitly mean "of the X coordinate" as in the Schnorr BIP? If it does, that would probably be useful, no?

sipa:
@kallewoof Integers modulo p internally are the data type 'secp256_fe`, and the Jacobi symbol is already implemented for them (search for "is_quad" somewhere in the source code, I can't look right now).

I don't think there is any use to tweaking private keys by raising them to a power. tweak_add is there for BIP32 and tweak_mul for BIP38.

As far as exposing Jacobi synbols are concerned, I think that's far too low level. It's something internal to the Schnorr scheme, so it'll be implemented as part of that, but not exposed.

That's perhaps a difference in philosophy, but I don't want to turn libsecp256k1 in a general math library. That brings in complexity, maintenance overhead for API stability, efficient but portable encodings for all internal data types, and attack surface.

kallewoof:
Gotcha! Thanks. I agree about the philosophy to not turn it into a math lib btw.

PR #557, opened 2018-09-22
Eliminate scratch memory used when generating contexts

apoelstra:
Builds on #553

### DIFF
--- a/src/ecmult_impl.h
+++ b/src/ecmult_impl.h
@@ -183,6 +183,7 @@ static void secp256k1_ecmult_odd_multiples_table_storage_var(const int n, secp25
     }
 
     /* Map `pj` back to our curve by multiplying its z-coordinate by `d.z`. */
+    zr = pj.z; /* save pj.z so we can use it to extract (d.z)^-1 from zi */
     secp256k1_fe_mul(&pj.z, &pj.z, &d.z);
     /* Directly set `pre[n - 1]` to `pj`, saving the inverted z-coordinate so
      * that we can combine it with the saved z-ratios to compute the other zs
@@ -193,7 +194,7 @@ static void secp256k1_ecmult_odd_multiples_table_storage_var(const int n, secp25
     secp256k1_ge_to_storage(&pre[n - 1], &p_ge);
 
     /* Compute the actual x-coordinate of D, which will be needed below. */
-    secp256k1_fe_inv_var(&d.z, &d.z);
+    secp256k1_fe_mul(&d.z, &zi, &zr);  /* d.z = 1/d.z */
     secp256k1_fe_sqr(&dx_over_dz_squared, &d.z);
     secp256k1_fe_mul(&dx_over_dz_squared, &dx_over_dz_squared, &d.x);
 
--- a/src/ecmult_impl.h
+++ b/src/ecmult_impl.h
@@ -137,24 +137,107 @@ static void secp256k1_ecmult_odd_multiples_table_globalz_windowa(secp256k1_ge *p
     secp256k1_ge_globalz_set_table_gej(ECMULT_TABLE_SIZE(WINDOW_A), pre, globalz, prej, zr);
 }
 
-static void secp256k1_ecmult_odd_multiples_table_storage_var(int n, secp256k1_ge_storage *pre, const secp256k1_gej *a, const secp256k1_callback *cb) {
-    secp256k1_gej *prej = (secp256k1_gej*)checked_malloc(cb, sizeof(secp256k1_gej) * n);
-    secp256k1_ge *prea = (secp256k1_ge*)checked_malloc(cb, sizeof(secp256k1_ge) * n);
-    secp256k1_fe *zr = (secp256k1_fe*)checked_malloc(cb, sizeof(secp256k1_fe) * n);
+static void secp256k1_ecmult_odd_multiples_table_storage_var(const int n, secp256k1_ge_storage *pre, const secp256k1_gej *a) {
+    secp256k1_gej d;
+    secp256k1_ge a_ge, d_ge, p_ge;
+    secp256k1_ge last_ge;
+    secp256k1_gej pj;
+    secp256k1_fe zi;
+    secp256k1_fe zr;
+    secp256k1_fe dx_over_dz_squared;
     int i;
 
-    /* Compute the odd multiples in Jacobian form. */
-    secp256k1_ecmult_odd_multiples_table(n, prej, zr, a);
-    /* Convert them in batch to affine coordinates. */
-    secp256k1_ge_set_table_gej_var(prea, prej, zr, n);
-    /* Convert them to compact storage form. */
-    for (i = 0; i < n; i++) {
-        secp256k1_ge_to_storage(&pre[i], &prea[i]);
+    VERIFY_CHECK(!a->infinity);
+
+    secp256k1_gej_double_var(&d, a, NULL);
+
+    /* First, we perform all the additions in an isomorphic curve obtained by multiplying
+     * all `z` coordinates by 1/`d.z`. In these coordinates `d` is affine so we can use
+     * `secp256k1_gej_add_ge_var` to perform the additions. For each addition, we store
+     * the resulting y-coordinate and the z-ratio, since we only have enough memory to
+     * store two field elements. These are sufficient to efficiently undo the isomorphism
+     * and recompute all the `x`s.
+     */
+    d_ge.x = d.x;
+    d_ge.y = d.y;
+    d_ge.infinity = 0;
+
+    secp256k1_ge_set_gej_zinv(&a_ge, a, &d.z);
+    pj.x = a_ge.x;
+    pj.y = a_ge.y;
+    pj.z = a->z;
+    pj.infinity = 0;
+
+    zr = d.z;
+    secp256k1_fe_normalize_var(&zr);
+    secp256k1_fe_to_storage(&pre[0].x, &zr);
+    secp256k1_fe_normalize_var(&pj.y);
+    secp256k1_fe_to_storage(&pre[0].y, &pj.y);
+
+    for (i = 1; i < n; i++) {
+        secp256k1_gej_add_ge_var(&pj, &pj, &d_ge, &zr);
+        secp256k1_fe_normalize_var(&zr);
+        secp256k1_fe_to_storage(&pre[i].x, &zr);
+        secp256k1_fe_normalize_var(&pj.y);
+        secp256k1_fe_to_storage(&pre[i].y, &pj.y);
     }
 
-    free(prea);
-    free(prej);
-    free(zr);
+    /* Map `pj` back to our curve by multiplying its z-coordinate by `d.z`. */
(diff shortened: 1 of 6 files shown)


src/group_impl.h
```diff
@@ -172,20 +172,23 @@ static void secp256k1_ge_set_all_gej_var(secp256k1_ge *r, const secp256k1_gej *a
     }
 }
 
-static void secp256k1_ge_set_table_gej_var(secp256k1_ge *r, const secp256k1_gej *a, const secp256k1_fe *zr, size_t len) {
+static void secp256k1_ge_set_table_gej_var(secp256k1_ge_storage *r, const secp256k1_gej *a, const secp256k1_fe *zr, size_t len) {
```
apoelstra: Oh, this function can be eliminated entirely actually.

src/ecmult_impl.h
```diff
@@ -137,24 +137,94 @@ static void secp256k1_ecmult_odd_multiples_table_globalz_windowa(secp256k1_ge *p
...
+    /* Directly set `pre[n - 1]` to `pj`, saving the inverted z-coordinate so
+     * that we can combine it with the saved z-ratios to compute the other zs
+     * without any more inversions. */
+    secp256k1_fe_inv_var(&zi, &pj.z);
+    secp256k1_ge_set_gej_zinv(&p_ge, &pj, &zi);
+    secp256k1_ge_from_storage(&last_ge, &pre[n - 1]);
+    secp256k1_ge_to_storage(&pre[n - 1], &p_ge);
+
+    /* Compute the actual x-coordinate of D, which will be needed below. */
+    secp256k1_fe_inv_var(&d.z, &d.z);
```
apoelstra: should batch this inversion with the one 5 lines up

gmaxwell:
Concept ACK.  Are you going to batch that extra inv?

apoelstra:
Yep, done.

apoelstra:
@sipa can you take a look at this?

src/group_impl.h
```diff
@@ -126,46 +126,48 @@ static void secp256k1_ge_set_gej_var(secp256k1_ge *r, secp256k1_gej *a) {
...
 
-static void secp256k1_ge_set_all_gej_var(secp256k1_ge *r, const secp256k1_gej *a, size_t len, const secp256k1_callback *cb) {
-    secp256k1_fe *az;
-    secp256k1_fe *azi;
+static void secp256k1_ge_set_all_gej_var(secp256k1_ge *r, const secp256k1_gej *a, size_t len) {
+    secp256k1_fe u;
     size_t i;
-    size_t count = 0;
-    az = (secp256k1_fe *)checked_malloc(cb, sizeof(secp256k1_fe) * len);
+    size_t last_i = (size_t)-1;
```
gmaxwell: Why no SIZE_MAX?

apoelstra:
Replaced all instances of `(size_t)-1)` with `SIZE_MAX`. Confirmed that `git grep -n '(size_t)\w*-1'` runs clean.

sipa:
Needs rebase on #553 (not urgent, I'll review soon rebased or not).

apoelstra:
Rebased.

src/group_impl.h
```diff
@@ -126,30 +126,50 @@ static void secp256k1_ge_set_gej_var(secp256k1_ge *r, secp256k1_gej *a) {
...
-static void secp256k1_ge_set_all_gej_var(secp256k1_ge *r, const secp256k1_gej *a, size_t len, const secp256k1_callback *cb) {
-    secp256k1_fe *az;
-    secp256k1_fe *azi;
+static void secp256k1_ge_set_all_gej_var(secp256k1_ge *r, const secp256k1_gej *a, size_t len) {
+    secp256k1_fe u;
     size_t i;
-    size_t count = 0;
-    az = (secp256k1_fe *)checked_malloc(cb, sizeof(secp256k1_fe) * len);
+    size_t last_i = SIZE_MAX;
+    size_t first_i = SIZE_MAX;
```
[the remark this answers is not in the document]
apoelstra: Yeah, though this was an algorithm-debugging check that probably should be removed now.

apoelstra:
Addressed all comments.

src/ecmult_impl.h
```diff
@@ -137,24 +137,94 @@ static void secp256k1_ecmult_odd_multiples_table_globalz_windowa(secp256k1_ge *p
...
+    secp256k1_fe_sqr(&dx_over_dz_squared, &d.z);
+    secp256k1_fe_mul(&dx_over_dz_squared, &dx_over_dz_squared, &d.x);
+
+    i = n - 1;
+    while (i > 0) {
+        secp256k1_fe zi2, zi3;
+        i--;
+        /* For the remaining points, we extract the z-ratio from the stored
```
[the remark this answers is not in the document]
apoelstra: Added comment.

src/ecmult_impl.h
```diff
@@ -137,24 +137,94 @@ static void secp256k1_ecmult_odd_multiples_table_globalz_windowa(secp256k1_ge *p
...
+         * x-coordinate, compute its z^-1 from that, and compute the full
+         * point from that: */
+        secp256k1_fe_mul(&zi, &zi, &last_ge.x);
+        secp256k1_fe_sqr(&zi2, &zi);
+        secp256k1_fe_mul(&zi3, &zi2, &zi);
+        /* To compute x, we observe that the z-ratio is simply `h` from
```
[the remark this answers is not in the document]
apoelstra: Expanded and rearranged this comment block.

src/group_impl.h
```diff
@@ -126,30 +126,50 @@ static void secp256k1_ge_set_gej_var(secp256k1_ge *r, secp256k1_gej *a) {
...
     }
+    if (last_i == SIZE_MAX) {
+        return;
+    }
+    secp256k1_fe_inv_var(&u, &r[last_i].x);
 
-    azi = (secp256k1_fe *)checked_malloc(cb, sizeof(secp256k1_fe) * count);
-    secp256k1_fe_inv_all_var(azi, az, count);
-    free(az);
+    i = last_i + 1;
```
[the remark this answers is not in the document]
apoelstra: Ah, nice

src/ecmult_impl.h
```diff
@@ -137,24 +137,94 @@ static void secp256k1_ecmult_odd_multiples_table_globalz_windowa(secp256k1_ge *p
...
+    d_ge.infinity = 0;
+
+    secp256k1_ge_set_gej_zinv(&a_ge, a, &d.z);
+    pj.x = a_ge.x;
+    pj.y = a_ge.y;
+    pj.z = a->z;
+    pj.infinity = 0;
+
+    zr = d.z;
+    secp256k1_fe_normalize(&zr);
```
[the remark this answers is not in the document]
apoelstra: done

sipa:
ACK, please squash the fixes in the respective commits?

It's a really cool trick; I hadn't expected it would be possible to reconstruct everything from Y and Z-ratios, but the formulas work out.

It seems this actually makes `secp256k1_context_create(SECP256K1_CONTEXT_VERIFY)` around 6% faster for me (when disabling static precomputation...).

apoelstra:
Squashed.

peterdettman:
I think it's a bit easier to understand if the z-ratios are stored with the y-values that they'll be used with (instead of off by 1). 

Like this: https://github.com/peterdettman/secp256k1/commit/f1ab49c5dfb11933ea64299c0c769a36eed4fa87

src/ecmult_impl.h
```diff
@@ -137,24 +137,108 @@ static void secp256k1_ecmult_odd_multiples_table_globalz_windowa(secp256k1_ge *p
...
+         *    rzr = d_x * z^2 - x
+         *
+         * where `d_x` is the x coordinate of `D` and `(x, z)` are Jacobian
+         * coordinates of our desired point.
+         *
+         * Rearranging and dividing by `z^2` to convert to affine, we get
+         *
+         *     x = d_x - rzr / z^2
+         *       = d_x - rzr * zi2
+         */
```
peterdettman: The comments here could be improved with respect to the `d.z` factors that are missing/implicit in the equations.

apoelstra: Sure, so above where I just say "this is equal to `h`" I could write out
```
rzr = h
    = d.x * z^2 - x * d.z^2
    = d.x * z^2 - x    because d.z = 1
```
Would that be sufficient?

apoelstra:
I think @peterdettman's commit is great, it does simplify the logic a fair bit.

src/ecmult_impl.h
peterdettman: I’m referring to the original d.z from the doubling at line 152 (call this d_z), not d_ge.z == 1. So the conversion to affine also involves this d_z:
`x_affine = d_x / d_z^2 - rzr / z^2 / d_z^2`
and note that:
`zi == 1 / z / d_z`
in each loop iteration.

apoelstra: Ah, right -- I say `d_x` in the comments but then use a variable called `dx_over_dz_squared`, which is confusing.

peterdettman: Yes, also the comments gloss over the 1/d_z factor in zi.

apoelstra:
Added both of Peter's commits squashed into one, and an addition commit that adds even more comments explaining how the effective-affine isomorphic curve plays into everything.

@peterdettman do you think this is better?

@sipa can you comment on this? We're back to "working backward" in the sense that I do a series of equations starting from abstract coordinates and ending up with values that have variable bindings, but it's all pretty explicit now.

src/ecmult_impl.h
```diff
@@ -190,6 +190,23 @@ static void secp256k1_ecmult_odd_multiples_table_storage_var(const int n, secp25
...
+     *           z-ratios we will obtain the inverse z-coordinate of all
+     *           other points. These inverses would have been expensive
+     *           to compute directly.
+     *     `d.z` is the multiplier that must be applied to all z-coordinates
+     *           to move from our isomorphic curve back to secp256k1. We
+     *           applied this multiplier above to the inverse-z-coordinate
+     *           computing `pre[n-1]`, and since all subsequent inverse-zs
+     *           are computed as multiples of this value, they will also
+     *           have the multiplier applied to them without any further
+     *           computation.
```
apoelstra: Actually this block is confusing and does not distinguish between the two curves. Will rewrite.

sipa:
@apoelstra Looks very readable now; I'll review in detail soon.

peterdettman:
Yes, better, thanks for improving that.

gmaxwell:
Nice!

sipa:
utACK b3bf5f99a3251e3d72ffde1f39158af6ea133e33

apoelstra:
Ready for merge?

gmaxwell:
Addition/subtraction with constants are effectively free except sometimes in the tightest of loops, the compiler will usually remove them, when it doesn't they usually get hidden by pipelines. This code is run-once code, it should be optimized for clarity and maintainability most and testability much more than for size (within reason) and performance.

PR #561, opened 2018-10-07
Respect LDFLAGS and #undef STATIC_PRECOMPUTATION if using basic config

DesWurstes:
Update: **This is a trimmed pull request with strong rationale.**

- Adding `--coverage` shouldn't reset `LDFLAGS`, this is definitely a typo
- The basic configuration should undefine `STATIC_PRECOMPUTATION`, as generating it is not supported and it complicates #549


roconnor-blockstream:
You probably also want to fix https://github.com/bitcoin-core/secp256k1/blob/b408c6a8b287003d1ade5709e6f7bc3c7f1d5be7/src/bench_internal.c#L189 as now the benchmark is invoking undefined behaviour rather than simply computing the wrong value.

gmaxwell:
ACK

PR #562, opened 2018-10-09
Make use of TAG_PUBKEY constants in secp256k1_eckey_pubkey_parse

Empact:
This results in more self-documenting code.

Constants defined here:
https://github.com/bitcoin-core/secp256k1/blob/1e6f1f5ad5e7f1e3ef79313ec02023902bf8175c/include/secp256k1.h#L175-L180


gmaxwell:
Meh. I've never been a fan of replacing simple immediate with construction-opaque macros.  But doing this here makes the code more consistent, so I will defer to the judgement that went into using EVEN and ODD in the surrounding code previously and ACK.  I reserve the right to whine about the conversion of other simple constants elsewhere. :P

PR #566, opened 2018-10-22
Enable context creation in preallocated memory

real-or-random:
@apoelstra 

This builds on #557.

Manually managing memory is always a pain in the ass in some way. I tried to keep the pain manageable. I'm open to suggestions to make this less ugly or error-prone.

to do:
 * tests
 * export functions

### DIFF
--- a/include/secp256k1_preallocated.h
+++ b/include/secp256k1_preallocated.h
@@ -1,0 +1,98 @@
+#ifndef SECP256K1_PREALLOCATED_H
+#define SECP256K1_PREALLOCATED_H
+
+#include "secp256k1.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/* The module provided by this header file is intended for settings in which it
+ * is not possible or desirable to rely on dynamic memory allocation. It provides
+ * functions for creating, cloning, and destroying secp256k1 context objects in a
+ * contiguous fixed-size block of memory provided by the caller.
+ *
+ * It is guaranteed that functions in this module will not call malloc or its
+ * friends realloc, calloc, and free.
+ */
+
+/** Determine the memory size of a secp256k1 context object to be created in
+ *  caller-provided memory.
+ *
+ *  The purpose of this function is to determine how much memory must be provided
+ *  to secp256k1_context_preallocated_create.
+ *
+ *  Returns: the required size of the caller-provided memory block
+ *  In:      flags:    which parts of the context to initialize.
+ */
+SECP256K1_API size_t secp256k1_context_preallocated_size(
+    unsigned int flags
+) SECP256K1_WARN_UNUSED_RESULT;
+
+/** Create a secp256k1 context object in caller-provided memory.
+ *
+ *  Returns: a newly created context object.
+ *  In:      prealloc: a pointer to a rewritable contiguous block of memory of
+ *                     size at least secp256k1_context_preallocated_size(flags)
+ *                     bytes, suitably aligned to hold an object of any type
+ *                     (cannot be NULL)
+ *           flags:    which parts of the context to initialize.
+ *
+ *  See also secp256k1_context_randomize.
+ */
+SECP256K1_API secp256k1_context* secp256k1_context_preallocated_create(
+    void* prealloc,
+    unsigned int flags
+) SECP256K1_ARG_NONNULL(1) SECP256K1_WARN_UNUSED_RESULT;
+
+/** Determine the memory size of a secp256k1 context object to be copied into
+ *  caller-provided memory.
+ *
+ *  The purpose of this function is to determine how much memory must be provided
+ *  to secp256k1_context_preallocated_clone when copying the context ctx.
+ *
+ *  Returns: the required size of the caller-provided memory block.
+ *  In:      ctx: an existing context to copy (cannot be NULL)
+ */
+SECP256K1_API size_t secp256k1_context_preallocated_clone_size(
+    const secp256k1_context* ctx
+) SECP256K1_ARG_NONNULL(1) SECP256K1_WARN_UNUSED_RESULT;
+
+/** Copy a secp256k1 context object into caller-provided memory.
+ *
+ *  Returns: a newly created context object.
+ *  Args:    ctx:      an existing context to copy (cannot be NULL)
+ *  In:      prealloc: a pointer to a rewritable contiguous block of memory of
+ *                     size at least secp256k1_context_preallocated_size(flags)
+ *                     bytes, suitably aligned to hold an object of any type
+ *                     (cannot be NULL)
+ */
+SECP256K1_API secp256k1_context* secp256k1_context_preallocated_clone(
+    const secp256k1_context* ctx,
+    void* prealloc
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_WARN_UNUSED_RESULT;
+
+/** Destroy a secp256k1 context object that has been created in
+ *  caller-provided memory.
+ *
+ *  The context pointer may not be used afterwards.
+ *
+ *  The context to destroy must have been created using
+ *  secp256k1_context_preallocated_create or secp256k1_context_preallocated_clone.
+ *  If the context has instead been created using secp256k1_context_create or
+ *  secp256k1_context_clone, the behaviour is undefined. In that case,
+ *  secp256k1_context_destroy must be used instead.
+ *
+ *  Args:   ctx: an existing context to destroy, constructed using
+ *               secp256k1_context_preallocated_create or
+ *               secp256k1_context_preallocated_clone (cannot be NULL)
+ */
+SECP256K1_API void secp256k1_context_preallocated_destroy(
+    secp256k1_context* ctx
+);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* SECP256K1_PREALLOCATED_H */
--- a/include/secp256k1_preallocated.h
+++ b/include/secp256k1_preallocated.h
@@ -12,7 +12,14 @@ extern "C" {
  * functions for creating, cloning, and destroying secp256k1 context objects in a
  * contiguous fixed-size block of memory provided by the caller.
  *
- * It is guaranteed that functions in this module will not call malloc or its
+ * Context objects created by functions in this module can be used like contexts
+ * objects created by functions in secp256k1.h, i.e., they can be passed to any
+ * API function that excepts a context object (see secp256k1.h for details). The
+ * only exception is that context objects created by functions in this module
+ * must be destroyed using secp256k1_context_preallocated_destroy (in this
+ * module) instead of secp256k1_context_destroy (in secp256k1.h).
+ *
+ * It is guaranteed that functions in by this module will not call malloc or its
  * friends realloc, calloc, and free.
  */
 
@@ -31,14 +38,28 @@ SECP256K1_API size_t secp256k1_context_preallocated_size(
 
 /** Create a secp256k1 context object in caller-provided memory.
  *
+ *  The caller must provide a pointer to a rewritable contiguous block of memory
+ *  of size at least secp256k1_context_preallocated_size(flags) bytes, suitably
+ *  aligned to hold an object of any type.
+ *
+ *  The block of memory is exclusively owned by the created context object during
+ *  the lifetime of this context object, which begins with the call to this
+ *  function and ends when a call to secp256k1_context_preallocated_destroy
+ *  (which destroys the context object again) returns. During the lifetime of the
+ *  context object, the caller is obligated not to access this block of memory,
+ *  i.e., the caller may not read or write the memory, e.g., by copying the memory
+ *  contents to a different location or trying to create a second context object
+ *  in the memory. In simpler words, the prealloc pointer (or any pointer derived
+ *  from it) should not be used during the lifetime of the context object.
+ *
  *  Returns: a newly created context object.
  *  In:      prealloc: a pointer to a rewritable contiguous block of memory of
  *                     size at least secp256k1_context_preallocated_size(flags)
- *                     bytes, suitably aligned to hold an object of any type
- *                     (cannot be NULL)
+ *                     bytes, as detailed above (cannot be NULL)
  *           flags:    which parts of the context to initialize.
  *
- *  See also secp256k1_context_randomize.
+ *  See also secp256k1_context_randomize (in secp256k1.h)
+ *  and secp256k1_context_preallocated_destroy.
  */
 SECP256K1_API secp256k1_context* secp256k1_context_preallocated_create(
     void* prealloc,
@@ -47,9 +68,6 @@ SECP256K1_API secp256k1_context* secp256k1_context_preallocated_create(
 
 /** Determine the memory size of a secp256k1 context object to be copied into
  *  caller-provided memory.
- *
- *  The purpose of this function is to determine how much memory must be provided
- *  to secp256k1_context_preallocated_clone when copying the context ctx.
  *
  *  Returns: the required size of the caller-provided memory block.
  *  In:      ctx: an existing context to copy (cannot be NULL)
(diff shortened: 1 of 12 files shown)


src/util.h
```diff
@@ -84,6 +84,27 @@ static SECP256K1_INLINE void *checked_realloc(const secp256k1_callback* cb, void
...
+ * leave some room because currently we don't care about a few bytes. */
+#define ALIGNMENT 16
+#endif
+
+#define ROUND_TO_ALIGN(size) (((size + ALIGNMENT - 1) / ALIGNMENT) * ALIGNMENT)
+
+static SECP256K1_INLINE void *manual_alloc(void** prealloc_ptr, size_t alloc_size, void* base, size_t max_size) {
+    size_t aligned_alloc_size = ROUND_TO_ALIGN(alloc_size);
+    void* ret = *prealloc_ptr;
+    CHECK((char*)*prealloc_ptr != NULL);
```
real-or-random: Todo: This check is not enough. The user could give us a NULL pointer, and then we want to call the callback and not just abort.

apoelstra: I think it's fine. We call this only internally, so actually even a `VERIFY_CHECK` would be sufficient.

In the public-API-facing function where this pointer originates we should use an `ARG_CHECK`.

src/secp256k1.c
```diff
@@ -64,6 +64,25 @@ static secp256k1_context secp256k1_context_no_precomp_ = {
 };
 secp256k1_context *secp256k1_context_no_precomp = &secp256k1_context_no_precomp_;
 
+size_t secp256k1_context_prealloc_size(unsigned int flags) {
```
apoelstra: This needs a corresponding entry in `include/secp256k1.h`.

src/secp256k1.c
```diff
@@ -83,29 +83,43 @@ size_t secp256k1_context_prealloc_size(unsigned int flags) {
...
+    secp256k1_context* ret = (secp256k1_context*)manual_alloc(&prealloc, sizeof(secp256k1_context), base, ctx_size);
+
     ret->illegal_callback = default_illegal_callback;
     ret->error_callback = default_error_callback;
 
     if (EXPECT((flags & SECP256K1_FLAGS_TYPE_MASK) != SECP256K1_FLAGS_TYPE_CONTEXT, 0)) {
             secp256k1_callback_call(&ret->illegal_callback,
                                     "Invalid flags");
-            free(ret);
+            /* Unreachable */
```
apoelstra: This isn't unreachable, is it? If the user provides bad flags it will trigger.

src/secp256k1.c
```diff
@@ -83,29 +83,43 @@ size_t secp256k1_context_prealloc_size(unsigned int flags) {
     return ret;
 }
 
-secp256k1_context* secp256k1_context_create(unsigned int flags) {
-    secp256k1_context* ret = (secp256k1_context*)checked_malloc(&default_error_callback, sizeof(secp256k1_context));
+secp256k1_context* secp256k1_context_prealloc_create(void* prealloc, unsigned int flags) {
```
apoelstra: Needs a corresponding entry in `include/secp256k1.h`, and we should `EXPECT` that `prealloc` is non-NULL.

src/secp256k1.c
real-or-random: Yep, "exporting functions" is on the todo list. :)

src/secp256k1.c
real-or-random: We call `default_illegal_callback` here, which aborts, so the return is unreachable.

(By the way I think the diff is somewhat misleading. I'm not adding `/* Unreachable */` because I'm removing the `free`; the two changes are unrelated.)

apoelstra: Oh, yeah, you're right.

real-or-random:
Addressed your comments including the memory leak (reason was that I didn't really implement context cloning) and tidied up a little.

Open things:
 - This removes the ability to clone `ecmult_context` and `ecmult_gen_context` individually without cloning an entire `secp256k1_context` in order to keep the code simple. Cloning the full `secp256k1_context` is still supported. We don't clone these "subcontexts" internally, and I don't see a reason why we would need this. Thoughts?
 - Do we need a `secp256k1_context_prealloc_clone()` to clone into preallocated memory?
 - One disadvantage of the approach that I implemented is that the user may call screw it up by calling `secp256k1_context_prealloc_size(flags)` with different flags than `secp256k1_context_prealloc_create(..., flags)`. But I think this is acceptable.
 - todo: Exporting functions in public header (I will do this at the end once we're clear about the other points)

apoelstra:
1. No, it's fine if we lose the ability to clone individual parts of a context object. We only had that separation for code cleanliness.
2. I'd like to have a `secp256k1_context_prealloc_clone` but I don't know if it'd be useful. Just in case.
3. Yeah, if the user is inconsistent with flags they'll get UB. Agreed that this is fine, that's a pretty silly thing to do.

src/ecmult_gen_impl.h
```diff
@@ -106,21 +107,15 @@ static int secp256k1_ecmult_gen_context_is_built(const secp256k1_ecmult_gen_cont
...
-                                               const secp256k1_ecmult_gen_context *src, const secp256k1_callback* cb) {
-    if (src->prec == NULL) {
-        dst->prec = NULL;
-    } else {
+static void secp256k1_ecmult_gen_context_finalize_memcpy(secp256k1_ecmult_gen_context *dst, const secp256k1_ecmult_gen_context *src) {
 #ifndef USE_ECMULT_STATIC_PRECOMPUTATION
-        dst->prec = (secp256k1_ge_storage (*)[64][16])checked_malloc(cb, sizeof(*dst->prec));
-        memcpy(dst->prec, src->prec, sizeof(*dst->prec));
+    if (src->prec != NULL) {
+        /* We cast to void* first to suppress a -Wcast-align warning in clang. */
```
apoelstra: Everything about this function is terrifying :) But I think it's correct.

real-or-random: I'm learning a lot about C...

real-or-random:
* squashed
 * implemented cloning into prealloc memory
 * added tests
 * exported API functions

I can drop the last commit which creates a separate header file. I think it's useful because the average user does not need the new function but the new header file somehow violates the idea that additional headers are modules.

We could make the life easier for the user if we store in the context whether we allocated the memory  or the user. (And then we need only one destroy function and the user cannot screw it up.)

src/ecmult_gen_impl.h
real-or-random: No but seriously, despite of the terrifying casts, I believe this memcpy cloning style is nicer because it's much less code and thus much less code to make mistakes. Open to suggestions of course.

gmaxwell:
The interface probably needs to document the required alignment of caller memory (at least size that it must be aligned similarly to the result of malloc, up to the largest primitive type).  Unless we want to handle aligning the struct ourselves. 

It would be useful to test this stuff on something that traps on unaligned reads (not x86, which does so only in freaky cases like unaligned SIMD loads using aligned instructions but only when they cross a page boundary)...

include/secp256k1_prealloc.h
```diff
@@ -0,0 +1,89 @@
...
+    const secp256k1_context* ctx
+) SECP256K1_ARG_NONNULL(1) SECP256K1_WARN_UNUSED_RESULT;
+
+/** Copy a secp256k1 context object into caller-provided memory.
+ *
+ *  Returns: a newly created context object.
+ *  Args:    ctx:      an existing context to copy (cannot be NULL)
+ *  In:      prealloc: a pointer to a rewritable contiguous block of memory of
+ *                     size at least secp256k1_context_prealloc_size(flags)
+ *                     bytes, suitably aligned to hold an object of any type
```
real-or-random: @gmaxwell I wrote this as documentation. Suggestions? "any primitive type" may be better than "any type".

real-or-random:
> It would be useful to test this stuff on something that traps on unaligned reads (not x86, which just runs slower but traps only in freaky cases like unaligned SIMD loads using aligned instructions but only when they cross a page boundary)...
Indeed. I haven't tested it but this looks promising for x86 and might do to job: https://stackoverflow.com/a/17748435/2725281

> I might also suggest calling the function something like prealloced not prealloc, functions are usually read as verbs that describe what they do. So I would assume that prealloc performs preallocation.
Good idea.

real-or-random:
> Indeed. I haven't tested it but this looks promising for x86 and might do to job: https://stackoverflow.com/a/17748435/2725281

Well no. It does what it promises, i.e., throwing SIGBUS at you when access unaligned memory, but there are too many false positives. If anyone is interested: You need to link statically (otherwise you get a SIGBUS in the dynamic linker), better disable openssl tests and don't use gmp (otherwise you get a SIGBUS in gmp). But even then, gcc will leave you with too many unaligned accesses in the binary.

apoelstra:
Can you try adding `-mno-unaligned-access` to the CFLAGS?

real-or-random:
That flag is not available for x86. I did some digging and I don't think there's a way to force gcc or clang to perform only aligned accesses. Unfortunately, qemu does not help either; it just won't trap.

real-or-random:
Renamed the functions and rebased on the changes in #557

apoelstra:
I think we're ready to define a public API for this

src/secp256k1.c
```diff
@@ -83,6 +83,17 @@ size_t secp256k1_context_prealloc_size(unsigned int flags) {
     return ret;
 }
 
+size_t secp256k1_context_prealloc_size_for_clone(const secp256k1_context* ctx) {
```
apoelstra: Should be marked `static` so it's clear this isn't exposed to the user.

apoelstra:
ACK except `static` nit.

I don't feel very strongly about it, but I'd prefer we drop the commit that uses a separate header file.

real-or-random:
fixed... (Haven't touched the separate header so far, I can still do that.)

real-or-random:
* rebased on master (no code changes besides merging)
 * squashed fixups (no code changes)
 * squashed the "rename from _prealloc to _preallocated" commit into the individual commits (no code changes)

include/secp256k1_preallocated.h
```diff
@@ -0,0 +1,89 @@
+#ifndef SECP256K1_preallocated_H
+#define SECP256K1_preallocated_H
+
+#include "secp256k1.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
```
jonasnick: Would be nice to have a 1 or 2 sentence explanation for what this module does.

sipa:
Concept ACK

real-or-random:
addressed @sipa's comments

real-or-random:
I forced-push (only changed the fixups). Changes:  
 * Replaced CHECKs by VERIFY_CHECKS in manual_alloc() because we use CHECKs only in tests AFAIU
 * Removed "Unreachable" comments... They're kind of inaccurate. This really depends on the behavior of the callbacks.
 * Added a note to the docs that malloc() is at most called once in the normal create and clone functions (as suggested by @gmaxwell) 

cc @apoelstra

include/secp256k1_preallocated.h
```diff
@@ -1,12 +1,21 @@
...
+#ifndef SECP256K1_PREALLOCATED_H
+#define SECP256K1_PREALLOCATED_H
 
 #include "secp256k1.h"
 
 #ifdef __cplusplus
 extern "C" {
 #endif
 
+/* The module provided by this header file is intended for settings, in which it
```
apoelstra: Should remove the `,`

apoelstra:
utACK nitfixes

gmaxwell:
> because we use CHECKs only in tests AFAIU

This concerned me some but so far I don't see a case where it's obviously wrong yet (haven't reviewed the whole thing yet).

But I think it might be useful to comment on runtime failure handling works.  I think our design is this:  The program can find itself in an impossible state (IS) as a result of memory corruption/hardware fault, miscompilation, operating system failure, or a serious programming error in the library itself or the caller (e.g. malloc fails, a non-nullable argument is null, a loop is out of range etc.).

A few kinds of IS get detected at runtime, particularly ones which get detected as a product of tests at the API boundary which wouldn't make any sense to make VERIFY_CHECKS (for multiple reasons including that those are only used in our own tests and couldn't detect bad interactions with a caller).

It's also generally considered impolite for libraries to be abort() prone: Many libraries haven't given this stuff much thought and have sometimes called abort() for normal error handling rather than actual corruption. You could imagine a signature verifier that aborted if the DER input didn't deserialize, instead of just returning false... and imagine how cross an application author would be to discover that behaviour in the field.  Even when abort() usage is intended to be limited to impossible states, sometimes the library authors make a mistake and a state is both possible and otherwise harmless... this, and performance, is why we prefer to do most of our impossible state detection via VERIFY_CHECKs that run only in our tests where mistakes in their logic won't introduce real bugs. So we generally only want to runtime detect IS in cases that have the best tradeoffs:  Cases that are really unambiguously wrong e.g. a null pointer we're about to dereference vs some complicated algebraic condition  and cases that arise out of interactions with the caller since our tests can't discover errors in the caller or in caller-library interactions are both candidates for runtime detection.

When an IS is detected we certainly don't want to just continue on like nothing bad happened. But we can't just return errors: (1) The IS may be detected in a location where it is not sensible to return an error (after all, the error is "impossible" so bending the design to accommodate returning might be a bad change), (2) the IS may happen inside an external interface which "cannot fail". (3) Returning on an IS could be confused for a normal false condition like an invalid signature (so IS happens and a CHECKSIG NOT passes when it shouldn't), (4) because IS is "impossible" applications would never be tested for handling them safely in any case. (5)  If an IS happened it may be the case that the entire process is dangerously corrupted and couldn't be trusted to do much error handling anyway.

So that's all well and good and suggests for abort() being the only reasonable way to handle an IS when detected.  However, even abort() may not be the safest thing for any particular application to do. For some device abort() might infinite loop and what we need to do is trigger a hardware reset. Aborting on a failure might, in some case, just end up bricking the device-- perhaps the IS failure should cause the device to restart with a null config.  It might amplify a benign bug in an optional component into a massive DOS, etc.  So we offer a callback to allow custom handling of these states.   The reason that we don't use CHECK in the library code is because we instead use error callbacks to allow that customization.

But what happens if the IS prevents us from getting access to the callback? E.g. what if our context is uninitialized or the callback pointer is null?   That is a case where using directly calling a default error handler is justified:  silently proceeding in a known corrupted state is still something we don't want to do. But we'd want to minimize the number of locations that do that and document, so that specialized applications that really want to be unconditionally abort() free, can go and make it happen and handle the cases that can't be addressed via a callback in whatever way makes sense for them. (E.g. by replacing the default handler with one suitable for their device).

Right now we have some cases where we can't use the callback where use the default handler, and a bunch of other places where we are essentially relying on a null pointer dereference to work well enough (e.g. a dereference of ctx) for the purpose of stopping execution.  Probably more of these places should be turned into explicitly calling the default handler because on some systems (esp embedded ones) a null pointer dereference will not crash.

CHECK is used in the tests because we can do whatever we want in the tests... and aborting on test failure is fine and generally interacts well with dynamic instrumentation tools like valgrind/afl/etc.

real-or-random:
fixed (and improved the comment about a single malloc too).

I can squash if someone else wants to have a closer look.

real-or-random:
Okay, I changed my mind and squashed.

sipa:
utACK, I really like the approach used to represent context objects now as a single allocated blob.

real-or-random:
Addressed @sipa's comment

real-or-random:
(Closed and opened to trigger travis rebuild)

elichai:
This enables using the library without it calling any mallocs, right?
If so then this would be awesome to `no-std` environments :)

real-or-random:
> This enables using the library without it calling any mallocs, right?
> If so then this would be awesome for `no-std` environments :)

Correct.

sipa:
@gmaxwell Anything that's left to address here?

gmaxwell:
ACK.  The documentation for this however need some guidance on the lifetime management of the caller provided memory (it could be done in another PR but it must be done).

It should make clear that the caller is obligated to make sure the buffer it provides lives at least as long as the ctx object and while the context exists the buffer is the exclusive property of that context (e.g. you cannot move it, you cannot create two contexts using the same buffer)... and that after destroying the context the caller it's up to the caller to free the buffer (if required).

[Basically don't assume that the user has a feel for how memory management works... everything that I just described is obvious if you have any idea of how it would be implemented, but we can't assume that the user has given it that much thought... or they might be use to programming languages that manage lifetimes for them]

We might also want to include a note that it's easier and less failure prone to use the normal functions in environments where dynamic allocation isn't a problem.

real-or-random:
@gmaxwell I added a commit to explain all of this. Can you re-ACK if it's okay?

elichai:
Any updates here? both this and #595 are what's missing to use this library in a `no-std` environment

real-or-random:
I pushed a fixup because more warnings appeared with gcc and ARM as target. I can squash when you looked at the diff / whenever you want.

gmaxwell:
ACK the fixups ack the text.

real-or-random:
squashed and rebased

PR #567, opened 2018-10-23
Correct order of libs returned on pkg-config --libs --static libsecp2…

pmienk:
…56k1 call.


pmienk:
Addresses issue #569.

gmaxwell:
utACK

real-or-random:
utACK

PR #558, opened 2018-09-25
 Add schnorrsig module which implements BIP-340 compliant signatures

jonasnick:
This PR implements signing, verification and batch verification as described in [BIP-340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki) in an experimental module named `schnorrsig`. It includes the test vectors and a benchmarking tool. 
This PR also adds a module `extrakeys` that allows [BIP-341](https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki)-style key tweaking.

(Adding ChaCha20 as a CSPRNG and batch verification was moved to PR #760).

In order to enable the module run `./configure` with `--enable-experimental --enable-module-schnorrsig`.

Based on apoelstra's work.

PR #568, opened 2018-10-24
Fix integer overflow in ecmult_multi_var when n is large

jonasnick:
Without this PR ecmult_multi could return wrong results. If the number of points `n` is large enough then some or all multiplications could be skipped or the function could end up in an infinite loop. This PR adds two checks to prevent `n` from wrapping around.

### DIFF
--- a/src/ecmult_impl.h
+++ b/src/ecmult_impl.h
@@ -1111,12 +1111,31 @@ static int secp256k1_ecmult_multi_simple_var(const secp256k1_ecmult_context *ctx
     return 1;
 }
 
+/* Compute the number of batches and the batch size given the maximum batch size and the
+ * total number of points */
+static int secp256k1_ecmult_multi_batch_size_helper(size_t *n_batches, size_t *n_batch_points, size_t max_n_batch_points, size_t n) {
+    if (max_n_batch_points == 0) {
+        return 0;
+    }
+    if (max_n_batch_points > ECMULT_MAX_POINTS_PER_BATCH) {
+        max_n_batch_points = ECMULT_MAX_POINTS_PER_BATCH;
+    }
+    if (n == 0) {
+        *n_batches = 0;
+        *n_batch_points = 0;
+        return 1;
+    }
+    /* Compute ceil(n/max_n_batch_points) and ceil(n/n_batches) */
+    *n_batches = 1 + (n - 1) / max_n_batch_points;
+    *n_batch_points = 1 + (n - 1) / *n_batches;
+    return 1;
+}
+
 typedef int (*secp256k1_ecmult_multi_func)(const secp256k1_ecmult_context*, secp256k1_scratch*, secp256k1_gej*, const secp256k1_scalar*, secp256k1_ecmult_multi_callback cb, void*, size_t);
 static int secp256k1_ecmult_multi_var(const secp256k1_ecmult_context *ctx, secp256k1_scratch *scratch, secp256k1_gej *r, const secp256k1_scalar *inp_g_sc, secp256k1_ecmult_multi_callback cb, void *cbdata, size_t n) {
     size_t i;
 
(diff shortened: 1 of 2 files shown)


apoelstra:
ACK

src/ecmult_impl.h
```diff
@@ -972,12 +972,32 @@ static size_t secp256k1_pippenger_max_points(secp256k1_scratch *scratch) {
     return res;
 }
 
+/* Compute the number and size of batches given the maximum number of points in a batch and the
+ * total number of points */
+static int secp256k1_ecmult_multi_batch_size_helper(size_t *n_batches, size_t *n_batch_points, size_t max_points, size_t n) {
+    if (max_points == 0) {
+        return 0;
+    } else if (max_points > ECMULT_MAX_POINTS_PER_BATCH) {
```
dcousens: unnecessary `else`?

real-or-random:
Oh I didn't want to bother you with nits only. I assumed that the missing parentheses are real bugs but no, we deal with unsigned integers. :)
ACK

src/ecmult_impl.h
```diff
@@ -972,12 +972,33 @@ static size_t secp256k1_pippenger_max_points(secp256k1_scratch *scratch) {
...
+ * total number of points */
+static int secp256k1_ecmult_multi_batch_size_helper(size_t *n_batches, size_t *n_batch_points, size_t max_points, size_t n) {
+    if (max_points == 0) {
+        return 0;
+    }
+    if (max_points > ECMULT_MAX_POINTS_PER_BATCH) {
+        max_points = ECMULT_MAX_POINTS_PER_BATCH;
+    }
+    /* Check overflow */
+    if (n > (size_t)-1 - (max_points-1)) {
```
gmaxwell: Unless we have a reason, I think we'd prefer to use stdint SIZE_MAX here.

real-or-random: SIZE_MAX requires C99 as far as I remember.

gmaxwell: @real-or-random  It is an stdint.h type, like uint32_t.   The codebase uses stdint.h ( https://github.com/bitcoin-core/secp256k1/blob/master/src/util.h#L15)  types everywhere.  In practice every compiler someone would want to use has this header (even non C99 compilers), but if someone did have a compiler without it, they'd have to provide a compatibility header.

real-or-random: Okay, I see. It's equivalent but yes, SIZE_MAX is obviously much nicer to read. I was the one who suggested `size_t(-1)` to @jonasnick in the Schnorr signature PR: https://github.com/bitcoin-core/secp256k1/pull/558#pullrequestreview-165077679  Maybe that's why @apoelstra used it in #557 too...

jonasnick:
Replaced `(size_t)-1` with `MAX_SIZE`

gmaxwell:
ACK

jimpo:
This seems fine, though you could also avoid the overflow checks entirely with

```c
n_batches = 1 + (n-1)/max_points;
n_batch_points = 1 + (n-1)/n_batches;
```

This should work in all cases since `n` is guaranteed to be `> 0`.

jonasnick:
I rewrote this PR with @jimpo's suggestion. I think this is better because the previous code was complex enough that we missed a potential bug [0] and the new code doesn't have unnecessary return 0's. Because `n = SIZE_MAX` is now actually possible of  I had to rewrite the tests to directly test the batch_size_helper.

[0] `if (n_batches == 0 || n > SIZE_MAX - (*n_batches-1)) {` (first asterisk missing)

real-or-random:
ACK

jimpo:
utACK

laanwj:
utACK c37e53d12b57a9f3ff248f1f0664872500cfd165, changes look correct to me

gmaxwell:
The addition of the trivial ecmult_multi algorithm in pr580 unfortunately created a trivial merge conflict for this PR, sorry about that. I am ACK on the PR, but need the rebase.

gmaxwell:
If you'd prefer I can make a fixup commit, whatever you'd be happiest with.

gmaxwell:
ACK

PR #571, opened 2018-10-31
Fix autoconf check for __builtin_expect

real-or-random:
Calling the compiler is not enough to check whether a function exists
because the compiler will assume an implicit declaration if the
function does not exist. We need the linker to check if the function
really exists.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -154,7 +154,7 @@ AC_ARG_WITH([asm], [AS_HELP_STRING([--with-asm=x86_64|arm|no|auto]
 AC_CHECK_TYPES([__int128])
 
 AC_MSG_CHECKING([for __builtin_expect])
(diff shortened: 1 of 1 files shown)


sipa:
@real-or-random Eh, `__builtin_expect` is expected to be a builtin for which no runtime symbol exists (you can't write `&__builtin_expect` for example), so if it compiles, we should be fine.

No objection to the change though, it's more obviously correct, but I don't think it's needed.

gmaxwell:
Or we could stop using autoconf for this and switch to compiler version sniffing,  gnupreq 3 is sufficient for expect.

real-or-random:
I don't know what gnupreq is but clang has __builtin_expect, too.

sipa:
@real-or-random The code here: https://github.com/sipa/minisketch/blob/master/src/util.h#L10L23 works on clang too, I think.

real-or-random:
Closed,  #579 is better

PR #572, opened 2018-11-01
Add anti nonce-sidechannel protocol to schnorrsigs

jonasnick:
Based on #588.

Adding anti nonce-sidechannel functionality was a request in the schnorrsig module PR. The use case is to prevent a signing device from exfiltrating the secret signing keys through biased signature nonces, for example after a bad firmware update. It uses sign-to-contract commitments to host supplied randomness which forces the signer to randomize the nonce.

In order to add more sign-to-contract commitments the host can create a merkle tree of data to be committed and randomness and send that to the signing device as "host randomness". 

See https://github.com/bitcoin-core/secp256k1/pull/572/files#diff-b19c5ee427283d4d82bc5beb4e2f4777R59 for the workflow.


jonasnick:
rebased on #558

gmaxwell:
I believe secp256k1_schnorrsig_anti_nonce_sidechan_client_commit is vulnerable to key thefts.

Host asks the client to sign the same message twice using different host randomnesses, the client will generate the same nonce as the host commitment is not part of the input to the derandomized RNG.  At the end, the host knows two signatures with the same key and message with a known offset between their nonces.  From this, I believe they can interpolate the secret key.

include/secp256k1_schnorrsig.h
```diff
@@ -0,0 +1,210 @@
...
+
+/** Verify a sign-to-contract commitment.
+ *
+ *  Returns: 1: the signature contains a commitment to data32
+ *           0: incorrect opening
+ *  Args:    ctx: a secp256k1 context object, initialized for verification.
+ *  In:      sig: the signature containing the sign-to-contract commitment (cannot be NULL)
+ *        data32: the 32-byte data that was committed to (cannot be NULL)
+ *       original_nonce: pointer to the original_nonce created when signing (cannot be NULL)
+ * negated_nonce: integer indicating if signing algorithm negated the nonce (can be NULL)
```
instagibbs: >(can be NULL)

Nah

Also specify what the integer must be to be considered true/false

jonasnick:
Closing in favor of #590

PR #578, opened 2018-12-14
Avoid implementation-defined and undefined behavior when dealing with sizes

real-or-random:
This is a result of auditing the code for overflow issues at random places. None of this is critical but I think all of it should be fixed.

I know this touches "red" code. I double-checked and triple-checked this but I can understand if some of the changes are not desirable because they change well-tested code.

Best reviewed in individual commits.

### DIFF
--- a/contrib/lax_der_parsing.c
+++ b/contrib/lax_der_parsing.c
@@ -32,7 +32,7 @@ int ecdsa_signature_parse_der_lax(const secp256k1_context* ctx, secp256k1_ecdsa_
     lenbyte = input[pos++];
     if (lenbyte & 0x80) {
         lenbyte -= 0x80;
-        if (pos + lenbyte > inputlen) {
+        if (lenbyte > inputlen - pos) {
             return 0;
         }
         pos += lenbyte;
@@ -51,7 +51,7 @@ int ecdsa_signature_parse_der_lax(const secp256k1_context* ctx, secp256k1_ecdsa_
     lenbyte = input[pos++];
     if (lenbyte & 0x80) {
         lenbyte -= 0x80;
-        if (pos + lenbyte > inputlen) {
+        if (lenbyte > inputlen - pos) {
             return 0;
         }
         while (lenbyte > 0 && input[pos] == 0) {
@@ -89,7 +89,7 @@ int ecdsa_signature_parse_der_lax(const secp256k1_context* ctx, secp256k1_ecdsa_
     lenbyte = input[pos++];
     if (lenbyte & 0x80) {
         lenbyte -= 0x80;
-        if (pos + lenbyte > inputlen) {
+        if (lenbyte > inputlen - pos) {
             return 0;
         }
         while (lenbyte > 0 && input[pos] == 0) {
--- a/src/ecdsa_impl.h
+++ b/src/ecdsa_impl.h
@@ -128,8 +128,11 @@ static int secp256k1_der_parse_integer(secp256k1_scalar *r, const unsigned char 
         /* Negative. */
         overflow = 1;
     }
-    while (rlen > 0 && **sig == 0) {
-        /* Skip leading zero bytes */
+    /* There is at most one leading zero byte:
+     * if there were two leading zero bytes, we would have failed and returned 0
+     * because of excessive 0x00 padding already. */
+    if (rlen > 0 && **sig == 0) {
+        /* Skip leading zero byte */
         rlen--;
         (*sig)++;
     }
--- a/src/ecdsa_impl.h
+++ b/src/ecdsa_impl.h
@@ -157,12 +157,8 @@ static int secp256k1_ecdsa_sig_parse(secp256k1_scalar *rr, secp256k1_scalar *rs,
     if (secp256k1_der_read_len(&rlen, &sig, sigend) == 0) {
         return 0;
     }
-    if (sig + rlen > sigend) {
-        /* Tuple exceeds bounds */
-        return 0;
-    }
-    if (sig + rlen != sigend) {
-        /* Garbage after tuple. */
+    if (rlen != (size_t)(sigend - sig)) {
+        /* Tuple exceeds bounds or garage after tuple. */
         return 0;
     }
 
--- a/src/ecdsa_impl.h
+++ b/src/ecdsa_impl.h
@@ -46,68 +46,73 @@ static const secp256k1_fe secp256k1_ecdsa_const_p_minus_order = SECP256K1_FE_CON
     0, 0, 0, 1, 0x45512319UL, 0x50B75FC4UL, 0x402DA172UL, 0x2FC9BAEEUL
 );
 
-static int secp256k1_der_read_len(const unsigned char **sigp, const unsigned char *sigend) {
-    int lenleft, b1;
-    size_t ret = 0;
+static int secp256k1_der_read_len(size_t *len, const unsigned char **sigp, const unsigned char *sigend) {
+    size_t lenleft;
+    unsigned char b1;
+    VERIFY_CHECK(len != NULL);
+    *len = 0;
     if (*sigp >= sigend) {
-        return -1;
+        return 0;
     }
     b1 = *((*sigp)++);
     if (b1 == 0xFF) {
         /* X.690-0207 8.1.3.5.c the value 0xFF shall not be used. */
-        return -1;
+        return 0;
     }
     if ((b1 & 0x80) == 0) {
         /* X.690-0207 8.1.3.4 short form length octets */
-        return b1;
+        *len = b1;
+        return 1;
     }
     if (b1 == 0x80) {
         /* Indefinite length is not allowed in DER. */
-        return -1;
+        return 0;
     }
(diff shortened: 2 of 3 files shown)


gmaxwell:
Aside, haven't reviewed yet... but your first commit messages appear to be exactly the sort of thing I want to see on a PR like that--  it gives a trace of the execution it's concerned with.   I want to merge it just so I can point people in the future at what this sort of change should look like.

src/ecdsa_impl.h
```diff
@@ -46,68 +46,71 @@ static const secp256k1_fe secp256k1_ecdsa_const_p_minus_order = SECP256K1_FE_CON
     0, 0, 0, 1, 0x45512319UL, 0x50B75FC4UL, 0x402DA172UL, 0x2FC9BAEEUL
 );
 
-static int secp256k1_der_read_len(const unsigned char **sigp, const unsigned char *sigend) {
-    int lenleft, b1;
-    size_t ret = 0;
+static int secp256k1_der_read_len(size_t *len, const unsigned char **sigp, const unsigned char *sigend) {
```
laanwj: Does this need a not-NULL check/assertion for `len`?

src/ecdsa_impl.h
```diff
@@ -46,68 +46,71 @@ static const secp256k1_fe secp256k1_ecdsa_const_p_minus_order = SECP256K1_FE_CON
...
-        return -1;
+        return 0;
     }
     while (lenleft > 0) {
-        ret = (ret << 8) | **sigp;
-        if (ret + lenleft > (size_t)(sigend - *sigp)) {
-            /* Result exceeds the length of the passed array. */
-            return -1;
-        }
+        *len = (*len << 8) | **sigp;
```
laanwj: No initial value is set for `*len`?

real-or-random: Good catch

src/ecdsa_impl.h
real-or-random: sounds like a good idea

contrib/lax_der_parsing.c
```diff
@@ -51,7 +51,7 @@ int ecdsa_signature_parse_der_lax(const secp256k1_context* ctx, secp256k1_ecdsa_
     lenbyte = input[pos++];
     if (lenbyte & 0x80) {
         lenbyte -= 0x80;
-        if (pos + lenbyte > inputlen) {
+        if (lenbyte > inputlen - pos) {
```
luke-jr: `pos` could possibly be larger than `inputlen` here, since it was just post-incremented on line 51 without a check.

contrib/lax_der_parsing.c
```diff
@@ -89,7 +89,7 @@ int ecdsa_signature_parse_der_lax(const secp256k1_context* ctx, secp256k1_ecdsa_
...
```
luke-jr: Same here, `pos` is incremented without check at line 89.

src/ecdsa_impl.h
```diff
@@ -66,7 +66,7 @@ static int secp256k1_der_read_len(const unsigned char **sigp, const unsigned cha
         return -1;
     }
     /* X.690-207 8.1.3.5 long form length octets */
-    lenleft = b1 & 0x7F;
+    lenleft = b1 & 0x7F; /* lenleft is at least 1 */
```
luke-jr: ACK, `b1` comes from an unsigned char, and the function returns before this point if `b1` < 0x80 (ruling out a value of 0) or `b1` == 0x80 (the only value that would yield a 0 from this mask)

src/ecdsa_impl.h
```diff
@@ -82,13 +82,13 @@ static int secp256k1_der_read_len(const unsigned char **sigp, const unsigned cha
     }
     while (lenleft > 0) {
         ret = (ret << 8) | **sigp;
-        if (ret + lenleft > (size_t)(sigend - *sigp)) {
```
luke-jr: This looks like it was wrong. If the remaining digits are all zero, the final size is just `ret`, not `ret + ...`.

If my understanding is correct, is there a possibility fixing this bug could create a a consensus incompatibility?

src/ecdsa_impl.h
```diff
@@ -126,8 +126,11 @@ static int secp256k1_der_parse_integer(secp256k1_scalar *r, const unsigned char
         /* Negative. */
         overflow = 1;
     }
-    while (rlen > 0 && **sig == 0) {
-        /* Skip leading zero bytes */
+    /* There is at most one leading zero byte:
+     * if there were two leading zero bytes, we would have failed and returned 0
+     * because of excessive 0x00 padding already. */
+    if (rlen > 0 && **sig == 0) {
+        /* Skip leading zero byte */
```
luke-jr: utACK

src/ecdsa_impl.h
```diff
@@ -155,12 +155,8 @@ static int secp256k1_ecdsa_sig_parse(secp256k1_scalar *rr, secp256k1_scalar *rs,
...
         return 0;
     }
-    if (sig + rlen > sigend) {
-        /* Tuple exceeds bounds */
-        return 0;
-    }
-    if (sig + rlen != sigend) {
-        /* Garbage after tuple. */
+    if (rlen != (size_t)(sigend - sig)) {
+        /* Tuple exceeds bounds or garage after tuple. */
```
luke-jr: utACK

src/hash_impl.h
```diff
@@ -131,7 +131,8 @@ static void secp256k1_sha256_transform(uint32_t* s, const uint32_t* chunk) {
 static void secp256k1_sha256_write(secp256k1_sha256 *hash, const unsigned char *data, size_t len) {
     size_t bufsize = hash->bytes & 0x3F;
     hash->bytes += len;
-    while (bufsize + len >= 64) {
+    VERIFY_CHECK(hash->bytes >= len);
+    while (len >= 64 - bufsize) {
```
luke-jr: utACK

src/ecdsa_impl.h
luke-jr: n.b. This returns via output pointer instead of return value, to prevent accidentally casting to int (or other wrong types) implicitly.

src/ecdsa_impl.h
real-or-random: I don't understand your sentence about the remaining digits but I think your understanding is wrong. It took me a time to understand what this code actually does. Checking `ret > (size_t)(sigend - *sigp)` instead would be enough but checking `ret + lenleft > (size_t)(sigend - *sigp)` gives you a (not very useful) earlier return in certain cases. My understanding is that this function was correct because this function still has to read `lenleft` bytes and then the caller will have to read at least the current value of `ret`. So if the sum of these two values exceeds the remaining bytes, it is correct to return -1 early. And if we don't return -1 early, then in the last iteration we have `lenleft == 0`, and then we check with effectively only `ret` on the left-hand side, which is the correct check.

It should be noted that none of this is code is consensus-critical (for Bitcoin) because Bitcoin relies on lax DER parsing function (as implemented in the `contrib` directory of secp256k1). The change may affect other projects though. AFAIK Zcash uses the functions in the PR here.

real-or-random: > This looks like it was wrong. If the remaining digits are all zero, the final size is just `ret`, not `ret + ...`.

Ah I think now I know what you are saying. It took me also a while to understand that code. The purpose of `ret  + lenleft` instead of `ret` is a (not really useful) early return: this function needs to read `lenleft` more bytes (or actually read `lenleft - 1` but we didn't increment `sigp*` yet), and the caller needs to read at least `ret` more bytes. So if `lenleft + ret` is more bytes than available, it is indeed correct to return -1 early.

contrib/lax_der_parsing.c
real-or-random: Right before line 51, we have `pos < inputlen`  since we have `pos != input` due to the check in line 48.
So after the increment in line 51, we have `pos <= inputlen`

contrib/lax_der_parsing.c
real-or-random: Same here too

real-or-random:
Addressed the comments

gmaxwell:
Can you squash?

real-or-random:
squashed

src/ecdsa_impl.h
```diff
@@ -144,18 +152,16 @@ static int secp256k1_der_parse_integer(secp256k1_scalar *r, const unsigned char
...
-    rlen = secp256k1_der_read_len(&sig, sigend);
-    if (rlen < 0 || sig + rlen > sigend) {
-        /* Tuple exceeds bounds */
+    if (secp256k1_der_read_len(&rlen, &sig, sigend) == 0) {
         return 0;
     }
-    if (sig + rlen != sigend) {
-        /* Garbage after tuple. */
+    if (rlen != (size_t)(sigend - sig)) {
+        /* Tuple exceeds bounds or garage after tuple. */
```
real-or-random: I admit that "garage after tuple" seems to be an unrealistic concern. 🔧🚗

PR #579, opened 2018-12-17
Use __GNUC_PREREQ for detecting __builtin_expect

real-or-random:


sipa:
utACK c663397f46152e96c548ba392858c730e132dd7a

jonasnick:
utACK c663397f46152e96c548ba392858c730e132dd7a

Replacing `#define EXPECT(x,c) __builtin_expect((x),(c))` with `#define EXPECT(x,c) (x)` works too.

PR #583, opened 2018-12-26
JNI: fix use sig array

fingera:


fingera:
https://github.com/bitcoin-core/secp256k1/blob/e34ceb333b1c0e6f4115ecbb80c632ac1042fa49/src/java/org/bitcoin/NativeSecp256k1Test.java#L137-L159
this is pub key tweak

gmaxwell:
I hate it when I accidentally press the [72] key. 

This change looks fine, but please don't make inaccurate commit messages.

gmaxwell:
Thanks for the update. Sorry for the slow response. While you're fixing this, can you also change the code to make use of the return values and not just silence ignoring them?

sipa:
ACK 74e2dbd68e07f752ac326a578e3071f9efa55e55. This is clearly an improvement.

Feel free to address to comment, but the serialization should never fail as the output array is large enough for any ECDSA signature.

real-or-random:
ACK 74e2dbd68e07f752ac326a578e3071f9efa55e55 I've read the code but haven't tested it

PR #587, opened 2019-01-27
Make randomization of a non-signing context a noop

### LINKED ISSUE #573: illegal-callback called with message that says "secp256k1_ecmult_context_is_built(&ctx->ecmult_ctx)"
trying to wrap secp256k1 in lisp and I am getting this message when trying to derive a public child key of a public parent key.
my code looks something like this(I have used cl-autowrap, have not given everthing nice names yet)

```
(with-context (ctx secp256k1.ffi:+secp256k1-context-sign+)
    (secp256k1.ffi:secp256k1-context-set-illegal-callback ctx
                                                          (autowrap:callback 'illegal-callback)
                                                          'nil)
    (autowrap:with-many-alloc ((pubkey 'secp256k1.ffi:secp256k1-pubkey)
                               (serialized ':unsigned-char 65)
                               (size-serialized ':unsigned-long 1))
      (unless (secp256k1.ffi:secp256k1-ec-pubkey-parse ctx pubkey key_ (length key))
        (error "when parsing pubkey"))
      (unless (secp256k1.ffi:secp256k1-ec-pubkey-tweak-add ctx pubkey hash_)
        (error "error when tweaking pubkey"))
      (setf (autowrap:c-aref size-serialized 0 :unsigned-long) 65)
      (unless (secp256k1.ffi:secp256k1-ec-pubkey-serialize ctx
                                                           serialized
                                                           size-serialized
                                                           pubkey
                                                           secp256k1.ffi:+secp256k1-ec-compressed+)
        (error "could not serialize new public key"))
      (values (cffi:foreign-array-to-lisp serialized `(:array :unsigned-char
                                                              ,(autowrap:c-aref size-serialized
                                                                                0
                                                                                :unsigned-long)))
              (subseq hash 32 64))))
```

i get the call to my illegal-callback after the call to tweak-add function and before the call to the ec-pubkey-serialize.

Questions:
what does this message mean? do i need to configure the context in some specific way? do I need to have a scratch-space set up? i have tried with one. 

I am running my lisp code in SBCL.

real-or-random:
Before this commit secp256k1_context_randomize called illegal_callback
when called on a context not initialized for signing. This is not
documented. Moreover, it is not desirable because non-signing contexts
may use randomization in the future.

This commit makes secp256k1_context_randomize a noop in this case. This
is safe because the context cannot be used for signing anyway.

This fixes #573 and it fixes rust-bitcoin/rust-secp256k1#82.

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -615,7 +615,7 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_tweak_mul(
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
 
 /** Updates the context randomization to protect against side-channel leakage.
- *  Returns: 1: randomization successfully updated
+ *  Returns: 1: randomization successfully updated or nothing to randomize
  *           0: error
  *  Args:    ctx:       pointer to a context object (cannot be NULL)
  *  In:      seed32:    pointer to a 32-byte random seed (NULL resets to initial state)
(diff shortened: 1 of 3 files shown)


gmaxwell:
I guess this seems reasonable to me. The only  concern I have is minor: if someone has a signing and verifying context, this makes it easier to randomize the wrong one.

sipa:
utACK 6198375218b8132f016b701ef049fb295ca28c95

jonasnick:
Was about to NACK this PR because being explicit is better in general. However, if verification contexts can be randomized in the future, it's better to start allow to randomizing everything always.

ACK 6198375218b8132f016b701ef049fb295ca28c95

gmaxwell:
ACK

PR #588, opened 2019-01-29
Allow sign-to-contract commitments in schnorrsigs

jonasnick:
This PR implements creation and verification of sign-to-contract commitments in the
schnorrsig module. The following snippet for example puts a commitment to `data32`
into the (Schnorr) signature.

```
/* Signer */
secp256k1_s2c_commit_context s2c_ctx;
secp256k1_s2c_commit_context_create(ctx, &s2c_ctx, data32);
secp256k1_schnorrsig_sign(sign, &sig, &nonce_is_negated, msg, sk1, NULL, &s2c_ctx);
secp256k1_s2c_commit_get_original_nonce(ctx, &s2c_original_nonce, &s2c_ctx);

/* Verifier */
secp256k1_schnorrsig_verify_s2c_commit(ctx, &sig, data32, &s2c_original_nonce, nonce_is_negated);
```

This PR is based on #558. The commits in this PR was previously included in #572.
The tests should provide full coverage of the reachable lines.

There is also functionality for doing pay-to-contract style commitments
(ec_commit), but it's not exposed yet. If it will be in the future it would
make sense to add an optional argument with a SHA midstate to allow hashing
arbitrary version prefixes before hashing the public key (which would be useful
in taproot for example).

Sign-to-contract and right now only work with schnorrsig_sign and
nonce_function_bipschnorr. It should be straightforward to add support for
ECDSA too in the future.

One thing to keep in mind is that for some users moving from ecdsa to
schnorrsigs requires reading the documentation carefully. There is undocumented
functionality in nonce_function_rfc6979 where the additional nonce data is
hashed into the secret. If someone uses that and then call schnorrsig_sign with
similar nonce data this results in a segmentation fault because
nonce_function_bipschnorr which is used by default in schnorrsig_sign expects
an s2c_commit_context object as nonce data and writes back to it.


src/secp256k1.c
```diff
@@ -610,6 +610,103 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
     return 1;
 }
 
+/* Compute an ec commitment tweak as hash(pubkey, data). */
+int secp256k1_ec_commit_tweak(const secp256k1_context *ctx, unsigned char *tweak32, const secp256k1_pubkey *pubkey, const unsigned char *data, size_t data_size) {
```
apoelstra: Should be `static`

gmaxwell:
If you want to harden against misuse of the nonce function arguments, you could add a constant uint64 as the first member of the s2c_commit_context  and test it.

Generally, I'd like to see us do that with all objects that we allow users to pass around: set a magic value when they're constructed, zero it when they're destroyed.  Check for it before other accesses... exception being purely internal objects or elements where a caller might plausibly have thousands of them in existence at once (such that the overhead becomes a concern).

src/secp256k1.c
```diff
@@ -610,6 +610,103 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
...
+        return 0;
+    }
+
+    /* Return commitment == commitment_tmp */
+    secp256k1_gej_set_infinity(&pj);
+    secp256k1_pubkey_load(ctx, &p, &commitment_tmp);
+    secp256k1_gej_add_ge(&pj, &pj, &p);
+    secp256k1_pubkey_load(ctx, &p, commitment);
+    secp256k1_ge_neg(&p, &p);
+    secp256k1_gej_add_ge(&pj, &pj, &p);
```
real-or-random: Maybe it's a good idea to implement an gej_cmp or even pubkey_cmp... No idea.

include/secp256k1.h
```diff
@@ -80,6 +80,26 @@ typedef struct {
...
+ *  secp256k1_s2c_commit_context_create the context can be given to a signing
+ *  algorithm via the nonce data argument.
+ *
+ *  This structure is not opaque, but it is strongly discouraged to copy it or read
+ *  or write to it directly. Use the secp256k1_s2c_commit_* instead to access a
+ *  sign-to-contract context.
+ *
+ *  The exact representation of data inside is implementation defined and not
+ *  guaranteed to be portable between different platforms or versions. It is however
+ *  guaranteed to be 128 bytes in size, and can be safely copied/moved.
```
real-or-random: one paragraph above, you say that copying is discouraged

src/modules/schnorrsig/tests_impl.h
```diff
@@ -713,15 +741,59 @@ void test_schnorrsig_sign_verify(secp256k1_scratch_space *scratch) {
...
+        data32_tmp[31] ^= 1;
+        CHECK(secp256k1_schnorrsig_verify_s2c_commit(ctx, &sig, data32_tmp, &s2c_original_nonce, nonce_is_negated) == 0);
+    }
+    {
+        /* verify_s2c_commit fails if signature does not commit to data */
+        secp256k1_schnorrsig sig_tmp;
+        sig_tmp = sig;
+        secp256k1_rand256(&sig_tmp.data[0]);
+        CHECK(secp256k1_schnorrsig_verify_s2c_commit(ctx, &sig_tmp, data32, &s2c_original_nonce, nonce_is_negated) == 0);
+    }
```
real-or-random: there should be a test that verification fails when nonce_is_negated is wrong

src/secp256k1.c
```diff
@@ -610,6 +610,103 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
...
+/* Compute an ec commitment tweak as hash(pubkey, data). */
+int secp256k1_ec_commit_tweak(const secp256k1_context *ctx, unsigned char *tweak32, const secp256k1_pubkey *pubkey, const unsigned char *data, size_t data_size) {
+    secp256k1_ge p;
+    unsigned char rbuf[33];
+    size_t rbuf_size = sizeof(rbuf);
+    secp256k1_sha256 sha;
+
+    if (data_size == 0) {
+        /* That's probably not what the caller wanted */
+        return 0;
```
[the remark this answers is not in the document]
jonasnick: VERIFY_CHECK only works in tests. ARG_CHECK crashes the caller's program. Not sure if that would be better given that there doesn't really seem to be precedent for aborting in such cases in libsecp. Also I don't see a good reason for aborting as this condition can be trivially recovered from (by just returning 0).

src/secp256k1.c
jonasnick: yeah, good idea. Maybe not in this PR.

include/secp256k1.h
jonasnick: good catch. I think there's no issue with copying that structure and I think I just copied this sentence from the MuSig session without thinking about it.

src/modules/schnorrsig/tests_impl.h
jonasnick: fixed

include/secp256k1.h

src/tests.c
```diff
@@ -2339,6 +2339,110 @@ void run_ec_combine(void) {
...
+    secp256k1_pubkey commitment;
+    unsigned char data[32];
+
+    /* Create random keypair */
+    secp256k1_rand256(seckey);
+    CHECK(secp256k1_ec_pubkey_create(ctx, &pubkey, seckey));
+    /* Create random data */
+    {
+        secp256k1_scalar d;
+        random_scalar_order_test(&d);
```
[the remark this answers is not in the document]

jonasnick:
I added a commit that adds a magic number to the sign to contract context struct.

jonasnick:
squashed

src/secp256k1.c
```diff
@@ -588,10 +589,189 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
...
+
+static uint64_t s2c_commit_context_magic = 0xd5bafd089f7e1c63;
+int secp256k1_s2c_commit_context_create(secp256k1_context *ctx, secp256k1_s2c_commit_context *s2c_ctx, const unsigned char *data32) {
+    secp256k1_sha256 sha;
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(s2c_ctx != NULL);
+    ARG_CHECK(data32 != NULL);
+
+    memcpy(s2c_ctx->magic, &s2c_commit_context_magic, sizeof(s2c_ctx->magic));
```
elichai: Probably me just not familiar enough with C but why copy from a `uint64_t` to a `unsigned char[8]` instead of just putting a `uint64_t` in the struct?

src/secp256k1.c
```diff
@@ -588,10 +589,189 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
...
+ * commitments.
+ */
+static int secp256k1_nonce_function_bipschnorr(const secp256k1_context *ctx, unsigned char *nonce32, const unsigned char *msg32, const unsigned char *key32, const unsigned char *algo16, void *data, unsigned int counter) {
+    if (!secp256k1_nonce_function_bipschnorr_no_s2c_tweak(ctx, nonce32, msg32, key32, algo16, data, counter)) {
+        return 0;
+    }
+    if (data != NULL) {
+        /* Create a sign-to-contract commitment by setting nonce32 <- nonce32 +
+         * hash(nonce32*G, s2c_ctx->data) */
+        secp256k1_s2c_commit_context *s2c_ctx = (secp256k1_s2c_commit_context *)data;
```
elichai: I think this is exactly the place to verify the magic number and fail if not right.

include/secp256k1.h
```diff
@@ -80,9 +80,34 @@ typedef struct {
...
+ *  guaranteed to be 136 bytes in size, and can be safely copied/moved.
+ */
+typedef struct {
+    /* magic is set during initialization. It allows functions casting to
+     * s2c_commit_contexts from a void pointer to check if they actually got an
+     * s2c_commit_context and if it has been initialized. */
+    unsigned char magic[8];
+    unsigned char data[32];
+    unsigned char data_hash[32];
+    secp256k1_pubkey original_pubnonce;
```
elichai: should the "public nonce" really be represented as `pubkey`? for me `ge` or `gej` would be more readable, but they're probably not exported....
(I got confused while diving into `secp256k1_ec_commit_seckey` and `secp256k1_ec_commit_tweak` about the whole public key thing)

src/secp256k1.c
jonasnick: yes, good catch

include/secp256k1.h
jonasnick: `ge` or `gej` are not exported. But, hm, perhaps this should be a pubnonce type if it already confuses people.

src/secp256k1.c
jonasnick: There was some reason I don't remember to use `uint64_t`. But I think using a byte array in the struct is arbitrary... probably easier to read if replaced by `uint64_t`.

elichai: Another thing, I'm pretty sure memcpy from `uint64_t`  to `unsigned char[8]` is platform defined(BE/LE)

jonasnick: Yes, but not really an issue as the doc clearly states

> The exact representation of data inside is implementation defined and not guaranteed to be portable between different platforms or versions.

real-or-random: It doesn't matter in the end but the code is certainly easier to read if it's the same type twice (no matter what type that is)

include/secp256k1.h
real-or-random: Hm, yes. A public key is a public key, and a nonce is a different thing. (I mean we can't enforce anything in C but maybe it's helpful to use different names for the two public types at least.)

Do we have the same problem in other places too?

real-or-random: On a second thought... if we really want to optimize for a simple API, then https://github.com/bitcoin-core/secp256k1/pull/589 is the way to go. There the entire problem is eliminated by hiding from the user that a signature contains a "nonce".

elichai: We do have that there too https://github.com/bitcoin-core/secp256k1/pull/589/files?file-filters%5B%5D=.c&file-filters%5B%5D=.h&file-filters%5B%5D=.include#diff-222d6d707e263c79c462fb223b0e3ddcR98

real-or-random: yes

elichai: fair

jonasnick:
Closing in favor of #589

PR #589, opened 2019-02-14
Allow sign-to-contract commitments in schnorrsigs [̶a̶l̶t̶e̶r̶n̶a̶t̶i̶v̶e̶]̶

jonasnick:
~~This is an alternative to #588 as requested by @real-or-random which I also slightly prefer~~ (EDIT: this seems to be generally preferred now). The main difference is that the sign-to-contract commitment step happens in the signature function and not the nonce function. Also the `nonce_is_negated` argument in `schnorrsig_sign` is replaced by an `s2c_opening` object. A new argument to `schnorrsig_sign` is added called `s2c_data`. There's no need to add a context argument to nonce functions. I also added parsing and serialization for `s2c_opening`s. Manual initialization of `s2c_opening` is not necessary anymore.

Example:
```
/* Signer */
secp256k1_s2c_opening opening;
unsigned char s2c_data[32];
secp256k1_schnorrsig_sign(sign, &sig, &opening, msg, sk1, &s2c_data, NULL, NULL);

/* Verifier */
secp256k1_schnorrsig_verify_s2c_commit(ctx, &sig, s2c_data, &opening);
```


gmaxwell:
That workflow doesn't make it look as easy to implement s2c anti-sidechannel blinding, if I understand it.

jonasnick:
Hm, I've done a quick and dirty rebase of the anti-nonce-sidechannel and it seems straight forward. It's even a bit simpler on the client side.

jonasnick:
I opened PR #590 to show the anti nonce-sidechannel protocol can be built on this PR in a similar way.

real-or-random:
Concept ACK

real-or-random:
I guess this is a more general discussion because it's also in ECDSA:
What's the exact benefit of the user being able to provide a nonce function (and possibly shoot herself in the foot?) I'm not sure if there are meaningful cases,so I'm somewhat curious if anybody currently  uses this in practice with ECDSA.

gmaxwell:
@real-or-random  IIRC it's actually used by joinmarket to do ECDH with the R value for a stealth payment donation feature.   (like one of the outputs pays to  P+H(kP)G  for some pubkey P).  The payments can be found by the recipient simply scanning every signature in the chain, performing ECDH with it, and checking the outputs...

Less cosmically RFC6979 is kind of absurd and actually makes signing measurably slower compared to just a simple hash-- for applications that care about signing speed (like matt's betterhash) substituting it is probably a good idea.  Not sure who if anyone is doing that right now, but some things probably should be.  (Though those same applications should probably also do batch nonce generation and other stuff that wouldn't really be simple with the current nonce function interface)

Re: footgunnery,  I haven't yet seen anyone footgun themselves that way. Creating a working nonce function takes a fair amount of work, so I think it's likely that if anyone bothers they at least have a fighting chance to get it right.  I think the best mental model is that we're trying to defend against users that are equal parts lazy and ignorant.   Give them a nonce argument and they will set the nonce to private key xor messagehash (actual example which demonstrates lazy+ignorant),  but give them a nonce function argument and they'll leave it null unless they actually have a real reason to do otherwise. :)

real-or-random:
Okay, convinced.

> I think the best mental model is that we're trying to defend against users that are equal parts lazy and ignorant.

That's indeed a good way to think about it. 

Maybe it's still a good idea to make it more prominent in the docs that people should use the default unless they really know what they are doing.

gmaxwell:
Sounds fine to me.  I also think we could explicitly say that the nonce function isn't an interface that we particularly care about keeping stable.

src/secp256k1.c
```diff
@@ -609,6 +609,103 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
...
+}
+
+/* Compute an ec commitment as pubkey + hash(pubkey, data)*G. */
+static int secp256k1_ec_commit(const secp256k1_context* ctx, secp256k1_pubkey *commitment, const secp256k1_pubkey *pubkey, const unsigned char *data, size_t data_size) {
+    unsigned char tweak[32];
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(commitment != NULL);
+    ARG_CHECK(pubkey != NULL);
+    ARG_CHECK(data != NULL);
```
apoelstra: We generally don't have `ARG_CHECK`s in `static` functions.

apoelstra: Same below in `ec_commit_seckey`

src/secp256k1.c
```diff
@@ -609,6 +609,103 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
...
+
+    if (!secp256k1_ec_commit(ctx, &commitment_tmp, pubkey, data, data_size)) {
+        return 0;
+    }
+
+    /* Return commitment == commitment_tmp */
+    secp256k1_gej_set_infinity(&pj);
+    secp256k1_pubkey_load(ctx, &p, &commitment_tmp);
+    secp256k1_gej_add_ge(&pj, &pj, &p);
```
apoelstra: We can use the `_var` variant of this function (same with below).

include/secp256k1.h
```diff
@@ -80,6 +80,26 @@ typedef struct {
...
+/** Data structure that holds a sign-to-contract ("s2c") opening. Sign-to-contract
+ *  allows a signer to commit to some data as part of a signature. It can be used as
+ *  an Out-argument in certain signing functions.
+ *
+ *  This structure is not opaque, but it is strongly discouraged to read or write to
+ *  it directly.
+ *
+ *  The exact representation of data inside is implementation defined and not
+ *  guaranteed to be portable between different platforms or versions. It is however
+ *  guaranteed to be 73 bytes in size, and can be safely copied/moved.
```
apoelstra: This guarantee is basically impossible to make for a data structure with multiple heterogeneous fields.

apoelstra: "safely copied/moved" is OK, but the exact size is implementation dependent.

src/secp256k1.c
```diff
@@ -706,6 +706,38 @@ static int secp256k1_ec_commit_verify(const secp256k1_context* ctx, const secp25
     return secp256k1_gej_is_infinity(&pj);
 }
 
+static uint64_t s2c_opening_magic = 0x5d0520b8b7f2b168;
```
apoelstra: Using `uint64_t` here rather than `char[8]` introduces an endianness dependency. Is this on purpose?

include/secp256k1_schnorrsig.h
```diff
@@ -61,19 +61,24 @@ SECP256K1_API int secp256k1_schnorrsig_parse(
...
- *       nonce_is_negated: a pointer to an integer indicates if signing algorithm negated the
- *                nonce (can be NULL)
+ *      s2c_data: pointer to an secp256k1_s2c_opening structure which can be
+ *                NULL but is required to be not NULL if this signature creates
+ *                a sign-to-contract commitment (i.e. the `s2c_data` argument
+ *                is not NULL).
  *  In:    msg32: the 32-byte message hash being signed (cannot be NULL)
  *        seckey: pointer to a 32-byte secret key (cannot be NULL)
+ *    s2c_data32: pointer to a 32-byte data to create an optional
+ *                sign-to-contract commitment to if non-NULL (non-NULL).
```
apoelstra: `(non-NULL)` should be removed.

include/secp256k1_schnorrsig.h
```diff
@@ -61,19 +61,24 @@ SECP256K1_API int secp256k1_schnorrsig_parse(
...
  *       noncefp: pointer to a nonce generation function. If NULL, secp256k1_nonce_function_bipschnorr is used
```
apoelstra: Can you add a comment to `noncefp` saying that it has to be `NULL` (or the default) if `s2c_data` is not NULL?

apoelstra:
`valgrind ./tests 1` shows a bunch of branches on uninitialized memory.

src/secp256k1.c
jonasnick: It's not purposefully endian dependent, but purposefully uint64. Just looks nicer and the object is documented to be "implementation defined" anyway. But agree char should be better.

src/modules/schnorrsig/main_impl.h
```diff
@@ -0,0 +1,392 @@
...
+    /* Fail if the secret key is invalid. */
+    if (overflow || secp256k1_scalar_is_zero(&x)) {
+        memset(sig, 0, sizeof(*sig));
+        return 0;
+    }
+
+    secp256k1_ecmult_gen(&ctx->ecmult_gen_ctx, &pkj, &x);
+    secp256k1_ge_set_gej(&pk, &pkj);
+
+    if(s2c_data32 != NULL) {
```
real-or-random: nit: spacing

include/secp256k1.h
gmaxwell: (if it wasn't clear, due to alignment padding, which could basically be anything)

src/modules/schnorrsig/main_impl.h
jonasnick: fixed

include/secp256k1_schnorrsig.h
jonasnick: fixed

include/secp256k1_schnorrsig.h
jonasnick: fixed

src/secp256k1.c
jonasnick: fixed

src/secp256k1.c
jonasnick: fixed

src/secp256k1.c
jonasnick: thought this could be non-static soon, but removed it nonetheless for now

include/secp256k1.h
jonasnick: yeah, not sure what's got into me

src/scalar_8x32_impl.h
```diff
@@ -718,4 +720,102 @@ SECP256K1_INLINE static void secp256k1_scalar_mul_shift_var(secp256k1_scalar *r,
...
+#endif
+
+static void secp256k1_scalar_chacha20(secp256k1_scalar *r1, secp256k1_scalar *r2, const unsigned char *seed, uint64_t idx) {
+    size_t n;
+    size_t over_count = 0;
+    uint32_t seed32[8];
+    uint32_t x0, x1, x2, x3, x4, x5, x6, x7, x8, x9, x10, x11, x12, x13, x14, x15;
+    int over1, over2;
+
+    memcpy((void *) seed32, (const void *) seed, 32);
```
elichai: Shouldn't you loop over `seed32` and put it into` LE32`?(otherwise the same seed will return different result on different endianesses)

src/secp256k1.c
```diff
@@ -588,10 +611,129 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
...
+    secp256k1_gej_set_infinity(&pj);
+    secp256k1_pubkey_load(ctx, &p, &commitment_tmp);
+    secp256k1_gej_add_ge_var(&pj, &pj, &p, NULL);
+    secp256k1_pubkey_load(ctx, &p, commitment);
+    secp256k1_ge_neg(&p, &p);
+    secp256k1_gej_add_ge_var(&pj, &pj, &p, NULL);
+    return secp256k1_gej_is_infinity(&pj);
+}
+
+static char s2c_opening_magic[8] = { 0x5d, 0x05, 0x20, 0xb8, 0xb7, 0xf2, 0xb1, 0x68 };
```
elichai: Here it's `char` but in the struct it's `unsigned char`.

src/modules/schnorrsig/main_impl.h
```diff
@@ -0,0 +1,392 @@
...
+    secp256k1_scalar_set_b32(&k, buf, NULL);
+    if (secp256k1_scalar_is_zero(&k)) {
+        return 0;
+    }
+
+    secp256k1_ecmult_gen(&ctx->ecmult_gen_ctx, &rj, &k);
+    secp256k1_ge_set_gej(&r, &rj);
+    if (s2c_opening != NULL) {
+        secp256k1_s2c_opening_init(s2c_opening);
+        if (s2c_data32 != NULL) {
```
elichai: Maybe add a check if `s2c_opening == NULL || s2c_data32 == NULL`? Because if that's true I think it means the user made a mistake

src/modules/schnorrsig/main_impl.h
```diff
@@ -0,0 +1,392 @@
...
+            /* Create sign-to-contract commitment */
+            secp256k1_pubkey_save(&s2c_opening->original_pubnonce, &r);
+            secp256k1_ec_commit_seckey(ctx, buf, &s2c_opening->original_pubnonce, s2c_data32, 32);
+            secp256k1_scalar_set_b32(&k, buf, NULL);
```
elichai: This is redundant, already did it here: https://github.com/jonasnick/secp256k1/blob/schnorrsig-s2c-only-new/src/modules/schnorrsig/main_impl.h#L114

include/secp256k1.h
```diff
@@ -80,6 +80,26 @@ typedef struct {
...
+ *  guaranteed to be portable between different platforms or versions. It can
+ *  be safely copied/moved.
+ */
+typedef struct {
+    /* magic is set during initialization */
+    unsigned char magic[8];
+    /* Public nonce before applying the sign-to-contract commitment */
+    secp256k1_pubkey original_pubnonce;
+    /* Integer indicating if signing algorithm negated the nonce */
+    unsigned char nonce_is_negated;
```
elichai: Is this a computational optimization to avoid computing the jacobi symbol on `secp256k1_schnorrsig_verify_s2c_commit`?

src/modules/schnorrsig/main_impl.h
```diff
@@ -0,0 +1,392 @@
...
+int secp256k1_schnorrsig_parse(const secp256k1_context* ctx, secp256k1_schnorrsig* sig, const unsigned char *in64) {
+    (void) ctx;
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(sig != NULL);
+    ARG_CHECK(in64 != NULL);
+    memcpy(sig->data, in64, 64);
+    return 1;
+}
+
+int secp256k1_schnorrsig_verify_s2c_commit(const secp256k1_context* ctx, const secp256k1_schnorrsig *sig, const unsigned char *data32, const secp256k1_s2c_opening *opening) {
```
elichai: Should this function be constant time? could `data32` actually be a secret worth guarding?

elichai:
(I also prefer this model over #588)

src/modules/schnorrsig/main_impl.h
real-or-random: I think you're right but it seems that it is constant time in `data32`.

include/secp256k1.h
jonasnick: Yes, we could could compute the ec_commitment of `original_pubnonce` and `data` then negate if this is not a valid nonce. But this prevents batch verification. I will mention this in the doc.

src/modules/schnorrsig/main_impl.h
jonasnick: You mean `CHECK((s2c_opening != NULL) == (s2c_data32 != NULL))` (both should be NULL or both should be non-NULL)? Yes, I'll add an ARG_CHECK

src/secp256k1.c
jonasnick: Switched to `uint64_t` because there's less room for mistakes and it's easier to read.

src/modules/schnorrsig/main_impl.h
jonasnick: fixed

include/secp256k1.h
jonasnick: added comment to explain this

include/secp256k1.h
```diff
@@ -97,7 +97,10 @@ typedef struct {
     uint64_t magic;
     /* Public nonce before applying the sign-to-contract commitment */
     secp256k1_pubkey original_pubnonce;
-    /* Integer indicating if signing algorithm negated the nonce */
+    /* Byte indicating if signing algorithm negated the nonce. Alternatively when
+     * verifying we could compute the EC commitment of original_pubnonce and the
+     * data and negate if this would not be a valid nonce. But this would prevent
+     * batch verification of sign-to-contract commitments. */
     unsigned char nonce_is_negated;
```
real-or-random: nit, feel free to ignore: Boolean values are typically represented by `int` and not `unsigned char`.

jonasnick: Yes but ints are the worst. Since `nonce_is_negated` is only ever 0 or 1 I changed this to int.

jonasnick:
rebased on schnorrsigs PR and squashed

include/secp256k1.h
```diff
@@ -444,6 +468,37 @@ SECP256K1_API int secp256k1_ecdsa_signature_serialize_compact(
...
+ *  Args:    ctx: a secp256k1 context object.
+ *  Out: opening: pointer to an opening object. If 1 is returned, it is set to a
+ *                 parsed version of input. If not, its value is undefined.
+ *  In:  input34: pointer to 34-byte array with a serialized opening
+ *
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_s2c_opening_parse(
+    const secp256k1_context* ctx,
+    secp256k1_s2c_opening* opening,
+    const unsigned char *input34
```
elichai: Is it worth it to optimize this down to 33 bytes by just using a single bit to mark if the nonce has been negated?(we can use the same byte that we use to signal the parity)

src/modules/schnorrsig/main_impl.h
```diff
@@ -0,0 +1,395 @@
...
+    secp256k1_ge_set_gej(&r, &rj);
+
+    if (s2c_opening != NULL) {
+        secp256k1_s2c_opening_init(s2c_opening);
+        if (s2c_data32 != NULL) {
+            /* Create sign-to-contract commitment */
+            secp256k1_pubkey_save(&s2c_opening->original_pubnonce, &r);
+            secp256k1_ec_commit_seckey(ctx, buf, &s2c_opening->original_pubnonce, s2c_data32, 32);
+            secp256k1_scalar_set_b32(&k, buf, NULL);
+            secp256k1_ecmult_gen(&ctx->ecmult_gen_ctx, &rj, &k);
```
elichai: Maybe instead of recomputing the secret key and multiplying by the generator you can compute `H(R, data)`*G and add that to the R you already have? 
(although since all of this is constant time it might result in no performance change? I'm not sure about it)

src/secp256k1.c
```diff
@@ -681,10 +705,130 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
...
+    secp256k1_pubkey commitment_tmp;
+
+    if (!secp256k1_ec_commit(ctx, &commitment_tmp, pubkey, data, data_size)) {
+        return 0;
+    }
+
+    /* Return commitment == commitment_tmp */
+    secp256k1_gej_set_infinity(&pj);
+    secp256k1_pubkey_load(ctx, &p, &commitment_tmp);
+    secp256k1_gej_add_ge_var(&pj, &pj, &p, NULL);
```
elichai: You can use `secp256k1_gej_set_ge()` instead of setting to infinity and adding

jonasnick: fixed

src/modules/schnorrsig/main_impl.h
elichai: I guess the performance will be a matter of will this wrap around the order or not. (if it will cause a wrap than `k*G` is gonna be faster, if it doesn't then calculating the `H(R,data)*G` and adding to R should be faster i think)

jonasnick: No. Adding two curve points is completely separate from "wrapping around the order".

jonasnick:
This PR is quite outdated. Sign-to-contract should use the schnorrsig_sign_custom API. PR #1018 is a WIP continuation of this work. Closing.

PR #590, opened 2019-02-15
Add anti nonce-sidechannel protocol to schnorrsigs [̶a̶l̶t̶e̶r̶n̶a̶t̶i̶v̶e̶]̶ 

jonasnick:
~~This is an alternative to #572~~ based on #589 (EDIT: this seems to be generally preferred now). It's a demonstration that the anti nonce-sidechannel protocol can be build just as well on #589 as on #588.


include/secp256k1_schnorrsig.h
```diff
@@ -56,6 +56,85 @@ SECP256K1_API int secp256k1_schnorrsig_parse(
...
+ *     commitment from step 2 and the signature and opening received in step 4. If verification does
+ *     not succeed, the protocol failed and can be restarted.
+ */
+
+/** Create a randomness commitment on the host as part of the Anti Nonce Sidechannel Protocol.
+ *
+ *  Returns 1 on success, 0 on failure.
+ *  Args:              ctx: pointer to a context object (cannot be NULL)
+ *  Out: rand_commitment32: pointer to 32-byte array to store the returned commitment (cannot be NULL)
+ *  In:             rand32: the 32-byte randomness to commit to (cannot be NULL)
```
real-or-random: Maybe it's even better to make this function responsible for calling `secp256k1_rand256` to avoid that the user screws up when producing randomness. 

(But then the user can't use his fancy hardware RNG... If we want to support that, then maybe there could also a callback to a randomness function, and if it's NULL, then just call `lsecp256k1_rand256`. But in general, I don't like the idea of the user providing randomness, unless really necessary.)

include/secp256k1_schnorrsig.h
```diff
@@ -56,6 +56,85 @@ SECP256K1_API int secp256k1_schnorrsig_parse(
...
+ *  randomness to the signing device client and the client commits to the randomness in the nonce
+ *  using sign-to-contract.
+ *  In order to make the randomness unpredictable, the host and client must engage in a
+ *  commit-reveal protocol as follows:
+ *  1. The host draws the randomness, commits to it with the `anti_nonce_sidechan_host_commit`
+ *     function and sends the commitment to the client.
+ *  2. The client commits to its sign-to-contract original nonce (which is the nonce without the
+ *     sign-to-contract tweak) using the hosts commitment by calling the
+ *     `secp256k1_schnorrsig_anti_nonce_sidechan_client_commit` function. The client sends the
+ *     rusulting commitment to the host
```
benma: rusulting -> resulting :)

include/secp256k1_schnorrsig.h
```diff
@@ -56,6 +56,85 @@ SECP256K1_API int secp256k1_schnorrsig_parse(
...
+ *  Args:           ctx: pointer to a context object (cannot be NULL)
+ *  Out:  client_commit: pointer to a pubkey where the clients public nonce will be
+ *                       placed. This is the public nonce before doing the
+ *                       sign-to-contract commitment to the hosts randomness (cannot
+ *                       be NULL)
+ *  In:           msg32: the 32-byte message hash to be signed (cannot be NULL)
+ *             seckey32: the 32-byte secret key used for signing (cannot be NULL)
+ *    rand_commitment32: the 32-byte randomness commitment from the host (cannot be NULL)
+ */
+SECP256K1_API int secp256k1_schnorrsig_anti_nonce_sidechan_client_commit(
```
benma: To keep the api / interface smaller, the client could also call the sign function and get the commitment from the signature. Would there be any security downside in that (except for code readability?).

jonasnick: That doesn't seem to work if I understand you correctly. The client needs the hosts randomness to produce the signature which it doesn't get until it sends the commitment.

benma: I meant a first of two calls to the signing function, with the first one being a plain call with no host randomness, just to extract the client commitment (`R = k*G`, `R` can be reconstructed using the sig).

jonasnick: If you call `sign` twice your wasting a lot of time producing a signature that's unnecessary?

benma: :+1: I see.

src/modules/schnorrsig/main_impl.h
```diff
@@ -0,0 +1,466 @@
...
+    ARG_CHECK(client_commit != NULL);
+    ARG_CHECK(msg32 != NULL);
+    ARG_CHECK(seckey32 != NULL);
+    ARG_CHECK(rand_commitment32 != NULL);
+
+    if (!secp256k1_nonce_function_bipschnorr(nonce32, msg32, seckey32, NULL, rand_commitment32, 0)) {
+        return 0;
+    }
+
+    secp256k1_scalar_set_b32(&k, nonce32, NULL);
```
benma: Is there an overflow check missing here?

real-or-random: No, the BIP draft intentionally does not check for overflow here. (It happens with negligible probability.)

benma: :+1:

jonasnick:
This thing needs a better name. Covert channel is a much better term for this than side channel. If no one comes up with something better I'll go on with `anti_covert_channel`.

jonasnick:
Rebased

src/modules/schnorrsig/main_impl.h
```diff
@@ -0,0 +1,469 @@
...
+    secp256k1_ge r;
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(secp256k1_ecmult_gen_context_is_built(&ctx->ecmult_gen_ctx));
+    ARG_CHECK(client_commit != NULL);
+    ARG_CHECK(msg32 != NULL);
+    ARG_CHECK(seckey32 != NULL);
+    ARG_CHECK(rand_commitment32 != NULL);
+
+    if (!secp256k1_nonce_function_bipschnorr(nonce32, msg32, seckey32, NULL, rand_commitment32, 0)) {
```
benma: since `secp256k1_schnorrsig_sign` accepts `ndata` and `s2c_data32` and combines them as input to the nonce function, would it make sense that this client_commit function also gets an `ndata` param which is mixed in in the same way?

src/secp256k1.c
```diff
@@ -681,10 +705,130 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
...
+    secp256k1_pubkey commitment_tmp;
+
+    if (!secp256k1_ec_commit(ctx, &commitment_tmp, pubkey, data, data_size)) {
+        return 0;
+    }
+
+    /* Return commitment == commitment_tmp */
+    secp256k1_gej_set_infinity(&pj);
+    secp256k1_pubkey_load(ctx, &p, &commitment_tmp);
+    secp256k1_gej_add_ge_var(&pj, &pj, &p, NULL);
```
elichai: You can use `secp256k1_gej_set_ge()` instead of setting to infinity and adding

PR #600, opened 2019-03-13
scratch space: use single allocation

apoelstra:


src/scratch_impl.h
```diff
@@ -28,6 +29,9 @@ static secp256k1_scratch* secp256k1_scratch_create(const secp256k1_callback* err
 static void secp256k1_scratch_destroy(secp256k1_scratch* scratch) {
     if (scratch != NULL) {
         VERIFY_CHECK(scratch->frame == 0);
+        if (memcmp(scratch->magic, "scratch", 8) != 0) {
+            return;
+        }
```
jonasnick: how about setting the magic here back to 0s?

src/scratch_impl.h
```diff
@@ -72,6 +88,10 @@ static void *secp256k1_scratch_alloc(secp256k1_scratch* scratch, size_t size) {
     size_t frame = scratch->frame - 1;
     size = ((size + ALIGNMENT - 1) / ALIGNMENT) * ALIGNMENT;
 
+    if (memcmp(scratch->magic, "scratch", 8) != 0) {
```
jonasnick: Perhaps this should happen before accessing `scratch->frame`?

src/tests.c
```diff
@@ -259,37 +259,46 @@ void run_scratch_tests(void) {
...
+    secp256k1_scratch_deallocate_frame(&none->error_callback, scratch);
+    CHECK(secp256k1_scratch_max_allocation(&none->error_callback, scratch, 0) == 1000);
+    CHECK(secp256k1_scratch_alloc(&none->error_callback, scratch, 500) == NULL);
 
     /* cleanup */
-    secp256k1_scratch_space_destroy(scratch);
+    secp256k1_scratch_space_destroy(none, scratch);
+    CHECK(ecount == 0);
+
+    /* try to use badly initialized scratch space */
```
jonasnick: should be done for the other functions as well not just scratch_max_allocation or we reduce fraction of lines covered by tests.

src/scratch.h
```diff
@@ -12,28 +12,29 @@
 /* The typedef is used internally; the struct name is used in the public API
  * (where it is exposed as a different typedef) */
 typedef struct secp256k1_scratch_space_struct {
-    void *data[SECP256K1_SCRATCH_MAX_FRAMES];
+    unsigned char magic[8];
```
real-or-random: Probably not this PR but what do people think about randomizing this? Then it could protect not only against accidental misuse but also against stack smashing etc. This is just a random thought, I haven't really thought about whether it makes sense or not.

src/scratch_impl.h
```diff
@@ -19,22 +19,32 @@ static secp256k1_scratch* secp256k1_scratch_create(const secp256k1_callback* err
     secp256k1_scratch* ret = (secp256k1_scratch*)checked_malloc(error_callback, sizeof(*ret));
```
real-or-random: Can we unify this malloc with the other one too? Also, `scratch_crate` kind of promises to allocate at most `max_size` bytes but we allocate more.

src/scratch_impl.h
```diff
@@ -44,15 +54,18 @@ static size_t secp256k1_scratch_max_allocation(const secp256k1_scratch* scratch,
...
     VERIFY_CHECK(scratch->frame < SECP256K1_SCRATCH_MAX_FRAMES);
 
-    if (n <= secp256k1_scratch_max_allocation(scratch, objects)) {
+    if (memcmp(scratch->magic, "scratch", 8) != 0) {
+        secp256k1_callback_call(error_callback, "invalid scratch space");
+        return 0;
+    }
+
+    if (n <= secp256k1_scratch_max_allocation(error_callback, scratch, objects)) {
         n += objects * ALIGNMENT;
```
real-or-random: Isn't that too much? To store an array of `objects` things of type `t`, we need `objects * sizeof(t)` bytes of contiguous aligned memory. But there's no need to care about alignment here, because the stack frames are aligned already, no?

src/scratch_impl.h
apoelstra: I guess we should rename `max_size` ... the question is whether it will be more annoying to the user if (a) the function allocates more than they tell it to; (b) the function allocates exactly what they say, but gives them less scratch space than they intended.

My guess is that (a) is the better way to go here, but we should document this.

As for unifying the `malloc` calls, yeah, sure, I can do that.

apoelstra:
Addressed jonas' nits.

src/scratch_impl.h
apoelstra: We allocate multiple objects within one frame.

apoelstra:
Addressed Tim's nits

apoelstra:
Tim, regarding doing these checks only at the endpoints, the problem is that the user-level "endpoints" are everything that might use a scratch space, including everything that might use ecmult_multi.  So this set is unbounded and it's something we'd have to maintain going forward.

This isn't a big deal; we already have this situation for e.g. context objects where every new function needs to ARG_CHECK that the appropriate precomp tables are initialized. But it's annoying and I think this situation is simpler/more maintainable.

real-or-random:
Uh, I haven't looked at it from that point of view. Agreed, checking it deep in the call tree needs less duplicated code. It's is less consistent with what we do now for ARG_CHECKs though. But as you say, that's maybe because it was not the best design choice.

apoelstra:
My view is that for the context stuff it's necessary to have so much duplication because different functions have different precomp requirements. In this case this is sanity checking that really shouldn't concern any code outside of the scratch space module.

src/scratch_impl.h
real-or-random: Ah nevermind, I just understood the API...

I guess `objects * (ALIGNMENT - 1)` is still enough but that's not much of a difference.

apoelstra: `(objects - 1) * ALIGNMENT` you mean :). I'm going to leave this be, because otherwise we get into thorny issues thinking about what happens when `objects == 0` and also we'd have to change it in a bunch of places. And it's only 16 bytes.

real-or-random: I really meant `objects * (ALIGNMENT - 1)`. We're rounding up to a multiple of ALIGNMENT, so at most by `ALIGNMENT - 1`... But yes, just leave it, it's not worth the hassle.

real-or-random:
It should be noted that this PR fixes the following bug (for which I had the description text in some open editor windows already but Andrew fixed it faster than I managed to report it. :+1:  )

1. Create context `ctx` with `default_error_callback`
2. Create scratch space `s` with `ctx`.
3. Change error_callback of `ctx` to `cb`
4. Trigger a call to `secp256k1_ecmult_multi_var(ctx,...`, which will call `scratch_space_allocate_frame`. Assume `checked_malloc` fails.

Expected:
`cb` is called as error callback

Actual:
`default_error_callback` is called as error callback

src/scratch_impl.h
apoelstra: Oh! Yeah, I see. But did you further mean `(objects - 1) * (ALIGNMENT - 1)` since the first object does not require alignment (as the frame will already be aligned)?

I'll add a commit to do `ALIGNMENT-1` since that's quick and easy and doesn't create any new edge cases.

real-or-random: > Oh! Yeah, I see. But did you further mean `(objects - 1) * (ALIGNMENT - 1)` since the first object does not require alignment (as the frame will already be aligned)?

Well, if the frame will already be aligned, then `(objects - 1) * (ALIGNMENT - 1)`  is fine and I think `objects == 0` should just be rejected. The problem is that now that we just have one malloc call, it's not guaranteed that the frame is aligned, not even the first one. (I think ROUND_TO_ALIGN from #566 helps here.)

apoelstra: Oh, shit, yes, I think that's actually a bug in the current code. I need to add alignment when "allocating" frames as well.

src/tests.c
```diff
@@ -259,37 +259,55 @@ void run_scratch_tests(void) {
...
+    CHECK(secp256k1_scratch_max_allocation(&none->error_callback, scratch, 0) == 1000);
 
     /* ...but pushing a new stack frame does affect the max allocation */
-    CHECK(secp256k1_scratch_allocate_frame(scratch, 500, 1 == 1));
-    CHECK(secp256k1_scratch_max_allocation(scratch, 1) < 500); /* 500 - ALIGNMENT */
-    CHECK(secp256k1_scratch_alloc(scratch, 500) != NULL);
-    CHECK(secp256k1_scratch_alloc(scratch, 500) == NULL);
+    CHECK(secp256k1_scratch_allocate_frame(&none->error_callback, scratch, 500, 1) == 1);
+    CHECK(secp256k1_scratch_max_allocation(&none->error_callback, scratch, 1) < 500); /* 500 - ALIGNMENT */
+    CHECK(secp256k1_scratch_alloc(&none->error_callback, scratch, 500) != NULL);
```
jonasnick: How about checking here that the result % ALIGNMENT = 0?

real-or-random:
I discusse this with @jonasnick: We're not sure

real-or-random:
I discussed this with @jonasnick: We're not sure what the point of the frames still is when there is only a single malloc now. Maybe there is one but it seems that everything that can be done with frames  can be done with similar programming ergonomics without frames.

If we keep frames, then I think the internal API should be documented a little more because it's not immediately clear how to use it. In particular the wording is somewhat confusing: we "allocate" frames and "alloc(cate)" objects but then e.g., what does the "allocation" in "max_allocation" then give you? All of this is well-defined in the end but one needs to read the code to figure it out.

src/tests.c
```diff
@@ -259,37 +259,55 @@ void run_scratch_tests(void) {
...
+    CHECK(secp256k1_scratch_alloc(&none->error_callback, scratch, 500) == NULL);
+    CHECK(secp256k1_scratch_max_allocation(&none->error_callback, scratch, 0) == 1000);
 
     /* ...but pushing a new stack frame does affect the max allocation */
-    CHECK(secp256k1_scratch_allocate_frame(scratch, 500, 1 == 1));
-    CHECK(secp256k1_scratch_max_allocation(scratch, 1) < 500); /* 500 - ALIGNMENT */
-    CHECK(secp256k1_scratch_alloc(scratch, 500) != NULL);
-    CHECK(secp256k1_scratch_alloc(scratch, 500) == NULL);
+    CHECK(secp256k1_scratch_allocate_frame(&none->error_callback, scratch, 500, 1) == 1);
+    CHECK(secp256k1_scratch_max_allocation(&none->error_callback, scratch, 1) < 500); /* 500 - ALIGNMENT */
```
real-or-random: `500 - (ALIGNMENT - 1)`?

src/tests.c
```diff
@@ -259,37 +259,55 @@ void run_scratch_tests(void) {
...
+    CHECK(secp256k1_scratch_alloc(&none->error_callback, scratch, 500) != NULL);
+    CHECK(secp256k1_scratch_alloc(&none->error_callback, scratch, 500) == NULL);
```
real-or-random: Is this supposed to fail because 500 is too large or because that's the second alloc? (Or both?)

src/tests.c
```diff
@@ -259,37 +259,55 @@ void run_scratch_tests(void) {
...
 
-    CHECK(secp256k1_scratch_allocate_frame(scratch, 500, 1) == 0);
+    CHECK(secp256k1_scratch_allocate_frame(&none->error_callback, scratch, 500, 1) == 0);
 
     /* ...and this effect is undone by popping the frame */
```
jonasnick: Perhaps check that effect is undone by comparing the resulting scratch space fields with original fields (`frame` and `data`)

src/tests.c
jonasnick: Could also be checked after `secp256k1_scratch_space_create`

src/tests.c
```diff
@@ -259,37 +259,55 @@ void run_scratch_tests(void) {
...
-    CHECK(secp256k1_scratch_max_allocation(scratch, 0) == 1000);
+    CHECK(secp256k1_scratch_alloc(&none->error_callback, scratch, 500) == NULL);
+    CHECK(secp256k1_scratch_max_allocation(&none->error_callback, scratch, 0) == 1000);
 
     /* ...but pushing a new stack frame does affect the max allocation */
-    CHECK(secp256k1_scratch_allocate_frame(scratch, 500, 1 == 1));
-    CHECK(secp256k1_scratch_max_allocation(scratch, 1) < 500); /* 500 - ALIGNMENT */
-    CHECK(secp256k1_scratch_alloc(scratch, 500) != NULL);
-    CHECK(secp256k1_scratch_alloc(scratch, 500) == NULL);
+    CHECK(secp256k1_scratch_allocate_frame(&none->error_callback, scratch, 500, 1) == 1);
```
real-or-random: Fun that the old code was (basically) equivalent... :D

src/scratch_impl.h
```diff
@@ -62,21 +77,34 @@ static int secp256k1_scratch_allocate_frame(secp256k1_scratch* scratch, size_t n
...
+
+    if (memcmp(scratch->magic, "scratch", 8) != 0) {
+        secp256k1_callback_call(error_callback, "invalid scratch space");
+        return NULL;
+    }
+
+    frame = scratch->frame - 1;
     size = ((size + ALIGNMENT - 1) / ALIGNMENT) * ALIGNMENT;
 
     if (scratch->frame == 0 || size + scratch->offset[frame] > scratch->frame_size[frame]) {
```
jonasnick: fwiw, this size check can be circumvented by setting size such that `size + scratch->offset[frame]` overflows

jonasnick:
Needs rebase

apoelstra:
Rebased; squashed fixups

apoelstra:
Will revisit to remove the idea of "frames" entirely.

jonasnick:
> it seems that everything that can be done with frames can be done with similar programming ergonomics without frames.

It has the small advantage that you can deallocate the whole frame and not have to deallocate all objects one by one. However, that could be done without the api requiring to give the frame size in advance when allocating it. This seems to be unnecessary right now. But I don't recall the reason for "frames" either.

src/scratch.h
apoelstra: Not this PR :)

src/tests.c
apoelstra: Because 500 is too large.

src/tests.c
apoelstra: Am I allowed to use % on a pointer? That seems unlikely to be meaningful.

src/scratch_impl.h
apoelstra: Fixed with checkpointing

src/tests.c
apoelstra: Fixed - check `scratch->alloc_size`

src/scratch_impl.h
apoelstra: Fixed by eliminating frames.

apoelstra:
Rebased and replaced frames with "checkpoints" which don't require pre-registering an allocation size and don't have a limit to how many are usable.

src/tests.c
apoelstra: Probably overkill now that only `alloc_size` ever actually changes. Added checks that `alloc_size` is 0 and nonzero when we expect it.

src/scratch.h
```diff
@@ -7,29 +7,31 @@
...
 typedef struct secp256k1_scratch_space_struct {
+    /** guard against interpreting this object as other types */
     unsigned char magic[8];
+    /** actual allocated data */
     void *data;
-    void *current_frame;
-    size_t offset[SECP256K1_SCRATCH_MAX_FRAMES];
-    size_t frame_size[SECP256K1_SCRATCH_MAX_FRAMES];
-    size_t frame;
+    /** amount that has been allocated (i.e. `date + offset` is the next
```
real-or-random: s/date/data

src/scratch_impl.h
```diff
@@ -15,70 +15,77 @@
...
-    size_t frame = scratch->frame - 1;
-    size = ((size + ALIGNMENT - 1) / ALIGNMENT) * ALIGNMENT;
 
-    if (scratch->frame == 0 || size + scratch->offset[frame] > scratch->frame_size[frame]) {
+    if (memcmp(scratch->magic, "scratch", 8) != 0) {
+        secp256k1_callback_call(error_callback, "invalid scratch space");
+        return NULL;
+    }
+
+    size = ((size + ALIGNMENT - 1) / ALIGNMENT) * ALIGNMENT;
```
real-or-random: once #566 is merged (probably today), this can be replaced by ROUND_TO_ALIGN, see https://github.com/bitcoin-core/secp256k1/blob/0522caac8f28e9eaa19bfebc88e463fc8d409b7c/src/util.h#L96

src/scratch_impl.h
```diff
@@ -15,70 +15,77 @@
...
+    if (size > scratch->max_size - scratch->alloc_size) {
         return NULL;
     }
-    ret = (void *) ((unsigned char *) scratch->data[frame] + scratch->offset[frame]);
+    ret = (void *) ((char *) scratch->data + scratch->alloc_size);
     memset(ret, 0, size);
```
real-or-random: Maybe we want to memset the entire scratch space when creating it. Seems simpler and also sets the padding bytes (that are there due to rounding to alignment)

apoelstra: I'm not sure if it'd be simpler - we'd also have to memset on dealloc, so it'd probably be just as much code and accounting.

src/scratch.h
apoelstra: fixe

src/scratch_impl.h
apoelstra: done

apoelstra:
Rebased, fixed date/data typo, used `ROUND_TO_ALIGN` in place of manual alignment calculations.

gmaxwell:
ACK

PR #607, opened 2019-03-30
Use size_t shifts when computing a size_t

sipa:
This was detected by compiling with MSVC; it triggers warning C4334.

I don't think this is necessary, as we know the maximum shift is a very small integer, but this makes the code more obviously correct.

### DIFF
--- a/src/ecmult_impl.h
+++ b/src/ecmult_impl.h
@@ -960,7 +960,7 @@ static size_t secp256k1_pippenger_scratch_size(size_t n_points, int bucket_windo
     size_t entries = n_points + 1;
 #endif
     size_t entry_size = sizeof(secp256k1_ge) + sizeof(secp256k1_scalar) + sizeof(struct secp256k1_pippenger_point_state) + (WNAF_SIZE(bucket_window+1)+1)*sizeof(int);
-    return ((1<<bucket_window) * sizeof(secp256k1_gej) + sizeof(struct secp256k1_pippenger_state) + entries * entry_size);
+    return (sizeof(secp256k1_gej) << bucket_window) + sizeof(struct secp256k1_pippenger_state) + entries * entry_size;
 }
 
 static int secp256k1_ecmult_pippenger_batch(const secp256k1_ecmult_context *ctx, secp256k1_scratch *scratch, secp256k1_gej *r, const secp256k1_scalar *inp_g_sc, secp256k1_ecmult_multi_callback cb, void *cbdata, size_t n_points, size_t cb_offset) {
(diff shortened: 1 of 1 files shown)


practicalswift:
utACK 1b8e50570c7227d958989d16fafa40313073a462

Rationale:
1. The change makes the code more obviously correct as noted.
2. The change increases the signal to noise ratio in compiler/static analyser output which makes it easier to spot real issues.

sipa:
Updated. Suggestion from IRC:

    < gmaxwell> sipa: why multiply instead shifting the sizeof?

src/ecmult_impl.h
```diff
@@ -1063,7 +1063,7 @@ static size_t secp256k1_pippenger_max_points(secp256k1_scratch *scratch) {
 #ifdef USE_ENDOMORPHISM
         entry_size = 2*entry_size;
 #endif
-        space_overhead = ((1<<bucket_window) * sizeof(secp256k1_gej) + entry_size + sizeof(struct secp256k1_pippenger_state));
+        space_overhead = sizeof(secp256k1_gej) << bucket_window + entry_size + sizeof(struct secp256k1_pippenger_state);
```
practicalswift: Precedence is hard :-)

practicalswift:
FWIW, I believe that Greg's suggestion (with the precedence fix in place) will generate the same code as 1b8e50570c7227d958989d16fafa40313073a462 under both `clang` and `gcc` under `-O1`.

src/ecmult_impl.h
```diff
@@ -960,7 +960,7 @@ static size_t secp256k1_pippenger_scratch_size(size_t n_points, int bucket_windo
     size_t entries = n_points + 1;
 #endif
     size_t entry_size = sizeof(secp256k1_ge) + sizeof(secp256k1_scalar) + sizeof(struct secp256k1_pippenger_point_state) + (WNAF_SIZE(bucket_window+1)+1)*sizeof(int);
-    return ((1<<bucket_window) * sizeof(secp256k1_gej) + sizeof(struct secp256k1_pippenger_state) + entries * entry_size);
+    return sizeof(secp256k1_gej) << bucket_window + sizeof(struct secp256k1_pippenger_state) + entries * entry_size;
```
practicalswift:
```suggestion
    return (sizeof(secp256k1_gej) << bucket_window) + sizeof(struct secp256k1_pippenger_state) + entries * entry_size;
```

sipa: I don't think excessive parentheses help readability.

sipa: Except in this case they're not actually excessive; they're necessary for correctness. Sorry!

gmaxwell:
ACK.  Has someone verified that this silences this particular warning from MSVC  (e.g. did we get all the cases?)

sipa:
@gmaxwell Yes, see 15703 in the Bitcoin Core repo; I added (an earlier version of) this change there as an extra commit and it silences AppVeyor.

real-or-random:
> ```
> < gmaxwell> sipa: why multiply instead shifting the sizeof?
> ```

The multiplication makes the purpose clearer, at least for stupid me.

Whatever, post-merge utACK. :)

PR #614, opened 2019-05-10
Use a static constant table for small ecmult WINDOW_G sizes.

gmaxwell:
For discussion for now: this need to be reworked some to play nicely with #600.

When WINDOW_G is small there is no reason to not pre-compute it: on almost all embedded devices flash is much 'cheaper' than ram, pre-computation can be burdensomely slow, and the code to do the precomputation might actually be larger than the table (I haven't checked).  We wouldn't want to put a 1.3 MB table in the source/library but a 1024 byte one hardly seems like a concern.

This is relevant to #603 @laanwj


gmaxwell:
I only tested this very minimally, so please don't immediately dump it on an embedded device and waste 4 hours troubleshooting. Test on a desktop first. :P

src/ecmult_impl.h
```diff
@@ -302,6 +377,17 @@ static void secp256k1_ecmult_context_init(secp256k1_ecmult_context *ctx) {
...
+#if ECMULT_WINDOW_SIZE <= ECMULT_STATIC_TABLE_MAX
+    (void)cb;
+    if (ctx->pre_g != NULL) {
+        return;
+    }
+    ctx->pre_g = (secp256k1_ge_storage (*)[ECMULT_TABLE_SIZE(WINDOW_G)])secp256k1_pre_g_static_context;
+#ifdef USE_ENDOMORPISM
+    ctx->pre_g_128 = (secp256k1_ge_storage (*)[ECMULT_TABLE_SIZE(WINDOW_G)])secp256k1_pre_g_128_static_context;
+#endif
+    return;
```
laanwj: This explicit return seems unnecessary as the rest of the `#ifdef` already covers the entire function.

src/ecmult_impl.h
```diff
@@ -302,6 +377,17 @@ static void secp256k1_ecmult_context_init(secp256k1_ecmult_context *ctx) {
 }
 
 static void secp256k1_ecmult_context_build(secp256k1_ecmult_context *ctx, const secp256k1_callback *cb) {
+#if ECMULT_WINDOW_SIZE <= ECMULT_STATIC_TABLE_MAX
```
laanwj: It'd be nice to have a convenience `#define STATIC_WINDOW_G` that can be tested instead of repeating this expression every time.

laanwj:
Concept ACK

gmaxwell:
@apoelstra Thoughts on how this should interact with the single allocation change?

real-or-random:
Concept ACK

> @apoelstra Thoughts on how this should interact with the single allocation change?

How is this related to scratch spaces?

gmaxwell:
@real-or-random oops should have really directed at #566  just had 'single allocation' on the brain.   When we do single allocation / no-allocation  API's we'll need to leave the memory for this table out from the allocation.  (which is also why I want to do this PR after merging those, as I don't want to disrupt them.)

real-or-random:
To adopt this to #566, you just need to do the accounting correctly here:
https://github.com/bitcoin-core/secp256k1/blob/b4d7c15da2b8eeaaed8e286c5ffadff89de03426/src/ecmult_impl.h#L297

gmaxwell:
Rebased, fixed endomorphism support, ifdefed out secp256k1_ecmult_odd_multiples_table_storage_var.

TODO: Update configure output, include a comment explaining how the constants are generated, and decide on how big to support. 6 is sounding pretty good based on llanwj's data.

@laanwj  Any interest in getting the actual binary size without this PR, with it and size 2 and size 6? on RV?

src/ecmult_impl.h
gmaxwell: fixed

src/ecmult_impl.h
gmaxwell: Will do.

real-or-random:
> include a comment explaining how the constants are generated

It does not hurt but this style of just hardcoding the constants is somewhat inconsistent with the one by gen_context. I like the approach here more because it's much simpler; gen_context creates a lot of complexity in the build system. Maybe we should also add a pre-generated file with gen_context constants to the repo in the future.

gmaxwell:
TODO: Update configure output, include a comment explaining how the constants are generated, and decide on how big to support. Better handling for exhaustive tests?

I'm unsure about gen_context the big reason that we don't ship a static table is because it's enormous (as in comparable to the size of the codebase enormous). This issue doesn't really arise with window_g.   We should probably change signing to use #546 and then we could have equivalent performance with a much smaller table, and then it might be less of a consideration.

real-or-random:
> I'm unsure about gen_context the big reason that we don't ship a static table is because it's enormous (as in comparable to the size of the codebase enormous).

Yep okay, I didn't see that point.

PR #617, opened 2019-05-14
Pass scalar by reference in secp256k1_wnaf_const()

real-or-random:
After this change, no struct or union is passed by value in the
entire codebase. This makes it easier to compile the library with
CompCert.


gmaxwell:
Concept ACK

sipa:
Concept ACK

gmaxwell:
ACK. Benchmarks slightly but consistently faster for me too.

PR #620, opened 2019-05-15
Install headers automatically

vmayoral:
This fix install all the headers under include/ into
/usr/local/include. The fix solves problems that arise
when building libraries that depend on secp256k1 such
as libbitcoin-system which require all the headers

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,6 +8,8 @@ else
 JNI_LIB =
 endif
 include_HEADERS = include/secp256k1.h
+include_HEADERS += include/secp256k1_ecdh.h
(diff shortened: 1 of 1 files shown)


gmaxwell:
ACK. This seems right to me, and I don't see any harm in installing headers for disabled modules.

real-or-random:
It's no real harm but won't users be confused when there are headers installed that actually won't work? I  think it's better to install the headers depending on whether the modules are enables. I can create a PR if you agree.

gmaxwell:
Apparently we already were.  :)

Issue #624, opened 2019-05-23
`make install` fails


doitian:
It passes the same file twice to `install` command.

```
/usr/bin/install -c -m 644 include/secp256k1.h include/secp256k1_ecdh.h include/secp256k1_recovery.h include/secp256k1_recovery.h '/usr/local/include'
/usr/bin/install: will not overwrite just-created '/usr/local/include/secp256k1_recovery.h' with 'include/secp256k1_recovery.h'
Makefile:1151: recipe for target 'install-includeHEADERS' failed
make[1]: *** [install-includeHEADERS] Error 1
make[1]: Leaving directory '/home/travis/secp256k1'
Makefile:1578: recipe for target 'install-am' failed
make: *** [install-am] Error 2
```

doitian:
It seems introduced in 91fae3ace0291b144b27fd8bbda509042f5400f1

doitian:
Reproduce:

    ./configure --enable-module-recovery

PR #625, opened 2019-05-24
Avoid generate duplicated files for install command.

doitian:
The module headers are already added in the top `Makefile.am`, which is
introduced in 91fae3ace0291b144b27fd8bbda509042f5400f1, but the module
`Makefile.am.include` will add the headers again.

Steps to reproduce the issue:

    ./autogen.sh
    ./configure --enable-module-recovery --prefix /tmp/secp256k1
    make
    make install

Failure message:

    /usr/bin/install -c -m 644 include/secp256k1.h include/secp256k1_ecdh.h include/secp256k1_recovery.h include/secp256k1_recovery.h '/usr/local/include'
    /usr/bin/install: will not overwrite just-created '/usr/local/include/secp256k1_recovery.h' with 'include/secp256k1_recovery.h'
    Makefile:1151: recipe for target 'install-includeHEADERS' failed
    make[1]: *** [install-includeHEADERS] Error 1
    make[1]: Leaving directory '/home/travis/secp256k1'
    Makefile:1578: recipe for target 'install-am' failed
    make: *** [install-am] Error 2

Closes https://github.com/bitcoin-core/secp256k1/issues/624

PR #626, opened 2019-05-24
Revert "Merge #620: Install headers automatically"

doitian:
This reverts commit 91fae3ace0291b144b27fd8bbda509042f5400f1, reversing
changes made to 5df77a0eda6e902a1aa9c6249cdeaec197b1e0cd.

See discussion in https://github.com/bitcoin-core/secp256k1/pull/625

After the change, if we enable any module, `make install` fails because of the
duplicated files in the command line arguments.

Closes https://github.com/bitcoin-core/secp256k1/issues/624

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -8,8 +8,6 @@ else
 JNI_LIB =
 endif
 include_HEADERS = include/secp256k1.h
-include_HEADERS += include/secp256k1_ecdh.h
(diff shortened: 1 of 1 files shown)


gmaxwell:
ACK.  (following the principle of revert errors eagerly and favor the status quo, if it turns out I'm missing some reason why this wasn't the best solution-- that's fine. We could switch to another solution later).

PR #627, opened 2019-05-25
Guard memcmp in tests against mixed size inputs.

### LINKED ISSUE #623: memcmp() may read unitiliazed memory in tests
If `len_der != len_der_lax`, then we may read uninitialized memory here:
https://github.com/bitcoin-core/secp256k1/blob/912680ed8675bbf9c74c4c112ca3c41d27a295d7/src/tests.c#L4318

This should probably be handled similar to
https://github.com/bitcoin-core/secp256k1/blob/912680ed8675bbf9c74c4c112ca3c41d27a295d7/src/tests.c#L4302

Same issues further below for the OpenSSL roundtrip.

Loosely related to #532 but I agree that #532 is not a good idea.

gmaxwell:
Reported by real-or-random.

Fixes #623.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -4315,7 +4315,7 @@ int test_ecdsa_der_parse(const unsigned char *sig, size_t siglen, int certainly_
     if (valid_der) {
         ret |= (!roundtrips_der_lax) << 12;
         ret |= (len_der != len_der_lax) << 13;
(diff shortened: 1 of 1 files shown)


real-or-random:
utACK

practicalswift:
utACK 248bffb0526ce8d829ce22f4d63d0d4a9ccfe137

Discussed in #532.

apoelstra:
ACK

PR #628, opened 2019-05-25
Fix ability to compile tests without -DVERIFY.

gmaxwell:
Broken by 3f3964e4.

It's important that the tests are also run without -DVERIFY due to
 the possibility that side-effects of a VERIFY_CHECK fix a bug that
 would otherwise be detected.

Use of the verify_check macro in tests isn't sufficient.


real-or-random:
utACK

apoelstra:
ACK

PR #629, opened 2019-05-26
Avoid calling _is_zero when _set_b32 fails.

gmaxwell:
Most of the codebase correctly used short-cutting to avoid calling
 _is_zero on possibly incompletely initialized elements, but a few
 places were missed.


real-or-random:
utACK

apoelstra:
ACK.

Did not check if there are more places that need fixing.

gmaxwell:
re-more places, I checked for them (which is why there is more than one in this PR) but didn't make a heroic effort. I essentially just looked at every _is_zero that wasn't already shortcutted to check if it shouldn't be. If there are places where the test after loading was something else, I might have missed it.

sipa:
utACK cd473e02c372217c3a6608ce5afaa543ed78f891

PR #448, opened 2017-03-15
Clear sensitive memory without getting optimized out.

isle2983:
(Previous discussion of this topic in #438 )

`explicit_bzero()` is used when available on the platform (glibc 2.25+), otherwise it falls back to an inline implementation. There are many possible versions of how the [inline could be implemented](https://gist.github.com/jiixyj/3e3389649c866f7ff7bd) but I have no reason to prefer one over another.

The commit stack does some refactoring in order to preserve the intended non-'cleanse' functionality done by several of the `_clear()` functions. I constructed the stack as unsquashed, incremental refactoring changes. Please let me know if a squash is preferred.

The measured delta in performance appears negligible. Trials were done inside a minimal headless VMs with Debian 8 in one for `glibc 2.19` and the current base of Arch Linux in the other for `glibc 2.25`. The full session output of the trials was logged for reference [and is available as a gist.](https://gist.github.com/isle2983/4883cad4d80952d64aba6362ab9f3322)

For reference to bare metal performance with the same CPU, timing on the underlying native Debian 8 install (with `glibc 2.19` and `gcc 4.9.2` running with `master`:
```
(wo/ endomorphism)
$ ./bench_sign
ecdsa_sign: min 55.3us / avg 55.4us / max 55.5us
$ ./bench_verify
ecdsa_verify: min 85.6us / avg 85.9us / max 86.3us
ecdsa_verify_openssl: min 476us / avg 484us / max 490us
```

Trials:

1) `master` w/ `glibc 2.19`, `gcc 4.9.2` w/ and wo/ endomorphisim
```
(wo/ endomorphism)
$ ./bench_sign
ecdsa_sign: min 55.9us / avg 56.4us / max 56.7us
$ ./bench_verify
ecdsa_verify: min 86.2us / avg 86.8us / max 87.7us
ecdsa_verify_openssl: min 487us / avg 488us / max 491us

(w/ endomorphism)
$ ./bench_sign
ecdsa_sign: min 56.1us / avg 56.6us / max 57.2us
$ ./bench_verify
ecdsa_verify: min 66.5us / avg 67.4us / max 68.8us
ecdsa_verify_openssl: min 487us / avg 493us / max 504us
```
2) a16034039f07572a64bc9704c23abb1fff9d70ad w/ `glibc 2.19` `gcc 4.9.2` w/ and wo/ endomorphism
```
(wo/ endomorphism)
./bench_sign
ecdsa_sign: min 55.8us / avg 56.5us / max 57.0us
[...]

PR #636, opened 2019-06-06
Clear sensitive memory without getting optimized out

### LINKED ISSUE #185: Memory zeroization improvements
Existing 'best effort' zeriozation for private data is hardly even best effort. At a minimum we should consider doing this via an extern-ed function and memset_s if available. No guarantees can still be provided, of course.

We might also consider wrapping the API entrance of private data handling functions like:

handle_data(){ 
  ret=handle_data_impl();
  handle_data_zero_stack(); 
  return ret;
}

Where _zero_stack uses slightly more stack than the whole callgraph for _impl and zeros it, in order to catch private data spilled onto the stack during execution before returning outside of our control.

I'm not sure where exactly where the line between best effort and security theatre is... there is only so much that can really be done (esp in portable code) here.

real-or-random:
This is a rebase of #448, including a few changes (mostly implementing what I described in #185). I haven't tested this on Windows and I actually haven't reviewed this in detail so far, but people can start to have a look.

Fixes #185 and closes #448.

### DIFF
--- a/src/field.h
+++ b/src/field.h
@@ -14,8 +14,8 @@
  *  - Each field element can be normalized or not.
  *  - Each field element has a magnitude, which represents how far away
  *    its representation is away from normalization. Normalized elements
- *    always have a magnitude of 1, but a magnitude of 1 doesn't imply
- *    normality.
+ *    always have a magnitude of 0 or 1, but a magnitude of 1 doesn't
+ *    imply normality.
  */
 
 #if defined HAVE_CONFIG_H
@@ -51,7 +51,8 @@ static int secp256k1_fe_normalizes_to_zero(secp256k1_fe *r);
  *  implementation may optionally normalize the input, but this should not be relied upon. */
 static int secp256k1_fe_normalizes_to_zero_var(secp256k1_fe *r);
 
-/** Set a field element equal to a small integer. Resulting field element is normalized. */
+/** Set a field element equal to a small integer. Resulting field element is normalized; it has
+ *  magnitude 0 if a == 0, and magnitude 1 otherwise. */
 static void secp256k1_fe_set_int(secp256k1_fe *r, int a);
 
 /** Sets a field element equal to zero, initializing all fields. */
--- a/src/field.h
+++ b/src/field.h
@@ -55,7 +55,7 @@ static int secp256k1_fe_normalizes_to_zero_var(secp256k1_fe *r);
  *  magnitude 0 if a == 0, and magnitude 1 otherwise. */
 static void secp256k1_fe_set_int(secp256k1_fe *r, int a);
 
-/** Sets a field element equal to zero, initializing all fields. */
+/** Clear a field element to prevent leaking sensitive information. */
 static void secp256k1_fe_clear(secp256k1_fe *a);
 
 /** Verify whether a field element is zero. Requires the input to be normalized. */
--- a/src/util.h
+++ b/src/util.h
@@ -11,10 +11,16 @@
 #include "libsecp256k1-config.h"
 #endif
 
+#include <string.h>
 #include <stdlib.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <limits.h>
+
+#if defined(_MSC_VER)
+// For SecureZeroMemory
+#include <Windows.h>
+#endif
 
 typedef struct {
     void (*fn)(const char *text, void* data);
@@ -173,6 +179,31 @@ static SECP256K1_INLINE void memczero(void *s, size_t len, int flag) {
         p++;
         len--;
     }
+}
+
+/* Cleanses memory to prevent leaking sensitive info. Won't be optimized out. */
+static SECP256K1_INLINE void memclear(void *ptr, size_t len) {
+#if defined(_MSC_VER)
+    /* SecureZeroMemory is guaranteed not to be optimized out by MSVC. */
+    SecureZeroMemory(ptr, n);
+#elif defined(__GNUC__)
+    /* We use a memory barrier that scares the compiler away from optimizing out the memset.
+     *
+     * Quoting Adam Langley <agl@google.com> in commit ad1907fe73334d6c696c8539646c21b11178f20f
+     * in BoringSSL (ISC License):
+     *    As best as we can tell, this is sufficient to break any optimisations that
+     *    might try to eliminate "superfluous" memsets.
+     * This method used in memzero_explicit() the Linux kernel, too. Its advantage is that it is
+     * pretty efficient, because the compiler can still implement the memset() efficently,
+     * just not remove it entirely. See "Dead Store Elimination (Still) Considered Harmful" by
+     * Yang et al. (USENIX Security 2017) for more background.
+     */
+    memset(ptr, 0, len);
+    __asm__ __volatile__("" : : "r"(ptr) : "memory");
+#else
+    void *(*volatile const volatile_memset)(void *, int, size_t) = memset;
+    volatile_memset(ptr, 0, len);
+#endif
 }
 
 #endif /* SECP256K1_UTIL_H */
--- a/src/ecmult_const_impl.h
+++ b/src/ecmult_const_impl.h
@@ -23,8 +23,8 @@
     VERIFY_CHECK(((n) & 1) == 1); \
     VERIFY_CHECK((n) >= -((1 << ((w)-1)) - 1)); \
     VERIFY_CHECK((n) <=  ((1 << ((w)-1)) - 1)); \
-    VERIFY_SETUP(secp256k1_fe_clear(&(r)->x)); \
-    VERIFY_SETUP(secp256k1_fe_clear(&(r)->y)); \
+    VERIFY_SETUP(secp256k1_fe_set_int(&(r)->x, 0)); \
+    VERIFY_SETUP(secp256k1_fe_set_int(&(r)->y, 0)); \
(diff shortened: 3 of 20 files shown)


src/util.h
```diff
@@ -125,6 +131,32 @@ static SECP256K1_INLINE void *manual_alloc(void** prealloc_ptr, size_t alloc_siz
     return ret;
 }
 
+/* Cleanses memory to prevent leaking sensitive info. Won't be optimized out. */
+#define SECP256K1_CLEANSE(v) secp256k1_cleanse(&(v), sizeof(v))
+static SECP256K1_INLINE void secp256k1_cleanse(void *ptr, size_t len) {
+#if defined(_MSC_VER)
+    /* SecureZeroMemory is guaranteed not to be optimized out by MSVC. */
+    SecureZeroMemory(ptr, n);
+#elif defined(__GNUC__)
```
real-or-random: I couldn't figure out when GCC started to support assembly but it's some version <= 2.95... So if you have such an old compiler, I guess you're on your own.

src/util.h
```diff
@@ -11,10 +11,16 @@
...
 #endif
 
+#include <string.h>
 #include <stdlib.h>
 #include <stdint.h>
 #include <stdio.h>
 
+#if defined(_MSC_VER)
+// For SecureZeroMemory
+#include <Windows.h>
```
real-or-random: I think that's okay but I'm not entirely sure if we want that. It's used in Bitcoin Core too.

gmaxwell:
I think using a sizeof on an argument is likely to introduce almost impossible to detect bugs when e.g. someone hands it the first element of an array (esp one passed from another function).  It doesn't feel particularly typesafe to me.

Is there a reason why you preferred this approach to having every subsystem (like field, scalar) define a clear function for its relevant type?

real-or-random:
No, I don't prefer that to be honest. It's just the approach taken in #448, and I was somewhat worried about it, too.

src/util.h
gmaxwell: I would be that this should be version gated, I'll see if I can figure out where thats supported to.

src/util.h
gmaxwell: GCC 2.95 is likely the earliest we'd would ever encounter someone wanting to support. (Haiku OS uses it still, I believe :) )

real-or-random:
I removed the macro from the PR. Also I removed `_fe_set_zero` from the PR (one can use `_fe_set_int`).

I'm happy to take suggestions for additional code locations where memory should be cleared.

src/util.h
```diff
@@ -125,6 +131,31 @@ static SECP256K1_INLINE void *manual_alloc(void** prealloc_ptr, size_t alloc_siz
...
+     * Quoting Adam Langley <agl@google.com> in commit ad1907fe73334d6c696c8539646c21b11178f20f
+     * in BoringSSL (ISC License):
+     *    As best as we can tell, this is sufficient to break any optimisations that
+     *    might try to eliminate "superfluous" memsets.
+     * This method used in memzero_explicit() the Linux kernel, too. Its advantage is that it is
+     * pretty efficient, because the compiler can still implement the memset() efficently,
+     * just not remove it entirely. See "Dead Store Elimination (Still) Considered Harmful" by
+     * Yang et al. (USENIX Security 2017) for more background.
+     */
+    memset(ptr, 0, len);
```
real-or-random: Now that "clearing" is cleanly separated from "setting to zero", it's a good idea to replace 0 with a different value here for testing that the code does not rely on the fact that the "_clear" functions set to 0. We could actually always use a different value here, or use a different value in (some runs of) tests. What do people think about this?

src/field.h
```diff
@@ -49,10 +49,11 @@ static int secp256k1_fe_normalizes_to_zero(secp256k1_fe *r);
  *  implementation may optionally normalize the input, but this should not be relied upon. */
 static int secp256k1_fe_normalizes_to_zero_var(secp256k1_fe *r);
 
-/** Set a field element equal to a small integer. Resulting field element is normalized. */
+/** Set a field element equal to a small integer. Resulting field element is normalized; it has
+ *  magnitude 0 if a == 0, and magnitude 1 otherwise. */
 static void secp256k1_fe_set_int(secp256k1_fe *r, int a);
```
real-or-random: (I wondered why this accepts `int`. An unsigned type seems more appropriate. For type safety, we would even want to exclude too large values and use something as small as `unsigned char`. But I guess this is slower and just not worth the hassle)?

src/util.h
elichai: Did you try checking the binary that it's actually not optimized out?

elichai:
What about clearing the memory of the temporary FE in the GE addition? https://github.com/bitcoin-core/secp256k1/blob/master/src/group_impl.h#L419 (and I think most of this file)

src/util.h
real-or-random: Not yet. But yes, we should do this with a few compilers.

real-or-random:
Thanks, collecting more:
https://github.com/bitcoin-core/secp256k1/blob/master/src/scalar_impl.h#L71
A lot in https://github.com/bitcoin-core/secp256k1/blob/master/src/scalar_8x32_impl.h  (also also 4x64)

And I think there more in the field module... I guess I'll just go through the entire codebase. :shrug:

gmaxwell:
@elichai  *_var are not used on secret inputs and probably should not waste time zeroizing anything.

src/util.h
jonasnick: > it's a good idea to replace 0 with a different value here for testing

I don't think it's a good idea to use a different value in testing. But I don't see a downside to just replace `0` with `42`. This also has the (slight) advantage that it'll make it easier to detect now improper usage of `*_clear` when rebasing -zkp on top of this. -zkp uses `*_clear` for initializing in many places.

jonasnick:
Results are positive and very similar to the above with clang version 8.0.0.

elichai:
@jonasnick did you try to benchmark if this affects performance in a non negligible way?

jonasnick:
On my system the memory barrier and volatile function pointer method have no measurable performance difference. However, using one of the methods vs. using nothing in `bench_sign` appears to be `1.2%` slower (`41.3us` vs `40.8us`, averaged over 6 runs).

jonasnick:
Looks like memsetting to `0x42` instead of `0` costs about 1% :/
```
ecdsa_sign: min 41.3us / avg 41.5us / max 42.2us
vs.
ecdsa_sign: min 40.9us / avg 41.1us / max 41.8us
```
(I changed run_benchmark count argument from 10 to 500)

sipa:
@jonasnick That sounds like a deal breaker.

real-or-random:
Note that's just signing time, not verification time.

But I agree, let's stick to 0, that's just simpler.

sipa:
Concept ACK.

There are a number of data types which end up being cleared using the generic `secp256k1_mem_clear()` function after this PR (including at least `secp256k1_sha256` and `secp256k1_rfc6979_hmac_sha256`). I think it would be cleaner to provide `*_clear()` functions for those as well, and use those where relevant.

real-or-random:
I rebased this and implemented @sipas suggestion for the hash API. Now the caller is responsible to call `_clean()` (instead of letting `_finalize`) do it. I think this is actually the right thing to do.



nit: Also renamed `secp256k1_mem_clear` to `memclear` now that we have also `memczero` in `util.h`

real-or-random:
> Going through the codebase to look for locations where a `_clear` is missing can happen in another PR.

Indeed, so this is ready for review.

For looking into more places in the future, in particular within some arithmetic function, I was wondering if "overdoing" this will introduce more stores to memory, which is not only slightly worse for security but will probably also affect performace. For example, the `explicit_bzero` function in Glibc has this problem:

> Also, explicit_bzero only operates on RAM. If a sensitive data object never needs to have its address taken other than to call explicit_bzero, it might be stored entirely in CPU registers until the call to explicit_bzero. Then it will be copied into RAM, the copy will be erased, and the original will remain intact. Data in RAM is more likely to be exposed by a bug than data in registers, so this creates a brief window where the data is at greater risk of exposure than it would have been if the program didn’t try to erase it at all. 

https://www.gnu.org/software/libc/manual/html_node/Erasing-Sensitive-Data.html

AFAIU and I've seen from looking at generated ASM this should not happen with our memory barrier approach but I'm not 100% sure and any insight will be helpful.

Note to myself: Cleaning registers is out of the scope currently but today the cited USENIX paper  reminded me of a simple but expensive way to do this: Just run the operation again. (For example, run another SHA256 transform on dummy midstate and dummy input.) This will reuse and hence overwrite the registers (modulo inlining etc).

real-or-random:
Apparently stores of 64-byte zero-over-zero can be faster than stores of other values, see https://twitter.com/BRIAN_____/status/1260913021116993536. This would be a reason not so overwrite with zeros.

But I think we shouldn't have secrets that are all zeroes.This would be relevant for indistinguishability-based primitives with arbitrary inputs, e.g.,  encryption, but not for signing and key exchange.

real-or-random:
> Conceptually I think it makes sense to have separate secure-erase wiping and zeroing of values. It needs a big redo by now though; are you still interesting in working on this?

Yeah, I certainly want to come back to this, but I don't want to commit to a specific timeline. Marking as draft for now.

siv2r:
Should this PR be closed since #1579 has replaced it?

PR #637, opened 2019-06-11
ecdsa sign-to-contract module, with anti nonce covert chan util functions

benma:
This is based on the description of the fix by Stepan: https://medium.com/cryptoadvance/hardware-wallets-can-be-hacked-but-this-is-fine-a6156bbd199

The protocol wording and functions are copied/adapted from Jonas
Nick's PRs which do the same for BIP-Schnorr:

https://github.com/bitcoin-core/secp256k1/pull/590

https://github.com/bitcoin-core/secp256k1/pull/590/commits/ae5fb7f8f143244de07e17fee6671e0dd68c81a6#diff-b19c5ee427283d4d82bc5beb4e2f4777R59

https://github.com/bitcoin-core/secp256k1/pull/590/commits/ae5fb7f8f143244de07e17fee6671e0dd68c81a6#diff-313ca26f0048bc16a608709915d0111eR70

1.
Add secp256k1_ecdsa_anti_nonce_sidechan_client_commit to return the
curve point committing to the signing client nonce.

This is a convenience function and can technically be emulated by
calling secp256k1_ecdsa_sign() and reconstructing the curve point from
the signature r/s values.

2.

secp256k1_ecdsa_sign_nonce_tweak_add, which is the same as
secp256k1_ecdsa_sign_nonce, but with an additional optional tweak parameter to
add to the nonce.

The nicer way to do this is to redefine `secp256k1_nonce_function` to
have a tweak param, but this would break API compatiblity. The way it
is implemented is fully backwards compatible.


benma:
cc @jonasnick 

Is this the right direction? If so I can add the host functions and some unit tests.

include/secp256k1.h
```diff
@@ -526,6 +526,37 @@ SECP256K1_API extern const secp256k1_nonce_function secp256k1_nonce_function_rfc
...
+ *             seckey32: the 32-byte secret key used for signing (cannot be NULL)
+ *              noncefp: pointer to a nonce generation function. If NULL, secp256k1_nonce_function_default is used
+ *    rand_commitment32: the 32-byte randomness commitment from the host (cannot be NULL)
+ */
+SECP256K1_API int secp256k1_ecdsa_anti_nonce_sidechan_client_commit(
+    const secp256k1_context* ctx,
+    secp256k1_pubkey *client_commit,
+    const unsigned char *msg32,
+    const unsigned char *seckey32,
+    secp256k1_nonce_function noncefp,
```
jonasnick: The reason why I'm not supporting arbitrary noncefps in my PR is to avoid giving a nonce function that doesn't make use of the `noncedata`. In the Schnorr case that would lead to nonce reuse and allows the host to extract the secret key. I suspect the same is true for ECDSA.

src/secp256k1.c
```diff
@@ -446,9 +446,52 @@ static int nonce_function_rfc6979(unsigned char *nonce32, const unsigned char *m
...
+    secp256k1_ge_set_gej(&r, &rj);
+    secp256k1_pubkey_save(client_commit, &r);
+    return 1;
+}
+
 int secp256k1_ecdsa_sign(const secp256k1_context* ctx, secp256k1_ecdsa_signature *signature, const unsigned char *msg32, const unsigned char *seckey, secp256k1_nonce_function noncefp, const void* noncedata) {
+    return secp256k1_ecdsa_sign_nonce_tweak_add(ctx, signature, msg32, seckey, noncefp, noncedata, NULL);
+}
+
+int secp256k1_ecdsa_sign_nonce_tweak_add(const secp256k1_context* ctx, secp256k1_ecdsa_signature *signature, const unsigned char *msg32, const unsigned char *seckey, secp256k1_nonce_function noncefp, const void* noncedata, const unsigned char* nonce_tweak32) {
```
jonasnick: If this function assumes that the caller gives a matching noncedata to the nonce_tweak instead of deriving the host commitment from the nonce_tweak by hashing it, the caller can't be stateless.

src/secp256k1.c
```diff
@@ -472,6 +522,9 @@ int secp256k1_ecdsa_sign(const secp256k1_context* ctx, secp256k1_ecdsa_signature
                 break;
             }
             secp256k1_scalar_set_b32(&non, nonce32, &overflow);
+            if (nonce_tweak32 != NULL) {
+                secp256k1_scalar_add(&non, &non, &nonce_tweak);
```
jonasnick: This is not a sign-to-contract commitment, but it works for the anti nonce covertchan because the host already has the nonce. I think it would be much better if `ecdsa_sign` would generally support sign to contract commitments and not only this protocol. Sign to contract would be `R = R1 + H(R1, k2)*G`.

include/secp256k1.h
```diff
@@ -526,6 +526,37 @@ SECP256K1_API extern const secp256k1_nonce_function secp256k1_nonce_function_rfc
...
+/** Compute commitment on the client as part of the ECDSA Anti Nonce Sidechannel Protocol.
+ *
+ * ECDSA Anti Nonce Sidechannel Protocol:
+ * 1. The host draws randomness `k2`, commits to it with sha256 and sends the commitment to the client.
+ * 2. The client commits to it's original nonce `k1` using the host commitment by calling
+ *    `secp256k1_ecdsa_anti_nonce_sidechan_client_commit`. The client sends the resulting commitment
+ *   `R1` to the host.
+ * 3. The host replies with `k2` generated in step 1.
+ * 4. The client checks that the host commitment from step 1 commits to `k2` from step 3 and signs
+ *    with `secp256k1_ecdsa_sign_nonce_tweak_add`, using the `k2` as the tweak and the host
```
jonasnick: In general I'm trying to avoid the name `tweak` because it carries little information.

src/secp256k1.c
benma: Thanks a lot for the great review! 

I can try to make a general sign to contract ecdsa sign function.

Peeking at your Schnorr implementation of it, I am unsure about a few things:

1. There is `secp256k1_s2c_opening`, which seems to return the original `R1` from the sign function. Why is this needed when there is already a function to produce the same commitment (`secp256k1_(schnorrsig|ecdsa)_anti_nonce_sidechan_client_commit`) beforehand?

2. What is `nonce_is_negated` and when is it set? [Here](https://github.com/bitcoin-core/secp256k1/pull/590/files#diff-313ca26f0048bc16a608709915d0111eR206) the check is `if (!secp256k1_fe_is_quad_var(&r.y)) {`, but I don't know what this means. Does it apply for ecdsa?

In the ecdsa case, do I need to add all of that, or is just adding `H(R1, k2)` instead of `k2` sufficient?

include/secp256k1.h
```diff
@@ -526,6 +526,37 @@ SECP256K1_API extern const secp256k1_nonce_function secp256k1_nonce_function_rfc
...
 
+/** Compute commitment on the client as part of the ECDSA Anti Nonce Sidechannel Protocol.
+ *
+ * ECDSA Anti Nonce Sidechannel Protocol:
+ * 1. The host draws randomness `k2`, commits to it with sha256 and sends the commitment to the client.
+ * 2. The client commits to it's original nonce `k1` using the host commitment by calling
+ *    `secp256k1_ecdsa_anti_nonce_sidechan_client_commit`. The client sends the resulting commitment
+ *   `R1` to the host.
+ * 3. The host replies with `k2` generated in step 1.
+ * 4. The client checks that the host commitment from step 1 commits to `k2` from step 3 and signs
```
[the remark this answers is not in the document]
benma: :+1: fixed

include/secp256k1.h
benma: fixed

include/secp256k1.h
benma: fixed

src/secp256k1.c
benma: fixed (copied the solution from your PR).

src/secp256k1.c
benma: I know pushed a commit which simply adds `H(R1, k2)` to the nonce to produce `R`. Seemed less intrusive and more efficient than computing `R2` and adding it to `R1` to get `R`.

benma: Question 2. is resolved by reading [BIP-Schnorr](https://github.com/sipa/bips/blob/bip-schnorr/bip-schnorr.mediawiki). Unless I misunderstand, it seems that it does not apply to ecdsa.

jonasnick: To question 1: True, but if you want sign-to-contract outside of the anti-nonce-sidechan context then it's much simpler to just have a single opening structure that contains the negated nonce and original nonce and that you can just give to s2c_verify.

include/secp256k1.h
```diff
@@ -580,16 +579,17 @@ SECP256K1_API int secp256k1_ecdsa_sign(
...
-SECP256K1_API int secp256k1_ecdsa_sign_nonce_tweak_add(
+SECP256K1_API int secp256k1_ecdsa_sign_to_contract(
     const secp256k1_context* ctx,
     secp256k1_ecdsa_signature *sig,
     const unsigned char *msg32,
     const unsigned char *seckey,
     secp256k1_nonce_function noncefp,
     const void *ndata,
-    const unsigned char* nonce_tweak32
+    const unsigned char* s2c_data32
```
jonasnick: nit: we should standardize on the position (in my PR it's before the noncefp, but I don't really care which way it is).

src/secp256k1.c
```diff
@@ -486,14 +482,56 @@ int secp256k1_ecdsa_anti_nonce_sidechan_client_commit(
 }
 
 int secp256k1_ecdsa_sign(const secp256k1_context* ctx, secp256k1_ecdsa_signature *signature, const unsigned char *msg32, const unsigned char *seckey, secp256k1_nonce_function noncefp, const void* noncedata) {
-    return secp256k1_ecdsa_sign_nonce_tweak_add(ctx, signature, msg32, seckey, noncefp, noncedata, NULL);
+    return secp256k1_ecdsa_sign_to_contract(ctx, signature, msg32, seckey, noncefp, noncedata, NULL);
 }
 
-int secp256k1_ecdsa_sign_nonce_tweak_add(const secp256k1_context* ctx, secp256k1_ecdsa_signature *signature, const unsigned char *msg32, const unsigned char *seckey, secp256k1_nonce_function noncefp, const void* noncedata, const unsigned char* nonce_tweak32) {
+/* Compute an ec commitment tweak as hash(pubkey, data). */
+static int secp256k1_ec_commit_tweak(const secp256k1_context *ctx, unsigned char *tweak32, const secp256k1_pubkey *pubkey, const unsigned char *data, size_t data_size) {
```
jonasnick: Looks like this is copy and pasted from my commit. Why not cherry-pick them? Would be easier to review.

src/secp256k1.c
```diff
@@ -446,11 +446,91 @@ static int nonce_function_rfc6979(unsigned char *nonce32, const unsigned char *m
...
+    const unsigned char *data,
+    size_t data_size) {
+    secp256k1_gej rp;
+    secp256k1_ge r;
+    secp256k1_ecmult_gen(&ctx->ecmult_gen_ctx, &rp, seckey);
+    secp256k1_ge_set_gej(&r, &rp);
+    return secp256k1_ec_commit_tweak(tweak32, &r, data, data_size);
+}
+
+int secp256k1_ecdsa_sign_to_contract(const secp256k1_context* ctx, secp256k1_ecdsa_signature *signature, const unsigned char *msg32, const unsigned char *seckey, secp256k1_nonce_function noncefp, const void* noncedata, const unsigned char* s2c_data32) {
```
jonasnick: So right now you can't get the original nonce when using pure sign-to-contract, unless you do `secp256k1_ecdsa_anti_nonce_sidechan_client_commit` ?

jonasnick:
By the way feel free to squash.

src/secp256k1.c
benma: I guess I only needed this function and not the rest, though now I see the commit is not that much bigger than this (https://github.com/bitcoin-core/secp256k1/commit/0b4bef4527e2b4694a83306f495ca61d574d2c1f), so I can do it anyway

src/secp256k1.c
benma: Yeah, Part of it is that it seemed redundant, and part of it is that I am not very knowledgeable about sign-to-contract, e.g. what other uses of it there are where one would not use something akin to `secp256k1_ecdsa_anti_nonce_sidechan_client_commit` beforehand anyway.

I can add an opening struct here too if you think it will be helpful, let me know.

benma:
> It looks like this is missing a verification function for ecdsa sign-to-contract.

> I know that this PR is an experiment but to move forward this PR needs API tests and functionality tests that should cover every reachable line.

I am very much interested in moving this from an experiment to production, so I will definitely add those. I will only get a chance to work on it after next week.

Should I move all of this over to a new module? I assume modifications directly to `secp256k1_ecdsa_sign` should be avoided, right?

src/secp256k1.c
jonasnick: I don't know a specific application for s2c besides the anti nonce sidechannel and committing to the signer set in threshold signatures. I'd add an opening struct argument.

jonasnick:
> Should I move all of this over to a new module?

I think that's a good idea. Having this as an experimental module would probably reduce the barrier for getting merged

benma:
@jonasnick I pushed a re-do, please take a look.

1. Since most of my PR is based on your Schnorr PR and your help, I added your name to the copyright header. Let me know if you are okay with this.
1. It now lives in the ecdsa_sign_to_contract module: `./configure --enable-module-ecdsa-sign-to-contract --enable-experimental`
1. Using "covert channel" instead of "side channel" now according to your comment in your PR.
1. I cherry-picked two of your commits to pull in the opening struct/functions and `secp256k1_ec_commit_seckey`
1. Tests added for commit/verify and the anti nonce covert channel workflow
1. `secp256k1_ecdsa_s2c_anti_nonce_covert_channel_host_commit` is literally the same as the one in the schnorr module, but copied since it's a module function (can think adding a module_common.c or similar at some point)

:warning: my biggest question mark is: the host verify function is slightly different to the Schnorr version: it checks that the x coordinate of `R1 + H(R1, k2)` matches `r` of the signature, whereas in in the Schnorr module, the check is for the full `R`, not just the `x`-coordinate. 
Rationale: reconstructing `R` from `(r,s)` is a [lot more involved](https://github.com/bitcoin-core/secp256k1/blob/master/src/ecdsa_impl.h#L221-L227) in ecdsa, and also needs the message, which the host does not need to know otherwise. Also, [Stepan's article](https://medium.com/cryptoadvance/hardware-wallets-can-be-hacked-but-this-is-fine-a6156bbd199) describes the verification like this.

Is this sufficient for security? Otherwise, we'd need to do one of:
- calculate the full `R` anyway like linked above
- encode the sign of the `R` (not the original pubnonce `R1`) in `opening->nonce_is_negated`
- ?

src/modules/ecdsa_sign_to_contract/main_impl.h
```diff
@@ -0,0 +1,230 @@
...
+ * Distributed under the MIT software license, see the accompanying   *
+ * file COPYING or http://www.opensource.org/licenses/mit-license.php.*
+ **********************************************************************/
+
+#ifndef SECP256K1_MODULE_ECDSA_SIGN_TO_CONTRACT_MAIN_H
+#define SECP256K1_MODULE_ECDSA_SIGN_TO_CONTRACT_MAIN_H
+
+#include "include/secp256k1_ecdsa_sign_to_contract.h"
+
+int secp256k1_ecdsa_s2c_sign(
```
jonasnick: The previous version in this PR was easier to review. I don't remember how exactly it looked like but you'd get there by doing something like this: 
* move this function to `secp256k1.c` and rename it to `secp256k1_ecdsa_sign_helper`. 
* change regular `ecdsa_sign` function to just call `ecdsa_sign_helper` function but with `s2c_data` and `s2c_opening` set to NULL.

Then we wouldn't have all this code duplication. The downside is that we wouldn't be able to fully test  `ecdsa_sign_helper` without enabling the s2c module.

benma: I copied it to the module with the goal of not touching `secp256k1_ecdsa_sign`. I thought that was the goal, otherwise wouldn't this nullify the purpose of an experimental module and increase the barrier to merge? 

For comparison, [the recoverable module also copied the whole function](https://github.com/bitcoin-core/secp256k1/blob/fa3301713549d118e57ebe6551d062903ddd6b63/src/modules/recovery/main_impl.h#L123) with the minimal change of extracting the `recid`.

Here, it is also a copy except for the `if(s2c_data32 != NULL) {` clauses and s2c_data32 related arg checks.

Imho it would make sense to postpone this refactoring to a future PR, and also cover the recoverable module. Since the tests for the helper and the main signing functions would overlap  >90%, the tests should also be refactored to not copy all the tests with slightly different call signature.

include/secp256k1_ecdsa_sign_to_contract.h
```diff
@@ -0,0 +1,129 @@
...
+#define SECP256K1_ECDSA_ANTI_COVERT_CHANNEL_H
+
+#include "secp256k1.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** Same as secp256k1_ecdsa_sign, but s2c_data32 is committed to by adding `hash(R1, s2c_data32)` to
+ *  the nonce generated by noncefp, with `ndata=hash(s2c_data32, ndata)`.
```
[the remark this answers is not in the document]
benma: Are we sure that there is no use case for `ndata` when you could also provide `s2c_data`? I am not familiar with uses of `ndata` in general, but if `s2c_data` alone is enough for all cases, then `ndata` could also be removed from `secp256k1_schnorrsig_sign`?

I can remove `noncefp`, but we could also keep it so other functions could be allowed in the future without breaking API compatibility.

include/secp256k1_ecdsa_sign_to_contract.h
```diff
@@ -0,0 +1,129 @@
...
+ *  Returns: 1: signature created
+ *           0: the nonce generation function failed, or the private key was invalid.
+ *  Args:    ctx:  pointer to a context object, initialized for signing (cannot be NULL)
+ *  Out:     sig:  pointer to an array where the signature will be placed (cannot be NULL)
+ *   s2c_opening:  pointer to an secp256k1_s2c_opening structure which can be
+ *                 NULL but is required to be not NULL if this signature creates
+ *                 a sign-to-contract commitment (i.e. the `s2c_data` argument
+ *                 is not NULL). nonce_is_negated is always 0 for ecdsa.
```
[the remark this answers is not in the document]
benma: removed mention of it

src/modules/ecdsa_sign_to_contract/main_impl.h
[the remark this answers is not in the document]
benma: imho long lines are less readable, but consistency first :+1: fixed

src/modules/ecdsa_sign_to_contract/main_impl.h
```diff
@@ -0,0 +1,230 @@
...
+    }
+    if (ret) {
+        secp256k1_ecdsa_signature_save(signature, &r, &s);
+    } else {
+        memset(signature, 0, sizeof(*signature));
+    }
+    return ret;
+}
+
+int secp256k1_ecdsa_s2c_verify_commit(
```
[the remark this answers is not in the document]
benma: fixed

src/modules/ecdsa_sign_to_contract/main_impl.h
```diff
@@ -0,0 +1,230 @@
...
+    secp256k1_scalar sigr, sigs;
+
+    if (!secp256k1_ec_commit(ctx, &commitment, &opening->original_pubnonce, data32, 32)) {
+        return 0;
+    }
+
+    /* Check that sigr (x coordinate of R) matches the x coordinate of the commitment. */
+    secp256k1_ecdsa_signature_load(ctx, &sigr, &sigs, sig);
+
+    secp256k1_pubkey_load(ctx, &commitment_ge, &commitment);
```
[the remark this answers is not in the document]
benma: fixed, though the negative branch cannot be hit I think, as `secp256k1_ec_commit()` always produces a valid pubkey.

src/modules/ecdsa_sign_to_contract/main_impl.h
```diff
@@ -0,0 +1,230 @@
...
+    secp256k1_fe_normalize(&commitment_ge.x);
+    secp256k1_fe_get_b32(b, &commitment_ge.x);
+    secp256k1_scalar_set_b32(&computed_r, b, NULL);
```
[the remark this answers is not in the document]
benma: Done, went with `memcmp`

benma:
Thanks again! I split it into two commits (s2c / nonce covert channel funcs) and added a lot of tests. There are very few lines which I'm not sure how to exercise, e.g. how to get a zero/overflowing nonce with `secp256k1_nonce_function_default`, but other than that everything should be covered.

include/secp256k1_ecdsa_sign_to_contract.h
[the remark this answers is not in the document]
jonasnick: `ndata` is for giving additional information to the nonce function. Since it is hashed with `s2c_data` I don't see how you'd transmit usable information to the nonce function in that way. `ndata` shouldn't be removed from `schnorrsig_sign` because `schnorrsig_sign` can be used with other nonce functions than just the default. Imo, simpler is better than optimizing for some unknown future usecase.

src/modules/ecdsa_sign_to_contract/main_impl.h
jonasnick: > wouldn't this nullify the purpose of an experimental module and increase the barrier to merge?

No because the point is that people can use sign-to-contract only if they pass the `--enable-experimental-modules` option. That the helper doesn't change ecdsa signing if s2c_data is not set is much easier to check in review than having to diff ecdsa_sign and ecdsa_s2c_sign.

@sipa Summarizing above discussion, do you prefer copying the ecdsa code into the ecdsa_s2c module or should we make the main ecdsa function accept `s2c_data` but don't expose the argument in secp256k1.h - only in secp256k1_ecdsa_sign_to_contract?

jonasnick:
> how to get a zero/overflowing nonce with secp256k1_nonce_function_default

Yeah, not possible. In the future we should consistently mark unreachable lines.

src/modules/ecdsa_sign_to_contract/main_impl.h
benma: Just to give a quick update: I have been a bit busy the past weeks, but I still intend to move this to production on the bitbox02 (prototype already working).

I will happily refactor the current code like you suggested, then we can compare. I expect to be able to do this two weeks from now.

In the meantime, I'd still be eager to hear @sipa's opinion on how to best structure this.

benma: I rebased this PR, and made PR to compare both alternatives: https://github.com/bitcoin-core/secp256k1/pull/669

include/secp256k1_ecdsa_sign_to_contract.h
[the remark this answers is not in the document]
benma: Dropped the `noncefp` and `ndata` params from the function in the alternative [here](https://github.com/bitcoin-core/secp256k1/pull/669). Can also do it for this version if we choose this alternative over the other.

benma:
Closing in favour of https://github.com/bitcoin-core/secp256k1/pull/669 as [discussed here](https://github.com/bitcoin-core/secp256k1/pull/637#discussion_r332613869). Thanks for the detailed review!

PR #638, opened 2019-06-12
Function to compute optimal ecmult_multi scratch  size for a number of points

jonasnick:
@DavidBurkett requested to allow computing the optimal scratch size for Schnorr batch verification (https://github.com/ElementsProject/secp256k1-zkp/issues/69). This PR is a prerequisite but also contains a bunch of other fixups.

The biggest part of this PR is to make computing the scratch size and its inverse more precise by not assuming maximum padding when aligning, but rather using the actual padding. This is not strictly necessary but removes a leaky abstraction and makes testing easier.


src/ecmult_impl.h
```diff
@@ -640,12 +640,18 @@ static void secp256k1_ecmult(const secp256k1_ecmult_context *ctx, secp256k1_gej
...
 #ifdef USE_ENDOMORPHISM
-    static const size_t point_size = (2 * sizeof(secp256k1_ge) + sizeof(secp256k1_gej) + sizeof(secp256k1_fe)) * ECMULT_TABLE_SIZE(WINDOW_A) + sizeof(struct secp256k1_strauss_point_state) + sizeof(secp256k1_gej) + sizeof(secp256k1_scalar);
+    size += ROUND_TO_ALIGN(n_points * 2 * ECMULT_TABLE_SIZE(WINDOW_A) * sizeof(secp256k1_ge));
 #else
-    static const size_t point_size = (sizeof(secp256k1_ge) + sizeof(secp256k1_gej) + sizeof(secp256k1_fe)) * ECMULT_TABLE_SIZE(WINDOW_A) + sizeof(struct secp256k1_strauss_point_state) + sizeof(secp256k1_gej) + sizeof(secp256k1_scalar);
+    size += ROUND_TO_ALIGN(n_points * ECMULT_TABLE_SIZE(WINDOW_A) * sizeof(secp256k1_ge));
 #endif
-    return n_points*point_size;
+    size += ROUND_TO_ALIGN(n_points * sizeof(struct secp256k1_strauss_point_state));
+    return size;
```
real-or-random: Hm, I'm somewhat unsure about this. It seems like a layer violation to care about the alignment here.

src/ecmult_impl.h
```diff
@@ -696,7 +702,9 @@ static int secp256k1_ecmult_strauss_batch_single(const secp256k1_callback* error
 }
 
 static size_t secp256k1_strauss_max_points(const secp256k1_callback* error_callback, secp256k1_scratch *scratch) {
-    return secp256k1_scratch_max_allocation(error_callback, scratch, STRAUSS_SCRATCH_OBJECTS) / secp256k1_strauss_scratch_size(1);
+    /* Call max_allocation with 0 objects because we've already accounted for
+     * alignment in strauss_scratch_size. */
+    return secp256k1_scratch_max_allocation(error_callback, scratch, 0) / secp256k1_strauss_scratch_size(1);
```
real-or-random: ... and wasn't the previous version (without changes in `strauss_scratch_size`) more precise?
We need to round up to the alignment only once per array (e.g., once for the scalars array). 

In the proposed revision, I think we overestimate the required padding a lot because we still call `strauss_scratch_size(1)` here but this has padding now.

src/ecmult_impl.h
```diff
@@ -1267,4 +1267,20 @@ static int secp256k1_ecmult_multi_var(const secp256k1_callback* error_callback,
...
+static size_t secp256k1_ecmult_multi_scratch_size(size_t n_points) {
+    if (n_points > ECMULT_MAX_POINTS_PER_BATCH) {
+        n_points = ECMULT_MAX_POINTS_PER_BATCH;
+    }
+    if (n_points >= ECMULT_PIPPENGER_THRESHOLD) {
+        int bucket_window = secp256k1_pippenger_bucket_window(n_points);
+        return secp256k1_pippenger_scratch_size(n_points, bucket_window);
+    } else {
+        return secp256k1_strauss_scratch_size(n_points);
+    }
```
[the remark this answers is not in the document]
jonasnick: What's an approach ACK?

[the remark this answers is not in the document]
jonasnick: Oh, I guess it's a concept ack. I was confused by other meanings of approach :D

src/ecmult_impl.h
jonasnick: This is correct - sorry I overlooked this. Will add fix and test.

src/ecmult_impl.h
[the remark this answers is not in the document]
real-or-random: I'm actively testing https://github.com/bitcoin/bitcoin/pull/16149 here

src/ecmult_impl.h
jonasnick: If we want to have a function that returns the required scratch space given a number of points (and we should) then we need to add the padding somewhere. While we can assume the worst case padding somewhere else I would prefer to have the `*_scratch_space` function return exact results. This makes it much easier to think about and also helps testing because now we can just check that what is allocated actually matches what we computed with `*_scratch_space` (see https://github.com/bitcoin-core/secp256k1/pull/638/commits/24553bfa1c34bcbe4820a783d66649b0a18affb7#diff-4655d106bf03045a3a50beefc800db21R2996). Do you have an alternative in mind?

src/ecmult_impl.h
```diff
@@ -1081,27 +1124,39 @@ static int secp256k1_ecmult_pippenger_batch_single(const secp256k1_callback* err
...
+         * space for the base point G. It's an upper bound because alignment is
+         * not taken into account. */
+        n_points = (space_for_points - entry_size)/entry_size;
+        if (n_points > 0
+              && space_for_points < secp256k1_pippenger_scratch_size_points(n_points, bucket_window, 1)) {
+            /* If there's not enough space after alignment is taken into
+             * account it suffices to decrease n_points by one. This is because
+             * the maximum padding required is less than an entry. */
+            n_points -= 1;
+            VERIFY_CHECK(space_for_points >= secp256k1_pippenger_scratch_size_points(n_points, bucket_window, 1));
```
real-or-random: After some discussion with @jonasnick, one of the things I'm not sure about is the added complexity in this function.
On the one hand, this function is accurate now and users of the function can rely on that. 
On the other hand, if we just call `secp256k1_scratch_max_allocation` with `PIPPENGER_SCRATCH_OBJECTS` instead of 0, we may return a value that is one too small. That's not terrible for performance but it potentially makes the function a little bit harder to use and test because you may need to remember that it is not accurate.

jonasnick: It seems like both hands are arguments in favor of the change  (i.e. calling `secp256k1_scratch_max_allocation` with `0`).

We need to do that for strauss anyway because otherwise ```
n_points == strauss_max_points(..., scratch_create(strauss_scratch_size(n_points)))
```
wouldn't hold.

**Comment by jonasnick** [comment]
I made a couple of changes and in order to avoid adding code that is deleted in later commits I force pushed, sorry. Summary of the changes:
* added function `alloc_size` to scratch space to compute actual size allocated for a given number of objects
* fixed bug in strauss_max_points (thanks @real-or-random) that vastly underestimated the number of points actually fitting into the scratch space. Also added test which would have caught this issue.
* added a verify check to ensure that the space required for a single point/entry is smaller than the worst case padding

File: src/ecmult_impl.h (hunk 2 above)
Comment (jonasnick) [reply src/ecmult_impl.h]: fixed

File: src/ecmult_impl.h (hunk 1 above)
Comment (jonasnick) [reply src/ecmult_impl.h]: Added function `alloc_size` to scratch space

File: src/ecmult_impl.h [hunk 5]
```diff
@@ -996,6 +996,24 @@ static size_t secp256k1_pippenger_scratch_size(size_t n_points, int bucket_windo
...
+    *state_space = (struct secp256k1_pippenger_state *) secp256k1_scratch_alloc(error_callback, scratch, sizeof(struct secp256k1_pippenger_state));
+    if (*points == NULL || *scalars == NULL || *state_space == NULL) {
+        secp256k1_scratch_apply_checkpoint(error_callback, scratch, scratch_checkpoint);
+        return 0;
+    }
+    (*state_space)->ps = (struct secp256k1_pippenger_point_state *) secp256k1_scratch_alloc(error_callback, scratch, entries * sizeof(struct secp256k1_pippenger_point_state));
+    (*state_space)->wnaf_na = (int *) secp256k1_scratch_alloc(error_callback, scratch, entries * WNAF_SIZE(bucket_window+1) * sizeof(int));
+    *buckets = (secp256k1_gej *) secp256k1_scratch_alloc(error_callback, scratch, sizeof(secp256k1_gej) << bucket_window);
+    if ((*state_space)->ps == NULL || (*state_space)->wnaf_na == NULL || *buckets == NULL) {
+        secp256k1_scratch_apply_checkpoint(error_callback, scratch, scratch_checkpoint);
```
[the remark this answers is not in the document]
jonasnick: fixed

src/ecmult_impl.h
```diff
@@ -1089,19 +1164,17 @@ static size_t secp256k1_pippenger_max_points(const secp256k1_callback* error_cal
...
-        if (space_overhead > max_alloc) {
+        space_constant = secp256k1_pippenger_scratch_size_constant(bucket_window);
+        if (space_constant + entry_size > max_alloc) {
             break;
         }
-        space_for_points = max_alloc - space_overhead;
+        space_for_points = max_alloc - space_constant;
 
-        n_points = space_for_points/entry_size;
+        n_points = (space_for_points - entry_size)/entry_size;
```
sipa: Is this right? It's equivalent to `space_for_points / entry_size - 1`.

sipa:
Concept ACK, I still need to go over the logic changes.

src/ecmult_impl.h
jonasnick: It is right. Simplified the line according to your suggestion.

real-or-random:
needs rebase

jonasnick:
Rebased and polished quite a bit. Also added fix for bug in master that we noticed before iirc. So to make sure it gets in I opened #1004.

Still, I didn't fully try to understand how this PR works. Also, it seems like `ecmult_multi_scratch_size` doesn't give the exact optimal result. That's because a scratch space of size `pippenger_scratch_size(n_points, bucket_window)` it may happen that `strauss_max_points(error_callback, scratch), n)` (the actual batch size) is smaller than `n_points`.

src/ecmult_impl.h
```diff
@@ -630,22 +671,75 @@ SECP256K1_INLINE static void secp256k1_ecmult_endo_split(secp256k1_scalar *s1, s
...
+ * Returns the scratch size required for a given number of points excluding
+ * base point G and excluding the parts not dependent on the number of points.
+ * If called with 0 n_points it'll return the size required for the base point
+ * G. If the as_allocated argument is 0, this function will return the sum of
+ * the sizes of the individual parts. Otherwise it will return the size that is
+ * actually allocated with secp256k1_scratch_alloc which is greater or equal.
+ */
+static size_t secp256k1_pippenger_scratch_size_points(size_t n_points, int bucket_window, int as_allocated) {
+    size_t entries = secp256k1_pippenger_entries(n_points);
+    /* 4 objects are accounted for in pippenger_scratch_size_constant */
```
robot-dreams, line 717: Nit: 2 objects?

src/ecmult_impl.h
robot-dreams: Is this assignment to `n_points` (along with the comment) redundant with above?

src/ecmult_impl.h
```diff
@@ -801,13 +820,17 @@ static int secp256k1_ecmult_pippenger_batch_single(const secp256k1_callback* err
     return secp256k1_ecmult_pippenger_batch(error_callback, scratch, r, inp_g_sc, cb, cbdata, n, 0);
 }
 
-/**
- * Returns the maximum number of points in addition to G that can be used with
- * a given scratch space. The function ensures that fewer points may also be
- * used.
+/* Returns the (near) maximum number of points in addition to G that can be
```
robot-dreams, line 823: Do you already know how it might fail to be a maximum? (No worries if not, I still want to revisit these details carefully.)

robot-dreams (commented):
Concept ACK

src/ecmult_impl.h
```diff
@@ -718,31 +820,51 @@ static int secp256k1_ecmult_pippenger_batch_single(const secp256k1_callback* err
...
+        /* Compute an upper bound for the number of points after subtracting
+         * space for the base point G. It's an upper bound because alignment is
+         * not taken into account. */
+        n_points = (space_for_points - entry_size)/entry_size;
+        if (n_points > 0
+              && space_for_points < secp256k1_pippenger_scratch_size_points(n_points, bucket_window, 1)) {
+            /* If there's not enough space after alignment is taken into
+             * account, it suffices to decrease n_points by one. This is because
+             * the maximum padding required is less than an entry. */
+            n_points -= 1;
```
robot-dreams, line 863: Do you know how it's possible to hit this branch in a test (I wasn't able to do so even by trying a huge number of scratch sizes)?

src/ecmult_impl.h
```diff
@@ -396,8 +437,27 @@ static int secp256k1_ecmult_strauss_batch_single(const secp256k1_callback* error
...
+     * worst case padding but in this function we want to be exact. */
+    size_t max_alloc = secp256k1_scratch_max_allocation(error_callback, scratch, 0);
+    size_t unpadded_single_size = secp256k1_strauss_scratch_size_raw(1, 0);
+    size_t n_points = max_alloc / unpadded_single_size;
+    if (n_points > 0
+              && max_alloc < secp256k1_strauss_scratch_size(n_points)) {
+        /* If there's not enough space after alignment is taken into
+         * account, it suffices to decrease n_points by one. This is because
+         * the maximum padding required is less than an entry. */
+        n_points -= 1;
```
robot-dreams, line 455: Do you know how it's possible to hit this branch in a test (I wasn't able to do so even by trying a huge number of scratch sizes)?

src/ecmult_impl.h
```diff
@@ -718,31 +820,51 @@ static int secp256k1_ecmult_pippenger_batch_single(const secp256k1_callback* err
...
 
     for (bucket_window = 1; bucket_window <= PIPPENGER_MAX_BUCKET_WINDOW; bucket_window++) {
         size_t n_points;
         size_t max_points = secp256k1_pippenger_bucket_window_inv(bucket_window);
         size_t space_for_points;
-        size_t space_overhead;
-        size_t entry_size = sizeof(secp256k1_ge) + sizeof(secp256k1_scalar) + sizeof(struct secp256k1_pippenger_point_state) + (WNAF_SIZE(bucket_window+1)+1)*sizeof(int);
+        size_t space_constant;
+        /* Compute entry size without taking alignment into account */
+        size_t entry_size = secp256k1_pippenger_scratch_size_points(0, bucket_window, 0);
```
robot-dreams, line 843: Style nit (feel free to ignore): Should this be called `point_size` instead to avoid confusion, since in other places you get 2(n+1) `entries` from the endomorphism?

robot-dreams (commented):
Looks good overall.

It's unfortunate that:
- Padding / alignment adds so much complexity to these calculations
- Pippenger sizes have this weird non-monotonic behavior

But I still think your change makes sense.

My only general feedback is that updating the scratch space usage would involve keeping a lot of different things in sync, similar to what @real-or-random mentioned at https://github.com/bitcoin-core/secp256k1/pull/1004#issuecomment-963010259. Is there a way to refactor or add comments to make the task easier in the future (e.g. by sharing code between `scratch_size_raw` and `batch_allocate`)?

jonasnick:
I rebased this to see how master affects this PR. Will still need to address review comments and add better explanations to the commits.

src/scratch_impl.h
```diff
@@ -96,4 +96,14 @@ static void *secp256k1_scratch_alloc(const secp256k1_callback* error_callback, s
     return ret;
 }
 
+static size_t secp256k1_scratch_alloc_size(size_t *sizes, size_t n_sizes) {
+    size_t i;
+    size_t sum = 0;
+
+    for (i = 0; i < n_sizes; i++) {
+        sum += ROUND_TO_ALIGN(sizes[i]);
```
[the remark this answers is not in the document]
jonasnick: I think I fixed this. Also added test.

src/ecmult_impl.h
jonasnick: fixed

fjahr:
I only found this randomly just a few days ago and compared the overlapping code to https://github.com/bitcoin-core/secp256k1/pull/1789. I found a small bug in my pippenger size calculation when comparing that to the code here. I guess this approach may not be so relevant anymore with the role of scratch space being in question now but I am happy to evaluate anything from this PR to be included in or aligned with #1789 if anyone thinks there is still something interesting here.

PR #639, opened 2019-06-13
Adding message length to the callbacks

elichai:
This will help a lot for downstream code that doesn't have easy access to libc(`strlen` etc.)

Any thoughts about this?

cc downstream https://github.com/rust-bitcoin/rust-secp256k1/pull/115

(I wasn't sure when is `CHECK` vs `VERIFY_CHECK` is used so would appreciate a feedback if I used it right)

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -272,7 +272,7 @@ SECP256K1_API void secp256k1_context_destroy(
  */
 SECP256K1_API void secp256k1_context_set_illegal_callback(
     secp256k1_context* ctx,
-    void (*fun)(const char* message, void* data),
+    void (*fun)(const char* message, size_t message_len, void* data),
     const void* data
 ) SECP256K1_ARG_NONNULL(1);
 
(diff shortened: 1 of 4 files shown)


real-or-random:
Ah, I considered this too in rust-bitcoin/rust-secp256k1#115. :)

The problem with this approach is that
1. It adds a dependency to strlen(). I think with all our recent efforts for supporting embedded systems, this part of the PR is rather a step back.
 2. It breaks the callback API, see https://github.com/bitcoin-core/secp256k1/pull/595 on discussions about whether it is okay to break this or not.

The first point is easily solved by implementing strlen on our own, it's just three lines. 
The second point is difficult.

My take is rather that downstream should implement strlen if it requires it. It's also just three lines there.

elichai:
hmm shouldn't embedded systems has some sort of a `libc`?
Altough I can see why people would want to compile to embedded without linking to anything.

if we need to implement that anyway then I guess downstream could implement this as easily as us.

real-or-random:
> Unless someone else will have a different opinion until tomorrow i'll close this PR.

Yeah or maybe wait somewhat longer, my single opinion is not authoritative of course (nor is anyone else's.)

PR #641, opened 2019-06-24
Adding random tests against a naive implementation

elichai:
Hi,
As suggested in #635
This adds ECDSA Signing + Verifying tests against a naive rust implementation (https://github.com/elichai/ecc-secp256k1)

What I added:

- New configure flag called `rust-naivetests`
- Cloning and building `ecc-secp256k1` from sources in the Makefile.
- Signing(ECDSA) with this library and verifying in `ecc-secp256k1`.
- Signing(ECDSA) with `ecc-secp256k1` and verifying with this library


.gitignore
```diff
@@ -48,3 +48,4 @@ build-aux/compile
 build-aux/test-driver
 src/stamp-h1
 libsecp256k1.pc
+ecc-secp256k1
```
real-or-random: nit: GitHub says "No newline at end of file"

Makefile.am
```diff
@@ -161,8 +161,9 @@ gen_%.o: src/gen_%.c
 $(gen_context_BIN): $(gen_context_OBJECTS)
 	$(CC_FOR_BUILD) $(CFLAGS_FOR_BUILD) $(LDFLAGS_FOR_BUILD) $^ -o $@
 
+ECC_SECP=libecc_secp256k1.a
```
real-or-random: nit: `ECC` does not say much, easily confused. Maybe `NAIVE` or similar?

Makefile.am
```diff
@@ -181,3 +182,12 @@ endif
...
 endif
+
+
+if ENABLE_RUST_NAIVETESTS
+$(ECC_SECP):
+	git clone https://github.com/elichai/ecc-secp256k1.git -b v0.1.0 2> /dev/null | true
+	cd ecc-secp256k1 && cargo build --release --features=ffi
+	cp ./ecc-secp256k1/target/release/libecc_secp256k1.a  .
+	cp ./ecc-secp256k1/ecc_secp256k1.h  ./src/
+endif
```
real-or-random: newline at end of file

Makefile.am
```diff
@@ -181,3 +182,12 @@ endif
 if ENABLE_MODULE_RECOVERY
 include src/modules/recovery/Makefile.am.include
 endif
+
+
+if ENABLE_RUST_NAIVETESTS
+$(ECC_SECP):
+	git clone https://github.com/elichai/ecc-secp256k1.git -b v0.1.0 2> /dev/null | true
```
jonasnick: Makefile needs to check that this matches a fixed hash. Otherwise malicious code can be easily added to your repo which would then be run on our machines.

nit: ./configure could check if git and cargo are available.

Makefile.am
real-or-random: This needs additional rules for `make clean`.

Makefile.am
```diff
@@ -181,3 +182,12 @@ endif
...
+	cd ecc-secp256k1 && cargo build --release --features=ffi
+	cp ./ecc-secp256k1/target/release/libecc_secp256k1.a  .
+	cp ./ecc-secp256k1/ecc_secp256k1.h  ./src/
```
jonasnick: Copying libraries and header files around will be confusing. Can't you avoid it by directly linking to the files in the `ecc-secp256k1` dir?

elichai: I thought it was simpler by yeah it's possible to link directly into the files there.
about the header, don't you want it to be under source control here?

Makefile.am
elichai: hmm so you suggest to hash the whole directory?

elichai: I'll look into checking that git and cargo are available, i'm pretty new to autotools so this was mostly me shooting in the dark :)

elichai:
The main reason I wrote that library was to learn so I tried to minimize dependencies and write everything myself,
I'll try to implement hmac myself there for the deterministic nonce later this week.

And then I'll add more tests in a later PR

Makefile.am
real-or-random: Maybe it would be simpler to just vendor the library, i.e., copy the source code into a subfolder.

.gitignore
elichai: Didn't know newline at the end is the standard. fixed.

Makefile.am
elichai: fixed.

Makefile.am
elichai: fixed.

elichai: fixed.
Do you think we should use `cargo clean` for cleanup or manually delete the `target/` dir?

elichai: or should it just remove the whole `ecc-secp256k1` directory?

elichai:
Two problems I have right now that I need to debug:
1. When I disable these tests I don't have a target for `$(NAIVE_SECP)` so it fails.
2. It can try to link against `libecc_secp256k1.a` before it finishes building it.

elichai:
Fixed almost all the tests.
What should I do with `distcheck`? https://travis-ci.org/elichai/secp256k1/jobs/549941744#L1126
And still not sure why the `HOST` are failing

elichai:
Implemented deterministic nonce in my implementation so in a future PR i'll add a bit by bit signature comparison

real-or-random:
distcheck: Probably depends on whether we want to vendor (copy to our tree) the files

real-or-random:
I have no idea why configure fails on the HOST (cross-compiling) builds; this works for me locally.

One issue: Your crate has a few dependencies, so we pull all of these whenever we build the test. That's potentially dangerous.

elichai:
> I have no idea why configure fails on the HOST (cross-compiling) builds; this works for me locally.
> 
> One issue: Your crate has a few dependencies, so we pull all of these whenever we build the test. That's potentially dangerous.

Hmm, technically the dependencies get verified against the hashes in `Cargo.lock`.
I can work on decreasing the dependencies (right now it's just sha2 and a big integer library, so I could work on implementing those myself(actually on my TODO, and I started implementing a U256 type but it isn't as easy as I thought))

And that's kind of the situation right now in rust in general.

elichai:
@real-or-random I removed almost all the dependencies in my library.
I currently only have https://crates.io/crates/rug.
If i'll have the time to implement the big number part myself I'll remove that too. 
But even now it's just these:
```
$ cargo clean && cargo build
Compiling libc v0.2.58
Compiling rug v1.4.0
Compiling ecc-secp256k1 v0.2.0 (/media/hdd/CLionProjects/elichai/ecc-secp256k1)
Compiling dirs v1.0.5
Compiling gmp-mpfr-sys v1.1.13
Finished release [optimized] target(s) in 3.96s
```

real-or-random:
That's super nice. I'm not saying that this is a strict requirement -- but I won't stop your motivation for removing them.  :)

@sipa @gmaxwell @apoelstra 
Can you comment on this approach in general to make sure @elichai is not wasting time going into the wrong direction here?

jonasnick:
Closing this for now. Naive cross-testing of ecdsa and BIP 340 is already covered by CryptoFuzz (and Wycheproof). Also, if we add cross-testing then it would be much preferable to have a C implementation (which contradicts what I said above, sorry).

PR #644, opened 2019-07-02
Avoid optimizing out a verify_check

### LINKED ISSUE #643: Null checks before dereferencing a pointer
Hi,
As far as I could see this doesn't currently affect anything in the code but there's a pointer dereference here:
https://github.com/bitcoin-core/secp256k1/blob/master/src/ecmult_impl.h#L398

and only a few lines afterwards we check if it's not null
https://github.com/bitcoin-core/secp256k1/blob/master/src/ecmult_impl.h#L406

elichai:
Before that even on debug the compiler could've assumed `a` isn't null and optimized `VERIFY_CHECK(a != NULL);` out.
This put the dereference after the check
Resolves #643

### DIFF
--- a/src/ecmult_impl.h
+++ b/src/ecmult_impl.h
@@ -395,7 +395,7 @@ static void secp256k1_ecmult_context_clear(secp256k1_ecmult_context *ctx) {
  *    than the number of bits in the (absolute value) of the input.
  */
 static int secp256k1_ecmult_wnaf(int *wnaf, int len, const secp256k1_scalar *a, int w) {
-    secp256k1_scalar s = *a;
+    secp256k1_scalar s;
     int last_set_bit = -1;
     int bit = 0;
     int sign = 1;
@@ -408,6 +408,7 @@ static int secp256k1_ecmult_wnaf(int *wnaf, int len, const secp256k1_scalar *a, 
 
     memset(wnaf, 0, len * sizeof(wnaf[0]));
 
+    s = *a;
     if (secp256k1_scalar_get_bits(&s, 255, 1)) {
         secp256k1_scalar_negate(&s, &s);
         sign = -1;


real-or-random:
ACK, I read the code

src/ecmult_impl.h
```diff
@@ -405,6 +405,7 @@ static int secp256k1_ecmult_wnaf(int *wnaf, int len, const secp256k1_scalar *a,
     VERIFY_CHECK(0 <= len && len <= 256);
     VERIFY_CHECK(a != NULL);
     VERIFY_CHECK(2 <= w && w <= 31);
+    s = *a;
```
jonasnick: nit: this seems like a pretty random location for this line. How about putting it right above the first usage of `s`?

elichai: I tried to make the change make sense by putting it right after the checks but you're right, in the future this should make sense without knowing there was some change. i'll fix.

gmaxwell:
This description is one which would incorrectly cause a overblown reaction.  I would have described it as "prevent the compiler from optimizing out an interface invariant assertion" or similar.  "checking null pointer before dereferencing" is a description people use when fixing serious vulnerabilities.

elichai:
When I named this I tried to think of a name that won't sound alarming but still describe what's the point.
I didn't try to make this sound like a "vulnerability". it's not. 
It's so nothing I wasn't even sure if it's worth a PR and that's why I opened an issue first #643

If anyone looking at this PR thought it is a vulnerability it's my fault. suggestions for a title?

gmaxwell:
No need to apologize, you did nothing wrong. We want people to treat serious issues seriously, but an unfortunate side effect is that a little extra effort needs to go into not falsely triggering serious issue detection.   My comment had a suggestion, "avoid optimizing out a verify_check" which was actually your goal!  :)   Generally it's most informative if changes are described "why" first either describe the "how" later (or just leave it to the code itself to describe). 

To add some additional colour which might be useful to you in the future there is actually a fair amount of machinery in the codebase already just to avoid optimizing out VERIFY_CHECKs (and even some ARG_CHECKs):  Most of the external interface is annotated with __nonnull__ in order to cause the compiler to warn at compile time on the caller's code when it can prove that an argument will be null.  Unfortunately that annotation is treated like a dereference when compiling the function and will cause null checks to be treated like dead code. As a result, when compiling the library itself those annotations are left out (triggered by SECP256K1_BUILD define).

elichai:
Yeah i'm still trying to figure out all the autotools and macro configurations in this library :)

sipa:
ACK 94ae7cbf83a34456e5cad721f61ea77fcc023a3f

PR #648, opened 2019-07-12
Prevent ints from wrapping around in scratch space functions

jonasnick:
This PR increases the general robustness of scratch spaces. It does not fix an existing vulnerability because scratch spaces aren't used anywhere in master. Additionally,  it must be prevented anyway that an attacker has (indirect) control over the arguments touched in this PR.

### DIFF
--- a/src/scratch_impl.h
+++ b/src/scratch_impl.h
@@ -11,7 +11,7 @@
 #include "scratch.h"
 
 static secp256k1_scratch* secp256k1_scratch_create(const secp256k1_callback* error_callback, size_t size) {
-    const size_t base_alloc = ((sizeof(secp256k1_scratch) + ALIGNMENT - 1) / ALIGNMENT) * ALIGNMENT;
+    const size_t base_alloc = ROUND_TO_ALIGN(sizeof(secp256k1_scratch));
     void *alloc = checked_malloc(error_callback, base_alloc + size);
     secp256k1_scratch* ret = (secp256k1_scratch *)alloc;
     if (ret != NULL) {
--- a/src/scratch_impl.h
+++ b/src/scratch_impl.h
@@ -72,7 +72,14 @@ static size_t secp256k1_scratch_max_allocation(const secp256k1_callback* error_c
 
 static void *secp256k1_scratch_alloc(const secp256k1_callback* error_callback, secp256k1_scratch* scratch, size_t size) {
     void *ret;
-    size = ROUND_TO_ALIGN(size);
+    size_t rounded_size;
+
+    rounded_size = ROUND_TO_ALIGN(size);
+    /* Check that rounding did not wrap around */
+    if (rounded_size < size) {
+        return NULL;
+    }
+    size = rounded_size;
 
     if (memcmp(scratch->magic, "scratch", 8) != 0) {
         secp256k1_callback_call(error_callback, "invalid scratch space");
--- a/src/scratch_impl.h
+++ b/src/scratch_impl.h
@@ -60,6 +60,10 @@ static size_t secp256k1_scratch_max_allocation(const secp256k1_callback* error_c
         secp256k1_callback_call(error_callback, "invalid scratch space");
         return 0;
     }
+    /* Ensure that multiplication will not wrap around */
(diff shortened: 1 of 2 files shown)


src/scratch_impl.h
```diff
@@ -60,6 +60,10 @@ static size_t secp256k1_scratch_max_allocation(const secp256k1_callback* error_c
         secp256k1_callback_call(error_callback, "invalid scratch space");
         return 0;
     }
+    /* Ensure that multiplication will not wrap around */
+    if (objects > SIZE_MAX/(ALIGNMENT - 1)) {
```
real-or-random: `ALIGNMENT == 1` should be handled.

src/tests.c
```diff
@@ -400,6 +400,15 @@ void run_scratch_tests(void) {
     secp256k1_scratch_space_destroy(none, scratch);
     CHECK(ecount == 5);
 
+    /* Test that large integers do not wrap around in a bad way */
+    scratch = secp256k1_scratch_space_create(none, 1000);
+    /* try max allocation with a large number of objects */
+    CHECK(!secp256k1_scratch_max_allocation(&none->error_callback, scratch, (SIZE_MAX / (ALIGNMENT - 1)) + 1));
+    /* try allocating SIZE_MAX */
+    CHECK(ROUND_TO_ALIGN(SIZE_MAX) == 0);
```
real-or-random: Couldn't `ROUND_TO_ALIGN(SIZE_MAX) == SIZE_MAX` be correct on some platforms?

jonasnick: On which platforms?

src/scratch_impl.h
jonasnick: I don't think that's necessary but I can fix it. Any suggestion for how to handle it? Is a VERIFY_CHECK enough?

src/tests.c
real-or-random: Judging from the definition of `ROUND_TO_ALIGN`, I think this happens if and only if `ALIGNMENT == 1`.

src/scratch_impl.h
real-or-random: Yeah, I don't know either how big of  a problem it is. But it's not impossible. I can imagine that someone defines  `ALIGNMENT == 1` because, e.g., on x86, unaligned access is not an issue.

I think we should either
  * `#error` out at compile-time (better than with at runtime `VERIFY_CHECK`)
  * or just handle it. An `if (ALIGNMENT == 1)` can anyway be optimized out then by the compiler.

jonasnick: Added commit to handle that at runtime now.

src/tests.c
jonasnick: removed line and added comment

real-or-random:
If you want, you can cherry-pick https://github.com/real-or-random/secp256k1/commit/00076ab2347104f8ab6d88b8aa08865d0b3fd93b; this does not need a separate PR.

src/tests.c
```diff
@@ -400,6 +400,16 @@ void run_scratch_tests(void) {
     secp256k1_scratch_space_destroy(none, scratch);
     CHECK(ecount == 5);
 
+    /* Test that large integers do not wrap around in a bad way */
+    scratch = secp256k1_scratch_space_create(none, 1000);
+    /* try max allocation with a large number of objects */
+    CHECK(!secp256k1_scratch_max_allocation(&none->error_callback, scratch, (SIZE_MAX / (ALIGNMENT - 1)) + 1));
```
real-or-random: `ALIGNMENT` could be 1 here too, yielding UB the tests.

jonasnick:
Squashed fixups and cherry-picked @real-or-random's commit such that tests pass now even with `#define ALIGNMENT 1`

real-or-random:
ACK I have read the code and tested it, also using valgrind, both with and without ALIGNMENT=1 

(Travis is failing but just in the Java builds and this is some unrelated issue with a missing java package.)

gmaxwell:
ACK

sipa:
ACK 60f7f2de5de917c2bee32a4cd79cc3818b6a94a0

PR #655, opened 2019-08-05
jni: Use only Guava for hex encoding and decoding

real-or-random:
This removes a dependency on javax.xml.bind, which is no longer
available in JDK >= 11, see
https://openjdk.java.net/jeps/320#Java-EE-modules .


real-or-random:
This fixes Travis, whose Java builds were failing recently.

sipa:
ACK 2abcf951af6a9e8aff7398eb9588a50339b720c7, tests pass.

PR #662, opened 2019-09-05
Add ecmult_gen, ecmult_const and ecmult to benchmark

jonasnick:
I was trying to determine the impact of ecmult_gen in schnorrsig signing and noticed that there is no way to bench this right now. The new benchmarks look like this:
```
$ ./bench_ecmult
ecmult_gen: min 20.9us / avg 21.2us / max 21.7us
ecmult_const: min 63.9us / avg 64.3us / max 64.8us
ecmult 1: min 49.4us / avg 49.7us / max 50.3us
ecmult 1g: min 39.8us / avg 40.0us / max 40.3us
ecmult 2g: min 27.2us / avg 27.3us / max 27.8us
ecmult_multi 1g: min 39.8us / avg 40.0us / max 40.2us
ecmult_multi 2g: min 27.2us / avg 27.4us / max 27.7us
ecmult_multi 3g: min 22.8us / avg 22.9us / max 23.1us
ecmult_multi 4g: min 20.6us / avg 20.8us / max 21.1us
ecmult_multi 5g: min 19.3us / avg 19.5us / max 19.7us
```

(Turns out ecmult_gen is 37% of the 55.8us that schnorrsig sign takes)

### DIFF
--- a/src/bench_ecmult.c
+++ b/src/bench_ecmult.c
@@ -232,7 +232,7 @@ static void bench_ecmult_multi_teardown(void* arg, int iters) {
 
 static void generate_scalar(uint32_t num, secp256k1_scalar* scalar) {
     secp256k1_sha256 sha256;
-    unsigned char c[11] = {'e', 'c', 'm', 'u', 'l', 't', 0, 0, 0, 0};
+    unsigned char c[10] = {'e', 'c', 'm', 'u', 'l', 't', 0, 0, 0, 0};
     unsigned char buf[32];
     int overflow = 0;
     c[6] = num;
--- a/src/bench_ecmult.c
+++ b/src/bench_ecmult.c
@@ -28,19 +28,26 @@ typedef struct {
     secp256k1_gej* expected_output;
     secp256k1_ecmult_multi_func ecmult_multi;
 
-    /* Changes per test */
+    /* Changes per benchmark */
     size_t count;
     int includes_g;
 
-    /* Changes per test iteration */
+    /* Changes per benchmark iteration, used to pick different scalars and pubkeys
+     * in each run. */
     size_t offset1;
     size_t offset2;
 
-    /* Test output. */
+    /* Benchmark output. */
     secp256k1_gej* output;
 } bench_data;
 
-static int bench_callback(secp256k1_scalar* sc, secp256k1_ge* ge, size_t idx, void* arg) {
+/* Hashes x into [0, POINTS) twice and store the result in offset1 and offset2. */
+static void hash_into_offset(bench_data* data, size_t x) {
+    data->offset1 = (x * 0x537b7f6f + 0x8f66a481) % POINTS;
+    data->offset2 = (x * 0x7f6f537b + 0x6a1a8f49) % POINTS;
+}
+
+static int bench_ecmult_multi_callback(secp256k1_scalar* sc, secp256k1_ge* ge, size_t idx, void* arg) {
     bench_data* data = (bench_data*)arg;
     if (data->includes_g) ++idx;
     if (idx == 0) {
@@ -53,7 +60,7 @@ static int bench_callback(secp256k1_scalar* sc, secp256k1_ge* ge, size_t idx, vo
     return 1;
 }
 
-static void bench_ecmult(void* arg, int iters) {
+static void bench_ecmult_multi(void* arg, int iters) {
     bench_data* data = (bench_data*)arg;
 
     int includes_g = data->includes_g;
@@ -62,19 +69,18 @@ static void bench_ecmult(void* arg, int iters) {
     iters = iters / data->count;
 
     for (iter = 0; iter < iters; ++iter) {
-        data->ecmult_multi(&data->ctx->error_callback, &data->ctx->ecmult_ctx, data->scratch, &data->output[iter], data->includes_g ? &data->scalars[data->offset1] : NULL, bench_callback, arg, count - includes_g);
+        data->ecmult_multi(&data->ctx->error_callback, &data->ctx->ecmult_ctx, data->scratch, &data->output[iter], data->includes_g ? &data->scalars[data->offset1] : NULL, bench_ecmult_multi_callback, arg, count - includes_g);
         data->offset1 = (data->offset1 + count) % POINTS;
         data->offset2 = (data->offset2 + count - 1) % POINTS;
     }
 }
 
-static void bench_ecmult_setup(void* arg) {
+static void bench_ecmult_multi_setup(void* arg) {
     bench_data* data = (bench_data*)arg;
-    data->offset1 = (data->count * 0x537b7f6f + 0x8f66a481) % POINTS;
-    data->offset2 = (data->count * 0x7f6f537b + 0x6a1a8f49) % POINTS;
+    hash_into_offset(data, data->count);
 }
 
-static void bench_ecmult_teardown(void* arg, int iters) {
+static void bench_ecmult_multi_teardown(void* arg, int iters) {
     bench_data* data = (bench_data*)arg;
     int iter;
     iters = iters / data->count;
@@ -102,7 +108,7 @@ static void generate_scalar(uint32_t num, secp256k1_scalar* scalar) {
     CHECK(!overflow);
 }
 
-static void run_test(bench_data* data, size_t count, int includes_g, int num_iters) {
+static void run_ecmult_multi_bench(bench_data* data, size_t count, int includes_g, int num_iters) {
     char str[32];
     static const secp256k1_scalar zero = SECP256K1_SCALAR_CONST(0, 0, 0, 0, 0, 0, 0, 0);
     size_t iters = 1 + num_iters / count;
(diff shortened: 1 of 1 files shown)


src/bench_ecmult.c
```diff
@@ -34,15 +35,144 @@ typedef struct {
...
 
     /* Test output. */
     secp256k1_gej* output;
 } bench_data;
 
-static int bench_callback(secp256k1_scalar* sc, secp256k1_ge* ge, size_t idx, void* arg) {
+/* Hashes x into [0, POINTS) twice and store the result in offset1 and offset2. */
+static void hash_into_offset(bench_data* data, size_t x) {
+    data->offset1 = (x * 0x537b7f6f + 0x8f66a481) % POINTS;
+    data->offset2 = (x * 0x7f6f537b + 0x6a1a8f49) % POINTS;
```
elichai: where's this thing from?
I see it's also in `bench_ecmult_setup` but can't find the source of this

jonasnick: I think is this is just a standard integer hash function (akin to https://en.wikipedia.org/wiki/Universal_hashing#Hashing_integers) whose parameters sipa selected. But looking at it now it may be give poor results because `POINTS` is not prime.

gmaxwell: I don't think points needs to be prime, so long as the multipliers are coprime to it.

jonasnick: fwiw they're all coprime to POINTS.

src/bench_ecmult.c
```diff
@@ -34,15 +35,144 @@ typedef struct {
...
+        secp256k1_ecmult(&data->ctx->ecmult_ctx, &data->output[iter], &data->pubkeys_gej[(data->offset1+iter) % POINTS], &data->scalars[(data->offset2+iter) % POINTS], NULL);
+    }
+}
+
+static void bench_ecmult_1_teardown(void* arg) {
+    bench_data* data = (bench_data*)arg;
+    bench_ecmult_teardown_helper(data, &data->offset1, &data->offset2, NULL);
+}
+
+static void bench_ecmult_1g(void* arg) {
```
real-or-random:
```suggestion
static void bench_ecmult_0g(void* arg) {
```

src/bench_ecmult.c
```diff
@@ -34,15 +35,144 @@ typedef struct {
...
+    secp256k1_scalar zero;
+    size_t iter;
+
+    secp256k1_scalar_set_int(&zero, 0);
+    for (iter = 0; iter < ITERS; ++iter) {
+        secp256k1_ecmult(&data->ctx->ecmult_ctx, &data->output[iter], NULL, &zero, &data->scalars[(data->offset1+iter) % POINTS]);
+    }
+}
+
+static void bench_ecmult_1g_teardown(void* arg) {
```
real-or-random:
```suggestion
static void bench_ecmult_0g_teardown(void* arg) {
```

and so on..

src/bench_ecmult.c
```diff
@@ -34,15 +35,144 @@ typedef struct {
...
+    bench_data* data = (bench_data*)arg;
+    bench_ecmult_teardown_helper(data, NULL, NULL, &data->offset1);
+}
+
+static void bench_ecmult_2g(void* arg) {
```
real-or-random: 1g

jonasnick:
@real-or-random I don't think so (unless we change it for ecmult_multi) but it took me a while to figure this out again. I added a commit that printfs and explanation. It's not pretty, but less confusing.

jonasnick:
squashed

src/bench_ecmult.c
```diff
@@ -129,21 +241,37 @@ static void run_test(bench_data* data, size_t count, int includes_g) {
...
     /* Run the benchmark. */
-    sprintf(str, includes_g ? "ecmult_%ig" : "ecmult_%i", (int)count);
-    run_benchmark(str, bench_ecmult, bench_ecmult_setup, bench_ecmult_teardown, data, 10, count * (1 + ITERS / count));
+    sprintf(str, includes_g ? "ecmult_multi %ig" : "ecmult_multi %i", (int)count);
+    run_benchmark(str, bench_ecmult_multi, bench_ecmult_multi_setup, bench_ecmult_multi_teardown, data, 10, count * (1 + ITERS / count));
+}
+
+static void generate_scalar(uint32_t num, secp256k1_scalar* scalar) {
+    secp256k1_sha256 sha256;
+    unsigned char c[11] = {'e', 'c', 'm', 'u', 'l', 't', 0, 0, 0, 0};
```
real-or-random: I know it's just code you move around but this annoys me every time I see it...
```suggestion
    unsigned char c[10] = {'e', 'c', 'm', 'u', 'l', 't', 0, 0, 0, 0};
```

It's not a bug because `c[9]` will be initialized to 0 automatically. But it's a code smell and it can't be wrong to fix it. 

ACK for the rest.

jonasnick: Added commit with fix

real-or-random:
ACK a419f3dddfc96b0c7454985a913b932b77b94b0b  I read the code and tried it

src/bench_ecmult.c
```diff
@@ -128,65 +265,71 @@ static void run_test(bench_data* data, size_t count, int includes_g, int num_ite
...
         } else if(have_flag(argc, argv, "simple")) {
-            printf("Using simple algorithm:\n");
+            printf("Using simple combined algorithm:\n");
             data.ecmult_multi = secp256k1_ecmult_multi_var;
-            secp256k1_scratch_space_destroy(data.ctx, data.scratch);
-            data.scratch = NULL;
         } else {
             fprintf(stderr, "%s: unrecognized argument '%s'.\n", argv[0], argv[1]);
-            fprintf(stderr, "Use 'pippenger_wnaf', 'strauss_wnaf', 'simple' or no argument to benchmark a combined algorithm.\n");
+            fprintf(stderr, "Use '--help', 'pippenger_wnaf', 'strauss_wnaf', 'simple' or no argument to benchmark a combined algorithm.\n");
```
real-or-random, line 297: simply call `help` here? Also, this is slightly confusing because it does not make clear that `simple` and no argument are the same thing.

src/bench_ecmult.c
```diff
@@ -128,65 +265,71 @@ static void run_test(bench_data* data, size_t count, int includes_g, int num_ite
...
             data.ecmult_multi = secp256k1_ecmult_pippenger_batch_single;
         } else if(have_flag(argc, argv, "strauss_wnaf")) {
             printf("Using strauss_wnaf:\n");
             data.ecmult_multi = secp256k1_ecmult_strauss_batch_single;
         } else if(have_flag(argc, argv, "simple")) {
-            printf("Using simple algorithm:\n");
+            printf("Using simple combined algorithm:\n");
             data.ecmult_multi = secp256k1_ecmult_multi_var;
-            secp256k1_scratch_space_destroy(data.ctx, data.scratch);
-            data.scratch = NULL;
```
real-or-random, line 159: That line got lost, and this was what made the difference between simple and combined? I see you merged this into "simple combined" in the printfs. Was this change on purpose?

jonasnick: Good catch. I also forgot what the simple algorithm actually does. I fixed this and added an explanation to the help().

src/bench_ecmult.c
jonasnick: okay

jonasnick:
@real-or-random care to  reACK :) ?

src/bench_ecmult.c
real-or-random: Ah sorry, my suggestion was just to replace the entire `else` branch with `help()`, not to replace the `--help` by `help`. Feel free to change back, or to ignore.

real-or-random:
ACK ee83dbf267b34bb4a01c872cd02bf1c8065e3166

src/bench_ecmult.c
jonasnick: Fixed

jonasnick:
Rebased to let CI do its thing

sipa:
While we're at it: are no-g variants not interesting?

jonasnick:
Hm with-g seems more interesting (schnorrsig batch verify, taproot tweaks, pedersen commitments,...) but we could add a flag or something for no-g (preferably in a separate PR imho).

real-or-random:
ACK 8f879c2887e166da2ec959ce78078f7b84ebfdf9

elichai:
tACK 8f879c2887e166da2ec959ce78078f7b84ebfdf9

nit, It is a little confusing that the first 4 benchmarks are unrelated to the flag being passed (Maybe we want to move the "Using XXX algorithm" print to just before the relevant benchmarks?)

real-or-random:
> nit, It is a little confusing that the first 5 benchmarks are unrelated to the flag being passed (Maybe we want to move the "Using XXX algorithm" print to just before the relevant benchmarks?)

Indeed but this can be done in a separate PR, too.

Let me take the opportunity to merge this PR now that we have two ACKs.

PR #664, opened 2019-09-15
Remove mention of ec_privkey_export because it doesn't exist

### LINKED ISSUE #663: secp256k1_ec_privkey_export mentioned in comment but not defined
See https://github.com/bitcoin-core/secp256k1/blob/96cd94e385f64c1936abf0d1e303d12d0f5da980/include/secp256k1.h#L173

jonasnick:
Fixes #663 
There is `ec_privkey_export_der` but it takes `0` for uncompressed and not `SECP256K1_EC_UNCOMPRESSED` (which is `2`).


real-or-random:
ACK https://github.com/bitcoin-core/secp256k1/pull/664/commits/59782c68b41e4262f003135717705990b3fdc3ae

> There is `ec_privkey_export_der` but it takes `0` for uncompressed and not `SECP256K1_EC_UNCOMPRESSED` (which is `2`).

Right, and anyway that function is only in `contrib`.

apoelstra:
utACK https://github.com/bitcoin-core/secp256k1/commit/59782c68b41e4262f003135717705990b3fdc3ae

PR #668, opened 2019-10-07
Return 0 if invalid seckey is given to ec_privkey_negate

### LINKED ISSUE #488: secp256k1_ec_privkey_negate and invalid privkeys
For keys greater than the order,  `secp256k1_scalar_negate`  returns strange results (understandable!).

But,  as a result,  `secp256k1_ec_privkey_negate`, when called with an invalid private key, returns `1` (OK),  and a ?private key?.

Tripped me up for a bit when generating some test fixtures for `secp256k1_ec_privkey_negate`.

``` 
0000000000000000000000000000000000000000000000000000000000000000 becomes 0000000000000000000000000000000000000000000000000000000000000000
ff00000000000000000000000000000000000000000000000000000000000000 becomes 00fffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141
ffff000000000000000000000000000000000000000000000000000000000000 becomes 0000fffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141
ffffff0000000000000000000000000000000000000000000000000000000000 becomes 000000fffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141
ffffffff00000000000000000000000000000000000000000000000000000000 becomes 00000000fffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141
ffffffffff000000000000000000000000000000000000000000000000000000 becomes 0000000000fffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141
ffffffffffff0000000000000000000000000000000000000000000000000000 becomes 000000000000fffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141
ffffffffffffff00000000000000000000000000000000000000000000000000 becomes 00000000000000fffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141
ffffffffffffffff000000000000000000000000000000000000000000000000 becomes 0000000000000000fffffffffffffffebaaedce6af48a03bbfd25e8cd0364141
ffffffffffffffffff0000000000000000000000000000000000000000000000 becomes 000000000000000000fffffffffffffebaaedce6af48a03bbfd25e8cd0364141
ffffffffffffffffffff00000000000000000000000000000000000000000000 becomes 00000000000000000000fffffffffffebaaedce6af48a03bbfd25e8cd0364141
ffffffffffffffffffffff000000000000000000000000000000000000000000 becomes 0000000000000000000000fffffffffebaaedce6af48a03bbfd25e8cd0364141
ffffffffffffffffffffffff0000000000000000000000000000000000000000 becomes 000000000000000000000000fffffffebaaedce6af48a03bbfd25e8cd0364141
ffffffffffffffffffffffffff00000000000000000000000000000000000000 becomes 00000000000000000000000000fffffebaaedce6af48a03bbfd25e8cd0364141
ffffffffffffffffffffffffffff000000000000000000000000000000000000 becomes 0000000000000000000000000000fffebaaedce6af48a03bbfd25e8cd0364141
ffffffffffffffffffffffffffffff0000000000000000000000000000000000 becomes 000000000000000000000000000000febaaedce6af48a03bbfd25e8cd0364141
ffffffffffffffffffffffffffffffff00000000000000000000000000000000 becomes fffffffffffffffffffffffffffffffe755db9cd5e9140777fa4bd19a06c8282
ffffffffffffffffffffffffffffffffff000000000000000000000000000000 becomes fffffffffffffffffffffffffffffffd765db9cd5e9140777fa4bd19a06c8282
ffffffffffffffffffffffffffffffffffff0000000000000000000000000000 becomes fffffffffffffffffffffffffffffffd755eb9cd5e9140777fa4bd19a06c8282
ffffffffffffffffffffffffffffffffffffff00000000000000000000000000 becomes fffffffffffffffffffffffffffffffd755dbacd5e9140777fa4bd19a06c8282
ffffffffffffffffffffffffffffffffffffffff000000000000000000000000 becomes fffffffffffffffffffffffffffffffd755db9ce5e9140777fa4bd19a06c8282
ffffffffffffffffffffffffffffffffffffffffff0000000000000000000000 becomes fffffffffffffffffffffffffffffffd755db9cd5f9140777fa4bd19a06c8282
ffffffffffffffffffffffffffffffffffffffffffff00000000000000000000 becomes fffffffffffffffffffffffffffffffd755db9cd5e9240777fa4bd19a06c8282
ffffffffffffffffffffffffffffffffffffffffffffff000000000000000000 becomes fffffffffffffffffffffffffffffffd755db9cd5e9141777fa4bd19a06c8282
ffffffffffffffffffffffffffffffffffffffffffffffff0000000000000000 becomes fffffffffffffffffffffffffffffffd755db9cd5e9140787fa4bd19a06c8282
ffffffffffffffffffffffffffffffffffffffffffffffffff00000000000000 becomes fffffffffffffffffffffffffffffffd755db9cd5e91407780a4bd19a06c8282
ffffffffffffffffffffffffffffffffffffffffffffffffffff000000000000 becomes fffffffffffffffffffffffffffffffd755db9cd5e9140777fa5bd19a06c8282
ffffffffffffffffffffffffffffffffffffffffffffffffffffff0000000000 becomes fffffffffffffffffffffffffffffffd755db9cd5e9140777fa4be19a06c8282
ffffffffffffffffffffffffffffffffffffffffffffffffffffffff00000000 becomes fffffffffffffffffffffffffffffffd755db9cd5e9140777fa4bd1aa06c8282
ffffffffffffffffffffffffffffffffffffffffffffffffffffffffff000000 becomes fffffffffffffffffffffffffffffffd755db9cd5e9140777fa4bd19a16c8282
ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff0000 becomes fffffffffffffffffffffffffffffffd755db9cd5e9140777fa4bd19a06d8282
ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff00 becomes fffffffffffffffffffffffffffffffd755db9cd5e9140777fa4bd19a06c8382

```

jonasnick:
Fixes #488.

Instead of just fixing the documentation it's much better for misuse resistance to return 0 if the seckey overflows. I think most people who use this function and expect it to return 1 always will prefer their code to crash instead of getting a wrong result.

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -549,7 +549,10 @@ SECP256K1_API int secp256k1_ecdsa_sign(
     const void *ndata
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
 
-/** Verify an ECDSA secret key.
+/** Verify an ECDSA secret key. It is valid if it is not 0 and less than the
+ *  secp256k1 curve order when interpreted as an integer (most significant byte
+ *  first). The probability of choosing a 32-byte string at random which is an
+ *  invalid secret key is negligible.
  *
  *  Returns: 1: secret key is valid
  *           0: secret key is invalid
@@ -579,9 +582,9 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_create(
  *
  *  Returns: 1 if seckey was successfully negated and 0 otherwise
  *  Args:   ctx:        pointer to a context object
- *  In/Out: seckey:     pointer to the 32-byte private key to be negated. The private key
- *                      interpreted as an integer (most significant byte first) must be less than
- *                      the curve order. (cannot be NULL)
+ *  In/Out: seckey:     pointer to the 32-byte private key to be negated. The private
+ *                      key should be valid according to secp256k1_ec_seckey_verify
+ *                      (cannot be NULL)
  */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_privkey_negate(
     const secp256k1_context* ctx,
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -577,9 +577,11 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_create(
 
 /** Negates a private key in place.
  *
- *  Returns: 1 always
+ *  Returns: 1 if seckey was successfully negated and 0 otherwise
  *  Args:   ctx:        pointer to a context object
- *  In/Out: seckey:     pointer to the 32-byte private key to be negated (cannot be NULL)
+ *  In/Out: seckey:     pointer to the 32-byte private key to be negated. The private key
+ *                      interpreted as an integer (most significant byte first) must be less than
+ *                      the curve order. (cannot be NULL)
  */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_privkey_negate(
     const secp256k1_context* ctx,
--- a/src/tests.c
+++ b/src/tests.c
@@ -1097,16 +1097,42 @@ void run_scalar_tests(void) {
 
 #ifndef USE_NUM_NONE
     {
-        /* A scalar with value of the curve order should be 0. */
+        /* Test secp256k1_scalar_set_b32 boundary conditions */
         secp256k1_num order;
-        secp256k1_scalar zero;
+        secp256k1_scalar scalar;
         unsigned char bin[32];
+        unsigned char bin_tmp[32];
         int overflow = 0;
+        static const secp256k1_scalar all_zeros_minus_order = SECP256K1_SCALAR_CONST(
+            0x00000000UL, 0x00000000UL, 0x00000000UL, 0x00000001UL,
+            0x45512319UL, 0x50B75FC4UL, 0x402DA173UL, 0x2FC9BEBEUL
+        );
+
+        /* A scalar set to 0s should be 0. */
+        memset(bin, 0, 32);
+        secp256k1_scalar_set_b32(&scalar, bin, &overflow);
+        CHECK(overflow == 0);
+        CHECK(secp256k1_scalar_is_zero(&scalar));
+
+        /* A scalar with value of the curve order should be 0. */
         secp256k1_scalar_order_get_num(&order);
         secp256k1_num_get_bin(bin, 32, &order);
-        secp256k1_scalar_set_b32(&zero, bin, &overflow);
+        secp256k1_scalar_set_b32(&scalar, bin, &overflow);
         CHECK(overflow == 1);
-        CHECK(secp256k1_scalar_is_zero(&zero));
+        CHECK(secp256k1_scalar_is_zero(&scalar));
+
+        /* A scalar with value of the curve order minus one should not overflow. */
+        bin[31] -= 1;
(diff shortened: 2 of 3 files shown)


src/tests.c
```diff
@@ -4104,6 +4104,26 @@ void run_eckey_edge_case_test(void) {
...
+    secp256k1_rand256_test(seckey);
+    memcpy(seckey_tmp, seckey, 32);
+
+    /* Verify negation changes the key and changes it back */
+    CHECK(secp256k1_ec_privkey_negate(ctx, seckey) == 1);
+    CHECK(memcmp(seckey, seckey_tmp, 32) != 0);
+    CHECK(secp256k1_ec_privkey_negate(ctx, seckey) == 1);
+    CHECK(memcmp(seckey, seckey_tmp, 32) == 0);
+
+    /* Negating an overflowing seckey does not work */
```
gmaxwell: Consistency of the actual behaviour would be better verified by a test that sets 16 bytes as 0xFF and then sets the remaining bytes randomly, in order to detect a change to a behaviour that  always sets the output to all 0xFF.

The documentation doesn't promise one behaviour vs the other, but preserving any entropy in the bad input is the safer behaviour (e.g. the data in it might get fed elsewhere into some rng hash or something) and the test should catch if it gets changed. 

A more extensive test would also(?) check the boundary conditions (smallest non-overflowing, smallest overflowing, maximum value). I don't think that's particularly important so long as scalar-set's overflow behaviour is checked precisely else where (I didn't look).

jonasnick: I improved the test according to your suggestion. `scalar_set_b32` just uses `secp256k1_scalar_reduce` which I assume is well tested but I added a commit with boundary condition tests explicitly for `scalar_set_b32`.

include/secp256k1.h
```diff
@@ -577,9 +577,11 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_create(
 
 /** Negates a private key in place.
  *
- *  Returns: 1 always
+ *  Returns: 1 if seckey was successfully negated and 0 otherwise
  *  Args:   ctx:        pointer to a context object
- *  In/Out: seckey:     pointer to the 32-byte private key to be negated (cannot be NULL)
+ *  In/Out: seckey:     pointer to the 32-byte private key to be negated. The private key
+ *                      interpreted as an integer (most significant byte first) must be less than
+ *                      the curve order. (cannot be NULL)
```
real-or-random: less than the curve order and not be 0?

(Actually we should just document what a valid private key is, e.g., here https://github.com/bitcoin-core/secp256k1/blob/master/include/secp256k1.h#L552)

jonasnick: It won't fail for 0 though. I think ideally we document what's a valid secret key is (as you said) in the doc of `seckey_verify` and then just refer to that instead of having to restate it over and over. I can rewrite this PR to do that.

jonasnick:
I added a commit to return 0 on an all 0 key because otherwise the result is an invalid seckey. Also moved definition of valid secret keys to verify_seckey as suggested by @real-or-random

real-or-random:
ACK 3f5512c81b17f1412d7992f8fe418a1ab629d955 I read the code and tested it, also with valgrind

It should be noted that this changes the observable behavior of ec_privkey_negate, which is also used in `CKey::Negate()` in Bitcoin Core, see https://github.com/bitcoin/bitcoin/blob/d0f81a96d9c158a9226dc946bdd61d48c4d42959/src/key.cpp#L168. But `CKey::Negate()` is only used in the tests, as far as I can tell.

jonasnick:
Superseded by #701

Issue #672, opened 2019-10-11
Storing and loading the private key


kroggen:
Hi!

Congrats for the great software! It is way faster than OpenSSL.

I was not able to find how to store and load the private key. Preferably in a small binary format.

kroggen:
Analyzing the code I could notice that the private key is not stored on the context. I was used with `mbed-crypto` where it is the opposite.

So we are responsible for generating the private key with a strong RNG, checking if it is valid with the `secp256k1_ec_seckey_verify` function (or `secp256k1_ec_privkey_verify` when renamed), storing it and loading it.

A documentation on README would be helpful.

Cheers,

real-or-random:
Have you seen the header file?
https://github.com/bitcoin-core/secp256k1/blob/master/include/secp256k1.h

It has a good documentation but I admit there's no general section on usage of the library.

Related to https://github.com/bitcoin-core/secp256k1/issues/184 .

NicholasFarrow:
For anyone else interested in how to generate keys using this library, feel free to checkout my [Bitcoin vanity address generator](https://github.com/NicholasFarrow/niceBit/) which I think provides a short and [reasonably well commented example](https://github.com/NicholasFarrow/niceBit/blob/master/niceBit.c) of how to generate private and public key pairs.

PR #673, opened 2019-10-11
Add basic usage information on README [skip ci]

kroggen:
This may not be so good documentation, but at least it is a start. Feel comfortable to fix it.

#672


kroggen:
What about changing "key generation" on README to "public key generation"?

real-or-random:
Closing in favor of #748.

PR #679, opened 2019-10-28
Add SECURITY.md

### LINKED ISSUE #646: We should add a SECURITY.md
It's now good practice to have a SECURITY.md file that explains a well-defined process for reporting vulnerabilities. Core has this too:
https://github.com/bitcoin/bitcoin/blob/master/SECURITY.md

This is also a good chance to think about the process, i.e., who should actually be informed about vulnerabilities in this library. This is not completely obvious since this library somehow belongs to Bitcoin Core (I mean the software, not the "organization"/group of people) but on the other hand is maintained separately.

jonasnick:
Fixes #646 

WIP because the secp256k1-security@bitcoincore.org email address doesn't exist yet. But it seems like the right place for vulnerability reports. security@bitcoincore.org would have the downside that it perhaps reaches more people than necessary. Ideally secp256k1-security would just forward to the three maintainers listed in SECURITY.md. @sipa @apoelstra is it okay to put you there? Fwiw I'm opting out for now because three people should be enough.

@sipa do you know who to talk to about adding secp256k1-security@bitcoincore.org and the specifics about how it would work?

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,3 +70,8 @@ Exhaustive tests
 With valgrind, you might need to increase the max stack size:
 
     $ valgrind --max-stackframe=2500000 ./exhaustive_tests
+
+Reporting a vulnerability
+------------
+
+See [SECURITY.md](SECURITY.md)
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,15 @@
+# Security Policy
+
+## Reporting a Vulnerability
+
+To report security issues send an email to secp256k1-security@bitcoincore.org (not for support).
+
+The following keys may be used to communicate sensitive information to developers:
+
+| Name | Fingerprint |
+|------|-------------|
+| Pieter Wuille | 133E AC17 9436 F14A 5CF1  B794 860F EB80 4E66 9320 |
+| Andrew Poelstra | 699A 63EF C17A D3A9 A34C  FFC0 7AD0 A91C 40BD 0091 |
+| Tim Ruffing | 09E0 3F87 1092 E40E 106E  902B 33BC 86AB 80FF 5516 |
+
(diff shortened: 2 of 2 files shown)


real-or-random:
> @sipa do you know who to talk to about adding [secp256k1-security@bitcoincore.org](mailto:secp256k1-security@bitcoincore.org) and the specifics about how it would work?

sipa says @laanwj may be the right person to ping here.

By the way, this PR is consistent with what was discussed in http://www.erisian.com.au/meetbot/bitcoin-core-dev/2019/bitcoin-core-dev.2019-08-15-19.00.log.html:
```
19:19:53 <real_or_random> also e.g., I have an open issue about a security.md file
19:20:34 <real_or_random> which raises the question who should be in there. secp256k1 maintainers or bitcoin-core maintainers?
[...]
19:22:24 <wumpus> secp256k1 issues should probably be reported to secp256k1 maintainers, in general
19:22:51 <sipa> agree; though very serious issues that impact bitcoin could of course also be reported to bitcoin core
19:23:09 <real_or_random> wumpus: yes this seems sensible they can escalate to core if necessary
[...]
19:23:44 <wumpus> yes, if it affects use in bitcoin, or is even an issue that threatens bitcoin, that seems an exception
```

Maybe we should add a note to the file that issues may be escalated to Bitcoin Core asking the reporter for permission.

laanwj:
Yes, I can set up a forward `secp256k1-security@bitcoincore.org`, just let me know who to include.

elichai:
Concept ACK. I think it's about time :)

laanwj:
The e-mail forward has been added (might take some time to propagate)

real-or-random:
ACK I looked at the diff and verified my fingerprint

sipa:
ACK, verified I can receive emails, and verified my fingerprint.

apoelstra:
ACK my fingerprint is correct and I can receive email.

PR #684, opened 2019-11-01
Make no-float policy explicit

### LINKED ISSUE #683: No-float policy for secp256k1
I think secp256k1 needs an explicit no-floating point policy, like the Linux kernel. This means no use of `float`, and no  `double` (and no `half`, no `long double` either).

- It has to be able to run in kernel context
- Boot and security processors tend to have no FP support (e.g. core 0 on the Sifive U54)
- FP hardware implementation is possibly inconsistent between platforms which is useless for cryptography

I mean for the main code; in the bench framework is fine, though with the assumption that it might be emulated and very slow.

(inspired by https://github.com/bitcoin-core/secp256k1/pull/658#issuecomment-548576364)

real-or-random:
We don't want floating types for various reasons, e.g.,
 - Their representation and often their behavior is implementation-defined.
 - Many targets don't support them.

Closes #683.

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,6 +23,7 @@ Implementation details
   * Extensive testing infrastructure.
   * Structured to facilitate review and analysis.
   * Intended to be portable to any system with a C89 compiler and uint64_t support.
+  * No use of floating types, except in benchmarks.
   * Expose only higher level interfaces to minimize the API surface and improve application security. ("Be difficult to use insecurely.")
 * Field operations
   * Optimized implementation of arithmetic modulo the curve's field size (2^256 - 0x1000003D1).


laanwj:
ACK

jonasnick:
ACK bae1bea3c4b46a2fb5ca76ff6bf1e98d43cff52f

gmaxwell:
After the fact ACK.

One thing to keep in mind WRT benchmarks, is that it's my experience that many embedded developers look at a library and if they say _ANY_ float at all they'll instantly just report up the chain "nope, can't be ported, uses floating point".  ... even if that float is some build time optional thing.

[I find they do the same for malloc/free and file I/O too.]

The explicit statement in the readme is probably good enough, but if you encounter this thinking don't be too surprised by it.

laanwj:
> One thing to keep in mind WRT benchmarks, is that it's my experience that many embedded developers look at a library and if they see ANY float at all they'll instantly just report up the chain "nope, can't be ported, uses floating point".

This is the only use of double in the test framework.
```c
src/bench.h:static double gettimedouble(void) {
src/bench.h:void print_number(double x) {
src/bench.h:    double y = x;
src/bench.h:    double min = HUGE_VAL;
src/bench.h:    double sum = 0.0;
src/bench.h:    double max = 0.0;
src/bench.h:        double begin, total;
src/bench.h:        begin = gettimedouble();
src/bench.h:        total = gettimedouble() - begin;
```
It'd be pretty straightforward to change this to use 64 bit microseconds instead, and get rid of the exception. I might have a stab at it.

PR #685, opened 2019-11-02
Fix issue where travis does not show the ./tests seed…

### LINKED ISSUE #645: Better output in failed tests
When a test fails, it currently outputs only the failed assertion (with file and line) to stderr. The seed is output to stdout only. Moreover, neither stdout nor stderr are output on `make check`. Instead, the stderr is written to `./test-suite.log`

As a consequence, a non-reproducible failure on Travis is somewhat useless to us because we just see the FAIL but we have no idea what failed because we don't see the contents of the log file. (This just happened downstream in secp256k1-zkp).  A similar issue was the failed PR build on a Fedora build server, which contained only stderr https://github.com/bitcoin-core/secp256k1/pull/558#issuecomment-495379966 and thus we didn't get the seed.

We should probably
 * output the seed to stderr, at least if the test fails.
 * output the log file in travis using `after_script` or `after_failure`, see https://docs.travis-ci.com/user/job-lifecycle/#the-job-lifecycle

jonasnick:
…by removing stdout buffering and always cat tests.log after a travis run. Fixes #645. 

I noticed that according to the [doc](https://www.gnu.org/software/automake/manual/html_node/Parallel-Test-Harness.html) tests.log should contain stdout as well as stderr. But it doesn't because stdout isn't flushed. I removed buffering completely to avoid having to call `fflush` twice.

Travis is instructed to always show the seed which seems helpful with `after_script` by `cat`ing `./tests.log`. In case the tests fail it looks like https://travis-ci.org/jonasnick/secp256k1/jobs/606446234.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -5174,6 +5174,9 @@ int main(int argc, char **argv) {
      * diagnostic information. Happens right at the start of main because
      * setbuf must be used before any other operation on the stream. */
     setbuf(stdout, NULL);
+    /* Also disable buffering for stderr because it's not guaranteed that it's
+     * unbuffered on all systems. */
+    setbuf(stderr, NULL);
 
     /* find iteration count */
     if (argc > 1) {
--- a/src/tests.c
+++ b/src/tests.c
@@ -5169,6 +5169,12 @@ void run_ecdsa_openssl(void) {
 int main(int argc, char **argv) {
     unsigned char seed16[16] = {0};
     unsigned char run32[32] = {0};
+
+    /* Disable buffering for stdout to improve reliability of getting
+     * diagnostic information. Happens right at the start of main because
+     * setbuf must be used before any other operation on the stream. */
+    setbuf(stdout, NULL);
+
     /* find iteration count */
     if (argc > 1) {
         count = strtol(argv[1], NULL, 0);
--- a/.travis.yml
+++ b/.travis.yml
@@ -77,7 +77,7 @@ matrix:
         apt:
           packages:
             - valgrind
-    
+
 before_install: mkdir -p `dirname $GUAVA_JAR`
 install: if [ ! -f $GUAVA_JAR ]; then wget $GUAVA_URL -O $GUAVA_JAR; fi
 before_script: ./autogen.sh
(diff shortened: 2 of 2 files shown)


src/tests.c
```diff
@@ -5208,6 +5208,7 @@ int main(int argc, char **argv) {
     }
     secp256k1_rand_seed(seed16);
 
+    setbuf(stdout, NULL);
```
real-or-random: C89 says `setbuf` should be called before any operation is performed on the stream. This means that it is currently in the right position but I think it would be less fragile to put at the beginning of `main()`. (For example, someone could have the idea of changing the `fprintf(stderr, ...)`  in line 5195 to `stdout`.) Also, I think this call deserves a comment with a rationale.

jonasnick: Good point. I moved the call to the beginning of main and added comments.

real-or-random:
ACK 8de439261a34faca17e513c0b554fa5526d22e1e I read the code and I checked that it does not break the tests

real-or-random:
Actually this would not fix the Fedora example in #645 because we still don't output the seed to stderr in case of failure. Maybe it's better to make `make check` cat the log files to stderr in case of failure.

jonasnick:
Good point, but I'm tempted to blame this on the user (for now). With automake you could run `make check` with `VERBOSE=1` which will print the stdout and stderr of a failed test to stdout. So not capturing stdout seems to be not a good idea when you need diagnostics anyway. Setting VERBOSE=1 should be possible in a build system and is probably useful for other automake based packages too.

If I understand correctly what you're asking for is to define a custom test runner for which my automake skills are insufficient. And it could be argued that such a non-standard customization isn't particularly helpful in the long run.

real-or-random:
Agreed, we don't want to fiddle around with automake if this can't be done easily.

real-or-random:
needs rebase

jonasnick:
rebased

gmaxwell:
Very nice change. Might it makes sense to also unbuffer stderr?

real-or-random:
> Very nice change. Might it makes sense to also unbuffer stderr?

It is probably unbuffered on most systems (https://stackoverflow.com/a/3746795) but yep, it won't hurt and it's good for consistency.

jonasnick:
rebased

real-or-random:
ACK a0771d15e67d3fe6ac1791f81d9731f73c550e5e I looked at the diff and checked that it does not break the tests

PR #688, opened 2019-11-05
Fix ASM setting in travis

jonasnick:
Without this PR the `ASM` setting isn't taken into account in travis.


real-or-random:
ACK 5c5f71eea5167b0dd9dbef246fc70132c50c9af3 I read the diff

(modulo the pending travis build)

Crazy that we missed this.

jonasnick:
Looks like it's working.

Issue #680, opened 2019-10-28
Change benchmark output to CSV


real-or-random:
Not crucial but nice to have. It's as readable as the current form for engineers but much easier to use in a spreadsheet if you want to look at the changes, which we do occasionally (that's what benchmarks are for, hm?).

_Originally posted by @elichai in https://github.com/bitcoin-core/secp256k1/pull/678#issuecomment-547062283_

gmaxwell:
I really dislike the bitcoin output, I think it's very hard to read and make sense of... if someone does this at least put the most important measurements first.   I'm dubious about the value of just bulk running these as most of their performance is a consequence of random code/data alignment and such... they jump around at random unless effort is made to deal with that.

real-or-random:
> In addition to a command line and CSV output, Will a JSON output format be helpful? ([like this](https://github.com/google/benchmark/blob/main/docs/user_guide.md?rgh-link-date=2021-10-13T14%3A11%3A16Z#output-formats))

It'd nice to have if it's not much more work but I think it's a stretch goal.

PR #689, opened 2019-11-05
Remove "except in benchmarks" exception for fp math

laanwj:
Convert `bench.h` to fixed-point math, removing all use of float math from the repository:

- Use 64-bit integer microsecond timestamps
- Use decimal fixed-point math for formatting numbers

It turned out to be a little trickier than I expected because of formatting and rounding. But, output should be the same before and after.

I used the following to test the number formatting: https://gist.github.com/laanwj/f971bfbe018e39c19677a21ff954d0c7

### DIFF
--- a/src/bench.h
+++ b/src/bench.h
@@ -7,43 +7,85 @@
 #ifndef SECP256K1_BENCH_H
 #define SECP256K1_BENCH_H
 
+#include <stdint.h>
 #include <stdio.h>
 #include <string.h>
-#include <math.h>
 #include "sys/time.h"
 
-static double gettimedouble(void) {
+static int64_t gettime_i64(void) {
     struct timeval tv;
     gettimeofday(&tv, NULL);
-    return tv.tv_usec * 0.000001 + tv.tv_sec;
+    return (int64_t)tv.tv_usec + (int64_t)tv.tv_sec * 1000000LL;
 }
 
-void print_number(double x) {
-    double y = x;
-    int c = 0;
-    if (y < 0.0) {
-        y = -y;
+#define FP_EXP (6)
+#define FP_MULT (1000000LL)
+
+/* Format fixed point number. */
+void print_number(const int64_t x) {
+    int64_t x_abs, y;
+    int c, i, rounding;
+    size_t ptr;
+    char buffer[30];
+
+    if (x == INT64_MIN) {
+        /* Prevent UB. */
+        printf("ERR");
+        return;
     }
-    while (y > 0 && y < 100.0) {
-        y *= 10.0;
+    x_abs = x < 0 ? -x : x;
+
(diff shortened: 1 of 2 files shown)


real-or-random:
Nice! Now I don't want to work on the related #680 because I may end up removing parts of your excellent formatting function...

laanwj:
I searched around a bit and it doesn't seem there's currently any way to tell the compiler (gcc or otherwise) to fail on any use on floating point. So we can't easily check this policy in the CI. Not that it matters, it's a relatively small project and changes here are well-reviewed.

> Nice! Now I don't want to work on the related #680 because I may end up removing parts of your excellent formatting function...

Haha yes I leave the "how to show the benchmark" part up to you. I think there's different preferences in that regard.

I made that it's straightforward to split up in a "how many decimals to show" and generic "format fixed-point with n decimals" part.

apoelstra:
FWIW the output of `git grep 'double '` is only two lines, both of which are comments. And `git grep float` is empty. We could just have CI check this :P

gmaxwell:
Very nice.  One thing that wasn't clear to me was what is protecting overflow in the fixed point scaling?

laanwj:
> Very nice. One thing that wasn't clear to me was what is protecting overflow in the fixed point scaling?

You mean this part?
```
    printf("%s: min ", name);
    print_number(min * FP_MULT / iter);
    printf("us / avg ");
    print_number(((sum * FP_MULT) / count) / iter);
    printf("us / max ");
    print_number(max * FP_MULT / iter)
```

The value is in microseconds and is multiplied with 1e6, so if my computation is correct (`(1<<64)/1e6/1e6`) it takes 9.2 megasecond (~1/3th year) until an overflow can happen. Do you think anyone is going to wait that long for a benchmark?

It would definitely be possible to add overflow checks here but it seemed overkill.

gmaxwell:
Indeed, that seems more than adequate. I should have done the math!

laanwj:
Anything left to do here?

real-or-random:
I think there's nothing left do to here. 

But can we get a second (untested) ACK here? @gmaxwell You apparently looked at it, do you mind ACKing it?

gmaxwell:
ACK

src/bench.h
```diff
@@ -9,41 +9,82 @@
```
real-or-random:
```suggestion
#include <stdint.h>
```

This currently builds fine because `util.h` is included in every `bench_` program. But that's somewhat fragile.

laanwj: thanks, added commit

real-or-random:
ACK bde2a32286c697dd1056aa3eb1ea2a5353f0bede I've read the code in detail and I've tested it. I haven't explicitly tested the formatting function with known/hardcoded inputs.

PR #690, opened 2019-11-08
Add valgrind check to travis

elichai:
As discussed in https://github.com/bitcoin-core/secp256k1/pull/687
This adds valgrind check to the repo.

It doesn't run on recovery+ecdh because of the time.
No openssl because of uninitialized mem.
I debated between with and without ASM, but decided with ASM because it might be more fragile(?).

I wasn't sure if I should pass `-DVALGRIND` via `CFLAGS` or `CPPFLAGS`, it seems like because this is only C then there shouldn't even be `CPPFLAGS` but looks like we use `CPPFLAGS` in other places for the preprocessor definitions.

If people are worried about the time it takes we can mark it as `allow_failure` although I don't think it's a problem here because there's only a handful of PRs and they're usually open for weeks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,10 +61,21 @@ matrix:
           packages:
             - gcc-multilib
             - libgmp-dev:i386
+    - compiler: gcc
+      env:
+        - FIELD=auto  BIGNUM=no  SCALAR=auto  ENDOMORPHISM=yes  STATICPRECOMPUTATION=yes  ASM=x86_64
+        - EXPERIMENTAL=no ECDH=no  RECOVERY=no VALGRIND=yes EXTRAFLAGS="--disable-openssl-tests CPPFLAGS=-DVALGRIND"  HOST= BUILD= JNI=no
+      addons:
+        apt:
+          packages:
+            - valgrind
+    
 before_install: mkdir -p `dirname $GUAVA_JAR`
 install: if [ ! -f $GUAVA_JAR ]; then wget $GUAVA_URL -O $GUAVA_JAR; fi
 before_script: ./autogen.sh
 script:
  - if [ -n "$HOST" ]; then export USE_HOST="--host=$HOST"; fi
  - if [ "x$HOST" = "xi686-linux-gnu" ]; then export CC="$CC -m32"; fi
- - ./configure --enable-experimental=$EXPERIMENTAL --enable-endomorphism=$ENDOMORPHISM --with-field=$FIELD --with-bignum=$BIGNUM --with-asm=$ASM --with-scalar=$SCALAR --enable-ecmult-static-precomputation=$STATICPRECOMPUTATION --with-ecmult-gen-precision=$ECMULTGENPRECISION --enable-module-ecdh=$ECDH --enable-module-recovery=$RECOVERY --enable-jni=$JNI $EXTRAFLAGS $USE_HOST && make -j2 $BUILD
+ - ./configure --enable-experimental=$EXPERIMENTAL --enable-endomorphism=$ENDOMORPHISM --with-field=$FIELD --with-bignum=$BIGNUM --with-asm=$ASM --with-scalar=$SCALAR --enable-ecmult-static-precomputation=$STATICPRECOMPUTATION --with-ecmult-gen-precision=$ECMULTGENPRECISION --enable-module-ecdh=$ECDH --enable-module-recovery=$RECOVERY --enable-jni=$JNI $EXTRAFLAGS $USE_HOST
+ - if [ -n "$BUILD" ]; then make -j2 $BUILD; fi
+ - if [ -n "$VALGRIND" ]; then make -j2 && travis_wait 30 valgrind --error-exitcode=42 ./tests 8; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,8 +63,16 @@ matrix:
             - libgmp-dev:i386
     - compiler: gcc
       env:
-        - FIELD=auto  BIGNUM=no  SCALAR=auto  ENDOMORPHISM=yes  STATICPRECOMPUTATION=yes  ASM=x86_64
-        - EXPERIMENTAL=no ECDH=no  RECOVERY=no VALGRIND=yes EXTRAFLAGS="--disable-openssl-tests CPPFLAGS=-DVALGRIND"  HOST= BUILD= JNI=no
+        - BIGNUM=no  ENDOMORPHISM=yes  ASM=x86_64 EXPERIMENTAL=yes ECDH=yes  RECOVERY=yes
+        - VALGRIND=yes EXTRAFLAGS="--disable-openssl-tests CPPFLAGS=-DVALGRIND" BUILD=
+      addons:
+        apt:
+          packages:
+            - valgrind
+    - compiler: gcc
+      env: # The same as above but without endomorphism.
+        - BIGNUM=no  ENDOMORPHISM=no  ASM=x86_64 EXPERIMENTAL=yes ECDH=yes  RECOVERY=yes
+        - VALGRIND=yes EXTRAFLAGS="--disable-openssl-tests CPPFLAGS=-DVALGRIND" BUILD=
       addons:
         apt:
           packages:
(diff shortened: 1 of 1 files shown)


real-or-random:
Concept ACK

> I wasn't sure if I should pass `-DVALGRIND` via `CFLAGS` or `CPPFLAGS`, it seems like because this is only C then there shouldn't even be `CPPFLAGS` but looks like we use `CPPFLAGS` in other places for the preprocessor definitions.

CPPFLAGS are the flags for the **C** **p**re**p**rocessor. If you need to define flags for a C++ compiler, then you want CXXFLAGS. (Yes, it's confusing...)

elichai:
LOOOL. everything makes sense now. Thanks :) 
(i've been passing C++ flags to both `CPPFLAGS` and `CXXFLAGS` for years :) )

.travis.yml
```diff
@@ -61,10 +61,21 @@ matrix:
           packages:
             - gcc-multilib
             - libgmp-dev:i386
+    - compiler: gcc
+      env:
+        - FIELD=auto  BIGNUM=no  SCALAR=auto  ENDOMORPHISM=yes  STATICPRECOMPUTATION=yes  ASM=x86_64
```
jonasnick: Imo easier to read if you don't redefined variables like field and scalar again.

.travis.yml
```diff
@@ -61,10 +61,21 @@ matrix:
...
+        - EXPERIMENTAL=no ECDH=no  RECOVERY=no VALGRIND=yes EXTRAFLAGS="--disable-openssl-tests CPPFLAGS=-DVALGRIND"  HOST= BUILD= JNI=no
```
jonasnick: I think we want ECDH and RECOVERY enabled here?

.travis.yml
```diff
@@ -61,10 +61,21 @@ matrix:
...
 install: if [ ! -f $GUAVA_JAR ]; then wget $GUAVA_URL -O $GUAVA_JAR; fi
 before_script: ./autogen.sh
 script:
  - if [ -n "$HOST" ]; then export USE_HOST="--host=$HOST"; fi
  - if [ "x$HOST" = "xi686-linux-gnu" ]; then export CC="$CC -m32"; fi
- - ./configure --enable-experimental=$EXPERIMENTAL --enable-endomorphism=$ENDOMORPHISM --with-field=$FIELD --with-bignum=$BIGNUM --with-asm=$ASM --with-scalar=$SCALAR --enable-ecmult-static-precomputation=$STATICPRECOMPUTATION --with-ecmult-gen-precision=$ECMULTGENPRECISION --enable-module-ecdh=$ECDH --enable-module-recovery=$RECOVERY --enable-jni=$JNI $EXTRAFLAGS $USE_HOST && make -j2 $BUILD
+ - ./configure --enable-experimental=$EXPERIMENTAL --enable-endomorphism=$ENDOMORPHISM --with-field=$FIELD --with-bignum=$BIGNUM --with-asm=$ASM --with-scalar=$SCALAR --enable-ecmult-static-precomputation=$STATICPRECOMPUTATION --with-ecmult-gen-precision=$ECMULTGENPRECISION --enable-module-ecdh=$ECDH --enable-module-recovery=$RECOVERY --enable-jni=$JNI $EXTRAFLAGS $USE_HOST
+ - if [ -n "$BUILD" ]; then make -j2 $BUILD; fi
+ - if [ -n "$VALGRIND" ]; then make -j2 && travis_wait 30 valgrind --error-exitcode=42 ./tests 8; fi
```
jonasnick: What does travis_wait do and why error-exitcode=42? Would be good to add a comment. Also, how about also doing the exhaustive tests?

.travis.yml
elichai: On my first tries it took ~25 minutes, but now with `./test 8` it might be faster, i'll try re-enabling them.

.travis.yml
elichai: about exhaustive test i'm not sure about time.
But `travis_wait` is because if 10 minutes pass without anything printed to stdout travis will terminate it, so `travis_wait` silently print an empty char every minute for the written amount(30m) so it won't timeout.
the error code, as you can see in the manual http://valgrind.org/docs/manual/manual-core.html is 0 by default, meaning the tests will pass even though valgrind found errors.

.travis.yml
jonasnick: It worked in my testing without endo and with schnorrsig. Without enabling all features we would still want the memsanitizer PR.

.travis.yml
jonasnick: Exhaustive tests are over pretty quickly. It worked for me with everything enabled (including schnorrsig and non-endo) and without `travis_wait`, which seems pretty useful given that the maximum allowed time  for a job is 50 minutes. I didn't know about the valgrind exit code - makes sense. Would still be good to add a comment explaining that.

.travis.yml
jonasnick: see https://travis-ci.org/jonasnick/secp256k1/jobs/609452969

.travis.yml
```diff
@@ -61,10 +61,36 @@ matrix:
...
+ - ./configure --enable-experimental=$EXPERIMENTAL --enable-endomorphism=$ENDOMORPHISM --with-field=$FIELD --with-bignum=$BIGNUM --with-asm=$ASM --with-scalar=$SCALAR --enable-ecmult-static-precomputation=$STATICPRECOMPUTATION --with-ecmult-gen-precision=$ECMULTGENPRECISION --enable-module-ecdh=$ECDH --enable-module-recovery=$RECOVERY --enable-jni=$JNI $EXTRAFLAGS $USE_HOST
+ - if [ -n "$BUILD" ]; then make -j2 $BUILD; fi
+ - # travis_wait extends the 10 minutes without output allowed (https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received)
+ - # the `--error-exitcode` is required to make the test fail if valgrind found errors, otherwise it'll return 0 (http://valgrind.org/docs/manual/manual-core.html)
+ - if [ -n "$VALGRIND" ]; then
+   make -j2 &&
+   travis_wait 30 valgrind --error-exitcode=42 ./tests 8 &&
+   travis_wait 30 valgrind --error-exitcode=42 ./exhaustive_tests 8;
```
jonasnick: nit: `exhaustive_tests` don't take an argument

.travis.yml
```diff
@@ -61,10 +61,36 @@ matrix:
...
+      env:
+        - BIGNUM=no  ENDOMORPHISM=yes  STATICPRECOMPUTATION=yes  ASM=x86_64
+        - EXPERIMENTAL=yes ECDH=yes  RECOVERY=yes VALGRIND=yes EXTRAFLAGS="--disable-openssl-tests CPPFLAGS=-DVALGRIND" BUILD=
+      addons:
+        apt:
+          packages:
+            - valgrind
+    - compiler: gcc
+      env: # The same as above but without endomorphism.
+        - BIGNUM=no  ENDOMORPHISM=no  STATICPRECOMPUTATION=yes  ASM=x86_64
```
jonasnick: nit: `STATICPRECOMPUTATION` is `yes` by default

.travis.yml
```diff
@@ -61,10 +61,36 @@ matrix:
...
+ - ./configure --enable-experimental=$EXPERIMENTAL --enable-endomorphism=$ENDOMORPHISM --with-field=$FIELD --with-bignum=$BIGNUM --with-asm=$ASM --with-scalar=$SCALAR --enable-ecmult-static-precomputation=$STATICPRECOMPUTATION --with-ecmult-gen-precision=$ECMULTGENPRECISION --enable-module-ecdh=$ECDH --enable-module-recovery=$RECOVERY --enable-jni=$JNI $EXTRAFLAGS $USE_HOST
+ - if [ -n "$BUILD" ]; then make -j2 $BUILD; fi
+ - # travis_wait extends the 10 minutes without output allowed (https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received)
+ - # the `--error-exitcode` is required to make the test fail if valgrind found errors, otherwise it'll return 0 (http://valgrind.org/docs/manual/manual-core.html)
+ - if [ -n "$VALGRIND" ]; then
+   make -j2 &&
+   travis_wait 30 valgrind --error-exitcode=42 ./tests 8 &&
```
jonasnick: We could optimize the `test runs` parameter because it's basically free. For example it looks like 16 would work just as well.

.travis.yml
jonasnick: nit: it's still present in the endo valgrind run

elichai: Totally forgot I split them in 2 :)

.travis.yml
```diff
@@ -61,10 +61,36 @@ matrix:
           packages:
             - gcc-multilib
             - libgmp-dev:i386
+    - compiler: gcc
+      env:
+        - BIGNUM=no  ENDOMORPHISM=yes  ASM=x86_64 EXPERIMENTAL=yes ECDH=yes  RECOVERY=yes
```
jonasnick: for the record: we *don't have to* disable gmp to get a valgrind pass

elichai: Yes, But I don't see a lot of value in running valgrind on `gmp` and I'd rather it tested our implementations.
altough there's quite a lot gmp wrapper code too.

gmaxwell: You might want a second run with endo off.  Otherwise you're not testing the most common config with it, and there have been issues in the past that were valgrind clean with endo on but not off and vice versa.  If speed is a concern, just decrease test count... running tests 4 twice is more testing than tests 8.  If for some reason only one can be run, endo=off is probably better.

elichai: There is :) https://github.com/bitcoin-core/secp256k1/pull/690/files#diff-354f30a63fb0907d4ad57269548329e3R74

real-or-random:
ACK dd98cc988f0fb3a0ab10bf1a4e28d2fbffd6c1e7 I looked at the diff

PR #696, opened 2019-11-28
Run a Travis test on s390x (big endian)

sipa:


sipa:
> checking whether byte ordering is bigendian... yes

\o/

elichai:
It worked on core :P
But still satisfying to see.
There's nothing more satisfying than seeing that your code works on an edge case you never tested before l

jonasnick:
Cool! What's missing for un-WIP'ing? How about turning on the ECDH and recovery module?

real-or-random:
> Cool! What's missing for un-WIP'ing? How about turning on the ECDH and recovery module?

@sipa Could you do this, and maybe also add a comment to the Travis file that this is a big-endian host?

.travis.yml
```diff
@@ -35,6 +35,9 @@ env:
 matrix:
   fast_finish: true
   include:
+    - compiler: gcc
+      env: HOST=s390x-unknown-linux-gnu DOCKER_NAME_TAG=s390x/ubuntu:18.04
```
MarcoFalke, line 39: What does `DOCKER_NAME_TAG` do?
```suggestion
      env: HOST=s390x-unknown-linux-gnu
```

sipa: I have no idea anymore, I assume I copied it from somewhere.

sipa:
Rebased, and added ECDH/recovery.

I did notice valgrind ctime test failure on this line in src/scalar_4x64_impl.h:

```c
    no |= (a->d[1] < SECP256K1_N_1);
```

so I've disabled the ctime test for now.

.travis.yml
```diff
@@ -83,6 +83,10 @@ matrix:
             - valgrind
             - libtool-bin
             - libc6-dbg:i386
+    # S390x build (big endian system)
+    - compiler: gcc
+      env: HOST=s390x-unknown-linux-gnu ECDH=yes RECOVERY=yes EXPERIMENTAL=yes CTIMETEST=no
```
real-or-random, line 88: travis still doesn't like you
```suggestion
      env: HOST=s390x-unknown-linux-gnu ECDH=yes RECOVERY=yes EXPERIMENTAL=yes CTIMETEST=
```
I admit that this is not intuitive...

real-or-random:
> ```c
>     no |= (a->d[1] < SECP256K1_N_1);
> ```

and two lines below too:

https://github.com/bitcoin-core/secp256k1/blob/master/src/scalar_4x64_impl.h#L65

> 
> so I've disabled the ctime test for now.

makes sense.

real-or-random:
ACK c5e43bb8d6c60c5f8ee38c89ac4d0d755e6515d0 Travis works and says it's big endian

The travis failure is a randomness test again (#68), see https://travis-ci.org/github/bitcoin-core/secp256k1/jobs/713125311#L497 

If you want, you can squash this. Otherwise I'll just merge it in the next days if noone objects.

Issue #598, opened 2019-03-09
Travis should build and test arm asm


gmaxwell:

gmaxwell:
Travis should also test make install.

benma:
Related: Travis does not seem to run `./tests`. Is there a reason for that?

real-or-random:
It does, see https://travis-ci.org/bitcoin-core/secp256k1/jobs/560256756#L708 for example.
It just runs the entire testsuite using `make check-TESTS`. Maybe we should just update the README.

benma:
Of course!

Updated README here: https://github.com/bitcoin-core/secp256k1/pull/652

real-or-random:
Travis should also test different window size settings.

real-or-random:
Travis has now (alpha) support for ARM: 
https://blog.travis-ci.com/2019-10-07-multi-cpu-architecture-support

elichai:
FWIW it's arm64, and we have asm only for arm, not arm64.

laanwj:
It can build and run ARM executables on ARM64 (FWIW this is what we do in bitcoin )

elichai:
Didn't know arm64 can execute arm binaries. i'll give this a try

laanwj:
I think it's a feature that many SoCs have, including the one Travis happens to use, not a general requirement for arm64

gmaxwell:
Even just having arm64 would be a big improvement. Sure, that wouldn't test that the arm ASM builds and works-- which is the critical thing motivating this issue, but there isn't anything blocking that as far as I know.

Similarly, there appears to be no BE host in this project's travis configuration. s390x is be, I believe. (lol)

elichai:
CC https://github.com/bitcoin-core/secp256k1/pull/696

real-or-random:
This has been solved by #930.

Issue #621, opened 2019-05-17
Is there an option to create a shared library for Android applications?


gecng:
I use `./configure --enable-jni --enable-experimental --enable-module-schnorr --enable-module-ecdh` and get one  secp256k1.so files.but there are many different CPU architectures in android like armeabi ,rmeabi-v7a,arm64-v8a ,x86 . Could you please tell me how should I do to create .so files for different CPU architectures. Thanks

gecng:
when  I use  `System.loadLibrary("javasecp256k1")` ， it throws exception 
java.lang.UnsatisfiedLinkError: dlopen failed: "/data/app/com.wallet.locker-Gpfu8hdv5zMz363H3eJC8Q==/lib/arm/libjavasecp256k1.so" is 64-bit instead of 32-bit.

sipa:
I don't much about Android, but generally I would expect you need a separate build/.so per architecture, and then ship the right one for the used architecture.

gecng:
> I don't much about Android, but generally I would expect you need a separate build/.so per architecture, and then ship the right one for the used architecture.


> I don't much about Android, but generally I would expect you need a separate build/.so per architecture, and then ship the right one for the used architecture.

yes,I really need a separate build/.so per architecture, but I don't know the way to build it

real-or-random:
`configure` has a `--host` option for cross compilation.

real-or-random:
For reference, I can build this for example by installing Android NDK 19 and then using
```
export PATH="$PATH:/opt/android-ndk/toolchains/llvm/prebuilt/linux-x86_64/bin"
./configure --enable-ecmult-static-precomputation --enable-experimental --with-bignum=no --with-asm=arm --host=arm-linux-androideabi CC=armv7a-linux-androideabi21-clang CFLAGS="-mthumb -march=armv7-a" CCASFLAGS="-Wa,-mthumb -Wa,-march=armv7-a"
```
See https://developer.android.com/ndk/guides/standalone_toolchain for more information, e.g., the right flags for other targets.

gecng:
> For reference, I can build this for example by installing Android NDK 19 and then using
> 
> ```
> export PATH="$PATH:/opt/android-ndk/toolchains/llvm/prebuilt/linux-x86_64/bin"
[...]

PR #698, opened 2019-12-05
travis: Add some arm builds

MarcoFalke:
Proof of concept for #598

Not sure which environment combinations make most sense (something with `ASM`?)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: c
 os: linux
+arch: amd64
 addons:
   apt:
     packages: libgmp-dev
@@ -61,6 +62,20 @@ matrix:
           packages:
             - gcc-multilib
             - libgmp-dev:i386
+    - compiler: clang
+      env: HOST=arm-linux-gnueabihf ASM=arm CC=arm-linux-gnueabihf-gcc EXPERIMENTAL=yes
+      arch: arm64
+      addons:
+        apt:
+          packages:
+            - gcc-arm-linux-gnueabihf
+    - compiler: gcc
+      env: HOST=arm-linux-gnueabihf ASM=arm CC=arm-linux-gnueabihf-gcc EXPERIMENTAL=yes ENDOMORPHISM=yes
+      arch: arm64
+      addons:
+        apt:
+          packages:
+            - gcc-arm-linux-gnueabihf
     - compiler: gcc
       env:
         - BIGNUM=no  ENDOMORPHISM=yes  ASM=x86_64 EXPERIMENTAL=yes ECDH=yes  RECOVERY=yes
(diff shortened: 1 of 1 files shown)


sipa:
We have optimized assembly for 32-bit ARM (not enabled by default, and not 64-bit iirc). It'd be neat to test that.

MarcoFalke:
ok, I set ASM=arm, but that fails:


```
$ if [ -n "$BUILD" ]; then make -j2 $BUILD; fi

  CC       src/bench_verify.o

  CC       src/bench_sign.o

gcc -I. -I./src -g -O2 -Wall -Wextra -Wno-unused-function -c src/gen_context.c -o gen_context.o

  CCAS     src/asm/field_10x26_arm.lo

src/asm/field_10x26_arm.s: Assembler messages:

src/asm/field_10x26_arm.s:1: Error: junk at end of line, first unrecognized character is `@'

src/asm/field_10x26_arm.s:18: Error: unknown pseudo-op: `.syntax'

src/asm/field_10x26_arm.s:19: Error: junk at end of line, first unrecognized character is `@'

src/asm/field_10x26_arm.s:20: Error: unknown pseudo-op: `.eabi_attribute'

src/asm/field_10x26_arm.s:21: Error: unknown pseudo-op: `.eabi_attribute'

src/asm/field_10x26_arm.s:24: Error: junk at end of line, first unrecognized character is `@'

src/asm/field_10x26_arm.s:27: Error: junk at end of line, first unrecognized character is `@'
...


```

See https://travis-ci.org/MarcoFalke/secp256k1/builds/621315153

elichai:
Tried that in the past and couldn't make the asm work.
Hope you will :)

real-or-random:
I haven't had a closer look now but this may help to figure out the right flags (even though it's for cross-compilation)
https://github.com/bitcoin-core/secp256k1/issues/621#issuecomment-495703399

laanwj:
I think you get those errors when rying to compile the 32-bit ARM assembly with a 64-bit targeted assembler.

MarcoFalke:
I might need to install libc from armhf. That doesn't work, so I reported the issue to travis: https://travis-ci.community/t/apt-addon-does-not-work-for-armhf-on-arm64/6374

real-or-random:
Did this work in the end? The travis output looks good, it was just aborted?

MarcoFalke:
It failed with 
```
/usr/lib/gcc-cross/arm-linux-gnueabihf/5/../../../../arm-linux-gnueabihf/bin/ld: cannot find crti.o: No such file or directory
```

https://travis-ci.org/MarcoFalke/secp256k1/jobs/624276878#L396

real-or-random:
Ah yes, thanks for clarifying!

PR #701, opened 2019-12-17
Make ec_ arithmetic more consistent and add documentation

### LINKED ISSUE #671: Make ec_ arithmetic more consistent and add documentation
As brought up by @afk11 on #secp256k1 there are a couple of undocumented and inconsistent surprises in the ec arithmetic functions (that is `secp256k1_ec_pub/privkey_neg`, `_tweak_add` and `tweak_mul`.

In particular:

1. Right now, `ec_privkey_neg`, `ec_privkey_tweak_add` and `ec_privkey_tweak_mul` do not fail if the seckey is invalid, but they may give unexpected results in that case (f.e. #488). In order to avoid that, the user should call `ec_seckey_verify` before calling these functions, but that's not documented. Alternatively, these functions could call `ec_seckey_verify` themselves. In any case we should define what a valid `seckey` is (in the documentation of `ec_seckey_verify` for example) and refer to that in the argument documentation of other functions.
2. `ec_privkey_tweak_add` and `ec_privkey_tweak_mul` can fail (return 0). As a result the seckey argument will be zeroized. This is not documented. There's conflicting ideas whether this is a good idea:
    * "preserving any entropy in the bad input is the safer behaviour (e.g. the data in it might get fed elsewhere into some rng hash or something)" (@gmaxwell in #668).
    * vs.
        ```
        17:25 < andytoshi> leaving it alone sounds very dangerous, presumably the original key will "work" in subsequent algos
        17:25 < andytoshi> while a 0 key is pretty visible and will cause failures pretty-much no matter what you do with it
        ```
        However, right now we wouldn't detect the 0 key in `ec_privkey_tweak_add` and `tweak_mul`

3. `privkey_tweak_add` and `privkey_tweak_mul` fail if the resulting privkey is 0 but not if it overflows. That's not ideal because you won't be able to sign for it. Additionally, the documentation for `privkey_tweak_add` is wrong: `returns: [...] 0 the resulting private key [...] would be invalid (only when the tweak is the complement of the corresponding private key)` (same for `pubkey_tweak_add`). An overflowing private key is also invalid.
4. All this behavior should be tested
5. See #670 for naming improvements (seckey vs. privkey)

jonasnick:
Fixes #671. Supersedes #668.

This PR unifies handling of invalid secret keys by introducing a new function `scalar_set_b32_secret` which returns false if the b32 overflows or is 0. By using this in `privkey_{negate, tweak_add, tweak_mul}` these function will now return 0 if the secret key is invalid which matches the behavior of `ecdsa_sign` and `pubkey_create`.

Instead of deciding whether to zeroize the secret key on failure, I only added documentation for now that the value is undefined on failure.

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -579,9 +579,12 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_create(
 
 /** Negates a private key in place.
  *
- *  Returns: 1 always
+ *  Returns: 0 if the given private key is invalid according to
+ *           secp256k1_ec_seckey_verify. 1 otherwise
  *  Args:   ctx:        pointer to a context object
- *  In/Out: seckey:     pointer to the 32-byte private key to be negated (cannot be NULL)
+ *  In/Out: seckey:     pointer to the 32-byte private key to be negated. The private
+ *                      key should be valid according to secp256k1_ec_seckey_verify
+ *                      (cannot be NULL)
  */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_privkey_negate(
     const secp256k1_context* ctx,
@@ -601,9 +604,10 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_negate(
 
 /** Tweak a private key by adding tweak to it.
  * Returns: 0 if the tweak was out of range (chance of around 1 in 2^128 for
- *          uniformly random 32-byte arrays, or if the resulting private key
- *          would be invalid (only when the tweak is the complement of the
- *          private key). 1 otherwise.
+ *          uniformly random 32-byte arrays, or if the given private key is
+ *          invalid according to secp256k1_ec_seckey_verify, or if the resulting
+ *          private key would be invalid (only when the tweak is the complement
+ *          of the private key). 1 otherwise.
  * Args:    ctx:    pointer to a context object (cannot be NULL).
  * In/Out:  seckey: pointer to a 32-byte private key.
  * In:      tweak:  pointer to a 32-byte tweak.
@@ -616,9 +620,10 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_privkey_tweak_add(
 
 /** Tweak a public key by adding tweak times the generator to it.
  * Returns: 0 if the tweak was out of range (chance of around 1 in 2^128 for
- *          uniformly random 32-byte arrays, or if the resulting public key
- *          would be invalid (only when the tweak is the complement of the
- *          corresponding private key). 1 otherwise.
+ *          uniformly random 32-byte arrays, or if the given private key is
+ *          invalid according to secp256k1_ec_seckey_verify, or if the resulting
+ *          public key would be invalid (only when the tweak is the complement
+ *          of the corresponding private key). 1 otherwise.
  * Args:    ctx:    pointer to a context object initialized for validation
  *                  (cannot be NULL).
  * In/Out:  pubkey: pointer to a public key object.
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -592,6 +592,13 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_create(
  *                      If this function returns 0, seckey will be some
  *                      unspecified value. (cannot be NULL)
  */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_seckey_negate(
+    const secp256k1_context* ctx,
+    unsigned char *seckey
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2);
+
+/** Same as secp256k1_ec_seckey_negate, but DEPRECATED. Will be removed in
+ *  future versions. */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_privkey_negate(
     const secp256k1_context* ctx,
     unsigned char *seckey
@@ -621,6 +628,14 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_negate(
  *                  32-byte arrays the chance of being out of range is
  *                  negligible (around 1 in 2^128). (cannot be NULL)
  */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_seckey_tweak_add(
+    const secp256k1_context* ctx,
+    unsigned char *seckey,
+    const unsigned char *tweak
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
+
+/** Same as secp256k1_ec_seckey_tweak_add, but DEPRECATED. Will be removed in
+ *  future versions. */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_privkey_tweak_add(
     const secp256k1_context* ctx,
     unsigned char *seckey,
@@ -655,6 +670,14 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_tweak_add(
  *                 32-byte arrays the chance of being out of range is
  *                 negligible (around 1 in 2^128). (cannot be NULL)
  */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_seckey_tweak_mul(
+    const secp256k1_context* ctx,
+    unsigned char *seckey,
+    const unsigned char *tweak
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
+
+/** Same as secp256k1_ec_seckey_tweak_mul, but DEPRECATED. Will be removed in
+ *  future versions. */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_privkey_tweak_mul(
     const secp256k1_context* ctx,
     unsigned char *seckey,
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -14,7 +14,7 @@ extern "C" {
  * 2. Array lengths always immediately the follow the argument whose length
  *    they describe, even if this violates rule 1.
  * 3. Within the OUT/OUTIN/IN groups, pointers to data that is typically generated
- *    later go first. This means: signatures, public nonces, private nonces,
+ *    later go first. This means: signatures, public nonces, secret nonces,
  *    messages, public keys, secret keys, tweaks.
  * 4. Arguments that are not data pointers go last, from more complex to less
  *    complex: function pointers, algorithm names, messages, void pointers,
@@ -531,7 +531,7 @@ SECP256K1_API extern const secp256k1_nonce_function secp256k1_nonce_function_def
 /** Create an ECDSA signature.
  *
  *  Returns: 1: signature created
- *           0: the nonce generation function failed, or the private key was invalid.
+ *           0: the nonce generation function failed, or the secret key was invalid.
  *  Args:    ctx:    pointer to a context object, initialized for signing (cannot be NULL)
  *  Out:     sig:    pointer to an array where the signature will be placed (cannot be NULL)
  *  In:      msg32:  the 32-byte message hash being signed (cannot be NULL)
@@ -574,7 +574,7 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_seckey_verify(
  *           0: secret was invalid, try again
  *  Args:   ctx:        pointer to a context object, initialized for signing (cannot be NULL)
  *  Out:    pubkey:     pointer to the created public key (cannot be NULL)
- *  In:     seckey:     pointer to a 32-byte private key (cannot be NULL)
+ *  In:     seckey:     pointer to a 32-byte secret key (cannot be NULL)
  */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_create(
     const secp256k1_context* ctx,
@@ -582,15 +582,15 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_create(
     const unsigned char *seckey
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
 
-/** Negates a private key in place.
+/** Negates a secret key in place.
  *
- *  Returns: 0 if the given private key is invalid according to
+ *  Returns: 0 if the given secret key is invalid according to
  *           secp256k1_ec_seckey_verify. 1 otherwise
  *  Args:   ctx:        pointer to a context object
- *  In/Out: seckey:     pointer to the 32-byte private key to be negated. The private
+ *  In/Out: seckey:     pointer to the 32-byte secret key to be negated. The secret
  *                      key should be valid according to secp256k1_ec_seckey_verify.
- *                      Value becomes unspecified if this function returns 0.
- *                      (cannot be NULL)
+ *                      If this function returns 0, seckey will be some
+ *                      unspecified value. (cannot be NULL)
  */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_privkey_negate(
     const secp256k1_context* ctx,
@@ -608,16 +608,18 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_negate(
     secp256k1_pubkey *pubkey
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2);
 
-/** Tweak a private key by adding tweak to it.
- * Returns: 0 if the tweak was out of range (chance of around 1 in 2^128 for
- *          uniformly random 32-byte arrays, or if the given private key is
- *          invalid according to secp256k1_ec_seckey_verify, or if the resulting
- *          private key would be invalid (only when the tweak is the complement
- *          of the private key). 1 otherwise.
+/** Tweak a secret key by adding tweak to it.
+ * Returns: 0 if the resulting secret key would be invalid (only when the tweak
+ *          is the negation of the secret key). 1 otherwise.
  * Args:    ctx:    pointer to a context object (cannot be NULL).
- * In/Out:  seckey: pointer to a 32-byte private key. Value becomes unspecified if this
- *                  function returns 0. (cannot be NULL)
- * In:      tweak:  pointer to a 32-byte tweak.
+ * In/Out:  seckey: pointer to a 32-byte secret key. The secret key should be
+ *                  valid according to secp256k1_ec_seckey_verify. If this
+ *                  function returns 0, seckey will be some unspecified
+ *                  value. (cannot be NULL)
+ * In:       tweak: pointer to a 32-byte tweak. Must be in the same range as secret
+ *                  keys (see secp256k1_ec_seckey_verify). For uniformly random
+ *                  32-byte arrays the chance of being out of range is
+ *                  negligible (around 1 in 2^128). (cannot be NULL)
  */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_privkey_tweak_add(
     const secp256k1_context* ctx,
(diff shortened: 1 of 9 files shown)


include/secp256k1.h
```diff
@@ -599,9 +602,10 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_negate(
 
 /** Tweak a private key by adding tweak to it.
  * Returns: 0 if the tweak was out of range (chance of around 1 in 2^128 for
- *          uniformly random 32-byte arrays, or if the resulting private key
- *          would be invalid (only when the tweak is the complement of the
- *          private key). 1 otherwise.
+ *          uniformly random 32-byte arrays, or if the given private key is
+ *          invalid according to secp256k1_ec_seckey_verify, or if the resulting
+ *          private key would be invalid (only when the tweak is the complement
```
real-or-random: Maybe "negated private key" instead of "complement"?

Is this really the only case? What about overflows?

include/secp256k1.h
```diff
@@ -614,9 +618,10 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_privkey_tweak_add(
 
 /** Tweak a public key by adding tweak times the generator to it.
  * Returns: 0 if the tweak was out of range (chance of around 1 in 2^128 for
- *          uniformly random 32-byte arrays, or if the resulting public key
- *          would be invalid (only when the tweak is the complement of the
- *          corresponding private key). 1 otherwise.
+ *          uniformly random 32-byte arrays, or if the given private key is
+ *          invalid according to secp256k1_ec_seckey_verify, or if the resulting
+ *          public key would be invalid (only when the tweak is the complement
```
real-or-random: same here

src/secp256k1.c
```diff
@@ -526,7 +526,9 @@ int secp256k1_ec_privkey_negate(const secp256k1_context* ctx, unsigned char *sec
     VERIFY_CHECK(ctx != NULL);
     ARG_CHECK(seckey != NULL);
 
-    secp256k1_scalar_set_b32(&sec, seckey, NULL);
+    if(!secp256k1_scalar_set_b32_seckey(&sec, seckey)) {
```
real-or-random: nit: spacing

include/secp256k1.h
real-or-random: This should also document what the valid range for a tweak is? (I guess every scalar?)

include/secp256k1.h
```diff
@@ -584,7 +584,8 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_create(
  *           secp256k1_ec_seckey_verify. 1 otherwise
  *  Args:   ctx:        pointer to a context object
  *  In/Out: seckey:     pointer to the 32-byte private key to be negated. The private
- *                      key should be valid according to secp256k1_ec_seckey_verify
+ *                      key should be valid according to secp256k1_ec_seckey_verify.
+ *                      Value becomes undefined if this function returns 0.
```
real-or-random: Can we say "unspecified"? "Undefined" comes with that weird terrifying sound of undefined behavior.

src/secp256k1.c
jonasnick: fixed

jonasnick:
@real-or-random  I added a commit that rephrases the docs. Let me know if you think that's better.

(Also, I replaced `secp256k1_rand256_test` with `secp256k1_rand256` because the former is apparently not random, may produce invalid seckeys and travis randomly hit that in one of the test configurations).

src/tests.c
```diff
@@ -1117,16 +1117,43 @@ void run_scalar_tests(void) {
...
+        secp256k1_scalar_set_b32(&scalar, bin, &overflow);
+        CHECK(overflow == 1);
+        CHECK(secp256k1_scalar_is_zero(&scalar));
+
+        /* A scalar with value of the curve order minus one should not overflow. */
+        bin[31] -= 1;
+        secp256k1_scalar_set_b32(&scalar, bin, &overflow);
+        CHECK(overflow == 0);
+        secp256k1_scalar_get_b32(bin_tmp, &scalar);
+        CHECK(memcmp(bin, bin_tmp, 32) == 0);
```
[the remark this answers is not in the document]
jonasnick: It's probably the only roundtrip because it's the only one which doesn't overlow and which can not be easily tested with `scalar_is_zero`.

jonasnick:
Removed the (attempted) constant time AND in `set_b32_seckey` as per @sipa's suggestion

real-or-random:
Is there a reason to believe that this will break downstream code (because it changes the API)?

include/secp256k1.h
```diff
@@ -597,57 +611,85 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_negate(
...
  * Args:   ctx:    pointer to a context object (cannot be NULL).
- * In/Out: seckey: pointer to a 32-byte private key.
- * In:     tweak:  pointer to a 32-byte tweak.
- */
+ * In/Out: seckey: pointer to a 32-byte secret key. If this function returns 0,
+ *                 seckey will be some unspecified value. (cannot be NULL).
+ * In:     tweak:  pointer to a 32-byte tweak. Must be in the same range as secret
+ *                 keys (see secp256k1_ec_seckey_verify). For uniformly random
+ *                 32-byte arrays the chance of being out of range is
+ *                 negligible (around 1 in 2^128). (cannot be NULL)
```
dr-orlovsky: while the changes are negligible, if one is aimed to write a deterministic code (like in the case of rust wrapper), it may be good to cover this case as well. 

It is possible to add to the docs that in case `tweak`<Z_p, the function will return 0?
Also, do you see any value of differentiating between `tweak`<Z_p and point at infinity cases? Right now both will return 0, but from the end-user perspective the first case requires change of the tweaking factor, while the second may be solved just by using a different public key...

real-or-random: > while the changes are negligible, if one is aimed to write a deterministic code (like in the case of rust wrapper), it may be good to cover this case as well.

Believe me, you don't care about a failure probability of 2^-128. It's astronomically small. It's much smaller than the probability of a hardware error (e.g., RAM returns wrong value). You can simply abort in this case.

include/secp256k1.h
apoelstra: Also, missing an `)` after "32-byte arrays". (This was a preexisting problem but we should fix it now.)

include/secp256k1.h
apoelstra: same `)` issue here

src/tests.c
```diff
@@ -3994,6 +3994,16 @@ void run_eckey_edge_case_test(void) {
     CHECK(secp256k1_ec_pubkey_tweak_mul(ctx, &pubkey, ctmp2) == 0);
     CHECK(memcmp(&pubkey, zeros, sizeof(pubkey)) == 0);
     memcpy(&pubkey, &pubkey2, sizeof(pubkey));
+    /* Overflowing key zeroizes */
+    memcpy(ctmp, orderc, 32);
+    memset(ctmp2, 0, 32);
+    ctmp2[31] = 0x01;
+    CHECK(secp256k1_ec_seckey_verify(ctx, ctmp2) == 1);
+    CHECK(secp256k1_ec_privkey_tweak_add(ctx, ctmp, ctmp2) == 0);
```
apoelstra: Can you add a check with `ctmp` and `ctmp2` swapped? Or maybe there is little value in this since both of these are invalid.

gmaxwell:
@jonasnick  the _test rng's are biased intentionally to hit corner cases and it worked perfectly there if it detected that your scalar generation was generating an invalid scalar.  Maybe you want random_scalar_order_test  or random_scalar_order instead?   It's better to make the tests not fail even with rare values because they can also be moved over to smaller groups where the failures would be common.

include/secp256k1.h
jonasnick: @apoelstra should be fixed in the latest commit.

src/tests.c
jonasnick: Are you suggesting to test this function with an overflowing tweak? That should happen a couple of lines below. I pushed a commit to make the comments more clear.

jonasnick:
@gmaxwell Thanks, `random_scalar_order` is almost what I was looking for and agree that it's better than `rand256`. I added a function similar to `random_scalar_order` but returning a byte array.

real-or-random:
Can you squash this?

jonasnick:
Rebasing required a bit of thought due to #710. Please have a look.

jonasnick:
I improved `ecdsa_sign`'s uses of `secp256k1_scalar_set_b32_seckey` a little bit.

src/tests.c
```diff
@@ -3936,45 +3989,64 @@ void run_eckey_edge_case_test(void) {
...
     /* Multiply tweak of zero zeroizes the output. */
-    CHECK(secp256k1_ec_privkey_tweak_mul(ctx, ctmp, ctmp2) == 0);
+    CHECK(secp256k1_ec_seckey_tweak_mul(ctx, ctmp, ctmp2) == 0);
     CHECK(memcmp(zeros, ctmp, 32) == 0);
     CHECK(secp256k1_ec_pubkey_tweak_mul(ctx, &pubkey, ctmp2) == 0);
     CHECK(memcmp(&pubkey, zeros, sizeof(pubkey)) == 0);
     memcpy(&pubkey, &pubkey2, sizeof(pubkey));
-    /* Overflowing key tweak zeroizes. */
+    /* If seckey_tweak_add or seckey_tweak_mul are called with an overflowing
+    seckey, the seckey is zeroized. */
```
real-or-random: Here they're not called with an overflowing seckey but the resulting seckey would overflow.

src/scalar_impl.h
```diff
@@ -55,6 +55,12 @@ static void secp256k1_scalar_order_get_num(secp256k1_num *r) {
 }
 #endif
 
+static int secp256k1_scalar_set_b32_seckey(secp256k1_scalar *r, const unsigned char *bin) {
+  int overflow;
+  secp256k1_scalar_set_b32(r, bin, &overflow);
+  return (!overflow) & (!secp256k1_scalar_is_zero(r));
```
real-or-random: nit: indentation
I only point it out here but it's wrong in many functions you added.

include/secp256k1.h
```diff
@@ -569,20 +572,31 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_seckey_verify(
...
+ *                      If this function returns 0, seckey will be some
+ *                      unspecified value. (cannot be NULL)
  */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_seckey_negate(
+    const secp256k1_context* ctx,
+    unsigned char *seckey
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2);
+
+/** Same as secp256k1_ec_seckey_negate, but DEPRECATED. Will be removed in
+ *  future versions. */
```
real-or-random: Can you change the calls in the codebase to  the `seckey` variants? For example https://github.com/bitcoin-core/secp256k1/blob/master/src/valgrind_ctime_test.c#L76

include/secp256k1.h
```diff
@@ -599,57 +613,85 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_negate(
     secp256k1_pubkey *pubkey
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2);
 
-/** Tweak a private key by adding tweak to it.
- * Returns: 0 if the tweak was out of range (chance of around 1 in 2^128 for
- *          uniformly random 32-byte arrays, or if the resulting private key
- *          would be invalid (only when the tweak is the complement of the
- *          private key). 1 otherwise.
+/** Tweak a secret key by adding tweak to it.
```
real-or-random: nit: Now that you're rewriting those, can you add an empty line here? (same for the other tweak functions)

src/eckey_impl.h
```diff
@@ -54,10 +54,7 @@ static int secp256k1_eckey_pubkey_serialize(secp256k1_ge *elem, unsigned char *p
 
 static int secp256k1_eckey_privkey_tweak_add(secp256k1_scalar *key, const secp256k1_scalar *tweak) {
     secp256k1_scalar_add(key, key, tweak);
-    if (secp256k1_scalar_is_zero(key)) {
-        return 0;
-    }
-    return 1;
```
real-or-random: Huh, shouldn't this have been caught by the new constant time test?

include/secp256k1.h
```diff
@@ -599,57 +613,85 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_negate(
...
     const unsigned char *tweak
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
 
 /** Tweak a public key by adding tweak times the generator to it.
- * Returns: 0 if the tweak was out of range (chance of around 1 in 2^128 for
- *          uniformly random 32-byte arrays, or if the resulting public key
- *          would be invalid (only when the tweak is the complement of the
- *          corresponding private key). 1 otherwise.
+ * Returns: 0 if the resulting public key would be invalid (only when the tweak
+ *          is the negation of the corresponding secret key). 1 otherwise.
```
real-or-random: What if the tweak itself is out of range?

include/secp256k1.h
```diff
@@ -569,20 +572,31 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_seckey_verify(
...
-/** Negates a private key in place.
+/** Negates a secret key in place.
  *
- *  Returns: 1 always
+ *  Returns: 0 if the given secret key is invalid according to
+ *           secp256k1_ec_seckey_verify. 1 otherwise
  *  Args:   ctx:        pointer to a context object
- *  In/Out: seckey:     pointer to the 32-byte private key to be negated (cannot be NULL)
+ *  In/Out: seckey:     pointer to the 32-byte secret key to be negated. The secret
+ *                      key should be valid according to secp256k1_ec_seckey_verify.
```
real-or-random: nit: "should" is somewhat imprecise. suggestion: "If the secret key is invalid according to secp256k1_ec_seckey_verify, this function returns 0 and seckey will be set to some  unspecified value."

include/secp256k1.h
```diff
@@ -599,57 +613,85 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_negate(
...
- *          would be invalid (only when the tweak is the complement of the
- *          private key). 1 otherwise.
+/** Tweak a secret key by adding tweak to it.
+ * Returns: 0 if the resulting secret key would be invalid (only when the tweak
+ *          is the negation of the secret key). 1 otherwise.
  * Args:    ctx:    pointer to a context object (cannot be NULL).
- * In/Out:  seckey: pointer to a 32-byte private key.
- * In:      tweak:  pointer to a 32-byte tweak.
- */
+ * In/Out:  seckey: pointer to a 32-byte secret key. The secret key should be
```
real-or-random: same as above

include/secp256k1.h
```diff
@@ -599,57 +613,85 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_negate(
     secp256k1_pubkey *pubkey
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2);
 
-/** Tweak a private key by adding tweak to it.
- * Returns: 0 if the tweak was out of range (chance of around 1 in 2^128 for
- *          uniformly random 32-byte arrays, or if the resulting private key
- *          would be invalid (only when the tweak is the complement of the
- *          private key). 1 otherwise.
+/** Tweak a secret key by adding tweak to it.
+ * Returns: 0 if the resulting secret key would be invalid (only when the tweak
```
real-or-random: What if the tweak itself is out of range?

include/secp256k1.h
```diff
@@ -599,57 +613,85 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_negate(
...
+ *          is the negation of the secret key). 1 otherwise.
  * Args:    ctx:    pointer to a context object (cannot be NULL).
- * In/Out:  seckey: pointer to a 32-byte private key.
- * In:      tweak:  pointer to a 32-byte tweak.
- */
+ * In/Out:  seckey: pointer to a 32-byte secret key. The secret key should be
+ *                  valid according to secp256k1_ec_seckey_verify. If this
+ *                  function returns 0, seckey will be some unspecified
+ *                  value. (cannot be NULL)
+ * In:       tweak: pointer to a 32-byte tweak. Must be in the same range as secret
```
real-or-random: "must" is similar to "should" here. I guess it's better to specify the consequences if it's not in range, then the function will return 0.

include/secp256k1.h
```diff
@@ -551,7 +551,10 @@ SECP256K1_API int secp256k1_ecdsa_sign(
     const void *ndata
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
 
-/** Verify an ECDSA secret key.
+/** Verify an ECDSA secret key. It is valid if it is not 0 and less than the
```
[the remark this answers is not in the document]

include/secp256k1.h
```diff
@@ -551,7 +551,10 @@ SECP256K1_API int secp256k1_ecdsa_sign(
...
+ *  secp256k1 curve order when interpreted as an integer (most significant byte
+ *  first). The probability of choosing a 32-byte string at random which is an
```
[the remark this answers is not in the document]

include/secp256k1.h
```diff
@@ -599,57 +613,85 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_negate(
...
- * In:      tweak:  pointer to a 32-byte tweak.
- */
+ * In/Out:  seckey: pointer to a 32-byte secret key. The secret key should be
+ *                  valid according to secp256k1_ec_seckey_verify. If this
+ *                  function returns 0, seckey will be some unspecified
+ *                  value. (cannot be NULL)
+ * In:       tweak: pointer to a 32-byte tweak. Must be in the same range as secret
+ *                  keys (see secp256k1_ec_seckey_verify). For uniformly random
+ *                  32-byte arrays the chance of being out of range is
+ *                  negligible (around 1 in 2^128). (cannot be NULL)
```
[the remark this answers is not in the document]
jonasnick: I don't understand what you're suggesting. Yes, the tweak must be in range. The resulting seckey will be in range if the tweak is not the negation of the seckey.

src/secp256k1.c
```diff
@@ -471,7 +471,7 @@ int secp256k1_ecdsa_sign(const secp256k1_context* ctx, secp256k1_ecdsa_signature
     secp256k1_scalar r, s;
     secp256k1_scalar sec, non, msg;
     int ret = 0;
-    int overflow = 0;
+    int is_sec_valid;
```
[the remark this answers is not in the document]
jonasnick: Afaics it's just wasted work if we fix the duplicated recovery code. Better to make recovery use the common code in a separate PR (right after this one).

src/tests.c
jonasnick: No, seckey is equal to orderc which is overflowing. Added `CHECK(secp256k1_ec_seckey_verify(ctx, ctmp) == 0);`

src/scalar_impl.h
jonasnick: Ugh, fixed, but only saw this in one other function (run_scalar_set_b32_seckey_tests)

include/secp256k1.h
jonasnick: Done

include/secp256k1.h
jonasnick: Done

src/eckey_impl.h
jonasnick: Presumably the compiler optimized away the branching

include/secp256k1.h
jonasnick: Added a commit to clarify

include/secp256k1.h
jonasnick: fixed in separate commit

include/secp256k1.h

include/secp256k1.h

include/secp256k1.h

src/secp256k1.c
```diff
@@ -556,17 +554,23 @@ int secp256k1_ec_pubkey_create(const secp256k1_context* ctx, secp256k1_pubkey *p
...
     secp256k1_scalar_negate(&sec, &sec);
     secp256k1_scalar_get_b32(seckey, &sec);
 
     secp256k1_scalar_clear(&sec);
-    return 1;
+    return ret;
+}
+
+int secp256k1_ec_privkey_negate(const secp256k1_context* ctx, unsigned char *seckey) {
+  return secp256k1_ec_seckey_negate(ctx, seckey);
```
real-or-random: nit: indentation

src/secp256k1.c
```diff
@@ -605,6 +609,10 @@ int secp256k1_ec_privkey_tweak_add(const secp256k1_context* ctx, unsigned char *
     return ret;
 }
 
+int secp256k1_ec_privkey_tweak_add(const secp256k1_context* ctx, unsigned char *seckey, const unsigned char *tweak) {
+  return secp256k1_ec_seckey_tweak_add(ctx, seckey, tweak);
```
real-or-random: here too

src/secp256k1.c
```diff
@@ -649,6 +657,10 @@ int secp256k1_ec_privkey_tweak_mul(const secp256k1_context* ctx, unsigned char *
     return ret;
 }
 
+int secp256k1_ec_privkey_tweak_mul(const secp256k1_context* ctx, unsigned char *seckey, const unsigned char *tweak) {
+  return secp256k1_ec_seckey_tweak_mul(ctx, seckey, tweak);
```
real-or-random: same

include/secp256k1.h
```diff
@@ -599,57 +616,93 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_negate(
...
- *          uniformly random 32-byte arrays, or if the resulting private key
- *          would be invalid (only when the tweak is the complement of the
- *          private key). 1 otherwise.
- * Args:    ctx:    pointer to a context object (cannot be NULL).
- * In/Out:  seckey: pointer to a 32-byte private key.
- * In:      tweak:  pointer to a 32-byte tweak.
- */
+/** Tweak a secret key by adding tweak to it.
+ *
+ *  Returns: 0 if the arguments are invalid or resulting secret key would be
```
real-or-random:
```suggestion
 *  Returns: 0 if the arguments are invalid or the resulting secret key would be
```

src/secp256k1.c
jonasnick: fixed

src/secp256k1.c
jonasnick: fixed

src/secp256k1.c
jonasnick: fixed

include/secp256k1.h

jonasnick:
@apoelstra would you mind reACKing?

PR #703, opened 2019-12-20
Overhaul README.md

real-or-random:
* Update feature list
* Be more positive about the state and quality of the library
* Mention ECDSA key operations explicitly in short library description
* Say "secret key" instead of "private key"

cc @gmaxwell who suggested a similar wording for the disclaimer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,17 +3,22 @@ libsecp256k1
 
 [![Build Status](https://travis-ci.org/bitcoin-core/secp256k1.svg?branch=master)](https://travis-ci.org/bitcoin-core/secp256k1)
 
-Optimized C library for EC operations on curve secp256k1.
+Optimized C library for ECDSA signatures and secret/public key operations on curve secp256k1.
 
-This library is a work in progress and is being used to research best practices. Use at your own risk.
+This library is intended to be the highest quality publicly available library for cryptography on the secp256k1 curve. However, the primary focus of its development has been for usage in the Bitcoin system and usage unlike Bitcoin's may be less well tested, verified, or suffer from a less well thought out interface. Correct usage requires some care and consideration that the library is fit for your application's purpose.
 
 Features:
 * secp256k1 ECDSA signing/verification and key generation.
-* Adding/multiplying private/public keys.
-* Serialization/parsing of private keys, public keys, signatures.
-* Constant time, constant memory access signing and pubkey generation.
-* Derandomized DSA (via RFC6979 or with a caller provided function.)
+* Additive and multiplicative tweaking of secret/public keys.
+* Serialization/parsing of secret keys, public keys, signatures.
+* Constant time, constant memory access signing and public key generation.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,17 +3,20 @@ libsecp256k1
 
 [![Build Status](https://travis-ci.org/bitcoin-core/secp256k1.svg?branch=master)](https://travis-ci.org/bitcoin-core/secp256k1)
 
-Optimized C library for EC operations on curve secp256k1.
+Optimized C library for ECDSA signatures and secret/public key operations on curve secp256k1.
```
real-or-random: If we do this change, we should also change the "GitHub description"

README.md
```diff
@@ -3,17 +3,20 @@ libsecp256k1
...
-* Constant time, constant memory access signing and pubkey generation.
-* Derandomized DSA (via RFC6979 or with a caller provided function.)
+* Additive and multiplicative tweaking of secret/public keys.
+* Serialization/parsing of secret keys, public keys, signatures.
+* Constant time, constant memory access signing and public key generation.
+* Derandomized ECDSA (via RFC6979 or with a caller provided function.)
 * Very efficient implementation.
+* Suitable for embedded systems.
+* Optional module for public key recovery.
+* Optional module for ECDH key exchange (experimental).
```
real-or-random: not sure if we really should mention experimental features

jonasnick: I think experimental modules can be mentioned, but only if we define what experimental means. We don't define that anywhere afaik so it would be good to do that in any case. I'm thinking of something like "don't use this in production, unless you know what you're doing", "the module may promise properties that don't actually hold (like constant timeness in ECDH), but we'd appreciate if you help test and review it".

real-or-random: "Experimental features have not received enough scrutiny to satisfy the standard of quality of this library but are made available for testing and review by the community. The APIs of these features should not be considered stable."

Ok?

jonasnick: sgtm

sipa:
ACK 2e759ec753446aab0272ba32c5f1b7dc3a4dc75c

hebasto:
>    * Say "secret key" instead of "private key"

While I was [porting](https://codeberg.org/guix/guix/pulls/5967) this change to the Guix `libsecp256k1` package description, the maintainer raised this [comment](https://codeberg.org/guix/guix/pulls/5967#issuecomment-10803065):

>  public/private are common terms, secret key not sure about.

Could @real-or-random or someone else provide a bit more detail to help justify the wording change?

real-or-random:
> > public/private are common terms, secret key not sure about.
> 
> Could @real-or-random or someone else provide a bit more detail to help justify the wording change?

There's no deep reason. It's about consistency. We need to pick one, and "secret" is a reasonable choice. I believe "secret" is much more common in the cryptographic literature. If anything, it matches the common abbreviation "sk".

edit:
And the "High-performance, high-assurance" thing is just our aspiration. If they want more neutral language, I wouldn't try to argue with them...

PR #713, opened 2020-01-16
Docstrings

benma:


benma:
As a newcomer to the library it is a bit hard to understand what the field and scalar functions do in terms of modulo reductions. The docstrings added here would have helped me, provided they are correct :smile:

src/scalar.h
```diff
@@ -32,7 +32,11 @@ static unsigned int secp256k1_scalar_get_bits(const secp256k1_scalar *a, unsigne
 /** Access bits from a scalar. Not constant time. */
 static unsigned int secp256k1_scalar_get_bits_var(const secp256k1_scalar *a, unsigned int offset, unsigned int count);
 
-/** Set a scalar from a big endian byte array. */
+/** Set a scalar from a big endian byte array. The scalar will be reduced modulo group order `n`.
+ * In:      bin:   pointer to a 32-byte array.
+ * Out:     r:     scalar to be loaded.
+ *          overflow: non-zero if the scalar was larger than `n` before reduction, zero otherwise.
```
elichai: Maybe change to "if not null is set to..."

benma: fixed by adding "(can be NULL)" as seen elsewhere docstrings.

src/scalar.h
```diff
@@ -32,7 +32,11 @@ static unsigned int secp256k1_scalar_get_bits(const secp256k1_scalar *a, unsigne
 /** Access bits from a scalar. Not constant time. */
 static unsigned int secp256k1_scalar_get_bits_var(const secp256k1_scalar *a, unsigned int offset, unsigned int count);
 
-/** Set a scalar from a big endian byte array. */
+/** Set a scalar from a big endian byte array. The scalar will be reduced modulo group order `n`.
+ * In:      bin:        pointer to a 32-byte array.
+ * Out:     r:          scalar to be loaded.
```
real-or-random: maybe s/loaded/set ?

benma: :heavy_check_mark:

real-or-random:
ACK dabfea7e217b129d10f0f787722626f388dddb5a

PR #719, opened 2020-02-17
Add Multiset/ECMH implementation

elichai:
Hi, the first commit is cherry-picked from https://github.com/Bitcoin-ABC/secp256k1/tree/master/src/modules/multiset
and  then I implemented a serliaze/parse functions and fixed wrong docs and some other minor stuff.
Best reviewed commit by commit :)


elichai:
Ops

PR #722, opened 2020-03-04
Context isn't freed in the ECDH benchmark

elichai:
Hi,
Currently the context is leaked in `bench_ecdh`.
Wanted to prevent this happening in the future by adding it to the valgrind run in travis.
for that I made `run_benchmark` pass the number of iterations as a variable into the benchmarks so that the iteration number it uses to print results actually matches the one used in the benchmark.
and added a way to pass `-DITERS=N` to change the baseline number of iterations used in the benchmarks(so that we can run the benchmarks with `ITERS=1` under travis to make sure they compile, run, don't leak or have UB)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ compiler:
   - gcc
 env:
   global:
-    - FIELD=auto  BIGNUM=auto  SCALAR=auto  ENDOMORPHISM=no  STATICPRECOMPUTATION=yes  ECMULTGENPRECISION=auto  ASM=no  BUILD=check  EXTRAFLAGS=  HOST=  ECDH=no  RECOVERY=no  EXPERIMENTAL=no CTIMETEST=yes
+    - FIELD=auto  BIGNUM=auto  SCALAR=auto  ENDOMORPHISM=no  STATICPRECOMPUTATION=yes  ECMULTGENPRECISION=auto  ASM=no  BUILD=check  EXTRAFLAGS=  HOST=  ECDH=no  RECOVERY=no  EXPERIMENTAL=no CTIMETEST=yes BENCH=yes SECP256K1_BENCH_ITERS=2
   matrix:
     - SCALAR=32bit    RECOVERY=yes
     - SCALAR=32bit    FIELD=32bit       ECDH=yes  EXPERIMENTAL=yes
@@ -25,7 +25,7 @@ env:
     - BIGNUM=no
     - BIGNUM=no       ENDOMORPHISM=yes RECOVERY=yes EXPERIMENTAL=yes
     - BIGNUM=no       STATICPRECOMPUTATION=no
-    - BUILD=distcheck CTIMETEST=
+    - BUILD=distcheck CTIMETEST= BENCH=
     - EXTRAFLAGS=CPPFLAGS=-DDETERMINISTIC
     - EXTRAFLAGS=CFLAGS=-O0
     - ECMULTGENPRECISION=2
@@ -94,6 +94,12 @@ script:
    travis_wait 30 valgrind --error-exitcode=42 ./tests 16 &&
    travis_wait 30 valgrind --error-exitcode=42 ./exhaustive_tests;
    fi
+ - if [ -n "$BENCH" ]; then
+     if [ -n "$VALGRIND" ]; then EXEC='libtool --mode=execute valgrind --error-exitcode=42'; else EXEC= ; fi &&
+     $EXEC ./bench_ecmult &>> bench.log && $EXEC ./bench_internal &>> bench.log && $EXEC ./bench_sign &>> bench.log && $EXEC ./bench_verify &>> bench.log &&
+     if [ "$RECOVERY" == "yes" ]; then $EXEC ./bench_recover &>> bench.log; fi &&
+     if [ "$ECDH" == "yes" ]; then $EXEC ./bench_ecdh &>> bench.log; fi;
+   fi
  - if [ -n "$CTIMETEST" ]; then
    libtool --mode=execute valgrind  ./valgrind_ctime_test &> valgrind_ctime_test.log;
    fi
@@ -102,3 +108,4 @@ after_script:
     - cat ./tests.log
     - cat ./exhaustive_tests.log
     - cat ./valgrind_ctime_test.log
+    - cat ./bench.log
--- a/src/bench.h
+++ b/src/bench.h
@@ -73,7 +73,7 @@ void print_number(const int64_t x) {
     printf("%s", &buffer[ptr]);
 }
 
-void run_benchmark(char *name, void (*benchmark)(void*), void (*setup)(void*), void (*teardown)(void*), void* data, int count, int iter) {
+void run_benchmark(char *name, void (*benchmark)(void*, int), void (*setup)(void*), void (*teardown)(void*, int), void* data, int count, int iter) {
     int i;
     int64_t min = INT64_MAX;
     int64_t sum = 0;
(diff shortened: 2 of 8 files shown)


real-or-random:
Concept ACK. 

This is neat. Travis fails at the moment: https://travis-ci.org/bitcoin-core/secp256k1/jobs/658249366#L447

A few observations:
 - If we already make `iter` configurable, wouldn't it be more convenient to obtain the parameter at runtime? 
 - Should we make `count` configurable too?

elichai:
>     * If we already make `iter` configurable, wouldn't it be more convenient to obtain the parameter at runtime?

That's what I did at first, but it somewhat collides with `have_flag` so I decided to start with a simpler version and if people think it's a useful feature I can implement have_flag in a way that will work both with the number of iterations and the flags without ambiguity(ie `--tests=N`)

>     * Should we make `count` configurable too?

Count doesn't seem to affect performance that match, so I didn't need to do that. but if we'll add runtime configurations we can definitely add that too.

elichai:
> Concept ACK.
> 
> This is neat. Travis fails at the moment: https://travis-ci.org/bitcoin-core/secp256k1/jobs/658249366#L447

Right, forgot I need to check if recovery and/or ecdh are compiled or not. that will be a big enough logic that I'm thinking of moving this to autotools maybe (although that will make it hard again to run under valgrind) Will fix this later/tomorrow and see how complex it will look

sipa:
configure: error: unrecognized option: `-DITERS=1'

jonasnick:
I agree that it would be better to make the benchmark iterations runtime configurable. Perhaps `getenv` would be an acceptable workaround?

real-or-random:
> I agree that it would be better to make the benchmark iterations runtime configurable. Perhaps `getenv` would be an acceptable workaround?

I think an environment variable is not bad because it can persist and it applies to all benchmarks. That fits our usage of the benchmarks.

src/bench.h
```diff
@@ -121,4 +121,13 @@ int have_flag(int argc, char** argv, char *flag) {
     return 0;
 }
 
+int get_iters(int default_iters) {
+    char* env = getenv("ITERS");
```
real-or-random: maybe change it to something with more meaning and less likely to collide with other software, e.g., `SECP256K1_BENCH_ITERS`

elichai: even though it's just the for the benchmarks? we also have `int have_flag(..);` and `void print_number(..);` in that header.

real-or-random: I'm talking about the environment variable, not the function. :)

elichai: Done

.travis.yml
```diff
@@ -94,6 +94,12 @@ script:
    travis_wait 30 valgrind --error-exitcode=42 ./tests 16 &&
    travis_wait 30 valgrind --error-exitcode=42 ./exhaustive_tests;
    fi
+ - if [ -n "$BENCH" ]; then
```
jonasnick: `BENCH` is never set

src/bench_internal.c
```diff
@@ -329,51 +329,53 @@ void bench_num_jacobi(void* arg) {
...
-    if (have_flag(argc, argv, "hash") || have_flag(argc, argv, "hmac")) run_benchmark("hash_hmac_sha256", bench_hmac_sha256, bench_setup, NULL, &data, 10, 20000);
-    if (have_flag(argc, argv, "hash") || have_flag(argc, argv, "rng6979")) run_benchmark("hash_rfc6979_hmac_sha256", bench_rfc6979_hmac_sha256, bench_setup, NULL, &data, 10, 20000);
+    if (have_flag(argc, argv, "hash") || have_flag(argc, argv, "sha256")) run_benchmark("hash_sha256", bench_sha256, bench_setup, NULL, &data, 10, iters);
+    if (have_flag(argc, argv, "hash") || have_flag(argc, argv, "hmac")) run_benchmark("hash_hmac_sha256", bench_hmac_sha256, bench_setup, NULL, &data, 10, iters);
+    if (have_flag(argc, argv, "hash") || have_flag(argc, argv, "rng6979")) run_benchmark("hash_rfc6979_hmac_sha256", bench_rfc6979_hmac_sha256, bench_setup, NULL, &data, 10, iters);
 
     if (have_flag(argc, argv, "context") || have_flag(argc, argv, "verify")) run_benchmark("context_verify", bench_context_verify, bench_setup, NULL, &data, 10, 20);
     if (have_flag(argc, argv, "context") || have_flag(argc, argv, "sign")) run_benchmark("context_sign", bench_context_sign, bench_setup, NULL, &data, 10, 200);
```
jonasnick: Is this and context_verify intentionally left with 200 iters? Would be more consistent to use `1 + iters/100` or something.

src/bench_ecmult.c
```diff
@@ -185,14 +186,17 @@ int main(int argc, char **argv) {
...
 
     for (i = 1; i <= 8; ++i) {
-        run_test(&data, i, 1);
+        run_test(&data, i, 1, iters);
     }
 
-    for (p = 0; p <= 11; ++p) {
-        for (i = 9; i <= 16; ++i) {
-            run_test(&data, i << p, 1);
+    if (iters > 1) {
```
jonasnick: Would be good to add comment why that's useful.

src/bench_sign.c
```diff
@@ -14,6 +14,7 @@ typedef struct {
     unsigned char key[32];
 } bench_sign;
 
+
```
jonasnick: nit: There are 2 more unnecessary newlines I think.

.travis.yml
elichai: oh I broke it when I replaced the precompiled macro with the env.
Thanks!

src/bench_internal.c
elichai: nope. missed it (I used regex to do the replacing and didn't consider so little zeroes :P )

gmaxwell:
FWIW, when you setup 'fast' CI tests, please at least run them with 2 iterations, not 1.  1 iteration will fail to catch some kinds of bugs in the test harnesses, e.g. use after free if the deallocations are in the wrong place with the testing loops, etc.  If 2 is too much, maybe resize the underlying work units so that they're not.

src/bench_sign.c
elichai: Done

src/bench_ecmult.c
elichai: Done

src/bench_internal.c
elichai: Done

.travis.yml
elichai: Done

jonasnick: Better to unset `BENCH` int the `BUILD=distcheck` case similar to `CTIMETEST` because the bench binaries are not build in that case.

elichai: Hmm so why didn't the tests fail

elichai: Maybe because there's `;` between the ifs? hmmm

elichai: Got it to fail: https://travis-ci.org/github/elichai/secp256k1/jobs/663963345
Will fix now.
Thanks.

real-or-random: ugh yep, I think we want `&&` to make sure that the final exit code of the statement.

I think it's cleaner to have individual lines in yaml even.

real-or-random: > I think it's cleaner to have individual lines in yaml even.
Then the execution continues. With `&&` we stop after the first failing command, which is okay but not optimal.

elichai: Can you split an `if` across different yaml lines?

real-or-random: Now, but you could different `if`s with combined conditions on  different lines. If you want to try, feel free to add a commit on top. Otherwise tell us and let's get his merged, it's certainly good as it is.

elichai: I'm kinda tired of fighting with travis, so unless there's any objections I prefer to get it in as is

real-or-random:
ACK 85b35afa765ba23ac3682cf15800769b0a3b834d I looked at the diff and tested the ecdh benchmark using valgrind

real-or-random:
This is minor inconvenience that I can fix on the way, e.g. in #723:
```
> SECP256K1_BENCH_ITERS= ./bench_ecdh 
[1]    96654 floating point exception (core dumped)  
```

PR #732, opened 2020-03-31
Retry if r is zero during signing

### LINKED ISSUE #720: Erroneous comment and VERIFY_CHECK in ecdsa_sig_sign
I've been following along the ECDSA signing code and have come across what I think is an incorrect comment. In `secp256k1_ecdsa_sig_sign` the code makes the following recommendation:

https://github.com/bitcoin-core/secp256k1/blob/96d8ccbd16090551aa003bfa4acd108b0496cb89/src/ecdsa_impl.h#L285-L293

However, these two conditions are not actually checked before calling. The main call doesn't:

https://github.com/bitcoin-core/secp256k1/blob/96d8ccbd16090551aa003bfa4acd108b0496cb89/src/secp256k1.c#L491-L510

Note that `sigr` is the nonce (R)'s x coordinate reduced modulo q as a scalar. The caller is only checking properties of the secret nonce not `sigr` (it doesn't even compute the point for R). Maybe this code was meant to check the secret nonce that's passed in instead?

There doesn't seem to be any problem with `sigr` overlfowing (the verification algorithm actually accounts for this possibility by adding the curve order to `sigr` if it doesn't work the first time!). `sigr` being 0 is a problem but the probability of this happening is so low I wouldn't bother accounting for it. I think this comment and VERIFY_CHECKS can be removed.

More generally there doesn't seem to be a problem with the secret nonce overflowing either but the while loop in the caller enforces that and that it's not 0.

real-or-random:
This reverts commit 25e3cfbf9b52d2f5afa543f967a73aa8850d2038. The reverted
commit was probably based on the assumption that this is about the touched
checks cover the secret nonce k instead of r, which is the x-coord of the public
nonce. A signature with a zero r is invalid by the spec, so we should return 0
to make the caller retry with a different nonce. Overflow is not an issue.

Then a second commit makes the resulting code constant-time in whether r is zero,
in order to pass the new constant-time test. 
    
Fixes #720.

### DIFF
--- a/src/ecdsa_impl.h
+++ b/src/ecdsa_impl.h
@@ -288,14 +288,6 @@ static int secp256k1_ecdsa_sig_sign(const secp256k1_ecmult_gen_context *ctx, sec
     secp256k1_fe_normalize(&r.y);
     secp256k1_fe_get_b32(b, &r.x);
     secp256k1_scalar_set_b32(sigr, b, &overflow);
-    if (secp256k1_scalar_is_zero(sigr)) {
-        /* P.x = order is on the curve, so technically sig->r could end up zero, which would be an invalid signature.
-         * This branch is cryptographically unreachable as hitting it requires finding the discrete log of P.x = N.
-         */
-        secp256k1_gej_clear(&rp);
-        secp256k1_ge_clear(&r);
-        return 0;
-    }
     if (recid) {
         /* The overflow condition is cryptographically unreachable as hitting it requires finding the discrete log
          * of some P where P.x >= order, and only 1 in about 2^127 points meet this criteria.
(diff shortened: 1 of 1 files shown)


elichai:
ACK 37ed51a7eafa5fd9955ceb66213251cdd8b8b8c4 makes sense.
took me a minute to understand the comment but it looks correct.

can we somehow test this condition under the exhaustive tests?

jonasnick:
If I understand correctly this issue may not show up in the exhaustive tests if the curve order (`EXHAUSTIVE_TEST_ORDER`) is not a valid X coordinate.

real-or-random:
> If I understand correctly this issue may not show up in the exhaustive tests if the curve order (`EXHAUSTIVE_TEST_ORDER`) is not a valid X coordinate.

Indeed. I'd be willing to write a test but I don't see a way to come up with a test vector. So this is ready for review.

PR #735, opened 2020-04-09
build: fix OpenSSL EC detection on macOS

### LINKED ISSUE #734: build: OpenSSL EC detection fails on macOS
```bash
./configure --enable-openssl-tests
...
checking for libcrypto... yes
checking for EC functions in libcrypto... no
configure: error: OpenSSL tests requested but OpenSSL with EC support is not available
```

config.log
```
configure:14672: checking for EC functions in libcrypto
configure:14696: gcc -c -O2 -fvisibility=hidden -std=c89 -pedantic -Wall -Wextra -Wcast-align -Wnested-externs -Wshadow -Wstrict-prototypes -Wno-unused-function -Wno-long-long -Wno-overlength-strings -W -g  conftest.c >&5
conftest.c:36:14: fatal error: 'openssl/ec.h' file not found
    #include <openssl/ec.h>
             ^~~~~~~~~~~~~~
1 error generated.
```

I can PR a fix shortly if no-one beats me to it.

fanquake:
Fixes #734.

The first commit fixes OpenSSL EC detection on macOS, by making `CRYPTO_CPPFLAGS` available during configure. Similar to how `GMP_CPPFLAGS` are used.

The second commit, which I'd like feedback on, fixes a compilation error:
```bash
  CC       src/bench_verify.o
src/bench_verify.c:15:10: fatal error: 'openssl/bn.h' file not found
#include <openssl/bn.h>
         ^~~~~~~~~~~~~~
1 error generated.
```

as `CRYPTO_CPPFLAGS` are then needed in bench_verify, and are not included otherwise. 

Not sure if this is the way you'd like to fix this, as its a bit of a hack for macOS.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -78,6 +78,8 @@ if USE_BENCHMARK
 noinst_PROGRAMS += bench_verify bench_sign bench_internal bench_ecmult
 bench_verify_SOURCES = src/bench_verify.c
 bench_verify_LDADD = libsecp256k1.la $(SECP_LIBS) $(SECP_TEST_LIBS) $(COMMON_LIB)
+# SECP_TEST_INCLUDES are only used here for CRYPTO_CPPFLAGS
+bench_verify_CPPFLAGS = -DSECP256K1_BUILD $(SECP_TEST_INCLUDES)
 bench_sign_SOURCES = src/bench_sign.c
 bench_sign_LDADD = libsecp256k1.la $(SECP_LIBS) $(SECP_TEST_LIBS) $(COMMON_LIB)
 bench_internal_SOURCES = src/bench_internal.c
--- a/build-aux/m4/bitcoin_secp.m4
+++ b/build-aux/m4/bitcoin_secp.m4
@@ -38,6 +38,8 @@ AC_DEFUN([SECP_OPENSSL_CHECK],[
   fi
 if test x"$has_libcrypto" = x"yes" && test x"$has_openssl_ec" = x; then
   AC_MSG_CHECKING(for EC functions in libcrypto)
+  CPPFLAGS_TEMP="$CPPFLAGS"
+  CPPFLAGS="$CRYPTO_CPPFLAGS $CPPFLAGS"
   AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
     #include <openssl/ec.h>
     #include <openssl/ecdsa.h>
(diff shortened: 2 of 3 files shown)


real-or-random:
Thanks! I'm fine with the code changes but I can't test on macOS. The second commit seems correct and I don't see why it should be a hack.

gwillen:
I have a mac (10.14.6) and am willing to test this -- can someone give me a fairly precise description of a negative test and a positive test for me to try?

fanquake:
@gwillen 

Master 4f27e344c69c33b4f3f448baa0196b9892287081:
```bash
./configure --enable-openssl-tests
...
checking for libcrypto... yes
checking for EC functions in libcrypto... no
configure: error: OpenSSL tests requested but OpenSSL with EC support is not available
```

This PR @ https://github.com/bitcoin-core/secp256k1/pull/735/commits/84b5fc5bc34cfb6e9df4f06b3b6fc98c3373dd0e. Configure works, compiling doesn't:
```bash
./configure --enable-openssl-tests
# make files
  CC       src/bench_verify.o
src/bench_verify.c:15:10: fatal error: 'openssl/bn.h' file not found
#include <openssl/bn.h>
         ^~~~~~~~~~~~~~
1 error generated.
```

This PR @ https://github.com/bitcoin-core/secp256k1/pull/735/commits/3b7d26b23c9811deee244ede7bb344bb43544153. Everything works as expected.

gwillen:
Is this all meant to be done with brew's openssl linked? On my system, brew says:

`openssl@1.1 is keg-only, which means it was not symlinked into /usr/local, because openssl/libressl is provided by macOS so don't link an incompatible version.`

And I get the configure error regardless of which version I check out.

fanquake:
> Is this all meant to be done with brew's openssl linked?

No

> I get the configure error regardless of which version I check out.

Did you run autogen.sh before running configure after checking out the changes?

gwillen:
Aha, nope I did not. That causes the behavior to be exactly as predicted! Configure works at 84b5fc5, build works at 3b7d26b. Let me know if you want me to check anything else.

real-or-random:
ACK 3b7d26b23c9811deee244ede7bb344bb43544153 the diff looks good to me

I think it's okay to merge this before now even though #738 is there. If we think that we can quickly merge #738, we can also do this instead.

PR #740, opened 2020-04-15
Make recovery/main_impl.h non-executable

elichai:
Opened it because of https://github.com/bitcoin/bitcoin/pull/18650

I assume it doesn't matter?
But because I'm not sure I preferred to open this then let the info go away in case someone thinks it does matter.


real-or-random:
I assume you checked all the files and this is the only with a questionable x?

elichai:
Yes.
```
find $(git ls-files) -executable
```

real-or-random:
ACK ffef45c98a55934ab555ecf96f0b19d0e4cfe830

PR #741, opened 2020-04-17
Remove unnecessary sign variable from wnaf_const

jonasnick:
There currently is a single branch in the `ecmul_const` function that is not being exercised by the tests. This branch is unreachable and therefore I'm suggesting to remove it.

For your convenience the paper the wnaf algorithm can be found [here (The Width-w NAF Method Provides Small Memory and Fast Elliptic Scalar Multiplications Secure against Side Channel Attacks)](http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.563.1267&rep=rep1&type=pdf). Similarly, unless I'm missing something important, I don't see how their algorithm needs to consider `sign(u[i-1])` unless `d` can be negative - which doesn't make much sense to me either.

### DIFF
--- a/src/ecmult_const_impl.h
+++ b/src/ecmult_const_impl.h
@@ -101,16 +101,22 @@ static int secp256k1_wnaf_const(int *wnaf, const secp256k1_scalar *scalar, int w
     /* 4 */
     u_last = secp256k1_scalar_shr_int(&s, w);
     do {
-        int sign;
         int even;
 
         /* 4.1 4.4 */
         u = secp256k1_scalar_shr_int(&s, w);
         /* 4.2 */
         even = ((u & 1) == 0);
-        sign = 2 * (u_last > 0) - 1;
-        u += sign * even;
-        u_last -= sign * even * (1 << w);
+        /* In contrast to the original algorithm, u_last is always > 0 and
+         * therefore we do not need to check its sign. In particular, it's easy
+         * to see that u_last is never < 0 because u is never < 0. Moreover,
+         * u_last is never = 0 because u is never even after a loop
+         * iteration. The same holds analogously for the initial value of
+         * u_last (in the first loop iteration). */
+        VERIFY_CHECK(u_last > 0);
+        VERIFY_CHECK((u_last & 1) == 1);
+        u += even;
+        u_last -= even * (1 << w);
 
         /* 4.3, adapted for global sign change */
         wnaf[word++] = u_last * global_sign;
--- a/src/tests.c
+++ b/src/tests.c
@@ -3190,6 +3190,7 @@ void test_constant_wnaf(const secp256k1_scalar *number, int w) {
     int skew;
     int bits = 256;
     secp256k1_scalar num = *number;
+    secp256k1_scalar scalar_skew;
 
     secp256k1_scalar_set_int(&x, 0);
     secp256k1_scalar_set_int(&shift, 1 << w);
@@ -3220,7 +3221,8 @@ void test_constant_wnaf(const secp256k1_scalar *number, int w) {
         secp256k1_scalar_add(&x, &x, &t);
     }
     /* Skew num because when encoding numbers as odd we use an offset */
-    secp256k1_scalar_cadd_bit(&num, skew == 2, 1);
+    secp256k1_scalar_set_int(&scalar_skew, 1 << (skew == 2));
+    secp256k1_scalar_add(&num, &num, &scalar_skew);
     CHECK(secp256k1_scalar_eq(&x, &num));
 }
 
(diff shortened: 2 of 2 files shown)


src/ecmult_const_impl.h
```diff
@@ -101,16 +101,19 @@ static int secp256k1_wnaf_const(int *wnaf, const secp256k1_scalar *scalar, int w
...
         u = secp256k1_scalar_shr_int(&s, w);
         /* 4.2 */
         even = ((u & 1) == 0);
-        sign = 2 * (u_last > 0) - 1;
-        u += sign * even;
-        u_last -= sign * even * (1 << w);
+        /* In contrast to the original algorithm, u_last is always > 0 and
+         * therefore we do not need to check its sign. In particular, it's easy
+         * to see that u_last is never < 0 because u is never < 0. Moreover,
+         * u_last is never = 0 because u is never even. */
```
real-or-random, line 113: I think "u is never even" is not true.

jonasnick: should say "u is never even after a loop iteration"

src/ecmult_const_impl.h
```diff
@@ -101,16 +101,19 @@ static int secp256k1_wnaf_const(int *wnaf, const secp256k1_scalar *scalar, int w
...
+        VERIFY_CHECK(u_last > 0);
```
real-or-random, line 114:
```suggestion
        VERIFY_CHECK(u_last > 0);
        VERIFY_CHECK(u_last & 1 == 0);
```

src/tests.c
```diff
@@ -3220,7 +3221,8 @@ void test_constant_wnaf(const secp256k1_scalar *number, int w) {
         secp256k1_scalar_add(&x, &x, &t);
     }
     /* Skew num because when encoding numbers as odd we use an offset */
-    secp256k1_scalar_cadd_bit(&num, skew == 2, 1);
+    secp256k1_scalar_set_int(&scalar_skew, 1 << (skew == 2));
+    secp256k1_scalar_add(&num, &num, &scalar_skew);
```
real-or-random, line 3225: Can you explain this change?

jonasnick: I added a test with `num = n-1`, but `cadd_bit` can not deal with overflows. Will add that to the commit description.

src/ecmult_const_impl.h
jonasnick: I would argue that this is excessive given that the `u_last > 0` is only there to show a single specific difference to the paper and the whole point of the algorithm is that `u` is never even. But I can see that this slightly helps review if you don't look at the paper. I added this line.

jonasnick: should be `VERIFY_CHECK((u_last & 1) == 1);` ("is never even")

real-or-random: Don't trust code I've written using the GitHub web interface.

src/tests.c
```diff
@@ -3332,13 +3334,19 @@ void run_wnaf(void) {
...
     test_constant_wnaf(&n, 4);
     n.d[0] = 2;
     test_constant_wnaf(&n, 4);
-    /* Test 0 */
+    /* Test -1, because it's a special case in wnaf_const */
+    n = secp256k1_scalar_one;
+    secp256k1_scalar_negate(&n, &n);
+    test_constant_wnaf(&n, 4);
+
+    /* Test 0 for fixed wnaf */
```
real-or-random, line 3349: This comment should be in above line 3337?

jonasnick: There's a difference between constant and "fixed" wnaf. This comment is about the fixed wnaf call below. If that's confusing I think we can just remove it.

real-or-random: Okay, I really got confused because the commit message said "This commit also adds a test for 0." and then my impression was that you added the test but the comment in different places. 

Don't remove it. The resulting code is not confusing, just the diff confused me.

real-or-random:
ACK modulo this:

It took me some more effort to think about the invariant in the first loop iteration, which depends on all that negation/flipping stuff before the loop. 

This made me add two tests tests, which for example catch a bug if we'd change `flip` to `secp256k1_scalar_is_high(&s)` here: https://github.com/bitcoin-core/secp256k1/pull/741/files#diff-bb69027d01eaa0554f24272ea41380a0L97`. So this was not covered by the tests previously. But it's not worth to have a separate PR. Can you cherry-pick the two top commits from mhttps://github.com/real-or-random/secp256k1/tree/ecmult_const_playground and squash this if you agree with the additions?

jonasnick:
Thanks for the additional edge case tests. Cherry-picked and squashed.

real-or-random:
ACK 37dba329c6cb0f7a4228a11dc26aa3a342a3a5d0 I verified the correctness of the change and claimed invariant by manual inspection. I tested the code, both with 32bit and 64bit scalars.

gmaxwell:
ACK.  I reviewed the code and agree that it cannot be negative. I also added a test to check vastly more values, tested with endomorphism on-and-off, and introduced several faults and confirmed that the tests work.

Issue #184, opened 2015-01-11
Documentation should provide example callers.


gmaxwell:
We've seen some people attempting to use the library calling into random internal functions. That should be harder to do now... but now that the basic signature interface is relatively safe to use (as such things go) there should be some easy copy and paste examples for correct usage.

jonasnick:
As an example we have someting like this in secp256k1-zkp to demonstrate musig: https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/src/modules/musig/example.c

NicholasFarrow:
For anyone else interested in how to generate keys using this library, feel free to checkout my [Bitcoin vanity address generator](https://github.com/NicholasFarrow/niceBit/) which I think provides a short and [reasonably well commented example](https://github.com/NicholasFarrow/niceBit/blob/master/niceBit.c) of how to generate private and public key pairs.

jonasnick:
I'd suggest to add a short and well documented example source file for every module and build and run it as part of `make check`. We can list the modules in the README and from there link to the examples and include files for more documentation.

PR #748, opened 2020-04-30
Add usage examples

elichai:
Resolves https://github.com/bitcoin-core/secp256k1/issues/184

I didn't add a sign/verify/recover example yet because I'd prefer that the example included hashing a message, 
but we don't provide a hash function so I decided to start only with key generation and ECDH, and get feedback.

I *did not* check the example on windows, but did try to implement it correctly, if anyone has access to a windows machine(and knows how to build on windows hehe) please test this :)

P.S. Should the usage examples be MIT or CC0(Public Domain)?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,11 +6,16 @@ exhaustive_tests
 precompute_ecmult_gen
 precompute_ecmult
 valgrind_ctime_test
+ecdh_example
+ecdsa_example
+schnorr_example
 *.exe
 *.so
 *.a
 *.csv
 !.gitignore
+*.log
+*.trs
 
 Makefile
 configure
--- a/Makefile.am
+++ b/Makefile.am
@@ -63,6 +63,7 @@ noinst_HEADERS += contrib/lax_der_parsing.h
 noinst_HEADERS += contrib/lax_der_parsing.c
 noinst_HEADERS += contrib/lax_der_privatekey_parsing.h
 noinst_HEADERS += contrib/lax_der_privatekey_parsing.c
+noinst_HEADERS += examples/random.h
 
 PRECOMPUTED_LIB = libsecp256k1_precomputed.la
 noinst_LTLIBRARIES = $(PRECOMPUTED_LIB)
@@ -137,6 +138,40 @@ endif
 exhaustive_tests_LDADD = $(SECP_LIBS) $(COMMON_LIB)
 exhaustive_tests_LDFLAGS = -static
 TESTS += exhaustive_tests
+endif
+
+if USE_EXAMPLES
+noinst_PROGRAMS += ecdsa_example
+ecdsa_example_SOURCES = examples/ecdsa.c
+ecdsa_example_CPPFLAGS = -I$(top_srcdir)/include
+ecdsa_example_LDADD = libsecp256k1.la
+ecdsa_example_LDFLAGS = -static
+if BUILD_WINDOWS
+ecdsa_example_LDFLAGS += -lbcrypt
+endif
+TESTS += ecdsa_example
+if ENABLE_MODULE_ECDH
+noinst_PROGRAMS += ecdh_example
+ecdh_example_SOURCES = examples/ecdh.c
+ecdh_example_CPPFLAGS = -I$(top_srcdir)/include
+ecdh_example_LDADD = libsecp256k1.la
+ecdh_example_LDFLAGS = -static
+if BUILD_WINDOWS
+ecdh_example_LDFLAGS += -lbcrypt
+endif
+TESTS += ecdh_example
+endif
+if ENABLE_MODULE_SCHNORRSIG
+noinst_PROGRAMS += schnorr_example
+schnorr_example_SOURCES = examples/schnorr.c
+schnorr_example_CPPFLAGS = -I$(top_srcdir)/include
+schnorr_example_LDADD = libsecp256k1.la
+schnorr_example_LDFLAGS = -static
+if BUILD_WINDOWS
+schnorr_example_LDFLAGS += -lbcrypt
+endif
+TESTS += schnorr_example
+endif
 endif
 
 ### Precomputed tables
--- a/README.md
+++ b/README.md
@@ -69,6 +69,14 @@ libsecp256k1 is built using autotools:
     $ make check  # run the test suite
     $ sudo make install  # optional
 
+Usage examples
+-----------
+  Usage examples can be found in the [examples](examples) directory. To compile them you need to configure with `--enable-examples`.
+  For experimental modules, you will also need `--enable-experimental` as well as a flag for each individual module, e.g. `--enable-module-ecdh`.
+  * [ECDSA example](examples/ecdsa.c)
+  * [Schnorr Signatures example](examples/schnorr.c)
+  * [Deriving a shared secret(ECDH) example](examples/ecdh.c)
+
 Test coverage
 -----------
 
--- a/configure.ac
+++ b/configure.ac
@@ -45,6 +45,8 @@ if test x"$ac_cv_prog_cc_c89" = x"no"; then
 fi
 AM_PROG_AS
 
+build_windows=no
+
 case $host_os in
   *darwin*)
      if  test x$cross_compiling != xyes; then
@@ -67,6 +69,9 @@ case $host_os in
          fi
        fi
      fi
+   ;;
+   cygwin*|mingw*)
+     build_windows=yes
    ;;
 esac
 
@@ -134,6 +139,11 @@ AC_ARG_ENABLE(exhaustive_tests,
     AS_HELP_STRING([--enable-exhaustive-tests],[compile exhaustive tests [default=yes]]),
     [use_exhaustive_tests=$enableval],
     [use_exhaustive_tests=yes])
+
+AC_ARG_ENABLE(examples,
+    AS_HELP_STRING([--enable-examples],[compile the examples [default=no]]),
+    [use_examples=$enableval],
+    [use_examples=no])
 
 AC_ARG_ENABLE(module_ecdh,
     AS_HELP_STRING([--enable-module-ecdh],[enable ECDH shared secret computation]),
@@ -393,6 +403,7 @@ AC_SUBST(SECP_CFLAGS)
 AM_CONDITIONAL([ENABLE_COVERAGE], [test x"$enable_coverage" = x"yes"])
 AM_CONDITIONAL([USE_TESTS], [test x"$use_tests" != x"no"])
 AM_CONDITIONAL([USE_EXHAUSTIVE_TESTS], [test x"$use_exhaustive_tests" != x"no"])
+AM_CONDITIONAL([USE_EXAMPLES], [test x"$use_examples" != x"no"])
 AM_CONDITIONAL([USE_BENCHMARK], [test x"$use_benchmark" = x"yes"])
 AM_CONDITIONAL([ENABLE_MODULE_ECDH], [test x"$enable_module_ecdh" = x"yes"])
 AM_CONDITIONAL([ENABLE_MODULE_RECOVERY], [test x"$enable_module_recovery" = x"yes"])
@@ -400,6 +411,7 @@ AM_CONDITIONAL([ENABLE_MODULE_EXTRAKEYS], [test x"$enable_module_extrakeys" = x"
 AM_CONDITIONAL([ENABLE_MODULE_SCHNORRSIG], [test x"$enable_module_schnorrsig" = x"yes"])
 AM_CONDITIONAL([USE_EXTERNAL_ASM], [test x"$use_external_asm" = x"yes"])
 AM_CONDITIONAL([USE_ASM_ARM], [test x"$set_asm" = x"arm"])
+AM_CONDITIONAL([BUILD_WINDOWS], [test "$build_windows" = "yes"])
 AC_SUBST(LIB_VERSION_CURRENT, _LIB_VERSION_CURRENT)
 AC_SUBST(LIB_VERSION_REVISION, _LIB_VERSION_REVISION)
 AC_SUBST(LIB_VERSION_AGE, _LIB_VERSION_AGE)
@@ -417,6 +429,7 @@ echo "  with external callbacks = $use_external_default_callbacks"
 echo "  with benchmarks         = $use_benchmark"
 echo "  with tests              = $use_tests"
 echo "  with coverage           = $enable_coverage"
+echo "  with examples           = $use_examples"
 echo "  module ecdh             = $enable_module_ecdh"
 echo "  module recovery         = $enable_module_recovery"
 echo "  module extrakeys        = $enable_module_extrakeys"
--- a/examples/ecdh.c
+++ b/examples/ecdh.c
@@ -1,0 +1,127 @@
+/*************************************************************************
+ * Written in 2020-2022 by Elichai Turkel                                *
+ * To the extent possible under law, the author(s) have dedicated all    *
+ * copyright and related and neighboring rights to the software in this  *
+ * file to the public domain worldwide. This software is distributed     *
+ * without any warranty. For the CC0 Public Domain Dedication, see       *
+ * EXAMPLES_COPYING or https://creativecommons.org/publicdomain/zero/1.0 *
+ *************************************************************************/
+
+#include <stdio.h>
+#include <assert.h>
+#include <string.h>
+
+#include <secp256k1.h>
+#include <secp256k1_ecdh.h>
+
+#include "random.h"
+
+
+int main(void) {
+    unsigned char seckey1[32];
+    unsigned char seckey2[32];
+    unsigned char compressed_pubkey1[33];
+    unsigned char compressed_pubkey2[33];
+    unsigned char shared_secret1[32];
+    unsigned char shared_secret2[32];
+    unsigned char randomize[32];
+    int return_val;
+    size_t len;
+    secp256k1_pubkey pubkey1;
+    secp256k1_pubkey pubkey2;
+
+    /* The specification in secp256k1.h states that `secp256k1_ec_pubkey_create`
+     * needs a context object initialized for signing, which is why we create
+     * a context with the SECP256K1_CONTEXT_SIGN flag.
+     * (The docs for `secp256k1_ecdh` don't require any special context, just
+     * some initialized context) */
+    secp256k1_context* ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN);
+    if (!fill_random(randomize, sizeof(randomize))) {
+        printf("Failed to generate randomness\n");
+        return 1;
+    }
+    /* Randomizing the context is recommended to protect against side-channel
+     * leakage See `secp256k1_context_randomize` in secp256k1.h for more
+     * information about it. This should never fail. */
+    return_val = secp256k1_context_randomize(ctx, randomize);
+    assert(return_val);
+
+    /*** Key Generation ***/
+
+    /* If the secret key is zero or out of range (bigger than secp256k1's
+     * order), we try to sample a new key. Note that the probability of this
+     * happening is negligible. */
+    while (1) {
+        if (!fill_random(seckey1, sizeof(seckey1)) || !fill_random(seckey2, sizeof(seckey2))) {
+            printf("Failed to generate randomness\n");
+            return 1;
+        }
+        if (secp256k1_ec_seckey_verify(ctx, seckey1) && secp256k1_ec_seckey_verify(ctx, seckey2)) {
+            break;
+        }
+    }
+
+    /* Public key creation using a valid context with a verified secret key should never fail */
+    return_val = secp256k1_ec_pubkey_create(ctx, &pubkey1, seckey1);
+    assert(return_val);
+    return_val = secp256k1_ec_pubkey_create(ctx, &pubkey2, seckey2);
+    assert(return_val);
+
+    /* Serialize pubkey1 in a compressed form (33 bytes), should always return 1 */
+    len = sizeof(compressed_pubkey1);
+    return_val = secp256k1_ec_pubkey_serialize(ctx, compressed_pubkey1, &len, &pubkey1, SECP256K1_EC_COMPRESSED);
+    assert(return_val);
+    /* Should be the same size as the size of the output, because we passed a 33 byte array. */
+    assert(len == sizeof(compressed_pubkey1));
+
+    /* Serialize pubkey2 in a compressed form (33 bytes) */
+    len = sizeof(compressed_pubkey2);
+    return_val = secp256k1_ec_pubkey_serialize(ctx, compressed_pubkey2, &len, &pubkey2, SECP256K1_EC_COMPRESSED);
+    assert(return_val);
+    /* Should be the same size as the size of the output, because we passed a 33 byte array. */
+    assert(len == sizeof(compressed_pubkey2));
+
+    /*** Creating the shared secret ***/
+
+    /* Perform ECDH with seckey1 and pubkey2. Should never fail with a verified
+     * seckey and valid pubkey */
+    return_val = secp256k1_ecdh(ctx, shared_secret1, &pubkey2, seckey1, NULL, NULL);
+    assert(return_val);
+
+    /* Perform ECDH with seckey2 and pubkey1. Should never fail with a verified
+     * seckey and valid pubkey */
+    return_val = secp256k1_ecdh(ctx, shared_secret2, &pubkey1, seckey2, NULL, NULL);
+    assert(return_val);
+
+    /* Both parties should end up with the same shared secret */
+    return_val = memcmp(shared_secret1, shared_secret2, sizeof(shared_secret1));
+    assert(return_val == 0);
+
+    printf("Secret Key1: ");
+    print_hex(seckey1, sizeof(seckey1));
+    printf("Compressed Pubkey1: ");
+    print_hex(compressed_pubkey1, sizeof(compressed_pubkey1));
+    printf("\nSecret Key2: ");
+    print_hex(seckey2, sizeof(seckey2));
+    printf("Compressed Pubkey2: ");
+    print_hex(compressed_pubkey2, sizeof(compressed_pubkey2));
+    printf("\nShared Secret: ");
+    print_hex(shared_secret1, sizeof(shared_secret1));
+
+    /* This will clear everything from the context and free the memory */
+    secp256k1_context_destroy(ctx);
+
+    /* It's best practice to try to clear secrets from memory after using them.
+     * This is done because some bugs can allow an attacker to leak memory, for
+     * example through "out of bounds" array access (see Heartbleed), Or the OS
+     * swapping them to disk. Hence, we overwrite the secret key buffer with zeros.
+     *
+     * TODO: Prevent these writes from being optimized out, as any good compiler
+     * will remove any writes that aren't used. */
+    memset(seckey1, 0, sizeof(seckey1));
+    memset(seckey2, 0, sizeof(seckey2));
+    memset(shared_secret1, 0, sizeof(shared_secret1));
+    memset(shared_secret2, 0, sizeof(shared_secret2));
+
+    return 0;
+}
--- a/examples/ecdsa.c
+++ b/examples/ecdsa.c
@@ -1,0 +1,137 @@
+/*************************************************************************
+ * Written in 2020-2022 by Elichai Turkel                                *
+ * To the extent possible under law, the author(s) have dedicated all    *
+ * copyright and related and neighboring rights to the software in this  *
+ * file to the public domain worldwide. This software is distributed     *
+ * without any warranty. For the CC0 Public Domain Dedication, see       *
+ * EXAMPLES_COPYING or https://creativecommons.org/publicdomain/zero/1.0 *
+ *************************************************************************/
+
+#include <stdio.h>
+#include <assert.h>
+#include <string.h>
+
+#include <secp256k1.h>
+
+#include "random.h"
+
+
+
+int main(void) {
+    /* Instead of signing the message directly, we must sign a 32-byte hash.
+     * Here the message is "Hello, world!" and the hash function was SHA-256.
+     * An actual implementation should just call SHA-256, but this example
+     * hardcodes the output to avoid depending on an additional library.
+     * See https://bitcoin.stackexchange.com/questions/81115/if-someone-wanted-to-pretend-to-be-satoshi-by-posting-a-fake-signature-to-defrau/81116#81116 */
+    unsigned char msg_hash[32] = {
+        0x31, 0x5F, 0x5B, 0xDB, 0x76, 0xD0, 0x78, 0xC4,
+        0x3B, 0x8A, 0xC0, 0x06, 0x4E, 0x4A, 0x01, 0x64,
+        0x61, 0x2B, 0x1F, 0xCE, 0x77, 0xC8, 0x69, 0x34,
+        0x5B, 0xFC, 0x94, 0xC7, 0x58, 0x94, 0xED, 0xD3,
+    };
+    unsigned char seckey[32];
+    unsigned char randomize[32];
+    unsigned char compressed_pubkey[33];
+    unsigned char serialized_signature[64];
+    size_t len;
+    int is_signature_valid;
+    int return_val;
+    secp256k1_pubkey pubkey;
+    secp256k1_ecdsa_signature sig;
+    /* The specification in secp256k1.h states that `secp256k1_ec_pubkey_create` needs
+     * a context object initialized for signing and `secp256k1_ecdsa_verify` needs
+     * a context initialized for verification, which is why we create a context
+     * for both signing and verification with the SECP256K1_CONTEXT_SIGN and
+     * SECP256K1_CONTEXT_VERIFY flags. */
+    secp256k1_context* ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN | SECP256K1_CONTEXT_VERIFY);
+    if (!fill_random(randomize, sizeof(randomize))) {
+        printf("Failed to generate randomness\n");
+        return 1;
+    }
+    /* Randomizing the context is recommended to protect against side-channel
+     * leakage See `secp256k1_context_randomize` in secp256k1.h for more
+     * information about it. This should never fail. */
+    return_val = secp256k1_context_randomize(ctx, randomize);
+    assert(return_val);
+
+    /*** Key Generation ***/
+
+    /* If the secret key is zero or out of range (bigger than secp256k1's
+     * order), we try to sample a new key. Note that the probability of this
+     * happening is negligible. */
+    while (1) {
+        if (!fill_random(seckey, sizeof(seckey))) {
+            printf("Failed to generate randomness\n");
+            return 1;
+        }
+        if (secp256k1_ec_seckey_verify(ctx, seckey)) {
+            break;
+        }
+    }
+
+    /* Public key creation using a valid context with a verified secret key should never fail */
+    return_val = secp256k1_ec_pubkey_create(ctx, &pubkey, seckey);
+    assert(return_val);
+
+    /* Serialize the pubkey in a compressed form(33 bytes). Should always return 1. */
+    len = sizeof(compressed_pubkey);
+    return_val = secp256k1_ec_pubkey_serialize(ctx, compressed_pubkey, &len, &pubkey, SECP256K1_EC_COMPRESSED);
+    assert(return_val);
+    /* Should be the same size as the size of the output, because we passed a 33 byte array. */
+    assert(len == sizeof(compressed_pubkey));
+
+    /*** Signing ***/
+
+    /* Generate an ECDSA signature `noncefp` and `ndata` allows you to pass a
+     * custom nonce function, passing `NULL` will use the RFC-6979 safe default.
+     * Signing with a valid context, verified secret key
+     * and the default nonce function should never fail. */
+    return_val = secp256k1_ecdsa_sign(ctx, &sig, msg_hash, seckey, NULL, NULL);
+    assert(return_val);
+
+    /* Serialize the signature in a compact form. Should always return 1
+     * according to the documentation in secp256k1.h. */
+    return_val = secp256k1_ecdsa_signature_serialize_compact(ctx, serialized_signature, &sig);
+    assert(return_val);
+
+
+    /*** Verification ***/
+
+    /* Deserialize the signature. This will return 0 if the signature can't be parsed correctly. */
+    if (!secp256k1_ecdsa_signature_parse_compact(ctx, &sig, serialized_signature)) {
+        printf("Failed parsing the signature\n");
+        return 1;
+    }
+
+    /* Deserialize the public key. This will return 0 if the public key can't be parsed correctly. */
+    if (!secp256k1_ec_pubkey_parse(ctx, &pubkey, compressed_pubkey, sizeof(compressed_pubkey))) {
+        printf("Failed parsing the public key\n");
+        return 1;
+    }
+
+    /* Verify a signature. This will return 1 if it's valid and 0 if it's not. */
+    is_signature_valid = secp256k1_ecdsa_verify(ctx, &sig, msg_hash, &pubkey);
+
+    printf("Is the signature valid? %s\n", is_signature_valid ? "true" : "false");
+    printf("Secret Key: ");
+    print_hex(seckey, sizeof(seckey));
+    printf("Public Key: ");
+    print_hex(compressed_pubkey, sizeof(compressed_pubkey));
+    printf("Signature: ");
+    print_hex(serialized_signature, sizeof(serialized_signature));
+
+
+    /* This will clear everything from the context and free the memory */
+    secp256k1_context_destroy(ctx);
+
+    /* It's best practice to try to clear secrets from memory after using them.
+     * This is done because some bugs can allow an attacker to leak memory, for
+     * example through "out of bounds" array access (see Heartbleed), Or the OS
+     * swapping them to disk. Hence, we overwrite the secret key buffer with zeros.
+     *
+     * TODO: Prevent these writes from being optimized out, as any good compiler
+     * will remove any writes that aren't used. */
+    memset(seckey, 0, sizeof(seckey));
+
+    return 0;
+}
--- a/examples/random.h
+++ b/examples/random.h
@@ -1,0 +1,73 @@
+/*************************************************************************
+ * Copyright (c) 2020-2021 Elichai Turkel                                *
+ * Distributed under the CC0 software license, see the accompanying file *
+ * EXAMPLES_COPYING or https://creativecommons.org/publicdomain/zero/1.0 *
+ *************************************************************************/
+
+/*
+ * This file is an attempt at collecting best practice methods for obtaining randomness with different operating systems.
+ * It may be out-of-date. Consult the documentation of the operating system before considering to use the methods below.
+ *
+ * Platform randomness sources:
+ * Linux   -> `getrandom(2)`(`sys/random.h`), if not available `/dev/urandom` should be used. http://man7.org/linux/man-pages/man2/getrandom.2.html, https://linux.die.net/man/4/urandom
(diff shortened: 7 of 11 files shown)


examples/random.h
```diff
@@ -0,0 +1,31 @@
...
+/* Returns 1 on sucess, and 0 on failure. */
+int fill_random(unsigned char* data, unsigned long size) {
+#if defined(_WIN32)
+    NTSTATUS res = BCryptGenRandom(NULL, data, size, BCRYPT_USE_SYSTEM_PREFERRED_RNG);
+    if (res != STATUS_SUCCESS) {
+        return 0;
+    }
+#elif defined(__linux__)
+    ssize_t res;
+    int fd = open("/dev/urandom", O_RDONLY);
```
jonasnick, line 20: Would be better if we'd use getrandom if available.

examples/random.h
```diff
@@ -0,0 +1,31 @@
+#if defined(_WIN32)
+#include <bcrypt.h>
+#elif defined(__linux__)
+#include <fcntl.h>
+#include <unistd.h>
+#else
+#error "Couldn't identify the OS"
```
jonasnick, line 7: Seems like this would happen on many common OSes (macOS, BSDs, ...). How about not compiling the examples by default?

elichai:
> Alternatively the module-specific examples could be in their respective module directory. But I think either way is fine.

I thought about it, but I think that for newcomers it's easier to see an examples directory at the top level and start from there.

examples/random.h
```diff
@@ -0,0 +1,31 @@
...
+    if (fd == -1 ) {
+        return 0;
+    }
+    res = read(fd, data, size);
+    close(fd);
+    if (res != (ssize_t)size) {
+        return 0;
+    }
+#endif
+    return 1;
```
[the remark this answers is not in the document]
elichai: That should never happen because there's an `#error` at the top of the file if none were taken.

examples/ecdh.c
```diff
@@ -0,0 +1,77 @@
...
+    unsigned char seckey1[32];
+    unsigned char seckey2[32];
+    unsigned char compressed_pubkey1[33];
+    unsigned char compressed_pubkey2[33];
+    unsigned char shared_secret1[32];
+    unsigned char shared_secret2[32];
+    size_t len;
+    secp256k1_pubkey pubkey1;
+    secp256k1_pubkey pubkey2;
+    secp256k1_context* ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN);
```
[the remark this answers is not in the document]
elichai: Great Idea! I'll try to phrase something :)

examples/random.h
elichai: I really wanted to use `getrandom` but it's not available in the libc in our travis.
We could add `dist: bionic` and that will probably solve this, but my thought was 
"if even our travis doesn't support it then we should probably not use this in the examples"

See: https://travis-ci.org/github/elichai/secp256k1/jobs/681424984#L403

examples/random.h
elichai: I skipped BSD because I didn't think anyone is really using it, but forgot about OSX.
I'll look into adding them both, see how compilcated it gets (I want to avoid something like https://github.com/bitcoin/bitcoin/blob/master/src/random.cpp and only support mostly recent OS versions)

examples/random.h
jonasnick: Good point. We can update this later.

examples/random.h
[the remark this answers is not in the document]
jonasnick: Yeah but if you add a branch above to add another OS, you may forget to add it here here as well (or accidentally branch on something else).

examples/ecdh.c
```diff
@@ -0,0 +1,77 @@
...
+    printf("Compressed Pubkey1: ");
+    print_hex(compressed_pubkey1, sizeof(compressed_pubkey1));
+    printf("\nSecret Key2: ");
+    print_hex(seckey2, sizeof(seckey2));
+    printf("Compressed Pubkey2: ");
+    print_hex(compressed_pubkey2, sizeof(compressed_pubkey2));
+    printf("\nShared Secret: ");
+    print_hex(shared_secret1, sizeof(shared_secret1));
+
+
```
jonasnick, line 75: We need to destroy the context here. Also can try to clear the secret keys (and add comment why we do that).

elichai: Weird that valgrind didn't complain about it.

elichai: Should I do try to do some best effort in that regard? (ie #636)

examples/random.h
real-or-random: Oh weird, it's really not in Ubuntu LTS 16.04. You need glibc 2.25 at least. -.- I wanted to point out `getrandom` too...

I think we should at least add an explanatory comment or a link to an explanation. And to be honest, I'd prefer to use `getrandom` here and have an example that breaks compilation if it's not available and refer people to `/dev/urandom` then.

elichai: I thought about that, but how do you propose to do this?
like should I involve autotools in this now?(to detect getrandom) I really preferred to keep the examples minimal and self standing.

another advantage of `/dev/urandom` is that I can just stick MacOS and BSD variants on it and it should work

real-or-random: > like should I involve autotools in this now?(to detect getrandom) I really preferred to keep the examples minimal and self standing.

Hm ok, what I had in mind was rather not compiling the examples at all in our build system. 
But I see that this is not great either, too.

examples/random.h
real-or-random: Okay, I hadn't read this conversation here before. Maybe it's really better not to compile the examples in our build system. If the concern is CI, we can compile them in the CI. If we want, we could add a super simple standalone Makefile that's entirely decoupled from our autoconf stuff.

elichai: I'm trying to redo the travis and at least add MacOS to the CI
hmm stand alone make can be interesting, should that compile libsecp by it self? 
I can write something simple.

Otherwise we can still compile it in our CI, turn it off by default and bump our travis to bionic

examples/ecdh.c
real-or-random: Shall I export a function to clear memory in #636?

real-or-random:
My feeling is that I think this entire issue of secret key handling deserves a broader discussion.

examples/keygen.c
```diff
@@ -0,0 +1,56 @@
...
+}
+
+int main(void) {
+    unsigned char seckey[32];
+    unsigned char compressed_pubkey[33];
+    unsigned char uncompressed_pubkey[65];
+    size_t len;
+    secp256k1_pubkey pubkey;
+    secp256k1_context* ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN);
+    while (1) {
```
real-or-random, line 24: I'd prefer not to loop here and instead bail out if we hit an invalid key.

 The probability of hitting an invalid key is negligible if your randomness is good.
So this loop is only relevant if your randomness is broken. And then it hides the fact that your randomness is broken.

elichai: FWIW:
https://github.com/bitcoin/bitcoin/blob/master/src/key.cpp#L157
https://github.com/rust-bitcoin/rust-secp256k1/blob/master/src/key.rs#L118
https://github.com/bitcoin-core/secp256k1/blob/f39f99be0e6add959f534c03b93044cef066fe09/src/tests.c#L130:L141

examples/ecdh.c
elichai: I think it's an improvement, although probably most serious users will have their own mechanism.

examples/random.h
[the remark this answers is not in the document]
elichai: Done

examples/ecdh.c
```diff
@@ -0,0 +1,77 @@
...
+    unsigned char shared_secret1[32];
+    unsigned char shared_secret2[32];
+    size_t len;
+    secp256k1_pubkey pubkey1;
+    secp256k1_pubkey pubkey2;
+    secp256k1_context* ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN);
+    while (1) {
+        if (!fill_random(seckey1, sizeof(seckey1)) || !fill_random(seckey2, sizeof(seckey2))) {
+            printf("Failed to generate randomness\n");
+                return 1;
```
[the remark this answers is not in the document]
elichai: Thanks. Done

examples/ecdh.c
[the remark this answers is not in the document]
elichai: Is it better?

examples/random.h
elichai: Added macOS and BSD, and will add macOS to the CI soon so we could actually test this.
rn they all use /dev/urandom, but if we bump to bionic I can replace with getrandom/getentropy

examples/ecdh.c
elichai: Done, I added memset with a TODO, so we could either copy your function to the example, or if we actually export it use that.

examples/random.h
jonasnick: I think the advantage of covering anything other than Linux for now is small compared to the maintenance overhead and bikeshedding. I'd suggest to do Linux only, use /dev/urandom, add comment to use `getrandom` if available, not compile by default, but add a travis test. Easy to revisit this later if there is demand for other systems but most important is the code, and not that it compiles on every user's system.

examples/ecdh.c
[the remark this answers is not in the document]
jonasnick: Yeah, I think it's helpful (if only for making sure that people pay attention to the flags).

examples/random.h
real-or-random: Oh I didn't see your comment here. I suggested Linux/MacOS/Windows here:
https://github.com/bitcoin-core/secp256k1/issues/749#issuecomment-623964688 But I'd also be okay with just Linux. I think the main point is that we good documentation.

And I have a somewhat strong opinion on `/dev/urandom` vs `getrandom`. The issue with `/dev/urandom` is that it never blocks, i.e., it doesn't block at early boot when there has never been enough entropy. That's rare for Linux on the desktop but may be the case for a lightweight Linux on some device. So I'd really prefer `getrandom` and a comment to use `/dev/urandom` if `getrandom` is not available. If we don't compile the examples, then it does not matter if it's not available.

As I said, I'll try to find some refs. If you want, go ahead with this PR and I'll open a PR on top of it.

examples/keygen.c
real-or-random: Hm, I see. I still think that my argument holds up. Maybe @sipa can chime in for the code in Core.

examples/random.h
jonasnick: Since we're updating travis to bionic in #750 (which I support), what you're suggesting sounds good to me (`getrandom` or bust) and test that in travis.

elichai: Done :)

elichai: Fixed, And added explanations about the different random sources for each platform

examples/keygen.c
NicholasFarrow: While the probability of an invalid key is negligible (something like 1/2^128?),  I feel that if the verify function is provided in the library then it could still be informative to include in the examples.

```
if (!secp256k1_ec_seckey_verify(ctx, seckey)) {
	printf("Invalid secret key\n");
	return 1;
}
```

examples/ecdh.c
```diff
@@ -0,0 +1,101 @@
...
+    unsigned char shared_secret2[32];
+    size_t len;
+    secp256k1_pubkey pubkey1;
+    secp256k1_pubkey pubkey2;
+    /* The docs in secp256k1.h above the `secp256k1_ec_pubkey_create` function say:
+     * "pointer to a context object, initialized for signing"
+     * Which is why we create a context for signing(SECP256K1_CONTEXT_SIGN).
+     * (The docs for `secp256k1_ecdh` don't require any special context, just some context) */
+
+    secp256k1_context* ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN);
```
real-or-random, line 40: The context should be randomized here.

elichai: Done

jonasnick:
Would be cool to revive this PR since it's really helpful, well-documented and provides best practices for getting randomness on different platforms. In particular, I don't think this PR should be blocked on the result of the key generation discussion. @elichai if you want to add examples for the schnorrsig module, feel free to do it, but we can also add this in a different PR.

elichai:
Rebased, and replaced the `keygen` example with a schnorr and ecdsa examples.
Please review and comment if you think something I've done isn't best practice or some comment's wording isn't good enough

examples/random.h
```diff
@@ -0,0 +1,54 @@
...
+    NTSTATUS res = BCryptGenRandom(NULL, data, size, BCRYPT_USE_SYSTEM_PREFERRED_RNG);
+    if (res == STATUS_SUCCESS) {
+        return 1;
+    } else {
+        return 0;
+    }
+#elif defined(__linux__) || defined(__FreeBSD__)
+    /* If `getrandom(2)` is not available you should fallback to /dev/urandom */
+    ssize_t res = getrandom(data, size, 0);
+    if (res == (ssize_t)size) {
```
jonasnick, line 39: `unsigned long` may not fit in an `ssize_t`. On 32 bit systems it can therefore happen that getrandom returns -1 but `fill_random` returns `1`. Perhaps change the `size` argument of `fill_random` to `size_t`, add `res != -1` to this `if` condition and, for WIN32, check whether size exceeds `unsigned long`?

Makefile.am
```diff
@@ -123,6 +124,31 @@ exhaustive_tests_LDFLAGS = -static
...
 endif
 
+if USE_EXAMPLES
+noinst_PROGRAMS += ecdsa_example
+ecdsa_example_SOURCES = examples/ecdsa.c
+ecdsa_example_CPPFLAGS = -I$(top_srcdir)/include $(SECP_INCLUDES)
+ecdsa_example_LDADD = libsecp256k1.la $(SECP_LIBS)
+ecdsa_example_LDFLAGS = -static
+TESTS += ecdsa_example
+endif
```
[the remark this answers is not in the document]
elichai: Nope, accidentally moved it when added schnorrsig

examples/random.h
real-or-random: Nice catch. I hate C integers.

real-or-random:
> P.S. Should the usage examples be MIT or CC0(Public Domain)?

I think we should make them CC0 to avoid any doubt.

examples/random.h
```diff
@@ -0,0 +1,54 @@
...
+#include <unistd.h>
+#else
+#error "Couldn't identify the OS"
+#endif
+
+
+/* Returns 1 on success, and 0 on failure. */
+static int fill_random(unsigned char* data, unsigned long size) {
+#if defined(_WIN32)
+    NTSTATUS res = BCryptGenRandom(NULL, data, size, BCRYPT_USE_SYSTEM_PREFERRED_RNG);
```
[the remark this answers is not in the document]
elichai: AFAIU this means use the default CPRNG, otherwise we can't pass in NULL as the `hAlgorithm` and we need to instead need to pass a specific algorithm

examples/ecdsa.c
```diff
@@ -0,0 +1,124 @@
...
+
+#include <stdio.h>
+#include <assert.h>
+#include <string.h>
+
+#include "random.h"
+#include "secp256k1.h"
+
+
+static void print_hex(unsigned char* data, size_t size) {
```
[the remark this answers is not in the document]
elichai: Should I move it to `random.h`? should I rename `random.h` to `util.h`?

examples/random.h
elichai: Argh,how do I check if `size` exceeds `ULONG_MAX` if the standard doesn't even tell me what's bigger `size_t` or `unsigned long`?.

What's the "right" way to do this here? :(

elichai:
Rebased and fixed review comments,
Feel free to suggest alternative phrasings to the comments I wrote (English is my 2nd language)

Makefile.am
```diff
@@ -123,6 +124,31 @@ exhaustive_tests_LDFLAGS = -static
 TESTS += exhaustive_tests
 endif
 
+if USE_EXAMPLES
+noinst_PROGRAMS += ecdsa_example
+ecdsa_example_SOURCES = examples/ecdsa.c
+ecdsa_example_CPPFLAGS = -I$(top_srcdir)/include $(SECP_INCLUDES)
+ecdsa_example_LDADD = libsecp256k1.la $(SECP_LIBS)
```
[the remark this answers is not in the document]
elichai: I removed `$(SECP_INCLUDES)` and `$(SECP_LIBS)`, anything else I should change in the automake?

examples/ecdh.c
```diff
@@ -0,0 +1,111 @@
...
+    if (!fill_random(randomize, sizeof(randomize))) {
+        printf("Failed to generate randomness\n");
+        return 1;
+    }
+    /* Randomizing the context is recommended to protect against side-channel leakage
+     * See `secp256k1_context_randomize` in secp256k1.h for more information about it
+     * Should never fail */
+    assert(secp256k1_context_randomize(ctx, randomize));
+
+    while (1) {
```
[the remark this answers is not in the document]
elichai: Added

examples/schnorr.c
```diff
@@ -0,0 +1,123 @@
...
+        printf("%02x", data[i]);
+    }
+    printf("\n");
+}
+
+int main(void) {
+    unsigned char msg_hash[32] = {0}; /* This should be a hash of the message. */
+    unsigned char seckey[32];
+    unsigned char randomize[32];
+    unsigned char auxilary_rand[32];
```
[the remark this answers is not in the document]
elichai: Fixed, Thanks!

examples/schnorr.c
```diff
@@ -0,0 +1,123 @@
...
+    /*** Key Generation ***/
+
+    /* If the secret key is zero or out of range (bigger than secp256k1 order), the probablity of this happening is negligble
+     * non the less we try generating a new key if failed  */
+    while (1) {
+        if (!fill_random(seckey, sizeof(seckey))) {
+            printf("Failed to generate randomness\n");
+            return 1;
+        }
+        /* Try to create a keypair wit a valid context, it should only fail if the seckey is zero or out of range. */
```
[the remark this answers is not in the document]
elichai: Fixed, Thanks!

examples/schnorr.c
```diff
@@ -0,0 +1,123 @@
...
+        if (secp256k1_keypair_create(ctx, &keypair, seckey)) {
+            break;
+        }
+    }
+
+    /* Extract the XOnly Public Key from the Keypair,
```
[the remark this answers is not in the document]
elichai: Fixed, Thanks!

examples/schnorr.c
```diff
@@ -0,0 +1,123 @@
...
+        return 1;
+    }
+    /* Randomizing the context is recommended to protect against side-channel leakage
+     * See `secp256k1_context_randomize` in secp256k1.h for more information about it
+     * Should never fail */
+    assert(secp256k1_context_randomize(ctx, randomize));
+
+    /*** Key Generation ***/
+
+    /* If the secret key is zero or out of range (bigger than secp256k1 order), the probablity of this happening is negligble
```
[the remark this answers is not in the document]
elichai: Rephrased, is it better now?

examples/schnorr.c
```diff
@@ -0,0 +1,123 @@
...
+     * non the less we try generating a new key if failed  */
```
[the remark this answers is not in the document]
elichai: I just removed that word :)

.gitignore
```diff
@@ -10,10 +10,14 @@ tests
 exhaustive_tests
 gen_context
 valgrind_ctime_test
+keygen_example
```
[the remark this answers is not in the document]
elichai: Thanks! Fixed

README.md
```diff
@@ -68,6 +68,13 @@ libsecp256k1 is built using autotools:
     $ make check
     $ sudo make install  # optional
 
+Usage Examples
+-----------
+  Usage Examples can be found in the [examples](examples) directory:
```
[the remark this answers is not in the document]
elichai: Added, Thanks :)

examples/ecdsa.c
```diff
@@ -0,0 +1,124 @@
+/**********************************************************************
+ * Copyright (c) 2020 Elichai Turkel	                              *
```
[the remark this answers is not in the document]
elichai: I don't see any trailing whitespaces now, is this still an issue? do you mean the way the copyright header is?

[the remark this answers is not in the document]
elichai: I don't see any trailing whitespaces now, is this still an issue? do you mean the way the copyright header is?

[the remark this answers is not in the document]
jonasnick: I don't see them either now.

examples/random.h
```diff
@@ -0,0 +1,54 @@
...
+ * Distributed under the MIT software license, see the accompanying   *
+ * file COPYING or http://www.opensource.org/licenses/mit-license.php.*
+ **********************************************************************/
+/*
+ * In this file only the best practice randomness sources are used
+ * Platform randomness sources:
+ * Linux   -> `getrandom(2)`(`sys/random.h`), if not available `/dev/urandom` should be used. http://man7.org/linux/man-pages/man2/getrandom.2.html, https://linux.die.net/man/4/urandom
+ * macOS   -> `getentropy(2)`(`sys/random.h`), if not available `/dev/urandom` should be used. https://www.unix.com/man-page/mojave/2/getentropy, https://opensource.apple.com/source/xnu/xnu-517.12.7/bsd/man/man4/random.4.auto.html
+ * FreeBSD -> `getrandom(2)`(`sys/random.h`), if not available `kern.arandom` should be used. https://www.freebsd.org/cgi/man.cgi?query=getrandom, https://www.freebsd.org/cgi/man.cgi?query=random&sektion=4
+ * OpenBSD -> `getentropy(2)`(`unistd.h`), if not available `/dev/urandom` should be used. https://man.openbsd.org/getentropy, https://man.openbsd.org/urandom
```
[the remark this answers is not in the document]
jonasnick: > I think the reason was that some platforms still use RC4 for that syscall:

  Urgh. But we're only targeting `APPLE` and `OpenBSD` here, whose manpages both recommend arc4random. But agree that `getentropy` should be ok too.

  How about we replace the sentence "In this file only the best practice randomness sources are used" at the top of the file with:
  "This file is an attempt at collecting best practice methods for obtaining randomness with different operating systems. It may be out-of-date. Consult the documentation of the operating system before considering to use the methods below."

examples/random.h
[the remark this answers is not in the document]
jonasnick: I see, thx

examples/schnorr.c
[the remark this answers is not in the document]
jonasnick: The sentence is better now. I pushed a commit that improves the punctuation (and layout) of the doc comments here https://github.com/jonasnick/secp256k1/commits/2020-07-docs-jn. Feel free to cherry-pick (and squash) if you agree.

examples/ecdsa.c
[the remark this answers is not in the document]
jonasnick: sgtm

Makefile.am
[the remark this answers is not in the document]
jonasnick: I think you can remove `ecdsa_example_CPPFLAGS` and `ecdh_example_CPPFLAGS`.

examples/random.h
jonasnick: What's wrong with `if (size > ULONG_MAX) return 0;`? Also, instead of checking `res != 1`, I think it would be better to check that `size <= SSIZE_MAX`.

elichai: I was worried about what if `size_t` is smaller than `unsigned long`, the constant will overflow and the condition won't be the one it seems to be.
But I re-read the integer promotion rules and I think it checks out, if I don't explicitly convert them to a specific type then they'll be converted to the integer with the higher rank, preserving the right values.

Makefile.am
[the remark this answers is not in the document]
elichai: I can't as they're importing `#include "secp256k1.h"`, I could remove it if I change it to `#include "include/secp256k1.h"` But I think actual users will pass `-Isecp256k1/secp256k1.h`

examples/random.h
real-or-random: Yep, sounds correct what you say about the promotion rules.

examples/random.h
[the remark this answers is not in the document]
elichai: That sounds good. I hate that OSes are so bad that no one can actually tell what it actually the best practice

examples/schnorr.c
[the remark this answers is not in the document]
elichai: Thanks! cherry-picked

elichai:
Rebasing on master again so the CI will work

examples/random.h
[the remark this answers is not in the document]
jonasnick: nit: better formatting
```
 * This file is an attempt at collecting best practice methods for obtaining
 * randomness with different operating systems. It may be out-of-date. Consult
 * the documentation of the operating system before considering to use the
 * methods below.
 *
 * Platform randomness sources:
 ```

File: examples/random.h [hunk 24]
```diff
@@ -0,0 +1,70 @@
...
+    NTSTATUS res = BCryptGenRandom(NULL, data, size, BCRYPT_USE_SYSTEM_PREFERRED_RNG);
+    if (res != STATUS_SUCCESS || size > ULONG_MAX) {
+        return 0;
+    } else {
+        return 1;
+    }
+#elif defined(__linux__) || defined(__FreeBSD__)
+    /* If `getrandom(2)` is not available you should fallback to /dev/urandom */
+    ssize_t res = getrandom(data, size, 0);
+    if (res != (ssize_t)size || size > SSIZE_MAX) {
```
jonasnick, line 45: Better to first check SSIZE, otherwise the function can return 1 even though not enough random bytes have been read.

Makefile.am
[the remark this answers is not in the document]
jonasnick: hm okay

examples/random.h
elichai: I don't mind checking it first, but how can it return 1 even though not enough random bytes have been read?
if `size` is bigger than `SSIZE_MAX` or `size` isn't the same as `res` this will return 0.

jonasnick: right, sorry, nevermind

elichai: So to access `SSIZE_MAX` I'll need to define some weird POSIX macros, I'll instead use `if (res < 0 || (size_t)res != size )` which should work assuming `ssize_t = signed size_t` which honestly I'm not 100% sure I can assume. sigh.
I could also cast to `unsigned long` as both `man size_t` and `man ssize_t` say that they're no greater than the width of `long`

elichai:
Added a commit fixing some mingw linking bugs:
1. Need to include `windows.h` and `ntstatus.h` even though they're used in `bcrypt.h` (https://sourceforge.net/p/mingw-w64/bugs/903/)
2. Had to explicitly link to `libbcrypt`

examples/schnorr.c
```diff
@@ -0,0 +1,119 @@
...
+    unsigned char msg_hash[32] = {0}; /* This should be a hash of the message. */
+    unsigned char seckey[32];
+    unsigned char randomize[32];
+    unsigned char auxiliary_rand[32];
+    unsigned char serialized_pubkey[32];
+    unsigned char signature[64];
+    int is_signature_valid;
+    secp256k1_xonly_pubkey pubkey;
+    secp256k1_keypair keypair;
+    /* The docs in secp256k1_extrakeys.h above the `secp256k1_keypair_create`
```
sipa, line 27: That's remarkably specific, and may go outdated if the header files are restructured a bit.

What about "The specification in secp256k1_extrakeys.h states that secp256k1_schnorrsig_verify needs ..."?

examples/schnorr.c
```diff
@@ -0,0 +1,119 @@
...
+        if (secp256k1_keypair_create(ctx, &keypair, seckey)) {
+            break;
+        }
+    }
+
+    /* Extract the X-only public key from the keypair. We pass NULL for
+     * `pk_parity` as we don't care about the parity of the key, only advanced
+     * users might care about the parity. This should never fail with a valid
+     * context and public key. */
+    assert(secp256k1_keypair_xonly_pub(ctx, &pubkey, NULL, &keypair));
```
sipa, line 64: assert with side effects

examples/schnorr.c
```diff
@@ -0,0 +1,119 @@
...
+
+    /* Serialize the public key. Should always return 1 for a valid public key. */
+    assert(secp256k1_xonly_pubkey_serialize(ctx, serialized_pubkey, &pubkey));
```
sipa, line 67: assert with side effects

examples/schnorr.c
```diff
@@ -0,0 +1,119 @@
...
+        return 1;
+    }
+
+    /* Generate a Schnorr signature `noncefp` and `ndata` allows you to pass a
+    * custom nonce function, passing `NULL` will use the BIP-340 safe default.
+    * BIP-340 recommends passing 32 bytes of randomness to the nonce function to
+    * improve security against side-channel attacks. Signing with a valid
+    * context, verified keypair and the default nonce function should never
+    * fail. */
+    assert(secp256k1_schnorrsig_sign(ctx, signature, msg_hash, &keypair, auxiliary_rand));
```
sipa, line 83: assert with side effects

examples/schnorr.c
```diff
@@ -0,0 +1,119 @@
...
+    printf("Public Key: ");
+    print_hex(serialized_pubkey, sizeof(serialized_pubkey));
+    printf("Signature: ");
+    print_hex(signature, sizeof(signature));
+
+    /* This will clear everything from the context and free the memory */
+    secp256k1_context_destroy(ctx);
+
+    /* It's best practice to try to remove secrets from memory after using them.
+     * This is done because some bugs can allow an attacker leak memory, for
```
sipa, line 110: Grammar seems off here.

Also: an attacker accessing memory through an exploit isn't the only reason; another one is the risk that the OS swaps out the application to disk, where it may end up on permanent storage (which might be less well-protected from attackers). Secrets in general should not be kept in memory longer than necessary.

Here and elsewhere.

examples/random.h
jonasnick: Looks good to me.

configure.ac
```diff
@@ -68,6 +70,10 @@ case $host_os in
        fi
      fi
    ;;
+   cygwin*|mingw*)
+
```
jonasnick, line 74: nit: unnecessary newline?

examples/schnorr.c
```diff
@@ -0,0 +1,119 @@
...
+     * SECP256K1_CONTEXT_SIGN and SECP256K1_CONTEXT_VERIFY flags. */
+    secp256k1_context* ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN | SECP256K1_CONTEXT_VERIFY);
+    if (!fill_random(randomize, sizeof(randomize))) {
+        printf("Failed to generate randomness\n");
+        return 1;
+    }
+    /* Randomizing the context is recommended to protect against side-channel
+     * leakage See `secp256k1_context_randomize` in secp256k1.h for more
+     * information about it. This should never fail. */
+    assert(secp256k1_context_randomize(ctx, randomize));
```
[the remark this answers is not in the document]
elichai: Oh you're 100% right

examples/schnorr.c
```diff
@@ -0,0 +1,119 @@
...
+ * Distributed under the CC0 software license, see the accompanying file *
+ * EXAMPLES_COPYING or https://creativecommons.org/publicdomain/zero/1.0 *
+ *************************************************************************/
+
+#include <stdio.h>
+#include <assert.h>
+#include <string.h>
+
+#include "random.h"
+#include "secp256k1.h"
```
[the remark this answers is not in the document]
elichai: Doesn't that mean that they've installed the library on the user's system? 
If I understood this correctly then if an app ships a vendored libsecp (like bitcoin core) it technically shouldn't be included via the system path

examples/schnorr.c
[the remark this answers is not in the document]
elichai: Do you think users should even assert the return value? if so then only in debug or also in release?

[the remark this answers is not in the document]
sipa: If you set -I arguments to the compiler correctly, there is no problem with using <...>.

The only difference is that "..." also looks in the current path, and as that's where one would expect the included headers, it's pointless and I would say confusing.

FWIW Bitcoin Core does *everything* using <...> includes, also internal headers.

examples/schnorr.c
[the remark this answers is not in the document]
sipa: If you set -I arguments to the compiler correctly, there is no problem with using <...>.

The only difference is that "..." also looks in the current path, and as that's where one would expect the included headers, it's pointless and I would say confusing.

FWIW Bitcoin Core does *everything* using <...> includes, also internal headers.

examples/schnorr.c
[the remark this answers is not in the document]
sipa: I think for `secp256k1_context_randomize` it's reasonable to change the API documentation that it always returns 1. I don't see under what conditions you'd ever want to return failure here. In that case, it seems unnecessary to check.

There are lots of operations that only returns 0 if you pass in invalid arguments (e.g. serializing a public key to an area that's too small). For those I think it can be recommended practice to check, but I wouldn't consider it incorrect usage not to check if you know your code will always provide enough space.

elichai:
I fixed the comments in a separate commit.
I'll squash them closer to merging as I assume the Review Club will give me a bunch more things to fix.

real-or-random:
We spend a lot of time here coming up with methods to obtain randomness on various OSes. If we're going to maintain this in this example file, could we just go ahead and maintain a real function? (I don't want to suggest a "yes" or "no", I'm really asking this as an open question.) I assume the overhead of maintaining a function is larger but I'm not sure to what extend.

examples/ecdh.c
```diff
@@ -0,0 +1,124 @@
...
+    /* This will clear everything from the context and free the memory */
+    secp256k1_context_destroy(ctx);
+
+    /* It's best practice to try to clear secrets from memory after using them.
+     * This is done because some bugs can allow an attacker to leak memory, for
+     * example through "out of bounds" array access (see Heartbleed), Or the OS
+     * swapping them to disk. Hence, we overwrite the secret key buffer with zeros.
+     *
+     * TODO: Prevent these writes from being optimized out, as any good compiler
+     * will remove any writes that aren't used. */
```
kristapsk, line 117: Check for C11 (`#if (__STDC_VERSION__ >=  201112L)`) could be added and then `memset_s()` used instead.

real-or-random: `memset_s()` is optional in C11, it's in Annex K. You need more than `_STDC_VERSION__ >= 201112L` to check for its presence, see for example https://en.cppreference.com/w/c/string/byte/memset. Also, almost no compiler supports Annex K, so I doubt it's useful in practice, see for example https://stackoverflow.com/a/50724865.

examples/ecdsa.c
```diff
@@ -0,0 +1,121 @@
...
+    int is_signature_valid;
+    secp256k1_pubkey pubkey;
+    secp256k1_ecdsa_signature sig;
+    /* The docs in secp256k1.h above the `secp256k1_ec_pubkey_create` function
+     * say: "pointer to a context object, initialized for signing" And the docs
+     * above the `secp256k1_ecdsa_verify` function say: "a secp256k1 context
+     * object, initialized for verification" which is why we create a context
+     * for both signing and verification with the SECP256K1_CONTEXT_SIGN and
+     * SECP256K1_CONTEXT_VERIFY flags. */
+    secp256k1_context* ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN | SECP256K1_CONTEXT_VERIFY);
```
theStack, line 33: nit: Could assert that the context was created successfully (here and in the two other examples)? I.e. `assert(ctx != NULL);`
IIUC returning NULL can only happen in the unlikely case that allocating dynamic memory fails, but it can't hurt. We also do that check in Bitcoin Core: https://github.com/bitcoin/bitcoin/blob/8f137e69caeb2a2ffe1aa930bd6fbc49cee4087c/src/key.cpp#L394-L395

theStack (commented):
Concept ACK

Looking forward to the review club tomorrow! 🚀

README.md
```diff
@@ -69,6 +69,13 @@ libsecp256k1 is built using autotools:
     $ make check  # run the test suite
     $ sudo make install  # optional
 
+Usage Examples
+-----------
+  Usage Examples can be found in the [examples](examples) directory, to compile them you need configure with `--enable-examples`.
```
robot-dreams, line 74:
```suggestion
  Usage Examples can be found in the [examples](examples) directory. To compile them you need to configure with `--enable-examples`.
  For experimental modules, you will also need `--enable-experimental` as well as a flag for each individual module, e.g. `--enable-module-ecdh`.
```

robot-dreams (commented):
Concept ACK, thanks for adding these helpful examples!

My comments are mostly cosmetic, and I think the PR should be ready to merge after review club.

examples/ecdsa.c
```diff
@@ -0,0 +1,121 @@
...
+#include <assert.h>
+#include <string.h>
+
+#include "random.h"
+#include "secp256k1.h"
+
+
+
+int main(void) {
+    unsigned char msg_hash[32] = {0}; /* This must be a hash of the message. otherwise ECDSA is easily broken. */
```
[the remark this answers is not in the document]
elichai: Can't this make some users think that signing directly on some meaningful string with length 32 is secure?
too bad we don't have access to sha2 here.

jesseposner:
Tested ACK 4c433823a85cac975b0746203d94ce041c10299d

examples/random.h
```diff
@@ -0,0 +1,70 @@
...
+    if (res == 0) {
+        return 1;
+    } else {
+        return 0;
+    }
+#endif
+    return 0;
+}
+
+static void print_hex(unsigned char* data, size_t size) {
```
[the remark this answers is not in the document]
jnewbery: I agree that this doesn't really fit in this file and should go in something like util.h. See also https://github.com/bitcoin-core/secp256k1/pull/748#discussion_r524634217.

examples/ecdsa.c
jonasnick: If that was the case then I would agree that the examples should do the check. However, `context_create` calls `checked_malloc` which will call the "error callback" if the result of malloc is NULL. The default behavior of the error callback is crashing. It could theoretically be overriden by the user to not crash (`context_set_error_callback`) but that would be inadvisable given that the API docs state that after the callback the originally called API function can do anything, in particular its return values are undefined. That's why the API doc for `secp256k1_context_create` says that it returns a "context object" and not a "context object or NULL".

examples/ecdsa.c
```diff
@@ -0,0 +1,127 @@
...
+#include <string.h>
+
+#include <secp256k1.h>
+
+#include "random.h"
+
+
+
+int main(void) {
+    unsigned char msg_hash[32] = {0}; /* This must be a hash of the message. otherwise ECDSA is easily broken. */
```
0xB10C, line 18: nit: It isn't immediately clear to me (without spending time thinking about it or looking it up) why ECDSA is easily broken if we sign the message rather than the hash of the message. Adding a keyword or simple and short explanation could be helpful.

0xB10C (commented):
Concept ACK, usage examples are helpful.

examples/random.h
```diff
@@ -0,0 +1,73 @@
...
+#if defined(_WIN32)
+    NTSTATUS res = BCryptGenRandom(NULL, data, size, BCRYPT_USE_SYSTEM_PREFERRED_RNG);
+    if (res != STATUS_SUCCESS || size > ULONG_MAX) {
+        return 0;
+    } else {
+        return 1;
+    }
+#elif defined(__linux__) || defined(__FreeBSD__)
+    /* If `getrandom(2)` is not available you should fallback to /dev/urandom */
+    ssize_t res = getrandom(data, size, 0);
```
siv2r, line 47: Is there any specific reason why `getrandom()` is used for Linux but `getentropy()` is used for Apple. Is it due to availability issues?

The GNU C library manual says this:
>Most applications should use getentropy. The getrandom function is intended for low-level applications which need additional control over blocking behavior.

https://www.gnu.org/software/libc/manual/html_node/Unpredictable-Bytes.html

Emzy:
Tested ACK 4c43382 (Ubuntu 20.04.3 LTS)

README.md
real-or-random: and s/Examples/examples

scgbckbone:
Tested ACK
* ubuntu 20.04 `Linux workstation 5.13.0-28-generic #31~20.04.1-Ubuntu SMP Wed Jan 19 14:08:10 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux`
* windows 10 `MINGW64_NT-10.0-19044 DESKTOP-UV75MOO 3.3.3-341.x86_64 2022-01-18 13:00 UTC x86_64 Msys`

examples/random.h
elichai: Hmm, mostly because I read the linux docs etc, and `getrandom` is a kernel thing, while `getentropy` is a glibc thing to emulate what OpenBSD does.
Also, most other libs also call `getrandom` on linux even though they call `getentropy` on OpenBSD/MacOS:
https://github.com/bitcoin-core/secp256k1/pull/748#discussion_r603295709

Not sure if this is convincing enough or not

elichai:
Fixed comments, and squashed

jonasnick (commented):
ACK beecb9a3b3f4936eadb678e7e931c90e8fa83de6

examples/ecdh.c
```diff
@@ -0,0 +1,124 @@
...
+        printf("Failed to generate randomness\n");
+        return 1;
+    }
+    /* Randomizing the context is recommended to protect against side-channel
+     * leakage See `secp256k1_context_randomize` in secp256k1.h for more
+     * information about it. This should never fail. */
+    return_val = secp256k1_context_randomize(ctx, randomize);
+    assert(return_val);
+
+    /*** Key Generation ***/
```
real-or-random, line 46: It may be good to add a header to every code section in this part, e.g., "Party 1: " and "Party 2:" or similar, to make sure.

This would require splitting the key generation loop but I think that's really better than. Real code wouldn't work like this, so I think we should avoid it in example code.

This is also something I can do in a follow-up PR.

real-or-random (commented):
@elichai Can you quickly fix the copyright and the msg_hash comment?

 I think then this is good to go. We had many eyes on it and I can address the other two comments in another PR.

examples/ecdsa.c
```diff
@@ -0,0 +1,135 @@
...
+    len = sizeof(compressed_pubkey);
+    return_val = secp256k1_ec_pubkey_serialize(ctx, compressed_pubkey, &len, &pubkey, SECP256K1_EC_COMPRESSED);
+    assert(return_val);
+    /* Should be the same size as the size of the output, because we passed a 33 byte array. */
+    assert(len == sizeof(compressed_pubkey));
+
+    /*** Signing ***/
+
+    /* Generate an ECDSA signature, note that even though here `msg_hash` is set
+    * to zeros, it MUST contain a hash, otherwise ECDSA is easily broken.
```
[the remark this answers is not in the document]
elichai: Thanks! fixed

examples/ecdh.c
```diff
@@ -0,0 +1,124 @@
+/*************************************************************************
+ * Copyright (c) 2020-2021 Elichai Turkel                                *
+ * Distributed under the CC0 software license, see the accompanying file *
+ * EXAMPLES_COPYING or https://creativecommons.org/publicdomain/zero/1.0 *
+ *************************************************************************/
```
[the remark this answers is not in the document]
elichai: Yeah, I hate dealing with licenses

real-or-random (approved):
ACK 7c9502cece9c9e8d811333f7ab5bb22f4eb01c04

As I said, I can create a follow-up PR.

The one CI failure for the sage prover is harmless and only because just sage on CI was introduced after the last rebase of this PR. It will work on master.

jonasnick (commented):
ACK 7c9502cece9c9e8d811333f7ab5bb22f4eb01c04

PR #750, opened 2020-05-03
Add macOS to the CI

elichai:
This adds macOS to the travis so we can test all the same configurations under macOS, both with Apple's clang and proper gcc on macOS.
It also runs the ctime test, and valgrind on the tests just like on linux.

The current travis script is pretty messy because we added more and more configuration over time each one required more complex ifs making it harder to read, and because of the somewhat complex logic it failed[1] on macOS(having an old bash version) so I decided to just throw it into a standalone sh script instead, that way it can be formatted nicely and with `-e` it will fail on every error without needing to put `&&` everywhere.

Some changed in the script:

* Replaced the usage of libtool with the locally generated libtool (https://github.com/bitcoin-core/secp256k1/pull/723#issuecomment-622999449)

* Added `--error-exitcode=42` to the ctime tests because they currently silently fail on `-O0` (https://travis-ci.org/github/bitcoin-core/secp256k1/jobs/681571334#L454) and disabled the ctime tests on `-O0`.

* Moved the valgrind tests to the matrix so that they'll run on both gcc and clang and on macOS.
(also, now that #710 is merged we always pass `-DVALGRIND` when the valgrind headers exist but I left the explicit CFLAGS in those tests anyway, there's no harm in explicitly doing that)

* Removed the use of `EXTRAFLAGS` for setting `CFLAGS`, it's enough to just set `CFLAGS` directly and it can cause troubles in sh (the whole `EXTRAFLAGS="--disable-openssl-tests CPPFLAGS=-DVALGRIND"`)

* We have to explicitly set the gcc version on macOS+gcc because macOS ship with a fake gcc which is basically just an alias to their clang compiler, and installing proper gcc from brew adds a `gcc-*` binary and doesn't replace the `gcc` binary, so we have to explicitly set `CC=gcc-9` under that scenario, so I also explicitly install `gcc@9` so it shouldn't break when macOS gets gcc-10.

* Bumped ubuntu to bionic because of https://github.com/bitcoin-core/secp256k1/pull/748#discussion_r418228255 (the end of `End of Standard Support` is in a year anyway) it's in a separate commit so that if anyone have concerns I'll just drop that commit.

1. https://travis-ci.org/github/elichai/secp256k1/jobs/681663742#L336


real-or-random:
That's very neat.  I'll have a closer look later. Let's try to get this merged before I continue to work on #723 because this here solves a bunch of the issues I encountered there.

contrib/travis.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/sh -ex
```
real-or-random, line 1: nit: it may be easier to spot for the reader to write
```suggestion
#!/bin/sh

set -e
set -x
```

contrib/travis.sh
```diff
@@ -0,0 +1,57 @@
...
+   make -j2
+# the `--error-exitcode` is required to make the test fail if valgrind found errors, otherwise it'll return 0 (http://valgrind.org/docs/manual/manual-core.html)
+   valgrind --error-exitcode=42 ./tests 16
+   valgrind --error-exitcode=42 ./exhaustive_tests
+   fi
+if [ -n "$BENCH" ]
+then
+    if [ -n "$VALGRIND" ]
+    then
+        EXEC='./libtool --mode=execute valgrind --error-exitcode=42'
```
real-or-random, line 37: Can you explain in a comment why we're using the internal one?

contrib/travis.sh
```diff
@@ -0,0 +1,57 @@
...
+    else
+        EXEC=
+    fi
```
real-or-random, line 40: I first wanted to write it would be cleaner so set `SECP256K1_BENCH_ITERS=2` here, for clarity. But I'm not entirely sure. If we set it here, it won't be configurable.  Bu then we should at least add a comment that the caller should set the ITERS to a low value. What do you think?

.travis.yml
```diff
@@ -71,41 +95,16 @@ matrix:
...
+# travis_wait extends the 10 minutes without output allowed (https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received)
+script: travis_wait 60 ./contrib/travis.sh
 
 after_script:
     - cat ./tests.log
     - cat ./exhaustive_tests.log
     - cat ./valgrind_ctime_test.log
     - cat ./bench.log
+    - $CC --version
+    - valgrind --version
```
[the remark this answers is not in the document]
elichai: I'll make sure it prints after apt/brew updates and if it does I'll move.

contrib/travis.sh
elichai: Hmm I think this script is 100% travis oriented anyway. so I'm not sure there's a meaningful difference.
If the script was written in a way that we could also run it locally pretty easily then maybe it would have (although the amount of env vars make it pretty hard to run locally)

real-or-random: > Hmm I think this script is 100% travis oriented anyway. so I'm not sure there's a meaningful difference.

Sure, and no need to change that right now. (Even though we want to change it in the future.) My main concern is that you don't see the low iteration count when looking at the script. Env variables are great in hiding details. It looks like we're just invoking the full benchmarks.  My maybe a comment is indeed the best.

real-or-random:
Fwiw, I think `travis_wait` does not work for us currently and just hides the timeout. It does not work with forking processes.

Proof: https://travis-ci.org/github/bitcoin-core/secp256k1/jobs/682613807#L1112

See 
 * https://github.com/travis-ci/travis-ci/issues/7431
 * https://github.com/travis-ci/travis-ci/issues/8526
 * https://github.com/travis-ci/travis-ci/issues/5281 

I had noticed this already when working on #723 but never wrote it down. If you want, you can give it a try. The issues above should provide some pointers, in particular PRs from other projects. But we can do this also in another PR. After this, was broken before this PR already.

real-or-random:
Any idea if that `dsymutil` thing here is specific to valgrind on MacOS? 
https://travis-ci.org/github/bitcoin-core/secp256k1/jobs/682613807#L1057

elichai:
> Any idea if that `dsymutil` thing here is specific to valgrind on MacOS?
> https://travis-ci.org/github/bitcoin-core/secp256k1/jobs/682613807#L1057

Sounds like it:
https://llvm.org/docs/CommandGuide/dsymutil.html

And here: https://valgrind.org/docs/manual/manual-core.html
> This option is only relevant when running Valgrind on Mac OS X.

.travis.yml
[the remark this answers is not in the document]
elichai: Just realized a problem with that.
We override `CC` in the sh script, so if we want to know what compiler is it using we must do it after the script :/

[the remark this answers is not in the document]
real-or-random: Well, okay, I guess then it's better not to move it. ^^

elichai:
Much better now :)
2-3 minutes per macOS job instead of ~11 minutes. now I hope the valgrind test won't time out

real-or-random:
Great, so I think this is ready (from my side at least ^^). One last question: Have you verified that a ct-test failure indeed makes the build fail? I guess this should be tested because this can be broken in subtle ways as I learned in my own PR...

elichai:
> Great, so I think this is ready (from my side at least ^^). One last question: Have you verified that a ct-test failure indeed makes the build fail? I guess this should be tested because this can be broken in subtle ways as I learned in my own PR...
<details>
<summary>I applied the following diff</summary>
```
diff --git a/src/field_5x52_impl.h b/src/field_5x52_impl.h
index 71aa550..348cca0 100644
--- a/src/field_5x52_impl.h
+++ b/src/field_5x52_impl.h
@@ -465,13 +465,12 @@ static SECP256K1_INLINE void secp256k1_fe_cmov(secp256k1_fe *r, const secp256k1_
 }
 
 static SECP256K1_INLINE void secp256k1_fe_storage_cmov(secp256k1_fe_storage *r, const secp256k1_fe_storage *a, int flag) {
-    uint64_t mask0, mask1;
-    mask0 = flag + ~((uint64_t)0);
-    mask1 = ~mask0;
-    r->n[0] = (r->n[0] & mask0) | (a->n[0] & mask1);
-    r->n[1] = (r->n[1] & mask0) | (a->n[1] & mask1);
-    r->n[2] = (r->n[2] & mask0) | (a->n[2] & mask1);
-    r->n[3] = (r->n[3] & mask0) | (a->n[3] & mask1);
+    if (flag) {
+        r->n[0] = a->n[0];
+        r->n[1] = a->n[1];
+        r->n[2] = a->n[2];
+        r->n[3] = a->n[3];
+    }
 }
```
</details>

And as you can see here https://travis-ci.org/github/elichai/secp256k1/builds/684392980 everything failed except the jobs with `FIELD=32bit`

contrib/travis.sh
```diff
@@ -0,0 +1,61 @@
...
+then
+    if [ -n "$VALGRIND" ]
+    then
+        # Using the local `libtool` because on macOS the system's libtool has nothing to do with GNU libtool
+        EXEC='./libtool --mode=execute valgrind --error-exitcode=42'
+    else
+        EXEC=
+    fi
+    SECP256K1_BENCH_ITERS="$ITERS" $EXEC ./bench_ecmult >> bench.log 2>&1
+    SECP256K1_BENCH_ITERS="$ITERS" $EXEC ./bench_internal >> bench.log 2>&1
```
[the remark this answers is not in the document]
elichai: I saw that option in shellcheck, but It didn't really seemed easier to read, I can do that anyway if you prefer.

About the ITERS, I do think it's readable right not (especially compared to before) but I can def export it right before instead.

contrib/travis.sh
```diff
@@ -0,0 +1,61 @@
...
+        SECP256K1_BENCH_ITERS="$ITERS" $EXEC ./bench_recover >> bench.log 2>&1
+    fi
+    if [ "$ECDH" = "yes" ]
+    then
+        SECP256K1_BENCH_ITERS="$ITERS" $EXEC ./bench_ecdh >> bench.log 2>&1
+    fi
+fi
+if [ -n "$CTIMETEST" ]
+then
+   ./libtool --mode=execute valgrind --error-exitcode=42 ./valgrind_ctime_test > valgrind_ctime_test.log 2>&1
```
real-or-random, line 60: nit: indentation

contrib/travis.sh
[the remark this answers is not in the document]
jonasnick: Yeah, doing both would be easier to read, better abstracted and more idiomatic imo.

real-or-random:
@elichai Is this ready for review?

elichai:
Yes

contrib/travis.sh
```diff
@@ -0,0 +1,65 @@
...
+if [ -n "$BENCH" ]
+then
+    if [ -n "$VALGRIND" ]
+    then
+        # Using the local `libtool` because on macOS the system's libtool has nothing to do with GNU libtool
+        EXEC='./libtool --mode=execute valgrind --error-exitcode=42'
+    else
+        EXEC=
+    fi
+    # This limits the iterations in the benchmark underneath to ITER(set in .travis.yml) iterations.
```
jonasnick, line 45: Took me a couple of seconds to understand what you mean by underneath (perhaps because it's usually only used with an object?). Maybe replace with below. Otherwise PR looks good!

elichai: Done

real-or-random:
ACK 71757da5ccece100b1eca6c70b4d87e2542cff97 I inspected the diff

PR #754, opened 2020-05-20
Fix uninit values passed into cmov

### LINKED ISSUE #753: Instances of operations on uninitialized values
Hi,
Currently we have 2 instances of operations on uninitialized values, in the cmov implementation.
all the `_cmov` implementations read from the resulting variable(ie `r->n[0] & mask0`) so if `r` points to uninitialized values this is undefined behavior [1] [2]
Unlike the `memcpy` debate, actual user operations and conditions on uninit values are assumed to not happen in the compiler and are used for optimizations (especially in clang)

**Instance 1:**
In ecdsa_sign, if `noncefp` returns 0 then we get to:
https://github.com/bitcoin-core/secp256k1/blob/41fc78560223aa035d9fe2cbeedb3ee632c740e2/src/secp256k1.c#L517-L518
with both `r` and `s` being unitialized. 
**Instance 2:**
In ecdh, we pass `&tmpa`  here: https://github.com/bitcoin-core/secp256k1/blob/41fc78560223aa035d9fe2cbeedb3ee632c740e2/src/ecmult_const_impl.h#L186 to the `ECMULT_CONST_TABLE_GET_GE` in an uninit state, which become the `r` value in the macro. 
Then in:
https://github.com/bitcoin-core/secp256k1/blob/41fc78560223aa035d9fe2cbeedb3ee632c740e2/src/ecmult_const_impl.h#L31-L32
 we cmov into `&(r)->x` which will perform operations on uninit values.
notice that under `VERIFY` we actually do clear those before:
https://github.com/bitcoin-core/secp256k1/blob/41fc78560223aa035d9fe2cbeedb3ee632c740e2/src/ecmult_const_impl.h#L26-L27

Possible solutions:
The only solution I can currently think of is default initializing those arguments.

Mitigations:
We could add `VALGRIND_CHECK_MEM_IS_DEFINED(r, sizeof(*r));` to all the cmovs but we only enable valgrind(`-DVALGRIND`) under tests, which use the VERIFY macro too, so that will not catch the second case.

[1] http://www.open-std.org/jtc1/sc22/wg14/www/docs/dr_451.htm
[2] https://markshroyer.com/2012/06/c-both-true-and-false/

elichai:
This should fix #753.
Used @peterdettman's solution here for the `ECMULT_CONST_TABLE_GET_GE` https://github.com/bitcoin-core/secp256k1/issues/753#issuecomment-631316091
and in ecdsa_sign I initialize `s` and `r` to a zero scalar.

The second commit adds a valgrind check to the cmovs that could've caught this (in ecdsa_sign, not in ecmult_const because there's a scalar clear there under `VERIFY_SETUP`)

### DIFF
--- a/src/field_10x26_impl.h
+++ b/src/field_10x26_impl.h
@@ -1097,6 +1097,7 @@ static void secp256k1_fe_sqr(secp256k1_fe *r, const secp256k1_fe *a) {
 
 static SECP256K1_INLINE void secp256k1_fe_cmov(secp256k1_fe *r, const secp256k1_fe *a, int flag) {
     uint32_t mask0, mask1;
+    VG_CHECK_VERIFY(r->n, sizeof(r->n));
     mask0 = flag + ~((uint32_t)0);
     mask1 = ~mask0;
     r->n[0] = (r->n[0] & mask0) | (a->n[0] & mask1);
@@ -1119,6 +1120,7 @@ static SECP256K1_INLINE void secp256k1_fe_cmov(secp256k1_fe *r, const secp256k1_
 
 static SECP256K1_INLINE void secp256k1_fe_storage_cmov(secp256k1_fe_storage *r, const secp256k1_fe_storage *a, int flag) {
     uint32_t mask0, mask1;
+    VG_CHECK_VERIFY(r->n, sizeof(r->n));
     mask0 = flag + ~((uint32_t)0);
     mask1 = ~mask0;
     r->n[0] = (r->n[0] & mask0) | (a->n[0] & mask1);
--- a/src/ecmult_const_impl.h
+++ b/src/ecmult_const_impl.h
@@ -14,7 +14,7 @@
 
 /* This is like `ECMULT_TABLE_GET_GE` but is constant time */
 #define ECMULT_CONST_TABLE_GET_GE(r,pre,n,w) do { \
-    int m; \
+    int m = 0; \
     /* Extract the sign-bit for a constant time absolute-value. */ \
     int mask = (n) >> (sizeof(n) * CHAR_BIT - 1); \
     int abs_n = ((n) + mask) ^ mask; \
@@ -25,7 +25,11 @@
     VERIFY_CHECK((n) <=  ((1 << ((w)-1)) - 1)); \
     VERIFY_SETUP(secp256k1_fe_clear(&(r)->x)); \
     VERIFY_SETUP(secp256k1_fe_clear(&(r)->y)); \
-    for (m = 0; m < ECMULT_TABLE_SIZE(w); m++) { \
+    /* Unconditionally set r->x = (pre)[m].x. r->y = (pre)[m].y. because it's either the correct one \
+     * or will get replaced in the later iterations, this is needed to make sure `r` is initialized. */ \
+    (r)->x = (pre)[m].x; \
+    (r)->y = (pre)[m].y; \
+    for (m = 1; m < ECMULT_TABLE_SIZE(w); m++) { \
         /* This loop is used to avoid secret data in array indices. See
          * the comment in ecmult_gen_impl.h for rationale. */ \
         secp256k1_fe_cmov(&(r)->x, &(pre)[m].x, m == idx_n); \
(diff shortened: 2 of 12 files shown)


src/field_10x26_impl.h
```diff
@@ -10,6 +10,10 @@
 #include "util.h"
 #include "field.h"
 
+#if defined(VALGRIND)
```
real-or-random, line 13: I think for now it's cleaner  to check also `VERIFY` here. (Same in the other files)

real-or-random:
> I think the valgrind checks need some more Concept ACKs because this is "new tooling" in some sense. I think they're really a good idea here because it's somewhat unexpected that cmov functions will read their target.

Ah yes, it seems that gcc 10 will just warn us here. (I get warnings only for the `ecdsa_sign` function and no for `ecmult_const`, with ecdh enabled). But since valgrind does not help for `ecmult_const`, it's maybe simpler not to include the valgrind checks. I don't have strong opinions here. Either way is okay with me.

sipa:
I think this is a grey area in the specification, see https://queue.acm.org/detail.cfm?id=3041020 (in C89 it's probably even less clear), because uintXX_t cannot have trap representations. The only question is whether just using an unspecified value on itself is UB, which seems up for debate.

Concept ACK on the changes for an abundance of caution.

I'm not convinced about the additional Valgrind check, because this seems to be outside of the scope of what Valgrind is even supposed to check (even without the additional check, it passing means there is no non-constant time behaviour; whether it's not triggering any other kind of UB beyond that seems more a job for sanitizers, who understand the source code).

elichai:
> Ah yes, it seems that gcc 10 will just warn us here.I get warnings only for the `ecdsa_sign` function and no for `ecmult_const`, with ecdh enabled. 

Do you get those warnings consistently? and on what parts of the code? I only got once on the `low_impl` implementation

> I think this is a grey area in the specification, see https://queue.acm.org/detail.cfm?id=3041020 (in C89 it's probably even less clear), because uintXX_t cannot have trap representations. The only question is whether just using an unspecified value on itself is UB, which seems up for debate.

I must say I wrote down a long post arguing that I do think it's UB and I think compilers (especially llvm/clang) are really harsh on uninitialized values but then I came across this thread: https://lists.llvm.org/pipermail/llvm-commits/Week-of-Mon-20180528/556592.html
So you're right, this isn't clear at all.

> I'm not convinced about the additional Valgrind check 

I'm not either, it was just the only test I could come up with to catch this in the future, but I can drop that commit if there's no consensus about it.

real-or-random:
> I'm not convinced about the additional Valgrind check, because this seems to be outside of the scope of what Valgrind is even supposed to check (even without the additional check, it passing means there is no non-constant time behaviour; whether it's not triggering any other kind of UB beyond that seems more a job for sanitizers, who understand the source code).

I don't think it's outside the scope of what valgrind is supposed to check. Valgrind tracks uninitialized memory (on a binary level, not on a C abstract machine level). It does not complain about the existence of uninitialized value, it complains if the externally visible behaviour of your binary depends on uninitialized values (e.g., if a branch depends on it, which valgrind always counts as observable.) This is not the case here as we discussed in #753.

But because valgrind tracks uninitialized memory, you can ask valgrind explicitly to check if a given piece of memory is defined, and that's what we're doing here. 

So that the check here passed is *not* related to the code being constant time (because it's not about a branch).


> I'm not either, it was just the only test I could come up with to catch this in the future, but I can drop that commit if there's no consensus about it.
Yes, shall we get the first commit merged if the second is debatable?

elichai:
FYI, this new valgrind check won't even run under the valgrind ctime test because it's guarded under VERIFY which isn't set for the ctime test.
but as @real-or-random this has nothing to do with constant/variable time but with alerting to future contributors in the future that it might be UB to pass uninit `r` here.

sipa:
> FYI, this new valgrind check won't even run under the valgrind ctime test because it's guarded under VERIFY which isn't set for the ctime test.

It seems rather pointless then?

elichai:
> It seems rather pointless then?

The point is that `valgrind ./tests` and/or `valgrind ./exhuastive_tests` will catch this. (or even running valgrind on a release binary)

sipa:
Ok, Concept ACK.

I think it's a bit ugly to have `#ifdef VALGRIND` all over the place, though.

Perhaps the VG_UNDEF/VG_CHECK macros from tests.c can be moved to util.h, so these checks can be just `#ifdef VERIFY`. Alternatively, there could be a VG_CHECK_VERIFY that is only defined when VERIFY is set even.

elichai:
@sipa Is this better?

sipa:
utACK f79a7adcf555ccc78b591850ea15c64fbfbca152

PR #755, opened 2020-05-27
Recovery sign, add to constant time test, and eliminate non ct operators.

elichai:
Hi,
The recovery module was overlooked in #708 and #710, so this adds it to the `valgrind_ctime_test` and replaces the secret dependent branching with the cmovs,
I created a new function `secp256k1_ecdsa_sign_inner` (feel free to bikeshed) which does the logic both for ecdsa_sign and for ecdsa_sign_recoverable, such that next time when things get changed/improved in ecdsa it will affect the recoverable signing too.


src/util.h
```diff
@@ -175,4 +175,12 @@ static SECP256K1_INLINE void memczero(void *s, size_t len, int flag) {
     }
 }
 
+/** If flag is true, set *r equal to *a; otherwise leave it. Constant-time. */
```
real-or-random, line 178: nit:  comment should also cover requirement to be defined, similar as in #754

elichai: You're right :D fixed.

src/util.h
```diff
@@ -175,4 +175,12 @@ static SECP256K1_INLINE void memczero(void *s, size_t len, int flag) {
     }
 }
 
+/** If flag is true, set *r equal to *a; otherwise leave it. Constant-time.  Both *r and *a must be initialized.*/
+static SECP256K1_INLINE void secp256k1_int_cmov(int *r, const int *a, int flag) {
+    uint32_t mask0, mask1;
+    mask0 = flag + ~((uint32_t)0);
```
real-or-random, line 181:
```suggestion
    unsigned int mask0, mask1;
    mask0 = flag + ~((unsigned int)0);
```
or shorter
```suggestion
    unsigned int mask0, mask1;
    mask0 = flag + ~0u;
```
is probably cleaner to review and more portable because `unsigned int` is guaranteed to have the same bitlength as `int`.

elichai: I wonder now if I'm sure this function is 100% safe for signed integers (I'll need to look again at the integer promotion rules etc)
I think it is safe though?
(flag will probably be promoted to unsigned, and then the overflow will be safe and I think all AND operations on ints are safe )

real-or-random: I thought about this too.

> (flag will probably be promoted to unsigned, and then the overflow will be safe and I think all AND operations on ints are safe )

Indeed, this was my understanding, too! (`flag` will not be "promoted" it will be "converted" to unsigned but this is just wording). The same is true in `(*r & mask0)` where `*r` will be converted to unsigned, and the same for `(*a & mask1)`, so everything on the right-hand side of the final assignment is unsigned arithmetic.

jonasnick: Would it make sense to test this function with `a` equal `INT_MIN` and `INT_MAX`?

real-or-random: I guess it can't hurt and it's cheap to do.

elichai: I'll add a test, FYI passing some flags to clang shows us exactly how the conversions happen here: https://godbolt.org/z/t9GHKK

Though I wouldn't mind changing it to something like this: https://godbolt.org/z/5NxyE5

jonasnick: Yeah explicit conversions would be nice. PR looks good. Thanks for the tests.

elichai: argh, this cmov is implementation defined for any negative value :/
See http://port70.net/~nsz/c/c89/c89-draft.html#3.2.1.2
> an unsigned integer is converted to its corresponding signed integer, if the value cannot be represented the result is implementation-defined. 

So even `(int)(unsigned int)(-1)` is implementation defined behavior.

real-or-random: Yeah, even though all non-obscure C implementations will do the most efficient thing possible, i.e., make the cast a noop on the bit level, which implies that casting back and forth is fine. So it will probably never be a problem for us. But it's cleaner not to rely on it. I suggest we simply `VERIFY` that inputs are nonnegative and a comment why this VERIFY makes sense?

elichai: Yeah, people on #C in freenode told me I'm being overly paranoid lol, VERIFY is more than enough

src/util.h
```diff
@@ -175,4 +175,12 @@ static SECP256K1_INLINE void memczero(void *s, size_t len, int flag) {
     }
 }
 
+/** If flag is true, set *r equal to *a; otherwise leave it. Constant-time.  Both *r and *a must be initialized.*/
+static SECP256K1_INLINE void secp256k1_int_cmov(int *r, const int *a, int flag) {
+    unsigned int mask0, mask1;
+    mask0 = flag + ~0u;
+    mask1 = ~mask0;
+    *r = (*r & mask0) | (*a & mask1);
+}
```
real-or-random, line 184: I'm sure people won't like this but this is perfectly well-defined. :p 
```suggestion
static SECP256K1_INLINE void secp256k1_int_cmov(int *r, const int *a, int flag) {
    int mask0, mask1;
    mask0 = 1 - flag;
    mask1 = flag;
    *r = (*r * mask0) + (*a * mask1);
}
```

elichai: Nice!
Interesting to think of a cmov using arithmetic and not bitwise masking.
Why won't people like it?

real-or-random: > Interesting to think of a cmov using arithmetic and not bitwise masking.

You see that often in pseudocode.

> Why won't people like it?

Integer multiplication is slower than `&`. But psst, let's not start the discussion and go with the `VERIFY`. ;)

real-or-random:
need rebase

elichai:
I first fixed the comments from https://github.com/bitcoin-core/secp256k1/pull/755#discussion_r431696407
so it will be easy to see the diff(https://github.com/bitcoin-core/secp256k1/compare/009a515800ddcb76230a5fd49a78c4ce22fae23b..dcc80cf985dfed5659699eca2f830857b5a42497)

Now I'll rebase

src/secp256k1.c
```diff
@@ -515,8 +514,26 @@ int secp256k1_ecdsa_sign(const secp256k1_context* ctx, secp256k1_ecdsa_signature
...
+    if (recid) {
+        const int zero = 0;
+        secp256k1_int_cmov(recid, &zero, !ret);
+    }
+    return ret;
+}
+
+int secp256k1_ecdsa_sign(const secp256k1_context* ctx, secp256k1_ecdsa_signature *signature, const unsigned char *msg32, const unsigned char *seckey, secp256k1_nonce_function noncefp, const void* noncedata) {
+    /* Default initialization here is important so we won't pass uninit values to the cmov in the end */
+    secp256k1_scalar r = secp256k1_scalar_zero, s = secp256k1_scalar_zero;
```
jonasnick, line 528: Is initializing twice intentional? Looks more like this crept in during the rebase.

elichai: Yep, crept in when rebased. Thanks

src/secp256k1.c
```diff
@@ -515,8 +514,25 @@ int secp256k1_ecdsa_sign(const secp256k1_context* ctx, secp256k1_ecdsa_signature
     secp256k1_scalar_clear(&msg);
     secp256k1_scalar_clear(&non);
     secp256k1_scalar_clear(&sec);
-    secp256k1_scalar_cmov(&r, &secp256k1_scalar_zero, !ret);
-    secp256k1_scalar_cmov(&s, &secp256k1_scalar_zero, !ret);
+    secp256k1_scalar_cmov(r, &secp256k1_scalar_zero, !ret);
+    secp256k1_scalar_cmov(s, &secp256k1_scalar_zero, !ret);
+    if (recid) {
+        const int zero = 0;
+        secp256k1_int_cmov(recid, &zero, !ret);
```
sipa, line 521: Why? recid is not secret.

elichai: The problem isn't `recid` being secret, it's `ret` being secret.

sipa: I don't think there is much of an issue with declassifying ret. If the API is used correctly, ret is always 1 anyway.

elichai: If we assume a user won't call `sign_*` without verifying the secret key first then yeah we can declassify `is_sec_valid` (which is what contaminates `ret`)
https://github.com/bitcoin-core/secp256k1/blob/a39c2b09de304b8f24716b59219ae37c2538c242/src/secp256k1.c#L510-L513

real-or-random: @sipa This is not the scope of this PR. If we want to change this, I think that's a different discussion. 

The current signing code on master is overly cautious and assumes `is_sec_valid` to be a secret. That's exactly the reason why we cmov `r` and `s` (introduced by #710), see also the corresponding [code comment](https://github.com/bitcoin-core/secp256k1/pull/755/files/2876af4f8da952e39c06bc229d68cd4892ea2c85#diff-0d77700fa6e05639431c962ffaa5365aL510-L512) and the [exact reasoning in the PR](https://github.com/bitcoin-core/secp256k1/pull/710#discussion_r374510230).

sipa: @real-or-random Thanks for the references, that makes sense. My point was mostly that at some point, the complexity of removing decreasingly relevant forms of non-ctness aren't worth it, but this isn't too bad yet.

PR #761, opened 2020-06-22
Add cmake support

xDimon:
Signed-off-by: Dmitriy Khaustov aka xDimon <khaustov.dm@gmail.com>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,6 +31,11 @@ libtool
 *.lo
 *.o
 *~
+
+build/
+cmake-build-*
+.idea/
+
 src/libsecp256k1-config.h
 src/libsecp256k1-config.h.in
 src/ecmult_static_context.h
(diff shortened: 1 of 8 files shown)


elichai:
As much as CMake support can be an interesting topic this is the wrong repo, we don't have multiset and schnorr in master :)

xDimon:
> As much as CMake support can be an interesting topic this is the wrong repo, we don't have multiset and schnorr in master :)

Ouch... Of course. I'll come back later with changes =)

real-or-random:
I know it's not great but we're actually reasonably happy with automake (and it integrates with Bitcoin Core), so I can strictly speaking only speak for myself but I don't think there's much interest in maintaining CMake support. 

So before you spend too much on this, be warned that the chances of this getting merged are not too high.

Issue #758, opened 2020-06-12
contrib stuff: Aftermath of Bitcoin Core subtree update


real-or-random:
see https://github.com/bitcoin/bitcoin/pull/19228#issuecomment-643406092

PR #765, opened 2020-06-29
remove dead store in ecdsa_signature_parse_der_lax

fanquake:
This change was made in bitcoin/bitcoin without upstreaming. So this is
a followup to the comment here: https://github.com/bitcoin/bitcoin/pull/19228#issuecomment-641795558.

See also: https://github.com/bitcoin/bitcoin/pull/11073.


elichai:
utACK f00d6575ca0dcca11e085b20afa4d73dc8742ddc, it does look like we don't use that assignment

real-or-random:
Part of #758 but #758 needs more work.

PR #769, opened 2020-07-21
Undef HAVE___INT128 in basic-config.h to fix gen_context compilation

### LINKED ISSUE #768: Issue compiling on OpenBSD 6.7
Since the secp256k1 subtree update in https://github.com/bitcoin/bitcoin/pull/19228, it hasn't been possible to cleanly compile Bitcoin Core on OpenBSD. Building fails with (excerpt from just building libsecp, `cc` is Clang 8.0.1):
```bash
./autogen.sh
./configure CC=cc MAKE=gmake
gmake -j6 V=1
...
gmake
gcc -I. -I./src -Wall -Wextra -Wno-unused-function -g -c src/gen_context.c -o gen_context.o
In file included from src/gen_context.c:16:
src/util.h:179: error: expected '=', ',', ';', 'asm' or '__attribute__' before 'uint128_t'
gmake: *** [Makefile:1689: gen_context.o] Error 1
```

libsecp does some native compiler detection in configure, and then uses that compiler specifically for the ecmult precomputation. i.e:
https://github.com/bitcoin-core/secp256k1/blob/3f4a5a10e43bfc8dae5b978cb39aa2dfbaf4d713/configure.ac#L186-L188
and then:
https://github.com/bitcoin-core/secp256k1/blob/3f4a5a10e43bfc8dae5b978cb39aa2dfbaf4d713/Makefile.am#L129-L130

The problem is that this ends up picking up and using OpenBSDs GCC 4.2.1, which barfs when it gets to util.h:
https://github.com/bitcoin-core/secp256k1/blob/3f4a5a10e43bfc8dae5b978cb39aa2dfbaf4d713/src/util.h#L179

I realise there might not be much that can be done here, given that it's the `AX_PROG_CC_FOR_BUILD` macro that is returning `gcc` to use. However thought I'd open an issue here for any thoughts, before we update our docs. I've suggested that affected users can just pass `CC_FOR_BUILD=cc` to configure when building. Issue here: https://github.com/bitcoin/bitcoin/issues/19559.

real-or-random:
Fixes #768.

### DIFF
--- a/src/basic-config.h
+++ b/src/basic-config.h
@@ -25,6 +25,7 @@
 #undef USE_SCALAR_INV_BUILTIN
 #undef USE_SCALAR_INV_NUM
 #undef ECMULT_WINDOW_SIZE
+#undef HAVE___INT128 /* used in util.h */
 
 #define USE_NUM_NONE 1
 #define USE_FIELD_INV_BUILTIN 1


sipa:
utACK

Issue #702, opened 2019-12-18
Compile time SHA256 override?


gmaxwell:
It's a bit silly that the library packs in its own sha256 even when the user has their own and on small devices it's a costly waste of space.  In places like Bitcoin Core the environment has a fast SIMD or SHA-NI sha256.  It would be nice if there were some -Dlibsecp2561k_sha256init=x  -Dlibsecp2561k_sha256update=y  -Dlibsecp2561k_sha256final=z  settings that could be used to compile time substitute in another library and leave the internal one out.

elichai:
Last time we discussed this on IRC some people said that providing 3 functions might be a bit much and having a single sha256 function instead is better.
I think that a single one might be a bit awkward because it requires allocating a big buffer on the stack that can contain everything (and then more things you need to clean).

What are your thoughts on this?

real-or-random:
I think you'll want this feature only if you seriously care about performance. And in this case, you probably have the more complex API with 3 functions anyway.

apoelstra:
Also, in practice a single function is often harder to use - if you want to hash a series of things it's nice to just throw them all into a hash engine ... if you need to create a buffer first then (depending on your language) you may need to think about buffer allocation and indexing.

Another alternate might be to just conditionally compile out the sha256 functions, and require the user provide replacements with exactly the correct names in order for linking to succeed. This is what we do for the default error callbacks for example.

gmaxwell:
> only if you seriously care about performance.

I think the vast majority of users would be ones that don't want to have an extra 19kilobytes in their binary.

> Another alternate might be to just conditionally compile out the sha256 functions, and require the user provide replacements with exactly the correct names in order for linking to succeed.

That would be ducky too.

elichai:
Unless anyone already works on this I'd like to give it a try :)

real-or-random:
> Unless anyone already works on this I'd like to give it a try :)

Could you look into this?

elichai:
> Could you look into this?

IIRC the problem I encountered was the SHA2 context.
I remember a few solutions but none of them was great:

1. Decide that the context needs to be size X, provide an opaque struct with unsigned array size X, and the implementer can memcpy this into his own Context back and forth (or type pun it? need to look into what the standard says on that).

2. The implementer also needs to provide a header with the name `sha256.h` that defines the struct+functions using the same names/symbols we use.

[...]

PR #770, opened 2020-07-21
Get rid of autoconf check for endianness

real-or-random:
This does not fix any particular issue but it's preferable to not
rely on autoconf. This avoids endianness mess for users on BE hosts
if they use their build without autoconf.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -493,8 +493,6 @@ if test x"$enable_module_recovery" = x"yes"; then
   AC_DEFINE(ENABLE_MODULE_RECOVERY, 1, [Define this symbol to enable the ECDSA pubkey recovery module])
 fi
 
-AC_C_BIGENDIAN()
-
 if test x"$use_external_asm" = x"yes"; then
   AC_DEFINE(USE_EXTERNAL_ASM, 1, [Define this symbol if an external (non-inline) assembly implementation is used])
 fi
--- a/src/hash_impl.h
+++ b/src/hash_impl.h
@@ -27,11 +27,11 @@
     (h) = t1 + t2; \
 } while(0)
 
-#ifdef WORDS_BIGENDIAN
-#define BE32(x) (x)
-#else
-#define BE32(p) ((((p) & 0xFF) << 24) | (((p) & 0xFF00) << 8) | (((p) & 0xFF0000) >> 8) | (((p) & 0xFF000000) >> 24))
-#endif
+/* swap byte order on a LE system, identity function on a BE system */
+SECP256K1_INLINE static uint32_t be32(uint32_t n) {
+    unsigned char *p = (unsigned char *)&n;
+    return ((uint32_t)p[0] << 24) | ((uint32_t)p[1] << 16) | ((uint32_t)p[2] << 8) | (uint32_t)p[3];
+}
 
 static void secp256k1_sha256_initialize(secp256k1_sha256 *hash) {
     hash->s[0] = 0x6a09e667ul;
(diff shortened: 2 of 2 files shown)


real-or-random:
I guess we should merge #696 first.

gmaxwell:
9% slowdown in hash_rfc6979_hmac_sha256 on i.MX6 (arm7; gcc 4.9.2).  Not needing autotools for target endianness detection sounds good but a slowdown on on a popular LE embedded target does not. endian.h perhaps?

real-or-random:
> i.MX6 (arm7; gcc 4.9.2).

Hm, I had looked at the ASM for recent compilers:
https://godbolt.org/z/ndjvKs
Note that there's a large difference between GCC 4 and newer versions, with and without `-mcpu=cortex-a9`.

Maybe I'm ignorant but my response would be that I don't care about the performance with an age-old GCC 4. Is GCC 4.9 common for this target?

real-or-random:
Hashing performance will even less be a concern with a resolution of https://github.com/bitcoin-core/secp256k1/issues/702

gmaxwell:
It's pretty normal to have older (or even awful) compilers on embedded platforms, esp arm because devices usually depend on a vendor kernel that stops getting upgraded to new major versions and hangs around with old operating systems. On novena, the latest supported OS is debian jessie.  The situation for toolchains you might find setup for hardware wallets and whatnot isn't that different a story.

If there is a choice between being fast on older compilers or new, sure pick new. Or between something much harder to maintain and a small slowdown on older compilers, sure. Bit 9% is pretty big, and a byteswap macro is about as idiomatic as you get. 

Small embedded devices don't usually have hashing hardware so 702 doesn't really apply,  unless you think that it's okay to just provide objectively worse stuff because the user can fix this libraries mistakes?

Besides, it appears to be a similar similar story elsewhere, if less dramatic. GCC 10.1.1 on PPC64LE:

Before: hash_rfc6979_hmac_sha256: min 5.96us / avg 5.96us / max 5.97us
After: hash_rfc6979_hmac_sha256: min 6.10us / avg 6.10us / max 6.10us

real-or-random:
@gmaxwell You mentioned endian.h but this seems to be a portability mess, too. Or which header do you refer to?

Here's another idea: What about sticking to autoconf but also let it define a macro if it figures out little-endian. Then we can  #error out at least (or try __BYTE_ORDER__ as a fallback, which is at least available on GCC/clang, and #error if that fails too). That's not very ergonomic but at least it errs on the side of caution, whereas we currently simply assume LE if autoconf was not used. Creating wrong hashes could end up terribly, e.g., if you create a schnorr sig with a user-specified nonce function, you could end up signing different hashes with the same nonce.  

For enhanced safety, we could even add a cheap check in the context creation that checks at runtime if the endianness matches what it thinks it has been compiled for. Reminds me of my suggestion in #702.

gmaxwell:
Sounds okay to me. Also, erroring on a non-configuration makes sense. Portability is kinda complicated-- in the sense that a lot of things are in theory non-portable but in practice are fine (and a few things are in theory portable but in practice are not).  I think things like __BYTE_ORDER__ are reasonably widely supported.   Anyone building for a really weird platform is going to be able to follow a clearly stated #error.

real-or-random:
>  Portability is kinda complicated-- in the sense that a lot of things are in theory non-portable but in practice are fine (and a few things are in theory portable but in practice are not).

I was specifically referring to `endian.h` not being available on Mac OS for example, which is a practical issue.

>  I think things like **BYTE_ORDER** are reasonably widely supported. 

By the way, I found this hell yesterday: :D 
https://gist.github.com/jtbr/7a43e6281e6cca353b33ee501421860c#file-endianness-h-L39 

I guess the combination of `__BYTE_ORDER__`  (just this none, none of the weird stuff behind the link) and `#error` is good enough.

elichai:
What do you think about this? https://godbolt.org/z/f9jbbP

real-or-random:
> What do you think about something like this? https://godbolt.org/z/hfYa7K
> It still has the conditional you were trying to avoid, but it doesn't rely on autotools.

Oh letting constant folding optimize the run-time away is really a neat idea. GCC 4.9 seems to do better here. 

GCC 4.4.7 x86 and ARM GCC 4.5.4 without `-mcpu=cortex-a9` still don't get this, but this seems good enough. Hey, even MSVC manages to optimize this... Or what do you think @gmaxwell ?

gmaxwell:
That looks like an extremely non-ideomatic way to do a byteswap and I wouldn't be particularly confident that some random compiler would handle it correctly, or that the optimization wouldn't randomly break when a butterfly flapped its wings in the amazon.

elichai:
> That looks like an extremely non-ideomatic way to do a byteswap and I wouldn't be particularly confident that some random compiler would handle it correctly, or that the optimization wouldn't randomly break when a butterfly flapped its wings in the amazon.

Yeah it's quite sketchy hehe, I first saw something like this in NIST submissions, it was way worse and had UB, it was something along these lines:
```c
typedef enum {BIG, LITTLE, MIDDLE} Endianess;

Endianess Endian() {
    unsigned char arr[8];
    *(unsigned int*)arr = 0x01020304;
    if (arr[0] == 0x04) {
        return LITTLE;
    } else if (arr[0] == 0x01) {
        return BIG;
    } else {
        return MIDDLE;
    }
}
```

sipa:
FWIW, it seems that literally every compiler on godbolt (including the obscure ones outside of clang/gcc/icc/msvc, and ones for less common platforms) supports `__BYTE_ORDER` (and it equalling either `__BIG_ENDIAN` or `__LITTLE_ENDIAN`).

Perhaps it's reasonable to just use that in non-autotools environments?

real-or-random:
> FWIW, it seems that literally every compiler on godbolt (including the obscure ones outside of clang/gcc/icc/msvc, and ones for less common platforms) supports `__BYTE_ORDER` (and it equalling either `__BIG_ENDIAN` or `__LITTLE_ENDIAN`).
> 
> Perhaps it's reasonable to just use that in non-autotools environments?

Nice, then let's just do this and #error if it's not defined.

real-or-random:
> FWIW, it seems that literally every compiler on godbolt (including the obscure ones outside of clang/gcc/icc/msvc, and ones for less common platforms) supports `__BYTE_ORDER` (and it equalling either `__BIG_ENDIAN` or `__LITTLE_ENDIAN`).

I guess you meant `__BYTE_ORDER__`, that's apparently defined almost everywhere. A notable exception is MSVC of course... But we can test for this with `_MSC_VER` and it then offers `_WIN32`, which is "Defined as 1 when the compilation target is 32-bit ARM, 64-bit ARM, x86, or x64. Otherwise, undefined."  https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=vs-2019

real-or-random:
Superseded by #787.

PR #773, opened 2020-07-26
Fix some compile problems on weird/old compilers.

gmaxwell:
The visibility attribute is a GCC 4+ feature.
GCC 2.95 also warns about the unsigned/signed comparision.


real-or-random:
ACK 1309c03c45beece646a7d21fdb6a0e3d38adee2b I inspected the diff

PR #774, opened 2020-07-27
tests: Abort if malloc() fails during context cloning tests

real-or-random:
Found by the clang static analyzer.

This is the worst true positive that it found. I feel somewhat proud.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -182,8 +182,10 @@ void run_context_tests(int use_prealloc) {
     ecount2 = 10;
     secp256k1_context_set_illegal_callback(vrfy, counting_illegal_callback_fn, &ecount);
     secp256k1_context_set_illegal_callback(sign, counting_illegal_callback_fn, &ecount2);
-    secp256k1_context_set_error_callback(sign, counting_illegal_callback_fn, NULL);
-    CHECK(vrfy->error_callback.fn != sign->error_callback.fn);
+    /* set error callback (to a function that still aborts in case malloc() fails in secp256k1_context_clone() below) */
+    secp256k1_context_set_error_callback(sign, secp256k1_default_illegal_callback_fn, NULL);
+    CHECK(sign->error_callback.fn != vrfy->error_callback.fn);
+    CHECK(sign->error_callback.fn == secp256k1_default_illegal_callback_fn);
 
     /* check if sizes for cloning are consistent */
     CHECK(secp256k1_context_preallocated_clone_size(none) == secp256k1_context_preallocated_size(SECP256K1_CONTEXT_NONE));
@@ -239,7 +241,8 @@ void run_context_tests(int use_prealloc) {
     }
 
     /* Verify that the error callback makes it across the clone. */
-    CHECK(vrfy->error_callback.fn != sign->error_callback.fn);
+    CHECK(sign->error_callback.fn != vrfy->error_callback.fn);
+    CHECK(sign->error_callback.fn == secp256k1_default_illegal_callback_fn);
     /* And that it resets back to default. */
     secp256k1_context_set_error_callback(sign, NULL, NULL);
     CHECK(vrfy->error_callback.fn == sign->error_callback.fn);


real-or-random:
By the way, the static analyzer is very easy to run and it's very easy to inspect the results:
https://clang-analyzer.llvm.org/scan-build.html 
You need clang 10.

gmaxwell:
I was going to comment that the cloning blocks seemed misplaced in the middle of error counting tests, but I see why they were added there: to check if clone copies the error callback.   Reasonable enough change.

For extra fun, it's possible to actually test handling of malloc failure:  https://github.com/xiph/opus/blob/master/tests/test_opus_api.c#L1751

src/tests.c
```diff
@@ -182,7 +182,8 @@ void run_context_tests(int use_prealloc) {
     ecount2 = 10;
     secp256k1_context_set_illegal_callback(vrfy, counting_illegal_callback_fn, &ecount);
     secp256k1_context_set_illegal_callback(sign, counting_illegal_callback_fn, &ecount2);
-    secp256k1_context_set_error_callback(sign, counting_illegal_callback_fn, NULL);
+    /* try to set error callback (to a function that still aborts in case malloc() fails in secp256k1_context_clone() below) */
```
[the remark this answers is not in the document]
real-or-random: Yeah, try and see if it works. It's all over that function:
https://github.com/bitcoin-core/secp256k1/blob/3b789d5edbf9faa356744d71ddd023ecf076ec90/src/tests.c#L303

But okay, I've removed it now and also made the checks more precise.

elichai:
> I run this with internal bignum, asm disabled (you get more false positives otherwise) on 64 bit without endo. Other configurations may be interesting, too, and I invite everyone to verify that the remaining issues are false positives.

Ha, that's interesting, I was spending half an hour trying to understand the 42 "bugs" it found and they didn't make any sense. disabling asm does removes most of them(when will clang tools will start understanding asm :( )

real-or-random:
> Fwiw, I've used scan-build a couple of times on this code base over the years so I doubt it needs clang 10.

You're right, and I was somewhat confused because I remember you showed me output of a toolthat looked very similar a while ago. I was totally confusing this with GCC 10, which has a new analyzer since v10 (`-fanalyzer`)... Anyway.

real-or-random:
> Ha, that's interesting, I was spending half an hour trying to understand the 42 "bugs" it found and they didn't make any sense. disabling asm does removes most of them(when will clang tools will start understanding asm :( )

Even with asm disabled, it shows 20 positive for me after the PR. So it could be interesting to look at all 42 ones and see if they are really false positives.

elichai:
> > Ha, that's interesting, I was spending half an hour trying to understand the 42 "bugs" it found and they didn't make any sense. disabling asm does removes most of them(when will clang tools will start understanding asm :( )
> 
> Even with asm disabled, it shows 20 positive for me after the PR. So it could be interesting to look at all 42 ones and see if they are really false positives.

yep, I'm trying to go over them right now (some are just we didn't check malloc returning NULL in the tests)

elichai:
tACK 2e1b9e0458317d03b682c1f5dd63aedb52c86b04
Can confirm that this PR made the following line disappear(in scan-view):
`Logic error | Dereference of null pointer | tests.c | counting_illegal_callback_fn | 43 | 82 `
Also the change makes sense (the alternative would be to check that the count wasn't incremented after each malloc)

src/tests.c
```diff
@@ -182,8 +182,10 @@ void run_context_tests(int use_prealloc) {
     ecount2 = 10;
     secp256k1_context_set_illegal_callback(vrfy, counting_illegal_callback_fn, &ecount);
     secp256k1_context_set_illegal_callback(sign, counting_illegal_callback_fn, &ecount2);
-    secp256k1_context_set_error_callback(sign, counting_illegal_callback_fn, NULL);
-    CHECK(vrfy->error_callback.fn != sign->error_callback.fn);
+    /* set error callback (to a function that still aborts in case malloc() fails in secp256k1_context_clone() below) */
+    secp256k1_context_set_error_callback(sign, secp256k1_default_illegal_callback_fn, NULL);
```
elichai, line 186: I see now that in other places we do `secp256k1_context_set_error_callback(sign, NULL, NULL);` instead of explicitly specifying the function

gmaxwell: NULL sets it to the default.

elichai: I missed the fact that this isn't actually the default function. but both will abort, so I think that was by mistake?

gmaxwell: No, this is intentional. The purpose of the next line is to verify that the setter actually sets (and isn't a no-op). So it needs to be set to something different to perform that test.

Some additional tests added after the initial setter-test were introduced call the clone function so they actually need the context to be functional.  They still logically belong using this context, however, because they're calling clone to verify (among other things) that the clone copies the configured setter function.

Your comments suggest that this is unclear enough that it deserves an explicit comment.  /* This uses secp256k1_default_illegal_callback_fn instead of secp256k1_default_error_callback_fn because it needs to invoke it with non-default function as the argument for the following setter and clone tests, but the configured function should still work correctly should malloc happen to get called during the tests. */

elichai: 3 lines above we setted the error callback to be `counting_illegal_callback_fn` so I do think that setting back to the default still tests the same thing.
But it doesn't really matter, so you can ignore my comment

elichai: I see now that I keep getting confused between error and illegal :) your response makes sense. thanks.

real-or-random: No three lines above it sets the illegal cb, not the error cb.

real-or-random: so you stick to your ACK?

elichai: yep. it was a nit anyway, and it turned out as a wrong nit :)
(but feel free to add a comment similar to what @gmaxwell was suggesting)

PR #779, opened 2020-07-29
travis: Fix argument quoting for ./configure

real-or-random:
When $USE_HOST or $EXTRAFLAGS are empty, we pass (due to quoting) an
empty string as a parameter to ./configure, which then believes we want
to use a deprecated syntax for specifing a host or a target and yells at us:
> configure: WARNING: you should use --build, --host, --target

The fixes are:
 - $EXTRAFLAGS could contain multiple flags and should not be quoted at all.
 - We can get rid of $USE_HOST by specifying --host="$HOST" directly.

### DIFF
--- a/contrib/travis.sh
+++ b/contrib/travis.sh
@@ -3,10 +3,6 @@
 set -e
 set -x
 
-if [ -n "$HOST" ]
-then
-    export USE_HOST="--host=$HOST"
-fi
 if [ "$HOST" = "i686-linux-gnu" ]
 then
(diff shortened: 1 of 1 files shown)


elichai:
so now it will pass `--host=`? does autotools handles that correctly?

real-or-random:
> so now it will pass `--host=`? does autotools handles that correctly?

Yep I checked this. You can also pass all of the others empty and it's like it was not passed at all, e.g., `--enable-endomormisphm=`.

PR #782, opened 2020-07-30
Check if variable=yes instead of if var is set in travis.sh

elichai:
Right now travis.sh checks if  `VALGRIND` and `CTIMETEST` are set or not, but in travis.yml we explictly set them to `yes` so I think it's more intuitive to check if they're equal to `yes` such that setting `CTIMETEST=no`(or anything other than "yes") will disable the test. 

pointed out by @real-or-random here: https://github.com/bitcoin-core/secp256k1/pull/696#discussion_r462053453


real-or-random:
I think that's an improvement but then we should do this for `BENCH` too.

real-or-random:
> I think that's an improvement but then we should do this for `BENCH` too.

Should we change the corresponding flags to `no` then, e.g., here? 
https://github.com/bitcoin-core/secp256k1/blob/master/.travis.yml#L34

elichai:
We could, but unlike `BIGNUM=no`  which passes to configure and it will check that it contains something valid, here we only check if it's yes or not, we don't check if it's `no` or something else.
but maybe it's still worth it for consistency and readability

real-or-random:
> but maybe it's still worth it for consistency and readability

That was my thought.

real-or-random:
@elichai do you want to make that change?

elichai:
> @elichai do you want to make that change?

Sorry, forgot about this. pushed now.

real-or-random:
ACK 34debf7a6d36bbd9a52e68e079ddfc446faf5bef

Would be nice to get a second ACK, then we can merge this before doing the travis changes of #813

real-or-random (approved):

jonasnick (approved):
ACK 34debf7a6d36bbd9a52e68e079ddfc446faf5bef

PR #783, opened 2020-07-30
Make the public API docs more consistent and explicit

elichai:
I went over the public API and added missing explanations on when a pointer can be null and when it cannot,
and added some missing checks for null ctx and null pubkey pointers.
 
Open questions IMHO:
1. Can `secp256k1_context_create` return NULL? right now it could return null if you replaced the callbacks at compile time to ones that do return(unlike the default ones which never return).
2. Related to the first, should we document that the callbacks should never return? (in the tests we use returning callbacks but we can violate our own API) right now we say the following:

> After this callback returns, anything may happen, including crashing.

Is this enough to document answer `no` for the first question and just saying that if the callback returned then you violated the API so `secp256k1_context_create` can return NULL even though it is promised not to?
Right now we AFAICT we never check if it returns null

Another nit I'm not sure about is wording `(does nothing if NULL)`/`(ignored if NULL)`/`(can be NULL)` 

More missing docs:
1. Documenting the `data` argument to the default nonce functions

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -226,7 +226,7 @@ SECP256K1_API secp256k1_context* secp256k1_context_create(
  *  memory allocation entirely, see the functions in secp256k1_preallocated.h.
  *
  *  Returns: a newly created context object.
- *  Args:    ctx: an existing context to copy (cannot be NULL)
+ *  Args:    ctx: an existing context to copy
  */
 SECP256K1_API secp256k1_context* secp256k1_context_clone(
     const secp256k1_context* ctx
@@ -278,11 +278,11 @@ SECP256K1_API void secp256k1_context_destroy(
  *  fails. In this case, the corresponding default handler will be called with
  *  the data pointer argument set to NULL.
  *
- *  Args: ctx:  an existing context object (cannot be NULL)
+ *  Args: ctx:  an existing context object.
  *  In:   fun:  a pointer to a function to call when an illegal argument is
  *              passed to the API, taking a message and an opaque pointer.
  *              (NULL restores the default handler.)
- *        data: the opaque pointer to pass to fun above.
+ *        data: the opaque pointer to pass to fun above, must be NULL for the default handler.
  *
  *  See also secp256k1_context_set_error_callback.
  */
@@ -302,12 +302,12 @@ SECP256K1_API void secp256k1_context_set_illegal_callback(
  *  for that). After this callback returns, anything may happen, including
  *  crashing.
  *
- *  Args: ctx:  an existing context object (cannot be NULL)
+ *  Args: ctx:  an existing context object.
  *  In:   fun:  a pointer to a function to call when an internal error occurs,
  *              taking a message and an opaque pointer (NULL restores the
  *              default handler, see secp256k1_context_set_illegal_callback
  *              for details).
- *        data: the opaque pointer to pass to fun above.
+ *        data: the opaque pointer to pass to fun above, must be NULL for the default handler.
  *
  *  See also secp256k1_context_set_illegal_callback.
  */
@@ -320,7 +320,7 @@ SECP256K1_API void secp256k1_context_set_error_callback(
 /** Create a secp256k1 scratch space object.
  *
  *  Returns: a newly created scratch space.
- *  Args: ctx:  an existing context object (cannot be NULL)
+ *  Args: ctx:  an existing context object.
  *  In:   size: amount of memory to be available as scratch space. Some extra
  *              (<100 bytes) will be allocated for extra accounting.
  */
@@ -480,8 +480,8 @@ SECP256K1_API int secp256k1_ecdsa_signature_serialize_compact(
  *  Returns: 1: correct signature
  *           0: incorrect or unparseable signature
  *  Args:    ctx:       a secp256k1 context object, initialized for verification.
- *  In:      sig:       the signature being verified (cannot be NULL)
- *           msghash32: the 32-byte message hash being verified (cannot be NULL).
+ *  In:      sig:       the signature being verified.
+ *           msghash32: the 32-byte message hash being verified.
  *                      The verifier must make sure to apply a cryptographic
  *                      hash function to the message by itself and not accept an
  *                      msghash32 value directly. Otherwise, it would be easy to
@@ -489,7 +489,7 @@ SECP256K1_API int secp256k1_ecdsa_signature_serialize_compact(
  *                      secret key. See also
  *                      https://bitcoin.stackexchange.com/a/81116/35586 for more
  *                      background on this topic.
- *           pubkey:    pointer to an initialized public key to verify with (cannot be NULL)
+ *           pubkey:    pointer to an initialized public key to verify with.
  *
  * To avoid accepting malleable signatures, only ECDSA signatures in lower-S
  * form are accepted.
@@ -515,8 +515,7 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ecdsa_verify(
  *                or copy if the input was already normalized. (can be NULL if
  *                you're only interested in whether the input was already
  *                normalized).
- *  In:   sigin:  a pointer to a signature to check/normalize (cannot be NULL,
- *                can be identical to sigout)
+ *  In:   sigin:  a pointer to a signature to check/normalize (can be identical to sigout)
  *
  *  With ECDSA a third-party can forge a second distinct signature of the same
  *  message, given a single initial signature, but without knowing the key. This
@@ -568,12 +567,16 @@ SECP256K1_API extern const secp256k1_nonce_function secp256k1_nonce_function_def
  *
  *  Returns: 1: signature created
  *           0: the nonce generation function failed, or the secret key was invalid.
- *  Args:    ctx:       pointer to a context object, initialized for signing (cannot be NULL)
- *  Out:     sig:       pointer to an array where the signature will be placed (cannot be NULL)
- *  In:      msghash32: the 32-byte message hash being signed (cannot be NULL)
- *           seckey:    pointer to a 32-byte secret key (cannot be NULL)
- *           noncefp:   pointer to a nonce generation function. If NULL, secp256k1_nonce_function_default is used
- *           ndata:     pointer to arbitrary data used by the nonce generation function (can be NULL)
+ *  Args:    ctx:       pointer to a context object, initialized for signing.
+ *  Out:     sig:       pointer to an array where the signature will be placed.
+ *  In:      msghash32: the 32-byte message hash being signed.
+ *           seckey:    pointer to a 32-byte secret key.
+ *           noncefp:   pointer to a nonce generation function. If NULL,
+ *                      secp256k1_nonce_function_default is used.
+ *           ndata:     pointer to arbitrary data used by the nonce generation function
+ *                      (can be NULL). If it is non-NULL and
+ *                      secp256k1_nonce_function_default is used, then ndata must be a
+ *                      pointer to 32-bytes of additional data.
  *
  * The created signature is always in lower-S form. See
  * secp256k1_ecdsa_signature_normalize for more details.
@@ -596,8 +599,8 @@ SECP256K1_API int secp256k1_ecdsa_sign(
  *
  *  Returns: 1: secret key is valid
  *           0: secret key is invalid
- *  Args:    ctx: pointer to a context object (cannot be NULL)
- *  In:      seckey: pointer to a 32-byte secret key (cannot be NULL)
+ *  Args:    ctx: pointer to a context object.
+ *  In:      seckey: pointer to a 32-byte secret key.
  */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_seckey_verify(
     const secp256k1_context* ctx,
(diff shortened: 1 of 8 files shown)


include/secp256k1.h
```diff
@@ -229,7 +229,7 @@ SECP256K1_API secp256k1_context* secp256k1_context_clone(
  *  be used instead.
  *
  *  Args:   ctx: an existing context to destroy, constructed using
- *               secp256k1_context_create or secp256k1_context_clone
+ *               secp256k1_context_create or secp256k1_context_clone (does nothing if NULL)
```
real-or-random, line 232: I don't think that we want to guarantee this. As I understand it, we don't define any behavior if you pass NULL, so it's UB (we're just nice currently and do nothing :))

include/secp256k1.h
```diff
@@ -268,7 +268,7 @@ SECP256K1_API void secp256k1_context_destroy(
  *  In:   fun:  a pointer to a function to call when an illegal argument is
  *              passed to the API, taking a message and an opaque pointer.
  *              (NULL restores the default handler.)
- *        data: the opaque pointer to pass to fun above.
+ *        data: the opaque pointer to pass to fun above (can be NULL for default handler)
```
real-or-random, line 271: I think we could say "must be NULL" for default handler because the default handler does not expect data. (Again, nothing bad happens if you don't obey.)

include/secp256k1.h
```diff
@@ -318,8 +318,8 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT secp256k1_scratch_space* secp256k1_sc
 /** Destroy a secp256k1 scratch space.
  *
  *  The pointer may not be used afterwards.
- *  Args:       ctx: a secp256k1 context object.
- *          scratch: space to destroy
+ *  Args:       ctx: a secp256k1 context object (cannot be NULL)
```
real-or-random, line 321: Well NULL is not a secp256k1, not sure if we need to say this everywhere.

But apparently we say this in some places, so we could make it consistent. But personally I'd prefer to omit it consistently.

include/secp256k1.h
```diff
@@ -318,8 +318,8 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT secp256k1_scratch_space* secp256k1_sc
...
+ *          scratch: space to destroy (does nothing if NULL)
```
real-or-random, line 322: same as with the context, not sure why we would want to guarantee this.

include/secp256k1.h
elichai: yeah I started by omitting it when it's obvious that it cannot be null, but then other places didn't omit so I settled on being explicit *everywhere*

tnawanna (approved):

sipa:
I think this is overkill, and in some cases makes the public description of the API commit to promises that we may not want to keep.

My preference would be to state once that unless explicitly indicated, pointer arguments cannot be NULL. What should be document is whether or not an argument is *intentionally optional* (as in: it still does something useful if the specified argument is NULL). Optional arguments are passed as NULL. Others are passed as non-NULL pointers to objects of the specified type.

jonasnick (commented):
Agree with @sipa, just documenting which arguments are optional is better. As far as wording goes, it'd make sense to state what happens if the argument is NULL. If applicable, something like `Ignored if NULL` (without parantheses preferably).

> Can secp256k1_context_create return NULL?
> Related to the first, should we document that the callbacks should never return?

Even if we mention that illegal callbacks should never return, `context_create` can still return NULL. That happens for example if malloc fails and the error callback is called. This seems to be already mentioned in the docs:

> When this [error] callback is triggered, the API function called is guaranteed not to cause a crash, though its return value and output arguments are undefined.

include/secp256k1.h
```diff
@@ -394,10 +396,10 @@ SECP256K1_API int secp256k1_ecdsa_signature_parse_compact(
...
  *
  *  Returns: 1 when the signature could be parsed, 0 otherwise.
- *  Args: ctx:      a secp256k1 context object
- *  Out:  sig:      a pointer to a signature object
- *  In:   input:    a pointer to the signature to be parsed
- *        inputlen: the length of the array pointed to be input
+ *  Args: ctx:      a secp256k1 context object.
+ *  Out:  sig:      a pointer to a signature object.
+ *  In:   input:    a pointer to the signature to be parsed.
+ *        inputlen: the length of the array pointed to be input.
```
jonasnick, line 402: Adding periods for every arguments seems a bit unnecessary. So far I've tried to follow the rule to add a period only if it follows an actual sentence (obj + verb).

include/secp256k1_extrakeys.h
```diff
@@ -72,13 +70,13 @@ SECP256K1_API int secp256k1_xonly_pubkey_serialize(
...
- *  Args:         ctx: pointer to a context object (cannot be NULL)
+ *  Args:         ctx: pointer to a context object.
  *  Out: xonly_pubkey: pointer to an x-only public key object for placing the
- *                     converted public key (cannot be NULL)
- *          pk_parity: pointer to an integer that will be set to 1 if the point
- *                     encoded by xonly_pubkey is the negation of the pubkey and
- *                     set to 0 otherwise. (can be NULL)
- *  In:        pubkey: pointer to a public key that is converted (cannot be NULL)
+ *                     converted public key.
+ *          pk_parity: Optional(if NULL): pointer to an integer that
```
jonasnick, line 76: How about `Ignored if NULL. Otherwise, pointer ....`

include/secp256k1_schnorrsig.h
```diff
@@ -69,15 +69,14 @@ SECP256K1_API extern const secp256k1_nonce_function_hardened secp256k1_nonce_fun
...
+ *  Out:   sig64: pointer to a 64-byte array to store the serialized signature.
+ *  In:    msg32: the 32-byte message being signed.
+ *       keypair: pointer to an initialized keypair.
  *       noncefp: pointer to a nonce generation function. If NULL, secp256k1_nonce_function_bip340 is used
  *         ndata: pointer to arbitrary data used by the nonce generation
- *                function (can be NULL). If it is non-NULL and
- *                secp256k1_nonce_function_bip340 is used, then ndata must be a
- *                pointer to 32-byte auxiliary randomness as per BIP-340.
+ *                function for secp256k1_nonce_function_bip340 it can either be NULL or
+ *                non-NULL and then ndata must be a pointer to 32-byte auxiliary randomness as per BIP-340.
```
jonasnick, line 79: How about

pointer to arbitrary data used by the nonce generation function. secp256k1_nonce_function_bip340 ignores ndata if it is NULL, otherwise it must be a pointer to 32-byte auxiliary randomness as per BIP-340.


Also I think we should try to keep the line width <= 100 chars.

jonasnick (commented):
Really good to get rid of the "(cannot be NULL)". There's one more of those in `secp256k1_ecdsa_signature_normalize` (see `sigin` argument) and more in `contrib/lax_der_privatekey_parsing.h`.

include/secp256k1.h
elichai: I mostly wanted consistency and thought that will be the less controversial way, but I'll remove that change

real-or-random:
Maybe it's will be a good idea to write the requirements on the context flags in more structured way. At the moment we say for example "initialized for verification". 

I believe it makes things easier for the user if we'd say "created with flag SECP256K1_CONTEXT_SIGN" or similar, and if this somehow stands out such that it can be parsed easily (be the human eye).

jonasnick:
Needs rebase

include/secp256k1.h
```diff
@@ -751,18 +750,18 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_context_randomize(
...
  *  Returns: 1: the sum of the public keys is valid.
  *           0: the sum of the public keys is not valid.
- *  Args:   ctx:        pointer to a context object
- *  Out:    out:        pointer to a public key object for placing the resulting public key
- *                      (cannot be NULL)
- *  In:     ins:        pointer to array of pointers to public keys (cannot be NULL)
- *          n:          the number of public keys to add together (must be at least 1)
+ *  Args:   ctx:        pointer to a context object.
+ *  Out:    out:        pointer to a public key object for placing the resulting public key.
+ *                     .
```
ariard, line 755: nit: i believe this dot is redundant

include/secp256k1.h
```diff
@@ -229,7 +231,7 @@ SECP256K1_API secp256k1_context* secp256k1_context_clone(
  *  be used instead.
  *
  *  Args:   ctx: an existing context to destroy, constructed using
- *               secp256k1_context_create or secp256k1_context_clone
+ *               secp256k1_context_create or secp256k1_context_clone.
  */
 SECP256K1_API void secp256k1_context_destroy(
```
ariard, line 236: Does `secp256k1_context_destroy`'s `ctx` deserves its own `SECP256K1_ARG_NONNULL` annotation?

edit: Have a look on `secp256k1_ecdsa_signature_normalize`'s second arg, missing too?
re-edit: `secp256k1_context_randomize`'s second arg?

include/secp256k1_preallocated.h
```diff
@@ -115,7 +115,7 @@ SECP256K1_API secp256k1_context* secp256k1_context_preallocated_clone(
  *
  *  Args:   ctx: an existing context to destroy, constructed using
  *               secp256k1_context_preallocated_create or
- *               secp256k1_context_preallocated_clone (cannot be NULL)
+ *               secp256k1_context_preallocated_clone.
  */
 SECP256K1_API void secp256k1_context_preallocated_destroy(
```
ariard, line 120: Same, missing `SECP256K1_ARG_NONNULL` on `secp256k1_context_preallocated_destroy`'s first arg ?

include/secp256k1_recovery.h
```diff
@@ -55,10 +56,10 @@ SECP256K1_API int secp256k1_ecdsa_recoverable_signature_convert(
...
- *  Out:  output64: a pointer to a 64-byte array of the compact signature (cannot be NULL)
- *        recid:    a pointer to an integer to hold the recovery id (can be NULL).
- *  In:   sig:      a pointer to an initialized signature object (cannot be NULL)
+ *  Args: ctx:      a secp256k1 context object.
+ *  Out:  output64: a pointer to a 64-byte array of the compact signature.
+ *        recid:    Optional(if NULL): a pointer to an integer to hold the recovery id.
+ *  In:   sig:      a pointer to an initialized signature object.
  */
 SECP256K1_API int secp256k1_ecdsa_recoverable_signature_serialize_compact(
     const secp256k1_context* ctx,
```
ariard, line 65: Well maybe here you need to drop the annotation for third arg?

ariard (commented):
Otherwise good to me, let me know if I'm completely astray w.r.t to `SECP256K1_ARG_NONNULL`

include/secp256k1.h
elichai: Good question if it shouldn't be null, right now it will work if it's not but we don't promise that so I think it should have.

As for the rest:
`secp256k1_ecdsa_signature_normalize` 2nd argument can be NULL if you just want to check for normalization without normalizing.
`secp256k1_context_randomize` you can pass NULL as the 2nd argument to reset to the initial state.

include/secp256k1_recovery.h
elichai: Good find! Thanks

include/secp256k1_preallocated.h
elichai: Added, Thanks!

include/secp256k1_recovery.h
elichai: I was wrong for writing `recid` is optional, so I removed that change. Thanks!

include/secp256k1_schnorrsig.h
elichai: That's better, Thanks!

include/secp256k1_extrakeys.h
elichai: That's better, Thanks!

include/secp256k1.h
elichai: I removed all the changes that simply added periods

elichai:
Rebased and fixed comments

include/secp256k1.h
jonasnick: I don't think it's worth changing the API of `context_destroy` and `preallocated_destroy` given that it is fine to give NULL pointers to `free()`.

jonasnick (commented):
ACK mod nit

include/secp256k1.h
ariard: Good point, though i don't know how "if ptr NULL, no operations is performed" guarantee holds beyond standardized *nix platforms.

ariard:
ACK adec5a16

Sorry for my annotation call around `context_destroy`/`preallocated_destroy` if you think it's redundant, feel free to drop it.

include/secp256k1.h
jonasnick: Probably best to mention this behavior in the doc ("Ignored if NULL" or something).

elichai: @jonasnick See https://github.com/bitcoin-core/secp256k1/pull/783#discussion_r462912364

jonasnick: If we don't mention this, we will have the exact same discussion again at some point. But fine either way imo.

jonasnick: @ariard that behavior is part of the C standard (http://port70.net/~nsz/c/c89/c89-draft.html#4.10.3.2)

elichai: Yeah, I wanted to make that promise too, but not too important (as long as it's explicitly either NON_NULL or allowed to be null, I prefer not to leave it in a limbo)

ariard: > @ariard that behavior is part of the C standard (http://port70.net/~nsz/c/c89/c89-draft.html#4.10.3.2)

Ah right, so "allowed to be null" recalling C standard?

real-or-random: @ariard Not sure if we're talking past each other. Sure the current `free(NULL)` implementation is fine. I was just pointing out that adding the promise that `secp256k1_context_destroy(NULL)` is a no-op will be an API change and I don't see a reason to make this change.

jonasnick: I still think it would be better to remove the `SECP256K1_ARG_NONNULL` because it is a backwards-incompatible API change.

jonasnick: Just to clarify, it's an API change since before 0881633d we didn't force that pointer args must not be NULL if not mentioned otherwise. But here we never mentioned "cannot be NULL" or similar.

jonasnick (commented):
ACK adec5a16383f1704d80d7c767b2a65d9221cee08

real-or-random:
> @real-or-random pointed out that while there wasn't a "cannot be NULL", an _existing_ context was required. So I think it's ok as is now.

I agree.

PR #787, opened 2020-08-07
Use preprocessor macros instead of autoconf to detect endianness

real-or-random:
This does not fix any particular issue but it's preferable to not
rely on autoconf. This avoids endianness mess for users on BE hosts
if they use their build without autoconf.

The macros are carefully written to err on the side of the caution,
e.g., we #error if the user manually configures a different endianness
than what we detect.

Supersedes #770 .

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -179,6 +179,20 @@ static SECP256K1_INLINE void *manual_alloc(void** prealloc_ptr, size_t alloc_siz
 SECP256K1_GNUC_EXT typedef unsigned __int128 uint128_t;
 #endif
 
+#if defined(__BYTE_ORDER__)
+# if defined(__ORDER_LITTLE_ENDIAN__) && __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__ && !defined(SECP256K1_LITTLE_ENDIAN)
+#  define SECP256K1_LITTLE_ENDIAN
+# elif defined(__ORDER_BIG_ENDIAN__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__ && !defined(SECP256K1_BIG_ENDIAN)
+#  define SECP256K1_BIG_ENDIAN
+# endif
+#endif
+#if defined(_MSC_VER) && defined(_WIN32) && !defined(SECP256K1_LITTLE_ENDIAN)
+# define SECP256K1_LITTLE_ENDIAN
+#endif
+#if defined(SECP256K1_LITTLE_ENDIAN) == defined(SECP256K1_BIG_ENDIAN)
+# error Please make sure that either SECP256K1_LITTLE_ENDIAN or SECP256K1_BIG_ENDIAN is set, see src/util.h.
+#endif
+
 /* Zero memory if flag == 1. Flag must be 0 or 1. Constant time. */
 static SECP256K1_INLINE void memczero(void *s, size_t len, int flag) {
     unsigned char *p = (unsigned char *)s;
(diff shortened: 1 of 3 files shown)


gmaxwell:
I like this approach.

gmaxwell:
Travis has a zero log apparently unexplained failure on s390 ... the only test that really mattered for this pr.  :)

I tested it by hand on s390 and it works. sooo ?!

sipa:
I checked this PR a few hours ago, and all tests had finished, except the s390x one, which hadn't started.

So I think there may be some provisioning issue that prevents travis from assigning an s390x worker, and it somehow timed out?

JeremyRubin:
This is also the case on Bitcoin Core right now too :/

real-or-random:
Yeah, first I felt brave when I squashed https://github.com/bitcoin-core/secp256k1/pull/696 and then merged it immediately. And then I got a notification from Travis that I broke master... :/ But yes, it's an issue on their side...

gmaxwell:
ACK

sipa:
Opened a PR for Bitcoin Core https://github.com/bitcoin/bitcoin/pull/19695 to test if this may impact any of the systems tested by its CI (which includes an MSVC based one).

gmaxwell:
It's _MSC_VER not _MSVC_VER  (thanks to sipa abusing bitcoin's CI setup to test this.)

real-or-random:
Fixed. @sipa can you retest on Core?

sipa:
@real-or-random Updated it.

real-or-random:
> @real-or-random Updated it.

It works. :tada:

src/util.h
```diff
@@ -179,6 +179,20 @@ static SECP256K1_INLINE void *manual_alloc(void** prealloc_ptr, size_t alloc_siz
...
+#  define SECP256K1_LITTLE_ENDIAN
+# elif defined(__ORDER_BIG_ENDIAN__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__ && !defined(SECP256K1_BIG_ENDIAN)
+#  define SECP256K1_BIG_ENDIAN
+# endif
+#endif
+#if defined(_MSC_VER) && defined(_WIN32) && !defined(SECP256K1_LITTLE_ENDIAN)
+# define SECP256K1_LITTLE_ENDIAN
+#endif
+#if defined(SECP256K1_LITTLE_ENDIAN) == defined(SECP256K1_BIG_ENDIAN)
+# error Please make sure that either SECP256K1_LITTLE_ENDIAN or SECP256K1_BIG_ENDIAN is set, see src/util.h.
```
sipa, line 193: I never knew that you could put an error message after `#error` without quotes around it.

In retrospect, not sure what I'd expect it to do. Give an error?

sipa:
ACK 0dccf98a21beb245f6cd9ed76fb7368529df09c7

gmaxwell:
ACK 0dccf98a21beb245f6cd9ed76fb7368529df09c7

gmaxwell:
This broke the build on GCC 2.95 on Debian i386 for me and on Haiku (BeOS).

sipa:
@gmaxwell Can you give the list of preprocessor defines on those?

gmaxwell:
```
    $ gcc -dM -E -x c -
    #define __linux__ 1 
    #define linux 1 
    #define __i386__ 1 
    #define __i386 1 
    #define __i686 1 
    #define __GNUC_MINOR__ 95 
    #define pentiumpro 1 
    #define __pentiumpro 1 
    #define i386 1 
    #define i686 1 
    #define __pentiumpro__ 1 
    #define __unix 1 
    #define __unix__ 1 
    #define __i686__ 1 
    #define __GNUC__ 2 
    #define __linux 1 
    #define __ELF__ 1 
    #define unix 1 
```

From the haiku I can't easily copy and paste, but it includes, __i586 __i386 _X86_ __i386__ __i586__ __pentium __pentium__

sipa:
@gmaxwell See #799

Issue #788, opened 2020-08-07
[Question] secp256k1_ecmult_const `k` times Infinity


lontivero:
Question: if `P` is the point at infinity and `k` is a scalar in the range `0 < k < N`, is `kP` also the point at infinity? 

In other words: should this test pass?
```c++
void ecmult_const_mult_inf_one(void) {
    secp256k1_scalar one = SECP256K1_SCALAR_CONST(0, 0, 0, 0, 0, 0, 0, 1);
    secp256k1_gej res1;
    secp256k1_ge res2;
    secp256k1_ge point;

    secp256k1_ge_set_infinity(&point);
    secp256k1_ecmult_const(&res1, &point, &one, 256);
    secp256k1_ge_set_gej(&res2, &res1);
    CHECK(secp256k1_ge_is_infinity(&res2));
}
```

sipa:
Yes, the point at infinity corresponds to the scalar 0, so multiplying it with anything cannot change it.

real-or-random:
In particular a multiplication with 1 should never change anything. So much for the math.

But I'm not sure whether `secp256k1_ecmult_const` is built to work with point with the point at infinity. (I believe no.) It's an internal function and it's currently only used in the ECDH code and there the point can't be infinity because the point was deserialized and we don't a serialization for infinity. If `secp256k1_ecmult_const` is not to be used with infinity, then we should probably document it.

lontivero:
Thank you both. I asked this here because I am tracking a problem from NBitcoin.Secp256k1 library (secp256k1 dotnet porting). `secp256k1_ecmult_const` is internal and should not be used with infinity. It is clear.

PR #789, opened 2020-08-07
Document that infinity must not be passed to ecmult_const

real-or-random:


real-or-random:
Related to #788.

gmaxwell:
What happens if you pass in a point at infinity? :)

gmaxwell:
Ah. Storage type doesn't currently encode infinities so the to precomputed tables have issues.  Other than that it looks like it takes a smaller change to fix infinity support, most of which is obviously more correct than not doing it.

gmaxwell:
Unless I'm mistaken, ecmult isn't constant time with respect to the point input.  If so, it could just return infinity in this case. It would be a smaller diff and less weird-surface to handle for future users of the function.

real-or-random:
> Unless I'm mistaken, ecmult isn't constant time with respect to the point input.

Oh that's a good point. I took the easy way because with constant-timeness in mind but yes, it's constant-time in the scalar only.

PR #794, opened 2020-08-09
Disable s390x again

sipa:
It seems Travis' s390x doesn't work reliably enough. Disable it again for now.


real-or-random:
ACK 124a757424af954420bc4b561f48aebeaec62f95 (if travis passes)

real-or-random:
Ok I've just restarted the errored job on master and it's working now. Let's hope it stays that way, otherwise we can resurrect this PR.

Issue #792, opened 2020-08-08
Document / assert non-portable assumptions


real-or-random:
> I think it's reasonable to require a two's complement implementation (we could verify it in a configure script, and in unit tests)

If you really have some exotic system, then you might not run ./configure or the tests either. I think the safest way is to use some C static assert hack e.g., https://stackoverflow.com/a/56742411. These looks very ugly but they work and the ugliness is hidden behind a macro. We should also check other simple assumptions such as `sizeof(int) == 4` etc. Alternatively, we could perform some self-tests at context creation time, and call the error callback if they fail. The simple ones will be optimized away by the compiler, and we should have some kind of self-test anyway if we want to implement a SHA256 override (see the discussion in #702).

_Originally posted by @real-or-random in https://github.com/bitcoin-core/secp256k1/pull/767#issuecomment-670979922_

@gmaxwell notes there that the static assert hack may not be the best idea.

real-or-random:
Here are some assumptions that come to my mind immediately:
 - Right shifts https://github.com/bitcoin-core/secp256k1/blob/d567b779fe446fd18820a9d2968ecb703c8dea19/src/ecmult_const_impl.h#L19
 - size of size_t, see for example https://github.com/bitcoin-core/secp256k1/blob/770b3dcd6f1edb0a6355a55d92b7bea1e9524042/src/ecmult_impl.h#L51 (also in contrib, see https://github.com/bitcoin/bitcoin/pull/10657/files)
 - `sizeof(int) <= 4`, otherwise `uint32_t` x `uint32_t` multiplication could result in signed overflow (This is even absurder when you state it this way.)
 - I would not be surprised if we assume `sizeof(int) >= 4` somewhere.
 - if we're pedantic then maybe also `CHAR_BITS`

This is a shot in the dark but @roconnor-blockstream  do you know some more?

gmaxwell:
> size of size_t, see for example 

Might be better to make that config macro get size_t's bits and base the limit on that.  Otherwise its reducing portability for unexposed configurability.

real-or-random:
> > size of size_t, see for example
> 
> Might be better to make that config macro get size_t's bits and base the limit on that. Otherwise its reducing portability for unexposed configurability.

Sorry, I can't follow. Which macro?

gmaxwell:
[...]

PR #798, opened 2020-08-13
Check assumptions on integer implementation at compile time

sipa:
A compile-time check is implemented in a new `src/assumptions.h` which verifies several aspects that are implementation-defined in C:
* size of bytes
* conversion between unsigned and (negative) signed types
* right-shifts of negative signed types.

### DIFF
--- a/src/assumptions.h
+++ b/src/assumptions.h
@@ -1,0 +1,74 @@
+/**********************************************************************
+ * Copyright (c) 2020 Pieter Wuille                                   *
+ * Distributed under the MIT software license, see the accompanying   *
+ * file COPYING or http://www.opensource.org/licenses/mit-license.php.*
+ **********************************************************************/
+
+#ifndef SECP256K1_ASSUMPTIONS_H
+#define SECP256K1_ASSUMPTIONS_H
+
+#include "util.h"
+
+/* This library, like most software, relies on a number of compiler implementation defined (but not undefined)
+   behaviours. Although the behaviours we require are essentially universal we test them specifically here to
+   reduce the odds of experiencing an unwelcome surprise.
+*/
+
+struct secp256k1_assumption_checker {
+    /* This uses a trick to implement a static assertion in C89: a type with an array of negative size is not
+       allowed. */
+    int dummy_array[(
+        /* Bytes are 8 bits. */
+        CHAR_BIT == 8 &&
+
+        /* Conversions from unsigned to signed outside of the bounds of the signed type are
+           implementation-defined. Verify that they function as reinterpreting the lower
+           bits of the input in two's complement notation. Do this for conversions:
+           - from uint(N)_t to int(N)_t with negative result
+           - from uint(2N)_t to int(N)_t with negative result
+           - from int(2N)_t to int(N)_t with negative result
+           - from int(2N)_t to int(N)_t with positive result */
+
+        /* To int8_t. */
+        ((int8_t)(uint8_t)0xAB == (int8_t)-(int8_t)0x55) &&
+        ((int8_t)(uint16_t)0xABCD == (int8_t)-(int8_t)0x33) &&
+        ((int8_t)(int16_t)(uint16_t)0xCDEF == (int8_t)(uint8_t)0xEF) &&
+        ((int8_t)(int16_t)(uint16_t)0x9234 == (int8_t)(uint8_t)0x34) &&
+
+        /* To int16_t. */
+        ((int16_t)(uint16_t)0xBCDE == (int16_t)-(int16_t)0x4322) &&
+        ((int16_t)(uint32_t)0xA1B2C3D4 == (int16_t)-(int16_t)0x3C2C) &&
+        ((int16_t)(int32_t)(uint32_t)0xC1D2E3F4 == (int16_t)(uint16_t)0xE3F4) &&
+        ((int16_t)(int32_t)(uint32_t)0x92345678 == (int16_t)(uint16_t)0x5678) &&
+
+        /* To int32_t. */
+        ((int32_t)(uint32_t)0xB2C3D4E5 == (int32_t)-(int32_t)0x4D3C2B1B) &&
+        ((int32_t)(uint64_t)0xA123B456C789D012ULL == (int32_t)-(int32_t)0x38762FEE) &&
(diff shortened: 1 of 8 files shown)


sipa:
It seems fairly easy to turn this into a compile time check, by using (that_expression - 1) as the size of array.

However, maybe we want to include endianness into the check, and I'm not sure that can be done at compile time.

gmaxwell:
> as the size of array.

Var arrays are C99  and "optional" in later standards and not supported by MSVC.

sipa:
@gmaxwell Yes, but the expression evaluates to a constant.

In fact, it has to be constant. A vararray of negative size could compile. So you have to do something like define a struct with a member array (where a vararray isn't allowed).

sipa:
@gmaxwell this seems to work in gcc/clang/icc/msvc: https://godbolt.org/z/6dsar5

gmaxwell:
Yeah, I tested it with GCC 2.95 on i386 and tinyc and it seems to be happy there too.  One downside is that when it fails you get an extremely opaque error.

sipa:
Added conversions from signed to signed, and a check for CHAR_BIT. Anything else? Do we have any unusual assumptions on the size of `int` etc?

real-or-random:
@sipa Are you aware of https://github.com/bitcoin-core/secp256k1/issues/792 ? :P

real-or-random:
> One downside is that when it fails you get an extremely opaque error.

We could use `static_assert` if it's supported by the compiler but I don't think it's worth the additional `#ifdef`s. These errors should be super rare, so it's acceptable that they're somewhat opaque. And it's probably better to have a compiler error instead of a runtime error.


I read up a little bit on these techniques. The issues with VLAs was something else: In some cases this array trick silently didn't do anything when the expression is non-constant, e.g., when you have a compiler like GCC that supports VLAs as an extension and you happen to pass a non-constant value, then it can't be a compiler error even if the size is negative. This hit the Linux kernel for example. 

But what we do here is robust against this failure mode because VLAs are never supported on a file level / global scope. I verified this in godbolt. Even GCC errors out if you pass a non-constant value. (Clang fails at another stage but the message is pretty harsh: `fields must have a constant size: 'variable length array in structure' extension will never be supported` :D) The difference in the Linux kernel is probably that they don't want to collect all the assumptions in a single place and they want a statement instead.

Approach ACK 

Maybe we want to move the assumptions to a separate header like in Bitcoin Core (https://github.com/bitcoin/bitcoin/blob/master/src/compat/assumptions.h). That's cleaner and it's easier to point people to the file, e.g., in the README.

gmaxwell:
A separate file might make it marginally more likely to get a more useful error message (e.g. telling you what file it was in) from more compilers.

sipa:
> A separate file might make it marginally more likely to get a more useful error message (e.g. telling you what file it was in) from more compilers.

That's a good point.

Moved.

gmaxwell:
You might want to a comment above it that says something like:

/* This library, like most software, relies on a number of compiler implementation defined (not undefined) behaviours.  Although the behaviours we require are essentially universal we test them specifically here to reduce the odds of experiencing an unwelcome surprise. */

sipa:
@gmaxwell Added.

src/assumptions.h
```diff
@@ -0,0 +1,74 @@
...
+        /* Bytes are 8 bits. */
+        CHAR_BIT == 8 &&
+
+        /* Conversions from unsigned to signed outside of the bounds of the signed type are
+           implementation-defined. Verify that they function as reinterpreting the lower
+           bits of the input in two's complement notation. Do this for conversions:
+           - from uint(N)_t to int(N)_t with negative result
+           - from uint(2N)_t to int(N)_t with negative result
+           - from int(2N)_t to int(N)_t with negative result
+           - from int(2N)_t to int(N)_t with positivie result */
```
gmaxwell, line 30: positive

sipa: Fixed.

gmaxwell:
Include it in gen_context.c too. Sorry for dribbling changes.

sipa:
Done. Also added to bench_internal.c, tests_exhaustive.c, and valgrind_ctime_test.c.

gmaxwell:
ACK fdb6dbc9e765588280e06d5a7050aa40d7e6c551

src/assumptions.h
```diff
@@ -0,0 +1,74 @@
...
+        ((int64_t)(((uint128_t)0xA1234567B8901234ULL << 64) + 0xC5678901D2345678ULL) == (int64_t)-(int64_t)0x3A9876FE2DCBA988ULL) &&
+        (((int64_t)(int128_t)(((uint128_t)0xB1C2D3E4F5A6B7C8ULL << 64) + 0xD9E0F1A2B3C4D5E6ULL)) == (int64_t)(uint64_t)0xD9E0F1A2B3C4D5E6ULL) &&
+        (((int64_t)(int128_t)(((uint128_t)0xABCDEF0123456789ULL << 64) + 0x0123456789ABCDEFULL)) == (int64_t)(uint64_t)0x0123456789ABCDEFULL) &&
+
+        /* To int128_t. */
+        ((int128_t)(((uint128_t)0xB1234567C8901234ULL << 64) + 0xD5678901E2345678ULL) == (int128_t)(-(int128_t)0x8E1648B3F50E80DCULL * 0x8E1648B3F50E80DDULL + 0x5EA688D5482F9464ULL)) &&
+#endif
+
+        /* Right shift on negative signed values is implementation defined. Verify that it
+           acts as a conversion to unsigned, right shift, and conversion back to signed.
```
real-or-random, line 62: I think this text is wrong. If it acted as a conversion to unsigned and then shift, they would exactly not be sign-extending, no?

sipa: You're right. And the fact that it's different is exactly the reason why we need signed types at all...

sipa: Fixed.

sipa:
@real-or-random Sounds good.

gmaxwell:
ACK 7c068998bac3e4a254d8542458b2068e38fca435

real-or-random:
ACK 7c068998bac3e4a254d8542458b2068e38fca435 code review and tested

Issue #780, opened 2020-07-29
API for context creation


real-or-random:
I'm arguably late to the party but I believe we should either talk about this now (before we do a release) or never:

> Taking a step back, I believe that even given the problems with C, upstream's API is not optimal here. A better API would be to ask the user to provide entropy already in `_context_create`, or pass NULL explicitly to opt out.

_Originally posted by @real-or-random in https://github.com/rust-bitcoin/rust-secp256k1/issues/225#issuecomment-655998672_

Note that this does not need to be a breaking change, we could for example add a new function and deprecate `secp256k1_context_create`.

elichai:
Do we want something like `secp256k1_context_create_randomized` which can accept either a seed or NULL, or should it enforce NULL? (the cost I think is that users that don't have access to an rng will pass either an array of zeros or a constant random seed, I don't think either will decrease the security, and a constant seed is probably better than nothing)

gmaxwell:
@real-or-random  Randomization should be ideally be updated for each signature not just once for the context, which is why it doesn't make sense to just have it as a context creation argument.

Ideally every signature would advance blinding to a new state, but doing that would write to the context, and the existing interface allows you to use the context from multiple threads.

This made more sense when the signing context had large and expensive to create precomputed tables in it rather than the tables being static.  Now the signing context is essentially just the randomness.  It would make a lot more sense in my view to now make signing contexts non-shared, and then they could advance their random state on each use,  and it would also be reasonable to generate randomness at creation time.

The problem I see with that is that the verify contexts still have huge expensive to create precomputed tables, and one probably doesn't want to have to generate a new one all the time... so it makes sense to share them.  This wouldn't be a big deal, except sign & verify contexts exist.

real-or-random:
> @real-or-random Randomization should be ideally be updated for each signature not just once for the context, which is why it doesn't make sense to just have it as a context creation argument.

Well agreed but I didn't think about removing  `_randomize`. We should keep this for sure, but it may still make sense to expect the randomness at context creation time.

> Ideally every signature would advance blinding to a new state, but doing that would write to the context, and the existing interface allows you to use the context from multiple threads.
> 
> This made more sense when the signing context had large and expensive to create precomputed tables in it rather than the tables being static. Now the signing context is essentially just the randomness. It would make a lot more sense in my view to now make signing contexts non-shared, and then they could advance their random state on each use, and it would also be reasonable to generate randomness at creation time.

[...]

PR #799, opened 2020-08-14
Add fallback LE/BE for architectures with known endianness + SHA256 selftest

sipa:
These are all the architecture macros I could find with known endianness. Use those as a fallback when __BYTE_ORDER__ isn't available.

See https://github.com/bitcoin-core/secp256k1/pull/787#issuecomment-673764652

It also adds a SHA256 selftest, so that improperly overriding the endianness detection will be detected at runtime.

### DIFF
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -19,6 +19,7 @@
 #include "eckey_impl.h"
 #include "hash_impl.h"
 #include "scratch_impl.h"
+#include "selftest.h"
 
 #if defined(VALGRIND)
 # include <valgrind/memcheck.h>
@@ -117,6 +118,9 @@ secp256k1_context* secp256k1_context_preallocated_create(void* prealloc, unsigne
     size_t prealloc_size;
     secp256k1_context* ret;
 
+    if (!secp256k1_selftest()) {
+        secp256k1_callback_call(&default_error_callback, "self test failed");
+    }
     VERIFY_CHECK(prealloc != NULL);
     prealloc_size = secp256k1_context_preallocated_size(flags);
     ret = (secp256k1_context*)manual_alloc(&prealloc, sizeof(secp256k1_context), base, prealloc_size);
--- a/src/util.h
+++ b/src/util.h
@@ -176,15 +176,26 @@ static SECP256K1_INLINE void *manual_alloc(void** prealloc_ptr, size_t alloc_siz
 # define SECP256K1_GNUC_EXT
 #endif
 
-#if defined(__BYTE_ORDER__)
-# if defined(__ORDER_LITTLE_ENDIAN__) && __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__ && !defined(SECP256K1_LITTLE_ENDIAN)
+/* If SECP256K1_{LITTLE,BIG}_ENDIAN is not explicitly provided, infer from various other system macros. */
+#if !defined(SECP256K1_LITTLE_ENDIAN) && !defined(SECP256K1_BIG_ENDIAN)
+/* Inspired by https://github.com/rofl0r/endianness.h/blob/9853923246b065a3b52d2c43835f3819a62c7199/endianness.h#L52L73 */
+# if (defined(__BYTE_ORDER__) && defined(__ORDER_LITTLE_ENDIAN__) && __BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__) || \
+     defined(_X86_) || defined(__x86_64__) || defined(__i386__) || \
+     defined(__i486__) || defined(__i586__) || defined(__i686__) || \
+     defined(__MIPSEL) || defined(_MIPSEL) || defined(MIPSEL) || \
+     defined(__ARMEL__) || defined(__AARCH64EL__) || \
+     (defined(__LITTLE_ENDIAN__) && __LITTLE_ENDIAN__ == 1) || \
+     (defined(_LITTLE_ENDIAN) && _LITTLE_ENDIAN == 1) || \
+     defined(_M_IX86) || defined(_M_AMD64) || defined(_M_ARM) /* MSVC */
 #  define SECP256K1_LITTLE_ENDIAN
-# elif defined(__ORDER_BIG_ENDIAN__) && __BYTE_ORDER__ == __ORDER_BIG_ENDIAN__ && !defined(SECP256K1_BIG_ENDIAN)
+# endif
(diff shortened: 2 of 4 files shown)


gmaxwell:
Great work! testing now.

gmaxwell:
I can confirm that this works on the systems that were broken for me before (debian + GCC 2.95 and 32-bit haiku). I'm installing a mips (be) host right now to check what macros it has.

real-or-random:
This could be helpful: https://github.com/rofl0r/endianness.h/blob/master/endianness.h#L52 but we should not overdo it I think .

sipa:
@real-or-random Might as well use something that has had other eyes on it. I've switched to it.

gmaxwell:
ACK 7fde5566a91ffc7fee4606a0a7b6dc6fb5b47444

looks fine, works on the weird things I've tested-- hakiu, gcc 2.95 i386, openwatcom, debian mips.

src/util.h
```diff
@@ -186,6 +186,26 @@ static SECP256K1_INLINE void *manual_alloc(void** prealloc_ptr, size_t alloc_siz
 #if defined(_MSC_VER) && defined(_WIN32) && !defined(SECP256K1_LITTLE_ENDIAN)
 # define SECP256K1_LITTLE_ENDIAN
 #endif
+#if !defined(SECP256K1_LITTLE_ENDIAN) && !defined(SECP256K1_BIG_ENDIAN)
```
real-or-random, line 189: Do we need this `#if` ? It breaks with the idea of the code above, which tries to detect from all available sources instead and then fail below if both are defined, e.g., if a user supplied `#define` is inconsistent with our detection. Seems a tiny bit safer.

gmaxwell: Hm? if the detection is wrong (e.g. because a system header sets something it shouldn't) the user should be able to override.

real-or-random: For example if the user wrongly sets `SECP256K1_BIG_ENDIAN`, then `__x86_64__` will not be checked at all due to this line (if I'm not getting it wrong entirely).

gmaxwell: That is my understanding of the behaviour. If some dumb header has __x86_64__ in it why shouldn't the user should be able to override?

I've seen stuff set things like that just to make stuff compile when some code was erroneously testing __i386__  to determine if the processor was 32-bit.

real-or-random: Ok, we can discuss this but it should be consistent at least. The existing code is different, see my commit message in #787.

> The macros are carefully written to err on the side of the caution,
e.g., we #error if the user manually configures a different endianness
than what we detect.

gmaxwell: What you're suggesting would override the BYTE_ORDER macro. I don't think it makes any sense at all to have arch guessing override explicitlyly set BYTE_ORDER or the user.  There is a difference between a conflicting setting and a guess conflicting with something explicit.

sipa: So really we have 3 potentially different sources:
* User manually setting `SECP256K1_{LITTLE,BIG}_ENDIAN`
* `__BYTE_ORDER__`
* The stuff I'm adding in this PR

Should all 3 be consistent with each other, if they're available? That means no mechanism is left for the user to override if the detection is somehow wrong. Should we have a USE_FORCE_{BIG,LITTLE}_ENDIAN for that case?

One possibility is just using SECP256K1_{BIG,LITTLE}_ENDIAN. If exactly one of those two is set, that's what is used. If not, all other sources that are found have to agree.

sipa:
I've changed the approach here to only check the system macros in case SECP256K1_{BIG,LITTLE}_ENDIAN aren't explicitly set, so it's possible to override a possible incorrect detection. I've also merged all the macros into one check for each of LE and BE.

In GCC/Clang we could always do the detection, and give a warning if it doesn't match an explicit SECP256K1_{BIG,LITTLE}_ENDIAN, but not all compilers have a way of doing so in a standard way.

real-or-random:
I'm ok with this but then we should really add a SHA256 self-test or something similar to context creation. For example, one scenario that concerns me is the user provides a custom nonce derivation function for Schnorr signing and for whatever reason uses the wrong endianness. 

Then the endianness affects the challenge hash but possibly not the nonce, so this could mean that two different challenge hashes are signed for the same nonce.

gmaxwell:
Self tests are good.  There could be a context creation argument to disable it if you really didn't want it.

How would you imagine the schnorr hash would change but the nonce wouldn't? maybe that's something that could be specifically fought against.

real-or-random:
> How would you imagine the schnorr hash would change but the nonce wouldn't?

As I wrote above, this could happen with a user-provided nonce function which just doesn't care about the endianness assumptions in our code. But that's the only scenario that comes to my mind.

sipa:
I've added a single SHA256 selftest, as that should be enough to detect an incorrectly-configured endianness. Other selftests could be added later.

src/secp256k1.c
```diff
@@ -144,8 +146,12 @@ secp256k1_context* secp256k1_context_preallocated_create(void* prealloc, unsigne
 }
 
 secp256k1_context* secp256k1_context_create(unsigned int flags) {
-    size_t const prealloc_size = secp256k1_context_preallocated_size(flags);
-    secp256k1_context* ctx = (secp256k1_context*)checked_malloc(&default_error_callback, prealloc_size);
+    size_t prealloc_size;
+    secp256k1_context* ctx;
+
+    CHECK(secp256k1_selftest());
```
real-or-random, line 152: `CHECK` is intended to be used in the tests only. It calls `abort()` (hardcoded) instead of the error callback, so can't even override this on compile time.

src/secp256k1.c
```diff
@@ -144,8 +146,12 @@ secp256k1_context* secp256k1_context_preallocated_create(void* prealloc, unsigne
...
+    prealloc_size = secp256k1_context_preallocated_size(flags);
+    ctx = (secp256k1_context*)checked_malloc(&default_error_callback, prealloc_size);
```
real-or-random, line 154: I think this is what we should do, namely call `default_error_callback`, which can at least be overridden at compile time.  We do this here because we can't call the error callback unfortunately. It's just not yet determined at context creation time. I'm going to add a note to #780, maybe we can improve this in the future.

elichai:
> @jonasnick right now the tests are hard to make run on any low memory target. 

Is this because of heap allocation? too much stack allocation? something else? if you could specify the requirements/how can I check it myself I can start working on it if it's feasible to fix without adding too much complexity

sipa:
I think this is ready.

jonasnick:
concept ACK

PR #801, opened 2020-08-20
removed dead assignment of "bits" var

whb07:
Toying around with Clang's "scan-build" static analyzer and found this dead assignment. If its meant to be a "zero out", the C/C++ specs state its up to the compiler to decide what to do with it. So by observing the "bits = 0" does nothing, it will just remove it from output. 

The `bits` is removed as the function is about to return.

Tests passing. 
Below is screenshot of scan-build logging the error:
Screen Shot 2020-08-20 at 5 37 59 PM

### DIFF
--- a/src/ecmult_gen_impl.h
+++ b/src/ecmult_gen_impl.h
@@ -151,7 +151,6 @@ static void secp256k1_ecmult_gen(const secp256k1_ecmult_gen_context *ctx, secp25
         secp256k1_ge_from_storage(&add, &adds);
         secp256k1_gej_add_ge(r, r, &add);
     }
-    bits = 0;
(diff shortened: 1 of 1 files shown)


real-or-random:
Thanks! That's indeed a dead store and it's meant as a zero-out but of course it doesn't work. This will be fixed by https://github.com/bitcoin-core/secp256k1/pull/636/files#diff-9d23f46e6342836fd7b34eaac40b4ada as soon as rebase this (hopefully soon). Until then, I'd personally to prefer to keep this line as a reminder unless there is another good reason to remove it.

whb07:
Oooh nice, I didn’t catch that PR yep sounds good. 👌

PR #802, opened 2020-08-23
Alpine installation script

jamesob:
Don't necessarily expect this to be merged, but thought this might be useful documentation for anyone else trying to build on Alpine Linux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,13 @@ libsecp256k1 is built using autotools:
     $ make check
     $ sudo make install  # optional
 
+Platform-specific installation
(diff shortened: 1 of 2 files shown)


sipa:
I'm not sure it makes sense to include instructions/scripts for specific platforms, unless those deviate significantly from what is standard for that platform.

Is there anything special here that is unexpected for Alpine?

jamesob:
Took me a little while to figure out the right constellation of apk packages to install, but this closed pull request should be sufficient breadcrumb for others.

PR #805, opened 2020-08-31
Remove the extremely outdated TODO file.

gmaxwell:
This had two things in it-- tests for the scalar/field code and
 constant time signing and keygen.

The signing and keygen have been thoroughly constant time for years
 and there are now powerful tests to verify it...  no further work
 on constant-time is needed at least on ordinary platforms (other
 sidechannels-- sure).

The scalar and field code have extensive tests.  They could use
 better static test vectors but they're well tested.

TODOs for the project are currently better documented on github
 right now.  This file could return in the future with current
 info, if needed.

### DIFF
--- a/TODO
+++ b/TODO
@@ -1,3 +1,0 @@
-* Unit tests for fieldelem/groupelem, including ones intended to
-  trigger fieldelem's boundary cases.
-* Complete constant-time operations for signing/keygen


real-or-random:
ACK https://github.com/bitcoin-core/secp256k1/pull/805/commits/1c325199d590e018cdfb5ea2ab541774009bf7f7

sipa:
Last updated in 2013 (b03d836be9e3779dd23bb9e3447d51ecb27f89e#diff-b7b1e314614cf326c6e2b6eba1540682)

ACK

PR #807, opened 2020-09-04
Make group order purely compile-time in exhaustive tests

sipa:
The current exhaustive tests rely on an order that's known at compile time (EXHAUSTIVE_TEST_ORDER), but it is still passed down as a runtime variable `order` as well.

Make this consistent by replacing all `order` into `EXHAUSTIVE_TEST_ORDER`. This should help the compiler avoid actual modulus operations, though I don't see any speedup.


sipa:
Ping @apoelstra.

sipa:
Superseded by #808.

PR #808, opened 2020-09-06
Exhaustive test improvements + exhaustive schnorrsig tests

sipa:
A few miscellaneous improvements:
* Just use EXHAUSTIVE_TEST_ORDER as order everywhere, rather than a variable
* Move exhaustive tests for recovery module to the recovery module directory
* Make `secp256k1_scalar_set_b32` detect overflow correctly for scalar_low (a comment in the recovery exhaustive test indicated why this was the case, but this looks incorrect).
* Change the small test groups so that they include a point with X coordinate 1.
* Initialize the RNG seed, allowing configurating from the cmdline, and report it.
* Permit changing the number of iterations (re-randomizing for each).
* Support splitting the work across cores from the cmdline.

And a big one:
* Add exhaustive tests for schnorrsig module (and limited ones for extrakeys).

### DIFF
--- a/sage/gen_exhaustive_groups.sage
+++ b/sage/gen_exhaustive_groups.sage
@@ -1,0 +1,129 @@
+# Define field size and field
+P = 2^256 - 2^32 - 977
+F = GF(P)
+BETA = F(0x7ae96a2b657c07106e64479eac3434e99cf0497512f58995c1396c28719501ee)
+
+assert(BETA != F(1) and BETA^3 == F(1))
+
+orders_done = set()
+results = {}
+first = True
+for b in range(1, P):
+    # There are only 6 curves (up to isomorphism) of the form y^2=x^3+B. Stop once we have tried all.
+    if len(orders_done) == 6:
+        break
+
+    E = EllipticCurve(F, [0, b])
+    print("Analyzing curve y^2 = x^3 + %i" % b)
+    n = E.order()
+    # Skip curves with an order we've already tried
+    if n in orders_done:
+        print("- Isomorphic to earlier curve")
+        continue
+    orders_done.add(n)
+    # Skip curves isomorphic to the real secp256k1
+    if n.is_pseudoprime():
+        print(" - Isomorphic to secp256k1")
+        continue
+
+    print("- Finding subgroups")
+
+    # Find what prime subgroups exist
+    for f, _ in n.factor():
+        print("- Analyzing subgroup of order %i" % f)
+        # Skip subgroups of order >1000
+        if f < 4 or f > 1000:
+            print("  - Bad size")
+            continue
+
+        # Iterate over X coordinates until we find one that is on the curve, has order f,
+        # and for which curve isomorphism exists that maps it to X coordinate 1.
+        for x in range(1, P):
+            # Skip X coordinates not on the curve, and construct the full point otherwise.
+            if not E.is_x_coord(x):
+                continue
+            G = E.lift_x(F(x))
+
+            print("  - Analyzing (multiples of) point with X=%i" % x)
+
+            # Skip points whose order is not a multiple of f. Project the point to have
+            # order f otherwise.
+            if (G.order() % f):
+                print("    - Bad order")
+                continue
+            G = G * (G.order() // f)
+
+            # Find lambda for endomorphism. Skip if none can be found.
+            lam = None
+            for l in Integers(f)(1).nth_root(3, all=True):
+                if int(l)*G == E(BETA*G[0], G[1]):
+                    lam = int(l)
+                    break
+            if lam is None:
+                print("    - No endomorphism for this subgroup")
+                break
+
+            # Now look for an isomorphism of the curve that gives this point an X
+            # coordinate equal to 1.
+            # If (x,y) is on y^2 = x^3 + b, then (a^2*x, a^3*y) is on y^2 = x^3 + a^6*b.
+            # So look for m=a^2=1/x.
+            m = F(1)/G[0]
+            if not m.is_square():
+                print("    - No curve isomorphism maps it to a point with X=1")
+                continue
+            a = m.sqrt()
+            rb = a^6*b
+            RE = EllipticCurve(F, [0, rb])
+
+            # Use as generator twice the image of G under the above isormorphism.
+            # This means that generator*(1/2 mod f) will have X coordinate 1.
+            RG = RE(1, a^3*G[1]) * 2
+            # And even Y coordinate.
+            if int(RG[1]) % 2:
+                RG = -RG
+            assert(RG.order() == f)
+            assert(lam*RG == RE(BETA*RG[0], RG[1]))
+
+            # We have found curve RE:y^2=x^3+rb with generator RG of order f. Remember it
+            results[f] = {"b": rb, "G": RG, "lambda": lam}
+            print("    - Found solution")
+            break
+
+    print("")
+
+print("")
+print("")
+print("/* To be put in src/group_impl.h: */")
+first = True
+for f in sorted(results.keys()):
+    b = results[f]["b"]
+    G = results[f]["G"]
+    print("#  %s EXHAUSTIVE_TEST_ORDER == %i" % ("if" if first else "elif", f))
+    first = False
+    print("static const secp256k1_ge secp256k1_ge_const_g = SECP256K1_GE_CONST(")
+    print("    0x%08x, 0x%08x, 0x%08x, 0x%08x," % tuple((int(G[0]) >> (32 * (7 - i))) & 0xffffffff for i in range(4)))
+    print("    0x%08x, 0x%08x, 0x%08x, 0x%08x," % tuple((int(G[0]) >> (32 * (7 - i))) & 0xffffffff for i in range(4, 8)))
+    print("    0x%08x, 0x%08x, 0x%08x, 0x%08x," % tuple((int(G[1]) >> (32 * (7 - i))) & 0xffffffff for i in range(4)))
+    print("    0x%08x, 0x%08x, 0x%08x, 0x%08x" % tuple((int(G[1]) >> (32 * (7 - i))) & 0xffffffff for i in range(4, 8)))
+    print(");")
+    print("static const secp256k1_fe secp256k1_fe_const_b = SECP256K1_FE_CONST(")
+    print("    0x%08x, 0x%08x, 0x%08x, 0x%08x," % tuple((int(b) >> (32 * (7 - i))) & 0xffffffff for i in range(4)))
+    print("    0x%08x, 0x%08x, 0x%08x, 0x%08x" % tuple((int(b) >> (32 * (7 - i))) & 0xffffffff for i in range(4, 8)))
+    print(");")
+print("#  else")
+print("#    error No known generator for the specified exhaustive test group order.")
+print("#  endif")
+
+print("")
+print("")
+print("/* To be put in src/scalar_impl.h: */")
+first = True
+for f in sorted(results.keys()):
+    lam = results[f]["lambda"]
+    print("#  %s EXHAUSTIVE_TEST_ORDER == %i" % ("if" if first else "elif", f))
+    first = False
+    print("#    define EXHAUSTIVE_TEST_LAMBDA %i" % lam)
+print("#  else")
+print("#    error No known lambda for the specified exhaustive test group order.")
+print("#  endif")
+print("")
--- a/src/group.h
+++ b/src/group.h
@@ -139,4 +139,15 @@ static void secp256k1_ge_storage_cmov(secp256k1_ge_storage *r, const secp256k1_g
 /** Rescale a jacobian point by b which must be non-zero. Constant-time. */
 static void secp256k1_gej_rescale(secp256k1_gej *r, const secp256k1_fe *b);
 
+/** Determine if a point (which is assumed to be on the curve) is in the correct (sub)group of the curve.
+ *
+ * In normal mode, the used group is secp256k1, which has cofactor=1 meaning that every point on the curve is in the
+ * group, and this function returns always true.
+ *
+ * When compiling in exhaustive test mode, a slightly different curve equation is used, leading to a group with a
+ * (very) small subgroup, and that subgroup is what is used for all cryptographic operations. In that mode, this
+ * function checks whether a point that is on the curve is in fact also in that subgroup.
+ */
+static int secp256k1_ge_is_in_correct_subgroup(const secp256k1_ge* ge);
+
 #endif /* SECP256K1_GROUP_H */
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -11,52 +11,38 @@
 #include "field.h"
 #include "group.h"
 
-/* These points can be generated in sage as follows:
+/* These exhaustive group test orders and generators are chosen such that:
+ * - The field size is equal to that of secp256k1, so field code is the same.
+ * - The curve equation is of the form y^2=x^3+B for some constant B.
+ * - The subgroup has a generator 2*P, where P.x=1.
+ * - The subgroup has size less than 1000 to permit exhaustive testing.
+ * - The subgroup admits an endomorphism of the form lambda*(x,y) == (beta*x,y).
  *
- * 0. Setup a worksheet with the following parameters.
- *   b = 4  # whatever secp256k1_fe_const_b will be set to
- *   F = FiniteField (0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F)
- *   C = EllipticCurve ([F (0), F (b)])
- *
- * 1. Determine all the small orders available to you. (If there are
- *    no satisfactory ones, go back and change b.)
- *   print C.order().factor(limit=1000)
- *
- * 2. Choose an order as one of the prime factors listed in the above step.
- *    (You can also multiply some to get a composite order, though the
- *    tests will crash trying to invert scalars during signing.) We take a
- *    random point and scale it to drop its order to the desired value.
- *    There is some probability this won't work; just try again.
- *   order = 199
- *   P = C.random_point()
- *   P = (int(P.order()) / int(order)) * P
- *   assert(P.order() == order)
- *
- * 3. Print the values. You'll need to use a vim macro or something to
- *    split the hex output into 4-byte chunks.
- *   print "%x %x" % P.xy()
+ * These parameters are generated using sage/gen_exhaustive_groups.sage.
  */
 #if defined(EXHAUSTIVE_TEST_ORDER)
-#  if EXHAUSTIVE_TEST_ORDER == 199
+#  if EXHAUSTIVE_TEST_ORDER == 13
 static const secp256k1_ge secp256k1_ge_const_g = SECP256K1_GE_CONST(
-    0xFA7CC9A7, 0x0737F2DB, 0xA749DD39, 0x2B4FB069,
-    0x3B017A7D, 0xA808C2F1, 0xFB12940C, 0x9EA66C18,
-    0x78AC123A, 0x5ED8AEF3, 0x8732BC91, 0x1F3A2868,
-    0x48DF246C, 0x808DAE72, 0xCFE52572, 0x7F0501ED
+    0xc3459c3d, 0x35326167, 0xcd86cce8, 0x07a2417f,
+    0x5b8bd567, 0xde8538ee, 0x0d507b0c, 0xd128f5bb,
+    0x8e467fec, 0xcd30000a, 0x6cc1184e, 0x25d382c2,
+    0xa2f4494e, 0x2fbe9abc, 0x8b64abac, 0xd005fb24
 );
-
-static const secp256k1_fe secp256k1_fe_const_b = SECP256K1_FE_CONST(0, 0, 0, 0, 0, 0, 0, 4);
-
-#  elif EXHAUSTIVE_TEST_ORDER == 13
+static const secp256k1_fe secp256k1_fe_const_b = SECP256K1_FE_CONST(
+    0x3d3486b2, 0x159a9ca5, 0xc75638be, 0xb23a69bc,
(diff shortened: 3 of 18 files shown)


jonasnick (approved):
ACK 634b3e9024340797b09105533ca63cf35cce427d

I don't really see the point of removing the order variable - was less strenous to read.

Did you notice the low order scalar_set_b32 issue in the exhaustive schnorrsig tests?
I couldn't find a historic reason for the old comment about the infinite loop but the test clearly show that it's wrong.

sipa:
> I don't really see the point of removing the order variable - was less strenous to read.

I'm also ok with replacing it with `order` everywhere, but in the current code it's a mix, which makes no sense IMO. In theory using a compile-time constant should be faster (modulus operations are very slow for the CPU, but modulus with a known constant can be rewritten using multiply/shift/subtract by the compiler), so I chose that side. I can't measure a difference in performance though, so it probably doesn't matter much.

> Did you notice the low order scalar_set_b32 issue in the exhaustive schnorrsig tests?

Sort of, I wanted to add a test that uses an out-of-range `s` value, but expected it would break things with the low-order scalar implementation that ignored overflows. In trying to fix that, I noticed it wasn't actually necessary in the first place.

sipa:
Added a commit that changes the exhaustive test groups so they include a point with X coordinate 1 (suggested by @gmaxwell).

gmaxwell:
nice, utack

src/group_impl.h
```diff
@@ -11,49 +11,125 @@
...
+ * - The curve equation is of the form y^2=x^3+B for some constant B.
+ * - The order is less than 1000 to permit exhaustive testing.
+ * - The order is a multiple of 3 plus 1, enabling the endomorphism optimization.
+ * - The generator is 2*P, where P has X coefficient equal to 1.
  *
- * 0. Setup a worksheet with the following parameters.
- *   b = 4  # whatever CURVE_B will be set to
- *   F = FiniteField (0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F)
- *   C = EllipticCurve ([F (0), F (b)])
+ * They can generated with the following Sage code:
```
jonasnick, line 21: We have a sage directory, perhaps best to move this over? My first attempt at copy & pasting this messed up indentation.

src/group_impl.h
```diff
@@ -11,49 +11,125 @@
 #include "field.h"
 #include "group.h"
 
-/* These points can be generated in sage as follows:
+/* These exhaustive group test orders and generators are chosen such that:
+ * - The field size is equal to that of secp256k1, so field code is the same.
+ * - The curve equation is of the form y^2=x^3+B for some constant B.
+ * - The order is less than 1000 to permit exhaustive testing.
+ * - The order is a multiple of 3 plus 1, enabling the endomorphism optimization.
+ * - The generator is 2*P, where P has X coefficient equal to 1.
```
jonasnick, line 19: coefficient -> coordinate?

jonasnick (commented):
Nice idea to create groups with an X=1 point!

src/group_impl.h
sipa: Yeah, done.

src/group_impl.h
sipa: Done.

sipa:
I've updated https://github.com/sipa/secp256k1/commits/202009_schnorrsig_exhaustive to build on this, including tests with signatures that have R.x=(fieldsize+1).

src/group_impl.h
```diff
@@ -11,49 +11,49 @@
...
- *   assert(P.order() == order)
- *
- * 3. Print the values. You'll need to use a vim macro or something to
- *    split the hex output into 4-byte chunks.
- *   print "%x %x" % P.xy()
+ * They can generated using the Sage code in sage/gen_exhaustive_groups.sage.
  */
 #if defined(EXHAUSTIVE_TEST_ORDER)
-#  if EXHAUSTIVE_TEST_ORDER == 199
+#  if EXHAUSTIVE_TEST_ORDER == 13
```
real-or-random, line 24: Shouldn't `EXHAUSTIVE_TEST_LAMBDA` chosen by these #ifs too?

src/group_impl.h
```diff
@@ -11,49 +11,49 @@
...
+    0xa2f4494e, 0x2fbe9abc, 0x8b64abac, 0xd005fb24
 );
-
-static const int CURVE_B = 4;
-#  elif EXHAUSTIVE_TEST_ORDER == 13
+static const secp256k1_fe secp256k1_fe_const_b = SECP256K1_FE_CONST(
+    0x3d3486b2, 0x159a9ca5, 0xc75638be, 0xb23a69bc,
+    0x946a45ab, 0x24801247, 0xb4ed2b8e, 0x26b6a417
+);
+#  elif EXHAUSTIVE_TEST_ORDER == 199
```
real-or-random, line 35: And have you tried running this with 199? Seems to take forever.

real-or-random (approved):
ACK aa48ca9c5b91f570eb2cb52636cd590eccea5de4 careful code inspection

My two comments are strictly speaking not this PR. If these are real issues, I guess you could address them here or in another PR, either is fine.

src/group_impl.h
jonasnick: I tried it on my laptop over night but terminated it the next morning.

sipa: I tested an earlier version of the extrakeys/schnorr exhaustive tests. The generic group and ECDSA ones are too slow with 199 (they include quintuple nested loops; the schnorr tests are only quartic).

sage/gen_exhaustive_groups.sage
```diff
@@ -0,0 +1,83 @@
...
+            # If (x,y) is on y^2 = x^3 + b, then (a^2*x, a^3*y) is on y^2 = x^3 + a^6*b.
+            # So look for m=a^2=1/x.
+            m = F(1)/G[0]
+            if not m.is_square():
+                continue
+            rb = b*m^3
+            RE = EllipticCurve(F, [0, rb])
+
+            # Use as generator twice the point with this low X coordinate (like secp256k1!)
+            RG = RE.lift_x(1) * 2
```
jonasnick, line 59: Why does `RG` have the same order as `G`?

jonasnick (approved):
ACK aa48ca9c5b91f570eb2cb52636cd590eccea5de4

sage/gen_exhaustive_groups.sage
sipa: It's the image of G under the isomorphism described in the comments. I can make it more explicit by computing it directly instead of re-lifting from 1.

src/group_impl.h
sipa: They're needed in scalar, which is independent from the group "module". Otherwise, done: I've added ifdefs there to pick the correct one automatically.

sage/gen_exhaustive_groups.sage
sipa: Done.

src/scalar_impl.h
```diff
@@ -254,6 +254,14 @@ static void secp256k1_scalar_inverse_var(secp256k1_scalar *r, const secp256k1_sc
 
 #ifdef USE_ENDOMORPHISM
 #if defined(EXHAUSTIVE_TEST_ORDER)
+#  if EXHAUSTIVE_TEST_ORDER == 13
+#    define EXHAUSTIVE_TEST_LAMBDA 9
+#  elif EXHAUSTIVE_TEST_ORDER == 199
+#    define EXHAUSTIVE_TEST_LAMBDA 92
+#  else
+#    error No known lambda for the specified exhaustive test group order.
+#endif
```
real-or-random: nit: indentation.

No lambda for order 7?

sipa: I just pushed another update.

The order 7 subgroup in fact does not have the endomorphism property (apparently all subgroups do, except the order 2, 3, and 7 ones).

sipa:
I made another change, making the exhaustive tests correctly initialize the RNG, and adding a way to split the workload (invoke with `./exhaustive_tests 1 "" 8 0` on CPU 0, `./exhaustive_tests 1 "" 8 1` on CPU 1, etc).

I'm done expanding the scope of this PR now. If this is too much I'm happy to split stuff off.

src/scalar_impl.h
real-or-random: Ok I see.

Can you fix the #endif indentation?

sipa: It's already fixed, I think?

sipa: Oh, it was fixed in a later commit only. Really fixed now.

real-or-random: Oh sorry, I thought I looked in the later commits too. For some reason I missed it there.

real-or-random:
Valgrind fails on Travis:
https://travis-ci.org/github/bitcoin-core/secp256k1/jobs/725488761#L573

> I made another change, making the exhaustive tests correctly initialize the RNG, and adding a way to split the workload (invoke with `./exhaustive_tests 1 "" 8 0` on CPU 0, `./exhaustive_tests 1 "" 8 1` on CPU 1, etc).

Really invoke the seed with "" ? Doesn't every instance get a different random seed then? 

> I'm done expanding the scope of this PR now. If this is too much I'm happy to split stuff off.

I think that's ok.

src/tests_exhaustive.c
```diff
@@ -327,184 +340,94 @@ void test_exhaustive_sign(const secp256k1_context *ctx, const secp256k1_ge *grou
...
-                    /* Verify by converting to a standard signature and calling verify */
-                    secp256k1_ecdsa_recoverable_signature_save(&rsig, &r_s, &s_s, recid);
-                    secp256k1_ecdsa_recoverable_signature_convert(ctx, &sig, &rsig);
-                    memcpy(&nonconst_ge, &group[sk_s], sizeof(nonconst_ge));
-                    secp256k1_pubkey_save(&pk, &nonconst_ge);
-                    CHECK(should_verify ==
-                          secp256k1_ecdsa_verify(ctx, &sig, msg32, &pk));
-                }
-            }
+    printf("test count = %i\n", count);
```
real-or-random, line 358: This is where valgrind on MacOS complains about uninit data.

This is spurious, see https://stackoverflow.com/questions/29312335/valgrind-on-os-x-yosemite-giving-bogus-errors ... @elichai Do you think we can update to MacOS 10.13 ? According to homebrew, valgrind supports this now.

real-or-random: Ah no, we're at 10.13 already... Upstream bug: https://bugs.kde.org/show_bug.cgi?id=399584 There they mention https://github.com/LouisBrunner/valgrind-macos as an alternative tap. Do you think it's worth trying?

gmaxwell:
I don't think exhaustive_tests should be actually random.  There is a need for 'random' values, but I don't see a lot of gain in having every run use different ones... and they make reproducing issues harder.   Is there a reason to not just give it a constant seed?

src/tests_exhaustive.c
elichai: We can try, idk how much will we trust it though (if someone is bored enough, he can go over their patches hehe https://github.com/LouisBrunner/valgrind-macos/commits/patches)

sage/gen_exhaustive_groups.sage
jonasnick: Oh I see now. You're specifically checking that `m` is square, i.e., that `a` exists.

sipa:
> Really invoke the seed with "" ? Doesn't every instance get a different random seed then?

Yes, for the random part. You can also specify a fixed seed by passing a non-empty hex string.

> I don't think exhaustive_tests should be actually random. There is a need for 'random' values, but I don't see a lot of gain in having every run use different ones... and they make reproducing issues harder. Is there a reason to not just give it a constant seed?

Unsure.

At least the gej tests are explicitly rescaled with a random z in every run, and unfortunately, those cannot be selected from an tractably-sized set for exhaustive tests.

The main thing these tests hope to reveal shouldn't be dependent on the choice of the seed, though, so perhaps it's ok to just pick a fixed seed.

real-or-random:
> > Really invoke the seed with "" ? Doesn't every instance get a different random seed then?
> 
> Yes, for the random part. You can also specify a fixed seed by passing a non-empty hex string.

Ok, yes. I somehow wrongly assumed that the randomness affects the mapping of work to CPUs, but it does not of course.  


> > I don't think exhaustive_tests should be actually random. There is a need for 'random' values, but I don't see a lot of gain in having every run use different ones... and they make reproducing issues harder. Is there a reason to not just give it a constant seed?
> 
> Unsure.
> 
> At least the gej tests are explicitly rescaled with a random z in every run, and unfortunately, those cannot be selected from an tractably-sized set for exhaustive tests.
> 
> The main thing these tests hope to reveal shouldn't be dependent on the choice of the seed, though, so perhaps it's ok to just pick a fixed seed. If so, we can probably also do without the iteration count.

"Unsure" was my first thought, too. Yes, the main thing here does not depend of on the choice of the seed but I still believe that randomness helps more than it hurts here. For example, it could detect a bug that occurs as a combination of special coordinate value 1 and some z value that we hit with probability 1/10000. This could be missed by the normal tests, even though they're randomized.

gmaxwell:
I suppose it doesn't hurt.  If a user reports a bug that only happens with some particular seeds and they fail to provide their seed-- well that's a bug which wouldn't have otherwise been detected at all.

Just please no "expected failures", the situation in CI  (common in bitcoin core but has happened here too) where people are responding to failures by retrying the CI rather than treating the failure as an emergency is a dangerous practice.

src/tests_exhaustive.c
sipa: Anyone have a clue why the `tests` binary doesn't have the exact same problem?

sipa: Seems that fixed it.

sipa:
Adding `setbuf(stdout, NULL);` seems to have fixed it...

real-or-random (approved):
ACK dd29d2ef298e92e9bae0bf35b269fb38dd79b23f code review and tests pass

sipa:
Rebased on the now-merged #558, and added exhaustive tests for schnorrsig module too.

src/modules/schnorrsig/tests_exhaustive_impl.h
```diff
@@ -0,0 +1,211 @@
...
+        memcpy(pk32, xonly_pubkey_bytes[d - 1], 32);
+        actual_d = parities[d - 1] ? EXHAUSTIVE_TEST_ORDER - d : d;
+        /* Iterate over the possible valid first 32 bytes in the signature, through their corresponding DL k.
+           Values above EXHAUSTIVE_TEST_ORDER/2 refer to the entries in invalid_pubkey_bytes. */
+        for (k = 1; k <= EXHAUSTIVE_TEST_ORDER / 2 + NUM_INVALID_KEYS; ++k) {
+            unsigned char sig64[64];
+            int actual_k = -1;
+            int e_done[EXHAUSTIVE_TEST_ORDER] = {0};
+            int e_count_done = 0;
+            if (skip_section(iter++)) continue;
```
jonasnick, line 92: `skip_section` takes a pointer, same in schnorrsig_sign

src/modules/schnorrsig/tests_exhaustive_impl.h
```diff
@@ -0,0 +1,211 @@
...
+        secp256k1_scalar_set_int(&scalar_i, i);
+        secp256k1_scalar_get_b32(buf, &scalar_i);
+        CHECK(secp256k1_keypair_create(ctx, &keypair[i - 1], buf));
+        CHECK(secp256k1_keypair_xonly_pub(ctx, &xonly_pubkey[i - 1], &parity[i - 1], &keypair[i - 1]));
+        CHECK(secp256k1_xonly_pubkey_serialize(ctx, xonly_pubkey_bytes[i - 1], &xonly_pubkey[i - 1]));
+    }
+
+    /* The keypairs/pubkey functions above have already been tested agains the group;
+       see src/modules/extrakeys/tests_exhaustive_impl.h. */
+    (void)group;
```
jonasnick, line 205: why not just remove this argument?

sage/gen_exhaustive_groups.sage
jonasnick: The new script output makes running this quite exciting!

jonasnick (commented):
Nice! LGTM mod nits

src/modules/schnorrsig/tests_exhaustive_impl.h
sipa: Fixed.

src/modules/schnorrsig/tests_exhaustive_impl.h
sipa: Yeah, done.

I wanted to keep the same "interface" for all exhaustive tests that are delegated to modules, but that's easy to add if we'd a lot more modules too.

src/tests_exhaustive.c
```diff
@@ -327,184 +343,114 @@ void test_exhaustive_sign(const secp256k1_context *ctx, const secp256k1_ge *grou
...
+
+    /* find random seed */
+    secp256k1_rand_init(argc > 2 ? argv[2] : NULL);
+
+    /* set up split processing */
+    if (argc > 4) {
+        num_cores = strtol(argv[3], NULL, 0);
+        this_core = strtol(argv[4], NULL, 0);
+        if (num_cores < 1 || this_core >= num_cores) {
+            fprintf(stderr, "Usage: %s [count] [seed] [numcores] [thiscore]", argv[0]);
```
gmaxwell, line 386: needs a \n at the end.

sipa: Fixed.

src/group.h
```diff
@@ -138,4 +138,7 @@ static void secp256k1_ge_storage_cmov(secp256k1_ge_storage *r, const secp256k1_g
 /** Rescale a jacobian point by b which must be non-zero. Constant-time. */
 static void secp256k1_gej_rescale(secp256k1_gej *r, const secp256k1_fe *b);
 
+/** Determine if a point (which is assumed to be on the curve) is in the group. */
```
real-or-random, line 141: I think "is in group" is somewhat imprecise (which group?), even for readers who know about exhaustive tests. And others may really wonder what the point of this function is. 

The same is true for the name of the function. When I saw this, I first assumed it checks if the input is really on the point.

sipa: Added some comments. WDYT?

jonasnick (approved):
ACK 76ac2bf214cd6a272176e90b98c84d34fe440f18

src/group.h
real-or-random: Yes, looks good! Now that you got a few more comments, I still think the function name can be improved. I don't have a great suggestion but something like  `ge_is_in_correct_subgroup`, `ge_is_in_subgroup`  would already make clear that this function is not about being on the curve or not. Note also the "is", I think that's more consistent with other names like `is_infinity` and makes clear that this is a getter.

gmaxwell:
I ran this for 199 without and with endo across 256 cores. Each process completed within +/-0.3%  of the others.  Total CPU time was 292 CPU-days (for each configuration), as expected-- it passed.

src/testrand_impl.h
```diff
@@ -111,4 +112,47 @@ static void secp256k1_rand_flip(unsigned char *b, size_t len) {
     b[secp256k1_rand_int(len)] ^= (1 << secp256k1_rand_int(8));
 }
 
+static void secp256k1_rand_init(const char* hexseed) {
```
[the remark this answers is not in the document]
sipa: You're right, but that's not specific to that function. It's in a file called `testrand` - should all functions in it be renamed to have `test` in the name?

src/tests_exhaustive.c
```diff
@@ -327,184 +343,114 @@ void test_exhaustive_sign(const secp256k1_context *ctx, const secp256k1_ge *grou
...
+    setbuf(stdout, NULL);
+    /* Also disable buffering for stderr because it's not guaranteed that it's
+     * unbuffered on all systems. */
+    setbuf(stderr, NULL);
+
+    /* find iteration count */
+    if (argc > 1) {
+        count = strtol(argv[1], NULL, 0);
+    }
+    printf("test count = %i\n", count);
```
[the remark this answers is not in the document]
sipa: Done.

src/tests_exhaustive.c
```diff
@@ -327,184 +343,114 @@ void test_exhaustive_sign(const secp256k1_context *ctx, const secp256k1_ge *grou
...
+
+    /* set up split processing */
+    if (argc > 4) {
+        num_cores = strtol(argv[3], NULL, 0);
+        this_core = strtol(argv[4], NULL, 0);
+        if (num_cores < 1 || this_core >= num_cores) {
+            fprintf(stderr, "Usage: %s [count] [seed] [numcores] [thiscore]\n", argv[0]);
+            return 1;
+        }
+        printf("running tests for core %lu out of %lu\n", (unsigned long)this_core, (unsigned long)num_cores);
```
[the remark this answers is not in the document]
sipa: I've changed it to `core 255 (out of [0..255])`. Does that seem better?

src/group.h
sipa: Done.

src/tests_exhaustive.c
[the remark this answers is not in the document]
gmaxwell: Seems fine to me!

src/testrand_impl.h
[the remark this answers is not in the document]
gmaxwell: That wouldn't be bad.  Mostly I raised this one becuase I thought it was likely to come to someone's attention if they searched the codebase for 'random', and in isolation it looks kind of bad! :)  (and was probably somewhat less of an issue previously when it was in tests.c's main.)

jonasnick (approved):
ACK 8b7dcdd955a4f57174f478e36bdae5b84784fb9c

real-or-random:
ACK 8b7dcdd955a4f57174f478e36bdae5b84784fb9c

src/testrand_impl.h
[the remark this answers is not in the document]
sipa: Done in a separate PR: https://github.com/bitcoin-core/secp256k1/pull/827

PR #814, opened 2020-09-13
tests: Initialize random group elements fully

real-or-random:
Also fix add a missing comment.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -86,6 +86,7 @@ void random_group_element_test(secp256k1_ge *ge) {
             break;
         }
     } while(1);
+    ge->infinity = 0;
 }
 
 void random_group_element_jacobian_test(secp256k1_gej *gej, const secp256k1_ge *ge) {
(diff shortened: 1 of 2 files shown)


gmaxwell:
utACK. Looks fine to me.

sipa:
utACK 5738e8622d8ba02caa984425c23c072a3f14352c

src/tests.c
```diff
@@ -86,6 +86,7 @@ void random_group_element_test(secp256k1_ge *ge) {
             break;
         }
     } while(1);
+    ge->infinity = 0;
```
real-or-random, line 89: Oh, on a second thought, this is not just necessary. `secp256k1_ge_set_xo_var` (a few lines above) sets `infinity` to 0.

I was playing around with multiple additional `VERIFY_CHECK`s when reviewing the schnorrsig PR and I wrongly believed that I found this bug ... 

Well whatever, the correct was correct before this PR and it's correct after this PR. If we want, we can reverse this commit in a later PR.

PR #816, opened 2020-09-14
Improve checks at top of _fe_negate methods

peterdettman:
In theory we could have a single static assertion that would ensure all of these are always true (for any magnitude up to the limit), but I think this small redundancy is clearer.

### DIFF
--- a/src/field_10x26_impl.h
+++ b/src/field_10x26_impl.h
@@ -391,6 +391,10 @@ SECP256K1_INLINE static void secp256k1_fe_negate(secp256k1_fe *r, const secp256k
 #ifdef VERIFY
     VERIFY_CHECK(a->magnitude <= m);
     secp256k1_fe_verify(a);
+    VERIFY_CHECK(0x3FFFC2FUL * 2 * (m + 1) >= 0x3FFFFFFUL * 2 * m);
+    VERIFY_CHECK(0x3FFFFBFUL * 2 * (m + 1) >= 0x3FFFFFFUL * 2 * m);
+    VERIFY_CHECK(0x3FFFFFFUL * 2 * (m + 1) >= 0x3FFFFFFUL * 2 * m);
+    VERIFY_CHECK(0x03FFFFFUL * 2 * (m + 1) >= 0x03FFFFFUL * 2 * m);
 #endif
     r->n[0] = 0x3FFFC2FUL * 2 * (m + 1) - a->n[0];
     r->n[1] = 0x3FFFFBFUL * 2 * (m + 1) - a->n[1];
--- a/src/field_5x52_impl.h
+++ b/src/field_5x52_impl.h
@@ -377,6 +377,9 @@ SECP256K1_INLINE static void secp256k1_fe_negate(secp256k1_fe *r, const secp256k
 #ifdef VERIFY
     VERIFY_CHECK(a->magnitude <= m);
     secp256k1_fe_verify(a);
+    VERIFY_CHECK(0xFFFFEFFFFFC2FULL * 2 * (m + 1) >= 0xFFFFFFFFFFFFFULL * 2 * m);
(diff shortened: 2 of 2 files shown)


robot-dreams (commented):
~ACK b06de0a9ee5af6f83821cb227ea3aba014f6d4ab~ edit: on second thought, I want to make sure I understand what's going on at magnitude 32 (for 10x26) and 2048 (for 5x52) before ACKing this change.

I ran a script to confirm that the field_10x26 check holds for all magnitudes in [0, 31], and that the field_5x52 check holds for all magnitudes in [0, 2047].

I also looked at the implementation of `fe_verify` to make sure the RHS of each check is indeed the largest possible value of the corresponding limb (given the magnitude m).

Since there was no CI run for this change, I rebased onto 49f608de474a89e5c742830c8ff5ac8978b5796a and ran `make check`—all tests passed locally.

real-or-random:
> I ran a script to confirm that the field_10x26 check holds for all magnitudes in [0, 31], and that the field_5x52 check holds for all magnitudes in [0, 2047].

Do you think it would make sense to post this here for other reviewers to check or even have this in the repo?


> Since there was no CI run for this change, I rebased onto 49f608d and ran make check—all tests passed locally.

I tried to trigger manually but it didn't work (https://cirrus-ci.com/build/5109653131493376). @peterdettman Can you rebase this on master so we get a CI run?

robot-dreams:
Sure, here you go! It's pretty straightforward—all I did was make sure that for every magnitude in the range, (1) the value doesn't overflow the word size, and (2) the inequality holds.

```python3
UINT32_MAX = (1 << 32) - 1
UINT64_MAX = (1 << 64) - 1

M_MAX_10x26 = 32
M_MAX_5x52 = 2048

def verify_ge_32(a, b):
    assert a <= UINT32_MAX
    assert b <= UINT32_MAX
    assert a >= b

def verify_ge_64(a, b):
    assert a <= UINT64_MAX
    assert b <= UINT64_MAX
    assert a >= b

for m in range(0, M_MAX_10x26):
    verify_ge_32(0x3FFFC2F * 2 * (m + 1), 0x3FFFFFF * 2 * m);
    verify_ge_32(0x3FFFFBF * 2 * (m + 1), 0x3FFFFFF * 2 * m);
    verify_ge_32(0x3FFFFFF * 2 * (m + 1), 0x3FFFFFF * 2 * m);
    verify_ge_32(0x03FFFFF * 2 * (m + 1), 0x03FFFFF * 2 * m);

for m in range(0, M_MAX_5x52):
    verify_ge_64(0xFFFFEFFFFFC2F * 2 * (m + 1), 0xFFFFFFFFFFFFF * 2 * m);
    verify_ge_64(0xFFFFFFFFFFFFF * 2 * (m + 1), 0xFFFFFFFFFFFFF * 2 * m);
    verify_ge_64(0x0FFFFFFFFFFFF * 2 * (m + 1), 0x0FFFFFFFFFFFF * 2 * m);
```

Note: The checks would fail at the "endpoints" `m = 32` (for 10x26) and `m = 2048` (for 5x52), but I think this is OK / expected because the magnitude of the result will be `m + 1`, so the final `fe_verify` check would have failed anyway. Maybe it's worth explicitly adding an extra check that `m` isn't too big, but I don't have a strong opinion about this.

peterdettman:
@real-or-random Rebased.

robot-dreams:
ACK 3e03435ba36f07ef956d4bfb71a3fa580ce29326, no difference aside from rebase

sipa:
utACK 515e7953cab4eb3be063fa3991e4e0663d3f04ae

real-or-random:
ACK 515e7953cab4eb3be063fa3991e4e0663d3f04ae bounds hold by inspection and by @robot-dreams 's script

I was a little bit skeptical that this will increase the running time of the tests but it doesn't make a difference on my machine. I had that thought because I recently noticed that `fe_verify` makes up for about 15% of the default `./tests 64`. I tried to optimize it but I failed to find a faster variant. (@peterdettman if you're in the mood...  :D)

PR #818, opened 2020-09-17
Add static assertion that uint32_t is unsigned int or wider

real-or-random:
Solves one item in #792 .

### DIFF
--- a/src/assumptions.h
+++ b/src/assumptions.h
@@ -7,6 +7,8 @@
 #ifndef SECP256K1_ASSUMPTIONS_H
 #define SECP256K1_ASSUMPTIONS_H
 
+#include <limits.h>
+
 #include "util.h"
 
 /* This library, like most software, relies on a number of compiler implementation defined (but not undefined)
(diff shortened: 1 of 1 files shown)


src/assumptions.h
```diff
@@ -19,7 +21,11 @@ struct secp256k1_assumption_checker {
        allowed. */
     int dummy_array[(
         /* Bytes are 8 bits. */
-        CHAR_BIT == 8 &&
+        (CHAR_BIT == 8) &&
+
+        /* No integer promotion for uint32_t. This ensures that we can multiply uintXX_t values where XX >= 32
+           without signed overflow, which would be undefined behaviour. */
+        (UINT_MAX <= UINT32_MAX) &&
```
roconnor-blockstream, line 28: I think `INT_MAX < UINT32_MAX` would more directly capture your intention here; however given the other rules of C, I do admit that what you have written ends up being equivalent.

real-or-random: Yeah, I think our intuition is just different. My thought was that if `uint32_t` is at least as large as `unsigned int`, then no promotion can happen at all, neither to `signed int` nor to `unsigned int`.

roconnor-blockstream: I think of the rule as if every value of a type can fit within a `signed int`, then it is subject to integer promotion.

sipa: Both seem reasonable and harmless to me.

real-or-random: Wanna ACK then?

elichai (approved):
Looks good.
ACK c0041b5cfca5efb160aa9a5616350069c89a8c29

sipa:
utACK c0041b5cfca5efb160aa9a5616350069c89a8c29

PR #821, opened 2020-09-19
travis: Explicitly set --with-valgrind

jonasnick:
Also remove CPPFLAGS=-DVALGRIND because that's redundant with when
configured with --enable-valgrind.


.travis.yml
```diff
@@ -17,7 +17,7 @@ compiler:
   - gcc
 env:
   global:
-    - WIDEMUL=auto  BIGNUM=auto  ENDOMORPHISM=no  STATICPRECOMPUTATION=yes  ECMULTGENPRECISION=auto  ASM=no  BUILD=check  EXTRAFLAGS=  HOST=  ECDH=no  RECOVERY=no SCHNORRSIG=no EXPERIMENTAL=no CTIMETEST=yes BENCH=yes ITERS=2
+    - WIDEMUL=auto  BIGNUM=auto  ENDOMORPHISM=no  STATICPRECOMPUTATION=yes  ECMULTGENPRECISION=auto  ASM=no  BUILD=check  ENABLE_VALGRIND=yes RUN_VALGRIND=no EXTRAFLAGS=  HOST=  ECDH=no  RECOVERY=no SCHNORRSIG=no EXPERIMENTAL=no CTIMETEST=yes BENCH=yes ITERS=2
```
luke-jr, line 20: Should it run valgrind to be sure it worked?

luke-jr (commented):
Nit: I think it would be better to rename `VALGRIND` in a separate commit

.travis.yml
jonasnick: It runs valgrind if `RUN_VALGRIND` is `yes` which happens in lines 39 and 40.

jonasnick:
Split into two commits.

.travis.yml
real-or-random: ah sure, I missed that this is the global line

sipa:
utACK 4eecb4d6ef6d4f18be8870a5929feb1dae376d15

elichai:
ACK 4eecb4d6ef6d4f18be8870a5929feb1dae376d15

PR #822, opened 2020-09-21
Increase precision of g1 and g2.

roconnor-blockstream:
This allows us to shift by 256+128 = 384 bits, which is a multiple of the limb size of the scalar representation.
This also happens to be the most precision possible for g2 that still fits into a 256-bit value.


roconnor-blockstream:
Historically speaking the values of `g1` and `g2` were set in https://github.com/bitcoin-core/secp256k1/pull/127 where they were copied from https://github.com/bitcoin-core/secp256k1/pull/21.  However https://github.com/bitcoin-core/secp256k1/pull/21 was using `secp256k1_num_mul` where there was some advantage to using small values for `g1` and `g2`. But with `secp256k1_scalar_mul_shift_var`, there is no advantage to using small values for `g1` and `g2`.

Instead there is a very slight advantage of shifting by a multiple of the limb size.

src/scalar_impl.h
```diff
@@ -270,19 +270,30 @@ static void secp256k1_scalar_split_lambda(secp256k1_scalar *r1, secp256k1_scalar
...
+ * Both lambda and beta are primitive cube roots of unity.  That is lamba^3 == 1 mod n and
+ * beta^3 == 1 mod p, where n is the curve order and p is the field order.
+ *
+ * Futhermore, because (X^3 - 1) = (X - 1)(X^2 + X + 1), the primitive cube roots of unity are
+ * roots of X^2 + X + 1.  Therefore lambda^2 + lamba == -1 mod n and beta^2 + beta == -1 mod p.
+ * (The other primitive cube roots of unity are lambda^2 and beta^2 respectively.)
+ *
+ * Let l = -1/2 + i*sqrt(3)/2, the complex root of X^2 + X + 1. We can define a ring
+ * homomorphism phi : Z[l] -> Z_n where phi(a + b*l) == a + b*lambda mod n. The kernel of phi
+ * is a lattice over Z[l]. A reduced basis of this lattice is generated by the values  a1 + b1*l
```
roconnor-blockstream, line 282: Rephrase as `This lattice is generated by the reduced basis {a1 + b1*l, a2 + b2*l} where:`.

src/scalar_impl.h
```diff
@@ -294,13 +305,130 @@ static void secp256k1_scalar_split_lambda(secp256k1_scalar *r1, secp256k1_scalar
...
+ *
+ * From Corollary 4, Corollary 5 and Lemma 6 we can conclude that
+ *
+ *  - either r1 <= (a1 + a2 + 1) / 2 or n - (a1 + a2 + 1)/2 <= r1
+ *  - either r2 <= (-b1 + b2) / 2 or n - (-b1 + b2)/2 <= r2.
  *
- * (Note that 'd' is also equal to the curve order here because [a1,b1] and [a2,b2] are found
- * as outputs of the Extended Euclidean Algorithm on inputs 'order' and 'lambda').
+ * Q.E.D.
  *
```
roconnor-blockstream, line 431: Delete this blank line.

src/scalar_impl.h
```diff
@@ -294,13 +305,130 @@ static void secp256k1_scalar_split_lambda(secp256k1_scalar *r1, secp256k1_scalar
...
+ * <=   {triangle inequality}
+ *    (-b1)*|a*b2/n - c1| + b2*|a*(-b1)/n - c2|
+ * <    {Lemma 1 and Lemma 2}
+ *    (-b1)*(2^-1 + 2^-129) + b2*(2^-1 + 2^-129)
+ *  =
+ *    (-b1 + b2)(2^-1 + 2^-129)
+ * <    {calculation}
+ *    (-b1 + b2 + 2)/2
+ *
+ * Corollary 4: |k2| <= (-b1 + b2)/2.
```
sipa, line 402: Reuse of number 4.

src/scalar_impl.h
```diff
@@ -294,13 +305,130 @@ static void secp256k1_scalar_split_lambda(secp256k1_scalar *r1, secp256k1_scalar
...
+ * g1 = round((2^384)*b2/d)
+ * g2 = round((2^384)*(-b1)/d)
+ *
+ * (Note that 'd' is also equal to the curve order, n, here because [a1,b1] and [a2,b2]
+ * can be found as outputs of the Extended Euclidean Algorithm on inputs 'n' and 'lambda').
+ *
+ * The function below splits a in r1 and r2, such that
+ * - r1 + lambda * r2 == a (mod n)
+ * - either 0 <= r1 <= (a1 + a2 + 1) / 2 or n - (a1 + a2 + 1)/2 <= r1 < n.
+ * - either 0 <= r2 <= (-b1 + b2) / 2 or n - (-b1 + b2)/2 <= r2 < n.
```
sipa, line 317: Maybe add here that this implies either `r1` or `-r1` is in range `[0..2^128-1]` (and similarly for `r2`), as that's what we ultimately care about.

src/scalar_impl.h
```diff
@@ -294,13 +305,130 @@ static void secp256k1_scalar_split_lambda(secp256k1_scalar *r1, secp256k1_scalar
...
+ *
+ * Proof.
+ *
+ * Let
+ *  - epsilon1 = |g1/2^384 - b2/d|
```
sipa, line 322: FWIW, epsilon1 and epsilon2 can be computed exactly, and are smaller than 2^-385.

* epsilon1 = 9722325547791800083761969591266328053001952849235129972941425083593053056369/4562440617622195218641171605700291324876190276497652512765600377628314188904572302704052175358268500448500017515730822496526831756972782880663319377092625557245625225500848619780486929676500992 =~ 2^-387.574091440504.
* epsilon2 = 31846006334326197673727480983349989536770120240999881061201988246081714647887/4562440617622195218641171605700291324876190276497652512765600377628314188904572302704052175358268500448500017515730822496526831756972782880663319377092625557245625225500848619780486929676500992 =~ 2^-385.862352326689.

src/scalar_impl.h
```diff
@@ -294,13 +305,130 @@ static void secp256k1_scalar_split_lambda(secp256k1_scalar *r1, secp256k1_scalar
...
+ * <=   {triangle inequality}
+ *    |c1 - a*g1/2^384| + |a*g1/2^384 - a*b2/d|
+ *  =
+ *    |c1 - a*g1/2^384| + a*|g1/2^384 - b2/d|
+ * <=   {property of rounding in c1 & definition of epsilon1
+ *    2^-1 + a*epsilon1
+ * <    {a < 2^256 and epsilon1 < 2^-385}
+ *    2^-1 + 2^256 * 2^-385
+ *  =
+ *    2^-1 + 2^-129
```
sipa, line 341: When using the exact value of epsilon1 here, I get 2^-10 + 2^-131.574091440504...

src/scalar_impl.h
```diff
@@ -294,13 +305,130 @@ static void secp256k1_scalar_split_lambda(secp256k1_scalar *r1, secp256k1_scalar
...
+ * <=   {triangle inequality}
+ *    |c1 - a*g2/2^384| + |a*g2/2^384 - a*(-b1)/d|
+ *  =
+ *    |c1 - a*g2/2^384| + a*|g2/2^384 - (-b1)/d|
+ * <=   {property of rounding in c1 & definition of epsilon2
+ *    2^-1 + a*epsilon2
+ * <    {a < 2^256 and epsilon2 < 2^-385}
+ *    2^-1 + 2^256 * 2^-385
+ *  =
+ *    2^-1 + 2^-129
```
sipa, line 357: When using the exact value of epsilon2 here, I get 2^-1 + 2^-129.862352326689...

src/scalar_impl.h
```diff
@@ -294,13 +305,130 @@ static void secp256k1_scalar_split_lambda(secp256k1_scalar *r1, secp256k1_scalar
...
+ *  =
+ *    |a1*(a*b2/n - c1) + a2*(a*(-b1)/n - c2)|
+ * <=   {triangle inequality}
+ *    a1*|a*b2/n - c1| + a2*|a*(-b1)/n - c2|
+ * <    {Lemma 1 and Lemma 2}
+ *    a1*(2^-1 + 2^-129) + a2*(2^-1 + 2^-129)
+ *  =
+ *    (a1 + a2)(2^-1 + 2^-129)
+ * <    {calculation}
+ *    (a1 + a2 + 3)/2
```
sipa, line 379: Using exact numbers I get (a1+a2)/2 + 0.3132789511161...

src/scalar_impl.h
roconnor-blockstream: Now that I see these written out like this, I can improve the bound on k1 by 1 using the fact that epsilon1 is so very small.  The bound on k2 is already more or less optimal with this proof technique.

src/scalar_impl.h
roconnor-blockstream: Oh I see you already wrote this.

roconnor-blockstream:
I've updated the proof.  Please note that `epsilon1` and `epsilon2` are now defined to be 2^256 times larger than they were before.  But don't worry, they are still pretty small values.

roconnor-blockstream:
I've added a `secp256k1_split_lambda_verify` function to verify the results against the bounds computed in the proof.

Be aware that the verification fails with GCC 9.3 and GCC 10.2 because these compilers are badly broken.  See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95189.

src/scalar_impl.h
```diff
@@ -293,16 +304,161 @@ static void secp256k1_scalar_split_lambda(secp256k1_scalar *r1, secp256k1_scalar
...
+ * From Lemma 3, Lemma 4, Lemma 5 and the definition of r2, we can conclude that
+ *
+ *  - either r1 < 2^128 or -r1 mod n < 2^128
+ *  - either r2 < 2^128 or -r2 mod n < 2^128.
+ *
+ * Q.E.D.
  */
 
+#ifdef VERIFY
+static void secp256k1_split_lambda_verify(const secp256k1_scalar *r1, const secp256k1_scalar *r2, const secp256k1_scalar *a) {
```
roconnor-blockstream, line 424: should be called `secp256k1_scalar_split_lambda_verify`.

src/scalar_impl.h
```diff
@@ -293,16 +304,161 @@ static void secp256k1_scalar_split_lambda(secp256k1_scalar *r1, secp256k1_scalar
  * The derivation is described in the paper "Efficient Software Implementation of Public-Key
  * Cryptography on Sensor Networks Using the MSP430X Microcontroller" (Gouvea, Oliveira, Lopez),
  * Section 4.3 (here we use a somewhat higher-precision estimate):
- * d = a1*b2 - b1*a2
- * g1 = round((2^272)*b2/d)
- * g2 = round((2^272)*b1/d)
+ * d = a1*b2 - b1*a
```
real-or-random, line 307:
```suggestion
 * d = a1*b2 - b1*a2
```

src/scalar_impl.h
```diff
@@ -293,16 +304,161 @@ static void secp256k1_scalar_split_lambda(secp256k1_scalar *r1, secp256k1_scalar
...
+ * g1 = round(2^384 * b2/d)
+ * g2 = round(2^384 * (-b1)/d)
+ *
+ * (Note that 'd' is also equal to the curve order, n, here because [a1,b1] and [a2,b2]
+ * can be found as outputs of the Extended Euclidean Algorithm on inputs 'n' and 'lambda').
+ *
+ * The function below splits a in r1 and r2, such that
```
real-or-random, line 314: In general, `a` and  (`a1`,`a2`) are confusing because they're not related. Maybe we can just rename `a1`,`a2`? They're used only in comments, so that should be easy.

real-or-random (commented):
Nice. I can create a follow up PR to add a sage script that computes all the constants here. 

One could write optimized multiplication routines for the current shorter `g1`, `g2`. But that's a lot of work for a tiny bit of performance so  unless we have this, I agree that the new constants are an improvement.

real-or-random:
> Be aware that the verification fails with GCC 9.3 and GCC 10.2 because these compilers are badly broken. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95189.

Yeah, I don't think we want "expected failures". But tbh, I don't know what to do. We could disable the tests for these GCC versions but this is not a good idea either. We could also write our own memcmp for the tests but this is strange, too.

src/scalar_impl.h
roconnor-blockstream: How do we feel about renaming the input parameter as `k`?

real-or-random: Totally fine, that matches the papers. I just didn't suggest this one because it's more work.

roconnor-blockstream: And rename it in the header file, `scalar.h` too?

roconnor-blockstream:
What would you do if we already had such a test in place before GCC released their broken compilers?

Keep in mind these broken compilers are already compromising our existing tests @ https://github.com/bitcoin-core/secp256k1/blob/770b3dcd6f1edb0a6355a55d92b7bea1e9524042/src/tests.c#L3458-L3460 for example.  It is simply that those tests are improperly succeeding instead of improperly failing.

real-or-random:
@roconnor-blockstream That's different. 

If the test succeeds incorrectly, then that's a problem because that single test does not work. And yes, maybe we should fix this.  

If the tests fail incorrectly, it's much worse. Then the test binary aborts early and a lot of the tests are simply skipped. Moreover,  people will be trained to ignore failing tests / Travis, which is bad.

elichai:
What about the other memcmp's we have in actual production code? (bip340 tag, tweak add check, scratch impl, sha256 selftest) does this bug affect those too?

roconnor-blockstream:
The fix for the tests that succeed incorrectly is the same as for fixing these tests.  Stop using GCC 9.3 and 10.2 or find some flags to disable the use of intrinsics for memcmp for those particular compiler versions that are broken.

real-or-random:
> The fix for the tests that succeed incorrectly is the same as for fixing these tests.

I agree.

>  Stop using GCC 9.3 and 10.2

We're not in control. We control our code, not the compiler. We can blacklist these compilers but that's ugly and people would override it. 

> or find some flags to disable the use of intrinsics for memcmp for those particular compiler versions that are broken.

`-fno-builtin-memcmp` should do the trick but then we need to detect GCC versions in autoconf. That's also  addiitonal code (autotools code) and work, and not everyone uses autoconf.

src/scalar_impl.h
real-or-random: Yes, consistency with the header is a good thing IMO.

roconnor-blockstream:
I propose we argue about `-fno-builtin-memcmp` for a few weeks until GCC 9 and GCC 10 have new point releases and then ignore the problem.

gmaxwell:
Concept ACK, but the test cannot be included as is-- not even in a couple weeks because the broken compilers have been massively shipped and some users will still be running them for a couple of years (e.g. imagine you setup an offline signer using install media...).

At a *minimum* the tests should add an explicit memcmp bug test so that the error the user gets is a useful report that tells them that their compiler is faulty and not all other tests are bypassed in case the user wants to continue with the broken compiler.  But I think the library should probably ship a workaround (see my comments in #823).

And congrats on finding a compiler bug (though you weren't the first in the case)--  good testing (sadly) finds compiler bugs.  The fact that none of the other tests find this suggest to me that there isn't enough mutation testing going on here (or we just got really unlucky, and none of the mutation hittable differences with constant comparisons have an early null).

real-or-random:
> The fact that none of the other tests find this suggest to me that there isn't enough mutation testing going on here (or we just got really unlucky, and none of the mutation hittable differences with constant comparisons have an early null).

We can for sure have better mutation testing but in our defense, we use `memcmp` in the tests mostly to assert equality and not inequality, and this is a natural thing to do. There are a few inequality checks with constants, e.g.,
https://github.com/bitcoin-core/secp256k1/blob/d7838ba6a6ac77cec173080f20efcd0e311ebfaa/src/modules/extrakeys/tests_impl.h#L432
But in this case, GCC emits a `memcmp` call, and I can't convince it to emit wrong code. It's not that easy to hit this bug, it seems to depend on specific contexts and optimizations.

roconnor-blockstream:
Tests now pass. :angry:

src/scalar_impl.h
real-or-random: I think there are a few places in this comment where it's still `a`.

sipa:
utACK d91408ea4c3661c172f6fa2da5f586b5096146eb

sipa:
If we want to add unit tests for lambda splitting that result in numbers very close to the bound:

* split(0x7fffffffffffffffffffffffffffffff5d576e7357a4501ddfe92f46681b20a0) = (0xa2a8918ca85bafe22016d0b917e4dd76, -0x59de565a2c9d0e2d4373f7623c1d7cd7) [where 0xa2a89... = k1_bound]
* split(0x2c9c52b33fa3cf1f5ad9e3fd77ed9ba54b294b893722e9a500e698ca4cf7632e) = (0x7221bf6b0087441437aa3fd4855ff261, 0x8a65287bd47179fb2be08846cea267eb) [where 0x8a52... = k2_bound-1]

gmaxwell:
That scalar that creates maximal values might be a useful scalar for ecmult tests and as a private key for ECDH too bad you don't achieve the k2_bound.

sipa:
Here is a commit to add tests for splitting and EC multiplications with those scalars: https://github.com/sipa/secp256k1/commits/202009_pr822. The endomorphism test fails when I introduce small deviations in g1/g2 (though only when the error is around ~2^128 at least).

Feel free to cherry pick, or I can open as a separate PR later.

sipa:
As it appears that all these large-output values follow a simple pattern (`(((ORDER+a)/2 + LAMBDA*b) % ORDER)` for small signed integers a and b), I've expanded the set a bit and added a comment.

roconnor-blockstream:
Should I add
```
#ifdef VERIFY
#include <string.h>
#endif
```
?

real-or-random:
> Should I add
> 
> ```
> #ifdef VERIFY
> #include <string.h>
> #endif
> ```
> 
> ?

I don't have strong opinions (and I assume noone has). If you add it, I'll remove it in #825 . :P

sipa:
> Should I add

If we expect to merge #825 afterwards, no need.

roconnor-blockstream:
I added the include of `string.h`.  I've updated the documentation.  I've pick @sipa's tests.

I expect the k2_bound is unachievable.  With some effort could probably enumerate the small number possible candidate points that could possibly achieve the k2_bound and eliminate them.

gmaxwell:
looks good to me

sipa:
ACK 57d4b0df2c1ba9c6718607b3c3175b25b93b120b

Reviewed the diff with the earlier version, created a merge of this with #826 and built/tested it (40000 runs of the unit tests with it).

@roconnor-blockstream It's my belief that the k2_bound is indeed unreachable; if it was, one of the inputs in the `scalars_near_split_bounds` constant now should trigger it.

gmaxwell:
For your consideration, I've made some test improvements: https://github.com/gmaxwell/secp256k1/commit/6c3e1fab91766aa0c424b678371a0c8099726655

gmaxwell:
The bounds check in secp256k1_scalar_split_lambda_verify now doesn't avoid triggering the GCC bug, it just falsely passes in the presence of it (e.g. set the two bounds to all zeros and it keeps passing).  Not ideal considering essentially all developers are using impacted compilers.  I didn't realize this at first and burnt a few cpu years running tests that could not fail. It appears to be right with the code in #825.

sipa:
PR that merges this (including @gmaxwell's extra tests) with others: #830

elichai:
Is it expected to see any meaningful difference in performance here? because I can't see any in the benchmakrs

real-or-random:
No, this just saves a few instructions and we run this only once per ecmult.

sipa:
Merged as part of #830.

PR #824, opened 2020-09-24
Return false from secp256k1_gej_is_valid_var when z is 0

nothingmuch:
Changes `secp256k1_gej_is_valid_var` to return false for cleared structs, which do not represent points on the curve.

Commits should probably be squashed to respect git bisect.

Context: https://github.com/zkSNACKs/WalletWasabi/pull/4430#issuecomment-698084278

### DIFF
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -271,7 +271,7 @@ static int secp256k1_gej_is_infinity(const secp256k1_gej *a) {
 
 static int secp256k1_gej_is_valid_var(const secp256k1_gej *a) {
     secp256k1_fe y2, x3, z2, z6;
-    if (a->infinity) {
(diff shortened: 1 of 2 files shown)


jonasnick (commented):
Thanks @nothingmuch. `secp256k1_gej_is_valid_var` is going to be removed in #808 because it's not meaningfully used in the library.

nothingmuch:
Ah, good to know. It's a shame search results don't cover PR diffs, as it would have come up for me...

I found this when I broke a contrived unit test, so it doesn't really matter for us. In anticipation of the corresponding NBitcoin API getting removed I think will just rework our code to only validate affine points as that is what is passed across program boundaries and where validation is needed, whereas Jacobian representation should only be the the result of conversion or group operations on validated points.

Thanks!

PR #825, opened 2020-09-24
Switch to our own memcmp function

### LINKED ISSUE #823: memcmp may be miscompiled by GCC
What about the other memcmp's we have in actual production code? (bip340 tag, tweak add check, scratch impl, sha256 selftest) does this bug affect those too?

_Originally posted by @elichai in https://github.com/bitcoin-core/secp256k1/pull/822#issuecomment-697221468_

context: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=95189

real-or-random:
Fixes #823.

Fwiw, I considered to make this a constant-time function but didn't do it in the end. Most constant-time `memcmp` replacements are actually only inequality checks but we need a real `memcmp` which outputs which one of the arguments is larger.

If you look at crypto libraries, there are a few candidates but I found this overkill for our use case 
https://github.com/jedisct1/libsodium/blob/a8fa837aacd310bc08fa72705a738fafc2513125/src/libsodium/sodium/utils.c#L239
https://cvsweb.openbsd.org/cgi-bin/cvsweb/src/lib/libc/string/timingsafe_memcmp.c?rev=1.1&content-type=text/x-cvsweb-markup

If we need a constant-time function in the future, it'll probably be only for equality/inequality, and then we can add the standard "OR of byte-wise XOR" implementation.

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -216,6 +216,21 @@ static SECP256K1_INLINE void memczero(void *s, size_t len, int flag) {
     }
 }
 
+/** Semantics like memcmp. Variable-time.
+   We use this to avoid possible compiler bugs with memcmp, e.g., a bug in certain versions of GCC 9 and 10. */
+int secp256k1_memcmp_var(const void *s1, const void *s2, size_t n) {
+    const unsigned char *p1 = s1, *p2 = s2;
+    size_t i;
+
+    for (i = 0; i < n; i++) {
+        int diff = p1[i] - p2[i];
+        if (diff != 0) {
+            return diff;
+        }
+    }
+    return 0;
+}
+
 /** If flag is true, set *r equal to *a; otherwise leave it. Constant-time.  Both *r and *a must be initialized and non-negative.*/
 static SECP256K1_INLINE void secp256k1_int_cmov(int *r, const int *a, int flag) {
     unsigned int mask0, mask1, r_masked, a_masked;
--- a/src/modules/ecdh/tests_impl.h
+++ b/src/modules/ecdh/tests_impl.h
@@ -80,7 +80,7 @@ void test_ecdh_generator_basepoint(void) {
         /* compute "explicitly" */
         CHECK(secp256k1_ec_pubkey_serialize(ctx, point_ser, &point_ser_len, &point[1], SECP256K1_EC_UNCOMPRESSED) == 1);
         /* compare */
-        CHECK(memcmp(output_ecdh, point_ser, 65) == 0);
+        CHECK(secp256k1_memcmp_var(output_ecdh, point_ser, 65) == 0);
 
         /* compute using ECDH function with default hash function */
         CHECK(secp256k1_ecdh(ctx, output_ecdh, &point[0], s_b32, NULL, NULL) == 1);
@@ -90,7 +90,7 @@ void test_ecdh_generator_basepoint(void) {
         secp256k1_sha256_write(&sha, point_ser, point_ser_len);
         secp256k1_sha256_finalize(&sha, output_ser);
         /* compare */
-        CHECK(memcmp(output_ecdh, output_ser, 32) == 0);
+        CHECK(secp256k1_memcmp_var(output_ecdh, output_ser, 32) == 0);
     }
 }
 
--- a/src/modules/extrakeys/main_impl.h
+++ b/src/modules/extrakeys/main_impl.h
@@ -121,7 +121,7 @@ int secp256k1_xonly_pubkey_tweak_add_check(const secp256k1_context* ctx, const u
     secp256k1_fe_normalize_var(&pk.y);
     secp256k1_fe_get_b32(pk_expected32, &pk.x);
 
-    return memcmp(&pk_expected32, tweaked_pubkey32, 32) == 0
+    return secp256k1_memcmp_var(&pk_expected32, tweaked_pubkey32, 32) == 0
             && secp256k1_fe_is_odd(&pk.y) == tweaked_pk_parity;
 }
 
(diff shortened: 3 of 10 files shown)


src/util.h
```diff
@@ -216,6 +216,21 @@ static SECP256K1_INLINE void memczero(void *s, size_t len, int flag) {
     }
 }
 
+/** Semantics like memcmp. Variable-time.
+   We use this to avoid possible compiler bugs with memcmp, e.g., a bug in certain versions of GCC 9 and 10. */
+int secp256k1_memcmp_var(const void *s1, const void *s2, size_t n) {
```
sipa, line 221: Make `static` please.

src/util.h
```diff
@@ -216,6 +216,21 @@ static SECP256K1_INLINE void memczero(void *s, size_t len, int flag) {
     }
 }
 
+/** Semantics like memcmp. Variable-time.
+   We use this to avoid possible compiler bugs with memcmp, e.g., a bug in certain versions of GCC 9 and 10. */
```
sipa, line 220: Link to the bug? So that it's easy to re-assess in the future whether this is still needed.

roconnor-blockstream:
Hmm, wasn't the plan to get rid of `string.h` entirely?  Thus needed to replace `memset` as well?

src/util.h
roconnor-blockstream: Are we required to follow the `const void *` signature?  Should we make it `const char *`?

real-or-random:
> Wasn't the plan to get rid of `string.h` entirely? Thus needed to replace `memset` as well?

@gmaxwell said in #823:
> (perhaps even memcpy too, and eliminate the string.h usage)

But this was not discussed further, so I sticked to only `memcmp` for this PR.

I think I'd like to see some benchmarks before we decide to get rid of more functions. `memcpy` is used in the hashing code, and `memset` will be used a lot in cleaning up secret after #636 (which I have been procrastinating on for too long, I know...). And this can make a difference, see e.g., https://github.com/bitcoin-core/secp256k1/pull/636#issuecomment-516031083.

roconnor-blockstream:
I don't understand what this PR is trying to achieve.  You are replacing occurrences of `memcmp` that are not believed to be miscompiled by GCC 9 & 10.

real-or-random:
> I don't understand what this PR is trying to achieve. You are replacing occurrences of `memcmp` that are not believed to be miscompiled by GCC 9 & 10.

I'm not the best person to answer this because I tend to agree, see https://github.com/bitcoin-core/secp256k1/issues/823#issuecomment-699082172. But since this PR does not make our code worse (no performance penalty, small amount of added code), I'm ok with working on it.

roconnor-blockstream:
I think the worry is about introducing code that unwittingly encounters this bug in the future (https://github.com/bitcoin-core/secp256k1/issues/823#issuecomment-699459852).  I thought maybe removing `string.h` might help achieving that because without the header file, it is difficult to use `memcmp` (though how to stop the reintroduction of the header?).  Otherwise, it seems like we need a linter step or something to disallow `memcmp` added to this PR.

real-or-random:
Well we could add some linter script that uses grep to check whether all instances of the `memcmp` string are actually `secp256k1_memcmp`. But that's ... weird. It doesn't understands comments, other identifiers, etc. Real linting is only possible in the compiler. It would be interesting to look into tools like clang-query which can give you a lot of insights internal from the compiler, and I'd love to use this as a linter for some other stuff, e.g., integer divisions. But this sounds like it's not in the scope of this PR. :P

real-or-random:
>  It would be interesting to look into tools like clang-query which can give you a lot of insights that are normally internal to the compiler, and I'd love to use this as a linter for some other stuff, e.g., integer divisions. But this sounds like it's not in the scope of this PR. :P

If someone wants to give it a try, this looks like a nice start: https://stackoverflow.com/a/31740134 maybe it's indeed a one-liner.

sipa:
You could also define memcmp to be a macro that just turns into an #error "Use secp256k1_memcmp_var instead please".

real-or-random:
> You could also define memcmp to be a macro that just turns into an #error "Use secp256k1_memcmp_var instead please".

Indeed. I'm not sure if I like this because that's UB according to the standard. Though I don't think any sane compiler in this world will have any issues with this.

roconnor-blockstream:
Technically the existence of a function named `memczero` is also UB, so I don't know by what measure we characterize some forms of UB over others.

real-or-random:
> Technically the existence of a function named `memczero` is also UB, so I don't know by what measure we characterize some forms of UB over others.

If I had known that this is UB when it we introduced `memczero`, I would have pointed this out.

gmaxwell:
> Technically the existence of a function named memczero

huh? wtf?  google returns nothing on "memczero" (except links to this library and copycats). How is this UB?

gmaxwell:
Damnit, I forgot that "mem[lowercase]" was reserved, fix it-- don't just gripe.

roconnor-blockstream:
I didn't mean to sound like I was griping.  I didn't know about the "mem[lowercase]" reserved function name thing until today. ... My reaction upon reading about it (and function names beginning with "is[lowercase]" and "to[lowercase]") was C's definition of UB was simply reaching absurd proportions and that it wasn't worth fixing.  From your reaction, it would seem that it isn't as absurd as I thought.

gmaxwell:
The bounds check in secp256k1_scalar_split_lambda_verify now doesn't avoid triggering the GCC bug, it just falsely passes in the presence of it (e.g. set the two bounds to all zeros and it keeps passing).  Not ideal considering essentially all developers are using impacted compilers.  I didn't realize this at first and burnt a few cpu years running tests that could not fail.

src/util.h
sipa: It seems that would need more code changes, as callers in some places expect to be able to pass a pointer to other objects.

sipa:
PR that merges this (with some comments addressed) with #822 and #826: #830

src/util.h
roconnor-blockstream: Oh I see.  This `void` thing is really subversive.

real-or-random:
Closing in favor of #830

PR #826, opened 2020-09-26
Rip out non-endomorphism code

sipa:
As the patent on the GLV optimization has expired, there is no need to keep the slower non-endomorphism code around anymore.


elichai:
There's nothing I love more than a PR that only deletes code :)

sipa:
> There's nothing I love more than a PR that only deletes code :)

Hey it also improves some comments.

.travis.yml
```diff
@@ -17,33 +17,29 @@ compiler:
...
     - ECMULTGENPRECISION=8
-    - VALGRIND=yes ENDOMORPHISM=yes BIGNUM=no ASM=x86_64 EXPERIMENTAL=yes ECDH=yes  RECOVERY=yes EXTRAFLAGS="--disable-openssl-tests" CPPFLAGS=-DVALGRIND BUILD=
     - VALGRIND=yes                  BIGNUM=no ASM=x86_64 EXPERIMENTAL=yes ECDH=yes  RECOVERY=yes EXTRAFLAGS="--disable-openssl-tests" CPPFLAGS=-DVALGRIND BUILD=
 matrix:
   fast_finish: true
   include:
     - compiler: clang
       os: linux
-      env: HOST=i686-linux-gnu ENDOMORPHISM=yes
+      env: HOST=i686-linux-gnu
```
elichai, line 42: This job can be completely removed, it's exactly the same as the one before

.travis.yml
```diff
@@ -63,7 +59,7 @@ matrix:
             - libtool-bin
             - libc6-dbg:i386
     - compiler: gcc
-      env: HOST=i686-linux-gnu ENDOMORPHISM=yes
```
elichai, line 66: This job can be completely removed, it's exactly the same as the one before

.travis.yml
sipa: One has gmp, the other doesn't. (same comment applies below).

elichai: Oh you're right, my mistake.

elichai:
tACK a523e2c9965ff43939abad555fcb8a8aed684d8e

two nits:
1. Can you also remove from here: https://github.com/bitcoin-core/secp256k1/blob/master/contrib/travis.sh#L16
2. Can you update the README to remove the "optionally":
> Optionally (off by default) use secp256k1's efficiently-computable endomorphism to split the P multiplicand into 2 half-sized ones.

practicalswift:
ACK a523e2c9965ff43939abad555fcb8a8aed684d8e (modulo merge conflict): diff looks correct

sipa:
Rebased, and:

> 1. Can you also remove from here: https://github.com/bitcoin-core/secp256k1/blob/master/contrib/travis.sh#L16
> 2. Can you update the README to remove the "optionally":

Done.

gmaxwell:
Looks good to me,  endo only increases a stripped Os build for me by 456 bytes-- so insignificant esp with respect to the performance difference, so I don't see any reason to keep the other code around even for code size constrained applications (you can get 456 bytes of savings in other ways with less performance impact).

jonatack (commented):
46f9e0e745accde97742cb542d3c779f5b99b4a8 patch diff LGTM

practicalswift:
re-ACK 46f9e0e745accde97742cb542d3c779f5b99b4a8

laanwj:
ACK 46f9e0e745accde97742cb542d3c779f5b99b4a8

> Looks good to me, endo only increases a stripped Os build for me by 456 bytes-- so insignificant esp with respect to the performance difference

I agree. I also looked at it from that angle, but this is a negligible increase in code size.

elichai:
re-ACK 46f9e0e745accde97742cb542d3c779f5b99b4a8

benthecarman:
ACK 46f9e0e

sipa:
I'd like #791 in before this, as should have all assurances we can get that the lambda-split values indeed stay in range.

sipa:
Closing this in favor of #830, which has it rebased on #822 and #825.

Issue #829, opened 2020-10-06
Rename memczero


roconnor-blockstream:
The list of reserved identifiers in C includes all functions names beginning with `mem` followed by a lowercase letter.  We are using the reserved function name `memczero`, which is undefined behaviour.

gmaxwell:
ACK fixing this.

I would gauge that this is extremely unlikely to ever cause issues (esp, not silent issues);  but the function name is a free choice so better to be pedantically correct.

Ideally the code would be checked by static analysis which is anal enough to catch things like this (some of which might be less benign than this one) and keeping this wrong would just cause the tool to produce uninteresting reports.

Reserved wildcard issues have been fixed before. E.g. #479

elichai:
wtf, as much as UB in C is interesting this is crazy, I though only `__` prefix is UB. sigh.

real-or-random:
If I understand C89 correctly, this is not an issue because `memczero` is static and as such has internal linkage and is not an "external name".

But the "mem..." identifiers are reserved as external names:
4.13: "All external names described below are reserved no matter what headers are included by the program. (...) 

3.1.2 defines "external name" to be "an identifier that has external linkage"
and
3.1.2.2:  "If the declaration of an identifier for an object or a function has file scope and contains the storage-class specifier static, the identifier has internal linkage."


4.1.2: "If the program defines an external identifier with the same name as a reserved external identifier, even in a semantically equivalent form, the behavior is undefined."

The language is pretty imprecise though. C99 and C11 make this clearer by explicitly saying which identifiers are reserved in which contexts. The result is the same in our case. 

That's enough language lawyering for today, even for my standards. No pun intended.

roconnor-blockstream:
C99, 7.3.1: "Each identifier with file scope listed in any of the following subclauses (including the future library directions) is reserved for use as a macro name and as an identifier with file scope in the same name space if any of its associated headers is included."

My reading of this was that defining the `memczero` function in any file that includes the `<string.h>` header is UB.  But I find this wording hard to read, and I may be mistaken.

real-or-random:
Oh yes, I think you're right!

I guess let's just fix it instead of debating the standard (even though the latter may be more entertaining). :D

roconnor-blockstream:
Should we expand the scope of this issue?

https://github.com/bitcoin-core/secp256k1/blob/master/src/ecmult_impl.h#L71

[...]

PR #833, opened 2020-10-15
Add simple static checker based on clang-query

real-or-random:
This add a simple static checker based on clang-query, which is a
tool that could be described as a "clever grep" for abstract syntax
trees (ASTs).

As an initial proof of usefulness, this commit adds these checks:
  - No uses of floating point types.
  - No use of certain reserved identifiers (e.g., "mem(...)", #829).
  - No use of memcmp (#823).

The checks are easily extensible.

The main purpose is to run the checker on CI, and this commit adds
the checker to the Travis CI script.

This currently requires clang-query version at least 10. (However,
it's not required not compile with clang version 10, or with clang
at all. Just the compiler flags must be compatible with clang.)
Clang-query simply uses the clang compiler as a backend for
generating the AST.

In order to determine the compile in which the code is supposed to be
compiled (e.g., compiler flags such as -D defines and include paths),
it reads a compilation database in JSON format. There are multiple ways
to generate this database. The easiest way to obtain such a database is
to use a tool that intercepts the make process and build the database.

On Travis CI, we currently use "bear" for this purpose. It's a natural
choice because there is an Ubuntu package for it. If you want to run
this locally, bear is a good choice but other tools such as compiledb
(Python) are available.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,8 @@ libtool
 *~
 *.log
 *.trs
+compile_commands.commands.json
+compile_commands.json
 src/libsecp256k1-config.h
 src/libsecp256k1-config.h.in
 src/ecmult_static_context.h
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,9 +9,11 @@ osx_image: xcode10.1
 addons:
   apt:
     packages:
+      - clang-tools-10
       - libgmp-dev
       - valgrind
       - libtool-bin
+      - bear
 compiler:
   - clang
   - gcc
@@ -43,10 +45,12 @@ matrix:
       addons:
         apt:
           packages:
+            - clang-tools-10
             - gcc-multilib
             - libgmp-dev:i386
             - valgrind
             - libtool-bin
+            - bear
             - libc6-dbg:i386
     - compiler: clang
       env: HOST=i686-linux-gnu
@@ -54,9 +58,11 @@ matrix:
       addons:
         apt:
           packages:
+            - clang-tools-10
             - gcc-multilib
             - valgrind
             - libtool-bin
+            - bear
             - libc6-dbg:i386
     - compiler: gcc
       env: HOST=i686-linux-gnu
@@ -64,6 +70,7 @@ matrix:
       addons:
         apt:
           packages:
+            - clang-tools-10
             - gcc-multilib
             - valgrind
             - libtool-bin
@@ -74,6 +81,7 @@ matrix:
       addons:
         apt:
           packages:
+            - clang-tools-10
             - gcc-multilib
             - libgmp-dev:i386
             - valgrind
@@ -104,5 +112,7 @@ after_script:
     - cat ./exhaustive_tests.log
     - cat ./valgrind_ctime_test.log
     - cat ./bench.log
+    - cat ./clang-query.log
+    - cat ./compile_commands.json
     - $CC --version
     - valgrind --version
--- a/clang-query.sh
+++ b/clang-query.sh
@@ -1,0 +1,39 @@
+#!/bin/sh
+
+set -u
+
+matcher=$(cat <<'EOF'
+set print-matcher true
+enable output detailed-ast
+
+### Expressions of any floating point type (unless in a system header)
+match expr(allOf(unless(isExpansionInSystemHeader()), hasType(realFloatingPointType())))
+
+### Calls to memcmp (secp256k1_memcmp_var should be used instead)
+match callExpr(callee(functionDecl(hasName("memcmp"))))
+
+### Reserved identifiers (unless in a system header) with external linkage or at file scope.
+# Any function is in file scope. Any variable with static storage (unless static local variable) is in file scope.
+# Allowed exceptions: __builtin_expect
+# We need the "::" due to LLVM bug 47879.
+match namedDecl(allOf(unless(isExpansionInSystemHeader()), anyOf(hasExternalFormalLinkage(), functionDecl(), varDecl(allOf(hasStaticStorageDuration(), unless(isStaticLocal())))), allOf(matchesName("^::(_|((mem|is|to|str|wcs)[a-z]))"), unless(hasAnyName("__builtin_expect")))))
+
+### Reserved type names (unless in a system header)
+# Allowed exceptions: uint128_t, int128_t, __int128_t, __uint128_t (the latter two are "implicit", i.e., added by the compiler)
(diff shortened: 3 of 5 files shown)


real-or-random:
Here's another interesting one:
`match binaryOperator(allOf(unless(isExpansionInSystemHeader()), hasOperatorName("/")))`

Unfortunately we use quite a few divisions, so this does not lead anywhere.

real-or-random:
grml, apparently clang-query uses a `compile_commands.json` file to infer compile options/flags, which I had lying around in my local tree from some old experiments in 2018... So I need to rework this.

gmaxwell:
> Unfortunately we use quite a few divisions, so this does not lead anywhere.

Is there any way to tell if the divisor is a compile time constant?   We generally don't want divisions ending up in the machine code (though if its in some setup routine it's not that big a deal) as they're non-constant time[1] and slow (e.g. 50-100 cycles on x86, worse on a lot of arm). ... but divisions with constants get converted by the compiler to strength reductions.  I *believe* the codebase is free or nearly free of divisions with non-constant divisors.

I've failed to yet fully impress my practice of always using masking over division onto Pieter yet... but there are some cases where masking doesn't easily work.
 
[1] also, stock valgrind ctime check will not catch divisions involving secret data.  I have a pretty trivial patch to valgrind on my laptop so that it does... and it passed for me as of the last time I ran it.  I'm not really eager to figure out how to make travis use a custom compiled copy of valgrind...

real-or-random:
> > Unfortunately we use quite a few divisions, so this does not lead anywhere.
> 
> Is there any way to tell if the divisor is a compile time constant? We generally don't want divisions ending up in the machine code (though if its in some setup routine it's not that big a deal) as they're non-constant time[1] and slow (e.g. 50-100 cycles on x86, worse on a lot of arm). ... but divisions with constants get converted by the compiler to multiples via strength reduction. I _believe_ the codebase is free or nearly free of divisions with non-constant divisors (maybe there is one in scratch space sizing or something like that).

Right, for example.
```
Match #103:

./src/ecmult_impl.h:1030:27: note: "root" binds here
    *n_batch_points = 1 + (n - 1) / *n_batches;
                          ^~~~~~~~~~~~~~~~~~~~
```

This is the reference for the matcher:
https://clang.llvm.org/docs/LibASTMatchersReference.html

You can match on "constantExpr" but this is not what we want. This matches constant expression as seen from the point of view of the AST, i.e., this matches places where only constant expressions are allowed (such as after `case`). 

This technique in general does not enable us to get insights on what the compiler will do with the code. It's really just pattern matching on the AST, so it's merely a clever grep. (One can do more advanced things with clang tooling but this requires more effort than writing matching expressions.)

What we could do is to check if the divisor is an integer literal, but even this is not precise we have also a few `(x-1)` divisors for example.

real-or-random:
This uses https://github.com/rizsotto/Bear, which intercepts make, to generate a JSON compilation database (see https://clang.llvm.org/docs/JSONCompilationDatabase.html). Bear is nice because we can easily install it on Travis. There are other tools that may work better when used locally, e.g., https://github.com/nickdiego/compiledb  .

real-or-random:
Ready to review.

Travis fails but this is intentionally. See it in action here:

https://travis-ci.org/github/bitcoin-core/secp256k1/jobs/736511002#L1254 (complains about `memczero` declaration, see #829)
https://travis-ci.org/github/bitcoin-core/secp256k1/jobs/736511002#L1313 (complains about reserved `_t` type declared in benchmarks)


I might add a commit that restructures the Travis output in general by adding folds to the main commands:
https://github.com/travis-ci/docs-travis-ci-com/issues/949#issuecomment-276755003
Then all the main commands be folded unless they fail, so see only the errors. Moreover we don't need the `cat`s at the end (which have confused people in the past) and we don't need to ugly hack that I introduced here to make the color output work with `cat`.)

sipa:
This is pretty cool. I'll have to play with it a bit.

real-or-random:
I added a long commit message and updated the PR description to this message.

elichai:
FWIW
> Names beginning with a capital ‘E’ followed a digit or uppercase letter
> Names that begin with either ‘is’ or ‘to’ followed by a lowercase letter
> Names that begin with ‘LC_’ followed by an uppercase letter
> Names of all existing mathematics functions suffixed with ‘f’ or ‘l’
> Names that begin with ‘SIG’ followed by an uppercase letter
> Names that begin with ‘SIG_’ followed by an uppercase letter
> Names beginning with ‘str’, ‘mem’, or ‘wcs’ followed by a lowercase letter
> Names that end with ‘_t’

real-or-random:
@elichai Yeah, where did you get this list from? Is this really all about all names? I thought `E` is only for macros. We can't check for macros using clang-query but we could add a simple grep for  `#define [[:space:]+]E` etc. I also have uncommitted changes to add  https://clang.llvm.org/docs/DiagnosticsReference.html#wreserved-id-macro if the compiler supports it (this covers only `_` and `__` macros).  Not sure if overkill but it's not a lot of work.

elichai:
> @elichai Yeah, where did you get this list from? Is this really all about all names? I thought `E` is only for macros. 

https://www.gnu.org/software/libc/manual/html_node/Reserved-Names.html

About `E` it's weird, in the gcc docs it says `for additional error code names` which are macros, and the exact wording of the standard is:(7.1.4)
> Macros that begin with E and a digit or E and an uppercase letter (followed by any combination of digits, letters, and underscore) may be added to the declarations in the <ermo. h> header.

**But** on 7.1.3 "Reserved identifies" it says:
> Each header declares or defines all identifiers listed in its associated subclause, and optionally declares or defines identifiers listed in its associated future library directions subclause and identifiers which are *always reserved either for any use* or for use as file scope identifiers.

This paper proposing to change that also agrees on that interpretation: http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2572.pdf (See "Future Library Directions")
> However, p1 makes it clear that all identifiers reserved from this subclause are reserved identifiers regardless of what header files are included, meaning that these rules apply to all C code

> In effect, these identifiers are reserved for all uses in C regardless of what header files (if any) are included

elichai:
Standard on 6.8.3 Semantics:
> The identifier immediately following the define is called the _macro name_
and also Aaron Ballman told me:
> macros are defined using an identifier, so "macro name" and "identifier" are interchangable
> "reserved for use as a macro name" is just saying how it's intended to be used, not that macro names are a special thing

So yeah :/

gmaxwell:
It makes sense:  if some macro is going to get called E1234  then you obviously can't have a variable named E1234 without trouble.

PR #835, opened 2020-10-20
Don't use reserved identifiers memczero and benchmark_verify_t

### LINKED ISSUE #829: Rename memczero

real-or-random:
As identified in #829 and #833. Fixes #829.

Since we touch this anyway, this commit additionally makes the
identifiers in the benchmark files a little bit more consistent.

This is necessary before we can merge #833. I preferred a separate PR because it makes it easier to see the results of Travis in #833.

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -260,14 +260,20 @@ static SECP256K1_INLINE void secp256k1_int_cmov(int *r, const int *a, int flag) 
 # define SECP256K1_WIDEMUL_INT128 1
 #elif defined(USE_FORCE_WIDEMUL_INT64)
 # define SECP256K1_WIDEMUL_INT64 1
-#elif defined(__SIZEOF_INT128__)
+#elif defined(UINT128_MAX) || defined(__SIZEOF_INT128__)
 # define SECP256K1_WIDEMUL_INT128 1
 #else
 # define SECP256K1_WIDEMUL_INT64 1
 #endif
 #if defined(SECP256K1_WIDEMUL_INT128)
+# if !defined(UINT128_MAX) && defined(__SIZEOF_INT128__)
 SECP256K1_GNUC_EXT typedef unsigned __int128 uint128_t;
 SECP256K1_GNUC_EXT typedef __int128 int128_t;
+#define UINT128_MAX ((uint128_t)(-1))
+#define INT128_MAX ((int128_t)(UINT128_MAX >> 1))
+#define INT128_MIN (-INT128_MAX - 1)
+/* No (U)INT128_C macros because compilers providing __int128 do not support 128-bit literals.  */
+# endif
 #endif
 
 #endif /* SECP256K1_UTIL_H */
--- a/src/util.h
+++ b/src/util.h
@@ -202,7 +202,7 @@ static SECP256K1_INLINE void *manual_alloc(void** prealloc_ptr, size_t alloc_siz
 #endif
 
 /* Zero memory if flag == 1. Flag must be 0 or 1. Constant time. */
-static SECP256K1_INLINE void memczero(void *s, size_t len, int flag) {
+static SECP256K1_INLINE void secp256k1_memczero(void *s, size_t len, int flag) {
     unsigned char *p = (unsigned char *)s;
     /* Access flag with a volatile-qualified lvalue.
        This prevents clang from figuring out (after inlining) that flag can
--- a/src/bench_sign.c
+++ b/src/bench_sign.c
@@ -12,11 +12,11 @@ typedef struct {
     secp256k1_context* ctx;
     unsigned char msg[32];
     unsigned char key[32];
-} bench_sign;
+} bench_sign_data;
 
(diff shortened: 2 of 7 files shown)


elichai:
FWIW http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2572.pdf

gmaxwell:
@elichai see #833

real-or-random:
@elichai @gmaxwell Mind to review? It's trivial.

elichai:
Just realized,
are these also UB? :O 
https://github.com/bitcoin-core/secp256k1/blob/ac05f61fcf639a15b5101131561620303e4bd808/src/util.h#L269-L270

elichai:
tACK e89278f211a526062745c391d48a7baf782b4b2b

real-or-random:
> Just realized,
> are these also UB? :O
> 
> https://github.com/bitcoin-core/secp256k1/blob/ac05f61fcf639a15b5101131561620303e4bd808/src/util.h#L269-L270

Yeah, probably. We use `__` types but we don't declare them, so this fine. But we define types reserved for the standard lib, so I guess this is evil, yes... And this is probably the only case where we may see a real clash in practice in the future, if some compiler defines both `int128_t` and `__int128_t`. (Even then is unlikely to cause issues because it's hopefully the same type.) What we could do is to guard these typedefs with `defined(INT128_MAX)`. Does this make sense? I could add this to the PR.

IMO the entire libary would be more readable if we used typedefs like i32 and u128. But we'd get a huge diff now, and would interrupt all PRs, pretty much like clang-format.

sipa:
@real-or-random Replacing things everywhere is invasive, but I think can reasonably change `uint128_t` into `uint128`.

gmaxwell:
I had thought the int128 typedefs only ran if they weren't already defined.  I agree clashing here is not great.

Personally I really don't like it when applications make local renames of all the generic standard types:  Then I have to go digging through the code to make sure their custom type is defined to be the thing I think it is and that it's consistently the thing I think it is and isn't something different on some platform or another.  Among other issues you end up with interface uglyness where your custom defined types end up in the interfaces and leak into applications.  It was excusable back before there were standard fixed-size types, but isn't really today, and I think today it mostly still lingers on as copying a style that doesn't make sense anymore.

This is distinct from e.g. making a type for a particular kind of data like a scalar or a field type, because at least that is defined to be fit for a particular purpose.  In most cases using a for-application-type is preferable to using generic types (and, in fact, mandated pretty much universally by MISRA C).

If a generic type needs to be renamed for some compatibility reason-- e.g. there isn't a universally available generic type with the required properties, then sure.

sipa:
ACK e89278f211a526062745c391d48a7baf782b4b2b

real-or-random:
> I had thought the int128 typedefs only ran if they weren't already defined. 

I added a commit that implements this. I also added `MAX` and `MIN` macros

Note: Currently there's no compiler that provides `(u)int128_t` because that will open a can of worms due to `(u)intmax_t`. See http://www.open-std.org/jtc1/sc22/wg14/www/docs/n2425.pdf .

src/util.h
```diff
@@ -260,14 +260,19 @@ static SECP256K1_INLINE void secp256k1_int_cmov(int *r, const int *a, int flag)
...
 #elif defined(USE_FORCE_WIDEMUL_INT64)
 # define SECP256K1_WIDEMUL_INT64 1
-#elif defined(__SIZEOF_INT128__)
+#elif defined(UINT128_MAX) || defined(__SIZEOF_INT128__)
 # define SECP256K1_WIDEMUL_INT128 1
 #else
 # define SECP256K1_WIDEMUL_INT64 1
 #endif
 #if defined(SECP256K1_WIDEMUL_INT128)
+# if !defined(UINT128_MAX) && defined(__SIZEOF_INT128__)
```
sipa, line 269: This `#if` is missing a corresponding `#endif`.

real-or-random: I hate the preprocessor.

jonasnick (commented):
Just to spell it out, the assumption is that a compiler that adds int128_t in the future also defines UINT128_MAX? 

Looks good in general. Not sure how useful it is to define constants we don't use.

real-or-random:
> Just to spell it out, the assumption is that a compiler that adds int128_t in the future also defines UINT128_MAX?

Yes. And this is implied by the standard as far as I understand it. 

> 
> Looks good in general. Not sure how useful it is to define constants we don't use.

Yeah, one of the reasons why I decided for it is that it increases the probability to have a clash be apparent at compile-time (and fail compilation).

jonasnick (commented):
ACK 1f4dd0383807bfb7fef884601357b4c629dfb566

sipa:
utACK 1f4dd0383807bfb7fef884601357b4c629dfb566

PR #839, opened 2020-10-26
Prevent arithmetic on NULL pointer if the scratch space is too small

Fabcien:
If the scratch space is too small when calling
`secp256k1_ecmult_strauss_batch()`, the `state.pre_a` allocation will
fail and the pointer will be `NULL`. This causes `state.pre_a_lam` to be
computed from the `NULL` pointer.

It is also possible that the first allocation to fail is for `state.ps`,
which will cause the failure to occur when in
`secp256k1_ecmult_strauss_wnaf()`.

The issue has been detected by UBSAN using Clang 10:
```
CC=clang \
CFLAGS="-fsanitize=undefined -fno-omit-frame-pointer" \
LDFLAGS="-fsanitize=undefined -fno-omit-frame-pointer" \
../configure

UBSAN_OPTIONS=print_stacktrace=1:halt_on_error=1 make check
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ env:
     - BUILD=distcheck WITH_VALGRIND=no CTIMETEST=no BENCH=no
     - CPPFLAGS=-DDETERMINISTIC
     - CFLAGS=-O0 CTIMETEST=no
+    - CFLAGS="-fsanitize=undefined -fno-omit-frame-pointer" LDFLAGS="-fsanitize=undefined -fno-omit-frame-pointer" UBSAN_OPTIONS="print_stacktrace=1:halt_on_error=1" BIGNUM=no ASM=x86_64 ECDH=yes RECOVERY=yes EXPERIMENTAL=yes SCHNORRSIG=yes CTIMETEST=no
     - ECMULTGENPRECISION=2
     - ECMULTGENPRECISION=8
     - RUN_VALGRIND=yes BIGNUM=no ASM=x86_64 ECDH=yes  RECOVERY=yes EXPERIMENTAL=yes SCHNORRSIG=yes EXTRAFLAGS="--disable-openssl-tests" BUILD=
--- a/src/ecmult_impl.h
+++ b/src/ecmult_impl.h
@@ -595,11 +595,11 @@ static int secp256k1_ecmult_strauss_batch(const secp256k1_callback* error_callba
     scalars = (secp256k1_scalar*)secp256k1_scratch_alloc(error_callback, scratch, n_points * sizeof(secp256k1_scalar));
     state.prej = (secp256k1_gej*)secp256k1_scratch_alloc(error_callback, scratch, n_points * ECMULT_TABLE_SIZE(WINDOW_A) * sizeof(secp256k1_gej));
     state.zr = (secp256k1_fe*)secp256k1_scratch_alloc(error_callback, scratch, n_points * ECMULT_TABLE_SIZE(WINDOW_A) * sizeof(secp256k1_fe));
-    state.pre_a = (secp256k1_ge*)secp256k1_scratch_alloc(error_callback, scratch, n_points * 2 * ECMULT_TABLE_SIZE(WINDOW_A) * sizeof(secp256k1_ge));
-    state.pre_a_lam = state.pre_a + n_points * ECMULT_TABLE_SIZE(WINDOW_A);
+    state.pre_a = (secp256k1_ge*)secp256k1_scratch_alloc(error_callback, scratch, n_points * ECMULT_TABLE_SIZE(WINDOW_A) * sizeof(secp256k1_ge));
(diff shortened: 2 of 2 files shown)


real-or-random:
Thanks for the RP!

Before I have to time a real reply, let me note for anyone reading here that this is not something to worry about because scratch spaces are used nowhere in the public API currently. This is dead code used for future extensions.

src/ecmult_impl.h
```diff
@@ -596,14 +596,15 @@ static int secp256k1_ecmult_strauss_batch(const secp256k1_callback* error_callba
     state.prej = (secp256k1_gej*)secp256k1_scratch_alloc(error_callback, scratch, n_points * ECMULT_TABLE_SIZE(WINDOW_A) * sizeof(secp256k1_gej));
     state.zr = (secp256k1_fe*)secp256k1_scratch_alloc(error_callback, scratch, n_points * ECMULT_TABLE_SIZE(WINDOW_A) * sizeof(secp256k1_fe));
     state.pre_a = (secp256k1_ge*)secp256k1_scratch_alloc(error_callback, scratch, n_points * 2 * ECMULT_TABLE_SIZE(WINDOW_A) * sizeof(secp256k1_ge));
-    state.pre_a_lam = state.pre_a + n_points * ECMULT_TABLE_SIZE(WINDOW_A);
     state.ps = (struct secp256k1_strauss_point_state*)secp256k1_scratch_alloc(error_callback, scratch, n_points * sizeof(struct secp256k1_strauss_point_state));
 
-    if (points == NULL || scalars == NULL || state.prej == NULL || state.zr == NULL || state.pre_a == NULL) {
+    if (points == NULL || scalars == NULL || state.prej == NULL || state.zr == NULL || state.pre_a == NULL || state.ps == NULL) {
```
real-or-random, line 601: Crazy that we missed this in #600.  

nit:
Maybe it will be better to call `secp256k1_scratch_alloc` also for `state.pre_a_lam`. This will be more consistent. I think the current code was good when we still had the endo ifdefs but it's no longer natural now that we removed these. But I don't have strong opinions on this. The version after the PR here is correct but invites doing NULL pointer arithmetic..

.travis.yml
```diff
@@ -31,6 +31,7 @@ env:
     - BUILD=distcheck WITH_VALGRIND=no CTIMETEST=no BENCH=no
     - CPPFLAGS=-DDETERMINISTIC
     - CFLAGS=-O0 CTIMETEST=no
+    - CFLAGS="-fsanitize=undefined -fno-omit-frame-pointer" LDFLAGS="-fsanitize=undefined -fno-omit-frame-pointer" UBSAN_OPTIONS="print_stacktrace=1:halt_on_error=1" CTIMETEST=no
```
real-or-random, line 34: I think this should go to a different PR (if we want this at all). Not sure how long this takes on Travis, and it may not run on the GCC version on Travis etc.

real-or-random (commented):
Concept ACK

.travis.yml
Fabcien: Here is the result on Travis: https://travis-ci.org/github/Fabcien/secp256k1-1/builds/738981454
UBSAN has a low overhead so the duration is only slightly impacted.

Fabcien:
@real-or-random Now calling `secp256k1_scratch_alloc` for `state.pre_a_lam`.
Let me know if you still want me to move the Travis change to its own PR. I personally prefer to keep the changes and the tests together.

real-or-random:
I'm ok with adding it here (can't speak for the others) but I think then it should look more like this line then (enable all the stuff)
`BIGNUM=no ASM=x86_64 ECDH=yes  RECOVERY=yes EXPERIMENTAL=yes SCHNORRSIG=yes`

And should we add `print_summary=1`? By the way, I can't find official docs on `UBSAN_OPTIONS`. Do you have a good ressource?

Fabcien:
The best resource I could find is the [documentation from LLVM](http://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html). The [GCC documentation](https://gcc.gnu.org/onlinedocs/gcc/Instrumentation-Options.html) explains the `halt_on_error` behavior.
I didn't know the `print_summary` option but it seems to be useless with `halt_on_error=1`, which needs to be set for the build to fail if UBSAN finds an issue.

sipa:
Concept ACK.

I think having unit tests with `-fsanitize=undefined` is a good idea in any case. Can you do that in a separate commit though? It's an unrelated change.

Fabcien:
@sipa Sure, done.

sipa:
ACK 29a299e373d5f0e326be74c514c7c70ddf50cce1. Reviewed the code changes and verified that building with these sanitizer flags catches the existing error, as well as a signed integer overflow if introduced.

sipa:
There is an issue with Travis' s390x build, which I've cancelled. CI succeeded for all other builds.

real-or-random:
ACK 29a299e373d5f0e326be74c514c7c70ddf50cce1 code inspection

real-or-random (approved):

jonasnick (approved):
utACK 29a299e373d5f0e326be74c514c7c70ddf50cce1

PR #840, opened 2020-10-26
Return NULL early in context_preallocated_create if flags invalid

real-or-random:
If the user passes invalid flags to _context_create, and the default
illegal callback does not abort the program (which is possible), then we
work with the result of malloc(0), which may be undefined behavior. This
violates the promise that a library function won't crash after the
illegal callback has been called.

This commit fixes this issue by returning NULL early in _context_create
in that case.


src/secp256k1.c
```diff
@@ -122,21 +124,21 @@ secp256k1_context* secp256k1_context_preallocated_create(void* prealloc, unsigne
...
-            secp256k1_callback_call(&ret->illegal_callback,
-                                    "Invalid flags");
-            return NULL;
-    }
-
     secp256k1_ecmult_context_init(&ret->ecmult_ctx);
     secp256k1_ecmult_gen_context_init(&ret->ecmult_gen_ctx);
 
+    /* Flags have been checked by secp256k1_context_preallocated_size. */
+    VERIFY_CHECK((flags & SECP256K1_FLAGS_TYPE_MASK) != SECP256K1_FLAGS_TYPE_CONTEXT);
```
sipa, line 141: I think you mean `==` here.

sipa (commented):
Concept ACK, but the newly introduced `VERIFY_CHECK` is wrong, causing tests to fail.

jonasnick (commented):
Looks good to me. Local tests pass with the `!=` -> `==` fix.

real-or-random:
Fixed...

sipa:
ACK ebfa2058e9cc2999dada47d2f1e1e5c0f4bcf619

jonasnick (commented):
ACK ebfa2058e9cc2999dada47d2f1e1e5c0f4bcf619

gmaxwell:
Should the header at secp256k1_preallocated.h indicate that this can return a null pointer on error?  (I guess it could before too, but it appears to be undocumented.)

real-or-random:
> Should the header at secp256k1_preallocated.h indicate that this can return a null pointer on error? (I guess it could before too, but it appears to be undocumented.)

The same is true for the the normal `_context_create`, which calls `_context_preallocated_create` (this is the case when we get `malloc(0)`. This issue has also been brought up in #783. What is true is that we say the return values of the function are undefined if the illegal callback fires. And this is enough IMO. 

But the docs are really difficult to understand here. We give a promise that the illegal callback "will only trigger for violations that are mentioned explicitly in the header." and this is not really mentioned (or at least) it's not very implicit. We may want to add a sentence for each function can calls the illegal callback but I don't know if we want to commit on this... I suggest to discuss this in #783.

gmaxwell:
K. Just thought I'd mention it.

PR #844, opened 2020-11-02
schnorrsig API overhaul

jonasnick:
This is a work in progress because I wanted to put this up for discussion before writing tests. It addresses the TODOs that didn't make it in the schnorrsig PR and changes the APIs of `schnorrsig_sign`, `schnorrsig_verify` and `hardened_nonce_function`.

- Ideally, the new `aux_rand32` argument for `sign` would be const, but didn't find a solution I was happy with.
- Support for variable length message signing and verification supports the [suggested BIP amendment](https://github.com/sipa/bips/issues/207#issuecomment-673681901) for such messages.
- ~~`sign_custom` with its opaque config object allows adding more arguments later without having to change the API again. Perhaps there are other sensible customization options, but I'm thinking of [sign-to-contract/covert-channel](https://github.com/bitcoin-core/secp256k1/pull/590) in particular. It would require adding the fields `unsigned char *s2c_data32` and `secp256k1_s2c_opening *s2c_opening` to the config struct. The former is the data to commit to and the latter is written to by `sign_custom`.~~ (EDIT: see below)

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -793,6 +793,31 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_combine(
     size_t n
 ) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
 
+/** Compute a tagged hash as defined in BIP-340.
+ *
+ *  This is useful for creating a message hash and achieving domain separation
+ *  through an application-specific tag. This function returns
+ *  SHA256(SHA256(tag)||SHA256(tag)||msg). Therefore, tagged hash
+ *  implementations optimized for a specific tag can precompute the SHA256 state
+ *  after hashing the tag hashes.
+ *
+ *  Returns 0 if the arguments are invalid and 1 otherwise.
+ *  Args:    ctx: pointer to a context object
+ *  Out:  hash32: pointer to a 32-byte array to store the resulting hash
+ *  In:      tag: pointer to an array containing the tag
+ *        taglen: length of the tag array
+ *           msg: pointer to an array containing the message
+ *        msglen: length of the message array
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_tagged_sha256(
+    const secp256k1_context* ctx,
+    unsigned char *hash32,
+    const unsigned char *tag,
+    size_t taglen,
+    const unsigned char *msg,
+    size_t msglen
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(5);
+
 #ifdef __cplusplus
 }
 #endif
--- a/include/secp256k1_schnorrsig.h
+++ b/include/secp256k1_schnorrsig.h
@@ -63,6 +63,35 @@ typedef int (*secp256k1_nonce_function_hardened)(
  */
 SECP256K1_API extern const secp256k1_nonce_function_hardened secp256k1_nonce_function_bip340;
 
+/** Data structure that contains additional arguments for schnorrsig_sign_custom.
+ *
+ *  A schnorrsig_extraparams structure object can be initialized correctly by
+ *  setting it to SECP256K1_SCHNORRSIG_EXTRAPARAMS_INIT.
+ *
+ *  Members:
+ *      magic: set to SECP256K1_SCHNORRSIG_EXTRAPARAMS_MAGIC at initialization
+ *             and has no other function than making sure the object is
+ *             initialized.
+ *    noncefp: pointer to a nonce generation function. If NULL,
+ *             secp256k1_nonce_function_bip340 is used
+ *      ndata: pointer to arbitrary data used by the nonce generation function
+ *             (can be NULL). If it is non-NULL and
+ *             secp256k1_nonce_function_bip340 is used, then ndata must be a
+ *             pointer to 32-byte auxiliary randomness as per BIP-340.
+ */
+typedef struct {
+    unsigned char magic[4];
+    secp256k1_nonce_function_hardened noncefp;
+    void* ndata;
+} secp256k1_schnorrsig_extraparams;
+
+#define SECP256K1_SCHNORRSIG_EXTRAPARAMS_MAGIC "\xda\x6f\xb3\x8c"
+#define SECP256K1_SCHNORRSIG_EXTRAPARAMS_INIT {\
+    SECP256K1_SCHNORRSIG_EXTRAPARAMS_MAGIC,\
+    NULL,\
+    NULL\
+}
+
 /** Create a Schnorr signature.
  *
  *  Does _not_ strictly follow BIP-340 because it does not verify the resulting
@@ -97,17 +126,15 @@ SECP256K1_API int secp256k1_schnorrsig_sign(
 /** Create a Schnorr signature with a more flexible API.
  *
  *  Same arguments as secp256k1_schnorrsig_sign except that it allows signing
- *  variable length messages and allows providing a different nonce derivation
- *  function with its own data argument.
+ *  variable length messages and accepts a pointer to an extraparams object that
+ *  allows customizing signing by passing additional arguments.
+ *
+ *  Creates the same signatures as schnorrsig_sign if msglen is 32 and the
+ *  extraparams.ndata is the same as aux_rand32.
  *
  *  In:     msg: the message being signed. Can only be NULL if msglen is 0.
  *       msglen: length of the message
- *      noncefp: pointer to a nonce generation function. If NULL,
- *               secp256k1_nonce_function_bip340 is used.
- *        ndata: pointer to arbitrary data used by the nonce generation function
- *               (can be NULL). If it is non-NULL and
- *               secp256k1_nonce_function_bip340 is used, then ndata must be a
- *               pointer to 32-byte auxiliary randomness as per BIP-340.
+ *  extraparams: pointer to a extraparams object (can be NULL)
  */
 SECP256K1_API int secp256k1_schnorrsig_sign_custom(
     const secp256k1_context* ctx,
@@ -115,8 +142,7 @@ SECP256K1_API int secp256k1_schnorrsig_sign_custom(
     const unsigned char *msg,
     size_t msglen,
     const secp256k1_keypair *keypair,
-    secp256k1_nonce_function_hardened noncefp,
-    void *ndata
+    secp256k1_schnorrsig_extraparams *extraparams
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(5);
 
 /** Verify a Schnorr signature.
--- a/include/secp256k1_schnorrsig.h
+++ b/include/secp256k1_schnorrsig.h
@@ -66,22 +66,38 @@ SECP256K1_API extern const secp256k1_nonce_function_hardened secp256k1_nonce_fun
  *  signature. Instead, you can manually use secp256k1_schnorrsig_verify and
  *  abort if it fails.
  *
- *  Otherwise BIP-340 compliant if the noncefp argument is NULL or
- *  secp256k1_nonce_function_bip340 and the ndata argument is 32-byte auxiliary
- *  randomness.
- *
  *  Returns 1 on success, 0 on failure.
  *  Args:    ctx: pointer to a context object, initialized for signing (cannot be NULL)
  *  Out:   sig64: pointer to a 64-byte array to store the serialized signature (cannot be NULL)
  *  In:    msg32: the 32-byte message being signed (cannot be NULL)
  *       keypair: pointer to an initialized keypair (cannot be NULL)
- *       noncefp: pointer to a nonce generation function. If NULL, secp256k1_nonce_function_bip340 is used
- *         ndata: pointer to arbitrary data used by the nonce generation
- *                function (can be NULL). If it is non-NULL and
- *                secp256k1_nonce_function_bip340 is used, then ndata must be a
- *                pointer to 32-byte auxiliary randomness as per BIP-340.
+ *    aux_rand32: 32 bytes of fresh randomness. While recommended to provide
+ *                this, it is only supplemental to security and can be NULL. See
+ *                BIP-340 "Default Signing" for a full explanation of this
+ *                argument and for guidance if randomness is expensive.
  */
 SECP256K1_API int secp256k1_schnorrsig_sign(
+    const secp256k1_context* ctx,
+    unsigned char *sig64,
+    const unsigned char *msg32,
+    const secp256k1_keypair *keypair,
+    unsigned char *aux_rand32
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
+
+/** Create a Schnorr signature with a more flexible API.
+ *
+ *  Same arguments as secp256k1_schnorrsig_sign except that it misses aux_rand32
+ *  and instead allows allows providing a different nonce derivation function
+ *  with its own data argument.
+ *
+ *  In: noncefp: pointer to a nonce generation function. If NULL,
+ *               secp256k1_nonce_function_bip340 is used
+ *        ndata: pointer to arbitrary data used by the nonce generation function
+ *               (can be NULL). If it is non-NULL and
+ *               secp256k1_nonce_function_bip340 is used, then ndata must be a
+ *               pointer to 32-byte auxiliary randomness as per BIP-340.
+ */
+SECP256K1_API int secp256k1_schnorrsig_sign_custom(
     const secp256k1_context* ctx,
     unsigned char *sig64,
     const unsigned char *msg32,
--- a/include/secp256k1_schnorrsig.h
+++ b/include/secp256k1_schnorrsig.h
@@ -24,7 +24,9 @@ extern "C" {
  *  Returns: 1 if a nonce was successfully generated. 0 will cause signing to
  *           return an error.
  *  Out:  nonce32: pointer to a 32-byte array to be filled by the function
- *  In:     msg32: the 32-byte message hash being verified (will not be NULL)
+ *  In:       msg: the message being verified. Is NULL if and only if msglen
+ *                 is 0.
+ *         msglen: the length of the message
  *          key32: pointer to a 32-byte secret key (will not be NULL)
  *     xonly_pk32: the 32-byte serialized xonly pubkey corresponding to key32
  *                 (will not be NULL)
@@ -38,7 +40,8 @@ extern "C" {
  */
 typedef int (*secp256k1_nonce_function_hardened)(
     unsigned char *nonce32,
-    const unsigned char *msg32,
+    const unsigned char *msg,
+    size_t msglen,
     const unsigned char *key32,
     const unsigned char *xonly_pk32,
     const unsigned char *algo,
@@ -66,6 +69,13 @@ SECP256K1_API extern const secp256k1_nonce_function_hardened secp256k1_nonce_fun
  *  signature. Instead, you can manually use secp256k1_schnorrsig_verify and
  *  abort if it fails.
  *
+ *  This function only signs 32-byte messages. If you have messages of a
+ *  different size (or the same size but without a context-specific tag
+ *  prefix), it is recommended to create a 32-byte message hash with
+ *  secp256k1_tagged_sha256 and then sign the hash. Tagged hashing allows
+ *  providing an context-specific tag for domain separation. This prevents
+ *  signatures from being valid in multiple contexts by accident.
+ *
  *  Returns 1 on success, 0 on failure.
  *  Args:    ctx: pointer to a context object, initialized for signing (cannot be NULL)
  *  Out:   sig64: pointer to a 64-byte array to store the serialized signature (cannot be NULL)
@@ -86,12 +96,14 @@ SECP256K1_API int secp256k1_schnorrsig_sign(
 
 /** Create a Schnorr signature with a more flexible API.
  *
- *  Same arguments as secp256k1_schnorrsig_sign except that it misses aux_rand32
- *  and instead allows allows providing a different nonce derivation function
- *  with its own data argument.
+ *  Same arguments as secp256k1_schnorrsig_sign except that it allows signing
+ *  variable length messages and allows providing a different nonce derivation
+ *  function with its own data argument.
  *
- *  In: noncefp: pointer to a nonce generation function. If NULL,
- *               secp256k1_nonce_function_bip340 is used
+ *  In:     msg: the message being signed. Can only be NULL if msglen is 0.
+ *       msglen: length of the message
+ *      noncefp: pointer to a nonce generation function. If NULL,
+ *               secp256k1_nonce_function_bip340 is used.
  *        ndata: pointer to arbitrary data used by the nonce generation function
  *               (can be NULL). If it is non-NULL and
  *               secp256k1_nonce_function_bip340 is used, then ndata must be a
@@ -100,11 +112,12 @@ SECP256K1_API int secp256k1_schnorrsig_sign(
 SECP256K1_API int secp256k1_schnorrsig_sign_custom(
     const secp256k1_context* ctx,
     unsigned char *sig64,
-    const unsigned char *msg32,
+    const unsigned char *msg,
+    size_t msglen,
     const secp256k1_keypair *keypair,
     secp256k1_nonce_function_hardened noncefp,
     void *ndata
-) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(5);
 
 /** Verify a Schnorr signature.
  *
@@ -112,15 +125,17 @@ SECP256K1_API int secp256k1_schnorrsig_sign_custom(
  *           0: incorrect signature
  *  Args:    ctx: a secp256k1 context object, initialized for verification.
  *  In:    sig64: pointer to the 64-byte signature to verify (cannot be NULL)
- *         msg32: the 32-byte message being verified (cannot be NULL)
+ *           msg: the message being verified. Can only be NULL if msglen is 0.
+ *        msglen: length of the message
  *        pubkey: pointer to an x-only public key to verify with (cannot be NULL)
  */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_schnorrsig_verify(
     const secp256k1_context* ctx,
     const unsigned char *sig64,
-    const unsigned char *msg32,
+    const unsigned char *msg,
+    size_t msglen,
     const secp256k1_xonly_pubkey *pubkey
-) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(5);
 
 #ifdef __cplusplus
 }
--- a/include/secp256k1_schnorrsig.h
+++ b/include/secp256k1_schnorrsig.h
@@ -23,14 +23,15 @@ extern "C" {
  *
  *  Returns: 1 if a nonce was successfully generated. 0 will cause signing to
  *           return an error.
- *  Out:     nonce32:   pointer to a 32-byte array to be filled by the function.
- *  In:      msg32:     the 32-byte message hash being verified (will not be NULL)
- *           key32:     pointer to a 32-byte secret key (will not be NULL)
- *      xonly_pk32:     the 32-byte serialized xonly pubkey corresponding to key32
- *                      (will not be NULL)
- *           algo16:    pointer to a 16-byte array describing the signature
- *                      algorithm (will not be NULL).
- *           data:      Arbitrary data pointer that is passed through.
+ *  Out:  nonce32: pointer to a 32-byte array to be filled by the function
+ *  In:     msg32: the 32-byte message hash being verified (will not be NULL)
+ *          key32: pointer to a 32-byte secret key (will not be NULL)
+ *     xonly_pk32: the 32-byte serialized xonly pubkey corresponding to key32
+ *                 (will not be NULL)
+ *           algo: pointer to an array describing the signature
+ *                 algorithm (will not be NULL)
+ *        algolen: the length of the algo array
+ *           data: arbitrary data pointer that is passed through
  *
  *  Except for test cases, this function should compute some cryptographic hash of
  *  the message, the key, the pubkey, the algorithm description, and data.
@@ -40,7 +41,8 @@ typedef int (*secp256k1_nonce_function_hardened)(
     const unsigned char *msg32,
     const unsigned char *key32,
     const unsigned char *xonly_pk32,
-    const unsigned char *algo16,
+    const unsigned char *algo,
+    size_t algolen,
     void *data
 );
 
@@ -51,10 +53,10 @@ typedef int (*secp256k1_nonce_function_hardened)(
  *  If a data pointer is passed, it is assumed to be a pointer to 32 bytes of
  *  auxiliary random data as defined in BIP-340. If the data pointer is NULL,
  *  the nonce derivation procedure follows BIP-340 by setting the auxiliary
- *  random data to zero. The algo16 argument must be non-NULL, otherwise the
- *  function will fail and return 0. The hash will be tagged with algo16 after
- *  removing all terminating null bytes. Therefore, to create BIP-340 compliant
- *  signatures, algo16 must be set to "BIP0340/nonce\0\0\0"
+ *  random data to zero. The algo argument must be non-NULL, otherwise the
+ *  function will fail and return 0. The hash will be tagged with algo.
+ *  Therefore, to create BIP-340 compliant signatures, algo must be set to
+ *  "BIP0340/nonce" and algolen to 13.
  */
 SECP256K1_API extern const secp256k1_nonce_function_hardened secp256k1_nonce_function_bip340;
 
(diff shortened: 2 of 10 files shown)


jonasnick:
ping :)

include/secp256k1_schnorrsig.h
```diff
@@ -50,59 +52,125 @@ typedef int (*secp256k1_nonce_function_hardened)(
...
+ *  arguments.
+ *
+ *  In: noncefp: pointer to a nonce generation function. If NULL,
+ *               secp256k1_nonce_function_bip340 is used
+ *        ndata: pointer to arbitrary data used by the nonce generation function
+ *               (can be NULL). If it is non-NULL and
+ *               secp256k1_nonce_function_bip340 is used, then ndata must be a
+ *               pointer to 32-byte auxiliary randomness as per BIP-340.
+ *       config: pointer to a config object.
+ */
```
LLFourn, line 146: Perhaps I am misunderstanding the doc comment syntax but something is not quite right here. It should say something like "instead of `aux_rand32` it accepts a custom config...". I think then the only "In" should be the config?

jonasnick: You're right, the doc for `ndata` and `noncefp` here should be removed in the last commit.

sipa:
Concept ACK

elichai:
Can we use an opaque "array" struct for the config instead of using malloc? this will still allow us to extend the struct while also supporting platforms without heap allocation

jonasnick:
@elichai Yeah good point, I agree that we should avoid dynamic memory allocation... and was hoping for suggestions :)

The opaque array struct wouldn't be fully backwards compatible because it only works if the caller is linked to the correct version of libsecp. In my limited understanding, I can count three reasonable options if we want to avoid malloc and be binary backwards compatible:

1. Use a versioned struct [as in the Linux kernel](https://lkml.org/lkml/2015/7/30/117). I implemented [this variant](https://stackoverflow.com/questions/25947997/how-does-sizeofstruct-help-provide-abi-compatibility#comment40623725_25947997) in a [branch](https://github.com/jonasnick/secp256k1/commits/schnorrsig-api-static) but I don't know enough of the implications, i.e. how well this works with [ffi](https://github.com/ElementsProject/secp256k1-zkp/pull/37).
2. Make space for future additions in the object by making the opaque byte array extra large or adding dummy members if we'd use a public struct instead.
3. Allocate it on the scratch space and work on optionally providing stack space to the scratch space (instead of having it malloc). That's not exactly backwards compatible either because the provided scratch space may become to small.

In case 1 and 2 we should document that the struct size may change between versions such that no one relies on this (otherwise it seems like this would become super [complex](https://sourceware.org/glibc/wiki/Development/Versioning_A_Structure))

include/secp256k1_schnorrsig.h
```diff
@@ -50,59 +52,125 @@ typedef int (*secp256k1_nonce_function_hardened)(
...
- *  schnorrsig_sign does not produce BIP-340 compliant signatures. The algo16
- *  argument must be non-NULL, otherwise the function will fail and return 0.
- *  The hash will be tagged with algo16 after removing all terminating null
- *  bytes. Therefore, to create BIP-340 compliant signatures, algo16 must be set
- *  to "BIP0340/nonce\0\0\0"
+ *  schnorrsig_sign does not follow the BIP-340 nonce derivation procedure
+ *  exactly. The algo16 argument must be non-NULL, otherwise the function will
+ *  fail and return 0. The hash will be tagged with algo16 after removing all
+ *  terminating null bytes. Therefore, to create BIP-340 compliant signatures,
+ *  algo16 must be set to "BIP0340/nonce\0\0\0"
```
real-or-random, line 59: Maybe it's easier than having a length but I think if we have a length arg for the message, we can have it here too. `\0\0\0` is somewhat strange. I mean it works but stripping trailing \0 bytes is very unconventional, I haven't seen this anywhere else.

real-or-random (commented):
Am I right that the user can still create config objects simply on the stack? This should be the case, we shouldn't force the user to have malloc.

But if the answer to this question is yes, then I think we could get rid of the create/destroy functions entirely. Config objects will be small, it's fine and natural to use the stack. And if the user really wants dynamic memory, she can call still call malloc.  


Or do we think the size of the config object isn't statically given?  This is true for example for the context object. Here, the size depends on the flags. (Although you could argue that there are only four types of contexts with known static sizes...) But here I don't think this will be case for config objects.

real-or-random:
Ah sorry, I haven't seen your previous comment @jonasnick.

I think what I have in mind is similar to 1. but even simpler.

As long as we keep this an opaque object with manipulation functions provided by us, and we give the caller a way to figure out it's size (e.g. for FFI/dynamic alloc), versioning should not be necessary. Am I right?

apoelstra:
@real-or-random the problem is that the "manipulation functions provided by us" need to use malloc.

jonasnick:
I implemented & pushed option 1, the versioned struct, after playing with it [here](https://github.com/jonasnick/versioned-param). The only downside is that if we add fields, we have to make sure that the new struct is in fact larger than the old one - on all supported compilers and platforms. We already make an assumption on the biggest alignment, so that should help figuring out when to add dummy fields.

I also mentioned above that there would be issues with FFI, but there are actually none. The concern in the issue I linked to was essentially that a struct could change out from under the bindings, but that's exactly what versioned structs prevent.

I'll add the missing tests and documentation.

> As long as we keep this an opaque object

@real-or-random callers can't put an opaque object on the stack because they can't statically determine its size.

jonasnick:
I pushed a new version of this PR with two major change:

1. The versioned struct for extra sign_custom parameters is removed. Versioned structs are useful for forward compatibility (i.e. application expecting newer library actually calls older library), but it's not clear what to do in case we add sign to contract members to the struct, but the library doesn't understand S2C yet. We can abort, but that's not really forward compatibility. Instead it's a simple struct now with a magic value which can be used to achieve backward compatibility. If we decide to change the struct, we also change the magic which then acts as a version number and allows libsecp to distinguish between different structs.
2. The `schnorrsig_sign` function only accepts 32-byte messages again. Also, I added the public `secp256k1_tagged_sha256` function to create tagged hashes according to BIP-340. The `schnorrsig_sign` doc explains that for messages that are not 32 bytes, one should use `secp256k1_tagged_sha256` with a domain seperator. `schnorrsig_sign_custom` still allows truly signing variable length messages. As such, with libsecp both methods for variable length messages can be used, but one is clearly recommended to not confuse users and that's the one that currently has the most adoption (Taproot, DLCs, Lightning offers). And the additional tag argument to `secp256k1_tagged_sha256` makes it harder to miss domain separation.

I also added tests and docs. Since this PR doesn't contradict the discussed BIP-340 changes, I removed the WIP status.

include/secp256k1_schnorrsig.h
jonasnick: fixed

include/secp256k1_schnorrsig.h
jonasnick: added `algolen` argument

include/secp256k1_schnorrsig.h
```diff
@@ -23,24 +23,28 @@ extern "C" {
...
- *  In:      msg32:     the 32-byte message hash being verified (will not be NULL)
- *           key32:     pointer to a 32-byte secret key (will not be NULL)
- *      xonly_pk32:     the 32-byte serialized xonly pubkey corresponding to key32
- *                      (will not be NULL)
- *           algo16:    pointer to a 16-byte array describing the signature
- *                      algorithm (will not be NULL).
- *           data:      Arbitrary data pointer that is passed through.
+ *  Out:  nonce32: pointer to a 32-byte array to be filled by the function.
+ *  In:       msg: the message being verified (will not be NULL)
+ *         msglen: the length of the message (will not be NULL)
```
jesseposner, line 28:
```suggestion
 *  In:       msg: the message hash being verified (will not be NULL)
 *         msglen: the length of the message hash (will not be NULL)
```

real-or-random: No, messages in BIP340 can be arbitrary bytestrings. They do not need to be obtained via hashing.

include/secp256k1_schnorrsig.h
```diff
@@ -23,24 +23,28 @@ extern "C" {
...
  *           return an error.
- *  Out:     nonce32:   pointer to a 32-byte array to be filled by the function.
- *  In:      msg32:     the 32-byte message hash being verified (will not be NULL)
- *           key32:     pointer to a 32-byte secret key (will not be NULL)
- *      xonly_pk32:     the 32-byte serialized xonly pubkey corresponding to key32
- *                      (will not be NULL)
- *           algo16:    pointer to a 16-byte array describing the signature
- *                      algorithm (will not be NULL).
- *           data:      Arbitrary data pointer that is passed through.
+ *  Out:  nonce32: pointer to a 32-byte array to be filled by the function.
```
jesseposner, line 26:
```suggestion
 *  Out:  nonce32: pointer to a 32-byte array to be filled by the function
```

include/secp256k1_schnorrsig.h
```diff
@@ -23,24 +23,28 @@ extern "C" {
...
+ *  In:       msg: the message being verified (will not be NULL)
+ *         msglen: the length of the message (will not be NULL)
+ *          key32: pointer to a 32-byte secret key (will not be NULL)
+ *     xonly_pk32: the 32-byte serialized xonly pubkey corresponding to key32
+ *                 (will not be NULL)
+ *           algo: pointer to an array describing the signature
+ *                 algorithm (will not be NULL).
```
jesseposner, line 33:
```suggestion
 *                 algorithm (will not be NULL)
```

include/secp256k1_schnorrsig.h
```diff
@@ -23,24 +23,28 @@ extern "C" {
...
+ *        algolen: the length of the algo array
+ *           data: Arbitrary data pointer that is passed through.
```
jesseposner, line 35:
```suggestion
 *           data: arbitrary data pointer that is passed through
```

include/secp256k1_schnorrsig.h
```diff
@@ -50,59 +54,112 @@ typedef int (*secp256k1_nonce_function_hardened)(
...
  */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_schnorrsig_verify(
     const secp256k1_context* ctx,
     const unsigned char *sig64,
-    const unsigned char *msg32,
+    const unsigned char *msg,
+    size_t msglen,
     const secp256k1_xonly_pubkey *pubkey
-) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(5);
```
jesseposner, line 162: Seems like `msglen` should always be non-NULL: 
```suggestion
) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
```

include/secp256k1_schnorrsig.h
```diff
@@ -50,59 +54,112 @@ typedef int (*secp256k1_nonce_function_hardened)(
...
+ *  different size, it is recommended to create a 32-byte message hash with
+ *  secp256k1_tagged_sha256 and then sign the hash. Tagged hashing allows
+ *  providing an application-specific tag for domain separation. This prevents
+ *  signatures from being valid in multiple applications by accident.
  *
  *  Returns 1 on success, 0 on failure.
  *  Args:    ctx: pointer to a context object, initialized for signing (cannot be NULL)
  *  Out:   sig64: pointer to a 64-byte array to store the serialized signature (cannot be NULL)
- *  In:    msg32: the 32-byte message being signed (cannot be NULL)
+ *  In:    msg32: the 32-byte message hash being signed (cannot be NULL)
```
jesseposner, line 109:
```suggestion
 *  In:    msg32: the 32-byte message being signed (cannot be NULL)
```

include/secp256k1_schnorrsig.h
jesseposner: Ah, right, hence this PR. Missed that somehow on my first pass.

real-or-random: Ah no, I think there's still a misunderstanding. So with the scheme in the current BIP340, you can already sign arbitrary *32-byte* messages (no matter if they're created via a hash function or not). 

There have been changes proposed to BIP340 (not yet in the BIP) that extend this to arbitrary-length messages. This is what is implemented by this PR. 

But BIP340 never required the messages to be created via a hash function. This is just a difference to the ECDSA API.

jesseposner: Thanks for clarifying. I'm assuming that despite the fact that BIP340, unlike ECDSA, doesn't require a hashed message, when the documentation on line 27 stated previously `the 32-byte message hash being verified` that was not an error, because for typical use cases (i.e. the message is not exactly 32-bytes) the message would need to be hashed.

However, with the introduction of variable length messages in this PR, the documentation no longer makes this assumption about the typical use case because it no longer requires a 32-byte message for the APIs that now accept a `msglen`. This seems to also be reflected in the insertion of `hash` in this PR on line 109, because that API still requires a 32-byte message. So the documentation logic of this PR seems to be that if `msglen` is specified, the documentation should describe it as simply a `message` whereas when a 32-byte message is required then the documentation should describe it as a `32-byte message hash` (even though technically a hash is not required, but for typical use cases it should be a hash because messages won't typically be strictly 32-bytes).

include/secp256k1_schnorrsig.h
```diff
@@ -50,59 +54,112 @@ typedef int (*secp256k1_nonce_function_hardened)(
...
  *  abort if it fails.
  *
- *  Otherwise BIP-340 compliant if the noncefp argument is NULL or
- *  secp256k1_nonce_function_bip340 and the ndata argument is 32-byte auxiliary
- *  randomness.
+ *  This function only signs 32-byte messages. If you have messages of a
+ *  different size, it is recommended to create a 32-byte message hash with
+ *  secp256k1_tagged_sha256 and then sign the hash. Tagged hashing allows
+ *  providing an application-specific tag for domain separation. This prevents
+ *  signatures from being valid in multiple applications by accident.
```
jesseposner, line 104: This paragraph might make it sound like application-specific tags are only relevant for non-32-byte messages, however, tags should be considered regardless of whether the message is hashed. 

Shouldn't the documentation recommend 1 of 2 options? (1) a tagged hash of the message with an application-specific tag or (2) an unhashed message with an application-specific tag prefix.

Perhaps this would be more clear: 
```suggestion
 *  This function only signs 32-byte messages. If you have messages of a
 *  different size (or the same size but without an application-specific tag prefix), it is recommended to create a 32-byte message hash with
 *  secp256k1_tagged_sha256 and then sign the hash. Tagged hashing allows
 *  providing an application-specific tag for domain separation. This prevents
 *  signatures from being valid in multiple applications by accident.
```

include/secp256k1_schnorrsig.h
jonasnick: I'm not sure I follow, what exactly is the possible documentation error? I agree that "message hash" for schnorrsig_sign's msg32 argument is misleading because it's not necessary that msg32 is one. Switched it back.

include/secp256k1_schnorrsig.h
jonasnick: done

include/secp256k1_schnorrsig.h
jonasnick: done

include/secp256k1_schnorrsig.h
jonasnick: done

include/secp256k1_schnorrsig.h
jonasnick: You're right, added your suggestion. The subtle recommendation right now from the api is the tagged hash, because that's the non `_custom` variant of schnorrsig sign.

include/secp256k1_schnorrsig.h
jesseposner: The possible documentation error I was referring to is that  currently line 27 on master says: https://github.com/bitcoin-core/secp256k1/blob/3a8b47bc6d1627a8f9f008234c5161538ef85a21/include/secp256k1_schnorrsig.h#L27 when it should probably be `the 32-byte message being verified`.

With your latest changes the documentation for both `secp256k1_nonce_function_hardened` and `secp256k1_schnorrsig_sign` now have `message` instead of `message hash` so it now looks good to me.

jonasnick:
Rebased on top of master because cirrus wasn't included in this branch.

apoelstra (approved):
ack 64503686a26e21dda422fe5377537601d0f86aa6

sipa:
API looks good, code review ACK. Feel like squashing in the fixup?

jonasnick:
Done.

src/bench_schnorrsig.c
```diff
@@ -60,7 +61,7 @@ int main(void) {
```
real-or-random, line 61:
```suggestion
    CHECK(MSGLEN >= 2);
```

src/bench_schnorrsig.c
```diff
@@ -26,7 +27,7 @@ typedef struct {
 void bench_schnorrsig_sign(void* arg, int iters) {
     bench_schnorrsig_data *data = (bench_schnorrsig_data *)arg;
     int i;
-    unsigned char msg[32] = "benchmarkexamplemessagetemplate";
+    unsigned char msg[MSGLEN] = "benchmarkexamplemessagetemplate";
```
real-or-random, line 30: maybe
```suggestion
    unsigned char msg[MSGLEN] = {0};
```
because people could set MSGLEN to < 32

include/secp256k1_schnorrsig.h
```diff
@@ -24,7 +24,8 @@ extern "C" {
  *  Returns: 1 if a nonce was successfully generated. 0 will cause signing to
  *           return an error.
  *  Out:  nonce32: pointer to a 32-byte array to be filled by the function
- *  In:     msg32: the 32-byte message hash being verified (will not be NULL)
+ *  In:       msg: the message being verified (will not be NULL)
```
real-or-random, line 27: will not be NULL unless msglen == 0

src/modules/schnorrsig/tests_impl.h
```diff
@@ -720,6 +749,14 @@ void test_schnorrsig_sign(void) {
...
     CHECK(secp256k1_schnorrsig_sign_custom(ctx, sig, msg, sizeof(msg), &keypair, &extraparams) == 1);
     CHECK(secp256k1_schnorrsig_verify(ctx, sig, msg, sizeof(msg), &pk));
+
+    /* When using the default nonce function, schnorrsig_sign_custom produces
+     * the same result as schnorrsig_sign with aux_rand = extraparams.ndata */
+    extraparams.noncefp = NULL;
+    extraparams.ndata = aux_rand;
+    CHECK(secp256k1_schnorrsig_sign_custom(ctx, sig, msg, sizeof(msg), &keypair, &extraparams) == 1);
+    CHECK(secp256k1_schnorrsig_sign(ctx, sig2, msg, &keypair, extraparams.ndata) == 1);
+    CHECK(memcmp(sig, sig2, sizeof(sig)) == 0);
```
real-or-random, line 759: secp256k1_memcmp_var

include/secp256k1_schnorrsig.h
```diff
@@ -50,59 +54,113 @@ typedef int (*secp256k1_nonce_function_hardened)(
...
  *
- *  Otherwise BIP-340 compliant if the noncefp argument is NULL or
- *  secp256k1_nonce_function_bip340 and the ndata argument is 32-byte auxiliary
- *  randomness.
+ *  This function only signs 32-byte messages. If you have messages of a
+ *  different size (or the same size but without an application-specific tag
+ *  prefix), it is recommended to create a 32-byte message hash with
+ *  secp256k1_tagged_sha256 and then sign the hash. Tagged hashing allows
+ *  providing an application-specific tag for domain separation. This prevents
+ *  signatures from being valid in multiple applications by accident.
```
real-or-random, line 105: okay maybe that's nitpicking but can you change this "in multiple applications or in multiple contexts within the same application by accident". I think people should understand that this can be meaningful within an applications (modulo the anyway imprecise meaning of this term).

src/tests.c
```diff
@@ -539,6 +539,37 @@ void run_rfc6979_hmac_sha256_tests(void) {
...
+    CHECK(secp256k1_tagged_sha256(none, hash32, NULL, 0, msg, sizeof(msg)) == 0);
+    CHECK(ecount == 2);
+    CHECK(secp256k1_tagged_sha256(none, hash32, tag, sizeof(tag), NULL, 0) == 0);
+    CHECK(ecount == 3);
+
+    /* Static test vector */
+    memcpy(tag, "tag", 3);
+    memcpy(msg, "msg", 3);
+    CHECK(secp256k1_tagged_sha256(none, hash32, tag, 3, msg, 3) == 1);
+    CHECK(memcmp(hash32, hash_expected, sizeof(hash32)) == 0);
```
real-or-random, line 570: same here. (I should really work on #833 :P)

real-or-random (commented):
I like this approach.  

Can you a explain how the extraparams struct is now forward-compatible or not? (Here on even as comments in the source?)

include/secp256k1_schnorrsig.h
```diff
@@ -50,11 +50,11 @@ typedef int (*secp256k1_nonce_function_hardened)(
  *
  *  If a data pointer is passed, it is assumed to be a pointer to 32 bytes of
  *  auxiliary random data as defined in BIP-340. If the data pointer is NULL,
- *  schnorrsig_sign does not produce BIP-340 compliant signatures. The algo16
- *  argument must be non-NULL, otherwise the function will fail and return 0.
- *  The hash will be tagged with algo16 after removing all terminating null
- *  bytes. Therefore, to create BIP-340 compliant signatures, algo16 must be set
- *  to "BIP0340/nonce\0\0\0"
+ *  schnorrsig_sign does not follow the BIP-340 nonce derivation procedure
+ *  exactly. The algo16 argument must be non-NULL, otherwise the function will
```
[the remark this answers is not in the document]
jonasnick: Yeah, I improved the sentence.

src/modules/schnorrsig/tests_impl.h
```diff
@@ -68,33 +69,38 @@ void run_nonce_function_bip340_tests(void) {
...
-    algo16[1] = 65;
-    CHECK(nonce_function_bip340(nonce, msg, key, pk, algo16, NULL) == 1);
-    /* Other algo16 is fine */
-    memset(algo16, 0xFF, 16);
-    CHECK(nonce_function_bip340(nonce, msg, key, pk, algo16, NULL) == 1);
+    /* NULL algo is disallowed */
+    CHECK(nonce_function_bip340(nonce, msg, key, pk, NULL, 0, NULL) == 0);
+    /* Empty algo is fine */
+    memset(algo, 0x00, algolen);
+    CHECK(nonce_function_bip340(nonce, msg, key, pk, algo, algolen, NULL) == 1);
```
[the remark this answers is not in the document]
jonasnick: Removed that. `algolen == 0` is tested a few lines below.

include/secp256k1_schnorrsig.h
```diff
@@ -66,22 +66,38 @@ SECP256K1_API extern const secp256k1_nonce_function_hardened secp256k1_nonce_fun
...
+    unsigned char *sig64,
+    const unsigned char *msg32,
+    const secp256k1_keypair *keypair,
+    unsigned char *aux_rand32
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
+
+/** Create a Schnorr signature with a more flexible API.
+ *
+ *  Same arguments as secp256k1_schnorrsig_sign except that it misses aux_rand32
+ *  and instead allows allows providing a different nonce derivation function
```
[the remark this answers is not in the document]
jonasnick: This is fixed in a later commit. Can I leave as is?

src/bench_schnorrsig.c
jonasnick: There was no intention to allow users to change MSGLEN and this wouldn't work since the benchmark only used `schnorrsig_sign`, not `schnorrsig_sign_custom`. It may be useful though, so I added a commit to allow that.

src/bench_schnorrsig.c
jonasnick: done

include/secp256k1_schnorrsig.h
jonasnick: good catch, fixed

src/modules/schnorrsig/tests_impl.h
jonasnick: ok

include/secp256k1_schnorrsig.h
jonasnick: Agree, but not really sure about your suggestion because we call it an "application-specific" tag. Instead, I replaced "application" with "context" which actually matches the wording in bip-340. What do you think?

src/tests.c

jonasnick:
There's _no_ forward compatibility, i.e. applications expecting a newer library than installed won't work. Adding a new field to the extraparams requires a bump of the minor version. Then, applications which need that feature can demand at least that version.

jonasnick:
Rebased to (hopefully) fix CI which was complaining about invalid bignum selection.

elichai:
I'm not sure I understand the problems this new API is solving
1. What's the point of exposing `secp256k1_tagged_sha256`? we currently don't expose regular sha256 so why should we expose a tagged one?.
2. What does the new `secp256k1_schnorrsig_extraparams` struct help solve? why is it better over the old API? (noncefp+ndata).

Thanks! (Other than that the implementation looks good)

jonasnick:
> What's the point of exposing secp256k1_tagged_sha256?

Because the recommended signing procedure is to hash the message with a context-specific tag and then sign that. If we didn't expose `secp256k1_tagged_sha256`, that would be unnecessarily difficult for users.

> What does the new secp256k1_schnorrsig_extraparams struct help solve?  I guess [...] it allows us to add new optional fields

Right, for example, we may want to use the extraparams to allow adding a sign-to-contract commitment to the signature or change the exact signing algorithm in other ways. The same could be achieved without extraparams by just adding entirely new signing functions `sign_1, ..., sign_n` with ever growing argument lists.

elichai:
> Because the recommended signing procedure is to hash the message with a context-specific tag and then sign that. If we didn't expose `secp256k1_tagged_sha256`, that would be unnecessarily difficult for users.

So the use case is something like the following?
```c
typedef struct {const secp256k1_context* ctx; unsigned char[32] hash;} custom_data;
int my_custom_nonce_func(unsigned char *nonce32, const unsigned char *msg, size_t msglen, const unsigned char *key32, const unsigned char *xonly_pk32, const unsigned char *algo, size_t algolen, void *data) {
    // Do custom own processing and result in a full message in the pair `(should_sign_ptr, should_sign_len)`.
    // ...
    custom_data* my_custom_data = (custom_data*)data;
    return secp256k1_tagged_sha256(my_custom_data->ctx, my_custom_data->hash, MY_OWN_APP_TAG, sizeof(MY_OWN_APP_TAG), should_sign_ptr, should_sign_len);
}
```

jonasnick:
The intended use case is simply signing non-32-byte messages:

```
unsigned char msg[33] = {0};
unsigned char msghash[32];
secp256k1_tagged_sha256(ctx, msghash, MY_OWN_APP_TAG, sizeof(MY_OWN_APP_TAG), msg, sizeof(msg));
secp256k1_schnorrsig_sign(ctx, ..., msghash, ...);
```

You could also use `sign_custom` with `msglen = 33` but as mentioned above "one is clearly recommended to not confuse users and that's the one that currently has the most adoption (Taproot, DLCs, Lightning offers). And the additional tag argument to secp256k1_tagged_sha256 makes it harder to miss domain separation."

include/secp256k1_schnorrsig.h
real-or-random: Indeed, that's better. I didn't want to suggest "context" because I thought it could be confused with the meaning of context as in "secp256k1 context" but that shouldn't be an issue.

include/secp256k1_schnorrsig.h
```diff
@@ -50,59 +54,113 @@ typedef int (*secp256k1_nonce_function_hardened)(
  *
  *  If a data pointer is passed, it is assumed to be a pointer to 32 bytes of
  *  auxiliary random data as defined in BIP-340. If the data pointer is NULL,
- *  schnorrsig_sign does not produce BIP-340 compliant signatures. The algo16
- *  argument must be non-NULL, otherwise the function will fail and return 0.
- *  The hash will be tagged with algo16 after removing all terminating null
- *  bytes. Therefore, to create BIP-340 compliant signatures, algo16 must be set
- *  to "BIP0340/nonce\0\0\0"
+ *  the nonce derivation procedure follows BIP-340 by setting the auxiliary
+ *  random data to zero. The algo argument must be non-NULL, otherwise the
```
real-or-random, line 58: I was about to suggest that maybe we could allow `algo == NULL` if `algolen == 0` But I'm not sure. If don't think we want to encourage users to use empty strings...

real-or-random:
Fixups are all  good, fine  to squash from my side.

jonasnick:
squashed

apoelstra:
Needs rebase

jonasnick:
rebased

jonasnick:
The new leak sanitizer in CI detected a missing `context_destroy` in the `tagged_sha256_tests`. Fixed & squashed.

apoelstra (approved):
ack 439ed2ff2fa3fe38bb0aeb26213d7f2263eca701

Glad we came to a "resolution" on the forward-compat issue.

include/secp256k1_schnorrsig.h
```diff
@@ -66,22 +66,38 @@ SECP256K1_API extern const secp256k1_nonce_function_hardened secp256k1_nonce_fun
...
  *  Out:   sig64: pointer to a 64-byte array to store the serialized signature (cannot be NULL)
  *  In:    msg32: the 32-byte message being signed (cannot be NULL)
  *       keypair: pointer to an initialized keypair (cannot be NULL)
- *       noncefp: pointer to a nonce generation function. If NULL, secp256k1_nonce_function_bip340 is used
- *         ndata: pointer to arbitrary data used by the nonce generation
- *                function (can be NULL). If it is non-NULL and
- *                secp256k1_nonce_function_bip340 is used, then ndata must be a
- *                pointer to 32-byte auxiliary randomness as per BIP-340.
+ *    aux_rand32: 32 bytes of fresh randomness. While recommended to provide
+ *                this, it is only supplemental to security and can be NULL. See
```
ariard, line 113: nits: "it is only a side-channel mitigation", "See BIP-340 "Default Signing"

include/secp256k1.h
```diff
@@ -793,6 +793,31 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_combine(
...
+ *  SHA256(SHA256(tag)||SHA256(tag)||msg). Therefore, tagged hash
+ *  implementations optimized for a specific tag can precompute the SHA256 state
+ *  after hashing the tag hashes.
+ *
+ *  Returns 0 if the arguments are invalid and 1 otherwise.
+ *  Args:    ctx: pointer to a context object
+ *  Out:  hash32: pointer to a 32-byte array to store the resulting hash
+ *  In:      tag: pointer to an array containing the tag
+ *        taglen: length of the tag array
+ *           msg: pointer to an array containing the message
```
ariard, line 809: `tag` and `msg` maybe annotate "cannot be NULL" ?

include/secp256k1_schnorrsig.h
```diff
@@ -24,7 +24,8 @@ extern "C" {
  *  Returns: 1 if a nonce was successfully generated. 0 will cause signing to
  *           return an error.
  *  Out:  nonce32: pointer to a 32-byte array to be filled by the function
- *  In:     msg32: the 32-byte message hash being verified (will not be NULL)
+ *  In:       msg: the message being verified. Can be NULL iff msglen is 0.
```
ariard, line 27: nit: s/iff/if/g

include/secp256k1_schnorrsig.h
```diff
@@ -66,6 +68,13 @@ SECP256K1_API extern const secp256k1_nonce_function_hardened secp256k1_nonce_fun
  *  signature. Instead, you can manually use secp256k1_schnorrsig_verify and
  *  abort if it fails.
  *
+ *  This function only signs 32-byte messages. If you have messages of a
+ *  different size (or the same size but without an context-specific tag
```
ariard, line 101: nit: s/an/a/g

ariard (commented):
Mostly nits, SGTM

include/secp256k1_schnorrsig.h
```diff
@@ -23,14 +23,15 @@ extern "C" {
...
- *      xonly_pk32:     the 32-byte serialized xonly pubkey corresponding to key32
- *                      (will not be NULL)
- *           algo16:    pointer to a 16-byte array describing the signature
- *                      algorithm (will not be NULL).
- *           data:      Arbitrary data pointer that is passed through.
+ *  Out:  nonce32: pointer to a 32-byte array to be filled by the function
+ *  In:     msg32: the 32-byte message hash being verified (will not be NULL)
+ *          key32: pointer to a 32-byte secret key (will not be NULL)
+ *     xonly_pk32: the 32-byte serialized xonly pubkey corresponding to key32
+ *                 (will not be NULL)
```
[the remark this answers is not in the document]
jonasnick: This is the type a function that is user provided and called from inside the library. Hence, the documentation says that this function will not be called with certain arguments set to NULL.

include/secp256k1.h
jonasnick: This was written with https://github.com/bitcoin-core/secp256k1/pull/783 already in mind which makes "cannot be NULL" implicit.

include/secp256k1_schnorrsig.h
jonasnick: Added the latter

include/secp256k1_schnorrsig.h
jonasnick: expanded to if and only if

include/secp256k1_schnorrsig.h
jonasnick: fixed

ariard:
utACK 66ff2fa6

Thanks for taking the nits.

apoelstra:
utACK the fixup commit

jonasnick:
squashed the comment fixup commit

apoelstra (approved):
ack 5f6ceafcfa46a69e901bed87e2c5f323b03b1e8c

ariard:
utACK 5f6ceaf

LLFourn (approved):
utACK 5f6ceafcfa46a69e901bed87e2c5f323b03b1e8c

lesliebranson (commented):
.

PR #845, opened 2020-11-05
Extract the secret key from a keypair

elichai:
With schnorrsig if you need to tweak the secret key (for BIP32) you must use the keypair API to get compatible secret/public keys which you do by calling `secp256k1_keypair_xonly_tweak_add()`, but after that there's no currently a way to extract the secret key back for storage.
so I added a `secp256k1_keypair_seckey` function to extract the key

### DIFF
--- a/src/modules/extrakeys/main_impl.h
+++ b/src/modules/extrakeys/main_impl.h
@@ -186,6 +186,16 @@ int secp256k1_keypair_create(const secp256k1_context* ctx, secp256k1_keypair *ke
     return ret;
 }
 
+int secp256k1_keypair_sec(const secp256k1_context* ctx, unsigned char *seckey, const secp256k1_keypair *keypair) {
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(seckey != NULL);
+    memset(seckey, 0, 32);
+    ARG_CHECK(keypair != NULL);
+
+    memcpy(seckey, &keypair->data[0], 32);
+    return 1;
+}
+
 int secp256k1_keypair_pub(const secp256k1_context* ctx, secp256k1_pubkey *pubkey, const secp256k1_keypair *keypair) {
     VERIFY_CHECK(ctx != NULL);
     ARG_CHECK(pubkey != NULL);
--- a/include/secp256k1_extrakeys.h
+++ b/include/secp256k1_extrakeys.h
@@ -165,6 +165,19 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_keypair_create(
     const unsigned char *seckey
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
 
+/** Get the secret key from a keypair.
+ *
+ *  Returns: 0 if the arguments are invalid. 1 otherwise.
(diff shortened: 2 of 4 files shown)


src/modules/extrakeys/main_impl.h
```diff
@@ -186,6 +186,16 @@ int secp256k1_keypair_create(const secp256k1_context* ctx, secp256k1_keypair *ke
     return ret;
 }
 
+int secp256k1_keypair_seckey(const secp256k1_context* ctx, unsigned char *seckey32, const secp256k1_keypair *keypair) {
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(seckey32 != NULL);
+    memset(seckey32, 0, 32);
+    ARG_CHECK(keypair != NULL);
+
+    memcpy(seckey32, &keypair->data[0], 32);
```
jonasnick, line 195: You may want to use` secp256k1_keypair_seckey_load` instead of memcpy because it correctly handles zeroed keypairs.

jonasnick (commented):
Concept ACK

Usually we test the functions more exhaustively - see for example the `keypair_pub` tests. We should have tests where the non-ctx arguments are NULL and check that the `seckey` is zeroed.

src/modules/extrakeys/main_impl.h
elichai: I change it, I've used `memcpy` because just underneath the function that extracts a pubkey just does a mempy:
https://github.com/bitcoin-core/secp256k1/blob/e89278f211a526062745c391d48a7baf782b4b2b/src/modules/extrakeys/main_impl.h#L195

elichai: And I'll add more tests

jonasnick: I see. Yeah, it should be consistent with the pubkey getter. Fine for now imo, since this outputs an invalid seckey, we don't mention a guarantee that a function fails if called with an invalid keypair and the current version is simpler than the alternative.

jonasnick (commented):
ACK 095f5eaa97dc639a8415cc52f640dafe37b4c245

real-or-random:
Concept ACK

I think this function should be added to the constant-time tests.

elichai:
> edit: And a nit on the naming: The other function is just `..._pub`, so should this be just `..._sec`? Not sure to be honest.

hmm I'm also not sure if it's better to stay "consistent" with the half name or write the "full name", I'll change this if anyone prefers a specific side

real-or-random:
@jonasnick  Was there a specific reason you choose the half name? 

(I know it's kind of annoying to have these debates but I think it's better to have them early than never.)

jonasnick:
> @jonasnick Was there a specific reason you choose the half name?

Yes, the name's in the module are generally too long already and `key` is already part of the name. `_sec` doesn't sound as nice, could also call it `_secret`. But I'd (slightly) prefer to be consistent.

sipa:
Concept ACK. Adding to constant-time tests would be good.

elichai:
@sipa @jonasnick Fixed

jonasnick (commented):
ACK 33cb3c2b1fc3f3fe46c6d0eab118248ea86c1f06

real-or-random (approved):
ACK 33cb3c2b1fc3f3fe46c6d0eab118248ea86c1f06 code inspection, tests pass

PR #846, opened 2020-11-09
ci: Run ASan/LSan and reorganize sanitizer and Valgrind jobs

real-or-random:
This enables asan (including lsan), and restructures the sanitizer builds. It also removes `-fno-omit-frame-pointer` again. This is debatable. The main reason for removing this is that GCC (but not clang) fails to build (runs out of registers) when combining asan, `-fno-omit-frame-pointer`, and the x86_64 asm.  But I believe  without `-fno-omit-frame-pointer`, the instrumented binary may be  closer to the real binary, which is good. If we get unusable debugging output on CI without frame pointers, we can still reproduce the build locally. 

 - `ECMULTGENPRECISION=8` : We should disable this entirely, not only on CI, because it's slow and needs too much stack space

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -5,7 +5,6 @@ env:
   ASM: no
   BUILD: check
   WITH_VALGRIND: yes
-  RUN_VALGRIND: no
   EXTRAFLAGS:
   HOST:
   ECDH: no
@@ -162,9 +161,8 @@ task:
     cpu: 1
     memory: 1G
   env:
-    QEMU_CMD: qemu-s390x
+    WRAPPER_CMD: qemu-s390x
     HOST: s390x-linux-gnu
-    BUILD:
     WITH_VALGRIND: no
     ECDH: yes
     RECOVERY: yes
@@ -185,9 +183,8 @@ task:
     cpu: 1
     memory: 1G
   env:
-    QEMU_CMD: qemu-arm
+    WRAPPER_CMD: qemu-arm
     HOST: arm-linux-gnueabihf
-    BUILD:
     WITH_VALGRIND: no
     ECDH: yes
     RECOVERY: yes
@@ -209,9 +206,8 @@ task:
     cpu: 1
     memory: 1G
   env:
-    QEMU_CMD: qemu-aarch64
+    WRAPPER_CMD: qemu-aarch64
     HOST: aarch64-linux-gnu
-    BUILD:
     WITH_VALGRIND: no
     ECDH: yes
     RECOVERY: yes
@@ -230,9 +226,8 @@ task:
     cpu: 1
     memory: 1G
   env:
-    WINE_CMD: wine64-stable
+    WRAPPER_CMD: wine64-stable
     HOST: x86_64-w64-mingw32
-    BUILD:
     WITH_VALGRIND: no
     ECDH: yes
     RECOVERY: yes
@@ -260,7 +255,8 @@ task:
   matrix:
     - name: "Valgrind (memcheck)"
       env:
-        RUN_VALGRIND: yes
+        # The `--error-exitcode` is required to make the test fail if valgrind found errors, otherwise it'll return 0 (https://www.valgrind.org/docs/manual/manual-core.html)
+        WRAPPER_CMD: "valgrind --error-exitcode=42"
     - name: "UBSan, ASan, LSan"
       env:
         CFLAGS: "-fsanitize=undefined,address"
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -13,7 +13,8 @@ env:
   EXPERIMENTAL: no
   CTIMETEST: yes
   BENCH: yes
-  ITERS: 2
+  TEST_ITERS:
+  BENCH_ITERS: 2
   MAKEFLAGS: -j2
 
 cat_logs_snippet: &CAT_LOGS
@@ -162,6 +163,7 @@ task:
     memory: 1G
   env:
     WRAPPER_CMD: qemu-s390x
+    TEST_ITERS: 16
     HOST: s390x-linux-gnu
     WITH_VALGRIND: no
     ECDH: yes
@@ -184,6 +186,7 @@ task:
     memory: 1G
   env:
     WRAPPER_CMD: qemu-arm
+    TEST_ITERS: 16
     HOST: arm-linux-gnueabihf
     WITH_VALGRIND: no
     ECDH: yes
@@ -207,6 +210,7 @@ task:
     memory: 1G
   env:
     WRAPPER_CMD: qemu-aarch64
+    TEST_ITERS: 16
     HOST: aarch64-linux-gnu
     WITH_VALGRIND: no
     ECDH: yes
@@ -227,6 +231,7 @@ task:
     memory: 1G
   env:
     WRAPPER_CMD: wine64-stable
+    TEST_ITERS: 16
     HOST: x86_64-w64-mingw32
     WITH_VALGRIND: no
     ECDH: yes
@@ -257,6 +262,7 @@ task:
       env:
         # The `--error-exitcode` is required to make the test fail if valgrind found errors, otherwise it'll return 0 (https://www.valgrind.org/docs/manual/manual-core.html)
         WRAPPER_CMD: "valgrind --error-exitcode=42"
+        TEST_ITERS: 16
     - name: "UBSan, ASan, LSan"
       env:
         CFLAGS: "-fsanitize=undefined,address"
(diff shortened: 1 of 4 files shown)


real-or-random:
A few things are weird on MacOS, and I don't really understand what's going on:
 * jobs 49 and 50 fail because lsan does not work with Apple's clang, see config.log. That's expected according to https://github.com/google/sanitizers/issues/1026
 * jobs 67 succeeds but 68 fails, with the same error message as above. But why the difference? I have no idea, they should use the same toolchain etc. Is this related to gmp? But 67 seem to fail with the "native compiler", and we don't set `$CC_FOR_BUILD` correctly for gcc. I've added this now.
 * I'm not convinced that the "gcc" jobs (including 67, 68) really use gcc. It's again Apple's fake gcc according to the `$CC --version` in `after_script`.  @elichai I added a `$CC --version` in the script, let's see if this makes a difference. (This will be obsolete anyway when we switch away from travis).


And job 32 times out... well that's not nice but at I understand what's going on ...

elichai:
>  * I'm not convinced that the "gcc" jobs (including 67, 68) really use gcc. It's again Apple's fake gcc according to the `$CC --version` in `after_script`.  @elichai I added a `$CC --version` in the script, let's see if this makes a difference. (This will be obsolete anyway when we switch away from travis).

IIRC that's why I did `CC=gcc-9` because OSX has an alias for `gcc=clang`
https://github.com/bitcoin-core/secp256k1/pull/750/files#diff-5c41b2883cfdc97e6b8a07ebe6456643de0160099b7f15ed04afeb2fd627f630R14

real-or-random:
> IIRC that's why I did `CC=gcc-9` because OSX has an alias for `gcc=clang`

Apparently that does not seem to work (anymore?).

I think we shouldn't try to fix this now but first move to another CI platform. For example, Cirrus seems to ship with GCC https://github.com/cirruslabs/cirrus-ci-docs/issues/67 .

real-or-random:
Rebased to work with Cirrus. I also updated the initial comment.

sipa:
utACK 1191bb07b0198ed605b30efebc7a542fbc14958f

.cirrus.yml
```diff
@@ -262,3 +243,53 @@ task:
...
+        WRAPPER_CMD: "valgrind --error-exitcode=42"
+    - name: "UBSan, ASan, LSan"
+      env:
+        # Use dummy wrapper command to run the tests outside of make check
+        WRAPPER_CMD: env
+        CFLAGS: "-fsanitize=undefined,address"
+        CFLAGS_FOR_BUILD: "-fsanitize=undefined,address"
+        UBSAN_OPTIONS: "print_stacktrace=1:halt_on_error=1"
+        ASAN_OPTIONS: "strict_string_checks=1:detect_stack_use_after_return=1:detect_leaks=1"
+        LSAN_OPTIONS: "use_unaligned=0"
```
jonasnick, line 274: that's the default according to https://github.com/google/sanitizers/wiki/AddressSanitizerLeakSanitizer

.cirrus.yml
```diff
@@ -262,3 +243,53 @@ task:
...
+    CTIMETEST: no
+    EXTRAFLAGS: "--disable-openssl-tests"
+  matrix:
+    - name: "Valgrind (memcheck)"
+      env:
+        # The `--error-exitcode` is required to make the test fail if valgrind found errors, otherwise it'll return 0 (https://www.valgrind.org/docs/manual/manual-core.html)
+        WRAPPER_CMD: "valgrind --error-exitcode=42"
+    - name: "UBSan, ASan, LSan"
+      env:
+        # Use dummy wrapper command to run the tests outside of make check
```
jonasnick, line 268: Remind me, what's the reason for not running this (and valgrind) in `make check`? If it was just about being able to set the test "count"/TEST_ITERS separately then we do have an env var for this now, and could remove it like this https://github.com/jonasnick/secp256k1/commits/202010_ci_matrix-jn (untested).

sipa: I think it's just that `make check` will try to run without the wrappers.

jonasnick: You mean `make check` is run without a wrapper even when a wrapper is set? Afaics no, because we unset `BUILD` whenever we set `WRAPPER_CMD`.

real-or-random: The main issue (and I think this is what Pieter is partly saying) is that `${WRAPPER_CMD} make check` typically won't work for all wrapper commands. 
 * `valgrind make check` will memcheck `make` but will not `memcheck` our tests because they'll be child processes. (Now you could try `valgrind --trace-children` but then it will still additionally memcheck `make` which gives false positives and wastes CI time.)  
 * `wine make check` just won't work at all because `make` is not a Windows executable.
 * I haven't tried QEMU but I expect the same as for Wine (because we cross-build).

What we could is to tell `make` to use the wrapper command. Now that I'm reading up on this again, I recall that it's not even more work to do this. There's the (obscurely named) `LOG_COMPILER` env variable, which does exactly this. You can even pass `LOG_COMPILER="./libtool --mode=execute valgrind"` etc, which we don't need now but maybe in the future. Running all tests using `make check` will have the benefit that the logs always end up in the same place in the CI output. Let me give it a shot.  

And thanks for reminding me of `SECP256K1_TEST_ITERS`. When I was writing this, I didn't know that this exists (I reinvented the name `TEST_ITERS`). So we should simply set this variable instead, like we do for the benchmarks. This has the nice side effect that it would apply to the ordinary `make check` builds if we'd like to set it there for some reason.

.cirrus.yml
real-or-random: Hm indeed. To be honest, I can't remember where this is from, it was already in my old branch. I think we should set it to `1` actually. Let me do this.

Some background: Both Valgrind and LSan can check for memory leaks, so we would not need LSan. But LSan is very cheap and none of these tools are perfect. So as long as we don't see false positives, it makes sense to just run LSan even though we have Valgrind. And it turns out that `use_unaligned=1` is indeed one thing where LSan has better coverage than Valgrind: https://stackoverflow.com/questions/33178242/valgrind-and-pragma-pack2  On my machine the difference in running time of `tests 64` was 99,95s with `use_unaligned=0` vs 100,02s `use_unaligned=1`, so this is just noise.

real-or-random:
updated as described above

ci/cirrus.sh
```diff
@@ -29,48 +29,32 @@ file *tests* || true
...
-if [ -n "$QEMU_CMD" ]
+if [ -n "$BUILD" ]
 then
-    $QEMU_CMD ./tests 16
-    $QEMU_CMD ./exhaustive_tests
+    make "$BUILD"
 fi
 
-if [ -n "$WINE_CMD" ]
+if [ -n "$WRAPPER_CMD" ]
```
jonasnick, line 44: If `LOG_COMPILER` works as described, can't we remove this branch? That's also why the CI jobs currently time out: the tests are being run twice.

ci/cirrus.sh
```diff
@@ -29,48 +29,32 @@ file *tests* || true
...
-    # the `--error-exitcode` is required to make the test fail if valgrind found errors, otherwise it'll return 0 (https://www.valgrind.org/docs/manual/manual-core.html)
-    valgrind --error-exitcode=42 ./tests 16
-    valgrind --error-exitcode=42 ./exhaustive_tests
-fi
+# This limits the iterations in the tests and benchmarks.
+export SECP256K1_TEST_ITERS="$TEST_ITERS"
+export SECP256K1_BENCH_ITERS="$BENCH_ITERS"
 
-if [ -n "$QEMU_CMD" ]
+if [ -n "$BUILD" ]
```
jonasnick, line 39: We don't need to check for this condition `$BUILD` is always set.

.cirrus.yml
jonasnick: cool, sgtm

ci/cirrus.sh
real-or-random: fixed

ci/cirrus.sh
real-or-random: fixed

real-or-random:
Hm, it's still timing out. But everything looks correct including the test iterations (like in the revision yesterday, when the longest job took ~ 40 min.) If I'm not mistaken, the only difference to now is that we run the tests in `make check` and this means the only real difference is that they run concurrently with `-j2` and we allocate only one CPU. I pushed a commit to see if this fixes the timeout.

real-or-random:
Ok, ignore my last comment. The tests were running twice because I screwed up when rebasing. I hope it's really ready now...

jonasnick (commented):
Nice!

ACK 02dcea1ad9441f857c7768e2b7d304bb19fd2a0c spot-checked ci output, checked that when `valgrind ./tests` crashes then `LOG_COMPILER=valgrind make check` also crashes.

sipa:
utACK 02dcea1ad9441f857c7768e2b7d304bb19fd2a0c

PR #848, opened 2020-11-20
use python3 print

fchapoton:


gmaxwell:
This is sage code, not python.

fchapoton:
sage is using now python3, since january 2020 and version 9.0. So this file is no longer working.

real-or-random:
Thanks, looks good! 

I noticed that `secp256k1.sage` does not work either with the current sage. This should also be easy to fix. Feel free to add a commit or indicate here that you're not willing to do this (then we can do it in another PR).

fchapoton:
Conversion to python3 is not an easy task. It usually requires care from somebody knowing the code.

jonasnick:
Merged as part of #849

PR #849, opened 2020-11-21
Convert Sage code to Python 3 (as used by Sage >= 9)

real-or-random:
This adds one more commit on top of #848 and squashes the commits.

Thanks to @fchapoton for spotting the issue and suggesting initial fixes.

### DIFF
--- a/sage/group_prover.sage
+++ b/sage/group_prover.sage
@@ -65,7 +65,7 @@ class fastfrac:
     return self.top in I and self.bot not in I
 
   def reduce(self,assumeZero):
-    zero = self.R.ideal(map(numerator, assumeZero))
+    zero = self.R.ideal(list(map(numerator, assumeZero)))
     return fastfrac(self.R, zero.reduce(self.top)) / fastfrac(self.R, zero.reduce(self.bot))
 
   def __add__(self,other):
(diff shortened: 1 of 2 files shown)


sage/group_prover.sage
```diff
@@ -100,7 +100,7 @@ class fastfrac:
     """Multiply something else with a fraction."""
     return self.__mul__(other)
 
-  def __div__(self,other):
+  def __truediv__(self,other):
```
jonasnick, line 103: This breaks sage 8.9. Can you leave the originas `__div__` definition (which shouldn't affect python 3) and add `__truediv__`? I tested that with this change both versions give the same output.

real-or-random: I can do this but I don't understand that this helps for sage 8.9 because it's based on python 2 and all the other changes should break it. So you're saying this works for you?

 https://wiki.sagemath.org/Python3-Switch

fchapoton: the changes about print, xrange and map are all harmless : they work with good old python2-based sage

real-or-random: Ok, I did the change.

sage/group_prover.sage
```diff
@@ -100,14 +100,19 @@ class fastfrac:
...
     if parent(other) == ZZ:
       return fastfrac(self.R,self.top,self.bot * other)
     if other.__class__ == fastfrac:
       return fastfrac(self.R,self.top * other.bot,self.bot * other.top)
     return NotImplemented
 
+  # Compatibility wrapper for Sage versions based on Python 2
+  def __div__(self,other):
+     """Divide two fractions."""
+     return __truediv__(self,other)
```
jonasnick, line 114: `NameError: global name '__truediv__' is not defined`. This should be `return self.__truediv__(other)`

real-or-random: Ok, I assumed I'm clever enough to write three lines of Python code without testing. But apparently I was wrong. (Sorry, it's a lot of work for me to install an older sage version...) Should work now.

jonasnick: Fwiw, `nix-shell -p sage` would work but the tests take forever. Therefore you can use `nix-shell '-p sage.overrideAttrs (oldAttrs: rec { buildInputs = [ makeWrapper ]; })'` to disable the tests.

jonasnick (approved):
ACK 13c88efed0005eb6745a222963ee74564054eafb

PR #850, opened 2020-11-22
add `secp256k1_ec_pubkey_cmp` method

apoelstra:


real-or-random:
Concept ACK

I would be easier to simply have a `compare` function which returns <0, ==0, or >0. Not sure which API is better. Did you think about this possibility?

 Implementation is then just a call to `memcmp`.

apoelstra:
I did think about this possibility, and if `secp256k1_memcmp_var` exposed this API then I probably would've just done a thin wrapper around that. But given that I had to do some extra work anyway, I felt that this API was more consistent than the libc one.

apoelstra:
I don't feel strongly about it, I'm happy to implement the memcmp API if that's what people want.

real-or-random:
> if `secp256k1_memcmp_var` exposed this API

Oh it does.

elichai:
I also think that exposing a memcmp-like single API is somewhat nicer (but this is also ok)

apoelstra:
Oh, neat, I misread the memcmp_var code. Ok, I'll just expose a single function then.

apoelstra:
Replaced with a single _cmp method.

jonasnick:
@apoelstra may I ask what you intend to use this for?

apoelstra:
@jonasnick it's needed for the `sortedmulti` descriptor and to use publickeys as keys in ordered sets (e.g. for PSBTs which should round-trip deterministically)

sipa:
@apoelstra The `sortedmulti` descriptor sorts lexicographically by the pubkey serialization, which means that uncompressed keys will sort after compressed ones.

apoelstra:
sigh, ok, I guess we cannot support comparisons in this library

real-or-random:
> sigh, ok, I guess we cannot support comparisons in this library

I'm not sure that `sortedmulti` stops us from exporting a comparison function that is compatible with compressed keys. This can be helpful for new applications which won't use the uncompressed serialization anyway, e.g., sorting keys in multisig. Moreover, orderings are in general helpful for some data structures, e.g., if you want to binary searches in a large set of keys.

A sad thing is that this ordering won't be compatible with the lexicographical ordering on x-only keys...

apoelstra:
Ironically, sorting by uncompressed encoding *would* be compatible with x-only keys.

elichai:
actually, what about adding these to xonly_pubkey too?

apoelstra:
sure

real-or-random:
A week ago I argued that this is helpful but I'm not so sure anymore.

If these functions are just serializing and memcmp'ing, they're not essential, so we could leave this to the user.

Pro: The user has to make a deliberate choice on an ordering, which may be good a thing here as things are complicated.
Con: We may end up with many different orderings in the wild.

apoelstra:
As a user I find it extremely irritating to need to serialize in order to compare keys for equality, so I think we should expose that....and then it's only epsilon more work to expose a total order.

real-or-random:
> As a user I find it extremely irritating to need to serialize in order to compare keys for equality, so I think we should expose that....

Agreed, equality is a no-brainer.

> and then it's only epsilon more work to expose a total order.

Indeed. I think in common cases when you want an order, you need it because you're going to serialize anyway, so the user probably won't be extremely irritated in that case. 


@sipa @jonasnick @elichai  What do you think? Should we expose a total order? (And if yes, which one? :P)

sipa:
@apoelstra Would it be useful to just have an equality test function? That avoids the rathole of trying to predict what ordering the user needs.

apoelstra:
Is there any actual opposition to merging this? Or just uncertainty about what ordering hypothetical users might want?

We could definitely use the xonly stuff (which is unambiguous) downstream now for MuSig so it would be cool to merge this.

apoelstra:
If it would be easier, I can remove the legacy stuff and only add the xonly comparison?

sipa:
utACK 48ac53b38068eb96474b785aaed4dbf1dc3dc586

This ordering is sufficiently natural that we can just merge it, I think.

src/modules/extrakeys/main_impl.h
```diff
@@ -55,6 +55,19 @@ int secp256k1_xonly_pubkey_serialize(const secp256k1_context* ctx, unsigned char
     return 1;
 }
 
+int secp256k1_xonly_pubkey_cmp(const secp256k1_context* ctx, const secp256k1_xonly_pubkey* pk1, const secp256k1_xonly_pubkey* pk2) {
+    unsigned char out1[32];
+    unsigned char out2[32];
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(pk1 != NULL);
+    ARG_CHECK(pk2 != NULL);
```
roconnor-blockstream, line 64: `ARG_CHECK` returns 0 for failure here, which isn't an appropriate return condition for a `cmp` function.  You can use `ARG_CHECK_NO_RETURN`, but you won't have a post-condition that the pointer is non-NULL, and you will still need to handle that case.  This can be done by zeroing the `out1` and `out2` buffers and only calling `secp256k1_xonly_pubkey_serialize` when `pk` is non NULL.

src/modules/extrakeys/main_impl.h
```diff
@@ -55,6 +55,19 @@ int secp256k1_xonly_pubkey_serialize(const secp256k1_context* ctx, unsigned char
...
+
+    CHECK(secp256k1_xonly_pubkey_serialize(ctx, out1, pk1));
```
roconnor-blockstream, line 66: Not sure why these are `CHECK` instead of `VERIFY_CHECK`.

jonasnick: `CHECK` and `VERIFY_CHECK` are both not right here because they abort if the check fails, making it impossible to test the function with illegal keys. I'd suggest to remove the `CHECK` and instead make sure that `out1` is all 0 if `pubkey_serialize` fails.

jonasnick:
@apoelstra I wrote a [fixup](https://github.com/jonasnick/secp256k1/commits/2020-11--pubkey-comparisons-jn) for this PR that makes the cmp functions define a consistent order even when dealing with NULL or illegal pubkeys by treating them as 0. It's used in [libsecp-zkp's musig aggregation PR](https://github.com/ElementsProject/secp256k1-zkp/pull/120/files#diff-5050fb6fb7b5cd270643e0ef126cc8b960fbdf10271257b525d00926a14620baR286) to add a sorting function for public keys.

real-or-random:
Oh CI fails again because this PR is too old and we switched to Cirrus CI. 

@apoelstra Can you quickly rebase this? This should solve it.

include/secp256k1.h
```diff
@@ -370,6 +371,21 @@ SECP256K1_API int secp256k1_ec_pubkey_serialize(
     unsigned int flags
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
 
+/** Compare two public keys using lexicographic (of compressed serialization) sort
```
jonasnick, line 374: nit: s/sort/order ? Same for xonly_pubkeys.

jonasnick (commented):
ACK mod nit

ramsemune:
C6S4P9U7

src/secp256k1.c
```diff
@@ -316,6 +316,28 @@ int secp256k1_ec_pubkey_serialize(const secp256k1_context* ctx, unsigned char *o
...
+        ARG_CHECK_NO_RETURN(pk[i] != NULL);
+        if (pk[i] == NULL
+                || !secp256k1_ec_pubkey_serialize(ctx, out[i], &out_size, pk[i], SECP256K1_EC_COMPRESSED)) {
+            out_size = sizeof(out[i]);
+            memset(out[i], 0, sizeof(out[i]));
+        }
+        VERIFY_CHECK(out_size == sizeof(out[i]));
+    }
+    return secp256k1_memcmp_var(out[0], out[1], sizeof(out[0]));
+}
```
real-or-random, line 339: I think we should call the `illegal_callback` also if a pubkey is non-null but invalid. This will almost always crash the program, which is much cleaner than trying to include invalid keys in the order.

Independently of that question, I'm not sure if it's a good idea to continue reading user-provided values if we know that something is wrong in the code of the user, so we should return immediately when we see an invalid value (after the `illegal_callback`).

According to the description in the header, should the `illegal_callback` ever return (which is not recommended), then the return value of our function is undefined. So `ARG_CHECK` would be fine, even if it returns 0. If we still don't like the return value of 0 in that case, we could something like      

```suggestion
int secp256k1_ec_pubkey_cmp(const secp256k1_context* ctx, const secp256k1_pubkey* pubkey0, const secp256k1_pubkey* pubkey1) {
    unsigned char out[2][33];
    size_t out_size = sizeof(out[0]);
    const secp256k1_pubkey* pk[2];
    int i;

    VERIFY_CHECK(ctx != NULL);
    pk[0] = pubkey0; pk[1] = pubkey1;
    for (i = 0; i < 2; i++) {
        int valid_pk;
        /* A public key that is NULL or invalid is treated as being smaller than
           any valid or invalid public key */
        ARG_CHECK_NO_RETURN(pk[i] != NULL);
        if (pk == NULL) return (i == 1 ? -1 : 1);
         
        valid_pk = secp256k1_ec_pubkey_serialize(ctx, out[i], &out_size, pk[i], SECP256K1_EC_COMPRESSED);
        ARG_CHECK_NO_RETURN(valid_pubkey);
        if (!valid_pubkey) return (i == 1 ? -1 : 1);
        
        VERIFY_CHECK(out_size == sizeof(out[i]));
    }
    return secp256k1_memcmp_var(out[0], out[1], sizeof(out[0]));
}
```

This also has the advantage that we'll never fire the `illegal_callback` more than once. I don't think we promise that it will fire only once but so far we have adhered to that convention, which is helpful when writing tests.

include/secp256k1.h
```diff
@@ -370,6 +371,21 @@ SECP256K1_API int secp256k1_ec_pubkey_serialize(
...
+ *
+ *  Returns: <0 if the first public key is less than the second
+ *           >0 if the first public key is greater than the second
+ *           0 otherwise
+ *  Args: ctx:      a secp256k1 context object.
+ *  In:   pubkey1:  first public key to compare
+ *        pubkey2:  second public key to compare
```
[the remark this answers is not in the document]
jonasnick: This is implicit as per the discussion in https://github.com/bitcoin-core/secp256k1/pull/783#issuecomment-695082220

src/secp256k1.c
jonasnick: This is implicit as per the discussion in https://github.com/bitcoin-core/secp256k1/pull/783#issuecomment-695082220

> I think we should call the illegal_callback also if a pubkey is non-null but invalid.

`ec_pubkey_serialize` will do that through `pubkey_load`. Or are you suggesting to do that explicitly here?

> So ARG_CHECK would be fine, even if it returns 0

Yes, the lib doesn't promise a proper return value in that case but if you give an inconsistent comparator to a sorting algorithm, it may never terminate. The goal was therefore to give a consistent response even if the keys are invalid. As far as I can see, your suggested code change doesn't achieve that because for two invalid pubkeys A and B we have A < B and B < A.

include/secp256k1.h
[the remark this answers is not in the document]
real-or-random: Okay, yes. Strictly speaking, we first need to add the line `Unless explicitly stated all pointer arguments must not be NULL.` in the header file (as done in #783) because currently we promise that the `illegal_callback` is only called for violations explicitly mentioned in the header.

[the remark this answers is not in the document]
jonasnick: added that line in #926

jonasnick:
Tim and I noticed that the comparison functions can be simplified because pubkey_serialize already ARG_CHECKs pk != NULL. We can rely on that because we check it in the API tests. I pushed fixups that also add comments on the rationale and fix the nits mentioned in this thread. @apoelstra if you think that makes sense, feel free to cherry-pick and squash.
https://github.com/jonasnick/secp256k1/tree/2020-11--pubkey-comparisons-jn

include/secp256k1.h
```diff
@@ -370,6 +371,21 @@ SECP256K1_API int secp256k1_ec_pubkey_serialize(
     unsigned int flags
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
 
+/** Compare two public keys using lexicographic (of compressed serialization) sort
+ *
+ *  Returns: <0 if the first public key is less than the second
+ *           >0 if the first public key is greater than the second
+ *           0 otherwise
```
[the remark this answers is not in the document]
real-or-random: Resolved in https://github.com/jonasnick/secp256k1/tree/2020-11--pubkey-comparisons-jn?rgh-link-date=2021-04-29T15%3A08%3A40Z

src/secp256k1.c
real-or-random: Resolved in https://github.com/jonasnick/secp256k1/tree/2020-11--pubkey-comparisons-jn

real-or-random:
@jonasnick Your branch looks good to me, I suggest we merge that one.

apoelstra:
Squashed in jonas' commits.

jonasnick (commented):
ACK 6eceec6d566898a5c157630e47f95b260767026b

elichai:
Code review ACK 6eceec6d566898a5c157630e47f95b260767026b

real-or-random (approved):
ACK 6eceec6d566898a5c157630e47f95b260767026b

include/secp256k1_extrakeys.h
```diff
@@ -67,6 +67,21 @@ SECP256K1_API int secp256k1_xonly_pubkey_serialize(
     const secp256k1_xonly_pubkey* pubkey
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
 
+/** Compare two x-only public keys using lexicographic order
+ *
+ *  Returns: <0 if the first public key is less than the second
+ *           >0 if the first public key is greater than the second
+ *           0 if the two public keys are equal
+ *  Args: ctx:      a secp256k1 context object.
```
roconnor-blockstream, line 75: non-null comments here too?

jonasnick: unnecessary if #926 gets merged (needs more ACKs :] )

PR #851, opened 2020-11-23
make test count iteration configurable by environment variable

apoelstra:
You can set the test iteration count by command-line parameter but it is very hard to control the iteration count used by `make check` and even harder to control the count when this is run through Bitcoin Core's build system. Using an environment var solves both problems.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -5632,6 +5632,10 @@ int main(int argc, char **argv) {
             count = strtol(env, NULL, 0);
         }
     }
+    if (count <= 0) {
+        fputs("An iteration count of 0 or less is not allowed.\n", stderr);
+        return EXIT_FAILURE;
+    }
     printf("test count = %i\n", count);
 
     /* find random seed */
--- a/src/tests.c
+++ b/src/tests.c
@@ -5626,6 +5626,11 @@ int main(int argc, char **argv) {
     /* find iteration count */
     if (argc > 1) {
         count = strtol(argv[1], NULL, 0);
+    } else {
+        const char* env = getenv("SECP256K1_TEST_ITERS");
+        if (env) {
+            count = strtol(env, NULL, 0);
+        }
     }
     printf("test count = %i\n", count);
 


real-or-random:
Nice, that's indeed useful. 

ACK 0ce45548813709d828cb3abcc7db4c9ce6e26907 I inspected the diff

Here are a few small annoying things that you can add to this PR if you want. If not, let's consider this a list for another PR:
 * Add check that aborts if not `count > 0`? Technically `count == 0` is fine but I think noone wants to run this on purpose as it just disables some tests arbitrary. 
 * I was tempted to say that we should have a `ITERS` also for the exhaustive tests but I think that's not true. The default of 2 makes a lot of sense for these, because the randomness is only used to randomize the representation in jacobian coordinates. But we should add the same `count > 0` check there.
 * Document the environment variables. Since we don't have a .md for tests, the canonical place is in the "usage" output. (This is not optimal because the benchmarks have an env variable but no "usage" output. Yeah, ok.)

jonasnick (approved):
ACK 0ce45548813709d828cb3abcc7db4c9ce6e26907

elichai:
utACK 0ce45548813709d828cb3abcc7db4c9ce6e26907

> * Add check that aborts if not `count > 0`? Technically `count == 0` is fine but I think noone wants to run this on purpose as it just disables some tests arbitrary.

I also wanted to do some sanitizing on the bench env var but hadn't gotten to doing it yet
https://github.com/bitcoin-core/secp256k1/blob/ca4906b02e69644d83e13b3c09cc9147fc1232a5/src/bench.h#L124

apoelstra:
deadalnix was using count==0 :) https://github.com/bitcoin-core/secp256k1/issues/528

elichai:
> deadalnix was using count==0 :) #528
lol
I didn't fix yet because I think those env vars are for "advanced users only" and only tests/benchmarks so it doesn't matter too much

apoelstra:
I'm happy to add a count>0 check if people think it's valuable.

I don't think it's worth exposing a way to change the exhaustive test iteration count. I didn't even know there was one.

Unfortunately there is no usage text for the tests binary :). I could add an unconditional message above the "count =" line.

sipa:
The count variable for the exhaustive tests changes the z coordinate for gej elements (which is randomized, not exhaustive).

I don't think there is much need to changing the default, though.

real-or-random:
> deadalnix was using count==0 :) #528

yeah but even he says:
> My concern here isn't really that 0 is a very useful thing to test for, but that it put the system is some weird state that reveal a bug either in libsecp256k1 or possibly in gcc's optimizer. 

So I think we should just enforce > 0. But either way is fine.

> I don't think there is much need to changing the default, though.

Indeed.

src/tests.c
```diff
@@ -5626,6 +5626,15 @@ int main(int argc, char **argv) {
     /* find iteration count */
     if (argc > 1) {
         count = strtol(argv[1], NULL, 0);
+    } else {
+        const char* env = getenv("SECP256K1_TEST_ITERS");
+        if (env) {
+            count = strtol(env, NULL, 0);
+        }
+    }
+    if (count == 0) {
```
real-or-random, line 5635: Maybe just
```suggestion
    if (count <= 0) {
```

real-or-random (approved):
ACK a729a6c6bfb07929c56222fc43cba790d9e5c083 I inspected the diff but didn't test

src/tests.c
```diff
@@ -5626,6 +5626,15 @@ int main(int argc, char **argv) {
...
     if (argc > 1) {
         count = strtol(argv[1], NULL, 0);
+    } else {
+        const char* env = getenv("SECP256K1_TEST_ITERS");
+        if (env) {
+            count = strtol(env, NULL, 0);
+        }
+    }
+    if (count <= 0) {
+        fputs("An iteration count of 0 is not allowed.\n", stderr);
```
jonasnick, line 5636: nit: `<= 0`

jonasnick (commented):
ACK mod nit

jonasnick (approved):
ACK f4fa8d226a95e42b252c07edb425c446370e01c0

real-or-random (approved):
ACK f4fa8d226a95e42b252c07edb425c446370e01c0

PR #852, opened 2020-11-25
Add sage script for generating scalar_split_lambda constants

real-or-random:
cc @roconnor-blockstream


roconnor-blockstream:
I recommend trying `LLL` to `replace find_split_constants_explicit_tof`.

Add/replace the BETA and LAMBDA assertions with an assertion that the value is a root of `X^2 + X + 1`.

real-or-random:
> I recommend trying `LLL` to `replace find_split_constants_explicit_tof`.

I think the current one is more instructive... I don't know, I don't want to spend a lot of time on this.

roconnor-blockstream:
I'm not really familiar with sage, and I haven't tested this PR, but it is okay by me.

sage/gen_split_lambda_constants.sage
```diff
@@ -0,0 +1,97 @@
...
+    B1, B2 = -B1, -B2
+
+    return A1, B1, A2, B2
+
+A1, B1, A2, B2 = find_split_constants_explicit_tof()
+
+# For extra fun, use an independent method to recompute the constants
+assert (A1, B1, A2, B2) == find_split_constants_gauss()
+
+assert A1*B2 - B1*A2 == N
```
roconnor-blockstream, line 56: If you want to add a comment, I think it would be that this determinant check is verifying that the vectors, um, bound? surround? a fundamental domain (of the ring homomorphism from Z[-1/2+i*sqrt(3)/2] to Zn).

roconnor-blockstream: If you move this assertion below the next two you could argue that this is proving that the two vectors are not only in the kernel but generate the kernel as well and phrase the comment that way.

(note the determinant also proves the vectors are linearly independent by being non-zero).

real-or-random: > If you want to add a comment, I think it would be that this determinant check is verifying that the vectors, um, bound? surround? a fundamental domain (of the ring homomorphism from Z[-1/2+i*sqrt(3)/2] to Zn).

Can you suggest a wording? I hadn't heard "fundamental domain" before in my life.

> If you move this assertion below the next two you could argue that this is proving that the two vectors are not only in the kernel but generate the kernel as well and phrase the comment that way.
> 
> (note the determinant also proves the vectors are linearly independent by being non-zero).

Indeed, that's easier. But we could still say why it's exactly n.

roconnor-blockstream: I don't know.  Maybe I'd say, "Check that he parallelogram generated by <A1, A2> and <B1,B2> is a fundamental domain that contains exactly N points."

roconnor-blockstream:
Another check to add is that `lambda*P = <beta*P.x, y>`.  Checking for `P=G` is probably sufficient.  The point is that the two roots of X^2+X+1 in Zn and Zp need to be matched up in accordance to the group operation on the secp256k1 curve.

sage/gen_split_lambda_constants.sage
```diff
@@ -0,0 +1,97 @@
...
+    return max(map(abs, v))
+
+def gauss_reduction(i1, i2):
+    v1, v2 = i1.copy(), i2.copy()
+    while True:
+        if inf_norm(v2) < inf_norm(v1):
+            v1, v2 = v2, v1
+        m = round( (v1[0]*v2[0] + v1[1]*v2[1]) / inf_norm(v1)**2 )
+        if m == 0:
+            return (v1, v2)
```
sipa, line 14: No need for parentheses.

sage/gen_split_lambda_constants.sage
```diff
@@ -0,0 +1,97 @@
...
+
+def find_split_constants_gauss():
+    """Find constants for secp256k1_scalar_split_lamdba using gauss reduction."""
+    (v11, v12), (v21, v22) = gauss_reduction([0, N], [1, int(LAMBDA)])
+
+    # We use related vectors in secp256k1_scalar_split_lambda
+    A1, B1 = -v21, -v11
+    A2, B2 = v22, -v21
+
+    return (A1, B1, A2, B2)
```
sipa, line 26: No need for parentheses.

sage/gen_split_lambda_constants.sage
```diff
@@ -0,0 +1,97 @@
...
+
+A1, B1, A2, B2 = find_split_constants_explicit_tof()
+
+# For extra fun, use an independent method to recompute the constants
+assert (A1, B1, A2, B2) == find_split_constants_gauss()
+
+assert A1*B2 - B1*A2 == N
+
+# Check that (A1, B1) and (A2, B2) are in the kernel
+assert Z(A1 + LAMBDA*B1) == Z(0)
```
sipa, line 59: Why the Z()?

real-or-random:
I addressed all of your comments.

sage/gen_split_lambda_constants.sage
```diff
@@ -0,0 +1,111 @@
...
+def inf_norm(v):
+    """Returns the infinity norm of a vector."""
+    return max(map(abs, v))
+
+def gauss_reduction(i1, i2):
+    v1, v2 = i1.copy(), i2.copy()
+    while True:
+        if inf_norm(v2) < inf_norm(v1):
+            v1, v2 = v2, v1
+        m = round( (v1[0]*v2[0] + v1[1]*v2[1]) // inf_norm(v1)**2 )
```
sipa, line 17: No need to round anymore with `//`.

sage/gen_split_lambda_constants.sage
```diff
@@ -0,0 +1,111 @@
...
+        if m == 0:
+            return v1, v2
+        v2[0] = v2[0] - m*v1[0]
```
sipa, line 20: `-=` ?

sipa:
Looks good, ACK 4e4158e653d902f7f2c619d8a76ddf658f200d9e

sage/gen_split_lambda_constants.sage
real-or-random: Oh, I thought `/` vs `//` is just a python2 compatibility thing but I think what you had in mind is to avoid floating-point arithmetic. But this changed the semantics.

The correct thing is to `round` (to the nearest int) and not `floor` (for a reason), so that's why I used `/` and `round` instead of `//`. (It just happens that the output is the same *for our specific input*.)
   
So I now changed this to a version which
 * rounds to the nearest integer (what the textbooks do TM [1]),
 * relies integer arithmetic instead of floating-point arithmetic, and
 * should work in Python 2 and Python 3 because it does not use `/`.


[1] https://en.wikipedia.org/wiki/Lattice_reduction#In_two_dimensions Ok not a book but you get the point.

roconnor-blockstream: Just saying, you could delete the whole thing and call `LLL` where all this has been worked out.

sipa:
ACK 329a2e0a3f2d9e936179cbf079773538f95bee33

CI failure is unrelated.

PR #854, opened 2020-12-03
Rename msg32 to msghash32 in ecdsa_sign/verify and add explanation

### LINKED ISSUE #307: Check argument names for confusability.
msg32 -- some people might think that this is the actual message, without hashing.

jonasnick:
This fixes #307 if there's nothing else that's confusing.

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -452,7 +452,14 @@ SECP256K1_API int secp256k1_ecdsa_signature_serialize_compact(
  *           0: incorrect or unparseable signature
  *  Args:    ctx:       a secp256k1 context object, initialized for verification.
  *  In:      sig:       the signature being verified (cannot be NULL)
- *           msg32:     the 32-byte message hash being verified (cannot be NULL)
+ *           msghash32: the 32-byte message hash being verified (cannot be NULL).
+ *                      The verifier must make sure to apply a cryptographic
+ *                      hash function to the message by itself and not accept an
+ *                      msghash32 value directly. Otherwise, it would be easy to
+ *                      create a "valid" signature without knowledge of the
+ *                      secret key. See also
+ *                      https://bitcoin.stackexchange.com/a/81116/35586 for more
+ *                      background on this topic.
  *           pubkey:    pointer to an initialized public key to verify with (cannot be NULL)
  *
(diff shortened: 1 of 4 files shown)


include/secp256k1.h
```diff
@@ -452,7 +452,13 @@ SECP256K1_API int secp256k1_ecdsa_signature_serialize_compact(
...
  *  Args:    ctx:       a secp256k1 context object, initialized for verification.
  *  In:      sig:       the signature being verified (cannot be NULL)
- *           msg32:     the 32-byte message hash being verified (cannot be NULL)
+ *           msghash32: the 32-byte message hash being verified. The verifier
+ *                      must make sure to apply a cryptographic hash function to
+ *                      the message by itself and not accept an msghash32 value
+ *                      directly. Otherwise, it would be easy to create a
+ *                      "valid" signature without knowledge of the secret key.
+ *                      See also https://bitcoin.stackexchange.com/a/81116/35586
+ *                      for more background on this topic. (cannot be NULL)
```
real-or-random, line 461: maybe move the "cannot be NULL" up again.

Otherwise it's "See also https://bitcoin.stackexchange.com/a/81116/35586 for more background on this topic. (cannot be NULL)", which confused me. We could also move the explanation to the body alternatively.

real-or-random (commented):
We should do the same for the functions in the recovery module.

One more thing that I noticed is that extrakeys module says "tweak32" whereas the other functions have just "tweak". I think "tweak32" is better. The number suffixes are really neat.

I think the only exception where we don't use them is "seckey". (Maybe because this thing is so common that people should not get it wrong.)

include/secp256k1.h
jonasnick: Fixed

jonasnick:
Good points. Renamed msg32 also in recovery module and added commit that renames tweak to tweak32.

real-or-random:
ACK 6e85d675aaf9dc17842096f9cbf8cfab216c9331 I inspected the diff

apoelstra (approved):
ack 6e85d675aaf9dc17842096f9cbf8cfab216c9331

gmaxwell:
looks fine to me.

PR #857, opened 2020-12-08
docs: fix simple typo, dependecy -> dependency

timgates42:
There is a small typo in src/group_impl.h.

Should read `dependency` rather than `dependecy`.


real-or-random (approved):
ACK 18aadf9d288a54533376cb94f655d059eb1f098e

real-or-random:
Thanks. 

Note: I removed the URL from your opening post here to avoid that our merge script will pick it up and include it in the git history.

PR #862, opened 2020-12-23
Autoconf improvements

real-or-random:
See individual commit messages. These are improvements in preparation of the switch to Cirrus CI. (Maybe I'll just open a PR on top of this one.)

The first commit made the difference between successful build https://cirrus-ci.com/task/6740575057608704 and unsuccessful build https://cirrus-ci.com/task/4909571074424832.

I've tested the second commit without cross-compilation and with cross-compilation for android (https://github.com/bitcoin-core/secp256k1/issues/621#issuecomment-495703399)

When working on the autoconf stuff, I noticed two things that I just want to write down here:
 - At some point we should update [build-aux/m4/ax_prog_cc_for_build.m4](https://www.gnu.org/software/autoconf-archive/ax_prog_cc_for_build.html). This is outdated, and [there have been a lot of fixes](https://github.com/autoconf-archive/autoconf-archive/pull/207) But the latest version is [broken](https://lists.gnu.org/archive/html/autoconf-archive-maintainers/2020-06/msg00002.html), so now is probably not the time.
 - The latest autoconf 2.70 deprecates `AC_PROG_CC_C89`. It's not needed anymore because `AC_PROG_CC` cares about testing for version support. This makes autoconf 2.70 output a warning that we should probably just ignore. We don't want to force users onto 2.70...

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -40,9 +40,9 @@ case $host_os in
          dnl These Homebrew packages may be keg-only, meaning that they won't be found
          dnl in expected paths because they may conflict with system files. Ask
          dnl Homebrew where each one is located, then adjust paths accordingly.
-
          openssl_prefix=`$BREW --prefix openssl 2>/dev/null`
          gmp_prefix=`$BREW --prefix gmp 2>/dev/null`
+         valgrind_prefix=`$BREW --prefix valgrind 2>/dev/null`
          if test x$openssl_prefix != x; then
            PKG_CONFIG_PATH="$openssl_prefix/lib/pkgconfig:$PKG_CONFIG_PATH"
            export PKG_CONFIG_PATH
@@ -51,6 +51,9 @@ case $host_os in
          if test x$gmp_prefix != x; then
            GMP_CPPFLAGS="-I$gmp_prefix/include"
            GMP_LIBS="-L$gmp_prefix/lib"
+         fi
+         if test x$valgrind_prefix != x; then
+           VALGRIND_CPPFLAGS="-I$valgrind_prefix/include"
          fi
        else
          AC_PATH_PROG([PORT],port,)
@@ -180,12 +183,15 @@ AC_ARG_WITH([valgrind], [AS_HELP_STRING([--with-valgrind=yes|no|auto],
 if test x"$req_valgrind" = x"no"; then
   enable_valgrind=no
 else
-  AC_CHECK_HEADER([valgrind/memcheck.h], [enable_valgrind=yes], [
+  SECP_VALGRIND_CHECK
+  if test x"$has_valgrind" != x"yes"; then
     if test x"$req_valgrind" = x"yes"; then
       AC_MSG_ERROR([Valgrind support explicitly requested but valgrind/memcheck.h header not available])
     fi
     enable_valgrind=no
-  ], [])
+  else
+    enable_valgrind=yes
+  fi
 fi
 AM_CONDITIONAL([VALGRIND_ENABLED],[test "$enable_valgrind" = "yes"])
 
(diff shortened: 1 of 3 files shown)


michaelfolkson:
Installing Valgrind on MacOS Big Sur to test this seems problematic https://stackoverflow.com/questions/65009780/macos-valgrind-alternative

real-or-random:
@michaelfolkson Indeed yeah, I'm using https://github.com/LouisBrunner/valgrind-macos on my branch. https://github.com/real-or-random/secp256k1/blob/202012-cirrusci/.cirrus.yml#L113

This seems to do the job. I hope it's stable enough that we won't have to fix this once per month.

michaelfolkson:
Required a reinstall of the MacOS CommandLineTools to install that LouisBrunner version of Valgrind. But this PR seems to work fine. Before this PR `CPPFLAGS` was empty on configuring and after this PR `CPPFLAGS` contains` -I/usr/local/opt/valgrind/include`

real-or-random:
> Required a reinstall of the MacOS CommandLineTools to install that LouisBrunner version of Valgrind.

Yep, this was necessary also here previously, see the commented lines in `.cirrus.yml`. But the Cirrus CI images have been upgraded now: https://github.com/cirruslabs/osx-images/issues/30

DvalCaruso (approved):

DvalCaruso (approved):

DvalCaruso (approved):

sipa:
@DvalCaruso Stop spamming this repository with meaningless operations. You will get banned if you continue.

sipa:
Is it possible to split the 2nd commit into ones that move stuff around, add comments, and provide the no-crosscompile special casing?

configure.ac
```diff
@@ -305,6 +263,79 @@ else
...
+    set_precomp=yes
+    if test x"${CC_FOR_BUILD+x}${CFLAGS_FOR_BUILD+x}${CPPFLAGS_FOR_BUILD+x}${LDFLAGS_FOR_BUILD+x}" != x; then
+      AC_MSG_WARN([CC_FOR_BUILD, CFLAGS_FOR_BUILD, CPPFLAGS_FOR_BUILD, and/or LDFLAGS_FOR_BUILD is set but ignored because we are not cross-compiling.])
+    fi
+    # If we're not cross-compiling, simply use the same compiler for building the static precompation code.
+    CC_FOR_BUILD="$CC"
+    CFLAGS_FOR_BUILD="$CFLAGS"
+    CPPFLAGS_FOR_BUILD="$CPPFLAGS"
+    LDFLAGS_FOR_BUILD="$LDFLAGS"
+  else
```
real-or-random, line 278: @sipa I think this is one part. But yes, I can split it if you want. I was somewhat lazy because I was hacking on the code at various places and then just had a bunch of changes...

sipa: @real-or-random Yes, sorry, I realize that that is what this part does. The code looks good, but it's easier to ascertain it doesn't introduce other changes if the move-only part and the rest are in separate commits.

real-or-random:
@sipa Ok, I split a commit. It's now easier to review this commit by commit. The second commit should introduce no behavioral changes.When viewing the (now) third commit, it's a good idea to hide whitespace changes because otherwise the diff will again be large due to changes in indentation.

sipa:
utACK 3c15130709da26a6d2f25a483aa45e14bf1e4feb?w=1

jonasnick:
utACK 3c15130709da26a6d2f25a483aa45e14bf1e4feb makes sense (with my very basic understanding of autoconf)

jonasnick (commented):
utACK 3c15130 makes sense (with my very basic understanding of autoconf)

real-or-random:
I think two utACKs are okay for this to be merged but let me rebase #864 on top of this, and then we can at least see this once being run on and tested on Cirrus CI. Then we can for example see if the brew commit still works, which was introduced to make the Cirrus PR work.

real-or-random:
> I think two utACKs are okay for this to be merged but let me rebase #864 on top of this, and then we can at least see this once being run on and tested on Cirrus CI. Then we can for example see if the brew commit still works, which was introduced to make the Cirrus PR work.

Ok, this work on CI: https://cirrus-ci.com/build/4517681179131904, merging then

Issue #870, opened 2021-01-20
IFDEFs beginning with underscores


roconnor-blockstream:
The IFDEFs in the header files in schnorrsig and extrakeys module directories all begin with `_`.

PR #874, opened 2021-01-23
Remove underscores from header defs.

roconnor-blockstream:
This makes them consistent with other files and avoids reserved identifiers.

Fixes <https://github.com/bitcoin-core/secp256k1/issues/870>.

See also <https://wiki.sei.cmu.edu/confluence/display/c/DCL37-C.+Do+not+declare+or+define+a+reserved+identifier>.

### DIFF
--- a/src/gen_context.c
+++ b/src/gen_context.c
@@ -47,8 +47,8 @@ int main(int argc, char **argv) {
         return -1;
     }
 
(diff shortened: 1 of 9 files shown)


real-or-random (approved):
utACK fb390c5299e999e06b7dff9e77e373600fae9fdf

jonasnick (commented):
ACK fb390c5299e999e06b7dff9e77e373600fae9fdf

PR #875, opened 2021-01-23
Avoid casting (void**) values.

roconnor-blockstream:
Replaced with an expression that only casts (void*) values.

While casting between `void *` and `foo *` generally allowed (upto alignment issues), the same cannot be said for casting between `void **` and `foo **` values, which is not okay in general.

While the C standard which says "A pointer to void shall have the same representation and alignment requirements as a pointer to a character type.” likely implies the cast is sound because `void *` and `char *` values have the same representation, perhaps it is better to rewrite the expression to not require a careful reading of the standard to be correct.


sipa:
utACK 27306186045955803ca4070783c6a28853003e6a

I can't imagine that on realistic platforms there is a difference between the two, so if the new code is more convincingly valid by a strict reading of the standard, let's use that.

real-or-random (approved):
utACK https://github.com/bitcoin-core/secp256k1/pull/875/commits/27306186045955803ca4070783c6a28853003e6a

This really reminds me of https://github.com/bitcoin-core/secp256k1/pull/566#discussion_r228187002.

jonasnick (approved):
utACK 27306186045955803ca4070783c6a28853003e6a

Issue #871, opened 2021-01-20
secp256k1_ge_set_gej_var missing declairation.


roconnor-blockstream:
The declaration for `secp256k1_ge_set_gej_var` is missing from `group.h`.

PR #877, opened 2021-01-24
Add missing secp256k1_ge_set_gej_var decl.

roconnor-blockstream:
Fixes <https://github.com/bitcoin-core/secp256k1/issues/871>.

### DIFF
--- a/src/group.h
+++ b/src/group.h
@@ -62,8 +62,11 @@ static int secp256k1_ge_is_valid_var(const secp256k1_ge *a);
 /** Set r equal to the inverse of a (i.e., mirrored around the X axis) */
 static void secp256k1_ge_neg(secp256k1_ge *r, const secp256k1_ge *a);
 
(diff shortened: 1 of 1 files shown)


sipa:
utACK 482e4a9cfcecad28c3b7e601667b3b41949f47fe

Seems obviously correct.

Note that these declarations don't really serve as anything more than somewhat machine-checked documentation. They're internal functions only declared and used afterwards within the same compilation unit. But this one is part of the (internally) exposed interface of the group "module", so it's better to have it explicit.

real-or-random (approved):
utACK https://github.com/bitcoin-core/secp256k1/pull/877/commits/482e4a9cfcecad28c3b7e601667b3b41949f47fe

jonasnick (commented):
ACK 482e4a9cfcecad28c3b7e601667b3b41949f47fe

PR #879, opened 2021-01-24
Avoid passing out-of-bound pointers to 0-size memcpy

### LINKED ISSUE #876: Potential UB in secp256k1_der_parse_integer?
Do we consider `memcpy(ra + 32 - rlen, *sig, rlen)` in `secp256k1_der_parse_integer` to be defined in the case of `rlen == 0`?

https://github.com/bitcoin-core/secp256k1/blob/98dac87839838b86094f1bccc71cc20e67b146cc/src/ecdsa_impl.h#L102-L104

https://github.com/bitcoin-core/secp256k1/blob/98dac87839838b86094f1bccc71cc20e67b146cc/src/ecdsa_impl.h#L143

Nothing high priority of course, but perhaps worth fixing? :)

This code was introduced as part of PR #334 ("Overhaul ECDSA signature parsing: strict DER, compact sigs, tests, lower-S") in 3bb9c447193a6d15726976c141f27d2c3ddd7991 back in 2015.

sipa:
Doing so could be considered UB in a pedantic interpretation of the standard. Avoid it.

Closes #876.

### DIFF
--- a/contrib/lax_der_parsing.c
+++ b/contrib/lax_der_parsing.c
@@ -121,7 +121,7 @@ int ecdsa_signature_parse_der_lax(const secp256k1_context* ctx, secp256k1_ecdsa_
     /* Copy R value */
     if (rlen > 32) {
         overflow = 1;
-    } else {
+    } else if (rlen) {
         memcpy(tmpsig + 32 - rlen, input + rpos, rlen);
     }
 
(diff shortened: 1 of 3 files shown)


practicalswift:
cr ACK 9570f674cc729cafcba65f4cce03552d9a6108f4: patch looks correct

real-or-random:
When we change the contrib functions, we should also consider changing the copies in Bitcoin Core, see https://github.com/bitcoin-core/secp256k1/pull/781 

Can you also change the comparison in the same function here https://github.com/bitcoin-core/secp256k1/blob/9570f674cc729cafcba65f4cce03552d9a6108f4/src/ecdsa_impl.h#L115

to read `rlen > (size_t)(sigend - *sig)` ? This will never compute an out-of-bounds pointer (which is another case of pedantic UB).

real-or-random:
@sipa ping

real-or-random:
ACK 9570f674cc729cafcba65f4cce03552d9a6108f4

I'll open a new PR for the other issue

PR #882, opened 2021-01-28
Use bit ops instead of int mult for constant-time logic in gej_add_ge

real-or-random:
Nothing is wrong with the current implementation. But it looks out of place because we generally prefer bit operations because they're faster and are more likely constant-time.

### DIFF
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -591,7 +591,7 @@ static void secp256k1_gej_add_ge(secp256k1_gej *r, const secp256k1_gej *a, const
     secp256k1_fe_cmov(&n, &m, degenerate);              /* n = M^3 * Malt (2) */
     secp256k1_fe_sqr(&t, &rr_alt);                      /* t = Ralt^2 (1) */
     secp256k1_fe_mul(&r->z, &a->z, &m_alt);             /* r->z = Malt*Z (1) */
-    infinity = secp256k1_fe_normalizes_to_zero(&r->z) * (1 - a->infinity);
+    infinity = secp256k1_fe_normalizes_to_zero(&r->z) & ~a->infinity;
     secp256k1_fe_mul_int(&r->z, 2);                     /* r->z = Z3 = 2*Malt*Z (2) */
     secp256k1_fe_negate(&q, &q, 1);                     /* q = -Q (2) */
     secp256k1_fe_add(&t, &q);                           /* t = Ralt^2-Q (3) */


sipa:
utACK e491d06b98c9caa5eab74e38ba8419b9fb3b5015. Seems obviously better.

src/group_impl.h
```diff
@@ -591,7 +591,7 @@ static void secp256k1_gej_add_ge(secp256k1_gej *r, const secp256k1_gej *a, const
     secp256k1_fe_cmov(&n, &m, degenerate);              /* n = M^3 * Malt (2) */
     secp256k1_fe_sqr(&t, &rr_alt);                      /* t = Ralt^2 (1) */
     secp256k1_fe_mul(&r->z, &a->z, &m_alt);             /* r->z = Malt*Z (1) */
-    infinity = secp256k1_fe_normalizes_to_zero(&r->z) * (1 - a->infinity);
+    infinity = secp256k1_fe_normalizes_to_zero(&r->z) & ~a->infinity;
```
elichai, line 594: Isn't `& !a->infinity` clearer?

real-or-random: I can't show you any evidence right but from my experience `!` is much more likely to make compilers introduce branches.

elichai: We already assume it is constant time in a few places:
```
$ rg "&= !"
src/modules/schnorrsig/main_impl.h
159:    ret &= !!noncefp(buf, msg32, seckey, pk_buf, bip340_algo16, ndata);
161:    ret &= !secp256k1_scalar_is_zero(&k);

src/secp256k1.c
370:    ret &= !overflow;
372:    ret &= !overflow;

src/tests_exhaustive.c
281:                    should_verify &= !secp256k1_scalar_is_high(&s_s);

src/modules/recovery/main_impl.h
49:    ret &= !overflow;
51:    ret &= !overflow;

src/modules/recovery/tests_exhaustive_impl.h
124:                    should_verify &= !secp256k1_scalar_is_high(&s_s);
```

sipa: @elichai Compilers don't translate every C operator the same way. We know the current code works (valgrind ctime test) with common compilers on common platforms, but that's no guarantee for the future, or a guarantee that a `~` is never compiled into a branch in other contexts.

There is no downside to using `~`, right?

elichai: I guess you're right. And I believe @real-or-random when he says that `!` is more likely to introduce branches :)

elichai:
ACK e491d06b98c9caa5eab74e38ba8419b9fb3b5015

src/group_impl.h
real-or-random: Good remark that we use `!` other places, I wasn't really aware of this anymore. But ok, I'd say let's keep it then.

jonasnick (commented):
ACK e491d06b98c9caa5eab74e38ba8419b9fb3b5015

PR #889, opened 2021-01-31
fix uninitialized read in tests

PiRK:
I tried compiling secp256k1 with clang and `-Wconditional-uninitialized` and got the following warning:

```
$ make check
  CC       src/tests-tests.o
src/tests.c:4336:15: warning: variable 'recid' may be uninitialized when used here [-Wconditional-uninitialized]
        CHECK(recid >= 0 && recid < 4);
              ^~~~~
./src/util.h:54:18: note: expanded from macro 'CHECK'
    if (EXPECT(!(cond), 0)) { \
                 ^~~~
./src/util.h:41:39: note: expanded from macro 'EXPECT'
#define EXPECT(x,c) __builtin_expect((x),(c))
                                      ^
src/tests.c:4327:14: note: initialize the variable 'recid' to silence this warning
    int recid;
             ^
              = 0
1 warning generated.
  CCLD     tests
make  check-TESTS
```

This initializes `recid` and adds the  `-Wconditional-uninitialized` flag when building with clang.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -80,6 +80,15 @@ AC_COMPILE_IFELSE([AC_LANG_SOURCE([[char foo;]])],
     ])
 
 saved_CFLAGS="$CFLAGS"
+CFLAGS="-Wconditional-uninitialized $CFLAGS"
+AC_MSG_CHECKING([if ${CC} supports -Wconditional-uninitialized])
+AC_COMPILE_IFELSE([AC_LANG_SOURCE([[char foo;]])],
+    [ AC_MSG_RESULT([yes]) ],
+    [ AC_MSG_RESULT([no])
+      CFLAGS="$saved_CFLAGS"
+    ])
+
+saved_CFLAGS="$CFLAGS"
 CFLAGS="-fvisibility=hidden $CFLAGS"
 AC_MSG_CHECKING([if ${CC} supports -fvisibility=hidden])
 AC_COMPILE_IFELSE([AC_LANG_SOURCE([[char foo;]])],
--- a/src/tests.c
+++ b/src/tests.c
@@ -4324,8 +4324,10 @@ void test_ecdsa_sign_verify(void) {
     secp256k1_scalar one;
     secp256k1_scalar msg, key;
     secp256k1_scalar sigr, sigs;
-    int recid;
     int getrec;
+    /* Initialize recid to suppress a false positive -Wconditional-uninitialized in clang.
+       VG_UNDEF ensures that valgrind will still treat the variable as uninitialized. */
+    int recid = -1; VG_UNDEF(&recid, sizeof(recid));
     random_scalar_order_test(&msg);
     random_scalar_order_test(&key);
     secp256k1_ecmult_gen(&ctx->ecmult_gen_ctx, &pubj, &key);


elichai:
I think this is a false positive, if `getrec` is 1 we pass a pointer to `recid` into `random_sign` and inside we write into it, and then we only call `CHECK` if `getrec` is 1, so it shouldn't call CHECK on an uninitialized `recid`

src/tests.c
```diff
@@ -4324,7 +4324,7 @@ void test_ecdsa_sign_verify(void) {
     secp256k1_scalar one;
     secp256k1_scalar msg, key;
     secp256k1_scalar sigr, sigs;
-    int recid;
+    int recid = 0;
```
real-or-random, line 4327: Can we set this init to `-1` instead to make sure that the `CHECK` below will fail if `recid` is never written to?

configure.ac
```diff
@@ -70,6 +70,9 @@ esac
 CFLAGS="-W $CFLAGS"
 
 warn_CFLAGS="-std=c89 -pedantic -Wall -Wextra -Wcast-align -Wnested-externs -Wshadow -Wstrict-prototypes -Wundef -Wno-unused-function -Wno-long-long -Wno-overlength-strings"
+if test x"$CC" = x"clang"; then
+    warn_CFLAGS="$warn_CFLAGS -Wconditional-uninitialized"
+fi
```
real-or-random, line 75: This is somewhat coarse: It won't match cases like `CC=clang-9` and it could match old clang versions which don't support the flag.

The proper way is to check support for the flag directly. Two possible approaches can be found in some of our PRs:

 * simple direct way: https://github.com/bitcoin-core/secp256k1/pull/833/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810 
 * using autoconf "libraries": see #744 

In this case, I'd prefer the simple direct way (but I may be biased here because it's my PR).

Alternatively, feel free to drop this commit and not add the warning at all.

real-or-random (commented):
So yes, this is a false positive but it's still a good idea to "fix" it. Ideally. the test should also check that `recid` is indeed written to by `secp256k1_ecdsa_sig_sign`.

PiRK:
Thanks for your reviews. I have updated the commits with your feedback.

real-or-random:
Thanks for the quick update. This looks good.

ACK 64ebb155a05f854df8010fe64f2b9d33d87d9a48 changes looks good but I haven't tested them

real-or-random (approved):

real-or-random (approved):

gmaxwell:
It would probably be a good idea to VG_UNDEF() that variable.

Unnecessarily initialization of variables is usually not the best practice. It causes actual harm because it undermines the sensitivity of undefined behaviour checking -- e.g. if there is some change to the code and the dummy initialization which isn't supposed to get used starts to get used, there isn't any way to detect it where otherwise with valgrind/ubsan it would get flagged.

Because this is in a test you have the option of VG_UNDEF(&recid,sizof()) it to restore the lost sensitivity at least for valgrind builds running in valgrind... and since this seems to allow turning on an additional warning it might be worth it.  OTOH, if the warning produces additional false positives in the future it'll probably have to get turned off.

It's probably worth testing this with multiple clang versions-- in the past compilers have sometimes introduced new warnings in forms that spewed false positives and then improved them.  If the codebase (with all the modules enabled) gets a lot of falses with an older but popular clang then it'll probably need to get compiler version restricted.

I tested this PR with ARM8 ubuntu clang 10 and all modules enabled and got no warnings.

src/tests.c
```diff
@@ -4324,7 +4324,7 @@ void test_ecdsa_sign_verify(void) {
     secp256k1_scalar one;
     secp256k1_scalar msg, key;
     secp256k1_scalar sigr, sigs;
-    int recid;
+    int recid = -1;
     int getrec;
     random_scalar_order_test(&msg);
```
real-or-random, line 4329: @gmaxwell Is this your suggestion?
```suggestion
    int getrec;
    /* Initialize recid to suppress a false positive -Wconditional-uninitialized in clang.
       VG_UNDEF ensures that valgrind will still treat the variable as uninitialized. */
    int recid = -1; VG_UNDEF(&recid, sizeof(recid));
    random_scalar_order_test(&msg);
```

real-or-random (commented):
Fwiw, I had checked that the current PR doesn't give further warnings on

```
clang version 11.0.1
Target: x86_64-pc-linux-gnu
```

real-or-random:
I wanted to write "let's ask Cirrus" about clang 9 but this PR didn't get a build. I'll try closing and opening this but I don't think it'll help.

real-or-random:
Oh I see the issue now: This PR is not rebased on the most recent master, so it just not have the new `.cirrus.yml`. 

@PiRK Can you rebase this on the current master? Sorry for all the fuss about this small change, we recently switched our CI from Travis to Cirrus and this hits us here.

PiRK:
Np, I will rebase it tomorrow. I'm on the road today.

src/tests.c
gmaxwell: yep!

PiRK: Thank you. I amended the first commit with your changes.

real-or-random:
Closing and reopening to trigger Cirrus CI, which now really uses clang (version 7).

real-or-random (approved):
ACK 99a1cfec1740a914aa416a87fd0acbde5426b969 code inspection

We now know that there are no additional warnings on clang 7 (Cirrus), clang 10 (@gmaxwell) and clang 11 (me). This should be enough.

So even there are two CI hiccups that could be solved by rebasing (the issue here is that .cirrus.yml is wrongly merged), these failures should disappear on master after merging. So I think we shouldn't bother @PiRK with a further rebase and just merge this.

jonasnick (approved):
No further warnings with clang version 11.1.1

ACK 99a1cfec1740a914aa416a87fd0acbde5426b969

PR #894, opened 2021-02-04
ctime_test: move context randomization test to the end

jonasnick:
I noticed this while reviewing https://github.com/ElementsProject/secp256k1-zkp/pull/117 and  finding some seemingly unnecessary `VALGRIND_MAKE_MEM_DEFINED` that I couldn't remove until I saw the bug.

### DIFF
--- a/src/valgrind_ctime_test.c
+++ b/src/valgrind_ctime_test.c
@@ -5,6 +5,8 @@
  ***********************************************************************/
 
 #include <valgrind/memcheck.h>
+#include <stdio.h>
+
 #include "include/secp256k1.h"
 #include "assumptions.h"
 #include "util.h"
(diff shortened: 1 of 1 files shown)


gmaxwell:
ah, indeed the comment said it needed to be last. :)  Did you look into why it wasn't causing failures with the schnorrsig test?

src/valgrind_ctime_test.c
```diff
@@ -25,16 +25,49 @@
...
+
+    /* Test context randomisation. Do this last because it leaves the context
+     * tainted. */
+    VALGRIND_MAKE_MEM_UNDEFINED(key, 32);
+    ret = secp256k1_context_randomize(ctx, key);
+    VALGRIND_MAKE_MEM_DEFINED(&ret, sizeof(ret));
+    CHECK(ret);
+
+    secp256k1_context_destroy(ctx);
+    return 1;
```
real-or-random, line 60: This must be left `0`. This is `main`. The issue here is valgrind exits with the exit code of the underlying process (unless there's a valgrind error, then we tell it to return 42). Successful runs will return 1 and that's why CI fails.

real-or-random (approved):
Concept ACK

nice catch and good idea to extract a function

src/valgrind_ctime_test.c
jonasnick: oops yeah that was unintentional

jonasnick:
> Did you look into why it wasn't causing failures with the schnorrsig test?

The reason is that in `ctime_test.c` we declassify public keys after creating them. In the case of the schnorrsig & extrakeys tests, they only operate on keypairs, and the public part of a keypair is declassified on `keypair_load` because it needs to branch on it.

src/valgrind_ctime_test.c
```diff
@@ -25,16 +25,49 @@
...
+
 int main(void) {
     secp256k1_context* ctx;
+    unsigned char key[32];
+    int ret, i;
+
+    if (!RUNNING_ON_VALGRIND) {
+        fprintf(stderr, "This test can only usefully be run inside valgrind.\n");
+        fprintf(stderr, "Usage: libtool --mode=execute valgrind ./valgrind_ctime_test\n");
+        exit(1);
```
real-or-random, line 38: Since you're touching this anyway, can you `#include <stdlib.h>` (or simply `return 1` here)?

real-or-random: Oh and we also need `stdio` for `fprintf`.

jonasnick: done

real-or-random (approved):
ACK 7d3497cdc4c747bdd51db70f42fe218622c3169f diff looks good

PR #899, opened 2021-02-26
WIP: Eliminate the prej array from ecmult_strauss_wnaf.

roconnor-blockstream:


roconnor-blockstream:
Address @sipa 's comments from #900.

Please review this PR first.

roconnor-blockstream:
@real-or-random observes that `STRAUSS_SCRATCH_OBJECTS` needs to be updated.

roconnor-blockstream:
Rebased and updated `STRAUSS_SCRATCH_OBJECTS`.

roconnor-blockstream:
rebased.

src/ecmult_impl.h
```diff
@@ -266,27 +281,29 @@ static void secp256k1_ecmult_strauss_wnaf(const struct secp256k1_strauss_state *
...
+        secp256k1_ge_globalz_fixup_table(ECMULT_TABLE_SIZE(WINDOW_A) * no, state->pre_a, state->aux);
     } else {
         secp256k1_fe_set_int(&Z, 1);
     }
 
     for (np = 0; np < no; ++np) {
         for (i = 0; i < ECMULT_TABLE_SIZE(WINDOW_A); i++) {
-            secp256k1_ge_mul_lambda(&state->pre_a_lam[np * ECMULT_TABLE_SIZE(WINDOW_A) + i], &state->pre_a[np * ECMULT_TABLE_SIZE(WINDOW_A) + i]);
+            secp256k1_ge pre_a_lam;
+            secp256k1_ge_mul_lambda(&pre_a_lam, &state->pre_a[np * ECMULT_TABLE_SIZE(WINDOW_A) + i]);
```
robot-dreams, line 298: Style nit (feel free to ignore): I'm on the fence about this, but do you think it's worth expressing this as just a field multiplication by `beta`?

```c
secp256k1_fe_mul(
    &state->aux[np * ECMULT_TABLE_SIZE(WINDOW_A) + i],
    &state->pre_a[np * ECMULT_TABLE_SIZE(WINDOW_A) + i],
    beta)
```

src/ecmult_impl.h
```diff
@@ -79,21 +81,19 @@ static void secp256k1_ecmult_odd_multiples_table(int n, secp256k1_gej *prej, sec
...
+        pre_a[i].x = a->x;
+        pre_a[i].y = a->y;
+        pre_a[i].infinity = 0;
     }
 
-    /*
-     * Each point in 'prej' has a z coordinate too small by a factor of 'd.z'. Only
-     * the final point's z coordinate is actually used though, so just update that.
-     */
-    secp256k1_fe_mul(&prej[n-1].z, &prej[n-1].z, &d.z);
```
robot-dreams, line 96: Is it worth keeping at least some version of this comment? For example:

```c
    /*
     * Multiply by d.z to undo the isomorphism that makes 'd' affine.
     */
```

src/group.h
```diff
@@ -64,12 +64,13 @@ static void secp256k1_ge_set_gej_var(secp256k1_ge *r, secp256k1_gej *a);
...
  *  the same global z "denominator". zr must contain the known z-ratios such
- *  that mul(a[i].z, zr[i+1]) == a[i+1].z. zr[0] is ignored. The x and y
- *  coordinates of the result are stored in r, the common z coordinate is
- *  stored in globalz. */
-static void secp256k1_ge_globalz_set_table_gej(size_t len, secp256k1_ge *r, secp256k1_fe *globalz, const secp256k1_gej *a, const secp256k1_fe *zr);
+ *  that mul(a[i].z, zr[i+1]) == a[i+1].z. zr[0] is ignored.
+ *  The common z coordinate is the (implicit) z coordinate of a[len-2], and therefore
+ *  the coordinates of a[len-1] are not changed.
+ */
+static void secp256k1_ge_globalz_fixup_table(size_t len, secp256k1_ge *a, const secp256k1_fe *zr);
```
robot-dreams, line 73: Is it also worth mentioning that `a[len - 1]` implicitly has `z = 1` (which is why it's enough for this function to accept a `_ge` rather than `_gej`, without even taking in a `Z` value)?

robot-dreams (commented):
Concept ACK, `secp256k1_ecmult_strauss_wnaf` indeed uses scratch space more efficiently with this change.

I ran `./bench_ecmult strauss_wnaf` at both 1cd7ab2c81b7a7089709b67638a596b8337d0b41 and 9a5a87e0f1276e0284446af1172056ea4693737f just to make sure there's no obvious regression in running time.

Unfortunately it looks like this might have some merge conflicts with #638 depending on which gets merged first.

src/ecmult_impl.h
```diff
@@ -56,12 +56,14 @@
...
  *  the values [1*a,3*a,...,(2*n-1)*a], so it space for n values. zr[0] will
- *  contain prej[0].z / a.z. The other zr[i] values = prej[i].z / prej[i-1].z.
- *  Prej's Z values are undefined, except for the last value.
+ *  contain pre_a[0].z / a.z. The other zr[i] values = pre_a[i].z / pre_a[i-1].z.
+ *  The a value will end up equal to pre_a[(2*n-1)].
+ *  The output pre_a array will have their z-coordinate omitted.
+ *  The omitted z-coordinate are implied by the final a value's z-coordinate and the zr array.
  */
-static void secp256k1_ecmult_odd_multiples_table(int n, secp256k1_gej *prej, secp256k1_fe *zr, const secp256k1_gej *a) {
+static void secp256k1_ecmult_odd_multiples_table(int n, secp256k1_ge *pre_a, secp256k1_fe *zr, secp256k1_gej *a) {
```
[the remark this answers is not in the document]
roconnor-blockstream: I have learned that returning structures is frowned upon in C.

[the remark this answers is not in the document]
robot-dreams: Oh yeah, I agree with you about that. What I was suggesting is to have an extra input parameter that's a pointer like `secp256k1_fe *global_z` (or maybe `z_final`).

src/ecmult_impl.h
roconnor-blockstream: I think it is  a good idea.  The major problem is that beta is not available as a global constant, but probably it should be.

src/group.h
roconnor-blockstream: This isn't correct.  The implicit z value is the "global Z" value returned in `secp256k1_ecmult_odd_multiples_table`.

robot-dreams: I'm probably using "implicit" incorrectly, I was taking into account this comment:

`Each point in 'prej' has a z coordinate too small by a factor of 'd.z'.`

https://github.com/bitcoin-core/secp256k1/blob/master/src/ecmult_impl.h#L93

roconnor-blockstream: The way I think of it is that a is an array of points with unknown z-coordinates, but known ratios of these z-coordinates.  But that is good enough to rescale all the points so that they end up all sharing the same z-coordinate of the last entry, whatever it is.

src/group.h
```diff
@@ -64,12 +64,13 @@ static void secp256k1_ge_set_gej_var(secp256k1_ge *r, secp256k1_gej *a);
...
 
-/** Bring a batch inputs given in jacobian coordinates (with known z-ratios) to
+/** Bring a batch inputs given in coordinates (with known z-ratios) to
  *  the same global z "denominator". zr must contain the known z-ratios such
- *  that mul(a[i].z, zr[i+1]) == a[i+1].z. zr[0] is ignored. The x and y
- *  coordinates of the result are stored in r, the common z coordinate is
- *  stored in globalz. */
-static void secp256k1_ge_globalz_set_table_gej(size_t len, secp256k1_ge *r, secp256k1_fe *globalz, const secp256k1_gej *a, const secp256k1_fe *zr);
+ *  that mul(a[i].z, zr[i+1]) == a[i+1].z. zr[0] is ignored.
+ *  The common z coordinate is the (implicit) z coordinate of a[len-2], and therefore
```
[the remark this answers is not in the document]
roconnor-blockstream: Fixed.

src/ecmult_impl.h
roconnor-blockstream: Done.

src/ecmult_impl.h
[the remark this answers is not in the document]
roconnor-blockstream: Done.  I'm a little tempted to take z as an input and rescale the a value by its value, but it is probably best to leave it as is.

src/ecmult_impl.h
roconnor-blockstream: Done.

src/field_impl.h
```diff
@@ -136,5 +136,9 @@ static int secp256k1_fe_sqrt(secp256k1_fe *r, const secp256k1_fe *a) {
 }
 
 static const secp256k1_fe secp256k1_fe_one = SECP256K1_FE_CONST(0, 0, 0, 0, 0, 0, 0, 1);
+static const secp256k1_fe secp256k1_const_beta = SECP256K1_FE_CONST(
+        0x7ae96a2bul, 0x657c0710ul, 0x6e64479eul, 0xac3434e9ul,
```
robot-dreams, line 140: Nit: deindent?

robot-dreams (commented):
OK, I took another look and I think I have a better understanding now. This PR looks pretty solid.

I left a few suggestions for comments / style, mostly because I thought some parts were confusing even before your PR. If you prefer I can also push them to some branch as a follow-up commit.

src/group_impl.h
```diff
@@ -160,27 +160,26 @@ static void secp256k1_ge_set_all_gej_var(secp256k1_ge *r, const secp256k1_gej *a
...
 
     if (len > 0) {
-        /* The z of the final point gives us the "global Z" for the table. */
-        r[i].x = a[i].x;
-        r[i].y = a[i].y;
         /* Ensure all y values are in weak normal form for fast negation of points */
-        secp256k1_fe_normalize_weak(&r[i].y);
-        *globalz = a[i].z;
-        r[i].infinity = 0;
+        secp256k1_fe_normalize_weak(&a[i].y);
```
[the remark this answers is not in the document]
peterdettman: The other elements will each get their y value scaled by multiplication in secp256k1_ge_set_gej_zinv, which will already give a magnitude of 1. Although this is somewhat implicit knowledge, the main consumer is the ECMULT_TABLE_GET_GE macro where the _fe_negate call requires the magnitude to be 1.

Unfortunately that's only IF a point is ever needed negated, so I think we could do with a static VERIFY check somewhere. https://github.com/bitcoin-core/secp256k1/pull/1032 should help lead in that direction.

roconnor-blockstream:
Thanks @robot-dreams.  I've taken your suggestions and made some minor adjustments to them.

src/field_impl.h
roconnor-blockstream: Done.

src/group.h
```diff
@@ -9,7 +9,10 @@
 
 #include "field.h"
 
-/** A group element of the secp256k1 curve, in affine coordinates. */
+/** A group element in affine coordinates on the secp256k1 curve,
+ *  or occasionally on an isomorphic curve of the from y^2 = x^3 + 7*t^6.
```
robot-dreams, line 13: Nit: from -> form

src/group.h
```diff
@@ -9,7 +9,10 @@
...
+ *  Note: For exhastive test mode, sepc256k1 is replaced by a small subgroup of a different curve.
```
robot-dreams, line 14: Nit: exhastive -> exhaustive, sepc256k1 -> secp256k1

robot-dreams:
ACK 5aea2e854e3db0d60416b3118eb5949b315103ef (from inspecting the range diff against ae332324f16f908ddf3b340274816c8edb26738a)

robot-dreams (commented):
ACK 5aea2e854e3db0d60416b3118eb5949b315103ef (from inspecting the range diff against ae332324f16f908ddf3b340274816c8edb26738a)

src/group.h
roconnor-blockstream: Done.

src/group.h
roconnor-blockstream: Done.

robot-dreams:
ACK b6ed302caf39e95b960ac4383654a710ab2d8737

src/ecmult_impl.h
```diff
@@ -56,44 +56,51 @@
...
     /*
-     * Perform the additions on an isomorphism where 'd' is affine: drop the z coordinate
-     * of 'd', and scale the 1P starting value's x/y coordinates without changing its z.
+     * Perform the additions using an isomorphism that divides the z coordinate by the
+     * constant d.z. The group law is the same in the image, and since 'd' maps to an
+     * affine point, addition becomes more efficient (using mixed coordinates).
+     *
+     *     phi(x, y, z) = (x, y, z/d.z)
+     *           phi(d) = (d.x, d.y, 1)
+     *           phi(a) = (a.x, a.y, a.z/d.z) = (a.x*d.z^2, a.y*d.z^3, a.z)
```
jonasnick, line 89:
```suggestion
     *     phi(x, y, z) = (x, y, z/d.z)
     *   d_ge := phi(d) = (d.x, d.y, 1)
```

src/ecmult_impl.h
```diff
@@ -56,44 +56,51 @@
...
-
-    secp256k1_ge_set_gej_zinv(&a_ge, a, &d.z);
-    prej[0].x = a_ge.x;
-    prej[0].y = a_ge.y;
-    prej[0].z = a->z;
-    prej[0].infinity = 0;
+    secp256k1_ge_set_xy(&d_ge, &d.x, &d.y);
+    secp256k1_ge_set_gej_zinv(&pre_a[0], a, &d.z);
+    secp256k1_gej_set_ge(&ai, &pre_a[0]);
+    ai.z = a->z;
```
jonasnick, line 94:
```suggestion
    secp256k1_ge_set_xy(&d_ge, &d.x, &d.y);
    /* Set pre_a[0] to the point
     *     (a.x*d.z^2, a.y*d.z^3, 1) = (a.x, a.y, 1/d.z).
     * As a result, the omitted z-coordinate z(pre_a[0]) is equal to a.z*d.z. */
    secp256k1_ge_set_gej_zinv(&pre_a[0], a, &d.z);
    /* ai := phi(a) = (a.x, a.y, a.z/d.z) = (a.x*d.z^2, a.y*d.z^3, a.z) */
    secp256k1_gej_set_ge(&ai, &pre_a[0]);
    ai.z = a->z;
```

jonasnick, line 94: micronit, but imho _slightly_ more helpful

jonasnick (commented):
Good stuff and despite being an optimization, this PR simplifies a lot of things as well. Also thanks for the new comments. They're really helpful.

ACK mod nits

src/ecmult_impl.h
```diff
@@ -106,13 +113,11 @@ static void secp256k1_ecmult_odd_multiples_table(int n, secp256k1_gej *prej, sec
...
 
     /* Compute the odd multiples in Jacobian form. */
-    secp256k1_ecmult_odd_multiples_table(ECMULT_TABLE_SIZE(WINDOW_A), prej, zr, a);
-    /* Bring them to the same Z denominator. */
-    secp256k1_ge_globalz_set_table_gej(ECMULT_TABLE_SIZE(WINDOW_A), pre, globalz, prej, zr);
+    secp256k1_ecmult_odd_multiples_table(ECMULT_TABLE_SIZE(WINDOW_A), pre, zr, globalz, a);
+    secp256k1_ge_globalz_fixup_table(ECMULT_TABLE_SIZE(WINDOW_A), pre, zr);
 }
 
 /** The following two macro retrieves a particular odd multiple from a table
```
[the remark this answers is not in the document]
real-or-random: Or just turn the macros into functions... I'm not aware of a real reason to use these procedural macros in modern C.*



* Unless you use macros to express things not expressible with a function, e.g., making up identifiers on the fly via string concatenation.

roconnor-blockstream:
Rebased, but I still need to address the outstanding comments.

src/ecmult_impl.h
```diff
@@ -56,44 +56,51 @@
...
     VERIFY_CHECK(!a->infinity);
 
     secp256k1_gej_double_var(&d, a, NULL);
 
     /*
-     * Perform the additions on an isomorphism where 'd' is affine: drop the z coordinate
-     * of 'd', and scale the 1P starting value's x/y coordinates without changing its z.
+     * Perform the additions using an isomorphism that divides the z coordinate by the
+     * constant d.z. The group law is the same in the image, and since 'd' maps to an
+     * affine point, addition becomes more efficient (using mixed coordinates).
```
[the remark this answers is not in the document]
roconnor-blockstream: I'm having trouble following your suggested wording.

We are working on the isomorphic curve y^2 = x^3 + 7*C^6 where C=d.z with our isomorphism sending secp256k1 to this curve is defined by phi(x,y) = (x*C^2, y*C^3).  The transformation of some generic point in Jacobian coordinates a := (ax, ay, az) is phi(a) = (ax*C^2,ay*C^3,az) or equivalently phi(a) = (ax,ay,az/C).  In particular a set of Jacobian coordinates for phi(d) is (x,y,1) which means the affine coordinates of phi(d) is (x,y).  The inverse isomorphism can be applied to a point in Jacobian coordinates by multiplying the z coordinate by C.

We have algorithms for adding two generic points in Jacobian coordinates, and a slightly optimized algorithm for adding a point in Jacobian coordinates with a point in affine coordinates.  The optimization amounts to noting that the z value of the point in affine coordinates is 1 and then removing all the places where you would otherwise multiply by 1.  Note that we are not adding two affine points, and thus field inversions are not under consideration, nor were they ever under consideration.

That all said, maybe I can try to expand upon this comment to make it more clear.  Doing operations on an isomorphic curve is definitely hard to follow.

src/ecmult_impl.h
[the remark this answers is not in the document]
roconnor-blockstream: I believe we got here as follows.  The macro was introduce in https://github.com/bitcoin-core/secp256k1/commit/b1483f874c6111b7d727d6c6cb3df35c16a880bc#diff-1f71a59e139573dfb0695be308c69455971378cd67ad7946a3703cb725b589fdR51 (expand the diff for ecmult.c), where there was a function name argument being passed in.  I suppose in that situation a macro makes sense to avoid the use of function pointers?

But this function argument was eliminated in https://github.com/bitcoin-core/secp256k1/commit/41f84554341c3c2f21324da53982b8da045f88e3#diff-dc206b3f5f973b97a0f5c1b12d705d0e58a1cb8c9684e481e7a34fc05e4716aaL63 at which point there is no longer a reason for this to continue to be a macro.

Thus it should be safe for me to convert it into an inline function.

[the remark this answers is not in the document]
roconnor-blockstream: Should I move the macros into function and #define ECMULT_TABLE_VERIFY?

roconnor-blockstream:
I've reworded some comments and made some of the suggested changes.

src/ecmult_impl.h
roconnor-blockstream: done.

src/group.h
```diff
@@ -64,12 +69,24 @@ static void secp256k1_ge_set_gej_var(secp256k1_ge *r, secp256k1_gej *a);
...
+ *  The zr[0] value is unused.
+ *
+ *  This function adjusts the coordinates of 'a' in place so that for all 'i', z(a[i]) == z(a[len-1]).
+ *  In other words, the initial value of z(a[len-1]) becomes the global z "denominator". Only the
+ *  a[i].x and a[i].y coordinates are explicitly modified; the adjustment of the omitted z coordinate is
+ *  implicit.
+ *
+ *  The coordinates of the final element a[len-1] are not changed.
+ */
+static void secp256k1_ge_globalz_fixup_table(size_t len, secp256k1_ge *a, const secp256k1_fe *zr);
```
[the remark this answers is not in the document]
roconnor-blockstream: done.

src/group_impl.h
```diff
@@ -7,7 +7,7 @@
 #ifndef SECP256K1_GROUP_IMPL_H
 #define SECP256K1_GROUP_IMPL_H
 
-#include "field.h"
+#include "field_impl.h"
```
roconnor-blockstream, line 10: So I had to make this change to get `precompute_ecmult_gen.c` to build.

I'm not sure what the rules on including *_impl files is.  Should `secp256k1_const_beta` (and `secp256k1_fe_one`) be moved from `field_impl.h` to `field.h` instead ?

real-or-random: Well, at the moment there are no written rules. I believe the rules should look like described here: https://github.com/bitcoin-core/secp256k1/issues/1039#issuecomment-992692922 but we don't adhere to them everywhere. 

And yes, moving the constants to `field.h` would take us a step closer to this. So if this works, feel free to do it here but otherwise let's stick with including the `_impl.h` file until we have a clean solution to #1039.

src/ecmult_impl.h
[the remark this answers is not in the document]
jonasnick: Sorry for the confusion. Your description of the isomorphism is correct. I am looking for a concise explanation of the optimization which is that we can use the faster `gej_add_ge` instead of `gej_add`. Surely we could use the former algorithm without the transformation by doing a field inversion to convert `d` into a regular `ge`. Perhaps this point is only confusing and not noteworthy.

src/ecmult_impl.h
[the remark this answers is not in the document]
jonasnick: That would be ideal I think, but turning the macros to functions isn't necessary for this PR. I just suggested isolating `ECMULT_TABLE_VERIFY` because it simplifies review of this PR.

src/ecmult_impl.h
```diff
@@ -56,71 +56,97 @@
...
-    secp256k1_ge_set_gej_zinv(&a_ge, a, &d.z);
-    prej[0].x = a_ge.x;
-    prej[0].y = a_ge.y;
-    prej[0].z = a->z;
-    prej[0].infinity = 0;
+    secp256k1_ge_set_xy(&d_ge, &d.x, &d.y);
+    /* Set pre_a[0] to the point
+     *     (a.x*C^2, a.y*C^3, 1) = (a.x, a.y, 1/C).
+     * As a result, the omitted z-coordinate z(pre_a[0]) is equal to a.z*C.
+      */
```
jonasnick, line 100: nit: not aligned

src/ecmult_impl.h
```diff
@@ -56,71 +56,97 @@
...
+    secp256k1_ge_set_gej_zinv(&pre_a[0], a, &d.z);
+    /* ai := phi(a) = (a.x, a.y, a.z/C) = (a.x*C^2, a.y*C^3, a.z) */
```
jonasnick, line 102: nit: definition of phi(a) already mentioned above. Perhaps just `ai := phi(a)`.

jonasnick (commented):
ACK mod nits

src/ecmult_impl.h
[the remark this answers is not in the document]
jonasnick: Your new explanation is very good.

src/ecmult_impl.h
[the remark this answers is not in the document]
jonasnick: Looks like you did both, thanks.

src/group_impl.h
roconnor-blockstream: constants are moved.

roconnor-blockstream:
I've rearranged the code a little bit and rewritten the comments a bit.

jonasnick (commented):
ACK b797a500ec194948eecbea8bd80f6b7d455f7ca2

sipa:
ACK b797a500ec194948eecbea8bd80f6b7d455f7ca2

PR #900, opened 2021-03-01
Eliminate the use of points and scalars from secp256k1_ecmult_strauss_batch.

roconnor-blockstream:
We have secp256k1_ecmult_strauss_wnaf invoke the callback itself.

This builds upon PR #899.

### DIFF
--- a/src/ecmult_impl.h
+++ b/src/ecmult_impl.h
@@ -233,7 +233,9 @@ struct secp256k1_strauss_state {
     struct secp256k1_strauss_point_state* ps;
 };
 
-static void secp256k1_ecmult_strauss_wnaf(const struct secp256k1_strauss_state *state, secp256k1_gej *r, size_t num, const secp256k1_gej *a, const secp256k1_scalar *na, const secp256k1_scalar *ng) {
+typedef int (secp256k1_ecmult_strauss_multi_callback)(secp256k1_scalar *sc, secp256k1_gej *pt, size_t idx, void *data);
+
+static int secp256k1_ecmult_strauss_wnaf(const struct secp256k1_strauss_state *state, secp256k1_gej *r, size_t num, secp256k1_ecmult_strauss_multi_callback cb, void *cbdata, size_t cb_offset, const secp256k1_scalar *ng) {
     secp256k1_ge tmpa;
     secp256k1_fe Z;
     /* Split G factors. */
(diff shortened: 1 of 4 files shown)


src/ecmult_impl.h
```diff
@@ -82,7 +82,7 @@
  *  contain prej[0].z / a.z. The other zr[i] values = prej[i].z / prej[i-1].z.
```
sipa: In commit c531ff8fb96cdaa871e637631a346bb1352cca3d:

The documentation above secp256k1_ecmult_odd_multiples_table is now outdated (it is only fixed in the next commit, I believe).

src/group.h
```diff
@@ -71,12 +71,11 @@ static void secp256k1_ge_set_gej_var(secp256k1_ge *r, secp256k1_gej *a);
...
 static void secp256k1_ge_set_all_gej_var(secp256k1_ge *r, const secp256k1_gej *a, size_t len);
 
-/** Bring a batch inputs given in jacobian coordinates (with known z-ratios) to
+/** Bring a batch inputs given in coordinates (with known z-ratios) to
  *  the same global z "denominator". zr must contain the known z-ratios such
- *  that mul(a[i].z, zr[i+1]) == a[i+1].z. zr[0] is ignored. The x and y
- *  coordinates of the result are stored in r, the common z coordinate is
- *  stored in globalz. */
-static void secp256k1_ge_globalz_set_table_gej(size_t len, secp256k1_ge *r, secp256k1_fe *globalz, const secp256k1_gej *a, const secp256k1_fe *zr);
+ *  that mul(a[i].z, zr[i+1]) == a[i+1].z. zr[0] is ignored.
```
sipa, line 76: In c531ff8fb96cdaa871e637631a346bb1352cca3d:

perhaps document which global z everything is set to?

src/ecmult_impl.h
roconnor-blockstream: It's probably because I don't know how to GitHub, but I intended for these comments to end up on #899 instead.

src/group.h
roconnor-blockstream: Done.

src/ecmult_impl.h
roconnor-blockstream: Done.

robot-dreams (commented):
I only reviewed the last 2 commits, since the rest are from #899 .

I think the callback/adaptor is a nice idea for optimizing space usage, but my main concern is about possibly breaking the "public" API (see inline comment).

A quick check of `./bench_ecmult strauss_wnaf` at three different versions shows that there are no immediately obvious performance regressions:

Plot

roconnor-blockstream:
I'm going to close this PR.  I don't feel like pursing it; I'm not really convinced it is worthwhile.  If, for whatever reason, someone disagree and does think it is worthwhile, feel free to take it over.

PR #901, opened 2021-03-02
ci: Switch all Linux builds to Debian and more improvements

real-or-random:
Best reviewed commit by commit


.cirrus.yml
```diff
@@ -101,14 +113,15 @@ task:
     - env:
         CC: gcc
```
jonasnick, line 114: Hm, your new `file` invocations shows that this actually builds 64-bit binaries https://cirrus-ci.com/task/5529748383203328?command=test#L212. Not sure if this is missing the `-m32` flag or if the automake `--host` option should have set that.

jonasnick (commented):
658faf66f757105280b021d0d0aba697bfce4f73 looks good to me mod comment

ci/cirrus.sh
```diff
@@ -25,14 +21,19 @@ sed -i'' -e 's@/usr/bin/file@$(which file)@g' configure
...
+make -j2
+
+# Print information about binaries so that we can see that the architecture is correct
+file ./*tests || true
+file ./bench_* || true
+
 if [ -n "$BUILD" ]
 then
-    make -j2 "$BUILD"
+    make "$BUILD"
```
jonasnick, line 32: Why remove `-j2`?

.cirrus.yml
real-or-random: Hm damn, I played a lot with this, and in the end I thought it works without the `-m32` flag. I'll have a look... 

By the way, a difference to the previous approach on Debian (what we had on Travis) is that this here doesn't use gcc-multilib because this is not compatible with installing the other s390x compiler (okay we could have two images) and I generally considered it to be more consistent: "multilib" is a thing we anyway can do only on x86(_64) and not for s390x (and arm and more archs, which we hopefully can add now by basically copy and paste).

ci/cirrus.sh
real-or-random: Oh yeah, so the call to `make` doing the actual compilation now is a few linesabove.

But yeah, `$BUILD` is usually `check` here and this can run in parallel too: https://www.gnu.org/software/automake/manual/html_node/Parallel-Test-Harness.html So we should probably do this. In fact, we should have maybe have -j5 or set it cleverly because on Linux we ask for 1 CPU but on MacOS we always get 4 CPUs.

.cirrus.yml
real-or-random: Ok, the issue was indeed that we set `CC` explicitly to `gcc`. When it's not set, configure finds the right `i686-linux-gn-gcc`. I've set it explicitly to `i686-linux-gnu-gcc` now.

ci/cirrus.sh
real-or-random: I changed the existing commit not to remove -j2 and I added a commit that improves parallelism on macOS. This can be improved further when we overhaul the CI script.

.cirrus.yml
jonasnick: Yup, thanks that appears to work.

.cirrus.yml
```diff
@@ -16,26 +16,38 @@ env:
   CTIMETEST: yes
   BENCH: yes
   ITERS: 2
-  # We only need the top commit
-  CIRRUS_CLONE_DEPTH: 1
+  MAKEFLAGS: -j2
```
jonasnick, line 19: What does this do?

real-or-random: `MAKEFLAGS` is a way to set flags for `make` globally. `make` will pick this up. See in 5.7.3 ian the manual (https://www.gnu.org/software/make/manual/make.html#index-MAKEFLAGS) or in Appendix A:

> The MAKEFLAGS variable can also be useful if you want to have certain options, such as ‘-k’ (see Summary of Options), set each time you run make. You simply put a value for MAKEFLAGS in your environment.

real-or-random: ugh but now I need to remove `-j2`  from the invocations in the script... Sorry, I screwed up with too many fixups  and squashes

jonasnick: Ah appendix A is helpful. The first sentence in the manual is "Flags [...] are passed automatically to the sub-make through the variable MAKEFLAGS." which would make this use of MAKEFLAGS rather esoteric.

real-or-random: Yeah, I admit the docs are somewhat obscure. (Maybe this was introduced as a way to pass stuff to sub-makes but it's widely used to specify -j2 for example, and there's nothing wrong with that.) See also POSIX: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/make.html

I'll update once again to remove the explicit flags from the command line.

real-or-random: fixed

jonasnick (approved):
ACK 9361f360bb04156c7a0fa8f2664680b74d463ed5

ci/cirrus.sh
```diff
@@ -21,22 +21,27 @@ valgrind --version || true
     --with-valgrind="$WITH_VALGRIND" \
     --host="$HOST" $EXTRAFLAGS
 
+make -j2
+
+# Print information about binaries so that we can see that the architecture is correct
+file *tests || true
```
sipa, line 28: Why the "|| true" ?

sipa (approved):
utACK 9361f360bb04156c7a0fa8f2664680b74d463ed5

ci/cirrus.sh
real-or-random: Otherwise the statement will fail if the wildcard does not match anything, and the files may not exist in every job. This is a little bit more meaningful in the line below when checking for benchmarks (because we have jobs with benchmarks disabled). But I just thought it's more consistent to apply the `|| true` in every line. In the future we may have special jobs without tests enabled (cttimetests only etc.)

PR #902, opened 2021-03-12
Add VERIFY_ONLY_CHECK that only evaluates in VERIFY mode

sipa:
The `VERIFY_CHECK` macro still evaluates its argument in non-VERIFY mode, to make sure that in case it has any side effects (intentionally or unintentionally) that are relevant to the code, these also apply in production builds.

This isn't always desirable. If nontrivial functions are invoked inside of them the compiler may be unable to optimize the call out, resulting in performance degradation or even the introduction of non-constant-time behavior where that is unexpected.

This adds a new `VERIFY_ONLY_CHECK` macro that does not evaluate its argument (it is still compiled inside an `if (0)` though to guarantee syntactic correctness), for this purpose, and changes some nontrivial calls to use it.

### DIFF
--- a/src/ecmult_const_impl.h
+++ b/src/ecmult_const_impl.h
@@ -129,7 +129,7 @@ static int secp256k1_wnaf_const(int *wnaf, const secp256k1_scalar *scalar, int w
     } while (word * w < size);
     wnaf[word] = u * global_sign;
 
-    VERIFY_CHECK(secp256k1_scalar_is_zero(&s));
+    VERIFY_ONLY_CHECK(secp256k1_scalar_is_zero(&s));
     VERIFY_CHECK(word == WNAF_SIZE_BITS(size, w));
     return skew;
 }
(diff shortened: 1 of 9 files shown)


real-or-random:
What do you think about this?

```
extern int not_supposed_to_survive;
#if defined(COVERAGE)
...
#elif defined(VERIFY)
#define VERIFY_CHECK CHECK
#define VERIFY_SETUP(stmt) do { stmt; } while(0)
#else
#define VERIFY_CHECK(cond) do { (void)(not_supposed_to_survive || (cond)); } while(0)
#define VERIFY_SETUP(stmt)
#endif
```

This a rather hacky but working way to let the compiler error out if it fails to prove that `cond` has no side-effects. (Taken from https://stackoverflow.com/a/35294344.) 

This compiles fine with on current master on clang and gcc with `-O1` and higher. Compilations fails when I add a dubious `VERIFY_CHECK` with a side-effect.

sipa:
@real-or-random Ha, nice!

But I think we'll still want a macro like this for cases where it can't be optimized out? Or do you think all cases now are actually removed?

We'd also need some define to explicitly disable VERIFY_CHECK for -O0 builds.

real-or-random:
> @real-or-random Ha, nice!
> 
> But I think we'll still want a macro like this for cases where it can't be optimized out? Or do you think all cases now are actually removed?

I think the fact that master compiles fine with this implies all cases are optimized everywhere. 


> We'd also need some define to explicitly disable VERIFY_CHECK for -O0 builds.

Indeed

sipa:
@real-or-random See #904.

sipa:
Closing in favor of #904.

PR #903, opened 2021-03-12
Make argument of fe_normalizes_to_zero{_var} const

sipa:
These functions do not modify their argument.

### DIFF
--- a/src/field.h
+++ b/src/field.h
@@ -43,13 +43,12 @@ static void secp256k1_fe_normalize_weak(secp256k1_fe *r);
 /** Normalize a field element, without constant-time guarantee. */
 static void secp256k1_fe_normalize_var(secp256k1_fe *r);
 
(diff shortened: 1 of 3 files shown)


apoelstra (approved):
ack 147494ce9975fea449381ec29af2e300dad2e7e2

peterdettman:
As the method comments say, it can optionally normalize the value. I believe this had something to do with one of the field implementations at the time (openssl, gmp??) not really having a good option with a const argument. Don't recall. Presumably it is a redundant consideration now.

sipa:
@peterdettman Ha, good point. At least the comments should be updated.

sipa:
Updated the function descriptions.

real-or-random (approved):
ACK 23c3fb629b905deebc4bcc9914bcfff7b9aedacd diff looks good

jonasnick (commented):
ACK 23c3fb629b905deebc4bcc9914bcfff7b9aedacd

PR #904, opened 2021-03-12
Automatically check that VERIFY_CHECKs are side effect free

sipa:
At configure time, detect whether our compiler is able to optimize out references to variables that only control execution of a side-effect free statement. If so, use that trick at compile time to verify that all our VERIFY_CHECK statements are in fact (provably) side-effect free.

This trick is explained on stackoverflow.com/a/35294344, and was suggested on https://github.com/bitcoin-core/secp256k1/pull/902#issuecomment-797736831.

### DIFF
--- a/build-aux/m4/bitcoin_secp.m4
+++ b/build-aux/m4/bitcoin_secp.m4
@@ -9,6 +9,20 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
 AC_MSG_RESULT([$has_64bit_asm])
 ])
 
+AC_DEFUN([SECP_CHECK_SIDE_EFFECT_FREE],[
+AC_MSG_CHECKING(whether compiler detects side effect free statements)
+AC_LINK_IFELSE([AC_LANG_PROGRAM([[
+  extern int should_not_survive;
+  static int side_effect_free(void) {
+      int x=3;
+      return x*x == 15;
+  }
+]],[[
+  (void)(should_not_survive || side_effect_free());
+  ]])],[has_check_side_effect_free=yes],[has_check_side_effect_free=no])
+AC_MSG_RESULT([$has_check_side_effect_free])
+])
+
 dnl
 AC_DEFUN([SECP_OPENSSL_CHECK],[
   has_libcrypto=no
--- a/configure.ac
+++ b/configure.ac
@@ -218,6 +218,13 @@ else
     CFLAGS="-O2 $CFLAGS"
 fi
 
+SECP_CHECK_SIDE_EFFECT_FREE
+if test x"$have_check_side_effect_free" = x"yes"; then
+  set_check_side_effect_free=yes
+else
+  set_check_side_effect_free=no
+fi
+
 if test x"$req_asm" = x"auto"; then
   SECP_64BIT_ASM_CHECK
   if test x"$has_64bit_asm" = x"yes"; then
@@ -292,6 +299,16 @@ if test x"$use_external_asm" = x"yes"; then
   AC_DEFINE(USE_EXTERNAL_ASM, 1, [Define this symbol if an external (non-inline) assembly implementation is used])
 fi
 
+case $set_check_side_effect_free in
+yes)
+  AC_DEFINE(CHECK_SIDE_EFFECT_FREE, 1, [Define this symbol to enable checks for side-effect free statements])
+  ;;
+no)
+  ;;
+*)
+  AC_MSG_ERROR([invalid selection for check_side_effect_free])
+  ;;
+esac
 
 # Select wide multiplication implementation
 case $set_widemul in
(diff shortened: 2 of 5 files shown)


apoelstra (approved):
ack 07d1a6e189cad3a411424b52e9ba333d33585596

Nice trick using autotools to detect whether the shouldn't_survive variable would actually survive or not!

real-or-random:
I like this approach because it's safe and we let the compiler prove things. I have a few minor suggestions (will post these later).

I'm somewhat concerned that there may be compiler (flags) in the wild that make the autoconf check pass but still fail on the real examples in our code base. It's somewhat hard to predict. Of course this wouldn't be terrible. We could try and see if we get those reports.

Maybe a more conservative approach would be to restrict it to gcc and clang, or to have a `--enable`/`--with` flag to force it off. Or approach it as "compiler-as-a-static-verifier" and enable this only in a special test build (that we also add to CI).

sipa:
A few changes:
* Added a configure flag to enable the checks (can be set to yes, no, or auto).
* By default it is off, to prevent it from biting people unexpectedly that try to compile on slightly less than platforms where the check may fail.
* Made the autodetection test code stricter, as it was causing issues at -O1 (where the autodetection succeeded, but compilation would fail). With the new check, will not detect the compiler as sufficiently powerful at -O1.
* Added now-needed CI configuration to enable it in most cases.

real-or-random:
I still seems to fail with ubsan. I guess ubsan disables some optimizations (there could be UB in the eliminated code).

Is there a specific reason to disable this on clang?

sipa:
> I still seems to fail with ubsan. I guess ubsan disables some optimizations (there could be UB in the eliminated code).

Pushed a change to make it 'auto' for those (which for the time being probably means 'no').

> Is there a specific reason to disable this on clang?

Yes, CI failed. Apparently clang's optimizer isn't powerful enough for the auto-detect test I have now.

real-or-random:
(I reported the MacOS failures at https://github.com/cirruslabs/osx-images/issues/33#issuecomment-799787772 .)

real-or-random:
> > Is there a specific reason to disable this on clang?
> 
> Yes, CI failed. Apparently clang's optimizer isn't powerful enough for the auto-detect test I have now.

Are you sure? I don't see any failure here in the previous CI runs that seems specific to clang https://cirrus-ci.com/github/bitcoin-core/secp256k1/pull/904. My local clang 11.1 seems clever enough to optimize it (ok that does not say much -- the Debian image on CI has clang 7...)

sipa:
@real-or-random Trying without `CHECK_SIDE_EFFECT_FREE=auto` for clang to see where it fails.

sipa:
Rebased after #693 merge. Had to re-introduce #902's `VERIFY_CHECK_ONLY` to give checks that only apply in VERIFY-mode (and/or can't be proven side-effect free). Given how many `#ifdef VERIFY; VERIFY_CHECK(...); #endif` patterns we already had that can be subsumed by that, I think this isn't too bad.

elichai:
If it works properly on all mainstream compilers then it's pretty cool :)
I am a little worried that some compiler at some version will be able to prove the example on autoconf but won't be able to prove a specific condition in our code, which will cause it to not compile.

But this worry isn't substantiated by anything.

real-or-random:
> By default it is off, to prevent it from biting people unexpectedly that try to compile on slightly less than platforms where the check may fail.

@elichai

elichai:
@real-or-random Oopps should've read the whole thread and not just the OP 😅
Code review ACK 75234ab0eb1c3f5e58c70711a2c0d7ba0689029a

real-or-random:
Here's a slightly improved definition that (on GCC) gives a diagnostic which is readable and is given at compile-time instead of link time.

```
#if SECP256K1_GNUC_PREREQ(4, 4)
    extern int secp256k1_not_supposed_to_survive()
        __attribute__((pure))
        __attribute__((error("VERIFY_CHECK cannot be proven to have no side effects")));
    #define ASSERT_NO_SIDE_EFFECT(cond) do { (void)(secp256k1_not_supposed_to_survive() || (cond)); } while(0)
#else
    extern int VERIFY_CHECK_cannot_be_proven_to_have_no_side_effects;
    #define ASSERT_NO_SIDE_EFFECT(cond) do { (void)(VERIFY_CHECK_cannot_be_proven_to_have_no_side_effects || (cond)); } while(0)
#endif

int main(void) {
    int a = 1;
    int b = 17;
    ASSERT_NO_SIDE_EFFECT(b + a);
    ASSERT_NO_SIDE_EFFECT(a = 0);
    return a;
} 
```

As a nit, I'd prefer `VERIFY_ONLY_CHECK` over `VERIFY_CHECK_ONLY` but if you agree, feel free to add a renaming commit on top, otherwise this could be too much work...)

PR #907, opened 2021-03-21
changed import to use brackets <> for openssl

whb07:
Changes:

OpenSSL imports were changed to use angle brackets for standardization. Typically `#import <foo>` is used for system and or 3rd party imports, rather than the local `#import "bar"`.

Tests:
Ran on Ubuntu 20.04 fine after building with `./configure --enable-openssl-tests`. Exhaustive and other flags were ran, as well as "make check".

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -21,10 +21,10 @@
 #include "util.h"
 
 #ifdef ENABLE_OPENSSL_TESTS
-#include "openssl/bn.h"
-#include "openssl/ec.h"
-#include "openssl/ecdsa.h"
-#include "openssl/obj_mac.h"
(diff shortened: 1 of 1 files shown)


real-or-random (approved):
ACK 4504472269df06b8765b134d41f86619cdcdf8f6 

Indeed, it doesn't matter here but it's cleaner to use the foo `<foo>` syntax in this case, and I don't think there there was a specific reason to use `"foo"`.

whb07:
Right! Standards make it easier to understand what’s going on.

jonasnick (commented):
ACK 4504472269df06b8765b134d41f86619cdcdf8f6

Issue #856, opened 2020-12-04
Add doc/release-process.md


jonasnick:
On our path towards a 1.0 release it would help to start with a 0.1 release soon. Ideally, we'd do this in a systematic way and first discuss how we want the process to look like initially.

It would make sense for the process to be similar to Bitcoin Core's. Here are some suggestions, based on my understanding of [bitcoin-maintainer-tools/make-tag.py](https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/a9eaddb7397f149121a4225b4096b04918d20362/make-tag.py) and [bitcoin/doc/release-process.md](https://github.com/bitcoin/bitcoin/blob/8f7b93047581c67f2133cdb8c7845471de66c30f/doc/release-process.md).

We would copy the start of Core's [configure.ac](https://github.com/bitcoin/bitcoin/blob/dca80ffb45fcc8e6eedb6dc481d500dedab4248b/configure.ac#L10) into ours:
```
define(_CLIENT_VERSION_MAJOR, 0)
define(_CLIENT_VERSION_MINOR, 99)
define(_CLIENT_VERSION_BUILD, 0)
define(_CLIENT_VERSION_RC, 0)
define(_CLIENT_VERSION_IS_RELEASE, false)
```
and replace our `AC_INIT` with
```
AC_INIT([libsecp256k1],m4_join([.], _CLIENT_VERSION_MAJOR, _CLIENT_VERSION_MINOR, _CLIENT_VERSION_BUILD)m4_if(_CLIENT_VERSION_RC, [0], [], [rc]_CLIENT_VERSION_RC))
```
We can print the version at the end of the configure script and a warning if `RELEASE` is not true.

Now in order to create a release candidate we

1. Start [writing release notes](https://github.com/bitcoin/bitcoin/blob/master/doc/release-process.md#write-the-release-notes)
2. Create a corresponding release branch
    1. Create a commit to fix the `VERSIONS` in configure.ac (including `RC=1`)
    2. Tag the commit with a version of make-tag.py that's adapted to libsecp
    3. Create github release
3. To create a new release candidate repeat steps 2.i through 2.iii
4. Once we're happy with the release candidate
    1. Archive the release notes for the new version to doc/release-notes/ (branch master and branch of the release)
    2. Create commit to set `IS_RELEASE` to `true` and `VERSION_RC` to `0`
    3. Tag the release
    4. Create a new GitHub release with a link to the archived release notes

[...]

PR #914, opened 2021-03-26
Introduce symbol to get version of library

rickmark:
This adds additional logic to autoconf for future support of versioning.

I picked 0.8 fairly arbitrarily to represent unstable, 0.9 would be the API stabilization series working to a 1.0

Also a NULL terminated C-String is exported with the symbol `secp256k1_version` that matches.


configure.ac
```diff
@@ -1,5 +1,13 @@
-AC_PREREQ([2.60])
-AC_INIT([libsecp256k1],[0.1])
+AC_PREREQ([2.69])
```
real-or-random, line 1: Why is this necessary?

configure.ac
```diff
@@ -1,5 +1,13 @@
...
+
+define(_LIBSECP256K1_VERSION_MAJOR, 0)
+define(_LIBSECP256K1_VERSION_MINOR, 8)
```
jonasnick, line 4: Should be 99 according to the preliminary plan (https://github.com/bitcoin-core/secp256k1/issues/856). I didn't have a particular reason for this besides matching Bitcoin Core's approach. But if I understand this PR correctly , `secp256k1_version` doesn't include `_LIBSECP256K1_VERSION_IS_RELEASE`. If minor version was 99 for non-releases, `secp256k1_version` can be used to distinguish between releases and non-releases.

configure.ac
rickmark: This was to do what bitcoin core was doing for version and release management, I think I can improve it though

configure.ac
rickmark: Yeah - I picked 0.8 to provide additional headroom for breaking API changes before stabilization.

0.8 could be cut as a branch now-ish

0.9 would represent a nearly complete API and allow for stabilization from 0.9 -> 0.9.9

and 1.0 would reflect final API / ABI

configure.ac
real-or-random: I think we shouldn't increase it to 2.69 without a good reason. I'm not sure if it really works with 2.60 but it's not a disaster if it fails.

real-or-random:
Closing due to no activity. It can always be reopened.

Issue #622, opened 2019-05-23
Add better build documentation


gmaxwell:
The library is designed to be easy to use without the build system, e.g. you can build the tests with nothing more than:

```
$ gcc -o tests src/tests.c -I. -Iinclude/ -DUSE_NUM_NONE -DUSE_FIELD_INV_BUILTIN -DUSE_SCALAR_INV_BUILTIN -DUSE_FIELD_10X26 -DUSE_SCALAR_8X32 -DVERIFY
```

We should document this fact better, because for many embedded systems the whole build system is overkill.  See also #184 

Related, we should also provide packaging instructions for common OSes, see #545

real-or-random:
I'm not sure if we want this but this could be made even easier with just `-DUSE_BASIC_CONFIG` if we include basic-config.h at the appropriate places. It's only used for gen-context at the moment.

gmaxwell:
We could also do some ifdef decision to use 32 vs 64 bit fields by default.

real-or-random:
We currently rely on ./configure to define WORDS_BIGENDIAN, which could lead to very annoying failures.

gmaxwell:
Maybe we should have a #warn on use of 32bit field/scalar when sizeof(void*)==8, I've found some cases of people publishing benchmarks that appear to have used 32bit libsecp256k1 (while comparing to avx2 code in another library...)

real-or-random:
https://github.com/bitcoin-core/secp256k1/issues/621#issuecomment-608348267

jonasnick:
I believe this is superseded by #929.

PR #916, opened 2021-04-10
add ECMULT_GEN_PREC_BITS to basic_config.h

voisine:
set ECMULT_GEN_PREC_BITS to the "auto" value of 4 in basic_config.h, so libsecp can be used without autoconf

### DIFF
--- a/src/basic-config.h
+++ b/src/basic-config.h
@@ -19,6 +19,7 @@
 
 #define USE_WIDEMUL_64 1
 #define ECMULT_WINDOW_SIZE 15
+#define ECMULT_GEN_PREC_BITS 4
 
 #endif /* USE_BASIC_CONFIG */
 


real-or-random:
Thanks for the PR, this is indeed a bug! However, fixing it requires some more cleanup (see CI failures). 

I'm currently playing around with some options. I should have something ready by tomorrow.

real-or-random:
By the way, how do you use this library without autoconf? Do you simply `#include` it or do you build it and then link it? In either case I'd we interested in the exact method, e.g.: How do you define your config? Do you include basic-config.h? What's the command line? 

This could help us towards #622.

voisine:
Thank you. Currently I just #include basic_config.h and then the appropriate libsecp implementations, but I eventually plan to add the source files to an external build system other than “make”

voisine:
Specifically I want to be able to build using Xcode on a clean OS X install without installing additional build tools like autoconf.

jonasnick:
Superseded by #918 (merged)

voisine:
Excellent. My commit even managed to survive into the final merge. Thank you.

PR #919, opened 2021-04-13
Replace gen_context.c with a Python implementation

sipa:
This removes the need for a separate native toolchain to build `gen_context`, and the issues around compilation options affecting one but not the other.

Alternative to #918.


sipa:
Also submitted the test improvements separately as #920.

apoelstra:
concept ACK. I think this would work well for rust-bitcoin.

fanquake:
Concept ACK - ran into issues running gen_context when cross-compiling with mingw-w64 while having Wine installed.

real-or-random:
@fanquake 

As an alternative, @roconnor-blockstream is working on extending gen_context in https://github.com/bitcoin-core/secp256k1/pull/956. The plan (in that PR or in a follow up PR) is to simply ship with precomputed headers then, which would removes the need for running gen_context (or python) entirely for all supported builds.

real-or-random:
Closing in favor of #988 .

PR #920, opened 2021-04-13
Test all ecmult functions with many j*2^i combinations

sipa:
Instead of just testing properties of the points xG for x=-36..36:
* also compute all xG where x=j*2^i for i=0..255 and odd j=1..255.
* test them against known exact results (SHA256 all of them, and compared against an independently created result)
* test all 4 ecmult functions (and for secp256k1_ecmult and secp256k1_ecmult_multi_var, both as G, and through the generic point input)

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -4489,37 +4489,89 @@ void run_wnaf(void) {
     CHECK(secp256k1_scalar_is_zero(&n));
 }
 
+static int test_ecmult_accumulate_cb(secp256k1_scalar* sc, secp256k1_ge* pt, size_t idx, void* data) {
+    const secp256k1_scalar* indata = (const secp256k1_scalar*)data;
+    *sc = *indata;
+    *pt = secp256k1_ge_const_g;
+    CHECK(idx == 0);
+    return 1;
+}
+
+void test_ecmult_accumulate(secp256k1_sha256* acc, const secp256k1_scalar* x, secp256k1_scratch* scratch) {
+    /* Compute x*G in 6 different ways, serialize it uncompressed, and feed it into acc. */
+    secp256k1_gej rj1, rj2, rj3, rj4, rj5, rj6, gj, infj;
+    secp256k1_ge r;
+    const secp256k1_scalar zero = SECP256K1_SCALAR_CONST(0, 0, 0, 0, 0, 0, 0, 0);
+    unsigned char bytes[65];
+    size_t size = 65;
+    secp256k1_gej_set_ge(&gj, &secp256k1_ge_const_g);
+    secp256k1_gej_set_infinity(&infj);
+    secp256k1_ecmult_gen(&ctx->ecmult_gen_ctx, &rj1, x);
+    secp256k1_ecmult(&rj2, &gj, x, &zero);
+    secp256k1_ecmult(&rj3, &infj, &zero, x);
+    secp256k1_ecmult_multi_var(NULL, scratch, &rj4, x, NULL, NULL, 0);
+    secp256k1_ecmult_multi_var(NULL, scratch, &rj5, &zero, test_ecmult_accumulate_cb, (void*)x, 1);
+    secp256k1_ecmult_const(&rj6, &secp256k1_ge_const_g, x, 256);
+    secp256k1_ge_set_gej_var(&r, &rj1);
+    ge_equals_gej(&r, &rj2);
+    ge_equals_gej(&r, &rj3);
+    ge_equals_gej(&r, &rj4);
+    ge_equals_gej(&r, &rj5);
+    ge_equals_gej(&r, &rj6);
+    if (secp256k1_ge_is_infinity(&r)) {
+        /* Store infinity as 0x00 */
+        const unsigned char zerobyte[1] = {0};
+        secp256k1_sha256_write(acc, zerobyte, 1);
+    } else {
+        /* Store other points using their uncompressed serialization. */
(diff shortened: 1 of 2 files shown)


real-or-random (approved):
ACK ec6dff0f532b9f029de6c64af1cd0480651f51f2 diff looks good

src/tests.c
```diff
@@ -3690,37 +3690,77 @@ void run_wnaf(void) {
...
+        secp256k1_sha256_write(acc, bytes, size);
+    }
+}
+
 void test_ecmult_constants(void) {
-    /* Test ecmult_gen() for [0..36) and [order-36..0). */
+    /* Test ecmult_gen for:
+     * - Numbers 0..36 and their negations
+     * - Numbers 2^i (with i=0..255)
+     * - Numbers 2^i + 2^j (with i=0..255, j=i+1..255)
```
jonasnick, line 3726: `i` only goes up to 254.

jonasnick (commented):
ACK mod nit

src/tests.c
real-or-random: Well yeah it's equivalent in the end.

jonasnick: If it's equivalent I'd prefer `i=0..42069`. Fwiw I'm only mentioning this because I added some printfs to this code and was confused for a moment why there was no `i=255`.

@sipa care to fix the micronit? Fine for me either way.

sipa: Oh, I was considering an alternative approach to instead test all numbers of the form `a*(2^N)^b a=1..M b=0..256/M`, for some M and N, as that'd give better coverage of the actual tables involved.

real-or-random:
@sipa Can you fix this? This will be helpful for testing #988 .

src/tests.c
```diff
@@ -3690,37 +3690,77 @@ void run_wnaf(void) {
     CHECK(secp256k1_scalar_is_zero(&n));
 }
 
+void test_ecmult_accumulate(secp256k1_sha256* acc, const secp256k1_scalar* x) {
+    /* Compute x*G in 4 different ways, serialize it uncompressed, and feed it into acc. */
```
robot-dreams, line 3694: Just confirming, `secp256k1_ecmult_multi_var` is out of scope as a 5th way for this test?

robot-dreams (commented):
ACK ec6dff0f532b9f029de6c64af1cd0480651f51f2

I don't have enough context to comment on the choice of test cases (e.g. whether there are other precomputed values that would be good to test), but the test itself looks good and passes locally.

I also checked against an [independent Python implementation](https://gist.github.com/robot-dreams/e5773c6cde579b6ea33a756b2d87cfe1) and got the exact same SHA256 output.

src/tests.c
sipa: I've rewritten it to use (j*2^i) for i=0..255 and odd j=1..255.

src/tests.c
sipa: Nope, added!

sipa:
Rebased, addressed some comments, and made some more changes.

sipa:
FWIW, the Python code to recreate the hash (in the https://github.com/bitcoin/bitcoin/tree/22.0/test/function directory):

```python
from test_framework import key
import hashlib

s = hashlib.sha256()

def add(s, v):
    v = v % key.SECP256K1_ORDER
    if v == 0:
        s.update(b"\x00")
    else:
        keyb = v.to_bytes(32, 'big')
        ec = key.ECKey()
        ec.set(keyb, False)
        s.update(ec.get_pubkey().get_bytes())

for i in range(0, 37):
    add(s, i)
    add(s, -i)

for b in range(256):
  for v in range(1, 256, 2):
    add(s, v << b)

print(s.hexdigest())
```

robot-dreams:
ACK 8ed167b4949898a531a57296b15f749037cc73d0

- Verified again with independent implementation, hash still matches
- Build timeout doesn't seem related to this change; `./configure --with-valgrind=yes; make; make check` passed locally

real-or-random:
Hm, the re-ran didn't work. @sipa Can you increase the timeout or decrease `TEST_ITERS` for the valgrind build?

real-or-random:
@sipa needs trivial rebase :/

sipa:
I don't think the CI failures are related to any change in this PR?

real-or-random:
@sipa You rebased on the wrong branch.

https://github.com/sipa/secp256k1/tree/202104_ecmult_bitcomb says "This branch is 1 commit ahead, 158 commits behind bitcoin-core:master."

real-or-random (approved):
ACK 5eb519e1f60c305e9240946d5773a635192d2a1a

src/tests.c
```diff
@@ -4489,37 +4489,89 @@ void run_wnaf(void) {
...
-            }
-            secp256k1_gej_add_ge(&r, &r, &ng);
-        }
-        CHECK(secp256k1_gej_is_infinity(&r));
-    }
-    for (i = 1; i <= 36; i++ ) {
+    secp256k1_sha256 acc;
+    unsigned char b32[32];
+    int i, j;
+    secp256k1_scratch_space *scratch = secp256k1_scratch_space_create(ctx, 65536);
```
jonasnick, line 4546: nit: thats more than 10 times what's required for a single point to go into the strauss batch codepath.

jonasnick (commented):
ACK 5eb519e1f60c305e9240946d5773a635192d2a1a

src/tests.c
sipa: Fair, but I think it's small enough not to matter.

PR #922, opened 2021-04-17
Add mingw32-w64/wine CI build

sipa:


.cirrus.yml
```diff
@@ -196,3 +196,25 @@ task:
...
+    WINE_CMD: wine
+    HOST: x86_64-w64-mingw32
+    BUILD:
+    WITH_VALGRIND: no
+    ECDH: yes
+    RECOVERY: yes
+    EXPERIMENTAL: yes
+    SCHNORRSIG: yes
+    CTIMETEST: no
+    CFLAGS: -O2 -static -static-libgcc
```
real-or-random, line 216: Do we need the static flags?  I guess it's just my lack of experience with mingw but maybe add a comment.

Also -O2 is our default, so we probably shouldn't set it explicitly here.

real-or-random:
Nice!

Concept ACK

real-or-random:
ok it seems there was a reason why -static was there:
https://cirrus-ci.com/task/5119836217933824?logs=test#L185
`libtool: warning: undefined symbols not allowed in x86_64-w64-mingw32 shared libraries; building static only` (and a few other messages then when linking the benchmarks)

Maybe `./configure --disable-shared` does the job then without specifying CFLAGS. In the end it doesn't matter for CI, I was just curious because we could use this to add build docs for mingw.

real-or-random:
Can you also change 
`file *tests || true` to `file *tests* || true` ? Otherwise it won't match the ".exe"...

sipa:
@real-or-random That wasn't the reason actually. I tried this first in the minisketch repository (see https://github.com/sipa/minisketch/pull/33), and there the issue is that C++ builds to building against shared libgcc by default (something related to being able to catch exceptions across libraries), but I couldn't get that to work with Wine. For C, static libgcc is the default, so this problem doesn't occur. I've just removed it.

real-or-random:
Ok makes sense. We can take care of more improvements to mingw later then (#923).

ACK mod the `file` thing.

sipa:
> Can you also change
> file *tests || true to file *tests* || true ? Otherwise it won't match the ".exe"...

Done.

real-or-random (approved):
ACK e2750b173527a80444929000b32ccd9b103878f7

jonasnick (commented):
utACK e2750b173527a80444929000b32ccd9b103878f7

jonasnick:
Hm any idea why wine complains so loudly (on cirrus). Is that supposed to be ignored?

```
it looks like wine32 is missing, you should install it.
as root, please execute "apt-get install wine32"
```

sipa:
It's only testing 64-bit binaries, so it doesn't need wine32 I think. We could probably just add it to the dockerfile to silence it.

real-or-random:
AFAIU wine32 would be required to run 32-bit binaries, which we don't need.

I'm not sure why Wine warns. Maybe because 32bit is still a common thing on Windows and it's an issue that many people run into?

.cirrus.yml
```diff
@@ -196,3 +196,24 @@ task:
...
   << : *CAT_LOGS
+
+task:
+  name: "x86_64 (mingw32-w64): Windows (Debian stable, Wine)"
+  container:
+    dockerfile: ci/linux-debian.Dockerfile
+    cpu: 1
+    memory: 1G
+  env:
+    WINE_CMD: wine
```
real-or-random, line 207: I think this would hide the warning.
```suggestion
    WINE_CMD: wine64-stable
```

ci/linux-debian.Dockerfile
```diff
@@ -10,4 +10,8 @@ RUN apt-get install --no-install-recommends --no-upgrade -y \
         make automake libtool pkg-config dpkg-dev valgrind qemu-user \
         gcc clang libc6-dbg \
         gcc-i686-linux-gnu libc6-dev-i386-cross libc6-dbg:i386 \
-        gcc-s390x-linux-gnu libc6-dev-s390x-cross libc6-dbg:s390x
+        gcc-s390x-linux-gnu libc6-dev-s390x-cross libc6-dbg:s390x \
+        wine gcc-mingw-w64-x86-64
```
real-or-random, line 14: (not strictly necessary but probably cleaner:)
```suggestion
        wine64 gcc-mingw-w64-x86-64
```

ci/linux-debian.Dockerfile
```diff
@@ -10,4 +10,8 @@ RUN apt-get install --no-install-recommends --no-upgrade -y \
...
+
+# Run a dummy command in wine to make it set up configuration
+RUN wine true || true
```
real-or-random, line 17:
```suggestion
RUN wine64-stable true || true
```

.cirrus.yml
sipa: IIRC I tried that, but it didn't work. I think invoking wine64 directly doesn't start wineserver.

real-or-random:
@sipa Ok, I played around with the docker container. If I'm not mistaken:

 * If you install `wine` (which pulls in `wine64` as dependency), then `wine64 <cmd>` (or `wine64-stable <cmd>`) works without the warning
 * If you install just `wine64`, then `WINESERVER=/usr/lib/wine/wineserver64 wine64-stable <cmd>` works without the warning. (Indeed, without the env variable wineserver can't be found.) 

Both are fine but the first is more natural...

By the way, `wine xcopy` works better as a dummy command. There's no `true` on windows.

sipa:
@real-or-random Ok, used your first variant, and also changed true -> xcopy.

real-or-random:
Ok, great. Now there seems to be one final hiccup:
```
WARNING: could not read 16 bytes from /dev/urandom; falling back to insecure PRNG
```
( https://cirrus-ci.com/task/4912068852711424?logs=test#L258 )

I can *sometimes* reproduce this in the Docker image and then the seed always has some random bytes in front followed by zeros. You can cherry-pick this fix: https://github.com/real-or-random/secp256k1/commit/625b1aa071a78dca3c2ca3377a3500d74914290b

sipa:
@real-or-random Cherry picked.

real-or-random (approved):
ACK ed5a199bed65bf084f34ce18d35807d31a1c75bb

jonasnick (commented):
utACK ed5a199bed65bf084f34ce18d35807d31a1c75bb

Issue #923, opened 2021-04-19
mingw builds


real-or-random:
#922 got me interested in mingw, so I started playing around and reading. Here are some notes that may save others time and should be converted into a PR at some point. I'd appreciate any comments from people more familiar with the topic (@gmaxwell, @theuni).

secp256k1 has been written to support mingw builds. As demonstrated in #922, these builds (still) work in general. But a few caveats apply.

Here are some simple issues to start with:
 - [x] The Makefile is not written with Windows in mind. The most obvious issue is that they ignore the `.exe` extension, e.g., in `make clean` and for `gen_context`. [Automake has explicit support for this](https://www.gnu.org/software/automake/manual/html_node/EXEEXT.html), which we can rely easily rely on.
 - [x] When cross-compiling on Linux, it may happen that `configure` believes that we're not cross-compiling because it's possible to execute windows binaries with an installed Wine and `binfmt_misc` enables. (This happens on my system but not on CI). See https://stackoverflow.com/questions/13150631/cross-compiling-why-checking-whether-we-are-cross-compiling-no#comment40154987_13150651. This by itself is not an issue: There's per se nothing wrong with configure building `gen_context.exe` and using wine to execute it. It's a little bit cumbersome though and together with the previous item that means that the `gen_context` invocation will fail. (#919 may improve this but could create other problems because invoking Python portably could be difficult.)  

The real fun starts with DLLs (shared libaries) which are a pain in the ass. When creating a DLL, one needs to declare exported symbols explictly using `__declspec (dllexport)`. We currently do this via in `secp256k1.h` via `SECP256K1_API` when `SECP256K1_BUILD` is defined.

[...]

PR #925, opened 2021-04-25
changed include statements without prefix 'include/' 

whb07:
Referencing #924 , this PR splits the two issues brought on to a smaller to digest change. What this does is removes the prefix "include/" when referencing the local library header files.

e.g:
from:
```cpp
#include "include/secp256k1.h"
```
to:
```cpp
#include "secp256k1.h"
```

Rationale besides styling and consistency across other files in the repo, it makes it easier for outside builds to properly locate the headers.

A live example seen here when attempting to build this library within bitcoin repo:
```sh
[ 14%] Building CXX object leveldb/CMakeFiles/leveldb.dir/util/bloom.cc.o
/tmp/bitcoin/src/secp256k1/src/secp256k1.c:7:10: fatal error: include/secp256k1.h: No such file or directory
    7 | #include "include/secp256k1.h"
      |          ^~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [secp256k1/CMakeFiles/Secp256k1.dir/build.make:76: secp256k1/CMakeFiles/Secp256k1.dir/src/secp256k1.c.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:537: secp256k1/CMakeFiles/Secp256k1.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....

```

### DIFF
--- a/src/modules/ecdh/main_impl.h
+++ b/src/modules/ecdh/main_impl.h
@@ -7,8 +7,8 @@
 #ifndef SECP256K1_MODULE_ECDH_MAIN_H
 #define SECP256K1_MODULE_ECDH_MAIN_H
 
-#include "include/secp256k1_ecdh.h"
-#include "ecmult_const_impl.h"
+#include "../../../include/secp256k1_ecdh.h"
+#include "../../ecmult_const_impl.h"
 
 static int ecdh_hash_function_sha256(unsigned char *output, const unsigned char *x32, const unsigned char *y32, void *data) {
     unsigned char version = (y32[31] & 0x01) | 0x02;
--- a/src/modules/extrakeys/main_impl.h
+++ b/src/modules/extrakeys/main_impl.h
@@ -7,8 +7,8 @@
 #ifndef SECP256K1_MODULE_EXTRAKEYS_MAIN_H
 #define SECP256K1_MODULE_EXTRAKEYS_MAIN_H
 
-#include "include/secp256k1.h"
-#include "include/secp256k1_extrakeys.h"
+#include "../../../include/secp256k1.h"
+#include "../../../include/secp256k1_extrakeys.h"
 
 static SECP256K1_INLINE int secp256k1_xonly_pubkey_load(const secp256k1_context* ctx, secp256k1_ge *ge, const secp256k1_xonly_pubkey *pubkey) {
     return secp256k1_pubkey_load(ctx, ge, (const secp256k1_pubkey *) pubkey);
--- a/src/modules/extrakeys/tests_exhaustive_impl.h
+++ b/src/modules/extrakeys/tests_exhaustive_impl.h
@@ -8,7 +8,7 @@
 #define SECP256K1_MODULE_EXTRAKEYS_TESTS_EXHAUSTIVE_H
 
 #include "src/modules/extrakeys/main_impl.h"
-#include "include/secp256k1_extrakeys.h"
+#include "../../../include/secp256k1_extrakeys.h"
 
 static void test_exhaustive_extrakeys(const secp256k1_context *ctx, const secp256k1_ge* group) {
     secp256k1_keypair keypair[EXHAUSTIVE_TEST_ORDER - 1];
--- a/src/modules/recovery/main_impl.h
+++ b/src/modules/recovery/main_impl.h
@@ -7,7 +7,7 @@
 #ifndef SECP256K1_MODULE_RECOVERY_MAIN_H
 #define SECP256K1_MODULE_RECOVERY_MAIN_H
 
-#include "include/secp256k1_recovery.h"
+#include "../../../include/secp256k1_recovery.h"
 
 static void secp256k1_ecdsa_recoverable_signature_load(const secp256k1_context* ctx, secp256k1_scalar* r, secp256k1_scalar* s, int* recid, const secp256k1_ecdsa_recoverable_signature* sig) {
     (void)ctx;
--- a/src/modules/recovery/tests_exhaustive_impl.h
+++ b/src/modules/recovery/tests_exhaustive_impl.h
@@ -8,7 +8,7 @@
 #define SECP256K1_MODULE_RECOVERY_EXHAUSTIVE_TESTS_H
 
 #include "src/modules/recovery/main_impl.h"
-#include "include/secp256k1_recovery.h"
+#include "../../../include/secp256k1_recovery.h"
 
 void test_exhaustive_recovery_sign(const secp256k1_context *ctx, const secp256k1_ge *group) {
     int i, j, k;
--- a/src/modules/schnorrsig/main_impl.h
+++ b/src/modules/schnorrsig/main_impl.h
@@ -7,9 +7,9 @@
 #ifndef SECP256K1_MODULE_SCHNORRSIG_MAIN_H
 #define SECP256K1_MODULE_SCHNORRSIG_MAIN_H
 
-#include "include/secp256k1.h"
-#include "include/secp256k1_schnorrsig.h"
-#include "hash.h"
+#include "../../../include/secp256k1.h"
+#include "../../../include/secp256k1_schnorrsig.h"
+#include "../../hash.h"
 
 /* Initializes SHA256 with fixed midstate. This midstate was computed by applying
  * SHA256 to SHA256("BIP0340/nonce")||SHA256("BIP0340/nonce"). */
--- a/src/modules/schnorrsig/tests_exhaustive_impl.h
+++ b/src/modules/schnorrsig/tests_exhaustive_impl.h
@@ -7,7 +7,7 @@
 #ifndef SECP256K1_MODULE_SCHNORRSIG_TESTS_EXHAUSTIVE_H
 #define SECP256K1_MODULE_SCHNORRSIG_TESTS_EXHAUSTIVE_H
 
-#include "include/secp256k1_schnorrsig.h"
+#include "../../../include/secp256k1_schnorrsig.h"
 #include "src/modules/schnorrsig/main_impl.h"
 
 static const unsigned char invalid_pubkey_bytes[][32] = {
--- a/src/modules/schnorrsig/tests_impl.h
+++ b/src/modules/schnorrsig/tests_impl.h
@@ -7,7 +7,7 @@
 #ifndef SECP256K1_MODULE_SCHNORRSIG_TESTS_H
 #define SECP256K1_MODULE_SCHNORRSIG_TESTS_H
 
-#include "secp256k1_schnorrsig.h"
+#include "../../../include/secp256k1_schnorrsig.h"
 
 /* Checks that a bit flip in the n_flip-th argument (that has n_bytes many
  * bytes) changes the hash function
--- a/contrib/lax_der_parsing.c
+++ b/contrib/lax_der_parsing.c
@@ -5,7 +5,6 @@
  ***********************************************************************/
 
 #include <string.h>
-#include <secp256k1.h>
 
 #include "lax_der_parsing.h"
 
--- a/contrib/lax_der_parsing.h
+++ b/contrib/lax_der_parsing.h
@@ -51,7 +51,9 @@
 #ifndef SECP256K1_CONTRIB_LAX_DER_PARSING_H
 #define SECP256K1_CONTRIB_LAX_DER_PARSING_H
 
+#ifndef SECP256K1_H
 #include <secp256k1.h>
+#endif
 
 #ifdef __cplusplus
 extern "C" {
--- a/contrib/lax_der_privatekey_parsing.c
+++ b/contrib/lax_der_privatekey_parsing.c
@@ -5,7 +5,6 @@
  ***********************************************************************/
 
 #include <string.h>
-#include <secp256k1.h>
 
 #include "lax_der_privatekey_parsing.h"
 
--- a/contrib/lax_der_privatekey_parsing.h
+++ b/contrib/lax_der_privatekey_parsing.h
@@ -28,7 +28,9 @@
 #ifndef SECP256K1_CONTRIB_BER_PRIVATEKEY_H
 #define SECP256K1_CONTRIB_BER_PRIVATEKEY_H
 
+#ifndef SECP256K1_H
 #include <secp256k1.h>
+#endif
 
 #ifdef __cplusplus
 extern "C" {
--- a/src/bench_ecdh.c
+++ b/src/bench_ecdh.c
@@ -6,8 +6,8 @@
 
 #include <string.h>
 
-#include "include/secp256k1.h"
-#include "include/secp256k1_ecdh.h"
+#include "../include/secp256k1.h"
+#include "../include/secp256k1_ecdh.h"
 #include "util.h"
 #include "bench.h"
 
--- a/src/bench_ecmult.c
+++ b/src/bench_ecmult.c
@@ -6,8 +6,8 @@
 #include <stdio.h>
 
 #include "secp256k1.c"
+#include "../include/secp256k1.h"
 
-#include "include/secp256k1.h"
 #include "util.h"
 #include "hash_impl.h"
 #include "field_impl.h"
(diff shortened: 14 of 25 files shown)


gmaxwell:
Alternatively, the internal usage could use correct relative paths, changing include/ to ../include/.  This is a smaller diff, and doesn't require any include settings in the build system to build the library:

E.g.

```
$ gcc -Wall -Wextra -Wno-unused-function -O2  -c ./src/secp256k1.c -DSECP256K1_BUILD -D ECMULT_GEN_PREC_BITS=4 -D ECMULT_WINDOW_SIZE=15 && ls secp256k1.o
secp256k1.o
```

I believe classically  "" includes didn't perform any search, so paths had to be explicit... though all modern compilers do search with it. 

As far as that Cmake build goes-- it looks broken.  the top level directory is (currently) supposed to be in the includes path.  I went to go find the cmake input to see if it was making more serious errors (e.g. failing to include SECP256K1_BUILD)  but I couldn't find it.

whb07:
> Alternatively, the internal usage could use correct relative paths, changing include/ to ../include/. This is a smaller diff, and doesn't require any include settings in the build system to build the library:

The deeper nested includes would end up needing more `../../../` as a prefix (don’t quote me on the preciseness of the example).

i agree there’s multiple ways to skin this cat, so let’s get something done?

i can make the changes be a relative path if you will it so.

real-or-random:
> Alternatively, the internal usage could use correct relative paths, changing include/ to ../include/. This is a smaller diff, and doesn't require any include settings in the build system to build the library:
> 
> E.g.
> 
> ```
> $ gcc -Wall -Wextra -Wno-unused-function -O2  -c ./src/secp256k1.c -DSECP256K1_BUILD -D ECMULT_GEN_PREC_BITS=4 -D ECMULT_WINDOW_SIZE=15 && ls secp256k1.o
> secp256k1.o
> ```
> 
> I believe classically "" includes didn't perform any search, so paths had to be explicit... though all modern compilers do search with it.

I noticed this include problem when looking into #923, and I agree we should aim for a solution that does not need `-I` when building without autotools. So I'd prefer to add `..` to the `#include`. Unless this results in some other problem but I don't see any.

whb07:
> I noticed this include problem when looking into #923, and I agree we should aim for a solution that does not need `-I` when building without autotools. So I'd prefer to add `..` to the `#include`. Unless this results in some other problem but I don't see any.

I have just made the changes to be a relative path as mentioned. The changes to the `Makefile.am` file have been reverted to reflect the state of the master branch.

The changes do fix the broken cmake build as well so thats a plus.

real-or-random:
The changes look fine. Let's see if the CI build works. If yes, can you squash this into a single commit?

whb07:
> The changes look fine. Let's see if the CI build works. If yes, can you squash this into a single commit?

Yep, I'll squash once the build passes.

Makefile.am
```diff
@@ -162,3 +162,4 @@ endif
 if ENABLE_MODULE_SCHNORRSIG
 include src/modules/schnorrsig/Makefile.am.include
 endif
+
```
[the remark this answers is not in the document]
whb07: Eagle eyes! Isn’t the standard to have a “\n” as the last line?

> 123 A source file that is not empty shall end in a new-line character, which shall not be immediately preceded by a backslash character before any such splicing takes place.

[quick-reference](http://www.c0x.coding-guidelines.com/5.1.1.2.html)

i got an editor and GitHub warning so I caved and added it 🤷‍♂️.

let me know what you want to do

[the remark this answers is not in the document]
gmaxwell: The file already ended with a newline, this is adding an additional blank line.

gmaxwell:
It should probably also change tests.c:

```c
#include "../contrib/lax_der_parsing.c"
#include "../contrib/lax_der_privatekey_parsing.c"
```

Though this isn't enough to compile tests.c with no -I lines to the compiler.  The issue is that the modules in contrib correctly use <secp256k1.h> as its 'end user code' -- but then that doesn't work as expected here.  Still, I think moving the issue back to there is more correct, and at least fewer -I are needed.

Makefile.am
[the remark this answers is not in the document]
whb07: You’re right! Gross, GitHub is showing that there isn’t one! Fixing it right now

whb07:
> It should probably also change tests.c:
> 
> ```c
> #include "../contrib/lax_der_parsing.c"
> #include "../contrib/lax_der_privatekey_parsing.c"
> ```
> 
> Though this isn't enough to compile tests.c with no -I lines to the compiler. The issue is that the modules in contrib correctly use <secp256k1.h> as its 'end user code' -- but then that doesn't work as expected here. Still, I think moving the issue back to there is more correct, and at least fewer -I are needed.


This is another good catch. Though the action isn't clear to me. You're saying we should change this to be a relative path as well `../contrib`?

gmaxwell:
Yes, it should be changed to ../contrib.

real-or-random:
We could do something like this in the contrib files:

```
/* #include secp256k1.h only when it hasn't been included yet.
   This enables this file to be #included directly in other project
   files (such as tests.c) without the need to set an explicit -I flag,
   which would be necessary to locate secp256k1.h.  */
#ifndef SECP256K1_H
#include <secp256k1.h>
#endif
```

whb07:
> We could do something like this in the contrib files:
> 
> ```
> /* #include secp256k1.h only when it hasn't been included yet.
>    This enables this file to be #included directly in other project
>    files (such as tests.c) without the need to set an explicit -I flag,
>    which would be necessary to locate secp256k1.h.  */
> #ifndef SECP256K1_H
> #include <secp256k1.h>
> #endif
> ```

Care to expand? I was able to properly build outside of the repo the executable `tests.c` and run it just fine post changes (including the "../contrib" ones).

real-or-random:
> > We could do something like this in the contrib files:
> > ```
> > /* #include secp256k1.h only when it hasn't been included yet.
> >    This enables this file to be #included directly in other project
> >    files (such as tests.c) without the need to set an explicit -I flag,
> >    which would be necessary to locate secp256k1.h.  */
> > #ifndef SECP256K1_H
> > #include <secp256k1.h>
> > #endif
> > ```
> 
> Care to expand? I was able to properly build outside of the repo the executable `tests.c` and run it just fine post changes (including the "../contrib" ones).

Maybe you have `secp256k1.h` installed in your system (in `/usr/include`)?

I think what should happen is that when trying to compile without `-I`, e.g., 
```
gcc  -c ./src/tests.c -DSECP256K1_BUILD -D ECMULT_GEN_PREC_BITS=4 -D ECMULT_WINDOW_SIZE=15
```
then gcc, when it encouters `#include <secp256k1.h>`, it looks up `secp256k1.h` in the system paths and would error out if the file can't be found.

See https://gcc.gnu.org/onlinedocs/cpp/Search-Path.html

gmaxwell:
@real-or-random  Seems fine to me, ever so slightly kludgy but argument free compiles are nice and the test harness is kind of inherently weird because its a test.

src/modules/ecdh/main_impl.h
```diff
@@ -7,7 +7,7 @@
 #ifndef SECP256K1_MODULE_ECDH_MAIN_H
 #define SECP256K1_MODULE_ECDH_MAIN_H
 
-#include "include/secp256k1_ecdh.h"
+#include "../include/secp256k1_ecdh.h"
 #include "ecmult_const_impl.h"
```
gmaxwell, line 11: ../../ecmult_const_impl.h

src/modules/schnorrsig/main_impl.h
```diff
@@ -7,8 +7,8 @@
 #ifndef SECP256K1_MODULE_SCHNORRSIG_MAIN_H
 #define SECP256K1_MODULE_SCHNORRSIG_MAIN_H
 
-#include "include/secp256k1.h"
-#include "include/secp256k1_schnorrsig.h"
+#include "../include/secp256k1.h"
+#include "../include/secp256k1_schnorrsig.h"
 #include "hash.h"
```
gmaxwell, line 12: ../../hash.h

src/modules/ecdh/main_impl.h
```diff
@@ -7,7 +7,7 @@
 #ifndef SECP256K1_MODULE_ECDH_MAIN_H
 #define SECP256K1_MODULE_ECDH_MAIN_H
 
-#include "include/secp256k1_ecdh.h"
+#include "../include/secp256k1_ecdh.h"
```
gmaxwell, line 10: ../../../include/

src/modules/extrakeys/main_impl.h
```diff
@@ -7,8 +7,8 @@
 #ifndef SECP256K1_MODULE_EXTRAKEYS_MAIN_H
 #define SECP256K1_MODULE_EXTRAKEYS_MAIN_H
 
-#include "include/secp256k1.h"
-#include "include/secp256k1_extrakeys.h"
+#include "../include/secp256k1.h"
+#include "../include/secp256k1_extrakeys.h"
```
gmaxwell, line 11: ../../../include/  for both

src/modules/extrakeys/tests_exhaustive_impl.h
```diff
@@ -8,7 +8,7 @@
 #define SECP256K1_MODULE_EXTRAKEYS_TESTS_EXHAUSTIVE_H
 
 #include "src/modules/extrakeys/main_impl.h"
-#include "include/secp256k1_extrakeys.h"
+#include "../include/secp256k1_extrakeys.h"
```
gmaxwell, line 11: ../../../include/

src/modules/recovery/main_impl.h
```diff
@@ -7,7 +7,7 @@
 #ifndef SECP256K1_MODULE_RECOVERY_MAIN_H
 #define SECP256K1_MODULE_RECOVERY_MAIN_H
 
-#include "include/secp256k1_recovery.h"
+#include "../include/secp256k1_recovery.h"
```
gmaxwell, line 10: ../../../include/

src/modules/recovery/tests_exhaustive_impl.h
```diff
@@ -8,7 +8,7 @@
 #define SECP256K1_MODULE_RECOVERY_EXHAUSTIVE_TESTS_H
 
 #include "src/modules/recovery/main_impl.h"
-#include "include/secp256k1_recovery.h"
+#include "../include/secp256k1_recovery.h"
```
gmaxwell, line 11: ../../../include/

src/modules/schnorrsig/main_impl.h
```diff
@@ -7,8 +7,8 @@
 #ifndef SECP256K1_MODULE_SCHNORRSIG_MAIN_H
 #define SECP256K1_MODULE_SCHNORRSIG_MAIN_H
 
-#include "include/secp256k1.h"
-#include "include/secp256k1_schnorrsig.h"
+#include "../include/secp256k1.h"
```
gmaxwell, line 10: ../../../include/

src/modules/schnorrsig/tests_exhaustive_impl.h
```diff
@@ -7,7 +7,7 @@
 #ifndef SECP256K1_MODULE_SCHNORRSIG_TESTS_EXHAUSTIVE_H
 #define SECP256K1_MODULE_SCHNORRSIG_TESTS_EXHAUSTIVE_H
 
-#include "include/secp256k1_schnorrsig.h"
+#include "../include/secp256k1_schnorrsig.h"
```
gmaxwell, line 10: ../../../include/

src/modules/schnorrsig/tests_impl.h
```diff
@@ -7,7 +7,7 @@
 #ifndef SECP256K1_MODULE_SCHNORRSIG_TESTS_H
 #define SECP256K1_MODULE_SCHNORRSIG_TESTS_H
 
-#include "secp256k1_schnorrsig.h"
+#include "../include/secp256k1_schnorrsig.h"
```
gmaxwell, line 10: ../../../include/

gmaxwell:
```
$ gcc -o ./tests ./src/tests.c -DECMULT_GEN_PREC_BITS=4 -DECMULT_WINDOW_SIZE=15 -DENABLE_MODULE_SCHNORRSIG -DENABLE_MODULE_EXTRAKEYS -DENABLE_MODULE_RECOVERY -DENABLE_MODULE_ECDH
```

should work without errors.  After applying my above review comments and the include work around mentioned by real-or-random (it's needed in four places in contrib, everywhere <secp256k1.h> is used) it does for me.

real-or-random:
This needs rebase now that #928 has been merged.

> @real-or-random Seems fine to me, ever so slightly kludgy but argument free compiles are nice and the test harness is kind of inherently weird because its a test.

That was also my thinking. Tt's not exactly elegant but it does the job in the end...

gmaxwell:
Great!

This just needs code like

```c
/* #include secp256k1.h only when it hasn't been included yet.
   This enables this file to be #included directly in other project
   files (such as tests.c) without the need to set an explicit -I flag,
   which would be necessary to locate secp256k1.h.  */
#ifndef SECP256K1_H
#include <secp256k1.h>
#endif
```

around the uses of ```#include <secp256k1.h>``` in contrib, and then

```
gcc -o ./tests ./src/tests.c -DECMULT_GEN_PREC_BITS=4 -DECMULT_WINDOW_SIZE=15 -DENABLE_MODULE_SCHNORRSIG -DENABLE_MODULE_EXTRAKEYS -DENABLE_MODULE_RECOVERY -DENABLE_MODULE_ECDH
```

will work.

real-or-random:
Argh, this needs rebase again (and CI failures just appeared because #925 was merged during CI was running...). Sorry that this needs so many ping pongs.

whb07:
> ```c
> #ifndef SECP256K1_H
> #include <secp256k1.h>
> #endif
> ```

This works, I'll tackle it tomorrow morning assuming this is the way to move forward. Sometimes i don't understand how the search paths are distinguished for these compilers, seems I don't read English too good when looking at their docs!

gmaxwell:
ACK on the update so far, looking forward to the rest. I'll review as soon as it's up. Thanks!

whb07:
This *should* be good to go now. For better context:

In `contrib/lax_der_parsing.c` and `contrib/lax_der_privatekey_parsing.c` the include statement for `secp256k1.h` has been removed and the header guard protected include for respective header files imports secp256k1.h instead.

This cleans up the unnecessary wordy include duplication and satisfies the definitions by being included in the header which is required for the source file to run.

As mentioned previously:

```bash
$ gcc -o ./tests ./src/tests.c -DECMULT_GEN_PREC_BITS=4 -DECMULT_WINDOW_SIZE=15 -DENABLE_MODULE_SCHNORRSIG -DENABLE_MODULE_EXTRAKEYS -DENABLE_MODULE_RECOVERY -DENABLE_MODULE_ECDH
```
&&
```bash
$ clang -o ./tests ./src/tests.c -DECMULT_GEN_PREC_BITS=4 -DECMULT_WINDOW_SIZE=15 -DENABLE_MODULE_SCHNORRSIG -DENABLE_MODULE_EXTRAKEYS -DENABLE_MODULE_RECOVERY -DENABLE_MODULE_ECDH
```

Run just fine now. I even ran the "include  what you use" tooling and it gives good-to-go.

whb07:
The test suite needs to be run again, seems like theres an underlying issue there.

real-or-random:
> The test suite needs to be run again, seems like theres an underlying issue there.

his failure is really spurious. If you look at the logs, all the tests pass, just Cirrus is confused about about result. (I reported it at 
https://github.com/cirruslabs/cirrus-cli/issues/373#issuecomment-832829630 .) Let me restart that ask, but let's not get that in our way.

real-or-random (approved):
ACK 3c90bdda95aa4e79ff33bfbbbe91872417650ae9 code looks good and even the tests compile fine now without `-I` args

gmaxwell:
ACK 3c90bdda95aa4e79ff33bfbbbe91872417650ae9

sipa:
Posthumous utACK 3c90bdda95aa4e79ff33bfbbbe91872417650ae9

PR #926, opened 2021-04-28
secp256k1.h: clarify that by default arguments must be != NULL

jonasnick:
The same file says that the illegal callback will only triger for violations
explicitly mentioned, which is not true without this commit because we often
don't mention that an argument is not allowed to be NULL.

This line is extracted from #783 in the hope that it gets merged faster because other PRs depend on it.

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -7,7 +7,9 @@ extern "C" {
 
 #include <stddef.h>
 
-/* These rules specify the order of arguments in API calls:
+/* Unless explicitly stated all pointer arguments must not be NULL.
+ *
+ * The following rules specify the order of arguments in API calls:
  *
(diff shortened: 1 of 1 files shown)


gmaxwell:
This seems fine to me, though FWIW, calling any the non-nullable inputs with a statically null input will cause warnings in the caller.  And the non-nullable arguments should all be documented as non-nullable with the annotations in the header (though not in doxygen, indeed).

jonasnick:
Agreed. Pushed an unrelated (slight) tweak.

real-or-random (approved):
ACK 0881633dfd0c530a915cf63be295f00841c94cc4

gmaxwell:
ACK 0881633dfd0c530a915cf63be295f00841c94cc4

PR #928, opened 2021-05-01
Define SECP256K1_BUILD in secp256k1.c directly.

### LINKED ISSUE #927: Make it impossible to build without SECP256K1_BUILD defined
If the library itself is compiled without -DSECP256K1_BUILD then it will be (minorly) miscompiled:  all the nullness checks on the interfaces which specify that their arguments be non-null will get optimized out (the exported symbols also won't get flagged for windows, but presumably where that matters it'll get noticed).

Is there any reason to not just have secp256k1.c #define SECP256K1_BUILD before including any headers and then drop any requirement for the build system?

If that can't be done, it should probably check and #error if it isn't defined.  But making it configuration-free seems much better to me.

gmaxwell:
This avoids building without it and makes it safer to use a custom
 building environment.  Test harnesses need to #include secp256k1.c
 first now.

Fixes #927

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -127,6 +127,17 @@ typedef int (*secp256k1_nonce_function)(
 #  define SECP256K1_INLINE inline
 # endif
 
+/** When this header is used at build-time the SECP256K1_BUILD define needs to be set
+ *  to correctly setup export attributes and nullness checks.  This is normally done
+ *  by secp256k1.c but to guard against this header being included before secp256k1.c
+ *  has had a chance to set the define (e.g. via test harnesses that just includes
+ *  secp256k1.c) we set SECP256K1_NO_BUILD when this header is processed without the
+ *  BUILD define so this condition can be caught.
+ */
+#ifndef SECP256K1_BUILD
+# define SECP256K1_NO_BUILD
+#endif
+
 #ifndef SECP256K1_API
 # if defined(_WIN32)
 #  ifdef SECP256K1_BUILD
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -4,6 +4,8 @@
  * file COPYING or https://www.opensource.org/licenses/mit-license.php.*
  ***********************************************************************/
 
+#define SECP256K1_BUILD
+
 #include "include/secp256k1.h"
 #include "include/secp256k1_preallocated.h"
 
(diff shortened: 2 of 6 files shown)


sipa:
Does this risk accidentally including secp256k1.h somewhere before secp256k1.c, thus getting the wrong interface definitions? If so, perhaps it's a possibility to have .h define a `SECP256K1_NO_BUILD` (if `SECP256K1_BUILD wasn't defined already). .c can then error out if `SECP256K1_NO_BUILD` was defined.

gmaxwell:
Indeed and the PR fixed a couple tests that were doing exactly that.  I didn't consider it much of a concern because the non-null warnings catch it in development... but you're absolutely right that this is easy to avoid and so I added your suggestion.

include/secp256k1.h
```diff
@@ -127,6 +127,10 @@ typedef int (*secp256k1_nonce_function)(
 #  define SECP256K1_INLINE inline
 # endif
 
+#ifndef SECP256K1_BUILD
+# define SECP256K1_NO_BUILD
+#endif
```
real-or-random, line 132: nit: I think this would be a good place to spend one or two sentences to explain the purpose of these macros. (My personal experience: I think I had two ask a few times before I understood the entire concept.)

real-or-random (commented):
ACK mod nit

include/secp256k1.h
gmaxwell: There is a comment below on the nullness macro. But I can add one to the guard about the guard specifically.

sipa:
utACK ae9e648526ceaf7cd97ba4dfe3c105db8e226c35

real-or-random (approved):
ACK ae9e648526ceaf7cd97ba4dfe3c105db8e226c35

src/secp256k1.c
```diff
@@ -4,6 +4,8 @@
  * file COPYING or https://www.opensource.org/licenses/mit-license.php.*
  ***********************************************************************/
 
+#define SECP256K1_BUILD
```
elichai, line 7: Shouldn't this be wrapped in `#ifndef SECP256k1_BUILD`?

gmaxwell: I don't see why-- if it was you wouldn't have spotted that your rust build was incorrectly setting it now that it shouldn't be set that way anymore.

elichai: I just thought it was common practice to wrap defines in ifndef in the case that the define acts as a flag and doesn't carry any additional value.
(I don't think there was any harm in the fact that we kept passing `-DSECP256K1_BUILD` to the compiler, as we don't really use "headers" in rust)

PR #930, opened 2021-05-02
Add ARM32/ARM64 CI

sipa:


.cirrus.yml
```diff
@@ -197,6 +197,55 @@ task:
...
+    EXPERIMENTAL: yes
+    SCHNORRSIG: yes
+    CTIMETEST: no
+  matrix:
+    - env: {}
+    - env: {ASM: arm}
+  << : *MERGE_BASE
+  test_script:
+    # https://sourceware.org/bugzilla/show_bug.cgi?id=27008
+    - rm /etc/ld.so.cache
```
real-or-random, line 222: I don't think we need these lines. This is really only an issue when using qemu with a different endianness, so this affects us only on s390x.

real-or-random (approved):
ACK mod nit

real-or-random:
Wait, the expectation was probably that this should break to due #931. 

I think this is related to the "shortcut" in configure.ac that simply uses the native toolchain to build gen_context when we're not cross-compiling. In this case, we are cross compiling, so we don't take that shortcut. 

I'll have a look.

real-or-random:
> Wait, the expectation was probably that this should break due to #931.
> 
> I think this is related to the "shortcut" in configure.ac that simply uses the native toolchain to build gen_context when we're not cross-compiling. In this case, we are cross compiling, so we don't take that shortcut.
> 
> I'll have a look.

Ok I think I was a little confused when I wrote that. The issue #931 appears when compiling gen_context for the build machine. On CI, we cross-compile for ARM, so building gen_context *for the CI* machine is never an issue.

I think no simple CI test could test for the breakage in #931. We could of course cross-compile `gen_context` and then run it on qemu. 

One possible way to do so without fiddling around with the build system would be to rely on `binfmt_misc`, a kernel feature which allows you to run commands like `./cmd` which then will be automatically wrapped, e.g., by qemu or wine. (This is why my machine will build and run `gen_context.exe` when cross-compiling for mingw...) But I heard `binfmt_misc` has some issues with Docker.

Another possible way is to simply add a CI job that manually cross-compiles `gen_context` for ARM and runs it on qemu. But this doesn't really test if the autoconf build works.

So I'm not convinced that testing this on CI is worth the hassle.

Anyway, my ACK on this PR is still valid.

sipa:
Indeed, I realized too late that #931 was in fact related to the build environment for gen_context rather than the ARM asm per se.

Still, these are very useful CI targets I think.

real-or-random:
> Still, these are very useful CI targets I think.

For sure, and this fixes #598.

gmaxwell:
if this supports a full docker environment could we just run qemu-system and run a whole emulated distribution? :P

.cirrus.yml
sipa: Done, let's see.

sipa:
@gmaxwell I think that's an option, but something I'd like to keep for later.

real-or-random (approved):
ACK 8bbad7a18e5dc5054b27ae44ea0c8dffe050f6bf CI output looks fine

jonasnick (commented):
ACK 8bbad7a18e5dc5054b27ae44ea0c8dffe050f6bf

PR #935, opened 2021-05-03
Fix gen_context/ASM build on ARM

### LINKED ISSUE #931: Arm ASM builds are broken
Gen_context tries to use ASM but doesn't get linked against it:

```
gcc -O2 -fvisibility=hidden -std=c89 -pedantic -Wall -Wextra -Wcast-align -Wnested-externs -Wshadow -Wstrict-prototypes -Wundef -Wno-unused-function -Wno-long-long -Wno-overlength-strings -W -g  gen_context.o -o gen_context
gen_context.o: In function `secp256k1_fe_mul':
/home/gmaxwell/secp256k1/src/field_10x26_impl.h:1078: undefined reference to `secp256k1_fe_mul_inner'
gen_context.o: In function `secp256k1_fe_sqr':
/home/gmaxwell/secp256k1/src/field_10x26_impl.h:1091: undefined reference to `secp256k1_fe_sqr_inner'
/home/gmaxwell/secp256k1/src/field_10x26_impl.h:1091: undefined reference to `secp256k1_fe_sqr_inner'
/home/gmaxwell/secp256k1/src/field_10x26_impl.h:1091: undefined reference to `secp256k1_fe_sqr_inner'
/home/gmaxwell/secp256k1/src/field_10x26_impl.h:1091: undefined reference to `secp256k1_fe_sqr_inner'
...
```

Reverting a3aa2628c7b675814157556d774872755c9f1aba fixes it.

real-or-random:
These are two attempts to fix #931.

@gmaxwell Can you test if *either* if the first commits solves the issue?
(The third commit is just a minor cleanup thing that I discovered on the way).

If yes, I'd probably prefer the first commit, which is a real fix. But we could still apply the second commit if people think that's a good idea.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -126,13 +126,13 @@ endif
 if USE_ECMULT_STATIC_PRECOMPUTATION
 CPPFLAGS_FOR_BUILD +=-I$(top_srcdir) -I$(builddir)/src
 
-gen_context_OBJECTS = gen_context.o
+gen_context_OBJECTS = $(COMMON_LIB) gen_context.o
 gen_context_BIN = gen_context$(BUILD_EXEEXT)
 gen_%.o: src/gen_%.c src/libsecp256k1-config.h
-	$(CC_FOR_BUILD) $(CPPFLAGS_FOR_BUILD) $(CFLAGS_FOR_BUILD) -c $< -o $@
+	$(CC_FOR_BUILD) $(DEFS) $(CPPFLAGS_FOR_BUILD) $(CFLAGS_FOR_BUILD) -c $< -o $@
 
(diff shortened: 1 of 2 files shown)


real-or-random:
Ok, CI answered my question: https://cirrus-ci.com/task/5313033309257728

The idea of the first commit is of course stupid and does not work when cross-compiling. We can't simply link the `$(COMMON_LIB)` (e.g., built for ARM) into `gen_context` (e.g., build for x86). 

The above CI failure is different but when I saw it, I still recognized my mistakes. So CI was a good [teddy bear](https://talkaboutquality.wordpress.com/2010/08/30/tell-it-to-your-teddy-bear/) here.

I'll open another PR with only the second commit and a rephrased comment.

PR #936, opened 2021-05-03
Fix gen_context/ASM build on ARM

real-or-random:
Obsoletes #935.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -101,7 +101,7 @@ if VALGRIND_ENABLED
 tests_CPPFLAGS += -DVALGRIND
 noinst_PROGRAMS += valgrind_ctime_test
 valgrind_ctime_test_SOURCES = src/valgrind_ctime_test.c
-valgrind_ctime_test_LDADD = libsecp256k1.la $(SECP_LIBS) $(SECP_LIBS) $(COMMON_LIB)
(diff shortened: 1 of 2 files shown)


gmaxwell:
I will test now.

gmaxwell:
ACK c8483520c9077905a1dc8b9adb88b6ea2a3bd9ef   looks good and works here. Undefign is kinda yuck, but it is already doing it and it's cleaner than the obvious alternatives.

sipa:
utACK c8483520c9077905a1dc8b9adb88b6ea2a3bd9ef. I verified that building still works on ARM64, but without asm of course.

Mental note for something to add: there is no sanity checking in ./configure when --with-asm=arm is selected, so it "succeeds" on ARM64, but compilation fails of course.

real-or-random:
> Mental note for something to add: there is no sanity checking in ./configure when --with-asm=arm is selected, so it "succeeds" on ARM64, but compilation fails of course.

In generally, this makes sense because `auto` is always the clever setting and setting a value explicitly just forces this value. Maybe we should rename simply rename the setting to `arm32`?

sipa:
@real-or-random Well, I think even when forcing we can check whether the build environment we have supports that option (we do for x86_64). Renaming to arm32 makes sense.

PR #937, opened 2021-05-04
Have ge_set_gej_var, gej_double_var and ge_set_all_gej_var initialize all fields of their outputs.

roconnor-blockstream:
Previous behaviour would not initialize `r->x` and `r->y` values in the case where infinity is passed in.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -3101,14 +3101,17 @@ void test_ge(void) {
 
     /* Test batch gej -> ge conversion with many infinities. */
     for (i = 0; i < 4 * runs + 1; i++) {
+        int odd;
         random_group_element_test(&ge[i]);
+        odd = secp256k1_fe_is_odd(&ge[i].x);
+        CHECK(odd == 0 || odd == 1);
         /* randomly set half the points to infinity */
-        if(secp256k1_fe_is_odd(&ge[i].x)) {
+        if (odd == i % 2) {
             secp256k1_ge_set_infinity(&ge[i]);
         }
         secp256k1_gej_set_ge(&gej[i], &ge[i]);
     }
-    /* batch invert */
+    /* batch convert */
     secp256k1_ge_set_all_gej_var(ge, gej, 4 * runs + 1);
     /* check result */
     for (i = 0; i < 4 * runs + 1; i++) {
--- a/src/tests.c
+++ b/src/tests.c
@@ -3115,6 +3115,17 @@ void test_ge(void) {
         ge_equals_gej(&ge[i], &gej[i]);
     }
 
+    /* Test batch gej -> ge conversion with all infinities. */
+    for (i = 0; i < 4 * runs + 1; i++) {
+        secp256k1_gej_set_infinity(&gej[i]);
+    }
+    /* batch convert */
+    secp256k1_ge_set_all_gej_var(ge, gej, 4 * runs + 1);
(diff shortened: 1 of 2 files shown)


gmaxwell:
utACK dd6c3de322740a3054cf6a1994a38dc8f201b473

real-or-random:
Concept ACK

Unless I'm mistaken, I think `VG_UNDEF`s here would be helpful. See https://github.com/bitcoin-core/secp256k1/pull/889#issuecomment-771386352.

gmaxwell:
This function exits without leaving anything uninitialized, so nothing to vg_undef here.   Aside, in the past I hadn't UNDEFed anywhere in the codebase itself, only in the tests to avoid injecting vg instrumentation in the production binary.. but we do that now for the CT tests and I think were were satisfied it was benign enough.

real-or-random:
> edit: oh I see, the set infinity could be undefing x/y.  I suppose it could be.

Right, that's what I had in mind. 

> But that should be a separate PR if it gets done.

Fair point.

real-or-random (approved):
utACK dd6c3de322740a3054cf6a1994a38dc8f201b473

gmaxwell:
utACK 45b6468d7e3ed9849ed474c71e9a9479de1a77db

real-or-random:
@roconnor-blockstream You could cherry pick the top two commits from https://github.com/real-or-random/secp256k1/commits/20210504_ge_set_gej_var

real-or-random:
Ah damn, this was not the latest version on my branch -.- Let me push the right one. Sorry.

real-or-random:
Now my branch should have the right commits... I somehow had lost them. (The reflog is nice sometimes. :))

gmaxwell:
ACK 14c9739a1fb485bb56dbe3447132a37bcbef4e22

src/tests.c
```diff
@@ -3101,14 +3101,17 @@ void test_ge(void) {
 
     /* Test batch gej -> ge conversion with many infinities. */
     for (i = 0; i < 4 * runs + 1; i++) {
+        int odd;
         random_group_element_test(&ge[i]);
+        odd = secp256k1_fe_is_odd(&ge[i].x);
+        CHECK(odd == 0 || odd == 1);
         /* randomly set half the points to infinity */
-        if(secp256k1_fe_is_odd(&ge[i].x)) {
+        if (odd == i % 2) {
```
sipa, line 3109: Nit: `% 2` seems superfluous here.

sipa:
utACK 14c9739a1fb485bb56dbe3447132a37bcbef4e22

src/tests.c
real-or-random: Ah I guess it depends on the thing you want to test. 

Before the change, the input consisted of some infinity points and some affine points *with even x coordinate*.
After the change, the input consists of some infinity points and some (arbitrary) affine points.

sipa: Well the line before is `CHECK(odd == 0 || odd == 1);`. The `% 2` is guaranteed to not have any effect.

sipa: Nevermind, I was just misreading this line as `if (odd % 2) {`.

real-or-random (approved):
ACK 14c9739a1fb485bb56dbe3447132a37bcbef4e22

PR #938, opened 2021-05-06
ci: Set -Werror for all CI builds

real-or-random:
Let's see if that passes...

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,21 +1,27 @@
 env:
-  WIDEMUL: auto
+  ### compiler options
+  HOST:
+  # Specific warnings can be disabled with -Wno-error=foo.
+  # -pedantic-errors is not equivalent to -Werror=pedantic and thus not implied by -Werror according to the GCC manual.
+  WERROR_CFLAGS: -Werror -pedantic-errors 
+  MAKEFLAGS: -j2
+  EXTRAFLAGS:
+  BUILD: check
+  ### secp256k1 config
   STATICPRECOMPUTATION: yes
   ECMULTGENPRECISION: auto
   ASM: no
-  BUILD: check
+  WIDEMUL: auto
   WITH_VALGRIND: yes
   RUN_VALGRIND: no
-  EXTRAFLAGS:
-  HOST:
-  ECDH: no
-  RECOVERY: no
-  SCHNORRSIG: no
-  EXPERIMENTAL: no
   CTIMETEST: yes
   BENCH: yes
   ITERS: 2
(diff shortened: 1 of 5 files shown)


real-or-random:
Uff ok, this is failing for multiple reasons:
 - [ ] setting `CFLAGS=-Werror` during `configure` is a bad idea because then all the compilation checks run by `configure` use this and produce wrong results (these are the macOS and x86 asm failures) 
 - [ ] the mingw libtool build produces warnings that are not our responsibility, see #923...

Fixing the first item is a little bit annoying. We could simply patch the generated `Makefile` but that's pretty ugly. So we'll better add some logic.

It would be nice to call `./configure` normally but then `make CFLAGS=-Werror` but this would override all the other settings in `CFLAGS`. According to https://www.gnu.org/software/automake/manual/html_node/Flag-Variables-Ordering.html, it would anyway be cleaner to set our `-Wall` etc in `AM_CFLAGS` but even then `CFLAGS=-Werror` would override `-O2` for example.

I believe a better way would be to have some maintainer variable with additional flags that can be set from the outside like `make EXTRACFLAGS=-Werror`. Or have some `--enable-werror` or `--enable-maintainer-mode` for `configure`. I'll probably try one of those options. Happy to hear opinions.

gmaxwell:
An ```enable-werror``` is the sort of thing I'd expect (maintainer mode sounds like it would get confused with AM_MAINTAINER_MODE).

The only concern I have with EXTRACFLAGS is just that it gets into my environment and then messes up my tests but is weird enough that I don't think to look for it. :P  Is there any particular reason it needs to rerun both ways such that rerunning configure is a concern?

However its setup it should be done in a way that makes it relatively easy to drop on some CI builds. Because if some build starts throwing false positives it may be best to ignore it (e.g. if it's S390 or some other weird config and silencing it requires hurting performance or crapping up the codebase), or maybe not, but the decision shouldn't be driven by having to all-or-nothing the Werror or dropping the build.

real-or-random:
> The only concern I have with EXTRACFLAGS is just that it gets into my environment and then messes up my tests but is weird enough that I don't think to look for it.

I ended up doing this (naming it `WERROR_CFLAGS`) because we anyway need some free text variable for specifying fine-grained `-Wno-error` flags, and having a user-facing configure option is not tremendously helpful anyway. If users want `-Werror` they can always set `CFLAGS`.

Let's see what Cirrus says.

real-or-random:
Closed in favor of https://github.com/bitcoin-core/secp256k1/pull/944

PR #939, opened 2021-05-06
Recommend VPATH (out-of-tree) builds in README

real-or-random:
These VPATH builds are supported natively by autotools, so we can
just rely on them.

This also removes the leftover `obj` directory and provides an empty `build`
directory instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,7 +14,6 @@ valgrind_ctime_test
 *.so
 *.a
 !.gitignore
-
 Makefile
 configure
 build/**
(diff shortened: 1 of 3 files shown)


real-or-random:
I think it's always a good idea to keep the tree as clean as possible. But that's my personal preference and yeah I agree, I guess we shouldn't recommend it to users if it's too uncommon and makes testing more complicated.

> Maybe then they'll open PRs adding obj to .gitignore, but in any case they'll be using something different from what gets tested.

Not sure what you're referring to. If I'm not mistaken, the `obj` directory is currently unused, we should remove it in any case.

gmaxwell:
Ah, the project uses subdir-objects so indeed obj is unused.  Sorry, I was confused by the fact that this PR removed it, as I assumed they it did so because they'd end up under the build directory. :)

real-or-random:
Closing in favor of #941.

PR #941, opened 2021-05-12
Clean up git tree

real-or-random:
This removes the ununsed `obj` directory. It also suggests in the README
to create the "coverage" files in a separate directory and adds the
coverage files to .gitignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,14 @@ libtool
 *~
 *.log
 *.trs
+
+coverage/
+coverage.html
+coverage.*.html
+*.gcda
+*.gcno
+*.gcov
+
 src/libsecp256k1-config.h
 src/libsecp256k1-config.h.in
 src/ecmult_static_context.h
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -33,6 +33,13 @@ libtool
 *~
 *.log
 *.trs
+
+coverage/
```
jonasnick, line 37: What do you do to get this? I just write gcovr output into a coverage.html. By the way, could you also add *.gcov?

real-or-random: Oh yes, this was left over from the earlier PR. And yes, I'll add .gcov.

real-or-random: Ok, I ported that one change that suggests to create coverage reports in a `coverage` subdir from #939 over to here. I think that one is still useful for users.

jonasnick (commented):
ACK 09b3bb8648fec903e4ac2ec1d047503d5f0f48d7

PR #942, opened 2021-05-12
Verify that secp256k1_ge_set_gej_zinv does not operate on infinity.

roconnor-blockstream:
a->x and a->y should not be used if the infinity flag is set.

### DIFF
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -67,6 +67,7 @@ static const secp256k1_fe secp256k1_fe_const_b = SECP256K1_FE_CONST(0, 0, 0, 0, 
 static void secp256k1_ge_set_gej_zinv(secp256k1_ge *r, const secp256k1_gej *a, const secp256k1_fe *zi) {
     secp256k1_fe zi2;
     secp256k1_fe zi3;
+    VERIFY_CHECK(!a->infinity);
     secp256k1_fe_sqr(&zi2, zi);
     secp256k1_fe_mul(&zi3, &zi2, zi);
     secp256k1_fe_mul(&r->x, &a->x, &zi2);
--- a/src/ecmult_const.h
+++ b/src/ecmult_const.h
@@ -14,6 +14,7 @@
  * Multiply: R = q*A (in constant-time)
  * Here `bits` should be set to the maximum bitlength of the _absolute value_ of `q`, plus
  * one because we internally sometimes add 2 to the number during the WNAF conversion.
+ * A must not be infinity.
  */
 static void secp256k1_ecmult_const(secp256k1_gej *r, const secp256k1_ge *a, const secp256k1_scalar *q, int bits);
 
--- a/src/ecmult_const_impl.h
+++ b/src/ecmult_const_impl.h
@@ -168,6 +168,7 @@ static void secp256k1_ecmult_const(secp256k1_gej *r, const secp256k1_ge *a, cons
      * that the Z coordinate was 1, use affine addition formulae, and correct
      * the Z coordinate of the result once at the end.
      */
+    VERIFY_CHECK(!a->infinity);
     secp256k1_gej_set_ge(r, a);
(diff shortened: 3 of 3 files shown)


roconnor-blockstream:
There is a minor question as to whether the last line should be replace with `r->infinity = 0;`

real-or-random (approved):
utACK https://github.com/bitcoin-core/secp256k1/commit/32338175cf1c1dca0b1bca7f2e1974297fc9e4d0

Setting the infinity flag: I think it's okay to keep this line. It errs on the safe side.

roconnor-blockstream:
rebased.

real-or-random:
> To this end I want to get to a state where `fe_clear` will set the memory to uninitialized for the purposes of valgrind, and use `set_int(...,0)` for those cases were were explicitly want the value to be 0.

This will be https://github.com/bitcoin-core/secp256k1/pull/636/commits/b33a8e49e89f5a1ea02a9b9b9b208cb4f3d59e44 (and maybe the following commit in that PR for other modules).

I can take these out of #636 after this PR.

src/group_impl.h
```diff
@@ -67,6 +67,7 @@ static const secp256k1_fe secp256k1_fe_const_b = SECP256K1_FE_CONST(0, 0, 0, 0,
 static void secp256k1_ge_set_gej_zinv(secp256k1_ge *r, const secp256k1_gej *a, const secp256k1_fe *zi) {
     secp256k1_fe zi2;
     secp256k1_fe zi3;
+    VERIFY_CHECK(!a->infinity);
```
robot-dreams, line 70: Do we need to account for the public-facing `secp256k1_ecmult_const`, where the user can possibly pass `a = infinity` and cause this check to fail?

https://github.com/bitcoin-core/secp256k1/blob/master/src/ecmult_const_impl.h#L172
https://github.com/bitcoin-core/secp256k1/blob/master/src/ecmult_impl.h#L113
https://github.com/bitcoin-core/secp256k1/blob/master/src/ecmult_impl.h#L81

roconnor-blockstream: `secp256k1_ecmult_odd_multiples_table` already has a VERIFY_CHECK for infinity, if I'm reading this correctly.

`src/modules/ecdh/main_impl.h` seems to be the only place where `secp256k1_ecmult_const` is used, and it is only used with a loaded public key, which is never infinity.

I'm somewhat skeptical that the table building function will work when given infinity.

Perhaps the documentation for `secp256k1_ecmult_const` should be updated to explicitly state that the input point cannot be infinity, and maybe a VERIFY_CHECK added there too.

robot-dreams: Oops, you're right, the failed check (which I triggered by calling `secp256k1_ecmult_const` with `a = infinity`) probably came from this path instead:

https://github.com/bitcoin-core/secp256k1/blob/master/src/ecmult_impl.h#L115
https://github.com/bitcoin-core/secp256k1/blob/master/src/group_impl.h#L163

It looks like all other calls (in `secp256k1_ecmult_odd_multiples_table` and `secp256k1_ge_set_all_gej_var`) are protected by a check `!a.infinity`.

robot-dreams: Adding a comment and a VERIFY_CHECK to `secp256k1_ecmult_const` sounds good to me, I will ACK after that.

roconnor-blockstream: Done.

robot-dreams:
ACK 099bad945e9a7c5237cdd764eca420285a9de279

real-or-random:
ACK 099bad945e9a7c5237cdd764eca420285a9de279 I inspected all call sites, they all ensure that a is not infinity

By the way, `secp256k1_ecmult_odd_multiples_table_globalz_windowa` is in the wrong module.  This function had been used within ecmult in the past but since 8c1c831bdb083dfa8b50fac16a0e17a7e1df4064, it is only used in ecmult_const. This also means its comment is wrong. We can fix this in another PR.

Issue #929, opened 2021-05-02
Tracking issue for non-autotool builds


real-or-random:
We started with 

```
gcc -Wall -Wextra -Wno-unused-function -O2 -I. -c ./src/secp256k1.c -DSECP256K1_BUILD -D ECMULT_GEN_PREC_BITS=4 -D ECMULT_WINDOW_SIZE=15
```

The goal is to get rid of most of the command line arguments, at least on GCC and clang:

 - [x] `-DSECP256K1_BUILD`: #928
 - [ ] `-I` #925 
 - [ ] `-D ECMULT_GEN_PREC_BITS=4 -D ECMULT_WINDOW_SIZE=15`: See discussion in #918. We could even print the config using `#pragma message`)
 - [ ] `-Wno-unused-function`: We could add a `#pragma` or better add `__attribute__ ((unsused))` to the relevant functions.
 - [ ] `-Wall -Wextra`: We could set this via `#pragmas` but maybe we should leave this to the user.
 - [ ] `-O2`: I think we should really leave this to the user.
 - [ ] Add section to README
 - [ ] Add CI jobs to test these builds

gmaxwell:
>  -Wno-unused-function: We could add a #pragma or better add __attribute__ ((unsused)) to the relevant functions.

There are no unused functions except for ones that are unneeded because of test harness or module configuration options.  I think they should be ifdefed out when their relevant modules are disabled.  This is ever so slightly less clean because it breaks module isolation but it has a big win that when the code changes to make them actually unused we'll get warnings, and I think that is well worth it.

>  -Wall -Wextra: We could set this via #pragmas but maybe we should leave this to the user.

Yeah leave that to the user. :P  Also -pedantic -std=c89 ... but the CI should be testing that way and probably with -Werror at least in one config since the CI's compiler is stable (never subject the general public to Werror, since god knows what their compiler will warn on--- but it's reasonable to expect development to be always warning clean on a single stable target compiler.)

One missing thing here is the loss of assembly optimizations.  Because arm requires an external assembler and 64-bit windows doesn't support inline asm at all, this is kinda unfortunate.   The performance loss on 32-bit arm from losing access to assembly is phenomenal. It's kinda dumb that joe-hardware-wallet-dev is going to almost double his signing time (schnorrsig_sign 872us vs 472us on a 1.2GHz arm7) just due to using a naive build process, and keep in mind HWW are much lower clockrate devices.

real-or-random:
[...]

Issue #896, opened 2021-02-17
build: configure script warnings on macOS


hebasto:
Configuring the recent master branch (24d1656c328c932cd533f6f4bf80b345b6591f0d) on macOS Big Sur 11.2.1 (20D74), and receiving 4 messages `rm: conftest.dSYM: is a directory`:

```
% ./configure -q
rm: conftest.dSYM: is a directory
rm: conftest.dSYM: is a directory
rm: conftest.dSYM: is a directory
rm: conftest.dSYM: is a directory

Build Options:
  with ecmult precomp     = yes
  with external callbacks = no
  with benchmarks         = yes
  with tests              = yes
  with openssl tests      = yes
  with coverage           = no
  module ecdh             = no
  module recovery         = no
  module extrakeys        = no
  module schnorrsig       = no

  asm                     = x86_64
  bignum                  = no
  ecmult window size      = 15
  ecmult gen prec. bits   = 4

  valgrind                = no
  CC                      = gcc
  CFLAGS                  = -O2 -fvisibility=hidden -std=c89 -pedantic -Wall -Wextra -Wcast-align -Wnested-externs -Wshadow -Wstrict-prototypes -Wundef -Wno-unused-function -Wno-long-long -Wno-overlength-strings -W -g
  CPPFLAGS                = -I/usr/local/Cellar/valgrind/3.16.1/include 
  LDFLAGS                 = 

  CC_FOR_BUILD            = gcc
  CFLAGS_FOR_BUILD        = -O2 -fvisibility=hidden -std=c89 -pedantic -Wall -Wextra -Wcast-align -Wnested-externs -Wshadow -Wstrict-prototypes -Wundef -Wno-unused-function -Wno-long-long -Wno-overlength-strings -W -g
  CPPFLAGS_FOR_BUILD      = -I/usr/local/Cellar/valgrind/3.16.1/include 
  LDFLAGS_FOR_BUILD       =
```

Detailed:
```
% ./configure
...
checking whether gcc understands -c and -o together... rm: conftest.dSYM: is a directory
yes
...
checking the name lister (/usr/bin/nm -B) interface... rm: conftest.dSYM: is a directory
BSD nm
...
checking for archiver @FILE support... rm: conftest.dSYM: is a directory
no
...
checking command to parse /usr/bin/nm -B output from gcc object... rm: conftest.dSYM: is a directory
ok
...
```

real-or-random:
[...]

PR #944, opened 2021-05-13
Various improvements related to CFLAGS

real-or-random:
- Fixes strange warnings in mingw builds (one item in #923).
 - Enables more and more fine-grained compiler warnings.
 - Fixes minor build issue on macOS  #896 on macOS.
 - Makes compiler warnings into errors on CI (`-Werror`, mentioned in https://github.com/bitcoin-core/secp256k1/issues/929#issuecomment-830861253).

### DIFF
--- a/build-aux/m4/bitcoin_secp.m4
+++ b/build-aux/m4/bitcoin_secp.m4
@@ -82,3 +82,19 @@ if test x"$has_valgrind" != x"yes"; then
   AC_CHECK_HEADER([valgrind/memcheck.h], [has_valgrind=yes; AC_DEFINE(HAVE_VALGRIND,1,[Define this symbol if valgrind is installed])])
 fi
 ])
+
+dnl SECP_TRY_APPEND_CFLAGS(flags, VAR)
+dnl Append flags to VAR if CC accepts them.
+AC_DEFUN([SECP_TRY_APPEND_CFLAGS], [
+  AC_MSG_CHECKING([if ${CC} supports $1])
+  SECP_TRY_APPEND_CFLAGS_saved_CFLAGS="$CFLAGS"
+  CFLAGS="$1 $CFLAGS"
+  AC_COMPILE_IFELSE([AC_LANG_SOURCE([[char foo;]])], [flag_works=yes], [flag_works=no])
+  AC_MSG_RESULT($flag_works)
+  CFLAGS="$SECP_TRY_APPEND_CFLAGS_saved_CFLAGS"
+  if test x"$flag_works" = x"yes"; then
+    $2="$$2 $1"
+  fi
+  unset flag_works
+  AC_SUBST($2)
+])
--- a/configure.ac
+++ b/configure.ac
@@ -8,10 +8,6 @@ AH_TOP([#define LIBSECP256K1_CONFIG_H])
 AH_BOTTOM([#endif /*LIBSECP256K1_CONFIG_H*/])
 AM_INIT_AUTOMAKE([foreign subdir-objects])
 
-# Set -g if CFLAGS are not already set, which matches the default autoconf
-# behavior (see PROG_CC in the Autoconf manual) with the exception that we don't
-# set -O2 here because we set it in any case (see further down).
-: ${CFLAGS="-g"}
 LT_INIT
 
 # Make the compilation flags quiet unless V=1 is used.
@@ -84,7 +80,6 @@ esac
 # These executables are compiled from C source file for which our flags may not be appropriate,
 # e.g., -std=c89 flag has lead to undesirable warnings in the past.
 #
-# TODO We still touch the CFLAGS for --coverage and -O0/-O2.
 # TODO We should analogously not touch CPPFLAGS and LDFLAGS but currently there are no issues.
 AC_DEFUN([SECP_TRY_APPEND_DEFAULT_CFLAGS], [
     # Try to append -Werror=unknown-warning-option to CFLAGS temporarily. Otherwise clang will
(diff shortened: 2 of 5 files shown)


real-or-random:
CI seems to be confused here because this is the same source branch as on #938 . Trying to close and reopen the PR to fix it.

real-or-random:
rebased

roconnor-blockstream:
LGTM. I did test building in under Nixos, but that was the extent of my testing.

build-aux/m4/bitcoin_secp.m4
```diff
@@ -82,3 +82,19 @@ if test x"$has_valgrind" != x"yes"; then
   AC_CHECK_HEADER([valgrind/memcheck.h], [has_valgrind=yes; AC_DEFINE(HAVE_VALGRIND,1,[Define this symbol if valgrind is installed])])
 fi
 ])
+
+dnl SECP_TRY_APPEND_CFLAGS([flags], VAR)
```
hebasto, line 86: From my understanding, the square brackets `[...]` in a comment, that describes a new macro, means that parameter is optional. They do not mean quoting. Right?

configure.ac
```diff
@@ -70,35 +66,40 @@ case $host_os in
    ;;
 esac
 
-CFLAGS="-W $CFLAGS"
```
hebasto, line 73: Maybe document the removing of `-W` option?

hebasto (commented):
Approach ACK 64ec678daaf215e007d44433cdc9b969b6e09098, tested on:
- Linux Mint 20.1 (x86_64)
- macOS 11.4 (20F71)

Warnings [introduced](https://github.com/bitcoin-core/secp256k1/pull/700#discussion_r577840652) in #700 (reported in #896) are gone.

The last two commits seems the one to me, maybe squash them?
But maybe another approach could be considered?

Why a CI is not a "user" from the Automake's point of view? If so, why to introduce the `WERROR_CFLAGS` variable, while we have a dedicated `CFLAGS` one?

real-or-random:
> Why a CI is not a "user" from the Automake's point of view? If so, why to introduce the `WERROR_CFLAGS` variable, while we have a dedicated `CFLAGS` one?

It may seem overly complicated but I think it's better for testing/CI because it does not interfere with autoconf's default for `CFLAGS` (which is `-g -O2` or just `-g` depending on the compiler). What we typically want to test in CI is building with unset `CFLAGS`. We could always set `-Werror -g -O2` instead of `-Werror` but that seems complicated too.

By the way, I think I overlooked this line: https://github.com/bitcoin-core/secp256k1/pull/944/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R397

Will address the other comments.

hebasto:
> > Why a CI is not a "user" from the Automake's point of view? If so, why to introduce the `WERROR_CFLAGS` variable, while we have a dedicated `CFLAGS` one?
> 
> It may seem overly complicated but I think it's better for testing/CI because it does not interfere with autoconf's default for `CFLAGS` (which is `-g -O2` or just `-g` depending on the compiler). What we typically want to test in CI is building with unset `CFLAGS`. We could always set `-Werror -g -O2` instead of `-Werror` but that seems complicated too.

Ah, indeed. This repo does not have the `--enable-werror` configure option as Bitcoin Core has :)

real-or-random:
Oh interesting, maybe I should just have looked at Core then... ;) 

For context:
https://github.com/bitcoin-core/secp256k1/pull/938#issuecomment-839632758

Anyway, this thing here works and I don't want to touch it further unless necessary.

configure.ac
real-or-random: Apparently I've documented it here already https://github.com/bitcoin-core/secp256k1/pull/944/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R95

hebasto: Ah, right.

build-aux/m4/bitcoin_secp.m4
real-or-random: fixed

real-or-random:
Addressed the comments. I also rebased on master (sorry, that wasn't even intentional...)

> The last two commits seems the one to me, maybe squash them?

Done.

hebasto:
ACK a5c6ff2917edf97e888c441eb545ac70083b5376

configure.ac
```diff
@@ -70,35 +66,40 @@ case $host_os in
...
+    # no matter if clang knows the flag or not.
+    SECP_TRY_APPEND_DEFAULT_CFLAGS_saved_CFLAGS="$CFLAGS"
+    SECP_TRY_APPEND_CFLAGS([-Werror=unknown-warning-option], CFLAGS)
+    
+    SECP_TRY_APPEND_CFLAGS([-std=c89 -pedantic -Wno-long-long -Wnested-externs -Wshadow -Wstrict-prototypes -Wundef], $1) # GCC >= 3.0, -Wlong-long is implied by -pedantic.
+    SECP_TRY_APPEND_CFLAGS([-Wno-overlength-strings], $1) # GCC >= 4.2, -Woverlength-strings is implied by -pedantic.
+    SECP_TRY_APPEND_CFLAGS([-Wall], $1) # GCC >= 2.95 and probably many other compilers
+    SECP_TRY_APPEND_CFLAGS([-Wno-unused-function], $1) # GCC >= 3.0, -Wunused-function is implied by -Wall.
+    SECP_TRY_APPEND_CFLAGS([-Wextra], $1) # GCC >= 3.4, this is the newer name of -W, which we don't use because older GCCs will warn about unused functions.
+    SECP_TRY_APPEND_CFLAGS([-Wcast-align=strict], $1) # GCC >= 8.0
```
jonasnick, line 96: clang doesn't have `-Wcast-align=strict` but only `-Wcast-align`. Since this PR replaces the latter with the former, when using clang there are no alignment warnings anymore.

real-or-random:
fixed

jonasnick (commented):
ACK 0302138f7508414e9e5212bc45b4ca4c0e5f081c

PR #951, opened 2021-06-14
configure: replace AC_PATH_PROG to AC_CHECK_PROG

UdjinM6:
~When libsecp256k1 is built as an external package for some another program it might be important to respect parent's `--prefix`. AC_CHECK_PROG does this correctly while AC_PATH_PROG finds system-wide installed binaries instead.~ See correct description of the issue in the comment below.

A patch like this applied to Bitcoin Core repo directly fixes `make check` for Bitcoin Core 0.19 and older on macOS. openssl was removed from dependencies in v0.20 so `make check` no longer fails on newer versions but it makes sense to fix libsecp256k1 anyway probably.

Example `make check` failure for Bitcoin Core 0.19:
```
Undefined symbols for architecture x86_64:
  "_BN_is_negative", referenced from:
      _test_ecdsa_der_parse in tests-tests.o
  "_ECDSA_SIG_get0", referenced from:
      _test_ecdsa_der_parse in tests-tests.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

Example configure logs for Bitcoin Core 0.19 with `--prefix`
(I also exposed `$enable_openssl_tests` like it's done in master to make it easier to compare)

Before:
```
...
checking for brew... no
...
=== configuring in src/secp256k1 (/path/to/bitcoin/src/secp256k1)
...
checking for brew... /usr/local/bin/brew
...
checking for libcrypto... yes
checking for EC functions in libcrypto... yes
...
Build Options:
...
  with openssl tests  = auto
...
```

After:
```
...
checking for brew... no
...
=== configuring in src/secp256k1 (/path/to/bitcoin/src/secp256k1)
...
checking for brew... no
checking for port... no
...
checking for libcrypto... yes
checking for EC functions in libcrypto... yes
...
Build Options:
...
  with openssl tests  = auto
...
```

Example configure logs for Bitcoin Core 0.21.99 (master) with `--prefix`

Before:
```
...
checking for brew... no
...
=== configuring in src/secp256k1 (/path/to/bitcoin/src/secp256k1)
...
checking for brew... /usr/local/bin/brew
...
checking for libcrypto... yes
checking for EC functions in libcrypto... yes
...
Build Options:
...
  with openssl tests      = yes
...
```

After:
```
...
checking for brew... no
...
=== configuring in src/secp256k1 (/path/to/bitcoin/src/secp256k1)
...
checking for brew... no
checking for port... no
...
checking for libcrypto... no
checking for openssl/crypto.h... no
...
Build Options:
...
  with openssl tests      = no
...
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -42,8 +42,8 @@ AM_PROG_AS
 case $host_os in
   *darwin*)
      if  test x$cross_compiling != xyes; then
-       AC_PATH_PROG([BREW],brew,)
-       if test x$BREW != x; then
+       AC_CHECK_PROG([BREW], brew, brew)
+       if test x$BREW = xbrew; then
          # These Homebrew packages may be keg-only, meaning that they won't be found
          # in expected paths because they may conflict with system files. Ask
          # Homebrew where each one is located, then adjust paths accordingly.
@@ -58,10 +58,10 @@ case $host_os in
            VALGRIND_CPPFLAGS="-I$valgrind_prefix/include"
          fi
        else
-         AC_PATH_PROG([PORT],port,)
+         AC_CHECK_PROG([PORT], port, port)
          # If homebrew isn't installed and macports is, add the macports default paths
          # as a last resort.
-         if test x$PORT != x; then
+         if test x$PORT = xport; then
            CPPFLAGS="$CPPFLAGS -isystem /opt/local/include"
            LDFLAGS="$LDFLAGS -L/opt/local/lib"
          fi


real-or-random:
Hm, I'd like to understand what's going on here. According to the [autoconf docs](https://www.gnu.org/software/autoconf/manual/autoconf-2.70/html_node/Generic-Programs.html), the two macros shouldn't differ in this respect, they both should search your PATH. And when looking at [their implementation](https://github.com/autotools-mirror/autoconf/blob/6d38e9fa2b39b3c3a8e4d6d7da38c59909d3f39d/lib/autoconf/programs.m4#L37), I don't see how they would differ...  

I guess in the end it makes sense to be consistent with Core. But I wonder why Core uses `AC_PATH_PROG` everywhere except when checking for brew, where it uses `AC_PATH_PROG`.


Is this maybe a caching issue? So after your PR, our config just picks up the cached value and that's why the results of the checks are consistent with Core? 

As far as I understand, any `--prefix` should be correct passed down here?
https://github.com/bitcoin/bitcoin/blob/master/configure.ac#L1918

Can you attach full `configure.log` files maybe?

UdjinM6:
@real-or-random thanks for reviewing! I dug a bit deeper now, see below.

> Hm, I'd like to understand what's going on here. According to the [autoconf docs](https://www.gnu.org/software/autoconf/manual/autoconf-2.70/html_node/Generic-Programs.html), the two macros shouldn't differ in this respect, they both should search your PATH. And when looking at [their implementation](https://github.com/autotools-mirror/autoconf/blob/6d38e9fa2b39b3c3a8e4d6d7da38c59909d3f39d/lib/autoconf/programs.m4#L37), I don't see how they would differ...
> 
> I guess in the end it makes sense to be consistent with Core. But I wonder why Core uses `AC_PATH_PROG` everywhere except when checking for brew, where it uses `AC_PATH_PROG`.

The difference is https://github.com/autotools-mirror/autoconf/blob/6d38e9fa2b39b3c3a8e4d6d7da38c59909d3f39d/lib/autoconf/programs.m4#L47 vs https://github.com/autotools-mirror/autoconf/blob/6d38e9fa2b39b3c3a8e4d6d7da38c59909d3f39d/lib/autoconf/programs.m4#L128 I think i.e. `AC_CHECK_PROG` overrides it with whatever value the user provides while `AC_PATH_PROG` only does this for paths. And Bitcoin Core has this https://github.com/bitcoin/bitcoin/blob/master/depends/config.site.in#L73-L76 so `AC_PATH_PROG` just ignores it I think.

> Is this maybe a caching issue? So after your PR, our config just picks up the cached value and that's why the results of the checks are consistent with Core?

Definitely not a caching issue imo https://github.com/autotools-mirror/autoconf/blob/6d38e9fa2b39b3c3a8e4d6d7da38c59909d3f39d/lib/autoconf/programs.m4#L83-L84 and https://github.com/autotools-mirror/autoconf/blob/6d38e9fa2b39b3c3a8e4d6d7da38c59909d3f39d/lib/autoconf/programs.m4#L139-L140.

> As far as I understand, any `--prefix` should be correct passed down here?
> https://github.com/bitcoin/bitcoin/blob/master/configure.ac#L1918

 I couldn't figure out what's going on _exactly_ initially tbh, I just noticed that going switching from `AC_PATH_PROG` to `AC_CHECK_PROG` fixed the issue and made things look much more consistent. And since it only happens when you specify `--prefix` I thought that it was somehow connected to the issue but it's not the case it seems.

real-or-random:
Urghs, thanks, I see. I think your analysis is not entirely correct though. The `BREW=no` line is in the depends builds (for dependencies (https://github.com/bitcoin/bitcoin/blob/master/depends/config.site.in#L73-L76) but secp256k1 is built as part of the real build. But `BREW` is set in the real build too.

Let me explain this in my own words, and feel free to use this (or a modified version) as an updated commit message:

---

Bitcoin Core's `configure` script uses `AC_CHECK_PROG` to find brew in the `PATH` [1]. If found, this macro will set `BREW=brew`. Ideally, secp256k1's `configure` script would simply pick this preset value of the `BREW` variable up instead of performing its own search. But this is not what happens because secp256k1's  `configure` uses `AC_PATH_PROG` instead which respects preset variable values (in this case for variable `BREW`) only if they are a valid path (i.e., they match `[\\/*] | ?:[\\/]*` [2]), and `brew` is not a path.

This commit changes `AC_PATH_PROG` to `AC_CHECK_PROG` to be consistent with Core's `AC_CHECK_PROG`. Both of these macros are supposed to find executables in the `PATH` but the difference is that former is supposed to return the full path whereas the latter is supposed to find only the program. As a result, the latter will accept even non-paths like `brew` as an override. Not knowing the full path is not an issue for the `configure` script because it will only execute `BREW` immediately afterwards, which works fine without the full path. (In particular, `PATH` cannot have changed in between [3].)

[1] https://github.com/bitcoin/bitcoin/blob/master/configure.ac#L684
[2] https://github.com/autotools-mirror/autoconf/blob/6d38e9fa2b39b3c3a8e4d6d7da38c59909d3f39d/lib/autoconf/programs.m4#L127
[3] https://github.com/bitcoin-core/secp256k1/commit/3ab1178d54029745219d67e6c305df4d7564e278

---

So assuming my analysis is correct (maybe you can test/verify, e.g., setting `BREW=/path/to/brew` or `BREW=/path/to/nonexisting/file` from the outside), this PR looks good to me.

To be honest, the behavior of autoconf is a little strange here. Say you would like to override some tool like brew to `my-own-brew`, which in your `PATH`. You think you could set `BREW=my-own-brew` (like you can set `CC=clang`) but this won't work if the config script uses `AC_PATH_PROG`...

UdjinM6:
> Urghs, thanks, I see. I think your analysis is not entirely correct though. The `BREW=no` line is in the depends builds (for dependencies (https://github.com/bitcoin/bitcoin/blob/master/depends/config.site.in#L73-L76) but secp256k1 is built as part of the real build. But `BREW` is set in the real build too.

Hmm... But building depends produces `depends/<host_prefix>/share/config.site` which I think is picked by the real build with `--prefix`. And for macOS https://github.com/bitcoin/bitcoin/blob/master/depends/config.site.in#L73-L76 transforms into
```
if test "darwin" = darwin; then
  BREW=no
  PORT=no
fi
```

real-or-random:
Hm lol, okay, I don't know. In the end, it does not matter if it's `BREW=no` or `BREW=brew`, both won't work. Can you maybe just `echo BREW` at the relevant point in secp256k1's configure? Then we see who's right at least. :D

UdjinM6:
I tweaked the commit message you proposed the way I understand the issue (at least the way I do it atm :) ), pls see a4642fa. 

Echoing `$BREW` right before `AC_CHECK_PROG` (`AC_PATH_PROG` before) in secp256k1's configure gives `no` (or whatever value I set in `depends/<host_prefix>/share/config.site`) for the build with `--prefix` and `` (empty) for the normal build.

real-or-random:
Ok, looks good.

ACK a4642fa15ee731b0a620a3f089826d556e5405f0



> Echoing `$BREW` right before `AC_CHECK_PROG` (`AC_PATH_PROG` before) in secp256k1's configure gives `no` (or whatever value I set in `depends/<host_prefix>/share/config.site`) for the build with `--prefix` and `` (empty) for the normal build.

Just for curiosity, what prefix do you set?

UdjinM6:
> maybe you can test/verify, e.g., setting `BREW=/path/to/brew` or `BREW=/path/to/nonexisting/file` from the outside

Setting both `BREW` and `PORT` to `/path/to/nonexisting/file` in `depends/<host_prefix>/share/config.site` (to test things only, normally we'd have to patch https://github.com/bitcoin/bitcoin/blob/master/depends/config.site.in#L73-L76 ofc) also fixes the issue even without this PR it seems. It's a bit weird to see `checking for brew... /path/to/nonexisting/file` in logs but everything else looks good I think (no openssl is found, as expected). If you feel like patching Bitcoin Core is the right way, I can close this one and submit a corresponding PR there instead.

> Just for curiosity, what prefix do you set?

In my case (macOS 10.14.6) it's `/path/to/bitcoin/depends/x86_64-apple-darwin18.7.0/`

real-or-random:
> > maybe you can test/verify, e.g., setting `BREW=/path/to/brew` or `BREW=/path/to/nonexisting/file` from the outside
> 
> Setting both `BREW` and `PORT` to `/path/to/nonexisting/file` in `depends/<host_prefix>/share/config.site` (to test things only, normally we'd have to patch [https://github.com/bitcoin/bitcoin/blob/master/depends/config.site.in#L73-L76](https://github.com/bitcoin/bitcoin/blob/master/depends/config.site.in?rgh-link-date=2021-06-15T17%3A19%3A34Z#L73-L76) ofc) also fixes the issue even without this PR it seems. It's a bit weird to see `checking for brew... /path/to/nonexisting/file` in logs but everything else looks good I think (no openssl is found on master, as expected). If you feel like patching Bitcoin Core is the right way to fix the issue, I can close this one and submit a corresponding PR there instead.

No I think it's totally fine to fix it here, if only to avoid the strange log output.  Autotools is a often somewhat arbitrary collection of things that someone need to work together and then I think in cases like this, there's no clear right place to fix the issue. We just need to fix it somewhere.

jonasnick (commented):
utACK a4642fa15ee731b0a620a3f089826d556e5405f0

Thanks @UdjinM6. Not sure I understand this issue fully (the detailed commit comment certainly helped), but should be good to merge.

PR #956, opened 2021-06-25
Static ecmult ctx 20210625

roconnor-blockstream:
Replace ecmult_context with a static array.


elichai:
Was curious, the binary size on my machine with all modules enabled before this change was: `672K`, in this branch: `1.6M`.

roconnor-blockstream:
That's expected since the `ecmult_context` is now statically allocated instead of dynamically allocated.  You should find that your runtime memory use is down by about the same amount.

elichai:
> That's expected since the `ecmult_context` is now statically allocated instead of dynamically allocated. You should find that your runtime memory use is down by about the same amount.
> 
> I'll also add that you can reduce the binary size by reducing the `ECMULT_WINDOW_SIZE` through the configuration parameter.


Yeah I know :) it wasn't a complaint or anything, just wanted to write it here for reference. less than a MB for simplifying the API and simplifying usage is pretty good

real-or-random:
@roconnor-blockstream Here's a branch with two fixups commits:
 https://github.com/real-or-random/secp256k1/commits/static_ecmult_ctx_20210625 

The first commit simply reverts your changes to the build system. 

The second commit introduces a very simple approach to change the build system: Just have a configure option `--enable-devtools` (default no), which enables building of `gen_pre_g` as a developer tool. The build system will build but never run `gen_pre_g`. This is by far the simplest we can do. Since the precomputed file will be shipped with the tree, only devs need to update the file and can do this manually by running `gen_pre_g`. Building `gen_pre_g` ignores issues of cross-compilation, i.e., if you compile for a foreign target, it will just compile `gen_pre_g` for that foreign target. Who cares, if you need to run `gen_pre_g`, simply don't cross-compile. By doing the same in a later PR for `gen_context` we could get of the entire cumbersome and error-prone machinery of detecting host compilers (`CFLAGS_FOR_BUILD` and friends). 

Nit on naming:
Maybe we could call this `gen_ecmult_pre_g` to make it clear from the name that it corresponds to the `ecmult.h` module. Otherwise people (including me) will constantly think that it is for `ecmult_gen.h` due to the `g` in the name. In the other PR we could then rename `gen_context` to `gen_ecmult_gen_context`. This will make things a little bit clearer. Naming is very confusing at the moment with even more issues, e.g., I claim `enable-ecmult-static-precomputation` should have been called `enable-ecmultgen-static-precomputation`.

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,249 @@
...
+int main(int argc, char **argv) {
+    const secp256k1_ge g = SECP256K1_G;
+    const secp256k1_ge g_13 = SECP256K1_G_ORDER_13;
+    const secp256k1_ge g_199 = SECP256K1_G_ORDER_199;
+    const int window_g_13 = 4;
+    const int window_g_199 = 8;
+    FILE* fp;
+    secp256k1_gej gj;
+    (void)argc;
+    (void)argv;
```
real-or-random, line 200: FWIW, `int main(void)` is valid C and simpler

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,249 @@
...
+void printTable(FILE* fp, const char * name, int window_g, const secp256k1_gej * gj, int withConditionals) {
+    int j;
+    int i;
+
+    secp256k1_ecmult_odd_multiples_table_storage(ECMULT_TABLE_SIZE(window_g), gj);
+    fprintf(fp, "static const secp256k1_ge_storage %s[ECMULT_TABLE_SIZE(WINDOW_G)] = {\n", name);
+    fprintf(fp," S(%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x)\n", SECP256K1_GE_STORAGE_CONST_GET(pre[0]));
+    j = 1;
+    for(i = 3; i <= window_g; ++i) {
+        if (withConditionals) {
```
real-or-random, line 167: I think we'd prefer snake_case. :P

real-or-random:
If you want, you could add a `.gitattributes` file with this content: 
```
src/ecmult_static_pre_g.h linguist-generated
```

This will exclude this file from GitHub language statistics (x% C Code, y% bash script, ...), and fold the file by default when showing diffs. But yeah, it's not the most important thing in the world. :P 

https://github.com/github/linguist/blob/master/docs/overrides.md

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,249 @@
+/*****************************************************************************************************
+ * Copyright (c) 2013, 2014, 2017, 2021 Pieter Wuille, Andrew Poelstra, Jonas Nick, Russell O'Connor *
+ * Distributed under the MIT software license, see the accompanying                                  *
+ * file COPYING or https://www.opensource.org/licenses/mit-license.php.                              *
+ *****************************************************************************************************/
+
+/* Autotools creates libsecp256k1-config.h, of which ECMULT_GEN_PREC_BITS is needed.
```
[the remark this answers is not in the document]
roconnor-blockstream: Fixed.

src/gen_ecmult_static_pre_g.c
roconnor-blockstream: Done.

src/gen_ecmult_static_pre_g.c
roconnor-blockstream: Done.

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,246 @@
...
+
+    fprintf(fp, "/* This file was automatically generated by gen_ecmult_static_pre_g. */\n");
+    fprintf(fp, "#ifndef SECP256K1_ECMULT_STATIC_PRE_G_H\n");
+    fprintf(fp, "#define SECP256K1_ECMULT_STATIC_PRE_G_H\n");
+    fprintf(fp, "#include \"src/group.h\"\n");
+    fprintf(fp, "#ifdef S\n");
+    fprintf(fp, "   #error macro identifier S already in use.\n");
+    fprintf(fp, "#endif\n");
+    fprintf(fp, "#define S(a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p) "
+                "SECP256K1_GE_STORAGE_CONST(0x##a##u,0x##b##u,0x##c##u,0x##d##u,0x##e##u,0x##f##u,0x##g##u,0x##h##u,0x##i##u,0x##j##u,0x##k##u,0x##l##u,0x##m##u,0x##n##u,0x##o##u,0x##p##u)\n");
```
roconnor-blockstream, line 214: Maybe break up this string across two lines.

src/tests.c
```diff
@@ -3384,6 +3384,72 @@ void run_group_decompress(void) {
...
+    secp256k1_ge p, q, gg;
+    secp256k1_fe dpx, dpy, dqx, dqy;
+    size_t i;
+
+    CHECK(0 < n);
+    secp256k1_ge_from_storage(&p, &pre_g[0]);
+    secp256k1_gej_set_ge(&g2, &p);
+    secp256k1_gej_double_var(&g2, &g2, NULL);
+    secp256k1_ge_set_gej_var(&gg, &g2);
+    secp256k1_fe_verify(&p.x);
```
roconnor-blockstream, line 3398: Move these there verification lines up to immediately after the `ge_from_storage`.

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,246 @@
...
+        }
+        if (with_conditionals) {
+            fprintf(fp,"#endif\n");
+        }
+    }
+    fprintf(fp, "};\n");
+
+}
+
+void printTwoTables(FILE* fp, int window_g, secp256k1_gej * gj, int with_conditionals) {
```
real-or-random, line 181: here also snake case and also in `printTwoTables`

real-or-random (commented):
should `#include <stdio.h>`

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,246 @@
...
+        secp256k1_fe_negate(&p_ge.x, &p_ge.x, 1);
+        secp256k1_fe_add(&p_ge.x, &dx_over_dz_squared);
+        /* y is stored_y/z^3, as we expect */
+        secp256k1_fe_mul(&p_ge.y, &p_ge.y, &zi3);
+        /* Store */
+        secp256k1_ge_to_storage(&pre[i], &p_ge);
+    }
+}
+
+void printTable(FILE* fp, const char * name, int window_g, const secp256k1_gej * gj, int with_conditionals) {
```
roconnor-blockstream, line 158: snake_case this function.

roconnor-blockstream: fixed.

src/gen_ecmult_static_pre_g.c
roconnor-blockstream: Fixed.

src/tests.c
roconnor-blockstream: fixed.

src/gen_ecmult_static_pre_g.c
roconnor-blockstream: fixed.

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,246 @@
...
+#undef USE_EXTERNAL_ASM
+
+#include "../include/secp256k1.h"
+#include "assumptions.h"
+#include "util.h"
+#include "field_impl.h"
+#include "group_impl.h"
+#include "ecmult.h"
+
+static secp256k1_ge_storage pre[ECMULT_TABLE_SIZE(ECMULT_WINDOW_SIZE) < ECMULT_TABLE_SIZE(8)
```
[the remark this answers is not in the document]
roconnor-blockstream: Done.

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,249 @@
...
+    secp256k1_ecmult_odd_multiples_table_storage(ECMULT_TABLE_SIZE(window_g), pre, gj);
+    fprintf(fp, "static const secp256k1_ge_storage %s[ECMULT_TABLE_SIZE(WINDOW_G)] = {\n", name);
+    fprintf(fp, " S(%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x)\n", SECP256K1_GE_STORAGE_CONST_GET(pre[0]));
+    j = 1;
+    for(i = 3; i <= window_g; ++i) {
+        if (with_conditionals) {
+            fprintf(fp, "#if ECMULT_TABLE_SIZE(WINDOW_G) > %ld\n", ECMULT_TABLE_SIZE(i-1));
+        }
+        for(;j < ECMULT_TABLE_SIZE(i); ++j) {
+            fprintf(fp, ",S(%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x)\n", SECP256K1_GE_STORAGE_CONST_GET(pre[j]));
```
real-or-random, line 173: I think these should be `%lx` (or `PRIu32N`) because `unsigned long` is guaranteed to be 32 bits but `unsigned int` isn't.

src/tests.c
```diff
@@ -3416,6 +3416,74 @@ void run_group_decompress(void) {
 
 /***** ECMULT TESTS *****/
 
+void test_pre_g_table(const secp256k1_ge_storage * pre_g, size_t n) {
```
real-or-random, line 3419: spacing around `*`

real-or-random (commented):
Code is good, here are some nits.

real-or-random:
I wonder what we should do with the context now... For example, the `ecdsa_verify` still says that it needs a verification context but that won't be true after this PR. Contexts require some discussion (see for example https://github.com/bitcoin-core/secp256k1/issues/780#issuecomment-670632456) but it should not delay this PR.  Maybe leave this as is now and postpone this until we made ecmult_gen stuff mandatory static?

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,249 @@
...
+
+    secp256k1_ecmult_odd_multiples_table_storage(ECMULT_TABLE_SIZE(window_g), pre, gj);
+    fprintf(fp, "static const secp256k1_ge_storage %s[ECMULT_TABLE_SIZE(WINDOW_G)] = {\n", name);
+    fprintf(fp, " S(%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x,%x)\n", SECP256K1_GE_STORAGE_CONST_GET(pre[0]));
+    j = 1;
+    for(i = 3; i <= window_g; ++i) {
+        if (with_conditionals) {
+            fprintf(fp, "#if ECMULT_TABLE_SIZE(WINDOW_G) > %ld\n", ECMULT_TABLE_SIZE(i-1));
+        }
+        for(;j < ECMULT_TABLE_SIZE(i); ++j) {
```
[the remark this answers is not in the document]
roconnor-blockstream: `j` is not reset each time this inner loop is executed.

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,249 @@
...
+    fp = fopen("src/ecmult_static_pre_g.h","w");
+    if (fp == NULL) {
+        fprintf(stderr, "Could not open src/ecmult_static_pre_g.h for writing!\n");
+        return -1;
+    }
+
+    fprintf(fp, "/* This file was automatically generated by gen_ecmult_static_pre_g. */\n");
+    fprintf(fp, "#ifndef SECP256K1_ECMULT_STATIC_PRE_G_H\n");
+    fprintf(fp, "#define SECP256K1_ECMULT_STATIC_PRE_G_H\n");
+    fprintf(fp, "#include \"src/group.h\"\n");
```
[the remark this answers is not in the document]
roconnor-blockstream: Fixed.

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,249 @@
...
+                "SECP256K1_GE_STORAGE_CONST(0x##a##u,0x##b##u,0x##c##u,0x##d##u,0x##e##u,0x##f##u,0x##g##u,"
+                "0x##h##u,0x##i##u,0x##j##u,0x##k##u,0x##l##u,0x##m##u,0x##n##u,0x##o##u,0x##p##u)\n");
+    fprintf(fp, "#if ECMULT_TABLE_SIZE(ECMULT_WINDOW_SIZE) > %ld\n", ECMULT_TABLE_SIZE(ECMULT_WINDOW_SIZE));
+    fprintf(fp, "   #error configuration mismatch, invalid ECMULT_WINDOW_SIZE. Try deleting ecmult_static_pre_g.h before the build.\n");
+    fprintf(fp, "#endif\n");
+    fprintf(fp, "#if defined(EXHAUSTIVE_TEST_ORDER)\n");
+    fprintf(fp, "#if EXHAUSTIVE_TEST_ORDER == 13\n");
+    fprintf(fp, "#define WINDOW_G %d\n", window_g_13);
+
+    secp256k1_gej_set_ge(&gj, &g_13);
```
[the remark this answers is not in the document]
roconnor-blockstream: Done.

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,249 @@
...
+    fprintf(fp, "#elif EXHAUSTIVE_TEST_ORDER == 199\n");
+    fprintf(fp, "#define WINDOW_G %d\n", window_g_199);
+
+    secp256k1_gej_set_ge(&gj, &g_199);
+    print_two_tables(fp, window_g_199, &gj, 0);
+
+    fprintf(fp, "#else\n");
+    fprintf(fp, "   #error No known generator for the specified exhaustive test group order.\n");
+    fprintf(fp, "#endif\n");
+    fprintf(fp, "#else\n");
```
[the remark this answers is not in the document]
roconnor-blockstream: Done.

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,249 @@
...
+        }
+        if (with_conditionals) {
+            fprintf(fp, "#endif\n");
+        }
+    }
+    fprintf(fp, "};\n");
+
+}
+
+void print_two_tables(FILE* fp, int window_g, secp256k1_gej * gj, int with_conditionals) {
```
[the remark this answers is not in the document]
roconnor-blockstream: Fixed.

src/gen_ecmult_static_pre_g.c
roconnor-blockstream: Fixed.

src/tests.c
roconnor-blockstream: Fixed.

roconnor-blockstream:
<sipa> on cortex-A73 (Odroid N2+), the optimal table size seems to be 22...
    <sipa> 196 us at 15, 189 us at 22
    <sipa> for an ecdsa verify
    <sipa> that's a 128 MiB table :s

src/gen_ecmult_static_pre_g.c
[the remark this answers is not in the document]
real-or-random: Oh sure!

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,252 @@
...
+/* Autotools creates libsecp256k1-config.h, of which ECMULT_WINDOW_SIZE is needed.
+   ifndef guard so downstream users can define their own if they do not use autotools. */
+#if !defined(ECMULT_WINDOW_SIZE)
+#include "libsecp256k1-config.h"
+#endif
+
+/* In principle we could use external ASM, but this yields only a minor speedup in
+   build time and it's very complicated. In particular when cross-compiling, we'd
+   need to build the external ASM for the build and the host machine. */
+#undef USE_EXTERNAL_ASM
```
real-or-random, line 19: Since this code is taken from `gen_context.c`, it should be updated if https://github.com/bitcoin-core/secp256k1/pull/965 is merged before this PR.

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,252 @@
...
+        secp256k1_fe_add(&p_ge.x, &dx_over_dz_squared);
+        /* y is stored_y/z^3, as we expect */
+        secp256k1_fe_mul(&p_ge.y, &p_ge.y, &zi3);
+        /* Store */
+        secp256k1_ge_to_storage(&pre[i], &p_ge);
+    }
+}
+
+void print_table(FILE *fp, const char *name, int window_g, const secp256k1_gej *gj, int with_conditionals) {
+    /* This array is only static because it may be unreasonably large to place on the stack. */
```
elichai, line 158: small nit, since this is a binary and not a library you can use malloc if you have concerns about sizes

roconnor-blockstream: I do think it is somewhat better as a static array, but if other people also prefer a dynamic array, I can change it.

Perhaps even better is to eliminate the allocation entirely and rewrite `secp256k1_ecmult_odd_multiples_table_storage` to be a simpler and more direct computation.  Since we have to call `ge_normalize` on every output anyways, there is little point in getting fancy.  `secp256k1_ecmult_odd_multiples_table_storage` is only the way it is because I have moved it from `ecmult_impl`.

src/ecmult_impl.h
```diff
@@ -44,28 +44,14 @@
...
- * If WINDOW_G > 27 and size_t has 32 bits, then the code is incorrect
- * because the size of the memory object that we allocate (in bytes)
- * will not fit in a size_t.
- * If WINDOW_G > 31 and int has 32 bits, then the code is incorrect
- * because certain expressions will overflow.
- */
-#if ECMULT_WINDOW_SIZE < 2 || ECMULT_WINDOW_SIZE > 24
-#  error Set ECMULT_WINDOW_SIZE to an integer in range [2..24].
+#if ECMULT_WINDOW_SIZE < WINDOW_G
+#  error ECMULT_WINDOW_SIZE to small for WINDOW_G.
```
sipa: s/to/too/

src/gen_ecmult_static_pre_g.c
sipa: #965 was in fact merged before this PR; what needs to be done?

src/gen_ecmult_static_pre_g.c
sipa: It's slightly strange to use a global object just because the size is too large; using the heap would be more natural I'd say. But also, the current code seems to work fine, no need to change it.

src/ecmult_impl.h
roconnor-blockstream: Fixed.

src/gen_ecmult_static_pre_g.c
roconnor-blockstream: I've cargo-cult-copied #965.

src/gen_ecmult_static_pre_g.c
roconnor-blockstream: The array has been nuked.

sipa:
utACK 8e9f75a5888a8ec549fe9026053051c3db7a1282

src/ecmult_impl.h
```diff
@@ -142,137 +121,6 @@ static void secp256k1_ecmult_odd_multiples_table_globalz_windowa(secp256k1_ge *p
     secp256k1_ge_globalz_set_table_gej(ECMULT_TABLE_SIZE(WINDOW_A), pre, globalz, prej, zr);
 }
 
-static void secp256k1_ecmult_odd_multiples_table_storage_var(const int n, secp256k1_ge_storage *pre, const secp256k1_gej *a) {
```
roconnor-blockstream, line 145: The comment 20ish lines above here is out of date.

roconnor-blockstream: Resolved.

laanwj:
Concept and code review ACK 8e9f75a5888a8ec549fe9026053051c3db7a1282
Embedding the read-only data is useful on embedded systems which have plenty of ROM/FLASH but limited RAM.

roconnor-blockstream:
Reworded the commit messages.

laanwj:
Thanks!
re-ACK f4d79009bff8db0d1e53178b39072e504ec49156

src/ecmult_impl.h
```diff
@@ -544,11 +318,12 @@ static void secp256k1_ecmult_strauss_wnaf(const secp256k1_ecmult_context *ctx, c
             }
         }
         if (i < bits_ng_1 && (n = wnaf_ng_1[i])) {
-            ECMULT_TABLE_GET_GE_STORAGE(&tmpa, *ctx->pre_g, n, WINDOW_G);
+            ECMULT_TABLE_GET_GE_STORAGE(&tmpa, secp256k1_pre_g, n, WINDOW_G);
             secp256k1_gej_add_zinv_var(r, r, &tmpa, &Z);
         }
         if (i < bits_ng_128 && (n = wnaf_ng_128[i])) {
-            ECMULT_TABLE_GET_GE_STORAGE(&tmpa, *ctx->pre_g_128, n, WINDOW_G);
+            VERIFY_CHECK(secp256k1_pre_g_128 != NULL);
```
jonasnick, line 325: What does this VERIFY_CHECK do? Isn't `secp256k1_pre_g_128` an array?

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,134 @@
...
+    const int window_g_199 = 8;
+    FILE* fp;
+
+    fp = fopen("src/ecmult_static_pre_g.h","w");
+    if (fp == NULL) {
+        fprintf(stderr, "Could not open src/ecmult_static_pre_g.h for writing!\n");
+        return -1;
+    }
+
+    fprintf(fp, "/* This file was automatically generated by gen_ecmult_static_pre_g. */\n");
```
jonasnick, line 96: I think we removed the comment that documented what the tables actually contain. How about adding something like
```
/* This file contains an array secp256k1_pre_g with odd multiples of the base point G and an array secp256k1_pre_g_128 with odd multiples of 2^128*G */
```

src/tests.c
jonasnick, line 3419: Took me embarassingly long to understand what is going on here. Perhaps worth adding an explanation like:
```
For all i in [0, n), p = pre_g[i], q = pre_g[i+1], gg = 2*pre_g[0] we have p + gg = q. Instead of more costly group operations, we check that the 3 points p, gg and -q are distinct, on the curve and on the same line.
```

jonasnick (commented):
I didn't have trouble to build (and run the tests) with window size 22 on my laptop with 32 GB of memory. It looked like it never used more than 20GB. So concept ACK. For posteriority (and this should probably be documented somewhere):
```
rm src/ecmult_static_pre_g.h
./configure --with-ecmult-window=22
make gen_ecmult_static_pre_g
./gen_ecmult_static_pre_g
# ecmult_static_pre_g.h is 294 MB
make -j 1 # make sure to use one job because otherwise it uses even more memory
```

roconnor-blockstream:
@jonasnick In principle we ought to be able to support upto a window size of 24 if you want to try that as well.

src/ecmult_impl.h
roconnor-blockstream: This is outdated and needs to be removed.
I used to have `secp256k1_pre_g_128` equal to `NULL` for `EXHAUSTIVE_TEST_ORDER`, but later I rewrote it to define it even in that case, even though it is not used at the moment.

roconnor-blockstream: Fixed.

src/tests.c
roconnor-blockstream: I added some more docs.

src/gen_ecmult_static_pre_g.c
roconnor-blockstream: Added a comment.

jonasnick (commented):
ACK 47f6185d622fa77a6dd6420aa667213f8d8b38f2

I tested with various ecmult_windows sizes.

src/tests.c
jonasnick: Thanks!

sipa:
@gmaxwell You want to try building with the table size set to 24?

src/ecmult.h
```diff
@@ -11,20 +11,26 @@
...
-    secp256k1_ge_storage (*pre_g_128)[]; /* odd multiples of 2^128*generator */
-} secp256k1_ecmult_context;
-
-static void secp256k1_ecmult_context_init(secp256k1_ecmult_context *ctx);
-static void secp256k1_ecmult_context_build(secp256k1_ecmult_context *ctx, void **prealloc);
-static void secp256k1_ecmult_context_finalize_memcpy(secp256k1_ecmult_context *dst, const secp256k1_ecmult_context *src);
-static void secp256k1_ecmult_context_clear(secp256k1_ecmult_context *ctx);
-static int secp256k1_ecmult_context_is_built(const secp256k1_ecmult_context *ctx);
+/* Noone will ever need more than a window size of 24. The code might
+ * be correct for larger values of ECMULT_WINDOW_SIZE but this is not
```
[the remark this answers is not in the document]
roconnor-blockstream: Fixed.

configure.ac
```diff
@@ -176,6 +176,7 @@ AC_ARG_WITH([asm], [AS_HELP_STRING([--with-asm=x86_64|arm|no|auto],
 AC_ARG_WITH([ecmult-window], [AS_HELP_STRING([--with-ecmult-window=SIZE|auto],
 [window size for ecmult precomputation for verification, specified as integer in range [2..24].]
 [Larger values result in possibly better performance at the cost of an exponentially larger precomputed table.]
+[For very large window sizes, use "make -j 1" to reduce memory use during compilation.]
 [The table will store 2^(SIZE-1) * 64 bytes of data but can be larger in memory due to platform-specific padding and alignment.]
 ["auto" is a reasonable setting for desktop machines (currently 15). [default=auto]]
```
sipa, line 181: I think it would be good to explain here that setting a value larger than 15 will require regeneration of ecmult_static_pre_g.h (and maybe instructions on how to do that).

Ping @gmaxwell.

src/gen_ecmult_static_pre_g.c
```diff
@@ -0,0 +1,137 @@
...
+    fprintf(fp, "#define SECP256K1_ECMULT_STATIC_PRE_G_H\n");
+    fprintf(fp, "#include \"group.h\"\n");
+    fprintf(fp, "#ifdef S\n");
+    fprintf(fp, "   #error macro identifier S already in use.\n");
+    fprintf(fp, "#endif\n");
+    fprintf(fp, "#define S(a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p) "
+                "SECP256K1_GE_STORAGE_CONST(0x##a##u,0x##b##u,0x##c##u,0x##d##u,0x##e##u,0x##f##u,0x##g##u,"
+                "0x##h##u,0x##i##u,0x##j##u,0x##k##u,0x##l##u,0x##m##u,0x##n##u,0x##o##u,0x##p##u)\n");
+    fprintf(fp, "#if ECMULT_TABLE_SIZE(ECMULT_WINDOW_SIZE) > %ld\n", ECMULT_TABLE_SIZE(ECMULT_WINDOW_SIZE));
+    fprintf(fp, "   #error configuration mismatch, invalid ECMULT_WINDOW_SIZE. Try deleting ecmult_static_pre_g.h before the build.\n");
```
sipa, line 110: Perhaps here too; just deleting isn't enough; it also needs a "make src/ecmult_static_pre_g.h" step.

sipa:
FWIW, I believe @gmaxwell successfully built with window size 24.

configure.ac
real-or-random: Should be ship a file with values up to 16 or 17 even though 15 is the default?

src/gen_ecmult_static_pre_g.c
roconnor-blockstream: Deleting `src/ecmult_static_pre_g.h` seems to cause it to be rebuilt.

My limited experiments suggests that reconfiguring causes `gen_ecmult_static_pre_g` to be rebuilt.

Is there some particular problem you are observing?

sipa: You're right, I'm unsure why I thought differently. I built from a clean checkout with the default first, then reconfigured with window size 20, rebuilt, got this error, deleted the `src/ecmult_static_pre_g.h` file, rebuilt again, and it worked.

configure.ac
sipa: I think 15 is enough.

src/gen_ecmult_static_pre_g.c
roconnor-blockstream: We don't ship with `gen_ecmult_static_pre_g` binaries, so you might need to try to reconfigure after creating that binary, but when I tested that situation it still worked for me.

configure.ac
roconnor-blockstream: Further elaborated instructions.

real-or-random:
A trivial rebase on master should fix the CI hiccups.

real-or-random:
ACK 20abd52c2e107e79391a19d2d2f8845e83858dea code inspection and tested some parameters

PR #959, opened 2021-06-28
tests: really test the non-var scalar inverse

niooss-ledger:
Function `test_inverse_scalar` contains:

    (var ? secp256k1_scalar_inverse_var : secp256k1_scalar_inverse_var)(&l, x);  /* l = 1/x */

The two sides of the condition are the same function. This seems to be an error, as there also exists a non-var function, named
`secp256k1_scalar_inverse`.

Make `test_inverse_scalar` use this other function when `var` is false.

This issue was found using clang's static analyzer, which reported a "Logic error: Identical expressions in conditional expression" (with checker `alpha.core.IdenticalExpr`).

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -2595,7 +2595,7 @@ void test_inverse_scalar(secp256k1_scalar* out, const secp256k1_scalar* x, int v
 {
     secp256k1_scalar l, r, t;
 
-    (var ? secp256k1_scalar_inverse_var : secp256k1_scalar_inverse_var)(&l, x);  /* l = 1/x */
+    (var ? secp256k1_scalar_inverse_var : secp256k1_scalar_inverse)(&l, x);  /* l = 1/x */
     if (out) *out = l;
     if (secp256k1_scalar_is_zero(x)) {
         CHECK(secp256k1_scalar_is_zero(&l));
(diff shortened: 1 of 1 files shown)


real-or-random (approved):
Whoa, nice catch!

ACK 41ed13942bd722ff0ee4d2eb8bef55155549afd6

real-or-random:
> This issue was found using clang's static analyzer, which reported a "Logic error: Identical expressions in conditional expression" (with checker `alpha.core.IdenticalExpr`).

Can you elaborate on how you used/called the tool? Maybe we could add it to our arsenal in CI if it doesn't give a gazillion false positives.

niooss-ledger:
> Can you elaborate on how you used/called the tool? Maybe we could add it to our arsenal in CI if it doesn't give a gazillion false positives.

I used the Ubuntu package `clang-tools` to install this tool. Then I ran this command line (on Ubuntu):

```sh
scan-build \
    --force-analyze-debug-code \
    -enable-checker security \
    -enable-checker unix \
    -enable-checker valist \
    -enable-checker alpha.core.BoolAssignment \
    -enable-checker alpha.core.CastSize \
    -enable-checker alpha.core.CastToStruct \
    -enable-checker alpha.core.DynamicTypeChecker \
    -enable-checker alpha.core.FixedAddr \
    -enable-checker alpha.core.IdenticalExpr \
    -enable-checker alpha.core.PointerArithm \
    -enable-checker alpha.core.PointerSub \
    -enable-checker alpha.core.SizeofPtr \
    -enable-checker alpha.core.TestAfterDivZero \
    -enable-checker alpha.security.ArrayBoundV2 \
    -enable-checker alpha.security.MallocOverflow \
    -enable-checker alpha.security.ReturnPtrRange \
    -enable-checker alpha.security.taint.TaintPropagation \
    -enable-checker alpha.unix.PthreadLock \
    -enable-checker alpha.unix.SimpleStream \
    -enable-checker alpha.unix.Stream \
    -enable-checker alpha.unix.cstring.BufferOverlap \
    -enable-checker alpha.unix.cstring.OutOfBounds \
    -enable-checker alpha.unix.cstring.NotNullTerminated \
    -maxloop 257 \
    -analyze-headers \
    -o "$(pwd)/scan-build-output" \
    sh -c 'export CFLAGS=-DVERIFY ; ./configure --enable-coverage=no --with-asm=no && make'
```

This command produces several HTML files in a new directory `scan-build-output`.

* `CFLAGS=-DVERIFY` is used to greatly reduce the number of "NULL pointer dereference" reports. By the way, it would be nice if there was a `configure` switch for this ;)
* `--enable-coverage=no` (which is the default value) is used to ensure that `VERIFY_CHECK` are compiled, in order to avoid "NULL pointer dereference" reports.
* `--with-asm=no` is used to avoid using the ASM implementation of some functions, that the static analyzer cannot use.

At first there were ~100 warnings, but many of them were duplicates because they occur in C header files.

real-or-random:
Thanks, that's helpful. Now I remember I used this before. I believe it's possible to eliminate most of the false positives.

>     * `CFLAGS=-DVERIFY` is used to greatly reduce the number of "NULL pointer dereference" reports. By the way, it would be nice if there was a `configure` switch for this ;)

Not a switch so far but the easiest way to set environment variables is to *pass* them to `configure`, e.g., `./configure CPPFLAGS=-DVERIFY`. (Strictly speaking, `-DVERIFY` is a C preprocessor flag, so it should go to  `CPPFLAGS` even though `CFLAGS` will also work in this case.)

jonasnick (commented):
ACK 41ed13942bd722ff0ee4d2eb8bef55155549afd6

jonasnick:
Thank you @niooss-ledger

sipa:
Posthumous ACK

PR #960, opened 2021-07-02
tests_exhaustive: check the result of secp256k1_ecdsa_sign

niooss-ledger:
Hello,

In `test_exhaustive_sign`, if `secp256k1_ecdsa_sign` fails, the signature which is then loaded by `secp256k1_ecdsa_signature_load` is garbage. Exit early with an error when this occurs.

By the way, I am wondering whether attribute `SECP256K1_WARN_UNUSED_RESULT` should be added to function `secp256k1_ecdsa_sign`: as (according to the documentation of this function) the nonce generation function may fail, it seems to be a good idea to force callers to check the value returned by this function. What do you think about this?

### DIFF
--- a/src/tests_exhaustive.c
+++ b/src/tests_exhaustive.c
@@ -302,6 +302,7 @@ void test_exhaustive_sign(const secp256k1_context *ctx, const secp256k1_ge *grou
             if (skip_section(&iter)) continue;
             for (k = 1; k < EXHAUSTIVE_TEST_ORDER; k++) {  /* nonce */
                 const int starting_k = k;
+                int ret;
                 secp256k1_ecdsa_signature sig;
                 secp256k1_scalar sk, msg, r, s, expected_r;
                 unsigned char sk32[32], msg32[32];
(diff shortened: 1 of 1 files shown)


real-or-random (approved):
utACK a1ee83c6546c65d8f5b32acc4a0e1740858ee7d6

real-or-random:
Thanks for the PR!

> By the way, I am wondering whether attribute `SECP256K1_WARN_UNUSED_RESULT` should be added to function `secp256k1_ecdsa_sign`: as (according to the documentation of this function) the nonce generation function may fail, it seems to be a good idea to force callers to check the value returned by this function. What do you think about this?

The default nonce generation function will fail only with astronomically low probability. So if you know that you have a valid secret key and you use the default nonce function (99% of the use cases), it's okay not to check the return value.

Having said that, I think we're not entirely consistent here... For example, the same argument would apply to `secp256k1_ec_seckey_verify` (https://github.com/bitcoin-core/secp256k1/blob/master/include/secp256k1.h#L632). Even `secp256k1_ec_pubkey_negate` 
https://github.com/bitcoin-core/secp256k1/blob/master/include/secp256k1.h#L650 has  `SECP256K1_WARN_UNUSED_RESULT` even if it's guaranteed to return 1 according to the docs...

Maybe we should have a look at this in https://github.com/bitcoin-core/secp256k1/pull/783 or in a follow up PR.

sipa:
ACK a1ee83c6546c65d8f5b32acc4a0e1740858ee7d6

PR #963, opened 2021-07-04
"Schnorrsig API overhaul" fixups

### LINKED ISSUE #962: SECP256K1_SCHNORRSIG_EXTRAPARAMS_INIT does not work with C++
```c
#include <secp256k1.h>
#include <secp256k1_schnorrsig.h>

int main(void)
{
    secp256k1_schnorrsig_extraparams extraparams = SECP256K1_SCHNORRSIG_EXTRAPARAMS_INIT;
    return 0;
}
```

This compiles as C but not as C++.

clang++:

```
x.cpp:6:52: error: initializer-string for char array is too long
    secp256k1_schnorrsig_extraparams extraparams = SECP256K1_SCHNORRSIG_EXTRAPARAMS_INIT;
                                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/secp256k1/include/secp256k1_schnorrsig.h:90:5: note: expanded from macro 'SECP256K1_SCHNORRSIG_EXTRAPARAMS_INIT'
    SECP256K1_SCHNORRSIG_EXTRAPARAMS_MAGIC,\
    ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/secp256k1/include/secp256k1_schnorrsig.h:88:48: note: expanded from macro 'SECP256K1_SCHNORRSIG_EXTRAPARAMS_MAGIC'
#define SECP256K1_SCHNORRSIG_EXTRAPARAMS_MAGIC "\xda\x6f\xb3\x8c"
                                               ^~~~~~~~~~~~~~~~~~
1 error generated.
```

g++

```
In file included from x.cpp:2:0:
x.cpp: In function ‘int main()’:
/secp256k1/include/secp256k1_schnorrsig.h:93:1: error: initializer-string for array of chars is too long [-fpermissive]
 }
 ^
x.cpp:6:52: note: in expansion of macro ‘SECP256K1_SCHNORRSIG_EXTRAPARAMS_INIT’
     secp256k1_schnorrsig_extraparams extraparams = SECP256K1_SCHNORRSIG_EXTRAPARAMS_INIT;
```

real-or-random:
Fixups for #844 .

Fixes #962 .

Now I wonder if a `static const` in the header file would be better than a macro.

By the way, an unrelated issue is that we should have `extern C` as we do in the main header...

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -320,3 +320,23 @@ task:
     - ./ci/cirrus.sh
   << : *CAT_LOGS
 
+task:
+  name: "C++ -fpermissive"
+  container:
+    dockerfile: ci/linux-debian.Dockerfile
+    cpu: 1
+    memory: 1G
+  env:
+    # ./configure correctly errors out when given CC=g++.
+    # We hack around this by passing CC=g++ only to make.
+    CC: gcc
+    MAKEFLAGS: -j2 CC=g++ CFLAGS=-fpermissive
+    WERROR_CFLAGS:
+    EXPERIMENTAL: yes
+    ECDH: yes
+    RECOVERY: yes
+    SCHNORRSIG: yes
+  << : *MERGE_BASE
+  test_script:
+    - ./ci/cirrus.sh
+  << : *CAT_LOGS
--- a/ci/linux-debian.Dockerfile
+++ b/ci/linux-debian.Dockerfile
@@ -13,6 +13,7 @@ RUN apt-get install --no-install-recommends --no-upgrade -y \
         git ca-certificates \
         make automake libtool pkg-config dpkg-dev valgrind qemu-user \
         gcc clang llvm libc6-dbg \
+        g++ \
         gcc-i686-linux-gnu libc6-dev-i386-cross libc6-dbg:i386 libubsan1:i386 libasan5:i386 \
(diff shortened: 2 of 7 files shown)


gmaxwell:
LGTM. Maybe you should add a CI build that builds the entire project/tests/benchmarks with a C++ compiler-- it should compile. It may need -fpermissive or a few extra explicit casts. (Don't use Werror, though).

elichai:
utACK 792d7c46668144e0ca2a321b0765935037258d2f
(currently the CI fails on master and this fixes it)

real-or-random:
Not ready to merge, I'll have some uncommitted additional changes but had to interrupt my work.

gmaxwell:
If this is going to take a while, it might make sense to revert the last change to master.

lesliebranson (commented):
.

jonasnick (commented):
Nice, CI output looks as expected.

Fwiw, the issue with CI on master after merging #844 wouldn't have been easily caught before pushing because the `-Wmissing-braces` warning doesn't appear with the version of clang installed on my system at least (I have version 12, appears with v7).

> By the way, an unrelated issue is that we should have extern C as we do in the main header...

We do already.

ACK 90e83449b2c2e4046af755e35fdeee579a468f31

Issue #817, opened 2020-09-15
Experimental features should be deprecated


rustyrussell:
If a feature is genuinely experimental, keep it in a separate branch.

Otherwise, it *will* be compiled in, as there may be a user. This becomes inevitable in larger systems: in particular, c-lightning uses both libsecp and libwally. The latter includes libsecp, and you can't have two libraries with the same symbols, so c-lightning needs to use the libwally one. Packaging systems fix this, but they have the same problem: they must configure in everything.

You can have boutique builds that omit features for embedded use, sure. But the default will be kitchen sink.

I know it's hard: labelling something experimental is a good way of avoiding commitment. But it is a mere fig leaf which doesn't work, and mature libraries don't get that luxury.

If you're not prepared to commit to an API (ideally an ABI too), you simply cannot include something. In that case, it needs to be a separate library (and if it gets folded in later, all the names have to change!).

real-or-random:
> If a feature is genuinely experimental, keep it in a separate branch.

I think this would make it easier to develop large features, e.g.. Schnorr signatures. Currently our model (and that of Bitcoin Core) is to have large atomic changes in a single PR developed by a single party who addresses review comments over and over. And reviewers keep reviewing the same stuff because people don't want to commit and the PR takes forever to be merged. I believe it's easier to develop large changes in a separate feature branch (or fork), but then create small PRs against this feature branch to avoid that PRs become huge, and later create a single large PR against master. Some things are simply too large to be handled in a single PR. 

This is what Bitcoin Core is effectively doing with secp256k1.

(Sorry, this is somewhat off-topic.)

sipa:
I think there are a few somewhat related issues here.

### "experimental" used to not mean very much

This library wasn't well maintained for a long time, and tat led to a few things remaining marked experimental for far longer than they should have been. This has improved lately, and ECDH (which was effectively stable for years) is now no longer experimental. I think the ARM assembly will in the near future become non-experimental as well, but I'd like to see CI support for it.

Beyond those two, the only remaining experimental feature is the recently-merged BIP340 support. It is experimental for two reasons:

* There are a few discussions still ongoing that may impact the API (variable-length message support, batch validation) without changes to the (currently supported) part of the scheme. It's not impossible to just add new functions for these once they're worked out, but given the next point, I think this may not be needed.
[...]

PR #964, opened 2021-07-07
Add release-process.md

### LINKED ISSUE #856: Add doc/release-process.md

jonasnick:
This is an attempt at a simple release process. Fixes #856. To keep it simple, there is no concept of release candidates for now.

The release version is determined by semantic versioning of the API. Since it does not seem to be a lot of work, it is proper to also version the ABI with the libtool versioning system. This versioning scheme (semver API, libtool versioning ABI) follows the suggestion in the [autotools mythbusters](https://autotools.io/libtool/version.html).

Experimental modules are a bit of a headache, as expected. This release process suggests to treat any change in experimental modules as backwards compatible. That way, users of stable modules are not bothered by frequent non-backwards compatible releases. But a downside is that one must not use experimental modules in shared libraries (which should be mentioned in the README?). It would be nice if we could make the schnorrsig module stable in the not too distant future (see also #817).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1,5 +1,15 @@
 AC_PREREQ([2.60])
-AC_INIT([libsecp256k1],[0.1])
+
+# The package (a.k.a. release) version is based on semantic versioning 2.0.0 of
+# the API. All changes in experimental modules are treated as
+# backwards-compatible and therefore at most increase the minor version.
+define(_PKG_VERSION_MAJOR, 0)
+define(_PKG_VERSION_MINOR, 1)
+define(_PKG_VERSION_BUILD, 0)
+define(_PKG_VERSION_IS_RELEASE, false)
+
+AC_INIT([libsecp256k1],m4_join([.], _PKG_VERSION_MAJOR, _PKG_VERSION_MINOR, _PKG_VERSION_BUILD)m4_if(_PKG_VERSION_IS_RELEASE, [true], [], [-pre]),[https://github.com/bitcoin-core/secp256k1/issues],[libsecp256k1],[https://github.com/bitcoin-core/secp256k1])
+
 AC_CONFIG_AUX_DIR([build-aux])
 AC_CONFIG_MACRO_DIR([build-aux/m4])
 AC_CANONICAL_HOST
--- a/configure.ac
+++ b/configure.ac
@@ -8,6 +8,15 @@ define(_PKG_VERSION_MINOR, 1)
 define(_PKG_VERSION_BUILD, 0)
 define(_PKG_VERSION_IS_RELEASE, false)
 
+# The library version is based on libtool versioning of the ABI. The set of
+# rules for updating the version can be found here:
+# https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html
+# All changes in experimental modules are treated as if they don't affect the
+# interface and therefore only increase the revision.
+define(_LIB_VERSION_CURRENT, 0)
+define(_LIB_VERSION_REVISION, 0)
+define(_LIB_VERSION_AGE, 0)
+
 AC_INIT([libsecp256k1],m4_join([.], _PKG_VERSION_MAJOR, _PKG_VERSION_MINOR, _PKG_VERSION_BUILD)m4_if(_PKG_VERSION_IS_RELEASE, [true], [], [-pre]),[https://github.com/bitcoin-core/secp256k1/issues],[libsecp256k1],[https://github.com/bitcoin-core/secp256k1])
 
 AC_CONFIG_AUX_DIR([build-aux])
@@ -391,6 +400,9 @@ AM_CONDITIONAL([ENABLE_MODULE_EXTRAKEYS], [test x"$enable_module_extrakeys" = x"
 AM_CONDITIONAL([ENABLE_MODULE_SCHNORRSIG], [test x"$enable_module_schnorrsig" = x"yes"])
 AM_CONDITIONAL([USE_EXTERNAL_ASM], [test x"$use_external_asm" = x"yes"])
 AM_CONDITIONAL([USE_ASM_ARM], [test x"$set_asm" = x"arm"])
+AC_SUBST(LIB_VERSION_CURRENT, _LIB_VERSION_CURRENT)
+AC_SUBST(LIB_VERSION_REVISION, _LIB_VERSION_REVISION)
+AC_SUBST(LIB_VERSION_AGE, _LIB_VERSION_AGE)
 
 # Make sure nothing new is exported so that we don't break the cache.
 PKGCONFIG_PATH_TEMP="$PKG_CONFIG_PATH"
--- a/doc/release-process.md
+++ b/doc/release-process.md
@@ -1,0 +1,14 @@
+# Release Process
+
+1. Open PR to master that
+   1. adds release notes to `doc/CHANGELOG.md` and
+   2. if this is **not** a patch release, updates `_PKG_VERSION_{MAJOR,MINOR}` and `_LIB_VERSIONS_*` in `configure.ac`
+2. After the PR is merged,
+   * if this is **not** a patch release, create a release branch with name `MAJOR.MINOR`.
+     Make sure that the branch contains the right commits.
+     Create commit on the release branch that sets `_PKG_VERSION_IS_RELEASE` in `configure.ac` to `true`.
+   * if this **is** a patch release, open a pull request with the bugfixes to the `MAJOR.MINOR` branch.
+     Also include the release note commit bump `_PKG_VERSION_BUILD` and `_LIB_VERSIONS_*` in `configure.ac`.
+4. Tag the commit with `git tag -s vMAJOR.MINOR.PATCH`.
+5. Push branch and tag with `git push origin --tags`.
+6. Create a new GitHub release with a link to the corresponding entry in `doc/CHANGELOG.md`.
--- a/Makefile.am
+++ b/Makefile.am
@@ -88,7 +88,7 @@ endif
 libsecp256k1_la_SOURCES = src/secp256k1.c
 libsecp256k1_la_CPPFLAGS = -I$(top_srcdir)/include -I$(top_srcdir)/src $(SECP_INCLUDES)
 libsecp256k1_la_LIBADD = $(SECP_LIBS) $(COMMON_LIB) $(PRECOMPUTED_LIB)
-libsecp256k1_la_LDFLAGS = -no-undefined
+libsecp256k1_la_LDFLAGS = -no-undefined -version-info $(LIB_VERSION_CURRENT):$(LIB_VERSION_REVISION):$(LIB_VERSION_AGE)
 
 if VALGRIND_ENABLED
 libsecp256k1_la_CPPFLAGS += -DVALGRIND
--- a/doc/CHANGELOG.md
+++ b/doc/CHANGELOG.md
@@ -1,0 +1,12 @@
+# Changelog
+
+This file is currently only a template for future use.
+
+Each change falls into one of the following categories: Added, Changed, Deprecated, Removed, Fixed or Security.
+
+## [Unreleased]
+
+## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
+
+### Added/Changed/Deprecated/Removed/Fixed/Security
+- [Title with link to Pull Request](https://link-to-pr)


rustyrussell:
This looks good!  Note that c-lightning has a tool which extracts Changelog entries from commit messages, and also a bot that tests they exist for each PR (or Changelog-None is in GitHub description of PR).

prusnak:
I suggest to use the format described here: https://keepachangelog.com/ - the guide also mentions that ChangeLog is not a git log dump :-)

rustyrussell:
> I suggest to use the format described here: https://keepachangelog.com/ - the guide also mentions that ChangeLog is not a git log dump :-)

Just to be clear: our tool extracts lives starting with Changelog-(Added|Fixed|Changed|Deprecated|Removed) and adds links to the PRs and sorts them :)

prusnak:
> our tool extracts lives

God, what kind of tools do you use?! 😁

jonasnick:
Thanks for the feedback. I added an `Unreleased` section to the changelog and categories as suggested by keepachangelog.com. I think in most cases the titles of PRs are sufficiently descriptive such that CHANGELOG entries can be title of PR + link.

configure.ac
```diff
@@ -1,5 +1,15 @@
...
+
+# The package (a.k.a. release) version is based on semantic versioning 2.0.0 of
+# the API. All changes in experimental modules are treated as
+# backwards-compatible and therefore at most increase the minor version.
+define(_PKG_VERSION_MAJOR, 0)
+define(_PKG_VERSION_MINOR, 1)
+define(_PKG_VERSION_BUILD, 0)
+define(_PKG_VERSION_IS_RELEASE, false)
+
+AC_INIT([libsecp256k1],m4_join([.], _PKG_VERSION_MAJOR, _PKG_VERSION_MINOR, _PKG_VERSION_BUILD)m4_if(_PKG_VERSION_IS_RELEASE, [true], [], [-pre]),[https://github.com/bitcoin-core/secp256k1/issues],[secp256k1],[https://github.com/bitcoin-core/secp256k1])
```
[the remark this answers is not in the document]
jonasnick: good point, fixed

sipa:
ACK procedure/changes, but somehow CI does not like this.

jonasnick (commented):
rebased

real-or-random (approved):
utACK 86d0b9eba0b191319c8b6dbfd9d014420f06c836

sipa:
Hmm, what is the expected difference between the package version and the library version? Why would they not be the same?

real-or-random:
@sipa See https://autotools.io/libtool/version.html

@jonasnick Sorry, needs trivial rebase again.

sipa:
Can we write in the release process document how each of the 6 numbers are supposed to be updated as a function of the changes in such releases?

sipa:
Another question (not necessarily something to be explicitly explained in the release process here, but still something that isn't currently clear to me): what branches structure are the release branches going to use? Possibilities:
* All release branches branch off master, and thus there is no (clean) way to create a `x.y.(z+1)` after `x.(y+1).z` has been released.
* The branch for `x.(y+1).0` branches off the `x.y.0` branch; the branch for `x.y.(z+1)` branches off the `x.y.z` one. Open questions with this:
  * How to obtain monotonically increasing library "revision" value in the presence of branches?
  * When creating a new branch by forking an already-released branch, would that imply resetting the IS_RELEASE variable? Or do we create the fork at the last commit before the one that sets the IS_RELEASE variable?

jonasnick (commented):
@sipa The rules are described in configure.ac ("The library version is based on libtool versioning of the ABI. The set of rules for updating the version can be found here: https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html") or do you mean copying over the rules to the release-process.md? I thought it's best to have the rules where they version is set because it serves both as a doc of the existing values as well as how to update.

> Another question

That's a good question. I'd like it answered in the release_process.md. How does Bitcoin Core do this? I would have thought everything is based off master except patches to earlier versions are done by branching off the earlier release branch. I think this would only have to be done for patches because it's unlikely we release a x.(y+1).z when (x+1).y.z is already released. Btw, "revision" isn't monotonically increasing, it's set to if there's any change to the interface (including additions).

sipa:
@jonasnick Yes, I meant including the actual rules for incrementing the library version somewhere, not just a reference. I find the rules described in that link kind of vague, and I also had never heard about the concept of library versions at all; if someone is in a similar situation, and just sees the link they may not realize it's actually explaining why it's different from the numbers above.

> That's a good question. I'd like it answered in the release_process.md. How does Bitcoin Core do this? I would have thought everything is based off master except patches to earlier versions are done by branching off the earlier release branch. I think this would only have to be done for patches because it's unlikely we release a x.(y+1).z when (x+1).y.z is already released. Btw, "revision" isn't monotonically increasing, it's set to if there's any change to the interface (including additions).

Bitcoin Core's release process is roughly:
* As a general rule, with very few exceptions, all PRs are made to the master branch.
* Every so often the master branch is forked off to create a major release branch (e.g. the [`22.x`](https://github.com/bitcoin/bitcoin/tree/22.x) branch). After that point, future changes in master *may* be backported to the `22.x` branch if they're bugfixes or somehow specifically relevant for that release. At some point, release candidates `v22.0rc1`, `v22.0rc2`, ... are created from this branch, and finally the final `v22.0` tag is created.
* Minor releases (e.g. `v22.1`) are created by continuing to backport changes (bugfixes, mostly) from the master branch into the `22.x` branch after `v22.0` is released, en then later in the same branch tagging `v22.1rc1`, `v22.1rc2`, ..., and `v22.1`. These can be made for multiple major release branches simultnaneously (e.g. a bugfix in master is sufficiently important that it deserves going into major releases 21 and 20 too, could be backported to all these branches, and then trigger release candidates/releases of new minor releases in all of them.

So the major difference with what you're proposing is that Bitcoin Core only has branches for major releases, not for very release, and that branch is then used for potentially multiple tags.

sipa:
@jonasnick Oh, I noticed that the link in the source is different from the one @real-or-random linked to. That one gives much more concrete instructions already.

jonasnick:
I don't see signficant down or upsides with either branching approach so I'd follow Bitcoin Core's process where applicable because that was the initial intention anyway.

> It's set to what if there is any change to the interface?

Lol, sorry, I have a new (terrible) keyboard. 0.

sipa:
Well, I think there are probably some differences that could justify deviating. For one, I don't think we expect many "major" releases (in the semver sense) after 1.0, certainly not two per year.

jonasnick:
Yes, I would have translated Core major -> secp minor and Core minor -> patch. So we'd make one branch per minor release and tag patch versions.

doc/release-process.md
```diff
@@ -0,0 +1,14 @@
+# Release Process
+
+1. Open PR to master that
+   1. adds release notes to `doc/CHANGELOG.md` and
+   2. if this is **not** a patch release, updates `_PKG_VERSION_{MAJOR,MINOR}` and `_LIB_VERSIONS_*` in `configure.ac`
+2. After the PR is merged,
+   * if this is **not** a patch release, create a release branch with name `MAJOR.MINOR.x`.
```
real-or-random, line 7: nit: So I assume `x` is a literal x here, right, so this is `1.2.x`. Then we could just call it `1.2`. I think this is in fact closer to Core (just `MAJOR.MINOR`), and this is how you usually abbreviate versions: Just drop the parts you don't care about.

real-or-random (approved):
I agree with all the additions, this looks good.

doc/release-process.md
```diff
@@ -0,0 +1,14 @@
+# Release Process
+
+1. Open PR to master that
+   1. adds release notes to `doc/CHANGELOG.md` and
```
t-bast, line 4: In several projects, I've found it quite painful to write all the release notes at once when we actually do a release (the burden of writing the release notes falls on the guy who publishes the release and needs to document other people's additions to the codebase).

Something that worked much better for me was to incrementally update the future release notes: each PR to master that will impact the next release should update the future release notes. This way when we want to publish a release, we only need to do a bit of clean-up, but the release notes are already written and up-to-date. It needs a bit of effort at first to remind contributors to do it, but in the end we usually end up with better release notes.

Feel free to ignore if you think that doesn't make sense for this project, but I felt it was worth sharing in case it can be helpful.

real-or-random: I think in principle I agree but I'm not really sure. My experience in this repo is that it happens quite often that you ask a  contributor to fix a nit before you ACK and then the PR is stalled for months. ;)

jonasnick: Agree on @real-or-random's point but we can still remind regular contributors to do this. I wouldn't want to force this on irregular contributors because it's additional work that discourages from contributing PRs in the first place. We could add a pull request template that optionally requests a changelog update.

One significant downside of this is that pull requests would all edit similar space in the same file which, when one PR is merged, would require rebasing every other PR.

doc/release-process.md
jonasnick: Core creates a MAJOR.x branch (where x is a literal x) according to @sipa's description ([core branches](https://github.com/bitcoin/bitcoin/branches)).

doc/release-process.md
real-or-random: > One significant downside of this is that pull requests would all edit similar space in the same file which, when one PR is merged, would require rebasing every other PR.

Apparently that's a real problem which led to a "crisis" in the GitLab project and needs an entirely non-trivial solution:
https://about.gitlab.com/blog/2018/07/03/solving-gitlabs-changelog-conflict-crisis/

I think then the maintainers should create the changelog... I suppose it will be easier for everyone (including maintainers).

sipa: Bitcoin Core uses `doc/release-notes-<PRNUMBER>.md` files which are added by individual PRs, and are then merged into a release release-notes.md by maintainers.

doc/release-process.md
sipa: I'm fine with either. I think Bitcoin Core went with "22.x" because branches that are just a number are a bit weird, but in the end it's all just arbitrary conventions.

sipa:
LGTM, feel free to squash.

doc/release-process.md
jonasnick: Ok, I removed the .x because it looks nicer.

doc/release-process.md
real-or-random: Just for reference, here's some docs on the process:
https://github.com/bitcoin/bitcoin/blob/23ae7931be50376fa6bda692c641a3d2538556ee/doc/release-process.md#write-the-release-notes
https://github.com/bitcoin/bitcoin/blob/23ae7931be50376fa6bda692c641a3d2538556ee/doc/release-notes.md

This has some helpful notes, e.g., a script to generate a (draft) changelog from git history. This may be helpful to us.

real-or-random (approved):
ACK c8fd83fb834321fcadd062709624f6fb9a95a40d

jonasnick:
rebased

apoelstra:
utACK 3ed0d02bf7690c68ba3ba74da765614809352049

Looks great!

real-or-random (approved):
ACK 3ed0d02bf7690c68ba3ba74da765614809352049

elichai:
ACK b4b02fd8c4276779f115af5985508f54ef9e51be

real-or-random:
@elichai wrong commit :P

elichai:
@real-or-random Ops, somehow I keep getting confused on the order of commits in github (in git the latest is on top, in github it's at the bottom)

ACK 3ed0d02bf7690c68ba3ba74da765614809352049

apoelstra:
@elichai if you have
```
[remote "origin"]
        fetch = +refs/pull/*:refs/remotes/pr/*
```
in your gitconfig, you can do something like `git rev-parse pr/964/head | xsel` to get the top commit ID.

You may still type the wrong PR number or write "ACK" in the wrong browser tab, but at least there's one less mistake you can make :)

sipa:
ACK 3ed0d02bf7690c68ba3ba74da765614809352049

PR #966, opened 2021-07-14
Make aux_rand32 arg to secp256k1_schnorrsig_sign const

sipa:
Since we know the default nonce function does not modify the passed aux randomness, we can mark it const in the API.

### DIFF
--- a/src/modules/schnorrsig/main_impl.h
+++ b/src/modules/schnorrsig/main_impl.h
@@ -185,8 +185,9 @@ int secp256k1_schnorrsig_sign_internal(const secp256k1_context* ctx, unsigned ch
     return ret;
 }
 
-int secp256k1_schnorrsig_sign(const secp256k1_context* ctx, unsigned char *sig64, const unsigned char *msg32, const secp256k1_keypair *keypair, unsigned char *aux_rand32) {
-    return secp256k1_schnorrsig_sign_internal(ctx, sig64, msg32, 32, keypair, secp256k1_nonce_function_bip340, aux_rand32);
+int secp256k1_schnorrsig_sign(const secp256k1_context* ctx, unsigned char *sig64, const unsigned char *msg32, const secp256k1_keypair *keypair, const unsigned char *aux_rand32) {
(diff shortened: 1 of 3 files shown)


src/modules/schnorrsig/main_impl.h
```diff
@@ -185,8 +185,8 @@ int secp256k1_schnorrsig_sign_internal(const secp256k1_context* ctx, unsigned ch
     return ret;
 }
 
-int secp256k1_schnorrsig_sign(const secp256k1_context* ctx, unsigned char *sig64, const unsigned char *msg32, const secp256k1_keypair *keypair, unsigned char *aux_rand32) {
-    return secp256k1_schnorrsig_sign_internal(ctx, sig64, msg32, 32, keypair, secp256k1_nonce_function_bip340, aux_rand32);
+int secp256k1_schnorrsig_sign(const secp256k1_context* ctx, unsigned char *sig64, const unsigned char *msg32, const secp256k1_keypair *keypair, const unsigned char *aux_rand32) {
+    return secp256k1_schnorrsig_sign_internal(ctx, sig64, msg32, 32, keypair, secp256k1_nonce_function_bip340, (unsigned char*)aux_rand32);
```
jonasnick, line 189: Perhaps add a comment here like

> We cast away const from the passed aux_rand32 argument since we know the default nonce function does not modify it.

jonasnick (commented):
You can add `const` to the `aux_rand` argument of `test_schnorrsig_bip_vectors_check_signing`. That way we will see a warning if this commit is reverted.

real-or-random (commented):
ACK mod what Jonas said

src/modules/schnorrsig/main_impl.h
sipa: Done.

sipa:
@jonasnick Done.

jonasnick:
CI fails with
```
cat tests.log || true
./tests: error while loading shared libraries: libasan.so.6: cannot open shared object file: No such file or directory
```

I believe this was fixed in master, so a rebase is likely to fix this.

PR #810, opened 2020-09-10
Avoid overly-wide multiplications in 5x52 field mul/sqr

peterdettman:
Speeds up bench_ecdh, bench_sign, bench_verify relative to master by 5+% at -O3, haswell.

PR #967, opened 2021-07-24
WIP WIP WIP 5x64 field representation

sipa:
This swaps out the 5x52 field with a 5x64 one, including inline assembly for x86_64. It also contains optimized x86_64 asm code by @kn-cs (not tested in this PR).

I'm just opening this to see if it doesn't break anything on the various platforms we have.


sipa:
So to give an idea of the status here:
* The code works, is generally a speedup, and is unlikely to change much; if it does, it's probably restricted to just better asm code
* There is an issue on macOS to figure out (linking fails; I wonder if the asm/linker detect platform-specific instructions, and need some kind of magic runtime detection annotation?)
* As is, this PR removes the old 5x52 representation. That may be undesirable, as at least on some (uncommon) platforms, the new code is (slightly) slower. This may be even more the case with #810 merged now.
* It's an open question how this is supposed to be reviewed or otherwise gained confidence in. I know there are formal methods of proving semantics of asm code, but I don't have experience with them. An alternative is adding lots of high-level description to the asm code to make it readable.
* This PR leaves the decision on which repr/impl to use up to the `./configure` step. That's useful in some settings, and enough to enable testing, but not what we want for normal production usage. I'm fine with that, but perhaps people want a clearer plan on how runtime autodetection etc would work before proceeding?

PR #969, opened 2021-08-19
ci: Fixes after Debian release

real-or-random:
Debian stable is now bullseye. This results in an update of compiler versions, which make some minor fixes necessary. See commit messages for details.

Tip: If you want to build the docker image locally, use `--pull` to force docker to pull the latest Debian image as a base, e.g., `docker build . -f ci/linux-debian.Dockerfile --pull`.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -5262,17 +5262,19 @@ void test_ecdsa_sign_verify(void) {
     secp256k1_scalar msg, key;
     secp256k1_scalar sigr, sigs;
     int getrec;
-    /* Initialize recid to suppress a false positive -Wconditional-uninitialized in clang.
-       VG_UNDEF ensures that valgrind will still treat the variable as uninitialized. */
-    int recid = -1; VG_UNDEF(&recid, sizeof(recid));
+    int recid;
     random_scalar_order_test(&msg);
     random_scalar_order_test(&key);
     secp256k1_ecmult_gen(&ctx->ecmult_gen_ctx, &pubj, &key);
     secp256k1_ge_set_gej(&pub, &pubj);
     getrec = secp256k1_testrand_bits(1);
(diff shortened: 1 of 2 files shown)


src/tests.c
```diff
@@ -5262,17 +5262,19 @@ void test_ecdsa_sign_verify(void) {
...
-    /* Initialize recid to suppress a false positive -Wconditional-uninitialized in clang.
-       VG_UNDEF ensures that valgrind will still treat the variable as uninitialized. */
-    int recid = -1; VG_UNDEF(&recid, sizeof(recid));
+    int recid;
     random_scalar_order_test(&msg);
     random_scalar_order_test(&key);
     secp256k1_ecmult_gen(&ctx->ecmult_gen_ctx, &pubj, &key);
     secp256k1_ge_set_gej(&pub, &pubj);
     getrec = secp256k1_testrand_bits(1);
-    random_sign(&sigr, &sigs, &key, &msg, getrec?&recid:NULL);
```
elichai, line 5273: fwiw, assigning the pointer to a variable also seems to solve this,
Isolated case: https://godbolt.org/z/q1T6eMccT

real-or-random: I see, yeah.   I tried a variant of this, which didn't work. I think both "fixes" are fine, I'm going to keep mine if noone objects. 

Interestingly, your test case shows that the variant appears also on clang 12. But I can't get valgrind to complain on clang 12 output...

jonasnick (commented):
ACK 5d5c74a057f3951677691113747952f4cbdde86b

Can confirm that clang 7, 11 and 12 don't complain.

elichai:
ACK 5d5c74a


It seems that removing the variable initialization and the valgrind macro even clang 12 warns, but this fixes it anyway.

PR #971, opened 2021-09-01
Dummy callbacks for tests

aahadley-anchorage:
If external callbacks are specified, the build can fail at link time since these functions aren't defined yet. Give the linker some definitions to make it happy.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -39,7 +39,7 @@ void ECDSA_SIG_get0(const ECDSA_SIG *sig, const BIGNUM **pr, const BIGNUM **ps) 
 #endif
 
 /* If external callbacks are specified, define these so we can build and test without linker errors */
-#if defined USE_EXTERNAL_DEFAULT_CALLBACKS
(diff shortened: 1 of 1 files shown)


real-or-random:
Hm, that's an issue indeed.

If you use the external callbacks, then you probably do because you don't have `printf` or `abort` on your platform. In tests, we use these function anyway, so I think this change is fine. But wouldn't it be simpler to `#undef USE_EXTERNAL_CALLBACKS`?

Ideally, we'd move into the other direction and make it possible to run the tests without `printf`. But that's more complicated. 

I wonder if could we just abuse the external callbacks for exactly that purpose? Instead of making `CHECK` call abort, we could make this macro reset the default callback and then invoke it. If you don't have `printf` or `abort`, you'd be able provide a different external callback... Except I don't know how to print the seed then or the final success message. Maybe it's just cleaner to let the user provide some print function...

real-or-random:
Closing in favor of #1306

PR #975, opened 2021-09-15
Add a `--enable-dev-mode` to enable everything

elichai:
When modifying the internals I usually want to run all the tests and make sure that I don't break any of the modules.
but I never remember exactly if `module` comes before or after the module name (i.e. `--enable-module-ecdh` vs `--enable-ecdh-module`) and it's a lot to enable all modules.

So I added a short simple flag `--enable-dev-mode` that enables everything.
(If we want to make it more configurable, we can set all default values before, then process `dev-mode`, and then process the rest of the flags without any `action-if-not-given`, so that further flags can override the default dev-mode, but because it's for development I think this is currently enough)

This also fixes the fact that we've assigned `use_experimental` but used `enable_experimental`.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -166,6 +166,18 @@ AC_ARG_ENABLE(external_default_callbacks,
     [use_external_default_callbacks=$enableval],
     [use_external_default_callbacks=no])
 
+AC_ARG_ENABLE(dev_mode,
+    AS_HELP_STRING([--enable-dev-mode],[enables all modules including experimental, for development use only [default=no]]),
+    [use_benchmark=$enableval
+     use_tests=$enableval
+     use_experimental=$enableval
+     use_exhaustive_tests=$enableval
+     enable_module_ecdh=$enableval
+     enable_module_recovery=$enableval
+     enable_module_extrakeys=$enableval
+     enable_module_schnorrsig=$enableval],
+    [])
+
 # Test-only override of the (autodetected by the C code) "widemul" setting.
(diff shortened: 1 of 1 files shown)


configure.ac
```diff
@@ -128,8 +128,8 @@ AC_ARG_ENABLE(openssl_tests,
 
 AC_ARG_ENABLE(experimental,
     AS_HELP_STRING([--enable-experimental],[allow experimental configure options [default=no]]),
-    [use_experimental=$enableval],
-    [use_experimental=no])
+    [enable_experimental=$enableval],
+    [enable_experimental=no])
```
real-or-random, line 132: Hm I was wondering why this has been working then since ever. The reason is that `enable_experimental` is automatically assigned, and so we don't need to set this at all here...

See  https://www.gnu.org/software/autoconf/manual/autoconf-2.70/html_node/Package-Options.html or https://autotools.io/autoconf/arguments.html .

And there's no reason to have sometimes `use` and sometimes `enable`. we may want to do some cleanup in our `configure.ac`. But that does not need to happen in this PR.

real-or-random (commented):
Concept ACK

sipa:
Concept ACK

jonasnick (commented):
concept ACK

(even more useful in secp-zkp)

practicalswift:
Concept ACK

real-or-random:
@elichai I still think this very useful. Are you planning to work on this? If not, maybe someone could take over.

real-or-random:
Closing in favor of #1079.

PR #976, opened 2021-09-15
`secp256k1_schnorrsig_sign_internal` should be static

elichai:
This function isn't used outside of this module so it should be declared static


real-or-random (approved):
ACK 72713872a8597884918bcf1edbc12f5c969ca680

jonasnick (commented):
ACK 72713872a8597884918bcf1edbc12f5c969ca680

sipa:
ACK

PR #977, opened 2021-09-20
Replace sprintf with snprintf to prevent buffer overflow (CWE-120)

GCrispino:
Hello!

This PR replaces `sprintf` calls in https://github.com/bitcoin-core/secp256k1/blob/master/src/bench_ecmult.c for `snprintf`. The reason behind this is that this file triggered (in another repository) a security alert for the common weakness [CWE-120](https://cwe.mitre.org/data/definitions/120.html) regarding buffer overflows when buffer length is not specified, which can be done with `snprintf`.

I'm not sure if this is correct or if this will solve the alerts, so feel free to point out any possible issues. If this is not correct, I can happily create an issue for this and close this PR.

### DIFF
--- a/src/bench_ecmult.c
+++ b/src/bench_ecmult.c
@@ -171,19 +171,20 @@ static void bench_ecmult_2g_teardown(void* arg, int iters) {
 }
 
 static void run_ecmult_bench(bench_data* data, int iters) {
-    char str[32];
-    sprintf(str, "ecmult_gen");
+    int bufsize = 32;
+    char str[bufsize];
+    snprintf(str, bufsize, "ecmult_gen");
     run_benchmark(str, bench_ecmult_gen, bench_ecmult_setup, bench_ecmult_gen_teardown, data, 10, iters);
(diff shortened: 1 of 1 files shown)


jonasnick (commented):
Hi @GCrispino,

just to clarify, the usage of sprintf here can not result in a buffer overflow. I agree that it's generally better use snprintf here (even if it was just for preventing to have this same discussion again). Because it's not part of the C89 standard library, I don't see a way to use snprintf that is not more complex than is justifiable.

GCrispino:
> Hi @GCrispino,
> 
> just to clarify, the usage of sprintf here can not result in a buffer overflow. I agree that it's generally better use snprintf here (even if it was just for preventing to have this same discussion again). Because it's not part of the C89 standard library, I don't see a way to use snprintf that is not more complex than is justifiable.

I see! Honestly I don't know much about the subject, I created this to help others that would maybe have the same issue, but only if it is correct. In this case, feel free to close it, or suggest another approach if you think of any.

Thanks for responding :)

PR #979, opened 2021-09-23
Native jacobi symbol algorithm

sipa:
This introduces variants of the vartime divsteps-based GCD algorithm used for modular inverses to compute Jacobi symbols. Changes compared to the normal vartime divsteps:
* Only positive matrices are used, guaranteeing that f and g remain positive.
* An additional jac variable is updated to track sign changes during matrix computation.
* There is (so far) no proof that this algorithm terminates within reasonable amount of time for every input, but experimentally it appears to almost always need less than 900 iterations. To account for that, only a bounded number of iterations is performed (1500), after which failure is returned. The field logic then falls back to using square roots to determining the result.
* The algorithm converges to f=g=gcd(f0,g0) rather than g=0. To keep this test simple, the end condition is f=1, which won't be reached if started with g=0. That case is dealt with specially.

This code is currently unused, except for tests. I don't aim for it to be merged until there is a need for it, but this demonstrates its feasibility.

In terms of performance:
```
field_inverse: min 1.76us / avg 1.76us / max 1.78us
field_inverse_var: min 0.991us / avg 0.993us / max 0.996us
field_jacobi_var: min 1.31us / avg 1.31us / max 1.31us
field_sqrt: min 4.36us / avg 4.37us / max 4.40us
```

while with the older (f24e122d13db7061b1086ddfd21d3a1c5294213b) libgmp based Jacobi code on the same system:
```
num_jacobi: min 1.53us / avg 1.54us / max 1.55us
```

### DIFF
--- a/doc/safegcd_implementation.md
+++ b/doc/safegcd_implementation.md
@@ -1,7 +1,7 @@
 # The safegcd implementation in libsecp256k1 explained
 
-This document explains the modular inverse implementation in the `src/modinv*.h` files. It is based
-on the paper
+This document explains the modular inverse and Jacobi symbol implementations in the `src/modinv*.h` files.
+It is based on the paper
 ["Fast constant-time gcd computation and modular inversion"](https://gcd.cr.yp.to/papers.html#safegcd)
 by Daniel J. Bernstein and Bo-Yin Yang. The references below are for the Date: 2019.04.13 version.
 
@@ -769,3 +769,51 @@ def modinv_var(M, Mi, x):
         d, e = update_de(d, e, t, M, Mi)
     return normalize(f, d, Mi)
 ```
+
+## 8. From GCDs to Jacobi symbol
+
+We can also use a similar approach to calculate Jacobi symbol *(x | M)* by keeping track of an
+extra variable *j*, for which at every step *(x | M) = j (g | f)*. As we update *f* and *g*, we
+make corresponding updates to *j* using
+[properties of the Jacobi symbol](https://en.wikipedia.org/wiki/Jacobi_symbol#Properties):
+* *((g/2) | f)* is either *(g | f)* or *-(g | f)*, depending on the value of *f mod 8* (negating if it's *3* or *5*).
+* *(f | g)* is either *(g | f)* or *-(g | f)*, depending on *f mod 4* and *g mod 4* (negating if both are *3*).
+
+These updates depend only on the values of *f* and *g* modulo *4* or *8*, and can thus be applied
+very quickly, as long as we keep track of a few additional bits of *f* and *g*. Overall, this
+calculation is slightly simpler than the one for the modular inverse because we no longer need to
+keep track of *d* and *e*.
+
+However, one difficulty of this approach is that the Jacobi symbol *(a | n)* is only defined for
+positive odd integers *n*, whereas in the original safegcd algorithm, *f, g* can take negative
+values. We resolve this by using the following modified steps:
+
+```python
+        # Before
+        if delta > 0 and g & 1:
+            delta, f, g = 1 - delta, g, (g - f) // 2
+
+        # After
+        if delta > 0 and g & 1:
+            delta, f, g = 1 - delta, g, (g + f) // 2
+```
+
+The algorithm is still correct, since the changed divstep, called a "posdivstep" (see section 8.4
+and E.5 in the paper) preserves *gcd(f, g)*. However, there's no proof that the modified algorithm
+will converge. The justification for posdivsteps is completely empirical: in practice, it appears
+that the vast majority of nonzero inputs converge to *f=g=gcd(f<sub>0</sub>, g<sub>0</sub>)* in a
+number of steps proportional to their logarithm.
+
+Note that:
+- We require inputs to satisfy *gcd(x, M) = 1*, as otherwise *f=1* is not reached.
+- We require inputs *x &neq; 0*, because applying posdivstep with *g=0* has no effect.
+- We need to update the termination condition from *g=0* to *f=1*.
+
+We account for the possibility of nonconvergence by only performing a bounded number of
+posdivsteps, and then falling back to square-root based Jacobi calculation if a solution has not
+yet been found.
+
+The optimizations in sections 3-7 above are described in the context of the original divsteps, but
+in the C implementation we also adapt most of them (not including "avoiding modulus operations",
+since it's not necessary to track *d, e*, and "constant-time operation", since we never calculate
+Jacobi symbols for secret data) to the posdivsteps version.
--- a/src/bench_internal.c
+++ b/src/bench_internal.c
@@ -219,6 +219,19 @@ static void bench_field_sqrt(void* arg, int iters) {
     CHECK(j <= iters);
 }
 
+static void bench_field_is_square_var(void* arg, int iters) {
+    int i, j = 0;
+    bench_inv *data = (bench_inv*)arg;
+    secp256k1_fe t = data->fe[0];
+
+    for (i = 0; i < iters; i++) {
+        j += secp256k1_fe_is_square_var(&t);
+        secp256k1_fe_add(&t, &data->fe[1]);
+        secp256k1_fe_normalize_var(&t);
+    }
+    CHECK(j <= iters);
+}
+
 static void bench_group_double_var(void* arg, int iters) {
     int i;
     bench_inv *data = (bench_inv*)arg;
@@ -371,6 +384,7 @@ int main(int argc, char **argv) {
     if (d || have_flag(argc, argv, "field") || have_flag(argc, argv, "mul")) run_benchmark("field_mul", bench_field_mul, bench_setup, NULL, &data, 10, iters*10);
     if (d || have_flag(argc, argv, "field") || have_flag(argc, argv, "inverse")) run_benchmark("field_inverse", bench_field_inverse, bench_setup, NULL, &data, 10, iters);
     if (d || have_flag(argc, argv, "field") || have_flag(argc, argv, "inverse")) run_benchmark("field_inverse_var", bench_field_inverse_var, bench_setup, NULL, &data, 10, iters);
+    if (d || have_flag(argc, argv, "field") || have_flag(argc, argv, "issquare")) run_benchmark("field_is_square_var", bench_field_is_square_var, bench_setup, NULL, &data, 10, iters);
     if (d || have_flag(argc, argv, "field") || have_flag(argc, argv, "sqrt")) run_benchmark("field_sqrt", bench_field_sqrt, bench_setup, NULL, &data, 10, iters);
 
     if (d || have_flag(argc, argv, "group") || have_flag(argc, argv, "double")) run_benchmark("group_double_var", bench_group_double_var, bench_setup, NULL, &data, 10, iters*10);
--- a/src/field_5x52_impl.h
+++ b/src/field_5x52_impl.h
@@ -664,4 +664,31 @@ static void secp256k1_fe_inv_var(secp256k1_fe *r, const secp256k1_fe *x) {
 #endif
 }
 
+static int secp256k1_fe_is_square_var(const secp256k1_fe *x) {
+    secp256k1_fe tmp;
+    secp256k1_modinv64_signed62 s;
+    int jac, ret;
+
+    tmp = *x;
+    secp256k1_fe_normalize_var(&tmp);
+    /* secp256k1_jacobi64_maybe_var cannot deal with input 0. */
+    if (secp256k1_fe_is_zero(&tmp)) return 1;
+    secp256k1_fe_to_signed62(&s, &tmp);
+    jac = secp256k1_jacobi64_maybe_var(&s, &secp256k1_const_modinfo_fe);
+    if (jac == 0) {
+        /* secp256k1_jacobi64_maybe_var failed to compute the Jacobi symbol. Fall back
+         * to computing a square root. This should be extremely rare with random
+         * input (except in VERIFY mode, where a lower iteration count is used). */
+        secp256k1_fe dummy;
+        ret = secp256k1_fe_sqrt(&dummy, &tmp);
+    } else {
+#ifdef VERIFY
+        secp256k1_fe dummy;
+        VERIFY_CHECK(jac == 2*secp256k1_fe_sqrt(&dummy, &tmp) - 1);
+#endif
+        ret = jac >= 0;
+    }
+    return ret;
+}
+
 #endif /* SECP256K1_FIELD_REPR_IMPL_H */
--- a/src/int128_struct_impl.h
+++ b/src/int128_struct_impl.h
@@ -189,10 +189,11 @@ static SECP256K1_INLINE int secp256k1_i128_eq_var(const secp256k1_int128 *a, con
    return a->hi == b->hi && a->lo == b->lo;
 }
 
-static SECP256K1_INLINE int secp256k1_i128_check_pow2(const secp256k1_int128 *r, unsigned int n) {
-   VERIFY_CHECK(n < 127);
-   return n >= 64 ? r->hi == (uint64_t)1 << (n - 64) && r->lo == 0
-                  : r->hi == 0 && r->lo == (uint64_t)1 << n;
+static SECP256K1_INLINE int secp256k1_i128_check_pow2(const secp256k1_int128 *r, unsigned int n, int sign) {
+    VERIFY_CHECK(n < 127);
+    VERIFY_CHECK(sign == 1 || sign == -1);
+    return n >= 64 ? r->hi == (uint64_t)sign << (n - 64) && r->lo == 0
+                   : r->hi == (uint64_t)((sign - 1) >> 1) && r->lo == (uint64_t)sign << n;
 }
 
 #endif
--- a/src/modinv32.h
+++ b/src/modinv32.h
@@ -35,4 +35,9 @@ static void secp256k1_modinv32_var(secp256k1_modinv32_signed30 *x, const secp256
 /* Same as secp256k1_modinv32_var, but constant time in x (not in the modulus). */
 static void secp256k1_modinv32(secp256k1_modinv32_signed30 *x, const secp256k1_modinv32_modinfo *modinfo);
 
+/* Compute the Jacobi symbol for (x | modinfo->modulus). x must be coprime with modulus (and thus
+ * cannot be 0, as modulus >= 3). All limbs of x must be non-negative. Returns 0 if the result
+ * cannot be computed. */
+static int secp256k1_jacobi32_maybe_var(const secp256k1_modinv32_signed30 *x, const secp256k1_modinv32_modinfo *modinfo);
+
 #endif /* SECP256K1_MODINV32_H */
--- a/src/modinv32_impl.h
+++ b/src/modinv32_impl.h
@@ -232,6 +232,21 @@ static int32_t secp256k1_modinv32_divsteps_30(int32_t zeta, uint32_t f0, uint32_
     return zeta;
 }
 
+/* inv256[i] = -(2*i+1)^-1 (mod 256) */
+static const uint8_t secp256k1_modinv32_inv256[128] = {
+    0xFF, 0x55, 0x33, 0x49, 0xC7, 0x5D, 0x3B, 0x11, 0x0F, 0xE5, 0xC3, 0x59,
+    0xD7, 0xED, 0xCB, 0x21, 0x1F, 0x75, 0x53, 0x69, 0xE7, 0x7D, 0x5B, 0x31,
+    0x2F, 0x05, 0xE3, 0x79, 0xF7, 0x0D, 0xEB, 0x41, 0x3F, 0x95, 0x73, 0x89,
+    0x07, 0x9D, 0x7B, 0x51, 0x4F, 0x25, 0x03, 0x99, 0x17, 0x2D, 0x0B, 0x61,
+    0x5F, 0xB5, 0x93, 0xA9, 0x27, 0xBD, 0x9B, 0x71, 0x6F, 0x45, 0x23, 0xB9,
+    0x37, 0x4D, 0x2B, 0x81, 0x7F, 0xD5, 0xB3, 0xC9, 0x47, 0xDD, 0xBB, 0x91,
+    0x8F, 0x65, 0x43, 0xD9, 0x57, 0x6D, 0x4B, 0xA1, 0x9F, 0xF5, 0xD3, 0xE9,
+    0x67, 0xFD, 0xDB, 0xB1, 0xAF, 0x85, 0x63, 0xF9, 0x77, 0x8D, 0x6B, 0xC1,
+    0xBF, 0x15, 0xF3, 0x09, 0x87, 0x1D, 0xFB, 0xD1, 0xCF, 0xA5, 0x83, 0x19,
+    0x97, 0xAD, 0x8B, 0xE1, 0xDF, 0x35, 0x13, 0x29, 0xA7, 0x3D, 0x1B, 0xF1,
+    0xEF, 0xC5, 0xA3, 0x39, 0xB7, 0xCD, 0xAB, 0x01
+};
+
 /* Compute the transition matrix and eta for 30 divsteps (variable time).
  *
  * Input:  eta: initial eta
@@ -243,21 +258,6 @@ static int32_t secp256k1_modinv32_divsteps_30(int32_t zeta, uint32_t f0, uint32_
  * Implements the divsteps_n_matrix_var function from the explanation.
  */
 static int32_t secp256k1_modinv32_divsteps_30_var(int32_t eta, uint32_t f0, uint32_t g0, secp256k1_modinv32_trans2x2 *t) {
-    /* inv256[i] = -(2*i+1)^-1 (mod 256) */
-    static const uint8_t inv256[128] = {
-        0xFF, 0x55, 0x33, 0x49, 0xC7, 0x5D, 0x3B, 0x11, 0x0F, 0xE5, 0xC3, 0x59,
-        0xD7, 0xED, 0xCB, 0x21, 0x1F, 0x75, 0x53, 0x69, 0xE7, 0x7D, 0x5B, 0x31,
-        0x2F, 0x05, 0xE3, 0x79, 0xF7, 0x0D, 0xEB, 0x41, 0x3F, 0x95, 0x73, 0x89,
-        0x07, 0x9D, 0x7B, 0x51, 0x4F, 0x25, 0x03, 0x99, 0x17, 0x2D, 0x0B, 0x61,
-        0x5F, 0xB5, 0x93, 0xA9, 0x27, 0xBD, 0x9B, 0x71, 0x6F, 0x45, 0x23, 0xB9,
-        0x37, 0x4D, 0x2B, 0x81, 0x7F, 0xD5, 0xB3, 0xC9, 0x47, 0xDD, 0xBB, 0x91,
-        0x8F, 0x65, 0x43, 0xD9, 0x57, 0x6D, 0x4B, 0xA1, 0x9F, 0xF5, 0xD3, 0xE9,
-        0x67, 0xFD, 0xDB, 0xB1, 0xAF, 0x85, 0x63, 0xF9, 0x77, 0x8D, 0x6B, 0xC1,
-        0xBF, 0x15, 0xF3, 0x09, 0x87, 0x1D, 0xFB, 0xD1, 0xCF, 0xA5, 0x83, 0x19,
-        0x97, 0xAD, 0x8B, 0xE1, 0xDF, 0x35, 0x13, 0x29, 0xA7, 0x3D, 0x1B, 0xF1,
-        0xEF, 0xC5, 0xA3, 0x39, 0xB7, 0xCD, 0xAB, 0x01
-    };
-
     /* Transformation matrix; see comments in secp256k1_modinv32_divsteps_30. */
     uint32_t u = 1, v = 0, q = 0, r = 1;
     uint32_t f = f0, g = g0, m;
@@ -297,7 +297,7 @@ static int32_t secp256k1_modinv32_divsteps_30_var(int32_t eta, uint32_t f0, uint
         VERIFY_CHECK(limit > 0 && limit <= 30);
         m = (UINT32_MAX >> (32 - limit)) & 255U;
         /* Find what multiple of f must be added to g to cancel its bottom min(limit, 8) bits. */
-        w = (g * inv256[(f >> 1) & 127]) & m;
+        w = (g * secp256k1_modinv32_inv256[(f >> 1) & 127]) & m;
         /* Do so. */
         g += f * w;
         q += u * w;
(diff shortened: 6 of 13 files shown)


robot-dreams (commented):
tACK 042f91bf4acca0e1d70ca6afffe045a525a7b614

`./tests` passed, including the additional checks I suggested.

I'm still brushing up on the divsteps algorithm, and will review this change again once I've worked through the various details / optimizations.

---

Benchmark results look good on x86_64 (MacBook Pro): native jacobi version is faster than libgmp. However, on arm64 (Mac M1), native jacobi version was slightly *slower*!

On M1 I also tried `./configure  --with-asm=arm --enable-experimental` to see if that would speed up the native jacobi version, but attempting to build resulted in a lot of compile errors in `field_10x26_arm.s`.

### MacBook Pro (x86_64)

**native jacobi branch**

```
$ ./bench_internal field
...
field_jacobi_var: min 2.03us / avg 2.25us / max 3.85us
field_sqrt: min 6.33us / avg 6.38us / max 6.46us

$ ./bench_internal jacobi
field_jacobi_var: min 2.00us / avg 2.06us / max 2.15us
```

**libgmp branch** f24e122d13db7061b1086ddfd21d3a1c5294213b:

```
$ ./bench_internal field
...
field_sqrt: min 6.05us / avg 6.13us / max 6.38us

$ ./bench_internal num
num_jacobi: min 2.44us / avg 2.47us / max 2.53us
```

### Mac M1 (arm64)

**native jacobi branch**

```
% ./bench_internal field
...
field_jacobi_var: min 1.54us / avg 1.55us / max 1.56us
field_sqrt: min 4.04us / avg 4.09us / max 4.29us

% ./bench_internal jacobi
field_jacobi_var: min 1.55us / avg 1.74us / max 3.18us
```

**libgmp branch** f24e122d13db7061b1086ddfd21d3a1c5294213b:

```
% ./bench_internal field
...
field_sqrt: min 4.04us / avg 4.04us / max 4.05us

% ./bench_internal jacobi
...
num_jacobi: min 1.41us / avg 1.41us / max 1.42us
```

real-or-random:
@robot-dreams Always nice to see new contributors in the repo. :) Just curious, are you simply helping or are you interested in getting this merged? If the latter, I'm further curious about your use case. Feel free just ignore my intrusive questions.

robot-dreams:
@real-or-random Thanks for the reply! I'm simply helping and try to learn about this repo, I don't have a personal use case.

src/modinv32.h
```diff
@@ -39,4 +39,7 @@ static void secp256k1_modinv32_var(secp256k1_modinv32_signed30 *x, const secp256
 /* Same as secp256k1_modinv32_var, but constant time in x (not in the modulus). */
 static void secp256k1_modinv32(secp256k1_modinv32_signed30 *x, const secp256k1_modinv32_modinfo *modinfo);
 
+/* Compute the Jacobi symbol for x (where gcd(x,p) == 1), 0 if it cannot be determined. */
+static int secp256k1_jacobi32_maybe_var(const secp256k1_modinv32_signed30 *x, const secp256k1_modinv32_modinfo *modinfo);
```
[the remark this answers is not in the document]
sipa: Good idea. Done.

src/tests.c
```diff
@@ -811,12 +811,29 @@ void test_modinv32_uint16(uint16_t* out, const uint16_t* in, const uint16_t* mod
...
+        uint16_to_signed30(&x, sqr);
+        CHECK(secp256k1_jacobi32_maybe_var(&x, &m) == nonzero);
+        if ((mod[0] & 3) == 3) {
+            /* If mod % 4 == 3, x and -x have opposite jacobi symbols. */
+            negone[0] = mod[0] - 1;
+            for (i = 1; i < 16; ++i) negone[i] = mod[i];
+            mulmod256(sqr, sqr, negone, mod);
+            uint16_to_signed30(&x, sqr);
+            CHECK(secp256k1_jacobi32_maybe_var(&x, &m) == -nonzero);
+        }
```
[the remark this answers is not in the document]
sipa: Nice catch, done. I was hoping to have a test that guarantees finding a -1 result, but that's hard without a per-modulus search for a non-square factor. This is pretty useful too, though.

src/field_5x52_impl.h
```diff
@@ -575,4 +575,25 @@ static void secp256k1_fe_inv_var(secp256k1_fe *r, const secp256k1_fe *x) {
...
+    secp256k1_fe_to_signed62(&s, &tmp);
+    ret = secp256k1_jacobi64_maybe_var(&s, &secp256k1_const_modinfo_fe);
+    if (ret == 0) {
+        /* secp256k1_jacobi64_maybe_var failed to compute the Jacobi symbol. Fall back
+         * to computing a square root. This should be extremely rare with random
+         * input. */
+        secp256k1_fe dummy;
+        ret = 2*secp256k1_fe_sqrt(&dummy, &tmp) - 1;
+    }
+    return ret;
```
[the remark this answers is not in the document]
sipa: Done.

sipa:
@robot-dreams Thanks for your comments!

The failure with asm=arm is expected, as Apple M1 is 64-bit ARM, while the assembly code used here is 32-bit.

sipa:
Made another change: in VERIFY mode, the upper bound on the number of posdivsteps iterations is reduced to ~750 (which is close to the median of how many iterations are needed for random input mod the field size), so that the uncomputable path is actually exercised in the tests.

src/modinv64_impl.h
```diff
@@ -294,6 +294,94 @@ static int64_t secp256k1_modinv64_divsteps_62_var(int64_t eta, uint64_t f0, uint
...
+            eta = -eta;
+            tmp = f; f = g; g = tmp;
+            tmp = u; u = q; q = tmp;
+            tmp = v; v = r; r = tmp;
+            /* Update bottom bit of jac: when swapping f and g, the Jacobi symbol changes sign
+             * if both f and g are 3 mod 4. */
+            jac ^= ((f & g) >> 1);
+            /* Use a formula to cancel out up to 6 bits of g. Also, no more than i can be cancelled
+             * out (as we'd be done before that point), and no more than eta+1 can be done as its
+             * will flip again once that happens. */
```
robot-dreams, line 345: Nit: `sign will flip`

robot-dreams (commented):
ACK 6e3fcc4f5149c96ab052f3312091dfc0b8eab397

I agree with your modifications to the original `safegcd` for calculating Jacobi symbol instead of modular inverse. I also replicated the logic for the 64-bit version in Python, and checked that the result matches a "slow" Jacobi function for a few randomly generated inputs (see https://gist.github.com/robot-dreams/ceb00162b80384f2ae1913aaa2b35e75).

One potential concern is that with this change, there are now 4 slightly different implementations of a similar algorithm (32/64-bit, inv/jacobi). Do you think there's a nice way to consolidate the logic, at least among a given word size? (If there's no nice way or if it's too much trouble, I'm fine with leaving it as is and relying on having good tests.)

Other questions:

- What was the motivation for using different approaches to clear the low-order bits of `g` between the 32 and 64 bit implementations (8-bit lookup table vs 4 and 6-bit tricks from Hacker's Delight)?

- I agree that replacing `f, g = g, -f` with `f, g = g, f` ensures that all matrix entries are always positive. But aside from actually observing that it converges to `f = 1` in practice, do you have any other intuitive justification for this? (The best I can come up with is that since you're constantly dividing `g` by 2 and swapping the roles of `f` and `g`, you're quickly moving towards a solution even if you replace `g - f` with `g + f` in the original GCD algorithm.)

- Is there already a writeup describing this change? If not, would it make sense to add a brief section to https://github.com/bitcoin-core/secp256k1/blob/master/doc/safegcd_implementation.md? (If you think so, I'd be happy to give that a shot.)

robot-dreams:
@sipa Thanks for responding to the comments!

- Yes, 32-bit makes sense (in retrospect the fact that the file was called `field_10x26` instead of `field_5x52` should have given that away)
- Great idea to reduce the number of iterations for validation!

sipa:
> One potential concern is that with this change, there are now 4 slightly different implementations of a similar algorithm (32/64-bit, inv/jacobi). Do you think there's a nice way to consolidate the logic, at least among a given word size? (If there's no nice way or if it's too much trouble, I'm fine with leaving it as is and relying on having good tests.)

Arguably there are 6 implementations (32/64-bit, and const/var/varpos divsteps), and they do share some of the code (in particular the matrix application), but not everything. I don't think there is an obvious way of merging more code without reducing performance, but I also haven't tried very hard.

> What was the motivation for using different approaches to clear the low-order bits of g between the 32 and 64 bit implementations (8-bit lookup table vs 4 and 6-bit tricks from Hacker's Delight)?

Just dumb benchmarking. The formula was faster on x86_64 systems we tried, and the table was faster on ARMv7 systems (it's explained in one of the commit messages that introduced it).

> I agree that replacing f, g = g, -f with f, g = g, f ensures that all matrix entries are always positive. But aside from actually observing that it converges to f = 1 in practice, do you have any other intuitive justification for this? (The best I can come up with is that since you're constantly dividing g by 2 and swapping the roles of f and g, you're quickly moving towards a solution even if you replace g - f with g + f in the original GCD algorithm.)

I don't, no. I do have a proof that any f>0,g>0 converges to f=g somewhere in a finite number of steps, but the upper bound is something like `f*g*log2(f*g)`, which is obviously useless for practical purposes. The technique explained in https://github.com/sipa/safegcd-bounds completely fails for posdivsteps (doesn't converge at all). With a lot of tricks I got it to converge, but in what appears quadratic (`O(log2(f)^2)`), and it's only tractable to evaluate for f up to 2^20 or so.

The difficulty is that there are both steps that take f and g further apart (g /= 2) and steps that bring them closer (g = (f+g)/2), and it's hard to infer tight bounds on how many times one can happen in between two executions of the other.

So in short, no; the actual argument why this works is purely empirical; all out of a 2.5 billion random inputs (with the field size modulus) need between 630 and 921 iterations.

> Is there already a writeup describing this change? If not, would it make sense to add a brief section to https://github.com/bitcoin-core/secp256k1/blob/master/doc/safegcd_implementation.md? (If you think so, I'd be happy to give that a shot.)

I think that would be valuable, yes. I haven't done so yet, but was considering it. Feel free to take a shot at it.

src/modinv64_impl.h
sipa: Done.

src/modinv32_impl.h
```diff
@@ -317,6 +317,83 @@ static int32_t secp256k1_modinv32_divsteps_30_var(int32_t eta, uint32_t f0, uint
...
+        /* Update the bottom bit of jac: when dividing g by an odd power of 2,
+         * if (f mod 8) is 3 or 5, the Jacobi symbol changes sign. */
+        jac ^= (zeros & ((f >> 1) ^ (f >> 2)));
+        /* We're done once we've done 60 posdivsteps. */
+        if (i == 0) break;
+        VERIFY_CHECK((f & 1) == 1);
+        VERIFY_CHECK((g & 1) == 1);
+        VERIFY_CHECK((u * f0 + v * g0) == f << (30 - i));
+        VERIFY_CHECK((q * f0 + r * g0) == g << (30 - i));
+        /* If eta is negative, negate it and replace f,g with g,-f. */
```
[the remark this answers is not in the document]
sipa: Done.

src/modinv64_impl.h
```diff
@@ -294,6 +294,94 @@ static int64_t secp256k1_modinv64_divsteps_62_var(int64_t eta, uint64_t f0, uint
...
+        /* Update the bottom bit of jac: when dividing g by an odd power of 2,
+         * if (f mod 8) is 3 or 5, the Jacobi symbol changes sign. */
+        jac ^= (zeros & ((f >> 1) ^ (f >> 2)));
+        /* We're done once we've done 60 posdivsteps. */
+        if (i == 0) break;
+        VERIFY_CHECK((f & 1) == 1);
+        VERIFY_CHECK((g & 1) == 1);
+        VERIFY_CHECK((u * f0 + v * g0) == f << (62 - i));
+        VERIFY_CHECK((q * f0 + r * g0) == g << (62 - i));
+        /* If eta is negative, negate it and replace f,g with g,-f. */
```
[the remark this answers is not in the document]
sipa: Done.

src/modinv32_impl.h
```diff
@@ -317,6 +317,83 @@ static int32_t secp256k1_modinv32_divsteps_30_var(int32_t eta, uint32_t f0, uint
...
+        /* Perform zeros divsteps at once; they all just divide g by two. */
+        g >>= zeros;
+        u <<= zeros;
+        v <<= zeros;
+        eta -= zeros;
+        i -= zeros;
+        /* Update the bottom bit of jac: when dividing g by an odd power of 2,
+         * if (f mod 8) is 3 or 5, the Jacobi symbol changes sign. */
+        jac ^= (zeros & ((f >> 1) ^ (f >> 2)));
+        /* We're done once we've done 60 posdivsteps. */
```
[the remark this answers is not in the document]
sipa: Done.

src/modinv64_impl.h
```diff
@@ -294,6 +294,94 @@ static int64_t secp256k1_modinv64_divsteps_62_var(int64_t eta, uint64_t f0, uint
...
+        /* Perform zeros divsteps at once; they all just divide g by two. */
+        g >>= zeros;
+        u <<= zeros;
+        v <<= zeros;
+        eta -= zeros;
+        i -= zeros;
+        /* Update the bottom bit of jac: when dividing g by an odd power of 2,
+         * if (f mod 8) is 3 or 5, the Jacobi symbol changes sign. */
+        jac ^= (zeros & ((f >> 1) ^ (f >> 2)));
+        /* We're done once we've done 60 posdivsteps. */
```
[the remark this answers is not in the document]
sipa: Done.

sipa:
Included @robot-dreams's update to the writeup.

src/modinv32_impl.h
```diff
@@ -584,4 +661,69 @@ static void secp256k1_modinv32_var(secp256k1_modinv32_signed30 *x, const secp256
     *x = d;
 }
 
+/* Compute the Jacobi symbol of x modulo modinfo->modulus (variable time). gcd(x,modulus) must be 1. */
```
robot-dreams, line 664: Nits (also applies to `modinv64_impl`):
- Is `gcd(x,modulus) must be 1` still true, since you're handling `x=0` here?
- Is it necessary to also mention `x must be normalized` here for consistency with `modinv32.h`?

robot-dreams (commented):
ACK 5af3d290b6eabd6dd5cb2565caab24c0ceb8b8f9 aside from comment nits—I checked the range diff against the old commit and it looks good.

src/modinv32_impl.h
sipa: I made a few changes to address this.

robot-dreams:
ACK 53b7ce28bef8008787c1aeb202ec17ef3a3b3325

sipa:
https://eprint.iacr.org/2021/1271 should be a much smaller changes, and faster. Closing.

src/bench_internal.c
```diff
@@ -218,6 +218,17 @@ void bench_field_sqrt(void* arg, int iters) {
...
+void bench_field_jacobi_var(void* arg, int iters) {
+    int i, j = 0;
+    bench_inv *data = (bench_inv*)arg;
+
+    for (i = 0; i < iters; i++) {
+        j += secp256k1_fe_jacobi_var(&data->fe[0]);
+        secp256k1_fe_add(&data->fe[0], &data->fe[1]);
+    }
+    CHECK(j <= iters);
+}
```
jonasnick, line 230: `./bench_internal jacobi` crashes on my machine when configured with CPPFLAGS=-DVERIFY. Backtrace:
```
src/field_5x52_impl.h:57: test condition failed: r == 1

Program received signal SIGABRT, Aborted.
0x00007ffff7e3fbc7 in __pthread_kill_implementation () from /nix/store/hsk71z8admvgykn7vzjy11dfnar9f4r1-glibc-2.35-163/lib/libc.so.6
(gdb) bt
#0  0x00007ffff7e3fbc7 in __pthread_kill_implementation () from /nix/store/hsk71z8admvgykn7vzjy11dfnar9f4r1-glibc-2.35-163/lib/libc.so.6
#1  0x00007ffff7df2b46 in raise () from /nix/store/hsk71z8admvgykn7vzjy11dfnar9f4r1-glibc-2.35-163/lib/libc.so.6
#2  0x00007ffff7ddd4b5 in abort () from /nix/store/hsk71z8admvgykn7vzjy11dfnar9f4r1-glibc-2.35-163/lib/libc.so.6
#3  0x0000000000402920 in secp256k1_fe_verify (a=<optimized out>) at src/field_5x52_impl.h:57
#4  0x000000000040c1f8 in secp256k1_fe_add (a=0x99, r=0x69) at src/field_5x52_impl.h:443
#5  bench_field_jacobi_var (arg=arg@entry=0x7fffffffcf20, iters=iters@entry=20000) at src/bench_internal.c:227
#6  0x0000000000413d5a in run_benchmark (name=0x416221 "field_jacobi_var", benchmark=0x40bdd0 <bench_field_jacobi_var>, setup=0x40cd70 <bench_setup>,
    teardown=0x0, data=0x7fffffffcf20, count=10, iter=20000) at src/bench.h:109
#7  0x000000000040147d in main (argc=2, argv=0x7fffffffd798) at src/bench_internal.c:393
```

src/tests.c
```diff
@@ -2854,8 +2894,10 @@ void run_sqrt(void) {
         for (j = 0; j < count; j++) {
             random_fe(&x);
             secp256k1_fe_sqr(&s, &x);
+            CHECK(secp256k1_fe_jacobi_var(&s) == 1);
             test_sqrt(&s, &x);
             secp256k1_fe_negate(&t, &s, 1);
+            CHECK(secp256k1_fe_jacobi_var(&t) == -1);
```
jonasnick, line 2900: Is it possible that that `secp256k1_fe_jacobi_var(&s)` is taking the slow `secp256k1_fe_sqrt` branch about 50% of the time? I discovered this on my machine with this simple patch that adds a printf:

```
diff --git a/src/field_5x52_impl.h b/src/field_5x52_impl.h
index 26e89123..d3eb3347 100644
--- a/src/field_5x52_impl.h
+++ b/src/field_5x52_impl.h
@@ -681,6 +681,7 @@ static int secp256k1_fe_jacobi_var(const secp256k1_fe *x) {
          * to computing a square root. This should be extremely rare with random
          * input. */
         secp256k1_fe dummy;
+        printf("slow");
         ret = 2*secp256k1_fe_sqrt(&dummy, &tmp) - 1;
 #ifdef VERIFY
     } else {
```

I'm quite confused because should be "extremely rare with random input", but `s` appears to be a random field element. Furthermore, "slow" isn't printed in the jacobi benchmarks unless I compile with `-DVERIFY` (which crashes after a few iterations, see other comment).

jonasnick: Ah ok, that's good. I did not looked into `secp256k1_jacobi64_maybe_var` so far and hadn't considered that its behavior would change on VERIFY.

src/field_5x52_impl.h
```diff
@@ -667,4 +667,32 @@ static void secp256k1_fe_inv_var(secp256k1_fe *r, const secp256k1_fe *x) {
...
+    tmp = *x;
+    secp256k1_fe_normalize_var(&tmp);
+    secp256k1_fe_to_signed62(&s, &tmp);
+    ret = secp256k1_jacobi64_maybe_var(&s, &secp256k1_const_modinfo_fe);
+    if (ret == -2) {
+        /* secp256k1_jacobi64_maybe_var failed to compute the Jacobi symbol. Fall back
+         * to computing a square root. This should be extremely rare with random
+         * input. */
+        secp256k1_fe dummy;
+        ret = 2*secp256k1_fe_sqrt(&dummy, &tmp) - 1;
```
real-or-random, line 684: Should instead have a function  `fe_is_square_var`? 

I think this will be better because it
 * avoids the return value translations here and below 
 * is more ergonomic to use (I think, remember that we had it this in the old version of BIP340 too),
 * and it aligns with the uses in ellswift: in the corresponding PR, we only test `>=0` and `<0`, so we actually don't even need the special casing for `==0` currently.

This is related to my comment below about the `jacp` function argument.

src/modinv64_impl.h
```diff
@@ -301,11 +303,100 @@ static int64_t secp256k1_modinv64_divsteps_62_var(int64_t eta, uint64_t f0, uint
...
+ * Output: t: transition matrix
+ * Return: final eta
+ */
+static int64_t secp256k1_modinv64_posdivsteps_62_var(int64_t eta, uint64_t f0, uint64_t g0, secp256k1_modinv64_trans2x2 *t, int *jacp) {
+    /* Transformation matrix; see comments in secp256k1_modinv64_divsteps_62. */
+    uint64_t u = 1, v = 0, q = 0, r = 1;
+    uint64_t f = f0, g = g0, m;
+    uint32_t w;
+    int i = 62, limit, zeros;
+    int jac = *jacp;
```
real-or-random, line 327:
```suggestion
    int jac = *jacp;
    VERIFY_CHECK(-1 <= jac && jac <= 1);
```

src/modinv64_impl.h
```diff
@@ -301,11 +303,100 @@ static int64_t secp256k1_modinv64_divsteps_62_var(int64_t eta, uint64_t f0, uint
...
     return eta;
 }
 
+/* Compute the transition matrix and eta for 62 posdivsteps (variable time, eta=-delta), and keeps track
+ * of the Jacobi symbol along the way. f0 and g0 must be f and g mod 2^64 rather than 2^62, because
+ * Jacobi tracking requires knowing (f mod 8) rather than just (f mod 2).
+ *
+ * Input:  eta: initial eta
+ *         f0:  bottom limb of initial f
+ *         g0:  bottom limb of initial g
```
real-or-random, line 317: should document the input/output parameter `jacp`. (And the variable name is confusing. This is not the jacobi symbol but its sign.)

src/modinv64_impl.h
```diff
@@ -626,4 +717,71 @@ static void secp256k1_modinv64_var(secp256k1_modinv64_signed62 *x, const secp256
     *x = d;
 }
 
+/* Do up to 25 iterations of 62 posdivsteps (up to 1550 steps; more is extremely rare) each until f=1.
+ * In VERIFY mode use a lower number of iterations (744, close to the median 756), so failure actually occurs. */
+#ifdef VERIFY
+#define JACOBI64_ITERATIONS 12
+#else
+#define JACOBI64_ITERATIONS 25
+#endif
```
real-or-random, line 726: On the one hand, I really like this trick. But on the other hand, it creates a difference between VERIFY and non-VERIFY (and who knows, maybe this influences loop unrolling etc.)

Would it be a good idea to keep this 25 in the real code but create a test-only version of it (either in the tests file or here) that just runs with 12? Or maybe better, make the number of iterations an argument of the function? Since it's a "maybe" function, it should be correct for each iteration value.

real-or-random (commented):
This is really good, I have almost only nits.

note to self: I haven't looked at the tests for secp256k1_jacobi32/64_maybe_var yet

src/modinv32_impl.h
```diff
@@ -317,6 +317,83 @@ static int32_t secp256k1_modinv32_divsteps_30_var(int32_t eta, uint32_t f0, uint
...
+ * of the Jacobi symbol along the way. f0 and g0 must be f and g mod 2^32 rather than 2^30, because
+ * Jacobi tracking requires knowing (f mod 8) rather than just (f mod 2).
+ *
+ * Input:  eta: initial eta
+ *         f0:  bottom limb of initial f
+ *         g0:  bottom limb of initial g
+ * Output: t: transition matrix
+ * Return: final eta
+ */
+static int32_t secp256k1_modinv32_posdivsteps_30_var(int32_t eta, uint32_t f0, uint32_t g0, secp256k1_modinv32_trans2x2 *t, int *jacp) {
```
real-or-random, line 330: Since this is so similar to the other function, have you considered using macros to avoid code duplication? (I'm not saying it's necessarily better.)

real-or-random: Makes sense!

src/modinv64_impl.h
real-or-random: Ok, I got it after reading your new comments.

src/modinv64_impl.h
real-or-random: > I feel like VERIFY mode isn't to test the compiler's correctness, but to test the correctness of our source code.

That's the way I see it. too. [1] I think the fundamental problem here is that we don't really run tests in non-VERIFY mode. I'll comment in https://github.com/bitcoin-core/secp256k1/issues/1037, which is somewhat related.

 > A copy in tests could go out of sync,  [...]

Agreed, that's bad. 

> If we don't like it, I think the best alternative is just having maintainers occasionally manually setting the iteration count lower, and seeing if tests still pass.

That's even worse, we'll forget to run it.

> and having the iteration count be an argument still means somewhere higher in the stack there needs to be a decision what to set it to, which I think has the same problems.

> So... I'm not sure. I do see the uneasiness about it, but I feel that the current approach is really the only one that guarantees (a) it's actually exercised in realistic settings and (b) using the same code as is used in production. 

Hm, yeah, I don't think (a) is a big concern here: What this VERIFY tests is just handling the loop termination in case of too many iterations, and that's a few lines of code which doesn't care too much if the failure happened in an realistic input or not. So what I imagined it is that we run it for values of the iteration count in the test, and this will be good enough to exercise this code.

This would have the advantage that it minimizes the difference between VERIFY and non-VERIFY mode. But after thinking more about it, I agree that it's a somewhat moot goal. [2] Your comment about loop unrolling is right and we have a lot of code locations where we add some substantial amount of code in VERIFY (you introduce another one here in this PR, which I don't mind at all.)  

[1] Though to some extent, VERIFY can also help checking compiler, e.g., when a function is miscompiles and its result is VERIFIED somewhere else.
[2] I think my brain mixed up my thoughts with the discussion in https://github.com/bitcoin-core/secp256k1/pull/1169#pullrequestreview-1224682895,  but this was about SECP256K1_CHECKMEM_ENABLED being defined, not VERIFY.

src/field_5x52_impl.h
real-or-random: makes sense

src/modinv64_impl.h
```diff
@@ -71,11 +71,13 @@ static int secp256k1_modinv64_mul_cmp_62(const secp256k1_modinv64_signed62 *a, i
     return 0;
 }
 
-/* Check if the determinant of t is equal to 1 << n. */
-static int secp256k1_modinv64_det_check_pow2(const secp256k1_modinv64_trans2x2 *t, unsigned int n) {
+/* Check if the determinant of t is equal to 1 << n. If abs, |det| == 1 << n. */
```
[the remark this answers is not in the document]
real-or-random: This one is still unresolved -- (Feel free to dismiss, I'm mostly checking if the PR is ready for review again.)

doc/safegcd_implementation.md
```diff
@@ -769,3 +769,30 @@ def modinv_var(M, Mi, x):
         d, e = update_de(d, e, t, M, Mi)
     return normalize(f, d, Mi)
 ```
+
+## 8. From GCDs to Jacobi symbol
+
+We can also use a similar approach to calculate Jacobi symbol *(x | M)* by keeping track of an extra variable *j*, for which at every step *(x | M) = j (g | f)*. As we update *f* and *g*, we make corresponding updates to *j* using [properties of the Jacobi symbol](https://en.wikipedia.org/wiki/Jacobi_symbol#Properties). In particular, we update *j* whenever we divide *g* by *2* or swap *f* and *g*; these updates depend only on the values of *f* and *g* modulo *4* or *8*, and can thus be applied very quickly. Overall, this calculation is slightly simpler than the one for modular inverse because we no longer need to keep track of *d* and *e*.
```
real-or-random, line 775: It wouldn't hurt to spell the two properties out here for self-containedness.

src/modinv32.h
```diff
@@ -35,4 +35,8 @@ static void secp256k1_modinv32_var(secp256k1_modinv32_signed30 *x, const secp256
 /* Same as secp256k1_modinv32_var, but constant time in x (not in the modulus). */
 static void secp256k1_modinv32(secp256k1_modinv32_signed30 *x, const secp256k1_modinv32_modinfo *modinfo);
 
+/* Compute the Jacobi symbol for (x | modinfo->modulus). Either x must be 0, or x must be coprime with
```
real-or-random, line 38: maybe this should spell out that the Jacobi symbol is only defined for odd x? (Do we need x > 1?)

real-or-random (commented):
Some final comments you can address or ignore. 

I'm running some testing locally and will then ACK if it's successful.

real-or-random (approved):
ACK ef2ac449d757189b67be27437824c2d0c0663bde diff and writeup is good and I tested every commit

src/modinv32.h
real-or-random: Okay yes, I don't immediately see why. If this is the case, then it should probably be explained.

(This can be done later as you said below.)

real-or-random:
needs (trivial) rebase

real-or-random (approved):
reACK 642b9f42  diff and writeup is good and I tested every commit

src/tests.c
```diff
@@ -1929,30 +1929,56 @@ static void run_int128_test_case(void) {
...
         }
-        CHECK(secp256k1_i128_check_pow2(&swz, pos) == expect);
+        CHECK(secp256k1_i128_check_pow2(&swz, pos, 1) == expect);
+    }
+    /* test secp256k1_i128_check_pow2 (sign == -1) */
+    {
+        int expect = (uc & 1);
+        int pos = ub % 127;
+        if (expect) {
+            /* If expect==1, set swz to exactly -(2 << pos). */
```
siv2r, line 2002: Shouldn't it be `-2^pos` instead of `-(2 << pos)`? Similarly, in the lines: 1973, 1937, 1947.

src/modinv32.h
jonasnick, line 38: Nit: The jacobi symbol (0|1) is commonly defined to be 1 (see e.g. [wikipedia](https://en.wikipedia.org/wiki/Jacobi_symbol)), but `secp256k1_jacobi_maybe_var` returns 0 in that case. I think it would be valuable to document `secp256k1_jacobi_maybe_var`'s behavior in this edge case.

jonasnick (commented):
ACK mod nit

I was able to [implement the algorithm](https://gist.github.com/jonasnick/beca92a9c18088c03146bf9dbb3d5afa) based on the descriptions and python snippets in `safegcd_implementation.md` and compare it against the C implementation.

src/modinv32.h
real-or-random: That's related to what I said there: https://github.com/bitcoin-core/secp256k1/pull/979#discussion_r1062661604. Though I don't understand why I said there that I don't see why the output is trivial. It's trivial, of course. Maybe what I meant is that I don't see immediately why the output is correct, which matches your comment here, @jonasnick.

Perhaps it's better to simply exclude (0|1) as a valid input, and then document and VERIFY_CHECK that. That makes it easier to recognize this corner case in future code. (And I feel it's hard to predict what definition of jac(0|1) future code would expect to work. It's better to make callers remind of the corner case explicitly.)

src/modinv32.h
```diff
@@ -35,4 +35,8 @@ static void secp256k1_modinv32_var(secp256k1_modinv32_signed30 *x, const secp256
 /* Same as secp256k1_modinv32_var, but constant time in x (not in the modulus). */
 static void secp256k1_modinv32(secp256k1_modinv32_signed30 *x, const secp256k1_modinv32_modinfo *modinfo);
 
+/* Compute the Jacobi symbol for (x | modinfo->modulus). x must be coprime with modulus (and thus cannot be 0).
```
jonasnick, line 38: Isn't 0 coprime with 1 (and no other positive integer)? The gcd is 1, since 1 divides 0 and 1 divides 1.

src/modinv32.h
real-or-random: Yes, that can be considered resolved. 

> The `secp256k1_modinv32_modinfo` requires an odd modulus already.

I wonder if it's a good idea to add a function `secp256k1_modinv32_modinfo` verify that checks consistency and call it on entry of every function that reads modinfo, similar to how we do this for other data structures. But if yes, that should happen in a separate PR.

src/modinv32.h
jonasnick: Maybe I'm missing something but this line still says `x coprime with modulus => x != 0`.

jonasnick:
reACK mod nit

jonasnick:
ACK ce3cfc78a6020d21be299e1e4f22cf8ef089194d

real-or-random (approved):
reACK ce3cfc78a6020d21be299e1e4f22cf8ef089194d diff and writeup is good and I tested every commit

src/int128_struct_impl.h
```diff
@@ -189,10 +189,11 @@ static SECP256K1_INLINE int secp256k1_i128_eq_var(const secp256k1_int128 *a, con
...
 
-static SECP256K1_INLINE int secp256k1_i128_check_pow2(const secp256k1_int128 *r, unsigned int n) {
-   VERIFY_CHECK(n < 127);
-   return n >= 64 ? r->hi == (uint64_t)1 << (n - 64) && r->lo == 0
-                  : r->hi == 0 && r->lo == (uint64_t)1 << n;
+static SECP256K1_INLINE int secp256k1_i128_check_pow2(const secp256k1_int128 *r, unsigned int n, int sign) {
+    VERIFY_CHECK(n < 127);
+    VERIFY_CHECK(sign == 1 || sign == -1);
+    return n >= 64 ? r->hi == (uint64_t)sign << (n - 64) && r->lo == 0
+                   : r->hi == (uint64_t)((sign - 1) >> 1) && r->lo == (uint64_t)sign << n;
```
roconnor-blockstream, line 196: There is no need to subtract 1 before doing a right shift.

real-or-random: solved in https://github.com/bitcoin-core/secp256k1/pull/1250

Issue #869, opened 2021-01-19
OpenSSL 3 will deprecate some of the functions we use in tests


rex4539:
```
/Users/rex/secp256k1 on branch master rex@MacBook-Pro-2018% ./autogen.sh 
glibtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'build-aux'.
glibtoolize: copying file 'build-aux/ltmain.sh'
glibtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'build-aux/m4'.
glibtoolize: copying file 'build-aux/m4/libtool.m4'
glibtoolize: copying file 'build-aux/m4/ltoptions.m4'
glibtoolize: copying file 'build-aux/m4/ltsugar.m4'
glibtoolize: copying file 'build-aux/m4/ltversion.m4'
glibtoolize: copying file 'build-aux/m4/lt~obsolete.m4'
configure.ac:15: installing 'build-aux/compile'
configure.ac:5: installing 'build-aux/config.guess'
configure.ac:5: installing 'build-aux/config.sub'
configure.ac:9: installing 'build-aux/install-sh'
configure.ac:9: installing 'build-aux/missing'
Makefile.am: installing 'build-aux/depcomp'
parallel-tests: installing 'build-aux/test-driver'
/Users/rex/secp256k1 on branch master rex@MacBook-Pro-2018% ./configure 
checking build system type... x86_64-apple-darwin20.3.0
checking host system type... x86_64-apple-darwin20.3.0
checking for a BSD-compatible install... /opt/local/bin/ginstall -c
checking whether build environment is sane... yes
checking for a thread-safe mkdir -p... /opt/local/bin/gmkdir -p
checking for gawk... gawk
checking whether make sets $(MAKE)... yes
checking whether make supports nested variables... yes
checking how to print strings... printf
checking whether make supports the include directive... yes (GNU style)
checking for gcc... gcc
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether gcc accepts -g... yes
checking for gcc option to accept ISO C89... none needed
checking whether gcc understands -c and -o together... rm: conftest.dSYM: is a directory
yes
checking dependency style of gcc... gcc3
checking for a sed that does not truncate output... /opt/local/bin/gsed
[...]

PR #983, opened 2021-09-28
[RFC] Remove OpenSSL testing support

sipa:
This removes the ability to test against OpenSSL, as well as the OpenSSL verification benchmark.

The motivation is that OpenSSL 3 is deprecating part of the API used here (see #869), and I'm not sure it's worth maintaining. We do lose the fact that this is the only test that verifies randomly-generated cases against an independent implementation. On the other hand, there are tons of existing fixed tests now that test all kinds of edge cases already.


real-or-random:
Hm yeah, this has a history. I'd probably be fine with removing this but when we last wanted to get rid of the OpenSSL tests (https://github.com/bitcoin-core/secp256k1/pull/635), the conclusion was that we rather want to replace it by some other implementation (https://github.com/bitcoin-core/secp256k1/issues/691). @elichai has two old PRs with different implementations (https://github.com/bitcoin-core/secp256k1/pull/738 and https://github.com/bitcoin-core/secp256k1/pull/641). Now there's also https://github.com/LLFourn/secp256kfun by @LLFourn, which would be another good option (but needs Rust).

I also suggested removing the benchmarks in https://github.com/bitcoin-core/secp256k1/issues/736. This issue didn't get a lot of feedback but I assume removing the benchmark is not controversial.

gmaxwell:
This looks good to me.  Cross-testing is important, but introduction of a new one shouldn't block removal of won whos interface instability has become a maintenance burden (particularly when  the alternative proposals add a rust dependency, which I think is extremely unattractive).

real-or-random (changes requested):
Concept ACK 

Now with #984, the tests even fail. This means that not only the API of OpenSSL has changed v3 but also the behavior. This makes the tests unsuitable for us. 

This PR should also remove this change to `Makefile.am` https://github.com/bitcoin-core/secp256k1/pull/735/files#diff-0462e381b2fb3286568215681c8983490a37ac9ae0f0c5ee304df7fa6426d4af

fanquake:
Concept ACK

jonasnick:
I agree that cross-testing is useful. I don't see such a problem with adapting the code calling into a library after a major release. But fine with me to remove it for now. Since the alternatives for cross-testing have tradeoffs as well, the easiest solution may be to reintroduce openssl 3 at some point. Perhaps once it is more popular. I don't have easy access to v3 right now to play with it.

Another quick fix would be to check that `OPENSSL_VERSION_NUMBER` in the `AC_COMPILE_IFELSE` statement of `SECP_OPENSSL_CHECK` is less than 3.

The commit lgtm.

sipa:
@real-or-random Added this:

```diff
 bench_verify_LDADD = libsecp256k1.la $(SECP_LIBS) $(SECP_TEST_LIBS) $(COMMON_LIB)
-# SECP_TEST_INCLUDES are only used here for CRYPTO_CPPFLAGS
-bench_verify_CPPFLAGS = $(SECP_TEST_INCLUDES)
 bench_sign_SOURCES = src/bench_sign.c
```

luke-jr:
Could just restrict it to the older version of OpenSSL?

jonasnick (commented):
ACK bc08599e776aff33c834ef829843ec5f629d1f39

Trinacristella (commented):
I'm new to a bunch of this but I know I can succeed

PR #987, opened 2021-10-04
Fix unused parameter warnings when building without VERIFY

jonasnick:
This commit makes `./configure --enable-coverage && make check` free of warnings.


practicalswift:
cr ACK 189f6bcfef6578b89e21f937b24060f74bd18f00

PR #988, opened 2021-10-08
Make signing table fully static

### LINKED ISSUE #692: Don't put an absurd amount of data onto the stack in some configs

### LINKED ISSUE #893: Fully static precomputation tables
Currently we offer `secp256k1_context_no_precomp` [1] but it's a "none" context.

If static precomputation (for ecmult_gen contexts) is enabled, we could easily offer a static `secp256k1_context_sign`. I believe we should do this and additionally offer static precomputation for ecmult contexts. This will make it possible to offer a static `secp256k1_context_verify` and a fully static  `secp256k1_context_sign_verify` depending on the compilation flags. 

I think this is very helpful for embedded targets and is probably much more convenient to use than the prealloc API [2], which requires to users to obtain the size of the context `secp256k1_context_preallocated_size()` *at runtime* and deal with manual allocation etc.

[1] Note that the docs can be improved here. They describe the context in terms of internals ("precomputed tables").  We should also describe this in terms of equivalent context flags. The same is true for the name of this thing. 
 https://github.com/bitcoin-core/secp256k1/blob/8c727b9087ae012d5bb77e0766c21da763561a3c/include/secp256k1.h#L186-L191
[2] Also with behind language bindings #892, see also @gmaxwell's comment there advocating for a fully static context

real-or-random:
This resolves #893,  resolves #692 (and also resolves bitcoin/bitcoin#22854).

- [x] Extract table generation to separate function in separate file (to be used by generation script and exhaustive tests)
- [x] Tidy up 
  - [x] Remove code that deals with non-static tables
  - [x] Make functions that need ecmult_gen not depend on signing context
  - [x] Rename stuff to make it fit the new structure and consistent with how we hande verification tables (#956)
- [x] Fix exhaustive tests
  - [x] Make table generation function take generator as input
  - [x] Overwrite the static tables with a table with custom generator in exhaustive tests
- [x] Overhaul script that generates table files
  - [x] Make table generation function take PREC_BITS as input (I have some code already, just not yet in this branch) 
  - [x] Change generation script to generate three tables (for all three values of ECMULT_GEN_PREC_BITS) 
- [x] Ship pre-built tables
  - [x] Add pregenerated table file to repo
  - [x] Remove generation of table file from build process (like in #956)
  - [x] Remove left-over stuff (e.g., detecting a compiler running on the build machine) from build system
- [x] Final cleanups (copyright headers, commit, messages, etc.)
- [ ] (separate PR:) Make sure link-time optimization remove corresponding static tables (and code) when no signing/verifcation function is called
- [ ] (separate PR:) Compile precomputation as a separate object file and link it (https://github.com/bitcoin-core/secp256k1/pull/988#issuecomment-977813538)
- [ ] (separate PR:) Document the backwards-compatible API changes made in this PR and in #956. 
  - [ ] Maybe deprecate the static context


fanquake:
Concept ACK from a Core POV. Thanks for following up with this.

real-or-random:
Ready for review. Note that this changes the stdout of `make` due to the use of `BUILT_SOURCES` in Automake (https://www.gnu.org/software/automake/manual/html_node/Sources.html) but that's purely cosmetic.

src/gen_ecmult_gen_static_prec_table.c
```diff
@@ -51,38 +44,42 @@ int main(int argc, char **argv) {
...
-    table = checked_malloc(&default_error_callback, ECMULT_GEN_PREC_TABLE_SIZE);
-    secp256k1_ecmult_gen_create_prec_table(table, &secp256k1_ge_const_g, bits);
-
-    for(outer = 0; outer != n; outer++) {
-        fprintf(fp,"{\n");
-        for(inner = 0; inner != g; inner++) {
-            fprintf(fp,"    SC(%uu, %uu, %uu, %uu, %uu, %uu, %uu, %uu, %uu, %uu, %uu, %uu, %uu, %uu, %uu, %uu)", SECP256K1_GE_STORAGE_CONST_GET(table[outer * g + inner]));
-            if (inner != g - 1) {
-                fprintf(fp,",\n");
+    for (bits = 2; bits <= 8; bits += 2) {
```
robot-dreams, line 60: What's the motivation for `bits += 2` instead of `bits *= 2` here, given that `ecmult_gen.h` has the following?

```
#if ECMULT_GEN_PREC_BITS != 2 && ECMULT_GEN_PREC_BITS != 4 && ECMULT_GEN_PREC_BITS != 8
#  error "Set ECMULT_GEN_PREC_BITS to 2, 4 or 8."
#endif
```

Does either this loop or the check in `ecmult_gen.h` need to be updated?

src/ecmult_gen.h
```diff
@@ -18,28 +18,13 @@
...
-     *   * U_i = U * 2^i, for i=0 ... PREC_N-2
-     *   * U_i = U * (1-2^(PREC_N-1)), for i=PREC_N-1
-     *   where U is a point with no known corresponding scalar. Note that sum(U_i, i=0 ... PREC_N-1) = 0.
-     * For each i, and each of the PREC_G possible values of n_i, (n_i * (PREC_G)^i * G + U_i) is
-     * precomputed (call it prec(i, n_i)). The formula now becomes sum(prec(i, n_i), i=0 ... PREC_N-1).
-     * None of the resulting prec group elements have a known scalar, and neither do any of
-     * the intermediate sums while computing a*G.
-     */
-    secp256k1_ge_storage (*prec)[ECMULT_GEN_PREC_N][ECMULT_GEN_PREC_G]; /* prec[j][i] = (PREC_G)^j * i * G + U_i */
+    int built;
```
robot-dreams, line 21: Nit: Is it worth having a comment with the definition of `built`?

src/ecmult_gen_impl.h
```diff
@@ -14,35 +14,33 @@
...
 static int secp256k1_ecmult_gen_context_is_built(const secp256k1_ecmult_gen_context* ctx) {
-    return ctx->prec != NULL;
-}
-
-static void secp256k1_ecmult_gen_context_finalize_memcpy(secp256k1_ecmult_gen_context *dst, const secp256k1_ecmult_gen_context *src) {
-    (void)dst, (void)src;
+    return ctx->built;
 }
 
 static void secp256k1_ecmult_gen_context_clear(secp256k1_ecmult_gen_context *ctx) {
```
robot-dreams, line 26: Does `ctx->built` need to be reset here?

robot-dreams (commented):
Thanks for making this really readable with atomic commits!

I still need to review this more carefully but here are some initial comments:

src/gen_ecmult_gen_static_prec_table.c
sipa: Just noticed this too; it adds a table for ECMULT_GEN_PREC_BITS=6, which isn't allowed.

sipa:
A suggestion for a follow-up PR: move all precomputed tables to a separate .c/.o file, so it can be compiled just once and included in all binaries?

src/gen_ecmult_gen_static_prec_table.c
real-or-random: Lol, will fix this.  That's great, so the file is even shorter!

src/ecmult_gen.h
real-or-random: Well, if you ask for it, then probably yes!

src/ecmult_gen_impl.h
real-or-random: Yes, fixing.

real-or-random:
> A suggestion for a follow-up PR: move all precomputed tables to a separate .c/.o file, so it can be compiled just once and included in all binaries?

Yep, I think separate .c/o. files for the ecmult vs ecmult_gen precomputation are still useful but linking is probably the way to go. But yes, this should go to a follow-up PR. I'm adding it to the list in the opening comment. This PR just makes the ecmult_gen tables use the same build method as the ecmult tables already use.

src/gen_ecmult_gen_static_prec_table.c
real-or-random: Went down from 1.8 MB to 1.4 MB

real-or-random:
Force-pushed to fix the initial comments by @robot-dreams.

src/ecmult_gen_prec.h
```diff
@@ -0,0 +1,16 @@
...
+ * Distributed under the MIT software license, see the accompanying    *
+ * file COPYING or https://www.opensource.org/licenses/mit-license.php.*
+ ***********************************************************************/
+
+#ifndef SECP256K1_ECMULT_GEN_PREC_H
+#define SECP256K1_ECMULT_GEN_PREC_H
+
+#include "ecmult_gen.h"
+
+static const size_t ECMULT_GEN_PREC_TABLE_SIZE = ECMULT_GEN_PREC_N(ECMULT_GEN_PREC_BITS) * ECMULT_GEN_PREC_G(ECMULT_GEN_PREC_BITS) * sizeof(secp256k1_ge_storage);
```
robot-dreams, line 12: Nit: I couldn't find any references to this; should it be either (1) removed or (2) be a `#define` with a `bits` parameter so that it can be used in `gen_ecmult_gen_static_prec_table.c`?

.gitignore
```diff
@@ -3,7 +3,7 @@ bench_ecmult
 bench_internal
 tests
 exhaustive_tests
-gen_context
+gen_ecmult_gen_static_prec_table
```
robot-dreams, line 6: It's too bad that `gen` stands for both "generate" (the static table) and "generator" (of a group) 😅 , but it's probably not worth changing, especially not as part of this PR.

robot-dreams (commented):
ACK f0318588ea5ec27dd2e6a074688e22f2bfae2b47 aside from nits.

I didn't review changes to build, CI, or test files; however, tests pass locally for me, and `make clean-precomp; make precomp` keeps the repo unchanged.

I also verified the precomputed tables against an independent implementation.

sipa (commented):
This is looking very good. I want to go over the test and allocation changes another time before I ACK, but I don't expect any issues.

src/modules/extrakeys/tests_impl.h
```diff
@@ -359,25 +359,32 @@ void test_keypair(void) {
     secp256k1_context *none = api_test_context(SECP256K1_CONTEXT_NONE, &ecount);
     secp256k1_context *sign = api_test_context(SECP256K1_CONTEXT_SIGN, &ecount);
     secp256k1_context *verify = api_test_context(SECP256K1_CONTEXT_VERIFY, &ecount);
+    secp256k1_context *sttc = secp256k1_context_clone(secp256k1_context_no_precomp);
```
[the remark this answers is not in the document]
real-or-random: Yeah, it's just "static". The tests in the other modules all use four letter names `none`, `sign`, `vrfy` (except the one here which has `verify`. -- I could change this), so I tried to stick to this convention. I could change this to `static_ctx` oder `no_precomp` or similar if you prefer. Unfortunately `static` itself is a keyword. ;)

.gitignore
real-or-random: Yeah, this confused me a lot when starting to work on the library a few years back. And `gen_context` should have been called `gen_gen_context` at least, even though that's ugly.

I tried to avoid this discussion in this PR (only renaming thins I'm touching anyway) and it may be worth unifying the names in the follow-up PR that switches to linking instead of including (e.g., I intentionally used "create" in a function name instead of "generate") Then we anyway need to touch both "generate" things again, so this would be a good opportunity.

src/gen_ecmult_gen_static_prec_table.c
```diff
@@ -0,0 +1,95 @@
...
+
+    fprintf(fp, "#define S(a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p) "
+            "SECP256K1_GE_STORAGE_CONST(0x##a##u,0x##b##u,0x##c##u,0x##d##u,0x##e##u,0x##f##u,0x##g##u,"
+            "0x##h##u,0x##i##u,0x##j##u,0x##k##u,0x##l##u,0x##m##u,0x##n##u,0x##o##u,0x##p##u)\n");
+
+    fprintf(fp, "static\n");
+    fprintf(fp, "#ifndef EXHAUSTIVE_TEST_ORDER\n");
+    fprintf(fp, "const\n");
+    fprintf(fp, "#endif\n");
+    fprintf(fp, "secp256k1_ge_storage secp256k1_ecmult_gen_prec_table[ECMULT_GEN_PREC_N(ECMULT_GEN_PREC_B)][ECMULT_GEN_PREC_G(ECMULT_GEN_PREC_B)] = {\n");
```
real-or-random, line 58: We shouldn't just make this non-const for the exhaustive test but also omit the initializer entirely. No need to have the static table in the exhaustive test binary just to overwrite it then.

real-or-random:
I forced-push to address the comments. I conversed the state before the rebase (old PR branch + fixups here): https://github.com/real-or-random/secp256k1/tree/202108-prebuilt-gen-context-fixups1 for easier re-review.

src/ecmult_gen.h
```diff
@@ -13,33 +13,20 @@
...
-     *   where U is a point with no known corresponding scalar. Note that sum(U_i, i=0 ... PREC_N-1) = 0.
-     * For each i, and each of the PREC_G possible values of n_i, (n_i * (PREC_G)^i * G + U_i) is
-     * precomputed (call it prec(i, n_i)). The formula now becomes sum(prec(i, n_i), i=0 ... PREC_N-1).
-     * None of the resulting prec group elements have a known scalar, and neither do any of
-     * the intermediate sums while computing a*G.
-     */
-    secp256k1_ge_storage (*prec)[ECMULT_GEN_PREC_N][ECMULT_GEN_PREC_G]; /* prec[j][i] = (PREC_G)^j * i * G + U_i */
-    secp256k1_scalar blind;
-    secp256k1_gej initial;
+    /* Whether the context has been built. */
```
robot-dreams, line 20: Looks like my previous comment got automatically hidden because it was in an already resolved conversation: https://github.com/bitcoin-core/secp256k1/pull/988/files#r757712377

In retrospect I shouldn't have asked for such a comment 😅 apologies for going back and forth, can you please either remove the comment after all, or if you prefer, replace it with something like this:

```c
/* Whether the blinding values have been generated. */
```

robot-dreams (commented):
ACK f26d43dad65c8aafdba057dde9cfa1adb4b58012, my remaining inline comments are non-blocking.

I re-reviewed and also looked at build, CI, test to the extent I could (the only things I didn't review this time are the CC_FOR_BUILD and USE_EXTERNAL_ASM commits).

I also checked that the following now works (no more warning about stack switching):

```
./configure --with-valgrind=yes --with-ecmult-gen-precision=8
make
valgrind ./exhaustive_tests
```

In a future PR, does it make sense to simplify `secp256k1_context_preallocated_size` and `secp256k1_context_preallocated_clone_size` since the size will always be the same, and the only use case now is to check that the flags are valid?

src/ecmult_gen.h
real-or-random: Hm, to be honest, I think the current comment is more precise. E.g., if you reset the blinding values, `built` will still be 1 but it's debatable whether the blinding values have been generated. I do think the current comment is helpful but of course I also don't insist on it and can remove it. :D

real-or-random: Ok, I simply kept the current comment.

src/modules/schnorrsig/tests_impl.h
```diff
@@ -157,42 +160,44 @@ void test_schnorrsig_api(void) {
...
+    CHECK(ecount == 3);
     CHECK(secp256k1_schnorrsig_sign_custom(sign, sig, msg, sizeof(msg), &invalid_keypair, &extraparams) == 0);
-    CHECK(ecount == 6);
+    CHECK(ecount == 4);
     CHECK(secp256k1_schnorrsig_sign_custom(sign, sig, msg, sizeof(msg), &keypairs[0], NULL) == 1);
-    CHECK(ecount == 6);
+    CHECK(ecount == 4);
     CHECK(secp256k1_schnorrsig_sign_custom(sign, sig, msg, sizeof(msg), &keypairs[0], &invalid_extraparams) == 0);
-    CHECK(ecount == 7);
+    CHECK(ecount == 5);
```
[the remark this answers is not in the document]
real-or-random: Fixed.

real-or-random:
Pushed another fixup to https://github.com/real-or-random/secp256k1/tree/202108-prebuilt-gen-context-fixups1 to address the missing line in the schnorrsig test and force-pushed the rebased version here.

fanquake:
I put together a Bitcoin Core [branch](https://github.com/fanquake/bitcoin/tree/test_secp256k1_988) to test these changes. It's master (https://github.com/bitcoin/bitcoin/commit/927a9b07776881d99340259f214c577537dbc77f) with the previous version of this PR (rebased on secp256k1 master) pulled into the subtree, and https://github.com/bitcoin/bitcoin/pull/23383/commits/89277969d7dca605b5d8a9097d19344a27876afc & https://github.com/bitcoin/bitcoin/pull/23383/commits/89e24fa6a6708f0e6a08e633f44070f73c480775 cherry-picked from https://github.com/bitcoin/bitcoin/pull/23383.

Checked that we no-longer have any issues like https://github.com/bitcoin/bitcoin/issues/22854. Also tested on an arm64 (M1) box.

Full Guix build also works (on x86_64):
```bash
real	134m9.592s
user	640m24.791s
sys	43m20.544s

bash-5.1# find guix-build-$(git rev-parse --short=12 HEAD)/output/ -type f -print0 | env LC_ALL=C sort -z | xargs -r0 sha256sum
de610013c9fc9f2c0ad54247deea5a481aeaf294f85b29da994e1c214af2eb20  guix-build-13027d952519/output/aarch64-linux-gnu/SHA256SUMS.part
05e26dc745d94168463ef2a82efbc8b180553919121e94fd753f7f6e11fc88d4  guix-build-13027d952519/output/aarch64-linux-gnu/bitcoin-13027d952519-aarch64-linux-gnu-debug.tar.gz
ea2abd132dab4eeb0d675d35548ca9aad3a1b83853fd0d4a24e9fd399cff3159  guix-build-13027d952519/output/aarch64-linux-gnu/bitcoin-13027d952519-aarch64-linux-gnu.tar.gz
dc1deace6cf7d392cfc8cf82fca35a9639cf9dcd30883ed0a50758ea3816ec03  guix-build-13027d952519/output/arm-linux-gnueabihf/SHA256SUMS.part
0fe3ad1c97be391038707fb8e76a2c60cc7666d40efbaf10a2b2a5084deb9b8a  guix-build-13027d952519/output/arm-linux-gnueabihf/bitcoin-13027d952519-arm-linux-gnueabihf-debug.tar.gz
a2015503d8aac931cb45047bd0096f0c5d04778889e5ac843458253a314787c2  guix-build-13027d952519/output/arm-linux-gnueabihf/bitcoin-13027d952519-arm-linux-gnueabihf.tar.gz
26d7af59f95300cd0933e288c5f84e4ba053b77716b3173152e93de32c890103  guix-build-13027d952519/output/dist-archive/bitcoin-13027d952519.tar.gz
08c6a552396d2b580d58169edec6992e21e51674a83927cf83e5dc417f469d3c  guix-build-13027d952519/output/powerpc64-linux-gnu/SHA256SUMS.part
7801a8c55d5aa424ac7dc412ed4532c311d05a4debc9ef8476c3537383240cab  guix-build-13027d952519/output/powerpc64-linux-gnu/bitcoin-13027d952519-powerpc64-linux-gnu-debug.tar.gz
e63ad72998c898b90a70e8916d4f0c87bc504b880059ac119ae5723fa26ea718  guix-build-13027d952519/output/powerpc64-linux-gnu/bitcoin-13027d952519-powerpc64-linux-gnu.tar.gz
b1e272e1bf6c8ea5155a943bfe6598c450d4ca3c7223324209ad58ed9366682c  guix-build-13027d952519/output/powerpc64le-linux-gnu/SHA256SUMS.part
a52b47f5ef50d43552a2050d9434c0ee48d61c0b765df063370e8f4da8113366  guix-build-13027d952519/output/powerpc64le-linux-gnu/bitcoin-13027d952519-powerpc64le-linux-gnu-debug.tar.gz
cd9d1beb51534ae31cd7d8c9c6479985b942369a5bab1ba4aad698331ba1930c  guix-build-13027d952519/output/powerpc64le-linux-gnu/bitcoin-13027d952519-powerpc64le-linux-gnu.tar.gz
21f3a1dcaee26d1b035f195367ed1315d83f753cb93a11ee82576411de78ae4c  guix-build-13027d952519/output/riscv64-linux-gnu/SHA256SUMS.part
7eb4d17dd6611db7b5167e237d36862893c8656ce6611804164673def7392789  guix-build-13027d952519/output/riscv64-linux-gnu/bitcoin-13027d952519-riscv64-linux-gnu-debug.tar.gz
c67d7721514da35e94d58a58a1530fbe4265abfee8c5ca4d9a8dcf17415c6756  guix-build-13027d952519/output/riscv64-linux-gnu/bitcoin-13027d952519-riscv64-linux-gnu.tar.gz
7a94d5a3e494f05077a49013a9341687a408f025ba4ca0c716ff4e881b8aef61  guix-build-13027d952519/output/x86_64-apple-darwin19/SHA256SUMS.part
25813a82754344e25d35b85c1ed01b94976dc1ab147a9066225d1015c11d4626  guix-build-13027d952519/output/x86_64-apple-darwin19/bitcoin-13027d952519-osx-unsigned.dmg
809ea888fcb883b8fc0e97b631d8b758d63cc040854dd3a5cf66b7b37e1287be  guix-build-13027d952519/output/x86_64-apple-darwin19/bitcoin-13027d952519-osx-unsigned.tar.gz
f0ae3c109deaa0d0798e5601964b05b1558e9789f65a672a83a47e7ff54650be  guix-build-13027d952519/output/x86_64-apple-darwin19/bitcoin-13027d952519-osx64.tar.gz
3aa4c0f0eb3944167a3b278f4c2826dfb8f18d1a7b1b33a8fa4a316eb99dfbe7  guix-build-13027d952519/output/x86_64-linux-gnu/SHA256SUMS.part
7a2bbcc399caf2fa17890cc29f46ad9ce940ac1a8f93706fb61ec70238b7ae6b  guix-build-13027d952519/output/x86_64-linux-gnu/bitcoin-13027d952519-x86_64-linux-gnu-debug.tar.gz
7cae0e347222e62dc4e7dba494b26164d5133f471d9e3e060046e805bd1fd142  guix-build-13027d952519/output/x86_64-linux-gnu/bitcoin-13027d952519-x86_64-linux-gnu.tar.gz
843dc5ff2a2618b08359debe12192be9f80ebd04ad7df8ec1d7bdb14837d8fdb  guix-build-13027d952519/output/x86_64-w64-mingw32/SHA256SUMS.part
56588107c596008f7b330b125f344a09fd70cbb0b8afb5ea5cdf9c5baa795abc  guix-build-13027d952519/output/x86_64-w64-mingw32/bitcoin-13027d952519-win-unsigned.tar.gz
6df35bd9df17912b8888087d6afb3382e3f3c197b131e3667442e73629727a2d  guix-build-13027d952519/output/x86_64-w64-mingw32/bitcoin-13027d952519-win64-debug.zip
e1b40b9b507e61018b3350a4e4dd142c13edf6807ba6cf273b4986cebdccec72  guix-build-13027d952519/output/x86_64-w64-mingw32/bitcoin-13027d952519-win64-setup-unsigned.exe
4d351114f736f21ffcd7e93f9eb1ebdbd7618db22f684e16e46e00c8287719ff  guix-build-13027d952519/output/x86_64-w64-mingw32/bitcoin-13027d952519-win64.zip
```

robot-dreams:
ACK 84b984ba0b69faf6d04483ee8d24d130dc762af6

sipa:
When compiling commit "refactor: Make PREC_BITS a parameter of ecmult_gen_build_prec_table" (ecff25fbd1998238d5fa67f74750f1857e1262e9):

```
src/gen_ecmult_gen_static_prec_table.c: In function ‘main’:
src/gen_ecmult_gen_static_prec_table.c:33:16: error: ‘ECMULT_GEN_PREC_B’ undeclared (first use in this function); did you mean ‘ECMULT_GEN_PREC_G’?
   33 |     int bits = ECMULT_GEN_PREC_B;
      |                ^~~~~~~~~~~~~~~~~
      |                ECMULT_GEN_PREC_G
src/gen_ecmult_gen_static_prec_table.c:33:16: note: each undeclared identifier is reported only once for each function it appears in
make: *** [Makefile:1704: gen_ecmult_gen_static_prec_table.o] Error 1
```

real-or-random:
> When compiling commit "refactor: Make PREC_BITS a parameter of ecmult_gen_build_prec_table" ([ecff25f](https://github.com/bitcoin-core/secp256k1/commit/ecff25fbd1998238d5fa67f74750f1857e1262e9)):

Forced-push to fix this intermediate commit. The new final tree at commit 56284c7d44c0ed46e636588bfbf6c403b7dfa6c1 is the same tree as at the previous commit 84b984ba0b69faf6d04483ee8d24d130dc762af6.

robot-dreams:
ACK 56284c7d44c0ed46e636588bfbf6c403b7dfa6c1

Confirmed that the final result is unchanged, and that every intermediate commit builds. (Some of the earlier commits fail `exhaustive_tests` but I don't have a strong opinion about this.)

real-or-random:
Pushed three further minimal fixups commits to https://github.com/real-or-random/secp256k1/tree/202108-prebuilt-gen-context-fixups1 and rebased here...

@robot-dreams Sorry, can you reACK? :)

src/gen_ecmult_gen_static_prec_table.c
```diff
@@ -0,0 +1,83 @@
...
+        fprintf(stderr, "Could not open %s for writing!\n", outfile);
+        return -1;
+    }
+
+    fprintf(fp, "/* This file was automatically generated by gen_ecmult_gen_static_prec_table. */\n");
+    fprintf(fp, "/* See ecmult_gen_impl.h for details about the contents of this file. */\n");
+    fprintf(fp, "#ifndef SECP256K1_ECMULT_GEN_STATIC_PREC_TABLE_H\n");
+    fprintf(fp, "#define SECP256K1_ECMULT_GEN_STATIC_PREC_TABLE_H\n");
+
+    fprintf(fp, "#include \"group.h\"\n");
```
robot-dreams, line 36: Nice catch!

robot-dreams (commented):
ACK 7dfceceea692c4118829c06128c41623b2373ac2 (based on range-diff between 56284c7d44c0ed46e636588bfbf6c403b7dfa6c1 and 7dfceceea692c4118829c06128c41623b2373ac2)

I agree with the removal of the conditional `#include "libsecp256k1-config.h"`, since `gen_ecmult_gen_static_prec_table.c` doesn't depend on the value of `ECMULT_GEN_PREC_BITS` to generate the output .h file.

I assume the generated file doesn't need a similar conditional include because you're assuming any file that imports `ecmult_gen_static_prec_table.h` must have already imported another header that has `ECMULT_GEN_PREC_BITS` defined?

real-or-random:
> I assume the generated file doesn't need a similar conditional include because you're assuming any file that imports `ecmult_gen_static_prec_table.h` must have already imported another header that has `ECMULT_GEN_PREC_BITS` defined?

Yeah, any file including `ecmult_gen_static_prec_table.h` this should have included `ecmult_gen.h`. To be honest, that's only half of the story. I opened #1039 for the other half...

sipa:
ACK 7dfceceea692c4118829c06128c41623b2373ac2

PR #990, opened 2021-10-17
Add comment on length checks when parsing ECDSA sigs

real-or-random:
I claim the check can be removed but I don't want to touch this
stable and well-tested code.

On the way, we fix grammar in another comment.


RandyMcMillan:
ACK e02f313

real-or-random:
Merging this because it's just a change in a comment. If we ever change the actual code, we should get more eyes on it.

PR #991, opened 2021-10-17
Merge all "external" benchmarks into a single bench binary

sipa:
This combines `bench_verify`, `bench_sign`, `bench_ecdh`, `bench_recovery`, and `bench_schnorrsig` into a single `bench` binary.

I don't think there is a good reason to have this many binaries, and it complicates build config and CI.


sipa:
@real-or-random How about adding a way to select which benchmark(s) to run using keywords, like `bench_intern` has?

sipa:
Rebased, and added the ability to select on the command line which benchmarks to run.

real-or-random (approved):
ACK af6abcb3d0097a7f7892fb8b54a4c6363e5c2c7f diff looks good, command line options work, valgrind is happy

siv2r (commented):
tACK af6abcb, the command-line options work as expected on my ubuntu machine. The diff looks good.

you can find my command line output [here](https://gist.github.com/siv2r/9c4e1a929f84ef9ca3b070a578f4bd2a).

These are the few things that I observed: (not sure if we need to change them)
1. Both `./bench` and `./bench ecdsa` runs all benchmark. (due to `have_flag("edcsa")` being present on both verify and sign benchmarks.
2. If a module is enabled (let's say schnorrsig), the benchmark of that module cannot be run separately as we do for sign and verify.

real-or-random:
Maybe I'm misunderstating your observations but I can't confirm. 

>     1. Both `./bench` and `./bench ecdsa` runs all benchmark. (_since the `have_flag("edcsa")` is present on both verify and sign benchmarks._)

Indeed but you can also specify `./bench ecdsa_verify` or `./bench ecdsa_sign.`

> 
>     2. If a module is enabled (let's say schnorrsig), the benchmark of that module cannot be run separately like we do for sign and verify.

You can also specify `./bench schnorrsig_verify` or `./bench schnorrsig_sign.`

siv2r:
> You can also specify `./bench schnorrsig_verify` or `./bench schnorrsig_sign.`

Ohh.. I didn't review the `src/modules/schnorrsig/bench_impl.h` properly. This works. Thanks!

> Indeed but you can also specify `./bench ecdsa_verify` or `./bench ecdsa_sign.`

Yes, I thought both the `./bench` and `./bench ecdsa` were equivalent. Now, I see that I was wrong. If we have a module enabled they act differently.

Everything looks good!

jonasnick (commented):
Concept ACK, played around with this a bit and it works as one would expect

A follow-up PR could add a `--help` which would also mention the `get_iters` environment variable (I have to look up in the source how it's called exactly every time).

PR #993, opened 2021-10-19
Enable non-experimental modules by default

real-or-random:
This has been discussed in https://github.com/bitcoin-core/secp256k1/issues/817#issuecomment-693198323 and I agree with the arguments brought up there. 

Alternatively, we could not enable them and add a discussion to the readme why we discourage people from using the modules. I believe enabling ECDH is not very controversial. But what about recovery? Do we want to leave it off and instead give a reason?

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -155,20 +155,20 @@ AC_ARG_ENABLE(examples,
     [SECP_SET_DEFAULT([enable_examples], [no], [yes])])
 
 AC_ARG_ENABLE(module_ecdh,
-    AS_HELP_STRING([--enable-module-ecdh],[enable ECDH module [default=no]]), [],
-    [SECP_SET_DEFAULT([enable_module_ecdh], [no], [yes])])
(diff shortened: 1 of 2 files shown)


real-or-random:
I propose doing this before the release. A release would hopefully trigger updates in a few packages and distributions (which currently use dates or commit id as version numbers), and if we believe that standard distributions should enable non-experimental modules (as discussed in the linked issue), then now is a good time.

Note: This PR needs an update, now that more (i.e., all) modules have been made non-experimental.

jonasnick:
> I propose doing this before the release.

Concept ACK. Do you also want to add an entry to the ChangeLog, similar to #1126?

real-or-random:
Updated. This now doesn't enable the ECDSA recovery module if this is controversial. Rationale is in the commit message.

jonasnick (commented):
ACK 41e8704b484652cf5bbb2b7ecc27feedc3cf0ae1

PR #994, opened 2021-10-20
x-only ECDH support

rustyrussell:
Technically this isn't required: you can always implement an x-only hashfn yourself, and convert to a compressed pubkey to call secp256k1_ecdh().  But I'm not smart enough to figure that out myself (thanks @jonasnick !) so here we are.

API question: should the xonly version take a scalar or a keypair?

Impl question: should this implementation live in ecdh, extrakeys, or even a new ecdh_extrakeys module?

Thanks!

### DIFF
--- a/src/modules/extrakeys/tests_impl.h
+++ b/src/modules/extrakeys/tests_impl.h
@@ -571,6 +571,35 @@ void test_keypair_add(void) {
     secp256k1_context_destroy(verify);
 }
 
+void test_xonly_ecdh(void) {
+#ifdef ENABLE_MODULE_ECDH
+    unsigned char sk1[32], sk2[32];
+    secp256k1_keypair keypair1, keypair2;
+    secp256k1_xonly_pubkey xpubkey1, xpubkey2;
+    secp256k1_pubkey pubkey1;
+    unsigned char output1[32], output2[32], output3[32];
+
+    /* Create random points */
+    secp256k1_testrand256(sk1);
+    secp256k1_testrand256(sk2);
+    CHECK(secp256k1_keypair_create(ctx, &keypair1, sk1) == 1);
+    CHECK(secp256k1_keypair_create(ctx, &keypair2, sk2) == 1);
+    CHECK(secp256k1_keypair_xonly_pub(ctx, &xpubkey1, NULL, &keypair1) == 1);
+    CHECK(secp256k1_keypair_xonly_pub(ctx, &xpubkey2, NULL, &keypair2) == 1);
+
+    /* They should get the same shared secret. */
+    CHECK(secp256k1_ecdh_xonly(ctx, output1, &xpubkey1, sk2, NULL, NULL) == 1);
+    CHECK(secp256k1_ecdh_xonly(ctx, output2, &xpubkey2, sk1, NULL, NULL) == 1);
+    CHECK(memcmp(output1, output2, sizeof(output1)) == 0);
+
+    /* We can also do ECDH via normal compressed pubkey functions, if we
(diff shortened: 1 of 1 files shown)


rustyrussell:
>  The way I see x-only pubkeys is that this is mostly an optimization when putting keys on chain.

I see things differently from you: compressed pubkeys are a vestigial tail, especially the 0x02/0x03 SEC1 encoding which is arbitrary.

As a simple crypto API user, 32-byte pubkeys are a minor but clear API improvement.  They align nicely, they index ~uniformly, and I expect with encouragement and tooling they will become the standard over time.

So, I've started using x-only pubkeys everywhere in new greenfields projects.  They're used in https://github.com/rustyrussell/centurymetadata (where I hit this ECDH issue) and in BOLT12 offers so far.  I'm considering them for my gossip update proposal, too.,

I think future generations will thank us for encouraging and supporting their ubiquitous adoption.

rustyrussell:
> Hm, I don't think it's strange. So far, we use 33 bytes pubkeys but hash 64 bytes. Is that strange? I guess not.

BTW, I think this is wrong: the default hashes 33 bytes, a-la a compressed key:

```
    unsigned char version = (y32[31] & 0x01) | 0x02;
    secp256k1_sha256 sha;
    (void)data;

    secp256k1_sha256_initialize(&sha);
    secp256k1_sha256_write(&sha, &version, 1);
    secp256k1_sha256_write(&sha, x32, 32);
    secp256k1_sha256_finalize(&sha, output);
```

real-or-random:
> As a simple crypto API user, 32-byte pubkeys are a minor but clear API improvement. They align nicely, they index ~uniformly, and I expect with encouragement and tooling they will become the standard over time.
> 

But then pragmatically speaking, what you really care about is that there's an ECDH function which takes x-only pubkeys as input and it doesn't really matter what is hashed. Or does it matter?

rustyrussell:
> > As a simple crypto API user, 32-byte pubkeys are a minor but clear API improvement. They align nicely, they index ~uniformly, and I expect with encouragement and tooling they will become the standard over time.
> 
> But then pragmatically speaking, what you really care about is that there's an ECDH function which takes x-only pubkeys as input and it doesn't really matter what is hashed. Or does it matter?

Yes, this is how I got into this mess. The current scheme breaks: given S1->PK1, S2->PK2, ECDH(S1,PK2) != ECHD(S2,PK1) unless PK1 and PK2 are both even (or maybe both odd, not sure).  I had to invert the secret if its pubkey was 03...

Further, interop matters, when setting standards (which is what libsecp256k1 is doing here, no doubt). We could also prepend a BIP-340 style tag, but prepending 02 or 03 is just a weird thing to lock in for future generations to ogle at?

I know @gmaxwell disliked the malleation which x-only hashing enabled, but that problem doesn't really exist in an x-only world.  With that discounted, the fact that there's a further optimization possible in future is compelling IMHO.

w0xlt:
I created an example of using x-only ECDH from this PR.
Implements the "[Basic Scheme](https://gist.github.com/RubenSomsen/c43b79517e7cb701ebf77eec6dbb46b8#basic-scheme)" mentioned in the article "Silent Payment".
https://github.com/w0xlt/secp256k1/blob/a9677ad9f064efd6c1f91afb9fa2f5d2ab43cd03/examples/spbs.c

More information at https://github.com/bitcoin-core/secp256k1/issues/1097#issuecomment-1086946055

real-or-random:
@rustyrussell We're trying to gauge demand for this. Are you still interested in this feature? If yes, do you agree that this PR could be closed in favor of #1198?

PR #995, opened 2021-10-20
build: stop treating schnorrsig, extrakeys modules as experimental

### LINKED ISSUE #992: Stabilization of schnorrsig
Beyond those two, the only remaining experimental feature is the recently-merged BIP340 support. It is experimental for two reasons:

* There are a few discussions still ongoing that may impact the API (variable-length message support, batch validation) without changes to the (currently supported) part of the scheme. It's not impossible to just add new functions for these once they're worked out, but given the next point, I think this may not be needed.
* It's very hard to be *sure* what exact semantics of BIP340 will actually activate on Bitcoin (or if it will at all), and if that ends up being different from what is implemented now, eventually this library will want to implement the final scheme. I think changes are very unlikely at this point, but it's also not exactly under our control.

_Originally posted by @sipa in https://github.com/bitcoin-core/secp256k1/issues/817#issuecomment-693033273_

jonasnick:
Fixes #992

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,9 +17,7 @@ Features:
 * Suitable for embedded systems.
 * Optional module for public key recovery.
 * Optional module for ECDH key exchange.
-* Optional module for Schnorr signatures according to [BIP-340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki) (experimental).
-
-Experimental features have not received enough scrutiny to satisfy the standard of quality of this library but are made available for testing and review by the community. The APIs of these features should not be considered stable.
+* Optional module for Schnorr signatures according to [BIP-340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki).
 
 Implementation details
 ----------------------
@@ -72,7 +70,6 @@ libsecp256k1 is built using autotools:
 Usage examples
 -----------
   Usage examples can be found in the [examples](examples) directory. To compile them you need to configure with `--enable-examples`.
-  For experimental modules, you will also need `--enable-experimental` as well as a flag for each individual module, e.g. `--enable-module-ecdh`.
   * [ECDSA example](examples/ecdsa.c)
   * [Schnorr Signatures example](examples/schnorr.c)
   * [Deriving a shared secret(ECDH) example](examples/ecdh.c)
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ Implementation details
   * Optimized implementation of arithmetic modulo the curve's field size (2^256 - 0x1000003D1).
     * Using 5 52-bit limbs (including hand-optimized assembly for x86_64, by Diederik Huys).
     * Using 10 26-bit limbs (including hand-optimized assembly for 32-bit ARM, by Wladimir J. van der Laan).
+      * This is an experimental feature that has not received enough scrutiny to satisfy the standard of quality of this library but is made available for testing and review by the community.
 * Scalar operations
   * Optimized implementation without data-dependent branches of arithmetic modulo the curve's order.
     * Using 4 64-bit limbs (relying on __int128 support in the compiler).
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -67,12 +67,12 @@ task:
   << : *LINUX_CONTAINER
   matrix: &ENV_MATRIX
     - env: {WIDEMUL:  int64,  RECOVERY: yes}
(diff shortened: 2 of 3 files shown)


real-or-random:
Needs trivial rebase. Sorry that I was sitting on this for so long...

real-or-random:
> Needs trivial rebase. Sorry that I was sitting on this for so long...

Again :/

real-or-random (approved):
ACK cd7285800287563e2a12a56dadee779dbe2330d4

One thing that I noticed is that there's no "natural counterpart" verification function for the normal signing function. Normal signing assumes 32 bytes but the verification function expects a length argument... But this is not a blocker. It's at most related to "elegance" of the API, and we may not care at all.

real-or-random:
> One thing that I noticed is that there's no "natural counterpart" verification function for the normal signing function. Normal signing assumes 32 bytes but the verification function expects a length argument... But this is not a blocker. It's at most related to "elegance" of the API, and we may not care at all.

Ok, I've just discussed this with @jonasnick somewhere else and here are our thoughts:

The current API is based on the discussion in  https://github.com/sipa/bips/issues/207 and the agreed plan there is to amend the BIP to accept variable length messages, and give recommendations for domain separation. (I think I promised to open a PR to the BIP and I failed -- sorry, will move this the top of my list). 

So taking this into account, the BIP340 signing and verifications algorithms (will) actually accept variable length messages.  This means that for my above ACK comment that not the var-len `_verify` function is off but actually the `_sign` function is not perfectly aligned with the standard. A possible enhancement of our API is to rename `schnorrsign_sign` to `schnorrsig_sign32` to emphasize it’s a specialization of BIP340 signing to  32-byte messages, and deprecate `schnorrsign_sign`. (I think we should not remove it given it's used widely and perfectly safe. But there's no need to have more than two (non-deprecated) signing functions. Callers who want variable length messages can call sign_custom with `NULL`. This is reasonable enough.

And @jonasnick points out that this way of thinking will also simplify advanced signing APIs, e.g., musig: Then the way to go for these is to provide *only* varlen functionality. This will save code and simplify these APIs by reducing the number of provided functions. For example, a call to `musig_nonce_gen` or similar functions is complex anyway, passing a “32” does not hurt. 

Let me remind people that supporting varlen is almost entirely orthogonal to domain separation. People can get domain separation wrong even with an API that supports only 32-byte messages.

@sipa Does this sound reasonable? If yes, do you think we should rename `schnorrsign_sign` to `schnorrsig_sign32` and deprecate the former? If yes, we should do it now (quickly) before stabilization. I just don't think we should hold this up for another month...

sipa:
I'm happy with either renaming or keeping things as they currently are.

real-or-random:
See the README issue mentioned in https://github.com/bitcoin-core/secp256k1/pull/1091#issuecomment-1069403585. Perfectly fine to solve this in a separate PR if you prefer, we just shouldn't forget about it.

jonasnick:
Fixed the README issue.

real-or-random (approved):
ACK 7f09d0f311117289719b690f91f6a907c2c6f3e2

Let's first get #1089 in.

README.md
```diff
@@ -72,7 +71,6 @@ libsecp256k1 is built using autotools:
 Usage examples
 -----------
   Usage examples can be found in the [examples](examples) directory. To compile them you need to configure with `--enable-examples`.
```
robot-dreams, line 73: Following these updated instructions only produces `example_ecdsa`. Should this also say something like:

The `schnorr` and `ecdh` modules aren't enabled by default, so to compile the corresponding examples you will also need `--enable-module-schnorr` and `--enable-module-ecdh`.

robot-dreams (commented):
Looks good, just one README comment

.cirrus.yml change looks OK: all CI checks passed, expected configurations were there (e.g. ASM:arm EXPERIMENTAL:yes), number of checks didn't change (this PR and #1089 both had 66 checks each)

configure.ac change looks OK: searched for `experimental` and didn't find anything else that needed changing

fanquake (approved):
ACK 7f09d0f311117289719b690f91f6a907c2c6f3e2 - When this is in, I think we'll do a subtree update in Core, and prune some build cruft on our side.

README.md
real-or-random: Indeed. I think this can be done in a separate PR because seems to be a more general issue. At the moment, we don't explain at all how to enable the modules. Maybe this should go to "build steps".

PR #998, opened 2021-10-21
.gitignore: add build-aux/ltmain.sh-e

Sjors:
When running [autogen](https://github.com/ElementsProject/libwally-core/blob/master/tools/autogen.sh) in [LibWally](https://github.com/ElementsProject/libwally-core), which uses secp256k1-zkp as a submodule, this file is left on my system. It also happens when I swap out secp256k1-zkp for secp256k1.

It does not happen when I compile secp256k1 directly, so perhaps this is an issue on the LibWally side.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,6 +51,7 @@ build-aux/config.sub
 build-aux/depcomp
 build-aux/install-sh
 build-aux/ltmain.sh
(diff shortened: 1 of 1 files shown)


Sjors:
cc @griffiths

PR #1000, opened 2021-10-26
Synthetic int128 type.

roconnor-blockstream:
Abstracts the int128 type and provides an native version, if available, or a implements it using a pair of int64_t's.

This is activated by setting the configuration flag `--with-test-override-wide-multiply=int128_struct`.

The primary purpose of this PR is to take advantage of MSVC's [umulh](https://docs.microsoft.com/en-us/cpp/intrinsics/umulh?view=msvc-170) intrinsic that we can use to simulate an int128 type which MSVC does not have (AFAIU). This PR lays out the groundwork for this level of MSVC support, but doesn't include the configuration logic to enable it yet.

For completeness, and implementation of `umulh` and `mulh` are also provided for compilers that support neither the intrinsic nor the int128 type (such as CompCert?).  This also opens up the possibility of removing the 32-bit field and scalar implementations should that ever be desired.


src/int128.h
```diff
@@ -0,0 +1,182 @@
...
+#include "util.h"
+
+/*
+#ifdef UINT128_MAX
+#undef UINT128_MAX
+*/
+
+#if defined(UINT128_MAX)
+typedef uint128_t secp256k1_uint128;
+typedef int128_t secp256k1_int128;
```
real-or-random, line 14: I think it's a idea to define our our own type with a separate prefix. I guess then the entire logic currently in `util.h` will conceptually belong here.

real-or-random (commented):
I think it would help readability to arrange the different implementations of the type in different files, similar to what we have a field and scalar "modules", where we have implementation files such as `scalar4x64_impl.h` but still a single `scalar.h` that makes sure that all function prototypes are identical.  Then you could remove a lot of the ifdefs within the functions.

src/int128.h
roconnor-blockstream: I am inclined to move the int128 logic from util.h into this file.  I just haven't done it yet.

roconnor-blockstream:
I'm moving this out of draft stage.  The coding is complete.  There are  a few tasks that remain.

- [ ] Double check that there is no signed integer overflow due to rearrangement of the order of some operations.
- [ ] We need someone to build this on MSVC.
- [ ] Make sure that the refactored code for the native int128 type isn't any slower.

To compile with the synthetic int128 type, you currently need to pass a configuration flag `--with-test-override-wide-multiply=int128_struct`.

real-or-random:
>     * [ ]  We need someone to build this on MSVC.

I have built this with MSVC 2019 on wine (https://github.com/mstorsjo/msvc-wine) and it works. Tests and exhaustive tests pass. 

For reference:
```
 ./configure --disable-benchmark --with-test-override-wide-multiply=int128_struct CC=/opt/msvc/bin/x64/cl.exe CFLAGS="-Za -O2 -w" LD=/opt/msvc/bin/x64/link.exe
```


For the Arch Linux users, there's an AUR package: https://aur.archlinux.org/packages/msvc-wine-git/

sipa:
FWIW, bitcoin core has an AppVeyor MSVC/Windows CI environment. Maybe it's worth looking into adding one for secp256k1 directly.

real-or-random:
Indeed. Let me note that also Cirrus apparently offers Windows though I have no idea if it's good and/or reliable:  https://cirrus-ci.org/guide/windows/

src/int128_struct_impl.h
```diff
@@ -0,0 +1,129 @@
+#ifndef SECP256K1_INT128_64_IMPL_H
+#define SECP256K1_INT128_64_IMPL_H
+
+#include "int128.h"
+
+#if defined(_M_X64) | defined(_M_ARM64) | defined(_WIN64) /* MSVC */
+  #include <intrin.h>
+  #define secp256k1_umulh __umulh
+  #define secp256k1_mulh __mulh
+#else
```
real-or-random, line 10:
```suggestion
#if defined(_M_X64) | defined(_M_ARM64) | defined(_WIN64) /* MSVC */
  #error
  #include <intrin.h>
  #define secp256k1_umulh __umulh
  #define secp256k1_mulh __mulh
#else
```

Follow-up: I confirmed that change results in an error in the the MSVC build, so the intrinsics should indeed be used.

src/int128.h
```diff
@@ -0,0 +1,60 @@
...
+#include "int128_native.h"
+#elif defined(SECP256K1_INT128_STRUCT)
+#include "int128_struct.h"
+#else
+#error "Please select int128 implementation"
+#endif
+
+static SECP256K1_INLINE uint64_t secp256k1_umulh(uint64_t a, uint64_t b);
+
+static SECP256K1_INLINE int64_t secp256k1_mulh(int64_t a, int64_t b);
```
sipa, line 20: This function does not appear to be used outside of `src/int128_struct_impl.h`, perhaps it can be considered a private implementation detail there, rather than exposing it?

src/int128.h
```diff
@@ -0,0 +1,60 @@
...
+
+static SECP256K1_INLINE void secp256k1_u128_from_u64(secp256k1_uint128 *r, uint64_t a);
+
+static SECP256K1_INLINE int secp256k1_u128_check_bits(secp256k1_uint128 *r, unsigned int n);
+
+static SECP256K1_INLINE void secp256k1_i128_mul(secp256k1_int128 *r, int64_t a, int64_t b);
+
+static SECP256K1_INLINE void secp256k1_i128_accum_mul(secp256k1_int128 *r, int64_t a, int64_t b);
+
+static SECP256K1_INLINE void secp256k1_i128_dissip_mul(secp256k1_int128 *r, int64_t a, int64_t b);
```
sipa, line 43: This function does not appear to be used outside of `src/int128_struct_impl.h`, perhaps it can be considered a private implementation detail there, rather than exposing it?

src/scalar_4x64_impl.h
```diff
@@ -172,27 +190,28 @@ static int secp256k1_scalar_cond_negate(secp256k1_scalar *r, int flag) {
...
 
 /** Add a*b to the number defined by (c0,c1,c2). c2 must never overflow. */
 #define muladd(a,b) { \
-    uint64_t tl, th; \
-    { \
-        uint128_t t = (uint128_t)a * b; \
-        th = t >> 64;         /* at most 0xFFFFFFFFFFFFFFFE */ \
-        tl = t; \
-    } \
+    uint64_t th = secp256k1_umulh(a, b);  /* at most 0xFFFFFFFFFFFFFFFE */ \
```
sipa, line 213: Really the `th` and `tl` computed here (and below) are the top and bottom of `secp256k1_uint128`. Perhaps it should be written in function of the actual `secp256k1_u128_...` functions, instead of the lower-level `secp256k1_umulh`? This would have as advantages:
* Not relying on the computer to recognize that `a * b` is computed twice and can be deduplicated.
* Permit removing `secp256k1_umulh` from the header.

src/int128.h
real-or-random: This hasn't been resolved yet.

roconnor-blockstream: Are you not happy with the changes to `util.h`?

real-or-random: I'd prefer to move the remaining `#ifdef` `WIDEMUL` logic in `util.h` to `int128_impl.h` (or `int128.h`?) because it belongs there conceptually.

I think `util.h` is just the kitchen sink for code pieces that don't have a proper home (or where we were too lazy to create a proper home. Now that you introduce `int128_impl.h` here, I think that's the place where those `#ifdefs` belong.

roconnor-blockstream: As it stands when SECP256K1_WIDEMUL_INT64 ends up being defined then `int128_impl.h` isn't even imported anywhere.  So I'm not sure what you are imagining.

src/scalar_4x64_impl.h
roconnor-blockstream: Done.

src/int128.h
roconnor-blockstream: Done.

src/int128.h
roconnor-blockstream: Done.

src/int128.h
real-or-random: Yes, I was wrong. See below though.

src/modinv64_impl.h
```diff
@@ -324,54 +327,64 @@ static void secp256k1_modinv64_update_de_62(secp256k1_modinv64_signed62 *d, secp
...
     /* Verify that the low 62 bits of the computation are indeed zero, and then throw them away. */
-    VERIFY_CHECK(((int64_t)cd & M62) == 0); cd >>= 62;
-    VERIFY_CHECK(((int64_t)ce & M62) == 0); ce >>= 62;
+    VERIFY_CHECK((secp256k1_i128_to_i64(&cd) & M62) == 0); secp256k1_i128_rshift(&cd, 62);
+    VERIFY_CHECK((secp256k1_i128_to_i64(&ce) & M62) == 0); secp256k1_i128_rshift(&ce, 62);
     /* Compute limb 1 of t*[d,e]+modulus*[md,me], and store it as output limb 0 (= down shift). */
-    cd += (int128_t)u * d1 + (int128_t)v * e1;
-    ce += (int128_t)q * d1 + (int128_t)r * e1;
+    secp256k1_i128_accum_mul(&cd, u, d1);
+    secp256k1_i128_accum_mul(&cd, v, e1);
```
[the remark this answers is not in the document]
real-or-random: Could we just introduce a temporary variable and sum up there first? Or better, add a `secp256k1_i128_accum_mul_twice` that will accumulate only in the int128 impl? This will save us the reviewer time here. I expect the compiler to prune it anyway.

(Or has anyone redone the bound analysis? I guess it's not hard but I'd need to refresh my memory here.)

roconnor-blockstream:
I notice that your PR still keeps the `USE_*_WIDEMUL_*` in `util.h`.  it still seems we cannot really move it out of `util.h` because it both selects between `INT128_STRUCT` and `INT128_NATIVE`, but can also select `WIDEMUL_INT64` which isn't really `INT128` related.

More specifically this block of CPP code is interpreting the `wideMultiply` configuration option, which I've designed to have 3 options (by adding that third option).

real-or-random:
> I notice that your PR still keeps the `USE_*_WIDEMUL_*` in `util.h`. it still seems we cannot really move it out of `util.h` because it both selects between `INT128_STRUCT` and `INT128_NATIVE`, but can also select `WIDEMUL_INT64` which isn't really `INT128` related.

Right. I think that's fine. 

So this means we have a couple of input #defines (preset in the compiler or set by autoconf, e..g, `USE_*_WIDEMUL_*`) and we have some CPP logic in file that turns those inputs into "nice" #ifdefs that can then be used in the rest of the code base, e.g., `INT128_NATIVE`. I believe that's a reasonable way of doing things.

(In the future we should move the CPP logic from `util.h` to a separate file. It somehow ended up in `util.h` though it does not belong there...)

roconnor-blockstream:
I've reviewed the changes in the association of `+` for signed int128 calculations to look for changes in overflow behavior.

Signed int128 is only used in the `src/modinv64_impl.h` file.  The only changes in association occur in the functions:

* `secp256k1_modinv64_update_de_62`
* `secp256k1_modinv64_update_fg_62`
* `secp256k1_modinv64_update_fg_62_var`

The values being summed are all of the form `(int128_t)u * ??? + (int128_t)v * ???` (or similarly for `q` and `r`). The only way the new association could cause an overflow (or underflow) is if this previous calculation depended on some cancellation within the above addition term to prevent the final accumulation from itself overflowing.  However, the previous calculation does not depend on any cancellation.  The only assumed constraint on `u` and `v` (resp. `q` and `r`) is that `|u|+|v| <= 2^62`, which implies nothing about the values of `u` and `v` causing cancellation.

My conclusion, after the above and re-reviewing the above functions, is that neither the new (or old) order of addition risks overflow.

real-or-random:
Indeed, we know nothing about the signs of the summands. To convince myself, I also redid the bounds analysis:

Counting value bits (i.e., bits that are not the sign bit):

For each step, `cd` starts with at most 52 bits.
Then we accumulate `(int128_t)u * ???`, where the factors have at most 62 bits (implied by `|u|+|v| <= 2^62`*) and 63 bits (trivial bound for `int64_t`).
Same with  `(int128_t)v * e0`.
Then we accumulate `(int128_t) modinfo->modulus.??? * md`, where the factors have at most 62 (modulus is in signed62) and 63 value bits.

We work with signed, so the value with the maximum absolute value representable in B bits is `-(1<<B)` (with absolute `1<<B`).  
This means the maximum absolute value `cd` can take is `(1<<52) + (1<<62) * (1<<63) + (1<<62) * (1<<63) + (1<<62) * (1<<63)` which is a 127 bit value and thus reprensentable in 127 value bits.

Same is true for `ce`.
 
*except `u==2^62`, which needs 63 bits but then `v==0`, which also works: `(1<<52) + (1<<63) * (1<<63) + 0 * (1<<63) + (1<<62) * (1<<63)` is also a 127 bit value.

src/int128_struct_impl.h
```diff
@@ -0,0 +1,133 @@
+#ifndef SECP256K1_INT128_STRUCT_IMPL_H
+#define SECP256K1_INT128_STRUCT_IMPL_H
+
+#include "int128.h"
+
+#if defined(_M_X64) | defined(_M_ARM64) | defined(_WIN64) /* MSVC */
```
real-or-random, line 6: Hm, I'm not sure about `_WIN64`. This is defined by icc for example, and as I understand also by mingw. (The Internet says that mingw, which is basically "gcc for windows", will support the MSVC intrinsicts, but it will also have a native 128 bit, so we'll probably want to use the latter.)

MSVC is what provides those intrinsics and `_MSC_VER` should really only be defined on MSVC. What about one of these two?

```
#if defined(_MSC_VER) && defined(_WIN64)
```
or
```
#if defined(_MSC_VER) && (defined(_M_X64) || defined(_M_ARM64))
```

Currently, the two versions should be equivalent. This may change if MSVC starts to support another 64 bit target in the future, and then I believe it is likely that `__mulh` und `__umulh` will exist, so I lean for the first option. But anyway, when that change happens to MSVC, we can deal with it...

Parts of my comment are guesswork so far, I believe we'll need to test compilation.

real-or-random:
@roconnor-blockstream Can you rebase this? This will ease benchmarking against master.

real-or-random:
Native 128bit performance looks good:

$ SECP256K1_BENCH_ITERS=1000000 ./bench_internal inverse

```
gcc 11.2, pr:

scalar_inverse                ,     2.40      ,     2.41      ,     2.42   
scalar_inverse_var            ,     1.74      ,     1.75      ,     1.81   
field_inverse                 ,     2.38      ,     2.39      ,     2.40   
field_inverse_var             ,     1.73      ,     1.73      ,     1.74

gcc 11.2, master:

scalar_inverse                ,     2.39      ,     2.39      ,     2.41   
scalar_inverse_var            ,     1.77      ,     1.77      ,     1.78   
field_inverse                 ,     2.37      ,     2.38      ,     2.39   
field_inverse_var             ,     1.76      ,     1.77      ,     1.82


clang 13.0.1, pr:

scalar_inverse                ,     2.70      ,     2.70      ,     2.72   
scalar_inverse_var            ,     1.69      ,     1.70      ,     1.70   
field_inverse                 ,     2.69      ,     2.70      ,     2.70   
field_inverse_var             ,     1.69      ,     1.69      ,     1.70   

clang 13.0.1, master:

scalar_inverse                ,     2.69      ,     2.70      ,     2.71   
scalar_inverse_var            ,     1.69      ,     1.69      ,     1.70   
field_inverse                 ,     2.69      ,     2.69      ,     2.69   
field_inverse_var             ,     1.68      ,     1.68      ,     1.68
```

But this is with asm on, I should have turned it off... Would be nice to see more benchmarks.

src/int128_struct_impl.h
```diff
@@ -0,0 +1,133 @@
...
+   return (int64_t)a->lo;
+}
+
+static SECP256K1_INLINE void secp256k1_i128_from_i64(secp256k1_int128 *r, int64_t a) {
+   r->hi = (uint64_t)(a >> 63);
+   r->lo = (uint64_t)a;
+}
+
+static SECP256K1_INLINE int secp256k1_i128_eq(const secp256k1_int128 *a, const secp256k1_int128 *b) {
+   return a->hi == b->hi && a->lo == b->lo;
```
real-or-random, line 125: I think we should name this `secp256k1_i128_eq_var`. Or try to make it const-time. A standard way would be:
```suggestion
static SECP256K1_INLINE int secp256k1_i128_eq(const secp256k1_int128 *a, const secp256k1_int128 *b) {
   return !((a->hi ^ b->hi) | (a->lo ^ b->lo));
```
Not sure if this works though.

src/int128_struct_impl.h
```diff
@@ -0,0 +1,133 @@
...
+}
+
+static SECP256K1_INLINE int secp256k1_i128_check_bit(secp256k1_int128 *r, unsigned int n) {
+   return n >= 64 ? r->hi == (uint64_t)1 << (n - 64) && r->lo == 0
```
real-or-random, line 129:
```suggestion
   VERIFY_CHECK(n < 128);
   return n >= 64 ? r->hi == (uint64_t)1 << (n - 64) && r->lo == 0
```
same for the unsigned version

src/int128_struct_impl.h
```diff
@@ -0,0 +1,133 @@
...
+
+static SECP256K1_INLINE void secp256k1_i128_det(secp256k1_int128 *r, int64_t a, int64_t b, int64_t c, int64_t d) {
+   secp256k1_i128_mul(r, a, d);
+   secp256k1_i128_dissip_mul(r, b, c);
+}
+
+/* Signed (arithmetic) right shift.
+ * Non-constant time in b.
+ */
+static SECP256K1_INLINE void secp256k1_i128_rshift(secp256k1_int128 *r, unsigned int b) {
```
real-or-random, line 105:
```suggestion
static SECP256K1_INLINE void secp256k1_i128_rshift(secp256k1_int128 *r, unsigned int b) {
    VERIFY_CHECK(b < 128);
```
same for the unsigned version

real-or-random (commented):
I haven't reviewed the struct implementations of `secp256k1_?128_accum/mul/mulh/rshift` algorithms in detail. I think it will be good to have some randomized unit tests here. (Maybe it's ok to just run them when we have a native type too and compare the result?). Most of these functions should be exercised by the current tests already but some may be not. For example, the `shift` functions are only called with specific shifts, so I think some branches will never be taken.

src/util.h
```diff
@@ -250,28 +250,23 @@ static SECP256K1_INLINE void secp256k1_int_cmov(int *r, const int *a, int flag)
     *r = (int)(r_masked | a_masked);
 }
 
-/* If USE_FORCE_WIDEMUL_{INT128,INT64} is set, use that wide multiplication implementation.
+/* If USE_FORCE_WIDEMUL_{INT128, INT128_STRUCT, INT64} is set, use that wide multiplication implementation.
  * Otherwise use the presence of __SIZEOF_INT128__ to decide.
```
real-or-random, line 254: I think we should detect 64-bit MSVC here and select `SECP256K1_INT128_STRUCT`.

roconnor-blockstream:
Rebased.

src/util.h
real-or-random: Ok sure, that will make things easier. :+1:

src/int128_struct_impl.h
roconnor-blockstream: I'm going to change it to `eq_var`, since it is far from obvious to me that the native version is constant time.

roconnor-blockstream: done.

src/int128_struct_impl.h
roconnor-blockstream: done.

src/int128_struct_impl.h
roconnor-blockstream: done.

src/int128.h
```diff
@@ -0,0 +1,55 @@
...
+ */
+static SECP256K1_INLINE void secp256k1_i128_rshift(secp256k1_int128 *r, unsigned int b);
+
+static SECP256K1_INLINE int64_t secp256k1_i128_to_i64(const secp256k1_int128 *a);
+
+static SECP256K1_INLINE void secp256k1_i128_from_i64(secp256k1_int128 *r, int64_t a);
+
+static SECP256K1_INLINE int secp256k1_i128_eq(const secp256k1_int128 *a, const secp256k1_int128 *b);
+
+static SECP256K1_INLINE int secp256k1_i128_check_bit(secp256k1_int128 *r, unsigned int n);
```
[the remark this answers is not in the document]
roconnor-blockstream: done.

src/util.h
real-or-random: > I think we should detect 64-bit MSVC here and select `SECP256K1_INT128_STRUCT`.

Note for the future PR: What we should actually do is to use to word size to decide which implementation to use, independently of the exact compiler.

src/int128_struct_impl.h
```diff
@@ -0,0 +1,137 @@
...
+/* Unsigned (logical) right shift.
+ * Non-constant time in n.
+ */
+static SECP256K1_INLINE void secp256k1_u128_rshift(secp256k1_uint128 *r, unsigned int n) {
+   VERIFY_CHECK(n < 128);
+   if (n >= 64) {
+     r->lo = (r->hi) >> (n-64);
+     r->hi = 0;
+   } else if (n > 0) {
+     r->lo = ((r->hi & (((uint64_t)1<<n)-1)) << (64-n)) | r->lo >> n;
```
roconnor-blockstream, line 52: @real-or-random Unsigned overflow from shifting is well defined right?  So I don't need the mask here?

real-or-random: Right, unsigned overflow is never UB, even when it comes from shifting. See C99: https://port70.net/~nsz/c/c99/n1256.html#6.5.7p4 

(The corresponding C89 part is a little bit strange since it doesn't mention the UB explicitly and it works under the assumption that `int` and `long` are the only non-promoted integer types: https://port70.net/~nsz/c/c89/c89-draft.html#3.3.7 I mean ok, `stdint.h` does not exist in C89...)

roconnor-blockstream: Oh I remember why I always write it this way, because we _technically_ don't know the size of a signed integer.

src/int128_struct_impl.h
```diff
@@ -0,0 +1,137 @@
...
+   r->lo -= lo;
+}
+
+static SECP256K1_INLINE void secp256k1_i128_det(secp256k1_int128 *r, int64_t a, int64_t b, int64_t c, int64_t d) {
+   secp256k1_i128_mul(r, a, d);
+   secp256k1_i128_dissip_mul(r, b, c);
+}
+
+/* Signed (arithmetic) right shift.
+ * Non-constant time in b.
```
roconnor-blockstream, line 105: `in n`.

src/int128_native_impl.h
```diff
@@ -0,0 +1,76 @@
...
+   *r = a;
+}
+
+static SECP256K1_INLINE int secp256k1_i128_eq_var(const secp256k1_int128 *a, const secp256k1_int128 *b) {
+   return *a == *b;
+}
+
+static SECP256K1_INLINE int secp256k1_i128_check_bit(const secp256k1_int128 *r, unsigned int n) {
+   VERIFY_CHECK(n < 128);
+   return (*r == (int128_t)1 << n);
```
roconnor-blockstream, line 73: 1 << 127 would cause signed integer overflow.

This never happens, but the VERIFY_CHECK should be reduced to n < 127.

roconnor-blockstream: Done.

src/int128_struct_impl.h
roconnor-blockstream: Done.

src/int128_struct_impl.h
laanwj: yes, unsigned overflow is always defined, it's wrap-around no matter what

roconnor-blockstream: Right.  But *technically speaking* a `signed int` could, in some unrealistic universe, be 65 bits, thus subjecting `r->hi` to signed integer promotion and thus causing signed integer overflow.

That said, if people would like me to rewrite this to assume 32-bit `signed int`s, I'm happy to change it.

src/scalar_4x64_impl.h
```diff
@@ -542,14 +568,18 @@ static void secp256k1_scalar_reduce_512(secp256k1_scalar *r, const uint64_t *l)
...
+    r->d[0] = secp256k1_u128_to_u64(&c128); secp256k1_u128_rshift(&c128, 64);
+    secp256k1_u128_accum_u64(&c128, p1);
+    secp256k1_u128_accum_mul(&c128, SECP256K1_N_C_1, p4);
+    r->d[1] = secp256k1_u128_to_u64(&c128); secp256k1_u128_rshift(&c128, 64);
+    secp256k1_u128_accum_u64(&c128, p2);
+    secp256k1_u128_accum_u64(&c128, p4);
+    r->d[2] = secp256k1_u128_to_u64(&c128); secp256k1_u128_rshift(&c128, 64);
+    secp256k1_u128_accum_u64(&c128, p3);
+    r->d[3] = secp256k1_u128_to_u64(&c128); secp256k1_u128_rshift(&c128, 64);
+    c = secp256k1_u128_to_u64(&c128);
```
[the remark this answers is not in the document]
roconnor-blockstream: The last line of the function exists outside this `#if` block (notice that the type of `c` is originally different between the two blocks).  So I cannot avoid making a `c` value here.   I suppose I could duplicate the last line if you think that is worthwhile.

src/scalar_4x64_impl.h
```diff
@@ -69,49 +70,62 @@ SECP256K1_INLINE static int secp256k1_scalar_check_overflow(const secp256k1_scal
...
+    secp256k1_u128_accum_u64(&t, r->d[2]);
+    secp256k1_u128_accum_u64(&t, ((uint64_t)((bit >> 6) == 2)) << (bit & 0x3F));
+    r->d[2] = secp256k1_u128_to_u64(&t); secp256k1_u128_rshift(&t, 64);
+    secp256k1_u128_accum_u64(&t, r->d[3]);
+    secp256k1_u128_accum_u64(&t, ((uint64_t)((bit >> 6) == 3)) << (bit & 0x3F));
+    r->d[3] = secp256k1_u128_to_u64(&t);
 #ifdef VERIFY
-    VERIFY_CHECK((t >> 64) == 0);
+    secp256k1_u128_rshift(&t, 64);
+    VERIFY_CHECK(secp256k1_u128_to_u64(&t) == 0);
```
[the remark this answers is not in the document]
roconnor-blockstream: I'll compromise with `VERIFY_CHECK(secp256k1_u128_hi_u64(&t) == 0);`.

src/scalar_4x64_impl.h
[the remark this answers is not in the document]
real-or-random: Oh,, I didn't didn't see that the ASM part uses the last line.  

> . I suppose I could duplicate the last line if you think that is worthwhile.

I don't think so.

src/int128_struct_impl.h
roconnor-blockstream: Done something similar to the suggestion.

src/int128_native_impl.h
```diff
@@ -0,0 +1,76 @@
...
+}
+
+static SECP256K1_INLINE void secp256k1_i128_accum_mul(secp256k1_int128 *r, int64_t a, int64_t b) {
+   int128_t ab = (int128_t)a * b;
+   VERIFY_CHECK(0 <= ab ? *r <= INT128_MAX - ab : INT128_MIN - ab <= *r);
+   *r += ab;
+}
+
+static SECP256K1_INLINE void secp256k1_i128_det(secp256k1_int128 *r, int64_t a, int64_t b, int64_t c, int64_t d) {
+   *r = (int128_t)a * d - (int128_t)b * c;
```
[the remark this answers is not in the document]
roconnor-blockstream: Done.

src/int128.h
```diff
@@ -0,0 +1,55 @@
...
+ */
+static SECP256K1_INLINE void secp256k1_i128_rshift(secp256k1_int128 *r, unsigned int b);
+
+static SECP256K1_INLINE int64_t secp256k1_i128_to_i64(const secp256k1_int128 *a);
+
+static SECP256K1_INLINE void secp256k1_i128_from_i64(secp256k1_int128 *r, int64_t a);
+
+static SECP256K1_INLINE int secp256k1_i128_eq_var(const secp256k1_int128 *a, const secp256k1_int128 *b);
+
+static SECP256K1_INLINE int secp256k1_i128_check_bit(const secp256k1_int128 *r, unsigned int n);
```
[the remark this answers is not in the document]
roconnor-blockstream: Done.

src/modinv64_impl.h
```diff
@@ -60,6 +70,13 @@ static int secp256k1_modinv64_mul_cmp_62(const secp256k1_modinv64_signed62 *a, i
     }
     return 0;
 }
+
+/* Check if the determinant of t is equal to 1 << n. */
+static int secp256k1_modinv64_det_bit(const secp256k1_modinv64_trans2x2 *t, unsigned int n) {
```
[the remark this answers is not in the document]
roconnor-blockstream: Done.

src/scalar_4x64_impl.h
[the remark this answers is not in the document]
roconnor-blockstream: Done.

src/int128_struct_impl.h
```diff
@@ -0,0 +1,137 @@
...
+static SECP256K1_INLINE void secp256k1_i128_mul(secp256k1_int128 *r, int64_t a, int64_t b) {
+   r->hi = (uint64_t)secp256k1_mulh(a, b);
+   r->lo = (uint64_t)a * (uint64_t)b;
+}
+
+static SECP256K1_INLINE void secp256k1_i128_accum_mul(secp256k1_int128 *r, int64_t a, int64_t b) {
+   uint64_t lo = (uint64_t)a * (uint64_t)b;
+   uint64_t hi = (uint64_t)secp256k1_mulh(a, b) + (~lo < r->lo);
+   VERIFY_CHECK((r->hi <= 0x7fffffffffffffffu && hi <= 0x7fffffffffffffffu) <= (r->hi + hi <= 0x7fffffffffffffffu));
+   VERIFY_CHECK((r->hi > 0x7fffffffffffffffu && hi > 0x7fffffffffffffffu) <= (r->hi + hi > 0x7fffffffffffffffu));
```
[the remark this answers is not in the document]
roconnor-blockstream: Done.

src/scalar_4x64_impl.h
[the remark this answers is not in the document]
roconnor-blockstream: Done.

real-or-random:
This is about 1% slower on my machine with gcc -O2 but about 1.5% faster with gcc -03. 

gcc 12.1.0 -O2, x86_64, asm disabled:
```

master:

ecdsa_verify                  ,    53.7       ,    53.9       ,    54.3    
ecdsa_sign                    ,    39.5       ,    39.7       ,    40.9    


pr:

ecdsa_verify                  ,    54.2       ,    54.4       ,    55.2    
ecdsa_sign                    ,    39.7       ,    39.9       ,    40.6  
```

gcc 12.1.0 -O3, x86_64, asm disabled:
```

master:

ecdsa_verify                  ,    52.2       ,    52.5       ,    53.2    
ecdsa_sign                    ,    39.6       ,    39.7       ,    40.3   

pr:

ecdsa_verify                  ,    51.6       ,    51.8       ,    52.4    
ecdsa_sign                    ,    39.5       ,    39.7       ,    40.4  
```



And it seems that gcc version 12 is finally faster at -O3 than at -O2...


Do we care about the loss in O2? I guess no.

roconnor-blockstream:
Is the generated asm different?

real-or-random (approved):
ACK 7edf1a373236f3200dee77aae12d9993b9d12995 careful code inspection, a lot of tests with ` --with-test-override-wide-multiply=int128_struct` (which is not covered by CI yet), I have *not* looked at the VST proofs in details

real-or-random:
> Is the generated asm different?

Yeah, we could check but I imagine it will be a lot of work. It would probably make sense to first run the internal benchmarks and see where we lose performance. And then look at the asm of these functions.

src/int128_struct_impl.h
```diff
@@ -0,0 +1,154 @@
...
+#endif
+
+static SECP256K1_INLINE void secp256k1_u128_mul(secp256k1_uint128 *r, uint64_t a, uint64_t b) {
+   r->hi = secp256k1_umulh(a, b);
+   r->lo = a * b;
+}
+
+static SECP256K1_INLINE void secp256k1_u128_accum_mul(secp256k1_uint128 *r, uint64_t a, uint64_t b) {
+   uint64_t lo = a * b;
+   r->hi += secp256k1_umulh(a, b) + (~lo < r->lo);
```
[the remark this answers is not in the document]
real-or-random: Yes, yours seems better, see https://godbolt.org/z/d593Yxocq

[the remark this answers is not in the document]
sipa: The generated code still uses two multiplication instructions it seems, which is very suboptimal (though obviously better than 4 32-bit multiplications).

I think the right approach is using `_mul128` and `_umul128` on MSVC actually, which compute both the low and high half. I think that all code in this int128_struct implementation which needs mulh/umulh also needs the lower half, so perhaps it's better to write things in function of those primitives?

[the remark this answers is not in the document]
roconnor-blockstream: As I recall `_mul128` and `_umul128` are not available on 32-bit platforms.  That said, it still makes sense to abstract through those function.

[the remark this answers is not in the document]
sipa: My understanding is:
* x64 has all four: `_mulh`, `_umulh`, `_mul128`, and `_umul128`.
* arm64 only has `_mulh` and `_umulh`.
* x86 and arm32 have none

The documentation is a bit inconsistent (one MSVC site says the _mulh are only available on x64, but another page lists them as available for arm64 too).

Still, this matches what the hardware provides. x64 does have a 64x64->128 instruction, while arm64 only has separate low and high 64x64->64 multiplication instructions.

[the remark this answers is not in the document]
sipa: Trying to make a patch that implements this.

src/int128_struct_impl.h
```diff
@@ -0,0 +1,154 @@
...
+}
+
+static SECP256K1_INLINE void secp256k1_i128_mul(secp256k1_int128 *r, int64_t a, int64_t b) {
+   r->hi = (uint64_t)secp256k1_mulh(a, b);
+   r->lo = (uint64_t)a * (uint64_t)b;
+}
+
+static SECP256K1_INLINE void secp256k1_i128_accum_mul(secp256k1_int128 *r, int64_t a, int64_t b) {
+   uint64_t lo = (uint64_t)a * (uint64_t)b;
+   uint64_t hi = (uint64_t)secp256k1_mulh(a, b) + (~lo < r->lo);
```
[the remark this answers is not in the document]
roconnor-blockstream: The VERIFY_CHECKS only operate on `hi` values, so I don't think the checks will need changes.

src/int128_struct_impl.h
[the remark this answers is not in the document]
sipa: See https://github.com/sipa/secp256k1/commits/pr1000_mul128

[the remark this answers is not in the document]
sipa: I made a few small changes to the branch above, but I'm not going to touch it anymore.

Feel free to ignore, include, or squash into this PR (perhaps subject to translating the changes to the formal proof code?).

src/int128_impl.h
```diff
@@ -0,0 +1,25 @@
+/***********************************************************************
+ * Copyright (c) 2014 Pieter Wuille                                    *
```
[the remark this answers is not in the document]
roconnor-blockstream: Arguably this file is a derived work of your `scalar_impl.h` file.  But I'm happy to remove it if you like.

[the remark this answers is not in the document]
roconnor-blockstream: Done.

src/int128_struct_impl.h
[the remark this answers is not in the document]
roconnor-blockstream: Merged your commit.

src/int128_struct_impl.h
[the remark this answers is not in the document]
roconnor-blockstream: Merged your commit.

src/int128_struct_impl.h
```diff
@@ -0,0 +1,154 @@
...
+   r->lo += a;
+}
+
+/* Unsigned (logical) right shift.
+ * Non-constant time in n.
+ */
+static SECP256K1_INLINE void secp256k1_u128_rshift(secp256k1_uint128 *r, unsigned int n) {
+   VERIFY_CHECK(n < 128);
+   if (n >= 64) {
+     r->lo = (r->hi) >> (n-64);
```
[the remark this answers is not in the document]
roconnor-blockstream: Merged your commit.

src/int128_struct_impl.h
```diff
@@ -0,0 +1,154 @@
...
+}
+
+static SECP256K1_INLINE void secp256k1_u128_accum_mul(secp256k1_uint128 *r, uint64_t a, uint64_t b) {
+   uint64_t lo = a * b;
+   r->hi += secp256k1_umulh(a, b) + (~lo < r->lo);
+   r->lo += lo;
+}
+
+static SECP256K1_INLINE void secp256k1_u128_accum_u64(secp256k1_uint128 *r, uint64_t a) {
+   r->hi += (r->lo > ~a);
```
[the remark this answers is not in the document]
roconnor-blockstream: Merged your commit.

src/int128.h
```diff
@@ -0,0 +1,62 @@
...
+
+#if defined(SECP256K1_INT128_NATIVE)
+#include "int128_native.h"
+#elif defined(SECP256K1_INT128_STRUCT)
+#include "int128_struct.h"
+#else
+#error "Please select int128 implementation"
+#endif
+
+static SECP256K1_INLINE void secp256k1_u128_mul(secp256k1_uint128 *r, uint64_t a, uint64_t b);
```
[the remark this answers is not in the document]
roconnor-blockstream: Done.

real-or-random:
@roconnor-blockstream Did you update your VST proofs to the changes suggested by Pieter? (If no, are planning to do so?)

src/int128.h
```diff
@@ -0,0 +1,79 @@
...
+
+/* Multiply two unsigned 64-bit values a and b and write the result to r. */
+static SECP256K1_INLINE void secp256k1_u128_mul(secp256k1_uint128 *r, uint64_t a, uint64_t b);
+
+/* Multiply two unsigned 64-bit values a and b and add the result to r.
+ * The final result is taken moduluo 2^128.
+ */
+static SECP256K1_INLINE void secp256k1_u128_accum_mul(secp256k1_uint128 *r, uint64_t a, uint64_t b);
+
+/* Add a unsigned 64-bit value a to r.
```
real-or-random, line 23:
```suggestion
/* Add an unsigned 64-bit value a to r.
```

src/int128.h
```diff
@@ -0,0 +1,79 @@
...
+
+/* Tests if r is strictly less than to 2^n.
+ * n must be strictly less than 128.
+ */
+static SECP256K1_INLINE int secp256k1_u128_check_bits(const secp256k1_uint128 *r, unsigned int n);
+
+/* Multiply two signed 64-bit values a and b and write the result to r. */
+static SECP256K1_INLINE void secp256k1_i128_mul(secp256k1_int128 *r, int64_t a, int64_t b);
+
+/* Multiply two unsigned 64-bit values a and b and add the result to r.
```
real-or-random, line 50:
```suggestion
/* Multiply two signed 64-bit values a and b and add the result to r.
```

src/int128.h
```diff
@@ -0,0 +1,79 @@
...
+ * Any overflow or underflow is undefined behaviour.
```
real-or-random, line 51: I think "any" means just in the final addition, not in the multiplication. (It's pretty clear if you think about it but may be worth clarifying.)

src/int128.h
```diff
@@ -0,0 +1,79 @@
...
+
+/* Compute a*d - b*c from signed 64-bit values and write the result to r. */
+static SECP256K1_INLINE void secp256k1_i128_det(secp256k1_int128 *r, int64_t a, int64_t b, int64_t c, int64_t d);
+
+/* Signed (arithmetic) right shift.
+ * Non-constant time in b.
+ */
+static SECP256K1_INLINE void secp256k1_i128_rshift(secp256k1_int128 *r, unsigned int b);
+
+/* Return the low 64-bits of a 128-bit value intepreted as an signed 64-bit value. */
```
real-or-random, line 63:
```suggestion
/* Return the low 64-bits of a 128-bit value interpreted as an signed 64-bit value. */
```

src/int128.h
```diff
@@ -0,0 +1,79 @@
...
+ */
+static SECP256K1_INLINE void secp256k1_u128_rshift(secp256k1_uint128 *r, unsigned int n);
+
+/* Return the low 64-bits of a 128-bit value as an unsigned 64-bit value. */
+static SECP256K1_INLINE uint64_t secp256k1_u128_to_u64(const secp256k1_uint128 *a);
+
+/* Return the high 64-bits of a 128-bit value as an unsigned 64-bit value. */
+static SECP256K1_INLINE uint64_t secp256k1_u128_hi_u64(const secp256k1_uint128 *a);
+
+/* Write a unsigned 64-bit value to r. */
```
real-or-random, line 39:
```suggestion
/* Write an unsigned 64-bit value to r. */
```

real-or-random (commented):
just some nits and typos in the new comments

src/int128_impl.h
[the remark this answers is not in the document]
real-or-random: Not sure if it's better without a copyright header.  

> Arguably this file is a derived work of your `scalar_impl.h` file. :) But I'm happy to remove it if you like.

Then you may be violating the MIT license because it requires you to keep the copyright notice. :D


For me personally, I don't care. In the interest of getting this merged, I'll ACK this with and without notice, and with any of your names. But if you ask me, we should really see if I can make copyright notices easier in the future.

src/int128.h
roconnor-blockstream: Done.

src/int128.h
roconnor-blockstream: Done.

src/int128.h
roconnor-blockstream: Done.

src/int128.h
roconnor-blockstream: Done.

src/int128.h
roconnor-blockstream: Done.

roconnor-blockstream:
I have not updated the proofs, but I do plan to do so.  I don't suggest waiting for them, but if you prefer to wait that's fine.

I'm working on a second pass through the proofs anyways in an attempt to add more automatic tactics to help me, so it is slow going.

roconnor-blockstream:
While I continue to refactor my proofs to automate the tedious bits, I have now got to the point where the revised code in the int128_struct file is now proved correct.

roconnor-blockstream:
I'm done my pass at refactoring my proof.  There isn't too much to say.  The new proofs are less tedious than they were before.  You are welcome to "browse" the proof at <https://htmlpreview.github.io/?https://github.com/ElementsProject/simplicity/blob/027e46ca91dc7eecf5a05707798edf17e72612c3/alectryon/verif_int128_impl.v.html> for whatever that's worth.  Again, the `*_spec` definitions are probably the most important bits as they define the pre and post-conditions of the functions that form the statements of what I've proved.

The most notable change is that the specification for `secp256k1_i128_to_i64` was amended to remove the precondition that the input be a 64-bit signed integer value.  Apparently libsecp256k1 on occasion will cast signed int128 numbers to signed int64 numbers, expecting to truncate the bits.  That seems kinda dicey to me, but okay.

real-or-random:
> Apparently libsecp256k1 on occasion will cast signed int128 numbers to signed int64 numbers, expecting to truncate the bits. That seems kinda dicey to me, but okay.

Do you have an example where this is unexpected (according to your impression)?

roconnor-blockstream:
Looking into it more closely it seems the sketchy cases of casting large signed int128 numbers to int64 is always then immediately masked with `& M62`, so it isn't so dicey in practice.

To check on this I created a commit that splits out [a `secp256k1_i128_to_u64` function](https://github.com/roconnor-blockstream/secp256k1/blob/da641861e84ada1e1298e6d0fa77b5b8c03daeb2/src/int128_native_impl.h#L62) and [adds a `VERIFY_CHECK` to the `secp256k1_i128_to_i64` function](https://github.com/roconnor-blockstream/secp256k1/blob/da641861e84ada1e1298e6d0fa77b5b8c03daeb2/src/int128_native_impl.h#L67).  This passes the tests.  It will fail the tests with only the `VERIFY_CHECK` added.

I'm not proposing we add that commit.  I only made it for validation purposes. Although we could add it if people really think that is a good idea, but that would be a separate PR.

chinaverse:
unsubscribe



---Original---
From: ***@***.***&gt;
Date: Tue, Oct 25, 2022 22:46 PM
To: ***@***.***&gt;;
Cc: ***@***.***&gt;;
Subject: Re: [bitcoin-core/secp256k1] Synthetic int128 type. (PR #1000)




 
Looking into it more closely it seems the sketchy cases of casting large signed int128 numbers to int64 is always then immediately masked with &amp; M62, so it isn't so dicey in practice.
 
To check on this I created a commit that splits out a secp256k1_i128_to_u64 function and adds a VERIFY_CHECK to the secp256k1_i128_to_i64 function.  This passes the tests.  It will fail the tests with only the VERIFY_CHECK added.
 
I'm not proposing we add that commit.  I only made it for validation purposes. Although we could add it if people really think that is a good idea, but that would be a separate PR.

src/int128.h
```diff
@@ -0,0 +1,79 @@
...
+/* Multiply two unsigned 64-bit values a and b and write the result to r. */
+static SECP256K1_INLINE void secp256k1_u128_mul(secp256k1_uint128 *r, uint64_t a, uint64_t b);
+
+/* Multiply two unsigned 64-bit values a and b and add the result to r.
+ * The final result is taken moduluo 2^128.
+ */
+static SECP256K1_INLINE void secp256k1_u128_accum_mul(secp256k1_uint128 *r, uint64_t a, uint64_t b);
+
+/* Add an unsigned 64-bit value a to r.
+ * The final result is taken moduluo 2^128.
```
jonasnick, line 24: nit: moduluo typo x2

jonasnick (commented):
I created a commit (https://github.com/jonasnick/secp256k1/commits/20211026_int128-jn) that adds int128_struct tests to CI (including an MSVC test that uses native (u)mul128). Feel free to cherry-pick and squash.

I also ran some benchmarks just to confirm that the changes to `field_5x52_int128` and `scalar_4x64` don't unexpectedly affect performance (https://gist.github.com/jonasnick/aed0f881aabe9b9f5bd769fcf3efbef8)

src/int128_struct_impl.h
```diff
@@ -0,0 +1,177 @@
...
+static SECP256K1_INLINE uint64_t secp256k1_umul128(uint64_t a, uint64_t b, uint64_t* hi) {
+    *hi = __umulh(a, b);
+    return a * b;
+}
+
+static SECP256K1_INLINE int64_t secp256k1_mul128(int64_t a, int64_t b, int64_t* hi) {
+    *hi = __mulh(a, b);
+    return a * b;
+}
+#    endif
```
jonasnick, line 23: Did we test this code path on ARM64 MSVC? Should we remove this until this is tested in CI?

src/int128.h
```diff
@@ -0,0 +1,79 @@
...
+/* Return the high 64-bits of a 128-bit value as an unsigned 64-bit value. */
+static SECP256K1_INLINE uint64_t secp256k1_u128_hi_u64(const secp256k1_uint128 *a);
+
+/* Write an unsigned 64-bit value to r. */
+static SECP256K1_INLINE void secp256k1_u128_from_u64(secp256k1_uint128 *r, uint64_t a);
+
+/* Tests if r is strictly less than to 2^n.
+ * n must be strictly less than 128.
+ */
+static SECP256K1_INLINE int secp256k1_u128_check_bits(const secp256k1_uint128 *r, unsigned int n);
```
jonasnick, line 45: `secp256k1_u128_check_bits, secp256k1_i128_dissip_mul, secp256k1_i128_det, secp256k1_i128_from_i64, secp256k1_i128_from_i64, secp256k1_i128_check_pow2` are not used when compiled without VERIFY. Should we put them behind `#ifdef VERIFY`?

jonasnick (commented):
I added some basic tests to [my branch](https://github.com/jonasnick/secp256k1/commits/20211026_int128-jn) mostly to confirm that the code is working as expected. But I think also generally it would be better if the code was tested. The tests also exercise a branch that was not executed before (`i128_rshift` with `b >= 64`).

> This also opens up the possibility of removing the 32-bit field and scalar implementations should that ever be desired.

I didn't really investigate this but benchmarks on my (64-bit) machine indicate that this won't be easy:

```
./configure --with-test-override-wide-multiply=int64 --with-asm=no
        ecdsa_verify                  103.0
./configure --with-test-override-wide-multiply=int128_struct --with-asm=no
        ecdsa_verify                  188.0
```

src/int128.h
roconnor-blockstream: I don't know.

src/int128_struct_impl.h
roconnor-blockstream: I'm not sure how our testing infrastructure works.

src/int128.h
jonasnick: One reason to do this would be if not doing it would negatively impact test coverage metrics. However, I just tested this with gcov and it doesn't care about these functions at all (presumably because they don't exist in the compilation result).

src/int128.h
roconnor-blockstream: Fixed.

roconnor-blockstream:
@jonasnick I've merged your tests into one commit.

src/int128_struct_impl.h
jonasnick: That's a sufficient approximate test. Thanks @sipa. Resolving this thread.

src/int128.h
jonasnick: It could be done to give a better indication for readers what this code is for, but not a big concern, I agree. Resolving this conversation.

jonasnick:
@sipa

> We could drop the optimization, but unless there is a concern that having code for optimization present would be misinterpreted as "supported", we could also just leave it for now.

I agree, my remark was mostly about adding code paths that have never been run at all.

src/modinv64_impl.h
```diff
@@ -7,17 +7,25 @@
 #ifndef SECP256K1_MODINV64_IMPL_H
 #define SECP256K1_MODINV64_IMPL_H
 
+#include "int128.h"
```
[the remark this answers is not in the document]
roconnor-blockstream: @real-or-random Since you wrote https://github.com/real-or-random/secp256k1/commit/af6cffda12e0bd2e595eae5ccdffec2d96b8eef3 do you wish to comment?

src/int128_struct_impl.h
roconnor-blockstream: Would you like me to add a comment "Beware of bugs in the above code;  We have only proved it correct, not tested it."?

roconnor-blockstream:
Presumably we can add configuration options so that our MSVC test infrastructure can at least exercise this code path on X86.

real-or-random:
> @roconnor-blockstream We could add a configure/config flag for forcing the use of _mulh/_umulh over _mul128/_umul128, and set it for one of our MSVC CI configs.

https://github.com/real-or-random/secp256k1/tree/202211-int128-mulh-override implemented here, ready to be cherry-picked

jonasnick:
ACK a340d9500a9c45e5c261174f48b3eb18b3b3647d

src/int128_impl.h
```diff
@@ -0,0 +1,18 @@
...
+
+#if defined(SECP256K1_WIDEMUL_INT128)
+#  if defined(SECP256K1_INT128_NATIVE)
+#    include "int128_native_impl.h"
+#  elif defined(SECP256K1_INT128_STRUCT)
+#    include "int128_struct_impl.h"
+#  else
+#    error "Please select int128 implementation"
+#  endif
+#endif
```
roconnor-blockstream, line 16: Although this file should never be included when `SECP256K1_WIDEMUL_INT128` is not defined, we should have an error message here just in case that situation arises.  Otherwise confusing error messages happen.

real-or-random: This file is actually included unconditionally in `secp256k1.c`. (One could argue that this is a bad idea, but the nice thing is that it keeps the preprocessor 
 logic

PR #1002, opened 2021-10-30
Make aux_rnd32==NULL behave identical to 0x0000..00.

sipa:
BIP340's default signing algorithm always requires an aux_rnd argument, but permits using an all-zero one when no randomness is available.

Make secp256k1_schnorrsig_sign follow this even when aux_rnd32==NULL, by treating the same as if an all-zero byte array was provided as input.

### DIFF
--- a/include/secp256k1_schnorrsig.h
+++ b/include/secp256k1_schnorrsig.h
@@ -111,7 +111,8 @@ typedef struct {
  *  In:    msg32: the 32-byte message being signed.
  *       keypair: pointer to an initialized keypair.
  *    aux_rand32: 32 bytes of fresh randomness. While recommended to provide
- *                this, it is only supplemental to security and can be NULL. See
+ *                this, it is only supplemental to security and can be NULL. A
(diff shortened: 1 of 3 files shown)


junderw:
ACK 5324f89

real-or-random:
Concept ACK. We should stay in line with the BIP.

jonasnick (commented):
ACK 5324f8942dd322448fae6c9b225ecac2854fa7e2

Without this PR, the doc for `secp256k1_nonce_function_bip340` is wrong:

>  *  If a data pointer is passed, it is assumed to be a pointer to 32 bytes of
>  *  auxiliary random data as defined in BIP-340. If the data pointer is NULL,
>  *  the nonce derivation procedure follows BIP-340 by setting the auxiliary
>  *  random data to zero.

elichai:
ACK 5324f8942dd322448fae6c9b225ecac2854fa7e2

real-or-random (approved):
utACK 5324f8942dd322448fae6c9b225ecac2854fa7e2

PR #1004, opened 2021-11-06
ecmult: fix definition of STRAUSS_SCRATCH_OBJECTS

jonasnick:
This bug was introduced in 7506e064d791e529d2e57bb52c156deb33b897ef by adding
an allocation but not updating the constant.

### DIFF
--- a/src/ecmult_impl.h
+++ b/src/ecmult_impl.h
@@ -47,7 +47,7 @@
 
 /* The number of objects allocated on the scratch space for ecmult_multi algorithms */
 #define PIPPENGER_SCRATCH_OBJECTS 6
-#define STRAUSS_SCRATCH_OBJECTS 6
+#define STRAUSS_SCRATCH_OBJECTS 7
 
 #define PIPPENGER_MAX_BUCKET_WINDOW 12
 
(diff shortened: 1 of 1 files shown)


robot-dreams:
Thanks for catching this, I confirmed that:
- `secp256k1_ecmult_strauss_batch` now has 7 calls to `secp256k1_scratch_alloc` instead of 6
- What matters for `STRAUSS_SCRATCH_OBJECTS` is the number of allocations (not the total size being allocated), since it's used to account for per-allocation alignment padding

I agree with @real-or-random 's suggestion. In addition, do the functions `secp256k1_strauss_scratch_size` and `secp256k1_pippenger_scratch_size` also need to be kept in sync when allocations are updated?

jonasnick:
Good idea. Added comments.

real-or-random (approved):
ACK 60bf8890df5360148df921f26d8dc4d667dd5926

real-or-random:
@robot-dreams  Want give a final review here?

robot-dreams:
ACK 60bf8890df5360148df921f26d8dc4d667dd5926

PR #1006, opened 2021-11-08
refactor: Use && in boolean context to avoid compiler warning

MarcoFalke:
This fixes a compiler warning:

```
./src/ecdsa_impl.h:312:12: warning: use of bitwise '&' with boolean operands [-Wbitwise-instead-of-logical]
    return !secp256k1_scalar_is_zero(sigr) & !secp256k1_scalar_is_zero(sigs);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                           &&

### DIFF
--- a/src/ecdsa_impl.h
+++ b/src/ecdsa_impl.h
@@ -309,7 +309,7 @@ static int secp256k1_ecdsa_sig_sign(const secp256k1_ecmult_gen_context *ctx, sec
     /* P.x = order is on the curve, so technically sig->r could end up being zero, which would be an invalid signature.
      * This is cryptographically unreachable as hitting it requires finding the discrete log of P.x = N.
      */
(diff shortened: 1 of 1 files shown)


MarcoFalke:
I didn't try to reproduce this, but I randomly saw it in logs on a 32-bit compilation with clang-14 (trunk) and some odd compiler flags like `-O0`.

sipa:
:)

MarcoFalke:
This is in clang-14 (for now): https://github.com/llvm/llvm-project/commit/f62d18ff140f67a8776a7a3c62a75645d8d540b5

sipa:
If there is a need to silence anything, I think the correct approach is suggested there: cast one of the inputs to `&` to int.

PR #1008, opened 2021-11-08
bench.c: add `--help` option and ci: move env variables

### LINKED ISSUE #1005: Improve documentation for benchmarks
A follow-up PR could add a `--help` which would also mention the `get_iters` environment variable (I have to look up in the source how it's called exactly every time).

_Originally posted by @jonasnick in https://github.com/bitcoin-core/secp256k1/pull/991#pullrequestreview-799481944_

We could also remove the indirection in `cirrus.sh` (https://github.com/bitcoin-core/secp256k1/blob/master/ci/cirrus.sh#L35-L37) and define the "right" constants directly in `cirrus.yml`. I tend to look the constants up in `cirrus.yml` and then see the wrong constants without the `SECP256K1_` prefix.

siv2r:
Fixes #1005

**Changes:**
- added `--help` option to `bench.c`
    - `help()` function prints the help to command-line
    - `have_invalid_args()` checks if the user has entered an invalid argument
- moved `secp256k1_bench_iters` and `secp256k1_test_iters` environment variables declaration to `.cirrus.yml`

### DIFF
--- a/src/bench.c
+++ b/src/bench.c
@@ -11,6 +11,51 @@
 #include "util.h"
 #include "bench.h"
 
+void help(int default_iters) {
+    printf("Benchmarks the following algorithms:\n");
+    printf("    - ECDSA signing/verification\n");
+
+#ifdef ENABLE_MODULE_ECDH
+    printf("    - ECDH key exchange (optional module)\n");
+#endif
+
+#ifdef ENABLE_MODULE_RECOVERY
+    printf("    - Public key recovery (optional module)\n");
+#endif
+
+#ifdef ENABLE_MODULE_SCHNORRSIG
+    printf("    - Schnorr signatures (optional module)\n");
+#endif
+
+    printf("\n");
+    printf("The default number of iterations for each benchmark is %d. This can be\n", default_iters);
+    printf("customized using the SECP256K1_BENCH_ITERS environment variable.\n");
+    printf("\n");
+    printf("Usage: ./bench [args]\n");
+    printf("By default, all benchmarks will be run.\n");
+    printf("args:\n");
+    printf("    help              : display this help and exit\n");
+    printf("    ecdsa             : all ECDSA algorithms--sign, verify, recovery (if enabled)\n");
+    printf("    ecdsa_sign        : ECDSA siging algorithm\n");
+    printf("    ecdsa_verify      : ECDSA verification algorithm\n");
+
+#ifdef ENABLE_MODULE_RECOVERY
+    printf("    ecdsa_recover     : ECDSA public key recovery algorithm\n");
+#endif
+
+#ifdef ENABLE_MODULE_ECDH
+    printf("    ecdh              : ECDH key exchange algorithm\n");
+#endif
+
+#ifdef ENABLE_MODULE_SCHNORRSIG
+    printf("    schnorrsig        : all Schnorr signature algorithms (sign, verify)\n");
+    printf("    schnorrsig_sign   : Schnorr sigining algorithm\n");
+    printf("    schnorrsig_verify : Schnorr verification algorithm\n");
+#endif
+
+    printf("\n");
+}
+
 typedef struct {
     secp256k1_context *ctx;
     unsigned char msg[32];
@@ -95,7 +140,52 @@ int main(int argc, char** argv) {
     bench_verify_data data;
 
     int d = argc == 1;
-    int iters = get_iters(20000);
+    int default_iters = 20000;
+    int iters = get_iters(default_iters);
+
+    /* Check for invalid user arguments */
+    char* valid_args[] = {"ecdsa", "verify", "ecdsa_verify", "sign", "ecdsa_sign", "ecdh", "recover",
+                         "ecdsa_recover", "schnorrsig", "schnorrsig_verify", "schnorrsig_sign"};
+    size_t valid_args_size = sizeof(valid_args)/sizeof(valid_args[0]);
+    int invalid_args = have_invalid_args(argc, argv, valid_args, valid_args_size);
+
+    if (argc > 1) {
+        if (have_flag(argc, argv, "-h")
+           || have_flag(argc, argv, "--help")
+           || have_flag(argc, argv, "help")) {
+            help(default_iters);
+            return 0;
+        } else if (invalid_args) {
+            fprintf(stderr, "./bench: unrecognized argument.\n\n");
+            help(default_iters);
+            return 1;
+        }
+    }
+
+/* Check if the user tries to benchmark optional module without building it */
+#ifndef ENABLE_MODULE_ECDH
+    if (have_flag(argc, argv, "ecdh")) { 
+        fprintf(stderr, "./bench: ECDH module not enabled.\n");
+        fprintf(stderr, "Use ./configure --enable-module-ecdh.\n\n");
+        return 1;
+    }
+#endif
+
+#ifndef ENABLE_MODULE_RECOVERY
+    if (have_flag(argc, argv, "recover") || have_flag(argc, argv, "ecdsa_recover")) { 
+        fprintf(stderr, "./bench: Public key recovery module not enabled.\n");
+        fprintf(stderr, "Use ./configure --enable-module-recovery.\n\n");
+        return 1;
+    }
+#endif
+
+#ifndef ENABLE_MODULE_SCHNORRSIG
+    if (have_flag(argc, argv, "schnorrsig") || have_flag(argc, argv, "schnorrsig_sign") || have_flag(argc, argv, "schnorrsig_verify")) { 
+        fprintf(stderr, "./bench: Schnorr signatures module not enabled.\n");
+        fprintf(stderr, "Use ./configure --enable-module-schnorrsig.\n\n");
+        return 1;
+    }
+#endif
 
     /* ECDSA verification benchmark */
     data.ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN | SECP256K1_CONTEXT_VERIFY);
--- a/src/bench.h
+++ b/src/bench.h
@@ -116,9 +116,35 @@ void run_benchmark(char *name, void (*benchmark)(void*, int), void (*setup)(void
 int have_flag(int argc, char** argv, char *flag) {
     char** argm = argv + argc;
     argv++;
-    while (argv != NULL && argv != argm) {
+    while (argv != argm) {
         if (strcmp(*argv, flag) == 0) {
             return 1;
+        }
+        argv++;
+    }
+    return 0;
+}
+
+/* takes an array containing the arguments that the user is allowed to enter on the command-line
+   returns:
+      - 1 if the user entered an invalid argument
+      - 0 if all the user entered arguments are valid */
+int have_invalid_args(int argc, char** argv, char** valid_args, size_t n) {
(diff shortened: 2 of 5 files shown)


src/bench.h
```diff
@@ -125,6 +125,28 @@ int have_flag(int argc, char** argv, char *flag) {
     return 0;
 }
 
+/* check for invalid command line argument */
```
real-or-random, line 128: You have a nice description of this function (including return values) in the commit message, but not here. Maybe move it here to a comment. It's not that helpful in the commit message.

src/bench.c
```diff
@@ -11,6 +11,31 @@
...
 #include "bench.h"
 
+void help(char **argv) {
+    printf("Benchmarks the following algorithms: \n");
+    printf("    1. ECDSA signing/verification\n");
+    printf("    2. ECDH key exchange (optional module)\n");
+    printf("    3. Public key recovery (optional module)\n");
+    printf("    4. Schnorr signatures (optional module)\n");
+    printf("\n");
+    printf("The default number of iterations for each benchmark is 20000. This can be customized\n");
```
real-or-random, line 21: Text looks nice here! Maybe make the default a variable and then print it here. Then whenever we change the default, the number gets updated automatically in the help text.

src/bench.c
```diff
@@ -11,6 +11,31 @@
 #include "util.h"
 #include "bench.h"
 
+void help(char **argv) {
+    printf("Benchmarks the following algorithms: \n");
```
real-or-random, line 15: nit: trailing space behind the `:`.

real-or-random (commented):
Thanks for the PR!

src/bench.c
```diff
@@ -97,6 +122,25 @@ int main(int argc, char** argv) {
...
+                         "ecdsa_recover", "schnorrsig", "schnorrsig_verify", "schnorrsig_sign"};
+    int valid_args_size = 11;
+    int invalid_args = have_invalid_args(argc, argv, valid_args, valid_args_size);
+
+    if(argc > 1) {
+        if(have_flag(argc, argv, "-h")
+           || have_flag(argc, argv, "--help")
+           || have_flag(argc, argv, "help")) {
+            help(argv);
+            return 1;
```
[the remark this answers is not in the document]
siv2r: `ls --help` does say:

> Exit status:
 0  if OK,
 1  if minor problems (e.g., cannot access subdirectory),
 2  if serious trouble (e.g., cannot access command-line argument).

I used `return 1` here since the [bench_ecmult.c](https://github.com/bitcoin-core/secp256k1/blob/master/src/bench_ecmult.c#L295) followed this conversion.

Shall I also modify `bench_ecmult's` help to `return 0`?

src/bench.h
```diff
@@ -125,6 +125,28 @@ int have_flag(int argc, char** argv, char *flag) {
...
+/* valid_arg_list = list of allowed user arguments,  n = size of valid_arg_list */
+int have_invalid_args(int argc, char** argv, char** valid_arg_list, int n) {
```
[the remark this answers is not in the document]
siv2r: Okay

src/bench.c
siv2r: Nice!

src/bench.c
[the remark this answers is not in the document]
real-or-random: Oh indeed, I think that's wrong in bench_ecmult too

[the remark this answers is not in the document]
siv2r: Okay, will update both.

src/bench.c
siv2r: I initially thought of creating a macro for the default value in `bench.h`. The other benchmarks (like `bench_ecmlult` uses 10000) use a different default iteration value so, I declared a local variable in the main function for the default value.

siv2r:
Incorporated all the suggestions :)

real-or-random (approved):
ACK 89939feb33c3e3b616848f23de9d624d1bec8060

siv2r:
There is one corner case that is not covered. The binary prints the table header if the user tries to benchmark an optional module without building it. Shall I add a check for this? Or is it too much?

The check would look something like this:
```c
#ifndef ENABLE_MODULE_SCHNORRSIG
    if (have_flag(argc, argv, "schnorrsig") || have_flag(argc, argv, "schnorrsig_sign") || have_flag(argc, argv, "schnorrsig_verify")) { 
         fprintf(stderr, "%s: schnorr module not enabled.\n\n", argv[0]);
         return 1;
    }
#endif
```

src/bench.c
```diff
@@ -11,6 +11,31 @@
...
+    printf("args:\n");
+    printf("    help              : display this help and exit\n");
+    printf("    ecdsa             : all ecdsa algorithms (sign, verify, recover)\n");
+    printf("    ecdsa_sign        : ecdsa siging algorithm\n");
+    printf("    ecdsa_verify      : ecdsa verification algorithm\n");
+    printf("    ecdsa_recover     : ecdsa public key recovery algorithm\n");
+    printf("    ecdh              : ecdh key exchange algorithm\n");
+    printf("    schnorrsig        : all schnorr signature algorithms (sign, verify)\n");
+    printf("    schnorrsig_sign   : schnorr sigining algorithm\n");
+    printf("    schnorrsig_verify : schnorr verification algorithm\n");
```
real-or-random, line 35: nit: Could capitalize ECDSA, ECDH and Schnorr here in the right column if you're on it.

siv2r: Sure!

real-or-random:
> There is one corner case that is not covered. The binary prints the table header if the user tries to benchmark an optional module without building it. Shall I add a check for this? Or is it too much?

It won't hurt but yeah since this is "only" the benchmarks, it's not really necessary (like my other two comments). This PR is clearly an improvement even without those issues addressed.

siv2r:
**New changes:**
      - check for optional modules
      - capitalized ECDH, ECDSA, Schnorr in help text
      - hardcoded `./bench` value

real-or-random (approved):
ACK c69060d303604f8dd03e915e9e6a8379076633ea

src/bench.c
```diff
@@ -11,6 +11,31 @@
 #include "util.h"
 #include "bench.h"
 
+void help(int default_iters) {
+    printf("Benchmarks the following algorithms:\n");
+    printf("    1. ECDSA signing/verification\n");
```
sipa, line 16: These lines can be surrounded with `#ifdef ENABLE_MODULE_ECDH` etc to make it only print the relevant ones (same below).

src/bench.c
```diff
@@ -11,6 +11,31 @@
...
+    printf("    2. ECDH key exchange (optional module)\n");
+    printf("    3. Public key recovery (optional module)\n");
+    printf("    4. Schnorr signatures (optional module)\n");
+    printf("\n");
+    printf("The default number of iterations for each benchmark is %d. This can be\n", default_iters);
+    printf("customized using the SECP256K1_BENCH_ITERS environment variable.\n");
+    printf("\n");
+    printf("Usage: ./bench [args]\n");
```
sipa, line 24: Common practice is to use `argv[0]` for printing the binary name (which keeps working if it is renamed etc, or on systems where executable have extensions).

src/bench.c
```diff
@@ -95,7 +120,52 @@ int main(int argc, char** argv) {
...
+            return 0;
+        } else if (invalid_args) {
+            fprintf(stderr, "./bench: unrecognized argument.\n\n");
+            help(default_iters);
+            return 1;
+        }
+    }
+
+    /* Check if the user tries to benchmark optional module without building it */
+    #ifndef ENABLE_MODULE_ECDH
```
sipa, line 146: I find it somewhat ugly that there is this much module-specific information inside `main()`, but I don't have a concrete suggestion on how to cleanly do it otherwise.

I'll try to address this in a follow-up PR if no simple solution can be used here.

src/bench.c
real-or-random: I specifically asked for the hardcoded earlier in this PR: https://github.com/bitcoin-core/secp256k1/pull/1008#discussion_r745698996

src/bench.c
siv2r: we can create a separate functions for these checks (like `int check_ecdh_enabled(argc,argv)`) in the `bench.h`. Not sure if this is a good approach.

src/bench.c
sipa: Ah, good point. I retract my comment.

siv2r:
**Update:**
- added `#ifdef ENABLE_MODULE_ECDH` etc to `void help()`, prints only the relevant module-info.

src/bench.c
real-or-random: I think when @sipa said "same below", he meant that the same could be done in the "args:" list printed further down. This is still missing. Changes look good otherwise.

siv2r: Oh, I forgot about the args list. Thanks!

siv2r:
> * added `#ifdef ENABLE_MODULE_ECDH` etc to `void help()`, prints only the relevant module-info.

I forgot to add these to the arg list, which has been taken care of now.

real-or-random (approved):
ACK 8d3efeeb1de16d8051ba874a10c44c16fe379530

src/bench.h
```diff
@@ -125,6 +125,32 @@ int have_flag(int argc, char** argv, char *flag) {
...
+   returns:
+      - 1 if the user entered an invalid argument
+      - 0 if all the user entered arguments are valid */
+int have_invalid_args(int argc, char** argv, char** valid_args, size_t n) {
+    size_t i;
+    int found_valid;
+    char** argm = argv + argc;
+    argv++;
+
+    while (argv != NULL && argv != argm) {
```
sipa, line 138: `argv` will never be NULL, as it is the result of applying `argv++;` one or more times on the input. Did you mean to write `*argv != NULL` here?

siv2r: Nice catch! 
I used `argv != NULL` since the [`have_flag`](https://github.com/bitcoin-core/secp256k1/blob/master/src/bench.h#L116) function used the same loop condition to traverse user arguments.

`*argv != NULL` should be used (as you mentioned) in both this function and the `have_flag` function.

siv2r: I initially assumed `argv` will behave as a normal 2d array so, I thought both `argv` and `*argv` will value will be the same (i.e. base address of argument string). That does not seem to be the case.

sipa: The type of `argv` is `char**`, so it's a pointer to pointer, not an array.

siv2r:
**Update:**
- removed `argv == NULL` check in the while loop traversing the user arguments. 

The `argv` cannot have a `NULL` value so, `*argv != NULL` might be the proper check here (as suggested in https://github.com/bitcoin-core/secp256k1/pull/1008#discussion_r762167607). Then, I realized we also don't need the `*argv != NULL` check in the while loop since, `argv != argm` condition will take care of this. The value of `*argm` will be `NULL` if `argm == argv + argc` ([C89 draft](http://port70.net/~nsz/c/c89/c89-draft.html#2.1.2.2))

sipa:
utACK 592661c22f56736099f83700be8cf280f8a963ff

real-or-random (approved):
ACK 592661c22f56736099f83700be8cf280f8a963ff

PR #1009, opened 2021-11-09
refactor: Use (int)&(int) in boolean context to avoid compiler warning

MarcoFalke:
This one should *really* be only a refactor with the goal to silence static analysis warnings. clang-14 (trunk) recently added one in commit https://github.com/llvm/llvm-project/commit/f62d18ff140f67a8776a7a3c62a75645d8d540b5 and I expect other tools will offer similar warnings.

Follow up to #1006, which was not a refactor.

### DIFF
--- a/src/ecdsa_impl.h
+++ b/src/ecdsa_impl.h
@@ -304,12 +304,12 @@ static int secp256k1_ecdsa_sig_sign(const secp256k1_ecmult_gen_context *ctx, sec
     high = secp256k1_scalar_is_high(sigs);
     secp256k1_scalar_cond_negate(sigs, high);
     if (recid) {
-            *recid ^= high;
+        *recid ^= high;
     }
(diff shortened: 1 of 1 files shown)


MarcoFalke:
I checked that this patch doesn't change the objdump, but 1006 did

jonasnick (commented):
ACK 16d132215cab68e57798927301268518bc1c3bf8

This is a should be a no-op also because `!` already results in type int. Quite a bit weird to have to add this cast, but apparently it can catch bugs in the real world (TM).

MarcoFalke:
> This is a should be a no-op also because ! already results in type int

Are you sure that `operator!` results in type `int`? At least in C++, it results in type `bool`: https://docs.microsoft.com/en-us/cpp/cpp/logical-negation-operator-exclpt

jonasnick:
See [C89 standard](https://port70.net/~nsz/c/c89/c89-draft.html#3.3.3.3). Also, there is no bool type in C89.

PR #1010, opened 2021-11-10
doc: Minor fixes in safegcd_implementation.md

robot-dreams:


doc/safegcd_implementation.md
```diff
@@ -576,7 +576,7 @@ i = N # divsteps left to do
 while True:
     # Get rid of all bottom zeros at once. In the first iteration, g may be odd and the following
     # lines have no effect (until "if eta < 0").
-    zeros = min(i, count_trailing_zeros(g))
+    zeros = i if g == 0 else min(i, count_trailing_zeros(g))
```
sipa, line 579: Hmm, it feels like this is mostly due to `count_trailing_zeros` giving the unintuitive *-1* as result when fed *0*.

Maybe we should have `count_trailing_zeros` return *N* on *0*?

robot-dreams: Yeah, that's much more clear!

doc/safegcd_implementation.md
```diff
@@ -570,7 +570,10 @@ bits efficiently, which is possible on most platforms; it is abstracted here as
 ```python
 def count_trailing_zeros(v):
     """For a non-zero value v, find z such that v=(d<<z) for some odd d."""
```
sipa, line 572: The comment would need to be updated.

real-or-random (approved):
ACK https://github.com/bitcoin-core/secp256k1/pull/1010/commits/dc9b6853b72b9a492cad230623670e89157525ca

sipa:
ACK dc9b6853b72b9a492cad230623670e89157525ca

PR #1011, opened 2021-11-10
ci: Enable -g if we set CFLAGS manually

real-or-random:
This enables sanitizers to output line numbers in stack traces.


jonasnick (commented):
ACK 72de1359e953390dc2f1ab5a59dd1a4057000acb

PR #1019, opened 2021-11-24
build: don't append valgrind CPPFLAGS if not installed (macOS)

fanquake:
Valgrinds CPPFLAGS, i.e `-I/usr/local/opt/valgrind/include`, are currently added to CPPFLAGS, regardless of whether valgrind is installed. This changes configure so that they are only added if valgrind is available. i.e the output of `brew list --versions valgrind` is non-null.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -43,8 +43,8 @@ case $host_os in
          # These Homebrew packages may be keg-only, meaning that they won't be found
          # in expected paths because they may conflict with system files. Ask
          # Homebrew where each one is located, then adjust paths accordingly.
-         valgrind_prefix=`$BREW --prefix valgrind 2>/dev/null`
-         if test x$valgrind_prefix != x; then
+         if $BREW list --versions valgrind >/dev/null; then
+           valgrind_prefix=`$BREW --prefix valgrind 2>/dev/null`
            VALGRIND_CPPFLAGS="-I$valgrind_prefix/include"
          fi
        else


real-or-random (approved):
ACK 214042a170c860523b7aad2a251b0dbfbaefb235

This is what is done in Core
https://github.com/bitcoin/bitcoin/commit/d3ef947524a07f8d7fbad5b95781ab6cacb1cb49

real-or-random:
@hebasto It's a trivial change but can you test this on MacOS so that we have at least one ACK from somehow who can actually run this on MacOS? :)

fanquake:
@real-or-random fwiw I also tested this on a macOS machine.

real-or-random:
> @real-or-random fwiw I also tested this on a macOS machine.

Okay sure. I'm fine merging as-is, and almost did, but then also remembered that @hebasto is quick usually.

configure.ac
```diff
@@ -43,8 +43,8 @@ case $host_os in
          # These Homebrew packages may be keg-only, meaning that they won't be found
          # in expected paths because they may conflict with system files. Ask
          # Homebrew where each one is located, then adjust paths accordingly.
-         valgrind_prefix=`$BREW --prefix valgrind 2>/dev/null`
-         if test x$valgrind_prefix != x; then
+         if $BREW list --versions valgrind >/dev/null; then
+           valgrind_prefix=`$BREW --prefix valgrind 2>/dev/null`
```
hebasto, line 47:
```suggestion
           valgrind_prefix=$($BREW --prefix valgrind 2>/dev/null)
```

See https://github.com/koalaman/shellcheck/wiki/SC2006

hebasto (approved):
ACK 214042a170c860523b7aad2a251b0dbfbaefb235, tested on macOS Big Sur 11.6.1 (20G224, Intel).

configure.ac
real-or-random: Yeah, the Bash FAQ says " `...` is the legacy syntax required by only the very oldest of non-POSIX-compatible bourne-shells." But then if we want to perform that cleanup, we should do this change in the entire file...

fanquake: Yep. I wasn't going to make other random cleanups.

PR #1020, opened 2021-11-24
doc: remove use of <0xa0> "no break space"

fanquake:
This is miscellaneous, but I don't think these were being used on purpose?


real-or-random (approved):
ACK 812ff5c74745e451f1a9de83b5bd0d0c18c75e5f

siv2r:
ACK 812ff5c

PR #1026, opened 2021-12-03
ecdh: Add test computing shared_secret=basepoint with random inputs

real-or-random:
I was playing around with the ECDH code and wrote this test on way.

### DIFF
--- a/src/modules/ecdh/tests_impl.h
+++ b/src/modules/ecdh/tests_impl.h
@@ -60,7 +60,7 @@ void test_ecdh_generator_basepoint(void) {
 
     s_one[31] = 1;
     /* Check against pubkey creation when the basepoint is the generator */
-    for (i = 0; i < 100; ++i) {
+    for (i = 0; i < 2 * count; ++i) {
         secp256k1_sha256 sha;
         unsigned char s_b32[32];
         unsigned char output_ecdh[65];
--- a/src/modules/ecdh/tests_impl.h
+++ b/src/modules/ecdh/tests_impl.h
@@ -123,10 +123,43 @@ void test_bad_scalar(void) {
     CHECK(secp256k1_ecdh(ctx, output, &point, s_overflow, ecdh_hash_function_test_fail, NULL) == 0);
 }
 
+/** Test that ECDH(sG, 1/s) == ECDH((1/s)G, s) == ECDH(G, 1) for a few random s. */
+void test_result_basepoint(void) {
+    secp256k1_pubkey point;
+    secp256k1_scalar rand;
+    unsigned char s[32];
+    unsigned char s_inv[32];
+    unsigned char out[32];
+    unsigned char out_inv[32];
+    unsigned char out_base[32];
+    int i;
+
+    unsigned char s_one[32] = { 0 };
+    s_one[31] = 1;
+    CHECK(secp256k1_ec_pubkey_create(ctx, &point, s_one) == 1);
+    CHECK(secp256k1_ecdh(ctx, out_base, &point, s_one, NULL, NULL) == 1);
+
+    for (i = 0; i < 2 * count; i++) {
+        random_scalar_order(&rand);
+        secp256k1_scalar_get_b32(s, &rand);
+        secp256k1_scalar_inverse(&rand, &rand);
+        secp256k1_scalar_get_b32(s_inv, &rand);
+
+        CHECK(secp256k1_ec_pubkey_create(ctx, &point, s) == 1);
+        CHECK(secp256k1_ecdh(ctx, out, &point, s_inv, NULL, NULL) == 1);
+        CHECK(secp256k1_memcmp_var(out, out_base, 32) == 0);
+
+        CHECK(secp256k1_ec_pubkey_create(ctx, &point, s_inv) == 1);
+        CHECK(secp256k1_ecdh(ctx, out_inv, &point, s, NULL, NULL) == 1);
+        CHECK(secp256k1_memcmp_var(out_inv, out_base, 32) == 0);
+    }
+}
+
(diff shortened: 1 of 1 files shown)


robot-dreams (commented):
ACK 18fe75b19e5cf5e1d4c6192e09a0d6346566b098

Agreed that we should get `ecdh(g^s, s_inv) == ecdh(g^s_inv, s) == ecdh(g, 1)` for any `s, s_inv` pair.

Non-blocking suggestion, would the intent be more clear if you explicitly calculate and compare against the expected constant value `ecdh(g, 1)`, instead of comparing `out` against the previous iteration's `out_eq`?

```c
void test_result_basepoint(void) {
    secp256k1_pubkey point;
    secp256k1_scalar rand;
    unsigned char s[32];
    unsigned char s_inv[32];
    unsigned char out[32];
    unsigned char out_inv[32];
    unsigned char out_base[32];
    int i;

    unsigned char s_one[32] = { 0 };
    s_one[31] = 1;
    CHECK(secp256k1_ec_pubkey_create(ctx, &point, s_one) == 1);
    CHECK(secp256k1_ecdh(ctx, out_base, &point, s_one, NULL, NULL) == 1);

    for (i = 0; i < 64; i++) {
        random_scalar_order(&rand);
        secp256k1_scalar_get_b32(s, &rand);
        secp256k1_scalar_inverse(&rand, &rand);
        secp256k1_scalar_get_b32(s_inv, &rand);

        CHECK(secp256k1_ec_pubkey_create(ctx, &point, s) == 1);
        CHECK(secp256k1_ecdh(ctx, out, &point, s_inv, NULL, NULL) == 1);
        CHECK(secp256k1_memcmp_var(out, out_base, 32) == 0);

        CHECK(secp256k1_ec_pubkey_create(ctx, &point, s_inv) == 1);
        CHECK(secp256k1_ecdh(ctx, out_inv, &point, s, NULL, NULL) == 1);
        CHECK(secp256k1_memcmp_var(out_inv, out_base, 32) == 0);
    }
}
```

real-or-random:
> Non-blocking suggestion, would the intent be more clear if you explicitly calculate and compare against the expected constant value `ecdh(g, 1)`, instead of comparing `out` against the previous iteration's `out_eq`?

Probably yes. I'll change this.

src/modules/ecdh/tests_impl.h
```diff
@@ -123,10 +123,40 @@ void test_bad_scalar(void) {
...
+void test_result_basepoint(void) {
+    secp256k1_pubkey point;
+    secp256k1_scalar rand;
+    unsigned char s[32];
+    unsigned char s_inv[32];
+    unsigned char out[32];
+    unsigned char out_eq[32];
+    int i;
+
+    for (i = 0; i < 64; i++) {
```
real-or-random, line 135: This should depend on `count`

robot-dreams: `test_ecdh_generator_basepoint` above also has a similar issue, right?

uvhw (changes requested):
@uvhw @bitchaincoin @totongcute

src/modules/ecdh/tests_impl.h
real-or-random: No, it turns out that's a known-output test with a fixed output.

real-or-random: > No, it turns out that's a known-output test with a fixed output.

Nevermind, I misremembered this. I saw a test where I assumed it should depend on `count` but it turned out to be a known-output test. But it wasn't this one.  

Fixed.

real-or-random:
Forced-pushed to rebase and add you as a coauthor.

robot-dreams:
ACK 6bc47a170319efb7bc6b07ce1eeba3dbbb5e6ca9

Test passes locally, first commit is directly from https://github.com/bitcoin-core/secp256k1/pull/1026#pullrequestreview-824152063 (only with iteration count changed).

Thanks for adding me, gotta pump those commit numbers up :)

real-or-random:
> Thanks for adding me, gotta pump those commit numbers up :)

I think it's good and right to acknowledge people. :)

But here's an off-topic rant about code stats, which I think are always misleading:
 - Commit scope is very different among contributors and stages of a project.
 - Line numbers are a stupid metric, too. Wanna have ten thousand lines? Change the format of the precomputed tables. 
 - Commits and line numbers say nothing about the time invested in a change or its value (whatever that even is).
 - Often less is more. Except for docs where more is often better. 
 - Common stats ignore code review (which is by far the largest chunk of time spent here) and conceptual contributions. 

So I think it's better for our wellbeing to ignore the numbers... At least that's why I decided to to.

robot-dreams:
Hopefully it's clear I was joking about pumping those rookie numbers up. I strongly agree with all of that! Goodhart's Law and etc. Especially about less is more:

> If we wish to count lines of code, we should not regard them as "lines produced" but as "lines spent": the current conventional wisdom is so foolish as to book that count on the wrong side of the ledger. --Dijkstra

real-or-random:
> Hopefully it's clear I was joking about pumping those rookie numbers up. 

Yes, I assumed you were joking. :) (Sorry if my post looked like the opposite... I was in the mood to write that rant down.)

jonasnick (commented):
Concept ACK

Would be helpful to add a comment describing what the test does, similar to what @robot-dreams  says above:
> [...] ecdh(g^s, s_inv) == ecdh(g^s_inv, s) == ecdh(g, 1) for any s, s_inv pair.

real-or-random:
@jonasnick Indeed! Fixed.

jonasnick (commented):
ACK 3531a43b5bc739838f5634afcfd02bdbef71b1ef

PR #1029, opened 2021-12-03
Simpler and faster ecdh skew fixup

peterdettman:
This PR adds a `_gej_cmov` method, with accompanying tests, and uses it to simplify the skew fixup at the end of `_ecmult_const`.

In the existing code, `_wnaf_const` chooses a skew of either 1 or 2, and `_ecmult_const` needs a call to `_ge_set_gej` (which does an expensive field inversion internally) and some overly-complicated conversions to/from `_ge_storage` so that `_ge_storage_cmov` can be used to select what value to add for the fixup.

This PR uses a simpler scheme where `_wnaf_const` chooses a skew of 0 or 1 and no longer needs special handling for scalars with value negative one. A new `_gej_cmov` method is used at the end of `_ecmult_const` for const-time optional addition to adjust the final result for the skew. Finally, the skew fixup is moved to before the global-Z adjustment, and the precomputed table entries (for 1P, &#955;(1P)) are used for the skew fixup, saving a field multiply and ensuring the fixup is done on the same isomorphism as the ladder.

The resulting `_wnaf_const` and `_ecmult_const` are shorter and simpler, and the ECDH benchmark is around 5% faster (64bit, i7). 

Edit: Updated description once the final scope was clear.

### DIFF
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -642,6 +642,14 @@ static void secp256k1_ge_from_storage(secp256k1_ge *r, const secp256k1_ge_storag
     r->infinity = 0;
 }
 
+static SECP256K1_INLINE void secp256k1_gej_cmov(secp256k1_gej *r, const secp256k1_gej *a, int flag) {
+    secp256k1_fe_cmov(&r->x, &a->x, flag);
+    secp256k1_fe_cmov(&r->y, &a->y, flag);
+    secp256k1_fe_cmov(&r->z, &a->z, flag);
+
+    r->infinity ^= (r->infinity ^ a->infinity) & flag;
+}
+
 static SECP256K1_INLINE void secp256k1_ge_storage_cmov(secp256k1_ge_storage *r, const secp256k1_ge_storage *a, int flag) {
     secp256k1_fe_storage_cmov(&r->x, &a->x, flag);
     secp256k1_fe_storage_cmov(&r->y, &a->y, flag);
--- a/src/ecmult_const_impl.h
+++ b/src/ecmult_const_impl.h
@@ -56,7 +56,6 @@ static void secp256k1_ecmult_odd_multiples_table_globalz_windowa(secp256k1_ge *p
     secp256k1_fe_cmov(&(r)->y, &neg_y, (n) != abs_n); \
 } while(0)
 
-
 /** Convert a number to WNAF notation.
  *  The number becomes represented by sum(2^{wi} * wnaf[i], i=0..WNAF_SIZE(w)+1) - return_val.
  *  It has the following guarantees:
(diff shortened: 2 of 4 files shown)


peterdettman:
Added another commit that changes the skew to be either 0 or 1. Total performance gain is maybe 5% now (less for 32bit).

apoelstra (approved):
ACK 6498a6c939f967de0ffe8d0475e9921a967080c9

These are great cleanups! LOL at the first commit -- I don't know why I didn't think to add a `gej_cmov`, instead adding 10+ lines of conversion logic so that I could use `ge_storage_cmov`.

apoelstra:
I can't recall exactly why my original code added 1 or 2, rather than 1 or 0. I remember thinking that "adding zero or non-zero can't be done in a constant-time way" but then I wound up needing to use a `cmov` anyway.

I think that it simply didn't occur to me to use your "always do a corrective addition, then use `cmov` to undo it if necessary" logic.

peterdettman:
> I can't recall exactly why my original code added 1 or 2, rather than 1 or 0

My somewhat foggy recollection is that much of the earliest EC side-channel literature was focused only on performing a constant number of point operations, and didn't initially care about memory timings, so they would even just use if/else to choose which of P or 2P to add. If it weren't for the endomorphism complicating things, we could probably modify the scalar instead (by adding N when it's even) and/or use a signed-digit encoding, which I would consider the "modern style".

src/group_impl.h
```diff
@@ -642,6 +642,14 @@ static void secp256k1_ge_from_storage(secp256k1_ge *r, const secp256k1_ge_storag
     r->infinity = 0;
 }
 
+static SECP256K1_INLINE void secp256k1_gej_cmov(secp256k1_gej *r, const secp256k1_gej *a, int flag) {
```
real-or-random, line 645: I think it's not strictly necessary but can you add a test for this?

real-or-random (commented):
Concept ACK

I read this code a few times and I always wondered about the 1, 2 thing.

src/group_impl.h
peterdettman: Done.

apoelstra:
ACK 1120790139baee6726ddf6eeb2d29a92bc174d4f

peterdettman:
Moved the skew fixup to before the global Z fixup by using the table entries. Saves an _fe_mul and is just a bit nicer to have all the operations on the same isomorphism. Also rebased.

apoelstra:
ACK e82144ed

Nifty! Agreed, it is easier to follow now with fewer temp variables.

real-or-random (approved):
ACK e82144edfb7673d9a5eeb2b556d08be5223835ac

peterdettman:
I've no more changes in mind.

sipa:
The PR description seems outdated now.

peterdettman:
> The PR description seems outdated now.

I've updated the description to match the full PR.

sipa:
ACK e82144edfb7673d9a5eeb2b556d08be5223835ac

PR #1031, opened 2021-12-04
Faster const-time modinv divsteps

peterdettman:
Changes to _divsteps_59 (_30) that give maybe 4% speed improvement to const-time modinv on 64 bit. I see a larger gain on 32 bit but measured on 64 bit so might not be real.

Start the result matrix scaled by 2^62 (resp. 2^30) and shift q, r down instead of u, v up at each step (should make life easier for vectorization). Since we're always shifting away the LSB of g, q, r, we can avoid doing a full negation for x, y, z (after a few tweaks). Then it makes sense to switch zeta back to delta (I confined this change to the local method for the moment).

### DIFF
--- a/doc/safegcd_implementation.md
+++ b/doc/safegcd_implementation.md
@@ -155,14 +155,14 @@ do one division by *2<sup>N</sup>* as a final step:
 ```python
 def divsteps_n_matrix(delta, f, g):
     """Compute delta and transition matrix t after N divsteps (multiplied by 2^N)."""
-    u, v, q, r = 1, 0, 0, 1 # start with identity matrix
+    u, v, q, r = 1<<N, 0, 0, 1<<N # start with identity matrix (scaled by 2^N)
     for _ in range(N):
         if delta > 0 and g & 1:
-            delta, f, g, u, v, q, r = 1 - delta, g, (g - f) // 2, 2*q, 2*r, q-u, r-v
+            delta, f, g, u, v, q, r = 1 - delta, g, (g-f)//2, q, r, (q-u)//2, (r-v)//2
         elif g & 1:
-            delta, f, g, u, v, q, r = 1 + delta, f, (g + f) // 2, 2*u, 2*v, q+u, r+v
+            delta, f, g, u, v, q, r = 1 + delta, f, (g+f)//2, u, v, (q+u)//2, (r+v)//2
         else:
-            delta, f, g, u, v, q, r = 1 + delta, f, (g    ) // 2, 2*u, 2*v, q  , r
+            delta, f, g, u, v, q, r = 1 + delta, f, (g  )//2, u, v, (q  )//2, (r  )//2
     return delta, (u, v, q, r)
 ```
 
@@ -414,9 +414,9 @@ operations (and hope the C compiler isn't smart enough to turn them back into br
 divstep can be written instead as (compare to the inner loop of `gcd` in section 1).
 
 ```python
-    x = -f if delta > 0 else f         # set x equal to (input) -f or f
+    x = f if delta > 0 else -f         # set x equal to (input) f or -f
     if g & 1:
-        g += x                         # set g to (input) g-f or g+f
+        g -= x                         # set g to (input) g-f or g+f
         if delta > 0:
             delta = -delta
             f += g                     # set f to (input) g (note that g was set to g-f before)
@@ -433,19 +433,21 @@ that *-v == (v ^ -1) - (-1)*. Thus, if we have a variable *c* that takes on valu
 Using this we can write:
 
 ```python
-    x = -f if delta > 0 else f
+    x = f if delta > 0 else -f
 ```
 
 in constant-time form as:
 
 ```python
-    c1 = (-delta) >> 63
+    # Compute c1=0 if delta>0 and c1=-1 if delta<=0.
+    c1 = (delta - 1) >> 63
     # Conditionally negate f based on c1:
     x = (f ^ c1) - c1
 ```
 
-To use that trick, we need a helper mask variable *c1* that resolves the condition *&delta;>0* to *-1*
-(if true) or *0* (if false). We compute *c1* using right shifting, which is equivalent to dividing by
+To use that trick, we need a helper mask variable *c1* that resolves the condition *&delta;&leq;0* to *-1*
+(if true) or *0* (if false). We compute *c1* by first subtracting *1*, which results in a negative value
+if and only if *&delta;&leq;0*. That is then right shifted, which is equivalent to dividing by
 the specified power of *2* and rounding down (in Python, and also in C under the assumption of a typical two's complement system; see
 `assumptions.h` for tests that this is the case). Right shifting by *63* thus maps all
 numbers in range *[-2<sup>63</sup>,0)* to *-1*, and numbers in range *[0,2<sup>63</sup>)* to *0*.
@@ -454,7 +456,7 @@ Using the facts that *x&0=0* and *x&(-1)=x* (on two's complement systems again),
 
 ```python
     if g & 1:
-        g += x
+        g -= x
 ```
 
 as:
@@ -463,7 +465,7 @@ as:
     # Compute c2=0 if g is even and c2=-1 if g is odd.
     c2 = -(g & 1)
     # This masks out x if g is even, and leaves x be if g is odd.
-    g += x & c2
+    g -= x & c2
 ```
 
 Using the conditional negation trick again we can write:
@@ -478,7 +480,7 @@ as:
 
 ```python
     # Compute c3=-1 if g is odd and delta>0, and 0 otherwise.
-    c3 = c1 & c2
+    c3 = ~c1 & c2
     # Conditionally negate delta based on c3:
     delta = (delta ^ c3) - c3
 ```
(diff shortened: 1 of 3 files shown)


sipa:
Wowwow, calm down!

sipa:
See https://github.com/sipa/secp256k1/commits/pr1031 for two commits on top of this. One renames your delta to theta (as the variable operated on is really defined as delta-1/2 compared to delta as defined in the paper). The other updates the writeup to reflect the improvements here.

Concept ACK in any case. I'll look over the code in more detail, but at least this convinces me it's correct.

real-or-random:
As I understand, this should always be faster and we don't need lots of benchmarks to confirm. 

>  The other updates the writeup to reflect the improvements here.

Nice, I wanted to ask for this.

sipa:
Updated my branch with additional assertions and a few more improvements to the documentation.

Code review ACK. I'm happy to do my changes on top in a separate PR, though it's probably better to include them, as a few code comments are out of that as-is in this PR. Feel free to cherry-pick or otherwise include my suggested changes.

peterdettman:
> Wowwow, calm down!

@sipa Never!

So, after I PR'ed this, I sat there looking at the code and thinking: you know, there's a lot of low zeros in u/v/q/r... you could almost fit an f/g in there...

(a few moments later)

https://github.com/peterdettman/secp256k1/tree/vec_modinv

I won't spoiler it, but you might like to benchmark const-time inverses in that branch...

sipa:
@peterdettman Nice!

Benchmarks on Ryzen 5950x (default compilation options, GCC 11.2.0, `SECP256K1_BENCH_ITERS=1000000 ./bench_internal inverse`:

```
master:
field_inverse                 ,     1.27      ,     1.27      ,     1.28   
field_inverse_var             ,     0.867     ,     0.871     ,     0.874  
pr1031:
field_inverse                 ,     1.26      ,     1.26      ,     1.27   
field_inverse_var             ,     0.846     ,     0.860     ,     0.868  
vec_modinv:
field_inverse                 ,     0.986     ,     0.995     ,     1.01   
field_inverse_var             ,     0.868     ,     0.870     ,     0.873  
vec_modinv, using the const-time matrix in the vartime code:
field_inverse                 ,     1.00      ,     1.00      ,     1.01   
field_inverse_var             ,     0.866     ,     0.871     ,     0.876  
```

This seems to at least suggest the option of throwing out the vartime matrix computation code, and using this new constant-time code for both?

peterdettman:
Wow, that's even more improvement than on my machine. Around 4900 cycles, right?

For the _var version, if copying the vec approach at minimum the vartime version would want to escape at the halfway mark if g1 is 0 (mod remaining steps) already, so I think there would still be a separate version. I haven't really looked at it though.

sipa:
@peterdettman A vartime version of the vec matrix code could perhaps also do more than 59 iterations.

peterdettman:
It could trivially do 60. The current core vec loop tops out at 30. In theory I think it can do 31, but it's not a given until I work it through.

sipa:
Benchmarks on Cortex-A53 (default compilation options, GCC 9.3.0, `./bench_internal inverse`):

```
master:
field_inverse                 ,    12.5       ,    12.5       ,    12.5    
field_inverse_var             ,     7.26      ,     7.27      ,     7.27
pr1031:
field_inverse                 ,    11.9       ,    11.9       ,    11.9    
field_inverse_var             ,     7.26      ,     7.27      ,     7.27 
vec_modinv:
field_inverse                 ,    11.1       ,    11.1       ,    11.1    
field_inverse_var             ,     7.26      ,     7.27      ,     7.27   
vec_modinv using const-time matrix for vartime:
field_inverse                 ,    11.2       ,    11.2       ,    11.2    
field_inverse_var             ,     9.83      ,     9.83      ,     9.83  
```

Looks like we shouldn't throw out the vartime matrix code just yet...

sipa:
ACK 6afd499f53e94d48a7fd90ff345d47101a6c6e41

@robot-dreams Perhaps you're interested in reviewing this too?

@peterdettman Given that your later (and much more significant) improvement builds on top of this one (at least conceptually, using initially-shifted u,v,q,r variables) and doesn't work for 32 bit yet, I'd vote to get this change merged as-is, and do the rest in a follow-up. WDYT?

peterdettman:
@sipa Yeah, we may as well merge this, since the other one is still under experimentation and will in any case require significant write-up and VERIFY checks.

doc/safegcd_implementation.md
```diff
@@ -651,37 +665,41 @@ Here we need the negated modular inverse, which is a simple transformation of th
...
+        x, y, z = f ^ c1, u ^ c1, v ^ c1
         c2 = -(g & 1)
-        # Conditionally add x, y, z to g, q, r.
-        g, q, r = g + (x & c2), q + (y & c2), r + (z & c2)
-        c1 &= c2                     # reusing c1 here for the earlier c3 variable
-        zeta = (zeta ^ c1) - 1       # inlining the unconditional zeta decrement here
+        # Conditionally subtract x, y, z from g, q, r.
+        g, q, r = g - (x & c2), q - (y & c2), r - (z & c2)
+        c3 = ~c1 & c2
+        # Conditionally negate delta, and then increment it by 1.
```
robot-dreams, line 696: Nit: emphasize that you're in fact negating delta (not theta)?

```suggestion
        # Conditionally negate delta (theta = delta - 1/2), and then increment it by 1.
```

robot-dreams (commented):
Thanks for the tag @sipa ! You're right, I'm definitely interested in this change 😄

I took an initial look at the new writeup and it looks correct so far, but I still need to sleep on it and then review the new C implementation carefully.

src/modinv64_impl.h
```diff
@@ -145,72 +145,72 @@ typedef struct {
     int64_t u, v, q, r;
 } secp256k1_modinv64_trans2x2;
 
-/* Compute the transition matrix and eta for 59 divsteps (where zeta=-(delta+1/2)).
+/* Compute the transition matrix and theta for 59 divsteps (where theta=delta-1/2)
  * Note that the transformation matrix is scaled by 2^62 and not 2^59.
```
robot-dreams, line 149: Is it worth elaborating on some of these parameter choices?

```suggestion
/* Compute the transition matrix and theta for 59 divsteps (where theta=delta-1/2)
 * We use 59 instead of 62 because we need 590 total divsteps.
 * However, we scale the transformation matrix by 2^62 and not 2^59 to
 * use the precomputed inverse of modulus mod 2^62.
 */
```

src/modinv32_impl.h
```diff
@@ -168,68 +168,71 @@ typedef struct {
...
-        VERIFY_CHECK((q * f0 + r * g0) == g << i);
-        /* Compute conditional masks for (zeta < 0) and for (g & 1). */
-        c1 = zeta >> 31;
+        /* f must always be odd */
+        VERIFY_CHECK((f & 1) == 1);
+        /* Minimum trailing zeros count for matrix elements decreases in each iteration */
+        VERIFY_CHECK(!((u | v | q | r) & (0xFFFFFFFFULL >> (i + 2))));
+        /* Applying the matrix so far to the initial f,g gives current f,g. */
+        VERIFY_CHECK((u >> (30 - i)) * f0 + (v >> (30 - i)) * g0 == f << i);
+        VERIFY_CHECK((q >> (30 - i)) * f0 + (r >> (30 - i)) * g0 == g << i);
```
robot-dreams, line 194: Just wondering, why don't these (and the analogous checks in `modinv64_impl.h`) cast to higher precision, like line 234 below?

robot-dreams (commented):
ACK 6afd499f53e94d48a7fd90ff345d47101a6c6e41

I checked the writeup updates, made sure the updated Python code produced the same transition matrices as before (using various small parameters), and then compared the Python vs. C implementations line by line.

Am I understanding correctly that removing 3 subtractions from the inner loop gives such a dramatic speedup?

src/modinv32_impl.h
peterdettman: They are equalities modulo 2^32 (resp. 2^64) only, because that's the precision we are given f0 and g0 to. Technically maybe we only even care about modulo 2^30 (resp. 2^62), but the extra bits are there and should agree anyway. If divsteps were being done at full precision (i.e. directly on the full-size F,G) the equality would still hold, but a significant feature of safegcd is to be able to operate in limited precision most of the time, only updating the full F,G values when we need more bits to work on.

The latter matrix determinant check is full precision because, well, we're checking the exact value and not making any assumptions that we got u/v/q/r right.

peterdettman:
> Am I understanding correctly that removing 3 subtractions from the inner loop gives such a dramatic speedup?

@sipa gave numbers from a different branch of mine that is more radical. This PR probably only gives a few percent, and maybe nothing if the processor could have executed them in parallel anyway.

src/modinv32_impl.h
robot-dreams: Thanks! Yes, the determinant check doesn't involve the low-order-only `f0`, `g0`, that makes sense.

src/modinv64_impl.h
peterdettman: How's this?

```
/* Compute the transition matrix and theta for 59 divsteps (where theta=delta-1/2)
 * Although only 59 divsteps are performed, the resulting transformation matrix
 * is scaled by 2^62 to allow reuse of _update_de_62 and _update_fg_62 between
 * _modinv64 and _modinv64_var.
```

peterdettman:
@sipa I'm currently assuming that you will get to the nits that @robot-dreams had above about safegcd_implementation.md, but I'll sort them out next week if not.

sipa:
See my updated branch https://github.com/sipa/secp256k1/commits/pr1031 . I've edited the "Update safegcd writeup to reflect the code"  commit to address @robot-dreams's comments.

doc/safegcd_implementation.md
```diff
@@ -155,14 +155,14 @@ do one division by *2<sup>N</sup>* as a final step:
 ```python
 def divsteps_n_matrix(delta, f, g):
     """Compute delta and transition matrix t after N divsteps (multiplied by 2^N)."""
-    u, v, q, r = 1, 0, 0, 1 # start with identity matrix
+    u, v, q, r = 2**N, 0, 0, 2**N # start with identity matrix
```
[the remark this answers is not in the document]
peterdettman: Done.

doc/safegcd_implementation.md
```diff
@@ -433,13 +433,13 @@ that *-v == (v ^ -1) - (-1)*. Thus, if we have a variable *c* that takes on valu
...
 ```python
-    x = -f if delta > 0 else f
+    x = f if delta > 0 else -f
 ```
 
 in constant-time form as:
 
 ```python
-    c1 = (-delta) >> 63
+    c1 = delta >> 63
```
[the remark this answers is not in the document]
peterdettman: Done (2 occurrences).

doc/safegcd_implementation.md
peterdettman: Now "Conditionally complement theta, and then increment it by 1.", since we are now using theta directly.

peterdettman:
Thanks, @sipa. Squashed one spelling fix into my last commit.

real-or-random:
Closing in favor of #1197

Issue #159, opened 2014-12-12
Possibility of magnitude restrictions on group element field values


peterdettman:
The group operations seem to generally assume input x,y,z can be multiplied directly i.e. there is an implicit mag 8 requirement, which could be worth adding verification for. In some group operations (broadly speaking those where special z values are in play, either gej_add_ge or coz) the input field values are normalized, usually to ensure that they are mag 6 or so (I am presupposing zero tests for h/i replacing u?/s? equality checks), but it varies. In practice the group ops don't produce higher than 6 mag anyway, so there's some waste here (and some performance-neutral jiggling might lower the output mags).

So I'm proposing to i) add pre/post verification for the current mag 8 assumptions, then ii) explore whether tightening the requirements can improve performance. At this stage, I'm only envisaging a simple global requirement, not op-specific bounds.

Thoughts?

sipa:
Fixed by #1348.

PR #1032, opened 2021-12-05
WIP Group verification

peterdettman:
Sets up pre- and post- method verification of _ge and _gej group elements. At the moment, this is concerned mainly with imposing a tighter limit (than the default) on the magnitudes of field elements x, y (,z).

Having guarantees about the magnitudes in input group elements can let us avoid some of the normalization calls needed at the start of several group addition methods, and perhaps e.g. use an alternative double algorithm. There may be a trade-off between the effort needed to get outputs to meet lower limits vs the benefits that provides to other methods.

### DIFF
--- a/src/field.h
+++ b/src/field.h
@@ -139,4 +139,11 @@ static void secp256k1_fe_half(secp256k1_fe *r);
  *  magnitude set to 'm' and is normalized if (and only if) 'm' is zero. */
 static void secp256k1_fe_get_bounds(secp256k1_fe *r, int m);
 
+#ifdef VERIFY
+
+/** Assert a context-specific limit m on the magnitude of a. m cannot be outside the intrinsic range of magnitudes supported by this field implementation. */
+static void secp256k1_fe_verify_magnitude(const secp256k1_fe *a, int m);
+
+#endif
+
 #endif /* SECP256K1_FIELD_H */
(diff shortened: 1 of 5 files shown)


peterdettman:
Removing _normalize_weak from several group add methods gives 2-3% speedup across major benchmarks (64 bit).

peterdettman:
Originally conceived many years ago now: https://github.com/bitcoin-core/secp256k1/issues/159 .

peterdettman:
Rebased and added some missing verify calls.

I've noted that there are several places where code directly manipulates the fields of group elements without calling a group method to do so. So the group structs are a bit too "open" at the moment. It should be possible to add suitable methods so that the group structs act more like abstract data types, and in particular so that we have a definite boundary at which to be able to place VERIFY calls in relation to group internals.

real-or-random:
@peterdettman This has "WIP" in the title but it looks pretty mature already. Can you comment on the status?

peterdettman:
@real-or-random See my previous comment; basically there are still quite a few unguarded local operations on group structs (i.e. not abstracted as group methods). These are not too difficult to track down comprehensively, but it occurs to me that, even once committed, we might need to allow some time for the abstraction to sink in to developers' minds before trying to exploit it (as per the "Save _normalize_weak..." commit) - there might be some backsliding. We could discuss ways of enforcing the abstraction in the language (or tooling), but the field implementations are in the same boat and just rely on "it being understood".

real-or-random:
Closing in favor of #1348

Issue #1001, opened 2021-10-28
Make fe magnitude implied statically


real-or-random:
See the discussion starting here: https://github.com/bitcoin-core/secp256k1/pull/943#issuecomment-944409385

peterdettman:
@roconnor-blockstream Surely not, e.g. what about _fe_cmov?

roconnor-blockstream:
I wasn't aware of that one.  (I'm less familiar with the non-verification related code.)

Make sense to patch that case by always setting the magnitude to the max of the two values?

peterdettman:
Yes, which was already done, then patched out, so this time with more comments I think!

real-or-random:
@peterdettman Can you elaborate a little bit more on the reasoning behind this suggestion?

I fully agree it's conceptually nicer. But as long as we don't have automated checks that prove that the magnitude is always low enough, does it really matter?

roconnor-blockstream:
The whole point of the magnitude field is to have a check that the tested code paths would not overflow even if different values were used.  It is a poor man's abstract interpretation.

But when we, say, set the magnitude of the output of cmov, based on the flag value, and the value of that flag value is depended on the specific initial value we are testing, then our "abstract interpretation" ends up being limited to only those value that would produce the same flag value at that call site.

I respect that our poor man's abstract interpretation is imperfect.  There are places were we explicitly branch on data (especially in the `_var` functions).  But we can at least strive to make our abstraction as abstract as reasonably possible.

(For illustration: the opposite course of action would be to make the magnitude based more and more on concrete values, but there is no point in doing that.  We already have a prefect measure of the a concrete magnitude: using the actual value itself!).

peterdettman:
I thought of the magnitude verification AS the automated checks. It's inherently worst-case, basically a static bounds check. That plus code coverage of all branches (and careful analysis of the very few loops) means there's no overflows in the field arithmetic. Without that property you need to be sure you have test cases that exercise every code path at the maximum possible magnitude, which is just a harder-to-prove/understand version of the same thing.

peterdettman:
> We already have a prefect measure of the a concrete magnitude: using the actual value itself!

Bingo.

real-or-random:
> I thought of the magnitude verification AS the automated checks. It's inherently worst-case, basically a static bounds check. 

Oh sure, I missed this part!

(My thinking was: it's only dynamic analysis anyway but if the values are statically determined, then that's indeed all we need...)

[...]

PR #1033, opened 2021-12-05
Add _fe_half and use in _gej_add_ge

peterdettman:
- Trades 1 _half for 3 _mul_int and 2 _normalize_weak

Gives around 2-3% faster signing and ECDH, depending on compiler/platform.


sipa:
ACK b6d109ddc80254e4ceedd7f22e6195a1ef37a087

I've written a basic test for the new function directly: https://github.com/sipa/secp256k1/commits/pr1033

peterdettman:
Thanks, @sipa . Merged your PR and added a benchmark entry also.

real-or-random (commented):
ACK mod nit. It may be a good idea to squash the last commit (update comments) into the first one.

real-or-random:
Looking at the comment in `gej_double`:
https://github.com/bitcoin-core/secp256k1/blob/master/src/group_impl.h#L274-L280

I wonder if `half` can save the normalization when switching to the other formula.

https://hyperelliptic.org/EFD/g1p/auto-shortw-jacobian-0.html#doubling-dbl-2009-l gives
```
A = X1^2
B = Y1^2
C = B^2
D = 2*((X1+B)^2-A-C)
E = 3*A
F = E^2
X3 = F-2*D
Y3 = E*(D-X3)-8*C
Z3 = 2*Y1*Z1
```

I think this is equivalent to

```
A = X1^2
B = Y1^2
C = B^2
D = (X1+B)^2-A-C
E = 3*(A/2)
F = E^2
X3 = F-D
Y3 = E*(D/2-X3)-C
Z3 = Y1*Z1
```

peterdettman:
> I wonder if `half` can save the normalization when switching to the other formula.

I might give it a try out of curiosity. It looks like the _half calls and extra _add calls can be paid for by removing several _mul_int calls, so maybe there's a net gain.

src/field.h
```diff
@@ -124,4 +124,6 @@ static void secp256k1_fe_storage_cmov(secp256k1_fe_storage *r, const secp256k1_f
 /** If flag is true, set *r equal to *a; otherwise leave it. Constant-time.  Both *r and *a must be initialized.*/
 static void secp256k1_fe_cmov(secp256k1_fe *r, const secp256k1_fe *a, int flag);
 
+static void secp256k1_fe_half(secp256k1_fe *r);
```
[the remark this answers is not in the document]
peterdettman: Done

peterdettman:
Squashed, added extra comments and extra VERIFY_CHECK that the low bit is zero before shifting it away.

real-or-random (approved):
ACK 0559a3d905ad9ec2e98faeb839e017349a18f8da

peterdettman:
My reasoning for the output magnitude in _fe_half. Please review, since the bound is exact (i.e. very tight).

Given the formula m_out = (m_in >> 1) + 1, we can just consider the worst case of an odd m_in. Also the top limb case will be the same as for the lower limbs, except with no incoming carry from above, and the lower limb(s) of P has a smaller value, so we deal only with a "middle limb" example where the added limb of P is maximal and a carry in from above is assumed.

Let odd m_in = 2.k + 1. Then the largest initial value for a lower limb per the magnitude constraints is (2.k + 1).(2.X), where X = 2^52-1 (resp. 2^26-1 in 32bit code). This value has potentially P added to it (worst case for an individual limb is +X), then is shifted down and then a carry is added. The carry will add 2^51 (resp. 2^25) == (X + 1)/2.

floor(((2.k + 1).(2.X) + X)/2) + (X + 1)/2

Since the carry is integral we can rearrange to this:

floor((4.k.X + 4.X + 1)/2)
= 2.k.X + 2.X
= 2.(k + 1).X

which is exactly the bound for the calculated output magnitude: floor((2.k + 1)/2 + 1 == k + 1

QED.

real-or-random:
Hm, it didn't occur to me that the analysis of the magnitude is that involved and I made a wrong assumption when reviewing this...

I believe your proof is correct but then we should maybe a polished version to a comment and introduce tests check the function with exercise the exact

peterdettman:
I agree, especially about wanting better tests. I have some here that use _fe_negate to get "OK" inputs, but ideally I'd like to be able to generate a field element with maximal limbs for a given magnitude, except that the LSB should be set so that the adding of P will be triggered.

Possibly we need some testing-oriented method(s) alongside _fe_verify to allow precise construction of otherwise maybe-unreachable representations. e.g a method to assert a new magnitude value would be helpful since we can generally get the limbs we want, but lose control of the magnitude. Then again, maybe it's easier to just directly manipulate the specific field representation(s) under SECP256K1_WIDEMUL_INT64/128 defines?

real-or-random:
I think you need to rebase (or force-push for some other reason) once here to make CI happy. A few tasks were failing due to #1047  being merged during the CI run of this PR. (Sorry for the CI mess again. Shouldn't happen for other PRs then at least...)

real-or-random:
> e.g a method to assert a new magnitude value would be helpful since we can generally get the limbs we want, but lose control of the magnitude. 

Sorry I can't follow here. When you "assert a new magnitude" value, wouldn't you have control then? 

Could we just add 2P (?) to increase the magnitude artificially for testing?

> Then again, maybe it's easier to just directly manipulate the specific field representation(s) under SECP256K1_WIDEMUL_INT64/128 defines?

You mean construct special values by by setting the fe members directly? Yeah, I think that's what we should do for testing.  go crypto has a similar thing, see this commit: https://github.com/golang/go/commit/d95ca9138026cbe40e0857d76a81a16d03230871

sipa:
I clearly also went too fast over the new bound.

Here is my reasoning for it. It applies to `x = r->n[i]` (for i=0..8) for the 32-bit code. Think of x as a real number.

Let `C = 2*0x3ffffff`.

On entrance to the function, we know that `x <= m*C`. In the first half of the function, at most `0x3ffffff` is added. Now `x <= (m+1/2)*C`. Then it is divided by two, leading to `x <= (m/2+1/4)*C`. Lastly, at most `1 << 25` is added (which is equal to `C/4 + 1/2`, giving `x <= (m/2+1/2)*C + 1/2`. This implies `x <= ceil(m/2 + 1/2)*C + 1/2`. And since `x`, `ceil(m/2 + 1/2)` and `C` are all integral, we can drop the `+ 1/2`, giving `x <= ceil(m/2 + 1/2)*C`, and `ceil(m/2 + 1/2)` equals the new magnitude `(m>>1)+1` for all integral m.

Let `D = 2*3fffff`.

For `x = r->n[8]` we instead have on entrance `x <= m*D`. In the first half of the function, at most `0x3fffff` is added. Now `x <= (m+1/2)*D`. Then it is divided by two, leading to `x <= (m/2+1/4)*D`. This implies `x <= ceil(m/2+1/4)*D`, and `ceil(m/2+1/4)` equals the new magnitude `(m>>1)+1` for all integral m.

Similar reasoning can be used for the 64-bit code.

It'd be good to document this reasoning in the code.

sipa:
@real-or-random By just following the "basic" doubling formula (https://en.wikipedia.org/wiki/Elliptic_curve_point_multiplication#Point_doubling), but using halving for the halving rather than bringing it to the Z coordinate, you get:

```
L = (3/2) * X1^2
S = Y1^2
T = X1*S
X3 = L^2 - 2*T
Y3 = L*(T - X3) - S^2
Z3 = Z1*Y1
```

which seems even simpler.

sipa:
Implemented here; seems to  work, but I can't really notice anything becoming faster. By operation count I would expect it to be slightly faster though: https://github.com/sipa/secp256k1/commits/202112_doublehalf. There may be some micro optimizations I've missed (e.g. I feel like it should be possible with one less negation), but it's not going to matter much.

peterdettman:
> Sorry I can't follow here. When you "assert a new magnitude" value, wouldn't you have control then?

Rephrased: we can generally create the limbs we want by simple addition through the _fe API, but not without the value of the magnitude variable going higher than we want. Therefore a new VERIFY-only method to let us just set the magnitude to what we want (this method would check the bounds of course) could be helpful.

However I think I'll start by using the direct-setting approach to get some worst-case tests in place

peterdettman:
> I feel like it should be possible with one less negation

`Y3 = -(S^2 + L*(X3 - T))`

Then only 2 negates needed, for T and Y3. Then you could choose to negate Z3 instead of Y3 (which is probably a speedup in itself).

By operations this should really be faster; I will test it shortly here. I wanted to note though that it also has a pleasant effect on the magnitudes of of the output field elements, which can be useful with https://github.com/bitcoin-core/secp256k1/pull/1032. With the input magnitudes of X1, Y1 constrained I think you could even negate them instead of T and Z3 and then the output magnitudes would get even smaller.

peterdettman:
> By operations this should really be faster; I will test it shortly here.

For me it seems an improvement, most noticeably in ecdh (as expected), which it makes around 1-2% faster.

peterdettman:
Cherry-picked new formula from @sipa and added a further refinement on top. Rebased to current master.

With new doubling formula ECDH is now around 4-5% faster (for the whole PR).

peterdettman:
Added specific test cases for maximal field elements (every limb at bound) and also worst-case field elements (subtract 1 from maximal low limb to ensure P is added and therefore carries all happen too).

peterdettman:
Added bounds analysis commentary to _fe_half methods and squashed into first commit.

sipa:
> Edit: Interestingly, moving the negate from Y3 to Z3 can't be done without changing some tests that appear to check for an exact expected z-ratio in some way that I haven't investigated.

Did you add a `secp256k1_fe_negate(rzr, &a->y)` in secp256k1_gej_double_var?

src/tests.c
```diff
@@ -47,6 +47,37 @@ static void uncounting_illegal_callback_fn(const char* str, void* data) {
     (*p)--;
 }
 
+void max_fe_for_magnitude(secp256k1_fe *r, int m) {
```
sipa, line 50: I think it would be cleaner to have this function be in the respective field_* files, and documented in field.h (it can still be surrounded by `#ifdef VERIFY` there).

peterdettman: Is tests.c only ever compiled with VERIFY defined?

sipa: It is right now, though in #1037 I suggested that maybe we also want tests that aren't. We can deal with that later, though.

real-or-random: I was asking myself the same question a few minutes ago. I think the answer is yes but we use  `#ifdef VERIFY` in a few other places (also introduced by you ;)), so it may be good to keep it like this. In fact I think we should run the tests also without `VERIFY`, e.g., to catch compiler bugs that don't appear when `VERIFY` is defined.

src/group_impl.h
```diff
@@ -326,8 +324,6 @@ static void secp256k1_gej_double_var(secp256k1_gej *r, const secp256k1_gej *a, s
 
     if (rzr != NULL) {
         *rzr = a->y;
-        secp256k1_fe_normalize_weak(rzr);
```
peterdettman, line 329: I haven't looked at the callers, but removing this normalize_weak is a little sketchy.

sipa: My belief is that this normalize_weak call only exists because of the multiply by 2 on the next line, which would otherwise operate on an input of unknown magnitude.

real-or-random: Maybe it's time to work on https://github.com/bitcoin-core/secp256k1/issues/1001 .

peterdettman:
> Did you add a `secp256k1_fe_negate(rzr, &a->y)` in secp256k1_gej_double_var?

Thanks, that was the problem. It didn't pan out faster though, despite my hope that scheduling a negate of Z3 along with other linear ops earlier in the method might be slightly faster than a final negate of Y3.

src/group_impl.h
peterdettman: I doubt there is a problem here, but performance impact is negligible and I don't want to be losing sleep, so I've restored the _normalize_weak. I think it can be cleaned up after/with #1001 and/or #1032.

src/tests.c
peterdettman: I've replaced the test method with a new field method _fe_get_bounds when VERIFY is defined. If the tests ever get run without VERIFY, we will presumably need a separate define to classify "test support code".

sipa:
The Sage group prover works with this new doubling formula:

```diff
--- a/sage/prove_group_implementations.sage
+++ b/sage/prove_group_implementations.sage
@@ -8,25 +8,20 @@ load("weierstrass_prover.sage")
 def formula_secp256k1_gej_double_var(a):
   """libsecp256k1's secp256k1_gej_double_var, used by various addition functions"""
   rz = a.Z * a.Y
-  rz = rz * 2
-  t1 = a.X^2
-  t1 = t1 * 3
-  t2 = t1^2
-  t3 = a.Y^2
-  t3 = t3 * 2
-  t4 = t3^2
-  t4 = t4 * 2
-  t3 = t3 * a.X
-  rx = t3
-  rx = rx * 4
-  rx = -rx
-  rx = rx + t2
-  t2 = -t2
-  t3 = t3 * 6
-  t3 = t3 + t2
-  ry = t1 * t3
-  t2 = -t4
-  ry = ry + t2
+  s = a.Y^2
+  l = a.X^2
+  l = l * 3
+  l = l / 2
+  t = -s
+  t = t * a.X
+  rx = l^2
+  rx = rx + t
+  rx = rx + t
+  s = s^2
+  t = t + rx
+  ry = t * l
+  ry = ry + s
+  ry = -ry
   return jacobianpoint(rx, ry, rz)
```

peterdettman:
Moved one of the _fe_negate calls in _gej_add_ge and updated the formula and some comments to properly reflect the current placement of negations. This was squashed into the first commit and the whole thing rebased.

peterdettman:
I've updated the sage formulae to reflect this PR's changes to _gej_double and _gej_add_ge. I've no more changes in mind.

real-or-random:
Please ignore the macOS CI failures. The Cirrus macOS VMs seems to have network issues... https://github.com/cirruslabs/osx-images/issues/42 macOS on CI is getting annoying on multiple fronts apparently.

sipa:
ACK b6631a809189fdbc9f31b710f995e7075160d536

Curious observation: in `secp256k1_gej_add_ge`, using `degenerate = secp256k1_fe_normalizes_to_zero(&m);` (dropping the `& secp256k1_fe_normalizes_to_zero(&rr)`) also works (and passes sage symbolic verification). I haven't though through why.

peterdettman:
@sipa An interesting observation. I have been playing around with reversing which of the two formulae is considered "primary" and which the "alternate", and presumably there is actually a set of cases where either formula works. I think there may be some small gains to be had here.

It might not be relevant, but do these sage scripts understand that the endomorphism exists so that (y1 == +/-y2) does not imply (x1 == x2)?

peterdettman:
> I have been playing around with reversing which of the two formulae is considered "primary" and which the "alternate"

Actually, on review, that's not quite what I was doing. Rather I had noted that in the degenerate case m and rr were 0 and so some of the cmov calls can actually be cadd or czero, which are marginally cheaper.

sipa:
> It might not be relevant, but do these sage scripts understand that the endomorphism exists so that (y1 == +/-y2) does not imply (x1 == x2)?

They reason over (fractions and ideals of) polynomials over Q. My intuition says that that will discover all properties that could hold in some field, but I don't know enough theory to know whether that's true. The script does find the error in the "old" gej_add_ge code, which was only wrong due to the existence of (x1,y) and (x2,-y) points, IIRC.

real-or-random:
needs (trivial) rebase :)

peterdettman:
Rebased.

jonasnick (commented):
This was fun to review. I get an 6% speedup in double and 2.5% in add.

I can't run the sage code. Tried to run it on two different implementations, but I'm getting a `no canonical coercion from Multivariate Polynomial Ring` error.

src/field_5x52_impl.h
```diff
@@ -56,6 +56,19 @@ static void secp256k1_fe_verify(const secp256k1_fe *a) {
     }
     VERIFY_CHECK(r == 1);
 }
+
+static void secp256k1_fe_get_bounds(secp256k1_fe *r, int m) {
```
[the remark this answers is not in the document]
peterdettman: I asked somewhere whether the tests are ever compiled without VERIFY and told they were not. Are you saying that --enable-coverage does do that?

src/field_10x26_impl.h
```diff
@@ -1133,6 +1151,80 @@ static SECP256K1_INLINE void secp256k1_fe_cmov(secp256k1_fe *r, const secp256k1_
...
+    secp256k1_fe_verify(r);
+    VERIFY_CHECK(r->magnitude < 32);
+#endif
+
+    /* Let m = r->magnitude
+     *     C = 0x3FFFFFFUL * 2
+     *     D = 0x03FFFFFUL * 2
+     *
+     * Initial bounds: t0..t8 <= C * m
+     *                     t9 <= D * m
```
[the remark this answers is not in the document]
peterdettman: Added a comment to that effect.

src/field_5x52_impl.h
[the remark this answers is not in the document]
jonasnick: Yes --enable-coverage builds the tests without VERIFY to get more accurate coverage statistics.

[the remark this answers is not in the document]
peterdettman: OK, I've made that change, but it would seem a good time to consider a `#define` that is set only when compiling tests.

jonasnick (commented):
ACK e848c3799c4f31367c3ed98d17e3b7de504d4c6e

Was able to run the updated sage scripts with #1068

real-or-random:
> Curious observation: in `secp256k1_gej_add_ge`, using `degenerate = secp256k1_fe_normalizes_to_zero(&m);` (dropping the `& secp256k1_fe_normalizes_to_zero(&rr)`) also works (and passes sage symbolic verification). I haven't thought through why.

I thought about this and I believe it works. The code currently switches to the alternative formula for lambda only if (R,M) = (0,0) but the alternative formula works whenever M = 0: Specifically, M = 0 implies y1 = -y2. If x1 = x2, then a = -b this is the r = infinity case that we handle separately. If x1 != x2, then the denominator in the alternative formula is non-zero, so this formula is well-defined. (And I understand this means that it gives the right result?) 

One needs to carefully check that the `infinity = ` assignment is still correct because now the definition of `m_alt` at this point in the code has changed. But this is true:

Case y1 = -y2 ==> degenerate = true ==> infinity = ((x1 -  x2)Z = 0) & ~a->infinity
a->infinity is handled separately. And if ~a->infinity, then Z = Z1 != 0, so infinity = (x1 - x2 = 0) = (a = -b) by case condition.

Case y1 != -y2 ==> degenerate = false ==> infinity = ((y1 + y2)Z = 0) & ~a->infinity.
a->infinity is handled separately. And if ~a->infinity, then Z = Z1 != 0, so infinity = (y1 + y2 = 0) = false by case condition.

real-or-random:
I pushed it here with a further change that saves the infinity variable: https://github.com/real-or-random/secp256k1/commits/202202-gej_add_ge  (Should not hold up this PR, I'm in the middle of reviewing it)

real-or-random (approved):
ACK e848c3799c4f31367c3ed98d17e3b7de504d4c6e

If you ever touch this again, maybe squash the commits that affect the doubling function. But no need to invalidate the ACKs for this. 

edit: Changed to PR title to reflect to change to _gej_double

sipa:
utACK e848c3799c4f31367c3ed98d17e3b7de504d4c6e

PR #1041, opened 2021-12-15
Follow-ups to making all tables fully static

sipa:
This PR implements a number of changes to follow up after merging #988:
* Naming consistency:
  * Renames all files related to the precomputated tables to have a consistent `precompute_ecmult` and `precompute_ecmult_gen` prefixes (avoiding the "gen" ambiguity).
  * Renames all precomputed files themselves to have consistent `precomputed_ecmult` and `precomputed_ecmult_gen` prefixes.
  * Merges the binaries precomputing the table files into one `precompute`.
* Make the tables for exhaustive tests be computed at runtime rather than compile time (this was already the case for ecmult_gen, but not ecmult). This is a preparation for the next commit, as the alternative would be to have separate precomputed libraries for the exhaustive tests and other binaries.
* Moves the actual tables to separate `precomputed_ecmult.c` and `precomputed_ecmult_gen.c` files, which are compiled only once as part of a new `libsecp256k1_precomputed.la`, included where relevant.


src/gen_ecmult_gen_static_prec_table.c
```diff
@@ -10,8 +10,8 @@
 #include "../include/secp256k1.h"
 #include "assumptions.h"
 #include "util.h"
-#include "group.h"
-#include "ecmult_gen.h"
```
real-or-random, line 14: nit: I think `ecmult_gen.h` is included for `ECMULT_GEN_PREC_G` and `ECMULT_GEN_PREC_N`.

Makefile.am
```diff
@@ -130,31 +130,24 @@ TESTS += exhaustive_tests
...
 # forced by deleting them, e.g., by invoking `make clean-precomp`.
-src/ecmult_static_pre_g.h:
-	$(MAKE) $(AM_MAKEFLAGS) gen_ecmult_static_pre_g$(EXEEXT)
-	./gen_ecmult_static_pre_g$(EXEEXT)
-src/ecmult_gen_static_prec_table.h:
-	$(MAKE) $(AM_MAKEFLAGS) gen_ecmult_gen_static_prec_table$(EXEEXT)
-	./gen_ecmult_gen_static_prec_table$(EXEEXT)
-
 PRECOMP = src/ecmult_gen_static_prec_table.h src/ecmult_static_pre_g.h
+$(PRECOMP)&:
```
real-or-random, line 156: is the `&` on purpose? I've never seen this syntax before.

real-or-random (commented):
That was quick1

Concept ACK

Makefile.am
sipa: Yes, it means: one command builds all these targets. Otherwise, if invoked with `-jN`, the submake commands gets invoked multiple times in parallel, which caused race condition failures on CI.

I guess I'll need another solution if mac doesn't support it.

real-or-random: Ah that explanation helped me search the web: https://stackoverflow.com/questions/63928066/what-is-the-canonical-way-to-simulate-grouped-targets-in-older-versions-of-make
And this confirms that it does not work on Mac Make.

real-or-random: I think you could make it a pattern rule, which have supported grouped targets for a long time, see last example here: https://www.gnu.org/software/make/manual/html_node/Pattern-Examples.html#Pattern-Examples

```
src/precomputed_%: 
	$(MAKE) $(AM_MAKEFLAGS) precompute$(EXEEXT)
	./precompute$(EXEEXT)
```

sipa: How does the syntax in your suggested code indicate grouping?

sipa: Oh, they always are. TIL

sipa: > Pattern rules may have more than one target; however, every target must contain a % character. Pattern rules are always treated as grouped targets (see Multiple Targets in a Rule) regardless of whether they use the : or &: separator. 

I interpret that as that you actually need multiple patterns on the left of the column. Otherwise rules like `%.o: %.c` would not do what we expect them to do.

real-or-random: Oh true... One could still abuse a pattern rule with % that essentially matches only once. But then not sure whether this is ugly than a stamp file.

sipa: Hmm, what about just

```make
precomp:
	$(MAKE) $(AM_MAKEFLAGS) precompute$(EXEEXT)
	./precompute$(EXEEXT)
```
?

sipa: Switched to a different approach, where `./precompute` is given the filename to generate, and it just generates that.

src/gen_ecmult_gen_static_prec_table.c
sipa: Fixed.

sipa:
Sigh, this isn't a solution either.

sipa:
Going to try a different approach.

jonasnick:
What was the problem with this approach? Just asking because I was in the middle of reviewing this.

sipa:
The problem was that if you build with "make -j", and it involves rebuilding the precomputed table files (4 in this PR as is), it'll trigger 4 invocations of a sub-make command to build the "precompute" binary, which end up overwriting each other, and race conditions.

The approach I'm going to try next is keeping precompute_ecmult and precompute_ecmult_gen separate, and only having them generate the precomputed_*.c file. The corresponding .h file can be a normal checked-in file in the repository.

jonasnick:
I see. Thanks.

PR #1042, opened 2021-12-17
Follow-ups to making all tables fully static

sipa:
This PR implements a number of changes to follow up after merging #988:

* Naming consistency:
  * All precomputed table files now have name `precomputed_*.*`
  * All source files related to the creation of the precomputed table files have name `precompute_*.*`.
  * All source files related to the computation of tables (whether they go in precomputed files or not) have name `*_compute_table.*`.
* Make the tables for exhaustive tests be computed at runtime rather than compile time (this was already the case for ecmult_gen, but not ecmult). This is a preparation for the next point, as the alternative would be to have separate precomputed libraries for the exhaustive tests and other binaries.
* Moves the actual tables to separate `precomputed_*.c` files, which are compiled only once as part of a new `libsecp256k1_precomputed.la`, included where relevant. The corresponding `precomputed_*.h` file are normal source files.

Retry of #1041.


real-or-random (commented):
ACK mod nits

real-or-random:
I think so far it was in theory possible to use libsecp by just including the `secp256k1.c` file, i.e., by including the source instead of linking binaries. I assume this is not a use case we care about (and I don't expect anyone to do this) but I would like to draw attention to this.

src/precompute_ecmult.c
```diff
@@ -20,39 +20,58 @@
...
+    for (j = 1; j < ECMULT_TABLE_SIZE(window_g); ++j) {
+        secp256k1_gej_set_ge(&gj, &ge);
+        secp256k1_gej_add_ge_var(&gj, &gj, &dgen, NULL);
+        secp256k1_ge_set_gej_var(&ge, &gj);
+        secp256k1_ge_to_storage(&table[j], &ge);
+    }
+}
+
+/* Like secp256k1_ecmult_compute_table, but one for both gen and gen*2^128. */
+static void secp256k1_ecmult_compute_two_tables(secp256k1_ge_storage* table, secp256k1_ge_storage* table_128, int window_g, const secp256k1_ge* gen) {
```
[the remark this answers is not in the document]
sipa: Done, added a commit to do so.

src/tests_exhaustive.c
```diff
@@ -389,8 +390,10 @@ int main(int argc, char** argv) {
         printf("running tests for core %lu (out of [0..%lu])\n", (unsigned long)this_core, (unsigned long)num_cores - 1);
     }
 
-    /* Recreate the ecmult_gen table using the right generator (as selected via EXHAUSTIVE_TEST_ORDER) */
+    /* Recreate the ecmult{,_gen} tables using the right generator (as selected via EXHAUSTIVE_TEST_ORDER) */
     secp256k1_ecmult_gen_create_prec_table(&secp256k1_ecmult_gen_prec_table[0][0], &secp256k1_ge_const_g, ECMULT_GEN_PREC_BITS);
+    secp256k1_ecmult_compute_two_tables(secp256k1_pre_g, secp256k1_pre_g_128, WINDOW_G, &secp256k1_ge_const_g);
+
```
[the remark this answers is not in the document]
sipa: The existing ecmult tables (from #956) contain separate tables for the exhaustive test case. So they're not using wrong tables.

The problem is that this approach isn't really compatible with the separately-compiled .c file with tables, unless there is a separately-compiled module for (each) exhaustive test generator. So instead I'm modifying the code to use the same approach for ecmult as ecmult_gen does since #988.

sipa:
@real-or-random Good to point that out, but I don't think it's really a concern. I think we should preserve (and improve) the ability to build without a complex build/config system around it, but whether that involves building one or two modules isn't the end of the world. If there is actual desire for that, it wouldn't be hard to make `src/precomputed_ecmult_{,gen_}.h` include the corresponding .c file.

src/ecmult_compute_table.h
```diff
@@ -0,0 +1,16 @@
+/*****************************************************************************************************
+ * Copyright (c) 2013, 2014, 2017, 2021 Pieter Wuille, Andrew Poelstra, Jonas Nick, Russell O'Connor *
+ * Distributed under the MIT software license, see the accompanying                                  *
+ * file COPYING or https://www.opensource.org/licenses/mit-license.php.                              *
+ *****************************************************************************************************/
+
+#ifndef SECP256K1_ECMULT_COMPUTE_TABLE_H
+#define SECP256K1_ECMULT_COMPUTE_TABLE_H
+
+/* Construct table of all odd multiples of gen in range 1..(2**(window_g-1)-1). */
```
jonasnick, line 10: `window_g-2`?

jonasnick (commented):
otherwise looks good

src/ecmult_compute_table.h
sipa: `ECMULT_TABLE_SIZE(w)` is `(1L << ((w)-2))`. But that's just the size of the table, not the range. As the entries are two apart (only odd values), I think the range extends to 2^(w-1)-1.

jonasnick (commented):
ACK e05da9e480de34129a170510a311abb204eefeb3

real-or-random:
ACK e05da9e480de34129a170510a311abb204eefeb3

hebasto:
It looks like this change [breaks](https://cirrus-ci.com/task/5951925467217920) building Bitcoin Core with MSVC:
```
...
     Creating library C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\build_msvc\x64\Release\bitcoind.lib and object C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\build_msvc\x64\Release\bitcoind.exp
libsecp256k1.lib(secp256k1.obj) : error LNK2019: unresolved external symbol secp256k1_pre_g referenced in function secp256k1_ecmult_strauss_wnaf [C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\build_msvc\bitcoind\bitcoind.vcxproj]
libsecp256k1.lib(secp256k1.obj) : error LNK2019: unresolved external symbol secp256k1_pre_g_128 referenced in function secp256k1_ecmult_strauss_wnaf [C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\build_msvc\bitcoind\bitcoind.vcxproj]
libsecp256k1.lib(secp256k1.obj) : error LNK2019: unresolved external symbol secp256k1_ecmult_gen_prec_table referenced in function secp256k1_ecmult_gen [C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\build_msvc\bitcoind\bitcoind.vcxproj]
C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\build_msvc\x64\Release\bitcoind.exe : fatal error LNK1120: 3 unresolved externals [C:\Users\ContainerAdministrator\AppData\Local\Temp\cirrus-ci-build\build_msvc\bitcoind\bitcoind.vcxproj]
```

Makefile.am
```diff
@@ -60,12 +64,17 @@ noinst_HEADERS += contrib/lax_der_parsing.c
 noinst_HEADERS += contrib/lax_der_privatekey_parsing.h
 noinst_HEADERS += contrib/lax_der_privatekey_parsing.c
 
+PRECOMPUTED_LIB = libsecp256k1_precomputed.la
+noinst_LTLIBRARIES = $(PRECOMPUTED_LIB)
+libsecp256k1_precomputed_la_SOURCES =  src/precomputed_ecmult.c src/precomputed_ecmult_gen.c
+libsecp256k1_precomputed_la_CPPFLAGS = $(SECP_INCLUDES)
```
real-or-random, line 70: @hebasto  This is because we now have a separate compilation unit for the precomputed tables. So Core needs to tell MSVC to build `libsecp256k1_precomputed` separately and link it to the library. 

But I have no idea about the required changes to https://github.com/bitcoin/bitcoin/blob/master/build_msvc/libsecp256k1/libsecp256k1.vcxproj that would be equivalent to this automake snippet.

If this is anyway maintained within the Bitcoin Core org, maybe we should move the vcxproj file to this repo in the future. This will help us add MSVC builds to the CI here.

hebasto: @sipsorcery

Could you look into this?

hebasto: @sipsorcery 

Thanks! It works flawlessly.

real-or-random: @sipsorcery That works but it's too much: You don't need the `precompute_` files, the `precomputed` files should be enough. 

Also, this will then recompile the `precomputed_` files every time, which makes recompilation somewhat slower than what we have in our Makefile. (This was the reason why we introduced the `libsecp256k1_precomputed` as a separate object.) But yeah, this only affects compilation time, so it's probably not worth to optimize the `vcxproj` for this.

hebasto:
@sipa 

During reviewing of the current [master branch](https://github.com/bitcoin-core/secp256k1/commit/427bc3cdcfbc74778070494daab1ae5108c71368), I've noticed that `extern "C" {...}` guarded by `#ifdef __cplusplus` were added to _non-public_ headers in the commit e05da9e480de34129a170510a311abb204eefeb3 "Fix c++ build".

But I cannot see any scenario where those headers, i.e., `src/precomputed_ecmult.h` and `src/precomputed_ecmult_gen.h`, are processed by a C++ compiler.

FWIW, no issues are observed when compiling Bitcoin Core after removal of `extern "C" {...}` from the mentioned headers.

Did I miss something?

real-or-random:
> But I cannot see any scenario where those headers, i.e., `src/precomputed_ecmult.h` and `src/precomputed_ecmult_gen.h`, are processed by a C++ compiler.

I haven't checked, but maybe those will would be relevant when compiling the `precomputed` compilation unit with a C compiler but the rest with a C++ compiler? But why should anyone do this? ping @sipa

PR #1044, opened 2021-12-18
Add another ecmult_multi test

sipa:
Add a large randomized test for ecmult_multi_* functions, which aims to test:
* Small and large numbers of (scalar,point) input pairs (from 0 to 128, vaguely exponentially distributed).
* Small and large numbers of (scalar,point) input pairs that individually multiply to zero (so scalar=0 or point=infinity), as those are handled specially in wnaf loops.
* With final output equal to infinity or not (roughly 50% chance).
* With the factor for G equal to 0 or not (roughly 50% chance).
* All single-batch algorithms (ecmult_multi_var, as well as ecmult_strauss_single_batch and ecmult_pippenger_single_batch).

I wrote this mostly to gain confidence in the handling of (0,p) or (s,infinity) inputs being changed by #899.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -4052,6 +4052,174 @@ void test_ecmult_multi(secp256k1_scratch *scratch, secp256k1_ecmult_multi_func e
     }
 }
 
+int test_ecmult_multi_random(secp256k1_scratch *scratch) {
+    /* Large random test for ecmult_multi_* functions which exercises:
+     * - Few or many inputs (0 up to 128, roughly exponentially distributed).
+     * - Few or many 0*P or a*INF inputs (roughly uniformly distributed).
+     * - Including or excluding an nonzero a*G term (or such a term at all).
+     * - Final expected result equal to infinity or not (roughly 50%).
+     * - ecmult_multi_var, ecmult_strauss_single_batch, ecmult_pippenger_single_batch
+     */
+
+    /* These 4 variables define the eventual input to the ecmult_multi function.
+     * g_scalar is the G scalar fed to it (or NULL, possibly, if g_scalar=0), and
+     * scalars[0..filled-1] and gejs[0..filled-1] are the scalars and points
+     * which form its normal inputs. */
+    int filled = 0;
+    secp256k1_scalar g_scalar = SECP256K1_SCALAR_CONST(0, 0, 0, 0, 0, 0, 0, 0);
+    secp256k1_scalar scalars[128];
+    secp256k1_gej gejs[128];
+    /* The expected result, and the computed result. */
+    secp256k1_gej expected, computed;
+    /* Temporaries. */
+    secp256k1_scalar sc_tmp;
+    secp256k1_ge ge_tmp;
+    /* Variables needed for the actual input to ecmult_multi. */
+    secp256k1_ge ges[128];
+    ecmult_multi_data data;
+
+    int i;
+    /* Which multiplication function to use */
+    int fn = secp256k1_testrand_int(3);
+    secp256k1_ecmult_multi_func ecmult_multi = fn == 0 ? secp256k1_ecmult_multi_var :
+                                               fn == 1 ? secp256k1_ecmult_strauss_batch_single :
+                                               secp256k1_ecmult_pippenger_batch_single;
+    /* Simulate exponentially distributed num. */
+    int num_bits = 2 + secp256k1_testrand_int(6);
+    /* Number of (scalar, point) inputs (excluding g). */
+    int num = secp256k1_testrand_int((1 << num_bits) + 1);
+    /* Number of those which are nonzero. */
+    int num_nonzero = secp256k1_testrand_int(num + 1);
+    /* Whether we're aiming to create an input with nonzero expected result. */
+    int nonzero_result = secp256k1_testrand_bits(1);
+    /* Whether we will provide nonzero g multiplicand. In some cases our hand
+     * is forced here based on num_nonzero and nonzero_result. */
+    int g_nonzero = num_nonzero == 0 ? nonzero_result :
+                    num_nonzero == 1 && !nonzero_result ? 1 :
+                    (int)secp256k1_testrand_bits(1);
+    /* Which g_scalar pointer to pass into ecmult_multi(). */
+    const secp256k1_scalar* g_scalar_ptr = (g_nonzero || secp256k1_testrand_bits(1)) ? &g_scalar : NULL;
+    /* How many EC multiplications were performed in this function. */
+    int mults = 0;
+    /* How many randomization steps to apply to the input list. */
+    int rands = (int)secp256k1_testrand_bits(3);
+    if (rands > num_nonzero) rands = num_nonzero;
+
+    secp256k1_gej_set_infinity(&expected);
+    secp256k1_gej_set_infinity(&gejs[0]);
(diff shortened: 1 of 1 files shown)


src/tests.c
```diff
@@ -4052,6 +4052,145 @@ void test_ecmult_multi(secp256k1_scratch *scratch, secp256k1_ecmult_multi_func e
...
+        secp256k1_gej_set_ge(&gejs[filled], &ge_tmp);
+        ++filled;
+    }
+    if (result_nonzero) {
+        /* Compute the expected result. */
+        CHECK(filled <= 1);
+        secp256k1_ecmult(&result, &gejs[0], &scalars[0], &g_scalar);
+        mults += filled + g_nonzero;
+    }
+    /* At this point we have (filled <= 2) and result equals the sum of all (scalars,gejs) pairs. */
```
jonasnick, line 4124: Is `(g_scalar, g)` part of "all pairs"? If no, it should be mentioned here because it's included in the result. If yes, then the comment above `filled = 0` should mention "excluding g".

src/tests.c
```diff
@@ -4052,6 +4052,145 @@ void test_ecmult_multi(secp256k1_scratch *scratch, secp256k1_ecmult_multi_func e
...
+    int result_nonzero = secp256k1_testrand_bits(1);
+    /* Whether we will provide nonzero g multiplicand. */
+    int g_nonzero = num_nonzero == 0 ? result_nonzero :
+                    num_nonzero == 1 && !result_nonzero ? 1 :
+                    (int)secp256k1_testrand_bits(1);
+    /* Which g_scalar pointer to pass into ecmult_multi(). */
+    const secp256k1_scalar* g_scalar_ptr = (g_nonzero || secp256k1_testrand_bits(1)) ? &g_scalar : NULL;
+    /* How many scalars/gej pairs have been filled. */
+    int filled = 0;
+    /* How many EC multiplications were performed in this function. */
```
jonasnick, line 4085: "excluding ecmult_multi" if this is intentional. But it seems like it would be preferable to include `ecmult_multi` (the nonzero pairs) because `multi` appears to be a proxy for time and most of the multiplications will happen inside `ecmult_multi`. Probably worth to add a note on what `mults` is for exactly.

jonasnick (commented):
I can't judge if this helps with coverage of corner cases in #899. But I'll take a look at that PR. 

This new test adds about 16% (7.5 seconds) to my `./tests`.

src/tests.c
sipa: Yes, it is. I've improved the comments.

src/tests.c
sipa: Nice catch. That wasn't intentional (and also the reason why this test added so much runtime). I've fixed it.

sipa:
I've improved comments, fixed the ecmult count, and now benchmark the added cost of this test at 2% of the total runtime (at iters=64; at iters=1 it should be much lower), running ~1400 instances of it.

sipa:
@jonasnick My goal isn't increasing test coverage as such, but just having something that tests many scenarios that involve `0*P` or `a*INF` terms in the beginning/end/middle of the list of terms. Just because all lines are hit doesn't mean those line are correct.

real-or-random:
This valgrind failure on CI is not a timeout: 
```
==2989== Conditional jump or move depends on uninitialised value(s)
==2989==    at 0x468DCD: secp256k1_ecmult_strauss_wnaf (ecmult_impl.h:235)
==2989==    by 0x43C309: secp256k1_ecmult (ecmult_impl.h:348)
==2989==    by 0x43C309: test_ecmult_multi_random (tests.c:4137)
==2989==    by 0x43E7F7: run_ecmult_multi_tests (tests.c:4424)
==2989==    by 0x465503: main (tests.c:6818)
```

sipa:
@real-or-random Fixed.

src/tests.c
```diff
@@ -4052,6 +4052,175 @@ void test_ecmult_multi(secp256k1_scratch *scratch, secp256k1_ecmult_multi_func e
...
+        CHECK(filled <= 1);
+        secp256k1_ecmult(&expected, &gejs[0], &scalars[0], &g_scalar);
+        mults += filled + g_nonzero;
+    }
+
+    /* At this point we have expected = scalar_g*G + sum(scalars[i]*gejs[i] for i=0..filled-1). */
+    CHECK(filled <= 1 + !nonzero_result);
+    CHECK(filled <= num_nonzero);
+
+    /* Add entries to sclalars,gejs so that there are num of them. All the added entries
```
jonasnick, line 4147:
```suggestion
    /* Add entries to scalars,gejs so that there are num of them. All the added entries
```

src/tests.c
```diff
@@ -4052,6 +4052,175 @@ void test_ecmult_multi(secp256k1_scratch *scratch, secp256k1_ecmult_multi_func e
...
+
+    secp256k1_gej_set_infinity(&expected);
+    secp256k1_gej_set_infinity(&gejs[0]);
+    secp256k1_scalar_set_int(&scalars[0], 0);
+
+    if (g_nonzero) {
+        /* If g_nonzero, set g_scalar to nonzero value r. */
+        random_scalar_order_test(&g_scalar);
+        if (!nonzero_result) {
+            /* And if 0 expected is desired, add a (a*r, -(1/a)*g) term to compensate. */
```
jonasnick, line 4116:
```suggestion
            /* And if expected=0 is desired, add a (a*r, -(1/a)*g) term to compensate. */
```

src/tests.c
```diff
@@ -4052,6 +4052,175 @@ void test_ecmult_multi(secp256k1_scratch *scratch, secp256k1_ecmult_multi_func e
...
+        /* Perturb all consecutive pairs of inputs:
+         * a*P + b*Q -> (a+b)*P + b*(Q-P). */
+        for (j = 0; j + 1 < num_nonzero; j += 2) {
+            secp256k1_gej gej;
+            secp256k1_scalar_add(&scalars[j], &scalars[j], &scalars[j+1]);
+            secp256k1_gej_neg(&gej, &gejs[j]);
+            secp256k1_gej_add_var(&gejs[j+1], &gejs[j+1], &gej, NULL);
+        }
+        /* Transform the last input: a*P -> (v*a) * ((1/v)*P). */
+        if (num_nonzero >= 1) {
```
jonasnick, line 4187: That condition seems to be true always.

jonasnick (commented):
@sipa Sure, I didn't mean to refer to metrics like line or branch coverage. I have looked at PR #899 enough now to see why this added test is useful. It exercises conditions that can uncover potential bugs in the rewritten loop of commit "Eliminate input_pos state field from ecmult_strauss_wnaf.". I think both the commit and this PR are worth having.

Can confirm the 2% test time increase. 7%  with 256 iters.

ACK mod nits

src/tests.c
sipa: Done.

src/tests.c
sipa: Done.

src/tests.c
sipa: Changed to an assertion.

sipa:
Idea for a follow-up: when testing ecmult_mult_var, also randomly choose the amount of scratch space, so it'll exercise splitting into batches.

jonasnick (commented):
ACK 22d25c8e0ab1d24f0f4a80fe016cbd71cd889866

PR #1046, opened 2021-12-20
ci: Optimize build matrix

real-or-random:
* Removes STATICPRECOMPUTATION from CI (left-over from #988) 
* Optimizes the Linux build matrix (more coverage with fewer jobs)
* Trims down the MacOS build matrix: MacOS jobs take roughly 10x time as compared Linux jobs, so I think we should run only some basic tests. And anyway, they target gcc and clang for x86_64, which is covered by the Linux jobs already. (And they currently fail due to some homebrew server being down but that's a different issue...)

I'm working on a further PR that adds MSVCs build on native Windows. This will give us a real new dimension in the configuration space.  And there's still the idea to build with multiple versions of gcc and clang.


real-or-random:
> MacOS jobs take roughly 10x time as compared Linux jobs, so I think we should run only some basic tests. And anyway, they target gcc and clang for x86_64, which is covered by the Linux jobs already. (And they currently fail due to some homebrew server being down but that's a different issue...)

Oh I think the reason is that we never update brew (and the catalina image provided by Cirrus is old and deprecated). Since the brew binary URL has moved, it always needs to fall back to building deps (autotools) from source, which currently fails because gnu.org is down...

Let's see if a newer brew fixes this. Then I may reconsider the commit that trims the macOS matrix.

We're stuck to old macOS Catalina because valgrind doesn't work on the latest versions and even the unoffical fork we use does only support Catalina (https://github.com/LouisBrunner/valgrind-macos/issues/19). Do you think we should stick to Catalina or move on (and drop Valgrind builds on macOS)?

real-or-random:
This is WIP.

I unmarked and marked again as draft trigger CI (which does not run on draft PRs...)

real-or-random:
Opened https://github.com/cirruslabs/osx-images/issues/41 to avoid adding `brew update` to every macOS task, which takes a few minutes.

PR #1047, opened 2021-12-21
ci: Various improvements

real-or-random:
This is a replacement for #1046. It contains only the changes that (I think) are unlikely to be debated and don't interfere with #995.  I'd like to get this in quickly to get a green badge again. (Using 2 CPUs instead of one for valgrind should do the trick.)

I'll open another PR on top of this to optimize the matrix. This will deserve more discussion and we should get #995 merged first.


real-or-random:
For more background on the "greedy" flag, see https://github.com/bitcoin/bitcoin/pull/23797.

.cirrus.yml
```diff
@@ -4,10 +4,10 @@ env:
   # Specific warnings can be disabled with -Wno-error=foo.
   # -pedantic-errors is not equivalent to -Werror=pedantic and thus not implied by -Werror according to the GCC manual.
   WERROR_CFLAGS: -Werror -pedantic-errors
-  MAKEFLAGS: -j2
+  MAKEFLAGS: -j4
```
elichai, line 7: Does the CI server actually have 4 CPUs?

.cirrus.yml
```diff
@@ -84,10 +88,7 @@ task:
 
 task:
   name: "i686: Linux (Debian stable)"
-  container:
-    dockerfile: ci/linux-debian.Dockerfile
-    cpu: 1
-    memory: 1G
+  << : *LINUX_CONTAINER
```
elichai, line 91: Is this a yaml trick?

.cirrus.yml
```diff
@@ -134,8 +135,9 @@ task:
   ##   - rm /tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress
   ##
   brew_valgrind_pre_script:
+    - brew update
     - brew config
-    - brew tap --shallow LouisBrunner/valgrind
+    - brew tap LouisBrunner/valgrind
```
elichai, line 140: Why not shallow?

elichai (commented):
Manually checked that the `ECMULT_TABLE_SIZE` changes are correct according to `(1 << (w-2))`

.cirrus.yml
real-or-random: Yes, we use this in other places, too.

.cirrus.yml
real-or-random: Shallow doesn't exist anymore in recent brew versions, see the commit message.

.cirrus.yml
real-or-random: We set `cpu: 1` or `cpu: 2` depending on the task. But the `greedy` flag will occasionally give us more CPUs when available. Increasing this to `-j4` makes sure we actually can make use of this.

.cirrus.yml
elichai: Found it: https://yaml.org/type/merge.html
This is interesting, I did not know yaml has this sort of things

real-or-random: Yeah, core uses it all over the place, so we're in good company.

elichai:
utACK 965f9480c417254318a3093936ee9748391b49b1
The C file changes are simple and correct, and it's great to test different window sizes

.cirrus.yml
```diff
@@ -66,12 +71,11 @@ task:
     - env: {WIDEMUL: int128,                 ECDH: yes, EXPERIMENTAL: yes, SCHNORRSIG: yes}
     - env: {WIDEMUL: int128,  ASM: x86_64}
     - env: {                  RECOVERY: yes,            EXPERIMENTAL: yes, SCHNORRSIG: yes}
-    - env: {                  STATICPRECOMPUTATION: no}
     - env: {BUILD: distcheck, WITH_VALGRIND: no, CTIMETEST: no, BENCH: no}
     - env: {CPPFLAGS: -DDETERMINISTIC}
     - env: {CFLAGS: -O0, CTIMETEST: no}
-    - env: { ECMULTGENPRECISION: 2 }
-    - env: { ECMULTGENPRECISION: 8 }
+    - env: { ECMULTGENPRECISION: 2, ECMULTWINDOW: 4 }
```
jonasnick, line 77: Better to check the edges? so `ECMULTWINDOW: 2`?

real-or-random: Indeed, let me just change to checking 2 (as edge case) and 4, which is a more reasonable value because it still needs almost no memory and may be a candidate for a "lowmem" setting here we may introduce in the future. (Rust bindings do the same, see https://github.com/rust-bitcoin/rust-secp256k1/blob/master/secp256k1-sys/build.rs#L45)

real-or-random: fixed with force-push

jonasnick (commented):
ACK b4ac1a1d5f4d51b9836ac310b78bc9d4256580c2

elichai:
utACK b4ac1a1d5f4d51b9836ac310b78bc9d4256580c2

sipa:
Posthumous ACK b4ac1a1d5f4d51b9836ac310b78bc9d4256580c2

src/precompute_ecmult.c
```diff
@@ -57,6 +57,8 @@ static void print_two_tables(FILE *fp, int window_g) {
 }
 
 int main(void) {
+    /* Always compute all tables for window sizes up to 15. */
+    int window_g = (ECMULT_WINDOW_SIZE < 15) ? 15 : ECMULT_WINDOW_SIZE;
```
binharebs, line 61: [15_]()

binharebs: - - [ ] **@[__]()**

src/precompute_ecmult.c
```diff
@@ -77,15 +79,15 @@ int main(void) {
     fprintf(fp, "#include \"ecmult.h\"\n");
     fprintf(fp, "#include \"precomputed_ecmult.h\"\n");
     fprintf(fp, "#define S(a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p) SECP256K1_GE_STORAGE_CONST(0x##a##u,0x##b##u,0x##c##u,0x##d##u,0x##e##u,0x##f##u,0x##g##u,0x##h##u,0x##i##u,0x##j##u,0x##k##u,0x##l##u,0x##m##u,0x##n##u,0x##o##u,0x##p##u)\n");
```
binharebs, line 81: ****

src/precompute_ecmult.c
```diff
@@ -77,15 +79,15 @@ int main(void) {
...
-    fprintf(fp, "#if ECMULT_TABLE_SIZE(ECMULT_WINDOW_SIZE) > %ld\n", ECMULT_TABLE_SIZE(ECMULT_WINDOW_SIZE));
```
binharebs, line 80: ****

src/precompute_ecmult.c
```diff
@@ -77,15 +79,15 @@ int main(void) {
...
+    fprintf(fp, "#if ECMULT_WINDOW_SIZE > %d\n", window_g);
```
binharebs, line 82: __

src/precompute_ecmult.c
```diff
@@ -77,15 +79,15 @@ int main(void) {
...
     fprintf(fp, "   #error configuration mismatch, invalid ECMULT_WINDOW_SIZE. Try deleting precomputed_ecmult.c before the build.\n");
     fprintf(fp, "#endif\n");
     fprintf(fp, "#ifdef EXHAUSTIVE_TEST_ORDER\n");
     fprintf(fp, "#    error Cannot compile precomputed_ecmult.c in exhaustive test mode\n");
     fprintf(fp, "#endif /* EXHAUSTIVE_TEST_ORDER */\n");
     fprintf(fp, "#define WINDOW_G ECMULT_WINDOW_SIZE\n");
 
-    print_two_tables(fp, ECMULT_WINDOW_SIZE);
```
binharebs, line 88: - [ ]

binharebs, line 88: 00

binharebs: @

binharebs (commented):
[995]()

src/precompute_ecmult.c
```diff
@@ -77,15 +79,15 @@ int main(void) {
     fprintf(fp, "#include \"ecmult.h\"\n");
```
roconnor-blockstream, line 79: `ecmult.h` is no longer needed if you remove `ECMULT_TABLE_SIZE`.

real-or-random: I think it's correct to use `ECMULT_TABLE_SIZE`.

roconnor-blockstream: Right sorry.  For some reason I thought all the users of `ECMULT_TABLE_SIZE` were eliminated, but that is not true.

PR #1048, opened 2021-12-21
ci: Optimize build matrix

real-or-random:
We should try to merge #995 before this PR to avoid conflicts.

This tries to get more coverage of "useful" configurations with fewer tasks. I think what I suggest here is good but on the other hand there's no real need to reduce the number of tasks right now, so please suggest additional configs if you think they're worth testing. Maybe we could add a `-UVERIFY` build? (Even nicer would be to run with and without VERIFY on the same seed and compare the results -- but that needs more work.)

This now always runs `make check` and then `make distcheck` but ensures that the latter skips the tests already run during `make check`.

I refrained from reducing the number of tasks on macOS further. Testing clang on macOS is useful because it's "Apple clang" and testing gcc is useful because it's gcc 9 (as opposed to gcc 10 on Linux).  Moreover, the changes in #1047 make macOS tasks quicker again. The annoying thing about these macOS tasks is not that they run too long, but just that there is more demand/scheduling pressure for macOS machines on Cirrus, so they often sit there for hours scheduled and waiting to be run. But there's nothing we can do about this.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -349,7 +349,7 @@ task:
   << : *CAT_LOGS
 
 task:
-  name: "C++ -fpermissive"
+  name: "C++ -fpermissive (entire project)"
   << : *LINUX_CONTAINER
   env:
     CC: g++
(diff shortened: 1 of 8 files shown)


real-or-random:
Hm, my plan was to always run `make distcheck` but I need to rethink that change. So `make distcheck` includes `make check` but does in a clean tree which is removed afterwards (including the test logs... -.-), so maybe it's not a good idea to always run it. On the other hand, this means we have this bug with test files being removed already now with the single distcheck task on CI: https://cirrus-ci.com/task/4941156000727040.

Maybe we should stick to the idea of always running distcheck and rescue the test logs. Or just introduce a separate distcheck build.

real-or-random:
This now runs `make check` and then `make distcheck` but ensures that the latter skips the tests already run during `make check`.  Rebased and ready for review.

real-or-random:
Yeah, this is obsolete. It still makes sense to take a look at the build matrix and see if we're happy with it. Let me track this in #1392.

PR #1052, opened 2021-12-23
Use xoshiro256++ instead of RFC6979 for tests

sipa:
Just some easy low-hanging fruit. It's complete overkill to use the RFC6979 RNG for our test randomness. Replace it with a modern non-cryptographic RNG with good properties. It's a few % speedup for me.

Given the internal naming of all these functions to be "testrand", I'm not concerned about the risk of someone using this for something that needs actual cryptographic randomness.

### DIFF
--- a/src/testrand_impl.h
+++ b/src/testrand_impl.h
@@ -109,7 +109,7 @@ static void secp256k1_testrand256_test(unsigned char *b32) {
 }
 
 static void secp256k1_testrand_flip(unsigned char *b, size_t len) {
-    b[secp256k1_testrand_int(len)] ^= (1 << secp256k1_testrand_int(8));
+    b[secp256k1_testrand_int(len)] ^= (1 << secp256k1_testrand_bits(3));
 }
 
 static void secp256k1_testrand_init(const char* hexseed) {
--- a/src/testrand_impl.h
+++ b/src/testrand_impl.h
@@ -14,35 +14,62 @@
 #include "testrand.h"
 #include "hash.h"
 
-static secp256k1_rfc6979_hmac_sha256 secp256k1_test_rng;
-static uint32_t secp256k1_test_rng_precomputed[8];
-static int secp256k1_test_rng_precomputed_used = 8;
+static uint64_t secp256k1_test_state[4];
 static uint64_t secp256k1_test_rng_integer;
 static int secp256k1_test_rng_integer_bits_left = 0;
 
 SECP256K1_INLINE static void secp256k1_testrand_seed(const unsigned char *seed16) {
-    secp256k1_rfc6979_hmac_sha256_initialize(&secp256k1_test_rng, seed16, 16);
-}
+    static const unsigned char PREFIX[19] = "secp256k1 test init";
+    unsigned char out32[32];
+    secp256k1_sha256 hash;
+    int i;
 
-SECP256K1_INLINE static uint32_t secp256k1_testrand32(void) {
-    if (secp256k1_test_rng_precomputed_used == 8) {
-        secp256k1_rfc6979_hmac_sha256_generate(&secp256k1_test_rng, (unsigned char*)(&secp256k1_test_rng_precomputed[0]), sizeof(secp256k1_test_rng_precomputed));
-        secp256k1_test_rng_precomputed_used = 0;
+    /* Use SHA256(PREFIX || seed16) as initial state. */
+    secp256k1_sha256_initialize(&hash);
+    secp256k1_sha256_write(&hash, PREFIX, sizeof(PREFIX));
+    secp256k1_sha256_write(&hash, seed16, 16);
+    secp256k1_sha256_finalize(&hash, out32);
+    for (i = 0; i < 4; ++i) {
+        uint64_t s = 0;
+        int j;
+        for (j = 0; j < 8; ++j) s = (s << 8) | out32[8*i + j];
+        secp256k1_test_state[i] = s;
     }
-    return secp256k1_test_rng_precomputed[secp256k1_test_rng_precomputed_used++];
+    secp256k1_test_rng_integer_bits_left = 0;
 }
 
-static uint32_t secp256k1_testrand_bits(int bits) {
-    uint32_t ret;
+SECP256K1_INLINE static uint64_t rotl(const uint64_t x, int k) {
+    return (x << k) | (x >> (64 - k));
+}
+
+SECP256K1_INLINE static uint64_t secp256k1_testrand64(void) {
+    /* Test-only Xoshiro256++ RNG. See https://prng.di.unimi.it/ */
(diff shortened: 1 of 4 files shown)


src/testrand_impl.h
```diff
@@ -14,22 +14,44 @@
...
+static uint64_t secp256k1_test_rng_key1, secp256k1_test_rng_key2;
+static uint64_t secp256k1_test_rng_cnt = 1;
 static uint64_t secp256k1_test_rng_integer;
 static int secp256k1_test_rng_integer_bits_left = 0;
 
-SECP256K1_INLINE static void secp256k1_testrand_seed(const unsigned char *seed16) {
-    secp256k1_rfc6979_hmac_sha256_initialize(&secp256k1_test_rng, seed16, 16);
+SECP256K1_INLINE static uint32_t secp256k1_testrand32(void) {
+    /* RNG based on https://arxiv.org/abs/2004.06278, using two separate keys, and
+     * only using odd counters to avoid entropy loss of key1. */
```
real-or-random, line 24: It may be worth stressing in the comment that it's for tests only, for someone who runs `grep RNG`...

elichai:
I'm curious why you choose this specific Rng? 
How does it compare to something like `XoShiRo256++`?

robot-dreams (commented):
Concept ACK

- What's meant by "avoid entropy loss of key1"?
- Section 4 of the linked paper says they ran tests with increments other than one. However, they didn't mention anything about using two different keys: how do you know, roughly, that it doesn't mess up any statistical properties?

sipa:
I just tried to insert additional entropy in the RNG because we already have a 128-bit seed. Perhaps that's overkill though, and I should use the paper's RNG unmodified.

The reasoning is this: the existing function uses the same key variable twice. If I replace the second use with a different key, it shouldn't worsen anything. However, if the counter is even now then `key1 * cnt` loses (at least) one bit of entropy (multiply by even number modulo a power of 2 is not reversible), so to counter that, only use odd counters. Neither of those changes should meaningfully impact runtime.

To be clear: my belief is that with two keys, but without only-odd counters, the result is still at least as good as the one-key version. But with only-odd counters it's somewhat better still.

robot-dreams:
Thanks for clarifying!

I have a slight preference for using the paper's RNG unmodified (possibly with odd counters, which is compatible with the paper), but I don't have a stronger reason than "on principle you shouldn't modify it" 🙃 so feel free to decide either way.

robot-dreams:
First commit looks good. Nits on second commit:
- Should commit message say `secp256k1_testrand_int(2**N) -> secp256k1_testrand_bits(N)` instead?
- It looks like `testrand_flip` still calls `testrand_int(8)`

Wait, just double checking, how much speedup did you see? Here's what I got (`%` = M1 Mac, `#` = Linux VM):

### At bcfbebe001404ce3eeb3bf3498caa5b2a6c28795 (new):

```
% time ./tests
./tests  28.53s user 3.27s system 99% cpu 31.878 total

# time ./tests
real	1m8.177s user	1m8.050s sys	0m0.104s

# time valgrind ./tests 2
real	11m54.970s user	10m58.498s sys	0m56.142s
```

### At 09971a3ffd053816286bcffcb2f15a10f7c3c098 (old):

```
% time ./tests
./tests  30.23s user 3.26s system 98% cpu 33.852 total

# time ./tests
real	1m7.266s user	1m7.136s sys	0m0.112s

# time valgrind ./tests 2
real	11m49.739s user	10m53.123s sys	0m56.321s
```

sipa:
Made a few changes:
* Added a separate commit that does testrand_int(2^N) -> testrand_bits(N) replacements
* Added a "Test-only" comment to the testrand32 implementation (the testrand.h file already mentioned this FWIW).
* Switched to "pure" Squares rand, with 64-bit seed. Also dropped the (cnt += 2), because the one-key version has no entropy loss with even cnt.

@elichai No good reason. I just googled for fast PRNGs. It has good randomness properties, and the counter-based design clearly has a 2^64 cycle, which should be enough.

@robot-dreams

On my Ryzen 5950X system, GCC 11.2.0, default config, `./tests`:
* Old: 0m36.216s
* New: 0m34.689s

src/testrand_impl.h
```diff
@@ -109,44 +121,45 @@ static void secp256k1_testrand256_test(unsigned char *b32) {
...
-            seed16[0] ^= t;
-            seed16[1] ^= t >> 8;
-            seed16[2] ^= t >> 16;
-            seed16[3] ^= t >> 24;
-            seed16[4] ^= t >> 32;
-            seed16[5] ^= t >> 40;
-            seed16[6] ^= t >> 48;
-            seed16[7] ^= t >> 56;
+        if (frand == NULL || fread(&seed8, 1, sizeof(seed8), frand) != sizeof(seed8)) {
+            fprintf(stderr, "WARNING: could not read 8 bytes from /dev/urandom; falling back to insecure PRNG\n");
```
robot-dreams, line 145: Nit: "falling back to time-based seed" since it's always an insecure PRNG?

robot-dreams (commented):
ACK 6ee0f009ae4026aabf7bde654c1a18f940f9b4ea

sipa:
Changed to using xoshiro256++. It's even (slightly) faster now, in my benchmarks.

Sorry for the many changes, I suspect I'm done now.

src/testrand_impl.h
```diff
@@ -109,7 +139,7 @@ static void secp256k1_testrand256_test(unsigned char *b32) {
 }
 
 static void secp256k1_testrand_flip(unsigned char *b, size_t len) {
-    b[secp256k1_testrand_int(len)] ^= (1 << secp256k1_testrand_int(8));
+    b[secp256k1_testrand_int(len)] ^= (1 << secp256k1_testrand_bits(3));
```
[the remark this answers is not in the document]
sipa: Done (in a separate commit).

real-or-random (approved):
utACK 77a19750b46916b93bb6a08837c26f585bd940fa

real-or-random:
> Changed to using xoshiro256++. It's even (slightly) faster now, in my benchmarks.

and still trivial to review given https://prng.di.unimi.it/xoshiro256plusplus.c :)

src/testrand_impl.h
sipa: Done.

robot-dreams:
ACK 77a19750b46916b93bb6a08837c26f585bd940fa

> and still trivial to review given https://prng.di.unimi.it/xoshiro256plusplus.c :)

Yes :)

Timing (n=1 sample size) are indeed slightly faster than before:

```
% time ./tests
./tests  28.31s user 3.28s system 99% cpu 31.766 total

# time ./tests
real	1m3.644s user	1m3.485s sys	0m0.140s

# time valgrind ./tests 2
real	11m21.063s user	10m25.844s sys	0m54.987s
```

elichai:
Post-Merge code review ACK 77a19750b46916b93bb6a08837c26f585bd940fa

Issue #1039, opened 2021-12-10
Make all source and header files self-contained


real-or-random:
This week I noticed that the includes in this project are a mess. Actually, each file should be self-contained, i.e., include the stuff that it needs (and ideally only the stuff that it needs) but at the moment neither of these are true for a lot of files in the repo. Just try compiling a random `.h` or `_impl.h` file.

This is not per se a problem for our "single translation unit" compilation (I learned that the cool kids call this ["unity builds"](https://en.wikipedia.org/wiki/Unity_build)) that we use in order to leverage the full potential of compiler optimizations without the need for LTO. The more important point (at least for me personally--though I may be the only regular contributor suffering here), it prevents the use of advanced tooling/IDE, e.g., my editor supports language servers such as ccls or clangd which "compile" the open file internally and provide a ton of features, e.g., completion, display of compile errors etc. But [none of this works properly](https://github.com/clangd/clangd/issues/45) if you can't compile individual files. You can say that's the fault of these tools but [self-contained files are also just good style and good engineering practice](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines.html#sf11-header-files-should-be-self-contained).

PR #924 was an attempt to clean up the includes but even the `include-what-you-use` tool there used there (or at least the way it has been used there) fixes includes only for the `.c` files.
  
Here are some things we'd need to do:
- [ ] Introduce header files for modules: Currently each `_impl.h` file in `src` has a corresponding `.h` file for other files to include. But we don't do this properly in `src/modules`. For example, the schnorrsig module uses function from the extrakeys module but there are only `_impl.h` files in  `src/modules/extrakeys` and no `.h` file for schnorrsig to include. 
- [ ] The same is true for some internal functions declared in `secp256k1.c`. There's no just header for other files to include.
- [ ] Then make sure that each file includes at least all the stuff it uses, i.e., compiles on its own. Maybe with the help of a tool like [clang-include-fixer](https://clang.llvm.org/extra/clang-include-fixer.html).
- [ ] ...and maybe enforce this on CI.
- [ ] Maybe make sure that each file includes at most the stuff it uses, e.g., with the help of a tool like [include-what-you-use](https://include-what-you-use.org/).

Again, I understand that I may the only one suffering, so fixing this may be on me but I still wanted to document my findings here. I don't except anyone disagreeing with the changes the mentioned here but please raise your hand if you do.

sipa:
It's complicated.

[...]

PR #1053, opened 2021-12-23
ecmult: move `_ecmult_odd_multiples_table_globalz_windowa`

### LINKED ISSUE #1035: Move secp256k1_ecmult_odd_multiples_table_globalz_windowa and fix docs
`secp256k1_ecmult_odd_multiples_table_globalz_windowa` is in the wrong module.  This function had been used within ecmult in the past but since 8c1c831bdb083dfa8b50fac16a0e17a7e1df4064, it is only used in ecmult_const. This also means its comment is wrong.

_Originally posted by @real-or-random in https://github.com/bitcoin-core/secp256k1/issues/942#issuecomment-987765042_

siv2r:
Fixes #1035

**Changes:**
    - move `secp256k1_ecmult_odd_multiples_table_globalz_windowa` function from ecmult to ecmult_const
    - remove outdated comment

### DIFF
--- a/src/ecmult_const_impl.h
+++ b/src/ecmult_const_impl.h
@@ -12,6 +12,22 @@
 #include "ecmult_const.h"
 #include "ecmult_impl.h"
 
+/** Fill a table 'pre' with precomputed odd multiples of a.
+ *
+ *  The resulting point set is brought to a single constant Z denominator, stores the X and Y
+ *  coordinates as ge_storage points in pre, and stores the global Z in globalz.
+ *  It only operates on tables sized for WINDOW_A wnaf multiples.
+ */
+static void secp256k1_ecmult_odd_multiples_table_globalz_windowa(secp256k1_ge *pre, secp256k1_fe *globalz, const secp256k1_gej *a) {
+    secp256k1_gej prej[ECMULT_TABLE_SIZE(WINDOW_A)];
+    secp256k1_fe zr[ECMULT_TABLE_SIZE(WINDOW_A)];
+
+    /* Compute the odd multiples in Jacobian form. */
+    secp256k1_ecmult_odd_multiples_table(ECMULT_TABLE_SIZE(WINDOW_A), prej, zr, a);
+    /* Bring them to the same Z denominator. */
+    secp256k1_ge_globalz_set_table_gej(ECMULT_TABLE_SIZE(WINDOW_A), pre, globalz, prej, zr);
+}
+
 /* This is like `ECMULT_TABLE_GET_GE` but is constant time */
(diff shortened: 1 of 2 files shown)


real-or-random:
Thanks! Can you rebase though? :P

siv2r:
> Can you rebase though?

Done :)

real-or-random (approved):
utACK 6287fa8c43736a9e187862a1ad0f7d159037e3fd

robot-dreams:
Concept ACK, thanks for addressing this!

Though should this wait for #899 to merge first to avoid a merge conflict in that one (which is a bit longer and already has two-ish ACKs)?

real-or-random:
> Though should this wait for #899 to merge first to avoid a merge conflict in that one (which is a bit longer and already has two-ish ACKs)?

Oh, thanks for pointing this out.

src/ecmult_const_impl.h
```diff
@@ -12,6 +12,22 @@
 #include "ecmult_const.h"
 #include "ecmult_impl.h"
 
+/** Fill a table 'pre' with precomputed odd multiples of a.
+ *
+ *  The resulting point set is brought to a single constant Z denominator, stores the X and Y
+ *  coordinates as ge_storage points in pre, and stores the global Z in rz.
```
real-or-random, line 18:
```suggestion
 *  coordinates as ge_storage points in pre, and stores the global Z in globalz.
```

Who would have expected that the global Z is stored in `globalz`?

real-or-random:
roconnor-blockstream told me that he doesn't mind rebasing #899 . 

By the way, it's debatable if this function should be prefixed with `ecmult` or `ecmult_const`... But I don't know, the enitre ecmult_const module is not really separated from the ecmult module (reuses `WINDOW_A` and the like). I believe this could be made cleaner but that could be done in a separate PR (see also #1039).

robot-dreams:
utACK 6287fa8c43736a9e187862a1ad0f7d159037e3fd (aside from @real-or-random 's suggested comment fixup)

I agree that:
- `secp256k1_ecmult_odd_multiples_table_globalz_windowa` is only called from `ecmult_const_impl.h` and thus should be moved there
- The function is no longer used for computation of `a*P + b*G`, thus the comment should be removed

siv2r:
Included the suggestion of https://github.com/bitcoin-core/secp256k1/pull/1053#discussion_r774774829

real-or-random (approved):
utACK 05e049b73c69002f498c3c9c21555fd91f95ccac

robot-dreams:
utACK 05e049b73c69002f498c3c9c21555fd91f95ccac (`diff` between removed and added lines is exactly as expected)

PR #1055, opened 2021-12-23
Initial release

jonasnick:
Based on #964


bgermann:
Any chance to make this happen?

hebasto:
I've tested 64e96eecb9127f420875a4345ba66335a4607014 with cross compiling for `--host=x86_64-w64-mingw32`.

The name of `dll` has been changed `bin/libsecp256k1-0.dll` --> `bin/libsecp256k1-1.dll`, reflecting the change in the `_LIB_VERSION_CURRENT` value.

doc/release-process.md
```diff
@@ -1,14 +1,17 @@
 # Release Process
 
 1. Open PR to master that
-   1. adds release notes to `doc/CHANGELOG.md` and
-   2. if this is **not** a patch release, updates `_PKG_VERSION_{MAJOR,MINOR}` and `_LIB_VERSIONS_*` in `configure.ac`
+   * adds release notes to `doc/CHANGELOG.md` (set the date of the new release to `YYYY-MM-DD`), and
+   * if this is **not** a patch release, updates `_PKG_VERSION_{MAJOR,MINOR}` and `_LIB_VERSIONS_*` in `configure.ac`.
```
hebasto, line 5:
```suggestion
   * if this is **not** a patch release, updates `_PKG_VERSION_{MAJOR,MINOR}` and `_LIB_VERSION_*` in `configure.ac`.
```

doc/release-process.md
```diff
@@ -1,14 +1,17 @@
...
-     Also include the release note commit bump `_PKG_VERSION_BUILD` and `_LIB_VERSIONS_*` in `configure.ac`.
-4. Tag the commit with `git tag -s vMAJOR.MINOR.PATCH`.
-5. Push branch and tag with `git push origin --tags`.
-6. Create a new GitHub release with a link to the corresponding entry in `doc/CHANGELOG.md`.
+     Create commit on the release branch that sets `_PKG_VERSION_IS_RELEASE` in `configure.ac` to `true` and replaces the dummy date in `doc/CHANGELOG.md` with the current date.
+   * if this **is** a patch release open a pull request that
+     * includes the bugfixes to the `MAJOR.MINOR` branch,
+     * includes the release note commit from master,
+     * replaces the dummy date in `doc/CHANGELOG.md` with the current date,
+     * bumps of `_PKG_VERSION_BUILD` and `_LIB_VERSIONS_*` in `configure.ac`.
```
hebasto, line 14:
```suggestion
     * bumps of `_PKG_VERSION_BUILD` and `_LIB_VERSION_*` in `configure.ac`.
```

hebasto (commented):
Tested e0ecf6deb3077edb9be803ac8ebb444b41d94be4 on macOS:
- the master branch:
```
% otool -L .libs/libsecp256k1.0.dylib 
.libs/libsecp256k1.0.dylib:
	/usr/local/lib/libsecp256k1.0.dylib (compatibility version 1.0.0, current version 1.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
```
- this PR:
```
% otool -L .libs/libsecp256k1.1.dylib
.libs/libsecp256k1.1.dylib:
	/usr/local/lib/libsecp256k1.1.dylib (compatibility version 2.0.0, current version 2.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1311.100.3)
```

Effectively, in`/usr/local/lib/libsecp256k1.<V>.dylib (compatibility version <X>.0.0, current version <X>.<Y>.0)`:
- V = _current_ - _age_
- X = _current_ + 1
- Y = _revision_

Is that what we want?

real-or-random:
I would need to check but have you read the discussion in https://github.com/bitcoin-core/secp256k1/pull/964#issuecomment-986248723 and the corresponding links?

https://autotools.info/libtool/version.html
https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html

The rules seem to be somewhat obscure at first glance of. But what you get there seems to correspond to what's described in the autotools mythbuster "Warning" paragraph for Linux.

real-or-random:
This may be interesting if we want to provide both libtool and cmake builds in the future: https://github.com/mesonbuild/meson/issues/1451#issuecomment-284845913

hebasto:
> I would need to check but have you read the discussion in [#964 (comment)](https://github.com/bitcoin-core/secp256k1/pull/964#issuecomment-986248723) and the corresponding links?
> 
> https://autotools.info/libtool/version.html https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html
> 
> The rules seem to be somewhat obscure at first glance of. But what you get there seems to correspond to what's described in the autotools mythbuster "Warning" paragraph for Linux.

Of course, I have.

My concerns are about V = _current_ - _age_, which will not increase monotonically, will it?

real-or-random:
> My concerns are about V = _current_ - _age_, which will not increase evenly, will it?

You mean monotonically? If we follow the algorithm in https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html, then it's easy to see that _current_ - _age_ will increase monotonically (if I'm not mistaken).

hebasto (approved):
ACK e0ecf6deb3077edb9be803ac8ebb444b41d94be4

real-or-random:
> You mean monotonically? If we follow the algorithm in [gnu.org/software/libtool/manual/html_node/Updating-version-info.html](https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html), then it's easy to see that _current_ - _age_ will increase monotonically (if I'm not mistaken).

Though I agree it can be strange due to setting `age = 0` in the last step. That means if we have a backwards-incompatible change (e.g., interface removed), we could have a hop from
 * `2:5:2` -> `0.2.5`  to 
 * `3:0:0` -> `3.0.0`.

So you skip over 1.x and 2.x entirely... I guess that's fine, just not "elegant". I'm pretty sure in that case one could also reset `current` to `1`, so one would have `1:0:0` -> `1.0.0`. For example, the rules in https://autotools.info/libtool/version.html are a little different and never set `age` to 0.... (It's all a huge mess... :shrug: )


This uses the translation for Linux (without the "+1" for Mac OS). See [the libtool source](https://github.com/autotools-mirror/libtool/blob/6132006b5fb6b95f31c30a972fbb829f93e1878b/build-aux/ltmain.in#L7001) for the ugly details.

I'm somewhat concerned that the first release commits us to this strange libtool convention with `current`, `revision`, `age` (which apparently is only used in libtool and no other build system). But on the other hand, I don't think there's a way to avoid this unless (we want to get rid of libtool as our main build system). I guess we need to apply tricks, e.g., in CMake to match libtool behavior, but AFAIU, you did this already in the CMake PR (I also found this https://cmake.org/cmake/help/latest/prop_tgt/MACHO_COMPATIBILITY_VERSION.html by the way). I guess we'd need to test it on multiple platforms.

hebasto:
FWIW, I've gathered details about potential future version changes, and their representations for different platforms using the current versioning implementation.

# Shared library ABI versioning across platforms

Note that the first 4 columns are identical for each platform-specific table.

## Linux -- `libsecp256k1.so.<V>`

| ABI change | Libtool diff | Libtool C:R:A | V=C-A | SONAME | SONAME.A.R |
|------------|:------------:|:-------------:|:-----:|--------|------------|
| Inception | | 0:0:0 | 0 | libsecp256k1.so.0 | libsecp256k1.so.0.0.0 |
| Initial release (this PR) | | 1:0:0 | 1 | libsecp256k1.so.1 | libsecp256k1.so.1.0.0 |
| No change | ++R | 1:1:0 | 1 | libsecp256k1.so.1 | libsecp256k1.so.1.0.1 |
| No change | ++R | 1:2:0 | 1 | libsecp256k1.so.1 | libsecp256k1.so.1.0.2 |
| BC * | ++C,R=0,++A | 2:0:1 | 1 | libsecp256k1.so.1 | libsecp256k1.so.1.1.0 |
| No change | ++R | 2:1:1 | 1 | libsecp256k1.so.1 | libsecp256k1.so.1.1.1 |
| BC | ++C,R=0,++A | 3:0:2 | 1 | libsecp256k1.so.1 | libsecp256k1.so.1.2.0 |
| BI ** | ++C,R=0,A=0 | 4:0:0 | 4 | libsecp256k1.so.4 | libsecp256k1.so.4.0.0 |

## macOS -- `libsecp256k1.<V>.dylib`

| ABI change | Libtool diff | Libtool C:R:A | V=C-A | compatibility_v=C+1.0.0 | current_v=C+1.R.0 |
|------------|:------------:|:-------------:|:-----:|:-----------------------:|:-----------------:|
| Inception | | 0:0:0 | 0 | 1.0.0 | 1.0.0 |
| Initial release (this PR) | | 1:0:0 | 1 | 2.0.0 | 2.0.0 |
| No change | ++R | 1:1:0 | 1 | 2.0.0 | 2.1.0 |
| No change | ++R | 1:2:0 | 1 | 2.0.0 | 2.2.0 |
| BC | ++C,R=0,++A | 2:0:1 | 1 | 3.0.0 | 3.0.0 |
| No change | ++R | 2:1:1 | 1 | 3.0.0 | 3.1.0 |
| BC | ++C,R=0,++A | 3:0:2 | 1 | 4.0.0 | 4.0.0 |
| BI | ++C,R=0,A=0 | 4:0:0 | 4 | 5.0.0 | 5.0.0 |

## Windows -- `libsecp256k1-<V>.dll`

| ABI change | Libtool diff | Libtool C:R:A | V=C-A |
|------------|:------------:|:-------------:|:-----:|
| Inception | | 0:0:0 | 0 |
| Initial release (this PR) | | 1:0:0 | 1 |
| No change | ++R | 1:1:0 | 1 |
| No change | ++R | 1:2:0 | 1 |
| BC | ++C,R=0,++A | 2:0:1 | 1 |
| No change | ++R | 2:1:1 | 1 |
| BC | ++C,R=0,++A | 3:0:2 | 1 |
| BI | ++C,R=0,A=0 | 4:0:0 | 4 |

---

\* BC -- backward-compatible

\** BI -- backward-incompatible

jonasnick:
I rebased the PR. After it is merged, the next step would be to create a release branch and tag the release. But after rereading the release process, I see two potential shortcomings:

- The version on the master branch is `MAJOR.MINOR.0-pre`. So the patch version is always 0 on master, even if we've actually released patches for `MAJOR.MINOR`. This might indicate at first glance that master somehow contains an older version. To avoid this and mark the unreleased version more clearly, we could use patch version 99 for the master branch.
- The release process currently does not describe when to replace the dummy date of the release in master. And doing this would be an extra step/PR. Maybe it's easier - at least for now - to avoid the dummy date entirely by using the current date as the release date in the PR to master. This may mean that that date needs to be updated multiple times while the release PR is not merged.

Bezzo100-lab (approved):

doc/release-process.md
```diff
@@ -1,14 +1,17 @@
 # Release Process
 
 1. Open PR to master that
-   1. adds release notes to `doc/CHANGELOG.md` and
-   2. if this is **not** a patch release, updates `_PKG_VERSION_{MAJOR,MINOR}` and `_LIB_VERSIONS_*` in `configure.ac`
+   * adds release notes to `doc/CHANGELOG.md` (set the date of the new release to dummy date `YYYY-MM-DD`), and
```
real-or-random, line 4: nit: new path of CHANGELOG.md (more instances below)

jonasnick:
Ok, I've made a couple of releases in my fork and updated the release process to include more git commands to avoid mistakes like forgetting to pull. I've also clarified a few steps to make it easier to follow.

CHANGELOG.md
```diff
@@ -0,0 +1,27 @@
...
+
+- Initial release
+
+## [0.1.0] - 2013-03-05 to 2022-12-08
+
+This version was never actually released.
+The number was given by the build system since the introduction of autotools in Jan 2014 (ea0fe5a5bf0c04f9cc955b2966b614f5f378c6f6).
+Therefore, this version number does not uniquely identify a set of source files.
+
+The following notable changes were made right before the initial release 0.2.0:
```
real-or-random, line 17: Yes, and I think that means we should list them under 0.2.0, no?

jonasnick:
Following a suggestion from @real-or-random I changed the proposed release process again. Instead of always having patch version 99 on master, the patch version in master is now always one higher than the most recent release. Besides being a bit more logical, this makes the process a tiny bit easier. The release process now also covers patch releases for old major or minor releases.

I'll still need to test this new process.

jonasnick:
Another suggestion from @real-or-random: clearly distinguish between "regular" and "maintenance" releases. This makes "regular" releases a bit simpler. I still need to test this.

CHANGELOG.md
```diff
@@ -1,11 +1,11 @@
...
-
-Each change falls into one of the following categories: Added, Changed, Deprecated, Removed, Fixed or Security.
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+## [0.2.0] - 2022-12-08
+
 ### Changed
  - Enable modules schnorrsig, extrakeys and ECDH by default in ./configure
```
real-or-random, line 10:
```suggestion
 - Enabled modules schnorrsig, extrakeys and ECDH by default in ./configure.
```
nit: seems that we use past tense everywhere else (and this fits the headings), so we could fix this in this PR.

CHANGELOG.md
```diff
@@ -16,7 +16,9 @@ Each change falls into one of the following categories: Added, Changed, Deprecat
 ### Added
  - Added `secp256k1_selftest`, to be used in conjunction with `secp256k1_context_static`.
 
-## [MAJOR.MINOR.PATCH] - YYYY-MM-DD
+## [0.1.0] - 2013-03-05 to 2021-12-25
+
+This version was never actually released.
```
real-or-random, line 21:
```suggestion
This version was in fact never released.
```
(actually sounds a little bit like it *was* released in some sense, just not properly...)

doc/release-process.md
```diff
@@ -1,14 +1,55 @@
...
-4. Tag the commit with `git tag -s vMAJOR.MINOR.PATCH`.
-5. Push branch and tag with `git push origin --tags`.
-6. Create a new GitHub release with a link to the corresponding entry in `doc/CHANGELOG.md`.
+This document outlines the process for releasing versions of the form `$MAJOR.$MINOR.$PATCH`.
+
+We distinguish between two types of releases: *regular* and *maintenance* releases.
+Regular releases are releases of a new major or minor version as well as patches of the most recent release.
+Maintenance releases, on the other hand, are required for patches of older releases.
+
+Note that adding a release entry to CHANGELOG.md requires a release date. This date should match the dates of the remaining steps in the release process, including the date of the tag and the GitHub release.
```
real-or-random, line 9:
```suggestion
Note that adding a release entry to [CHANGELOG.md](../CHANGELOG.md) requires a release date. This date should match the dates of the remaining steps in the release process, including the date of the tag and the GitHub release.
```

doc/release-process.md
```diff
@@ -1,14 +1,55 @@
...
+## Regular release
+
+1. Open PR to the master branch that includes the following commits:
+   1. A commit (with commit message `"release: prepare for $MAJOR.$MINOR.$PATCH"`, for example) that
+       * adds release notes to [CHANGELOG.md](../CHANGELOG.md) and
+       * updates `_PKG_VERSION_*`, `_LIB_VERSION_*`, and sets `_PKG_VERSION_IS_RELEASE` to `true` in `configure.ac`.
+   2. A commit (with commit message `"release: update master after $MAJOR.$MINOR.$PATCH"`, for example) that sets `_PKG_VERSION_IS_RELEASE` to `false` and `_PKG_VERSION_PATCH` to `$PATCH + 1` and increases `_LIB_VERSION_REVISION`.
+2. After the PR is merged, tag the commit:
+   ```
+   BRANCH_COMMIT=<commit of step 1.1>
```
real-or-random, line 22:
```suggestion
   BRANCH_COMMIT=<commit of step 1.i> 
```
That's how it's rendered.

doc/release-process.md
```diff
@@ -1,14 +1,55 @@
...
+
+This process also assumes that there will be no minor releases for old major releases.
+
+## Regular release
+
+1. Open PR to the master branch that includes the following commits:
+   1. A commit (with commit message `"release: prepare for $MAJOR.$MINOR.$PATCH"`, for example) that
+       * adds release notes to [CHANGELOG.md](../CHANGELOG.md) and
+       * updates `_PKG_VERSION_*`, `_LIB_VERSION_*`, and sets `_PKG_VERSION_IS_RELEASE` to `true` in `configure.ac`.
+   2. A commit (with commit message `"release: update master after $MAJOR.$MINOR.$PATCH"`, for example) that sets `_PKG_VERSION_IS_RELEASE` to `false` and `_PKG_VERSION_PATCH` to `$PATCH + 1` and increases `_LIB_VERSION_REVISION`.
```
real-or-random, line 19:
```suggestion
   2. A commit (with commit message `"release: Bump version after $MAJOR.$MINOR.$PATCH"`, for example) that sets `_PKG_VERSION_IS_RELEASE` to `false` and `_PKG_VERSION_PATCH` to `$PATCH + 1` and increases `_LIB_VERSION_REVISION`.
```
Mentioning "master" feels somewhat wrong because the commit will end up on other branches, too. I think this is minimally clearer. 

(My suggestion includes starting with a capital letter after the colon in the commit messages, because I find this more readable. But this is just my preference, feel free to stick to lower case.)

doc/release-process.md
```diff
@@ -1,14 +1,55 @@
...
+3. Push tag:
+   ```
+   git push git@github.com:bitcoin-core/secp256k1.git v$MAJOR.$MINOR.$PATCH
+   ```
+4. Create a new GitHub release with a link to the corresponding entry in [CHANGELOG.md](../CHANGELOG.md).
+
+## Maintenance release
+
+Note that some releases do not require maintenance, even if they include bugs.
+This is the case when users of the library are able to update to a newer, compatible release that does not include the bug.
```
real-or-random, line 34:
```suggestion
This is the case when users of the library are able to update to a newer, backwards-compatible release that does not include the bug.
```

doc/release-process.md
```diff
@@ -1,14 +1,55 @@
...
+
+1. Open PR to the master branch that includes a commit (with commit message `"release: prepare for $MAJOR.$MINOR.$PATCH"`, for example) that adds release notes to [CHANGELOG.md](../CHANGELOG.md).
+2. If `$PATCH = 1`, create maintenance branch `$MAJOR.$MINOR`:
+     ```
+     BRANCH_COMMIT=<id of commit that set _PKG_VERSION_* to $MAJOR.$MINOR>
```
real-or-random, line 39: Couldn't we just use the tag here?
```suggestion
     BRANCH_COMMIT=v$MAJOR.$MINOR.$PATCH
```
If yes, then maybe it fits on a single line. Not sure if that's more readable.

real-or-random (commented):
That looks really good, the split into two sections is much easier to read.

A high-level thought for maintenance releases: In principle, we could add multiple PRs to a maintenance branch before we do the PR that cut the release. And this is true even for the first maintenance release. So a more general process: is
 - Create a branch from the tag, bump PATCH+1 on this branch and set Release = false
 - (Work on the branch)
 - Open a PR that creates the release ("prepare" commit + second commit that bumps PATCH+1) again.\\

jonasnick:
Testing was ok. I released 0.2.0, 0.3.0 and 0.2.1 (maintenance) in my fork which unveiled a minor bug in the process (fixed now).

CHANGELOG.md
jonasnick: Fixed (the example at keepachangelog.com uses both present and past tense)

CHANGELOG.md
jonasnick: fixed

doc/release-process.md
jonasnick: fixed

doc/release-process.md
jonasnick: fixed

doc/release-process.md
jonasnick: true, fixed (and sticked with lowercase for now)

doc/release-process.md
jonasnick: Chose sipa's suggestion below.

doc/release-process.md
jonasnick: done (should be `v$MAJOR.$MINOR.0`)

doc/release-process.md
```diff
@@ -1,14 +1,55 @@
...
+   ```
+3. Push tag:
+   ```
+   git push git@github.com:bitcoin-core/secp256k1.git v$MAJOR.$MINOR.$PATCH
+   ```
+4. Create a new GitHub release with a link to the corresponding entry in [CHANGELOG.md](../CHANGELOG.md).
+
+## Maintenance release
+
+Note that some releases do not require maintenance, even if they include bugs.
```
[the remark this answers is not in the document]
jonasnick: done

jonasnick:
@real-or-random It's true that the process doesn't work if a maintenance release requires multiple PRs. Agree that we can adjust this once we have to do such a release.

doc/release-process.md
```diff
@@ -1,14 +1,54 @@
...
+Regular releases are releases of a new major or minor version as well as patches of the most recent release.
+Maintenance releases, on the other hand, are required for patches of older releases.
+
+Note that adding a release entry to [CHANGELOG.md](../CHANGELOG.md) requires a release date. This date should match the dates of the remaining steps in the release process, including the date of the tag and the GitHub release.
+
+This process also assumes that there will be no minor releases for old major releases.
+
+## Regular release
+
+1. Open PR to the master branch that includes the following commits:
```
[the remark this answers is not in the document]
real-or-random: Urghs, this is a good point. (And this was not a problem with the previous approach...)

> Don't care about it

 Well we anyway need to open the PR for the right date to get the release date right... So yeah, this is doable but it still seems error-prone. (And releases are the one thing where we don't want to make mistakes, because we cannot correct them without another release.)

> Use two separate PRs, and tag the merge-commit of the first.

Ok, that works. I tried to avoid two PRs but ok.  What if we have a single PR, merge that locally with the merge tool, tag the merge, let the merging maintainer add the "bump after release" commit manually, and then push the result. That saves the second PR and having that one commit with PR is harmless. (And if there's a problem with the "bump after release" commit, then we can just fix it on master and it won't affect the release.) 

> Use two commits (or at least the first one) that's not part of a PR, but just immediately pushed by a maintainer to the master branch.

Is that what I just wrote above? I'm not sure. You say at least the first commit should be manual and not part of the PR, and I don't see how that would work. (In my suggestion, the *second* commit is done manually.)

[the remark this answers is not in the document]
real-or-random: > I don't see why the second commit matters? It's past the tag, so it can't affect it.
> 
> All that matters is that there is a stable point on the master branch after the first commit. That can be achieved by it being in a separate PR, and using the merge commit of that PR as the tag, or by just pushing a commit and marking that as the tag.

Yes, we're in full agreement here! 

So I think we should
1. tag the merge commit of a PR
2. then get the "release: update master ..." commit on master (with or without a PR). 

Now that I think about  this again, I tend so say let's make a PR so that we never break the implicit "don't push without PR" rule (this also means other maintainers will see it etc.) But to make sure the process is not overly complicated and interactive, I suggest that the maintainer preparing the release simply merges this second PR without an ACK. (And indeed, even if we then spot a mistake in the second PR, it won't affect the release).

hebasto:
nit: I found the change type order [here](https://keepachangelog.com/en/1.0.0/#how) more preferable. Maybe follow it and reorder in our CHANGELOG.md as well like that:
- Added
- Changed
- Deprecated

?

configure.ac
```diff
@@ -4,20 +4,20 @@ AC_PREREQ([2.60])
...
 
 # The library version is based on libtool versioning of the ABI. The set of
 # rules for updating the version can be found here:
 # https://www.gnu.org/software/libtool/manual/html_node/Updating-version-info.html
 # All changes in experimental modules are treated as if they don't affect the
 # interface and therefore only increase the revision.
-define(_LIB_VERSION_CURRENT, 0)
-define(_LIB_VERSION_REVISION, 0)
+define(_LIB_VERSION_CURRENT, 1)
+define(_LIB_VERSION_REVISION, 1)
```
hebasto, line 17: Why has `revision` been incremented? Incrementing of `current` implies setting `revision` to zero.

doc/release-process.md
```diff
@@ -1,14 +1,54 @@
...
-4. Tag the commit with `git tag -s vMAJOR.MINOR.PATCH`.
-5. Push branch and tag with `git push origin --tags`.
-6. Create a new GitHub release with a link to the corresponding entry in `doc/CHANGELOG.md`.
+This document outlines the process for releasing versions of the form `$MAJOR.$MINOR.$PATCH`.
+
+We distinguish between two types of releases: *regular* and *maintenance* releases.
+Regular releases are releases of a new major or minor version as well as patches of the most recent release.
+Maintenance releases, on the other hand, are required for patches of older releases.
+
+Note that adding a release entry to [CHANGELOG.md](../CHANGELOG.md) requires a release date. This date should match the dates of the remaining steps in the release process, including the date of the tag and the GitHub release.
```
hebasto, line 9: Is ti worth mentioning in the `CHANGELOG.md` for each regular release whether it is ABI backward compatible or not?
A clearl statement about ABI backward compatibility will also help to modify `_LIB_VERSION_*` variables properly.

doc/release-process.md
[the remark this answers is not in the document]
jonasnick: That's quite unfortunate, I did not consider this.

> So I think we should
> 1. tag the merge commit of a PR
> 2. then get the "release: update master ..." commit on master (with or without a PR).

One issue with this approach is that `IS_RELEASE` is set to true in the merge commit as well as in the commit that set it to true. Therefore, we have two (maybe significantly) different states of the repo that are marked as `IS_RELEASE`. Another problem with this approach is that the repo is locked after the first PR until the second commit lands in the repo (unless you ignore `IS_RELEASE` and `VERSION_PATCH` for a while).

Another approach is that we revert the release process back to a version where we set the patch version of master to `99` and branch off from before we set `IS_RELEASE` which is preserved [here](https://github.com/jonasnick/secp256k1/blob/initial-release-backup2/doc/release-process.md). This version, however, has some subtle edge cases that made me prefer the current version of the release process.

I would have no problem with the approach that locks the repo until the PR is merged ("don't care about it") if we could reasonably ensure that no mistakes will happen. But I have no better ideas than mentioning it in the release process and in the OP of the PR.

configure.ac
jonasnick: Did you notice that `CURRENT` and `REVISION` are incremented in different commits? The idea is that after we create the release commit we immediately increment the package `PATCH` version and library `REVISION` to have sensible version numbers in master that are distinct from the release.

doc/release-process.md
jonasnick: It would make more sense to me to do the opposite and mention that we make little guarantees about the library version. We don't have methods to accurately determine ABI compatibility and so we can either put reasonable effort into attempting proper library versioning or not having library versions at all. So far the consensus seemed to be that we do the former.

jonasnick:
@hebasto Sure, we can do that. I don't have a particular preference about the order except that I wouldn't agree with the "security-last" approach of keepachangelog.com.

configure.ac
hebasto: > Did you notice that `CURRENT` and `REVISION` are incremented in different commits?

I didn't (

In that case, changes make sense.

doc/release-process.md
jonasnick: OTOH if our resonable-effort approach to library versioning would indicate that it's not consistent with semantic versioning, then I agree that it makes sense to mention that in the changelog. You think that's worth adding to the `release-process.md`?

jonasnick: > I don't think it's unreasonable to at least try and have a very good chance of getting it right

Good, then there doesn't seem to be disagreement.

If we want to mention ABI compatibility in every  release note, we could add a section `### ABI compatibility` (or so) to the release notes that (somewhat cautiously) mentions how we set the library version.

doc/release-process.md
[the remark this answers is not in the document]
jonasnick: As per our discussion elsewhere I changed the release process to @real-or-random's suggestion above.

doc/release-process.md
jonasnick: I added a section `### ABI compatibility` to the changelog and mention it in the release process.

doc/release-process.md
```diff
@@ -1,14 +1,53 @@
...
+   ```
+3. Open a PR to the master branch with a commit (using message `"release: bump version after $MAJOR.$MINOR.$PATCH"`, for example) that sets `_PKG_VERSION_IS_RELEASE` to `false` and `_PKG_VERSION_PATCH` to `$PATCH + 1` and increases `_LIB_VERSION_REVISION`.
+  * If other maintainers are not present to approve the PR, it can be merged without ACKs.
+4. Create a new GitHub release with a link to the corresponding entry in [CHANGELOG.md](../CHANGELOG.md).
+
+## Maintenance release
+
+Note that bugfixes only need to be backported to releases for which no compatible release without the bug exists.
+
+1. Open PR to the master branch that includes a commit (with commit message `"release: prepare for $MAJOR.$MINOR.$PATCH"`, for example) that adds release notes to [CHANGELOG.md](../CHANGELOG.md).
```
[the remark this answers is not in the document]
real-or-random: > I'd say the only required step is the creation of a `3.1` branch, branching off the `v3.1.0` merge commit, and adding the release notes relevant for `v3.1.1` there.

Hm yeah, I don't know. So far I held the belief that it's a good thing to treat the changelog more like a "global" thing that lists all releases (regular and maintenance across branches). Because it makes sense to keep a list of all releases somehwere. And then the canonical place for it is master. (Or outside the repo, e.g., in a wiki, but that comes with other difficulties.) 

How do other projects handle this?

(Happy to postpone this and all other things related to  maintenance releases to the future)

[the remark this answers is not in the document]
jonasnick: > How about moving step (1) to be after step (6), and calling it Add $MAJOR.$MINOR.$PATCH to release notes? Step (1) happens in a different branch than everything else, so the order shouldn't matter.

sgtm

doc/release-process.md
```diff
@@ -1,14 +1,53 @@
...
+You should coordinate with the other maintainers on the release date, if possible.
+This date will be part of the release entry in [CHANGELOG.md](../CHANGELOG.md) and it should match the dates of the remaining steps in the release process (including the date of the tag and the GitHub release).
+It is best if the maintainers are present during the release, so they can help ensure that the process is followed correctly and, in the case of a regular release, they are aware that they should not modify the master branch between merging the PR in step 1 and the PR in step 3.
+
+This process also assumes that there will be no minor releases for old major releases.
+
+## Regular release
+
+1. Open a PR to the master branch with a commit (using message `"release: prepare for $MAJOR.$MINOR.$PATCH"`, for example) that
+       * adds release notes to [CHANGELOG.md](../CHANGELOG.md) (make sure to include an entry for `### ABI Compatibility`) and
```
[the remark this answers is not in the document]
jonasnick: Fixed.

doc/release-process.md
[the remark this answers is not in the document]
jonasnick: done

jonasnick:
@hebasto Thanks. I fixed the formatting.

CHANGELOG.md
```diff
@@ -0,0 +1,28 @@
...
+
+## [Unreleased]
+
+## [0.2.0] - 2022-12-08
+
+### Added
+ - Added `secp256k1_selftest`, to be used in conjunction with `secp256k1_context_static`.
+
+### Changed
+ - Enabled modules schnorrsig, extrakeys and ECDH by default in ./configure.
```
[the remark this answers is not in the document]
jonasnick: Fixed

jonasnick:
Set the release date to 2022-12-12 (today).

real-or-random (approved):
ACK 38c9096236cd0c45db2aa258888caf97fbd92b78

hebasto (approved):
ACK 38c9096236cd0c45db2aa258888caf97fbd92b78, I have reviewed the changes, except for Add/Changed//Deprecated paragraphs in the CHANGELOG.md, and they look OK.

The entire release process differs from one in the Bitcoin Core repo, but it still looks reasonable.

jonasnick:
@real-or-random noticed that the distribution (`make dist`) is missing some files. I added a commit to fix that.

real-or-random (approved):
ACK 90618e9263ebc2a0d73d487d6d94fd3af96b973c

PR #1056, opened 2021-12-26
Save negations in var-time group addition

peterdettman:
- Updated _gej_add_var, _gej_add_ge_var, _gej_add_zinv_var
- 2 fewer _fe_negate in each method
- Updated operation counts and standardize layout
- Added internal benchmark for _gej_add_zinv_var

benchmark_internal shows about 2% speedup in each method as a result (64bit).

### DIFF
--- a/sage/prove_group_implementations.sage
+++ b/sage/prove_group_implementations.sage
@@ -40,29 +40,26 @@ def formula_secp256k1_gej_add_var(branch, a, b):
   s2 = s2 * a.Z
   h = -u1
   h = h + u2
-  i = -s1
-  i = i + s2
+  i = -s2
+  i = i + s1
   if branch == 2:
     r = formula_secp256k1_gej_double_var(a)
     return (constraints(), constraints(zero={h : 'h=0', i : 'i=0', a.Infinity : 'a_finite', b.Infinity : 'b_finite'}), r)
   if branch == 3:
     return (constraints(), constraints(zero={h : 'h=0', a.Infinity : 'a_finite', b.Infinity : 'b_finite'}, nonzero={i : 'i!=0'}), point_at_infinity())
-  i2 = i^2
+  t = h * b.Z
+  rz = a.Z * t
   h2 = h^2
+  h2 = -h2
   h3 = h2 * h
-  h = h * b.Z
-  rz = a.Z * h
   t = u1 * h2
-  rx = t
-  rx = rx * 2
+  rx = i^2
   rx = rx + h3
-  rx = -rx
-  rx = rx + i2
-  ry = -rx
-  ry = ry + t
-  ry = ry * i
+  rx = rx + t
+  rx = rx + t
+  t = t + rx
+  ry = t * i
   h3 = h3 * s1
-  h3 = -h3
   ry = ry + h3
   return (constraints(), constraints(zero={a.Infinity : 'a_finite', b.Infinity : 'b_finite'}, nonzero={h : 'h!=0'}), jacobianpoint(rx, ry, rz))
 
@@ -80,28 +77,25 @@ def formula_secp256k1_gej_add_ge_var(branch, a, b):
   s2 = s2 * a.Z
   h = -u1
   h = h + u2
-  i = -s1
-  i = i + s2
+  i = -s2
+  i = i + s1
   if (branch == 2):
     r = formula_secp256k1_gej_double_var(a)
     return (constraints(zero={b.Z - 1 : 'b.z=1'}), constraints(zero={a.Infinity : 'a_finite', b.Infinity : 'b_finite', h : 'h=0', i : 'i=0'}), r)
   if (branch == 3):
     return (constraints(zero={b.Z - 1 : 'b.z=1'}), constraints(zero={a.Infinity : 'a_finite', b.Infinity : 'b_finite', h : 'h=0'}, nonzero={i : 'i!=0'}), point_at_infinity())
-  i2 = i^2
-  h2 = h^2
-  h3 = h * h2
   rz = a.Z * h
+  h2 = h^2
+  h2 = -h2
+  h3 = h2 * h
   t = u1 * h2
-  rx = t
-  rx = rx * 2
+  rx = i^2
   rx = rx + h3
-  rx = -rx
-  rx = rx + i2
-  ry = -rx
-  ry = ry + t
-  ry = ry * i
+  rx = rx + t
+  rx = rx + t
+  t = t + rx
+  ry = t * i
   h3 = h3 * s1
-  h3 = -h3
   ry = ry + h3
   return (constraints(zero={b.Z - 1 : 'b.z=1'}), constraints(zero={a.Infinity : 'a_finite', b.Infinity : 'b_finite'}, nonzero={h : 'h!=0'}), jacobianpoint(rx, ry, rz))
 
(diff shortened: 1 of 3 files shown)


real-or-random:
Woho, we'll soon need a few more Peters to review all those improvements! :)

sipa:
Can you update the symbolic versions of this code in `sage/prove_group_implementations.sage` too?

sipa:
ACK 66a412939351396a6351b062dff820989a63cc56

I re-derived these formulas by manipulating the sage code until it matches the forms in this PR, and then comparing the result with the C implementations.

sipa:
ACK e6ed3705b602cd8e51612ce6d3728654bc2d1aec. Verified it's just a rebase of 66a412939351396a6351b062dff820989a63cc56

src/bench_internal.c
```diff
@@ -245,6 +245,15 @@ void bench_group_add_affine_var(void* arg, int iters) {
     }
 }
 
+void bench_group_add_zinv_var(void* arg, int iters) {
+    int i;
+    bench_inv *data = (bench_inv*)arg;
+
+    for (i = 0; i < iters; i++) {
+        secp256k1_gej_add_zinv_var(&data->gej[0], &data->gej[0], &data->ge[1], &data->fe[0]);
```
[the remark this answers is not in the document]
peterdettman: Done.

sage/prove_group_implementations.sage
```diff
@@ -85,43 +82,40 @@ def formula_secp256k1_gej_add_ge_var(branch, a, b):
...
-    return (constraints(), constraints(nonzero={b.Infinity : 'b_infinite'}), a)
   if branch == 1:
     bzinv2 = bzinv^2
     bzinv3 = bzinv2 * bzinv
     rx = b.X * bzinv2
     ry = b.Y * bzinv3
     rz = 1
     return (constraints(), constraints(zero={b.Infinity : 'b_finite'}, nonzero={a.Infinity : 'a_infinite'}), jacobianpoint(rx, ry, rz))
+  if branch == 0:
+    return (constraints(), constraints(nonzero={b.Infinity : 'b_infinite'}), a)
```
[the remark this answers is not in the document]
peterdettman: It was just for consistency across the several methods, but on review 'b' is from a lookup table and should never actually be infinity in current usages. So as little as it matters, the check should come after checks that might actually happen (if we can't just require it to be non-infinity anyway).

I've adopted your fixup for the sage script, thanks.

peterdettman:
Rebased with tweak to benchmark and updated sage  file from @real-or-random .

real-or-random:
@peterdettman Nice. :) Can you squash this? I think at least the `fixup!` should be squashed but it's probably also good to change  the sage stuff in the same commit as the C code, then C and sage are always consistent.

sage/prove_group_implementations.sage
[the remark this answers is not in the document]
real-or-random: > It was just for consistency across the several methods, but on review 'b' is from a lookup table and should never actually be infinity in current usages. So as little as it matters, the check should come after checks that might actually happen (if we can't just require it to be non-infinity anyway).

We could VERIFY_CHECK it. But the branch can probably be predicted then, so I don't think it hurts (and it's certainly safer).

peterdettman:
Squashed everything into a single commit.

real-or-random (approved):
ACK 2d0e1154c1939e8e7ffb05f64faf3f626fcb7000

sage/prove_group_implementations.sage
```diff
@@ -80,43 +77,42 @@ def formula_secp256k1_gej_add_ge_var(branch, a, b):
...
-    return (constraints(), constraints(nonzero={b.Infinity : 'b_infinite'}), a)
-  if branch == 1:
+  if branch == 0 or branch == 1:
     bzinv2 = bzinv^2
     bzinv3 = bzinv2 * bzinv
     rx = b.X * bzinv2
     ry = b.Y * bzinv3
     rz = 1
+    if branch == 0:
+      return (constraints(), constraints(nonzero={a.Infinity : 'a_infinite', b.Infinity : 'b_infinite'}), point_at_infinity())
```
jonasnick, line 112: Why is this branch inside this block `branch == 0 or branch == 1` and not just before? It doesn't use any of the local variables. Also, why does it exist at all, given that there's no corresponding branch in the C code?

real-or-random: This branch corresponds to https://github.com/bitcoin-core/secp256k1/blob/2d0e1154c1939e8e7ffb05f64faf3f626fcb7000/src/group_impl.h#L455 in the C code, which is equivalent to
```
if (b->infinity) {
    r ->infinity = 1
} else {
    b ->infinity = 1
}
```

So the mapping then is  
 * `(C) a->infinity && b->infinity` -> `(sage) branch == 0`
 * `(C) a->infinity && !b->infinity` -> `(sage) branch == 1`

When I wrote this I assumed that in sage, we need to write it as an explicit branch because we have no infinity field but instead `point_at_infinity()` and `jacobianpoint(x, y, z)` constructors. But now I see that it's not entirely true. The constructor is actually [`jacobianpoint(x, y, z, infinity)`](https://github.com/bitcoin-core/secp256k1/blob/master/sage/weierstrass_prover.sage#L18), so there's a much simpler way that's closer to the sage code, see
https://github.com/real-or-random/secp256k1/commit/0946486cec44c99befe89fbb6ae5be8f48e33dcf .

@peterdettman Can you pick that commit and squash again?

sage/prove_group_implementations.sage
```diff
@@ -288,7 +280,7 @@ if __name__ == "__main__":
   success = True
   success = success & check_symbolic_jacobian_weierstrass("secp256k1_gej_add_var", 0, 7, 5, formula_secp256k1_gej_add_var)
   success = success & check_symbolic_jacobian_weierstrass("secp256k1_gej_add_ge_var", 0, 7, 5, formula_secp256k1_gej_add_ge_var)
-  success = success & check_symbolic_jacobian_weierstrass("secp256k1_gej_add_zinv_var", 0, 7, 5, formula_secp256k1_gej_add_zinv_var)
+  success = success & check_symbolic_jacobian_weierstrass("secp256k1_gej_add_zinv_var", 0, 7, 6, formula_secp256k1_gej_add_zinv_var)
```
real-or-random, line 283: Anyway, I forgot to make the corresponding change below for the `--exhaustive` version. But it doesn't matter as the fixup mentioned above reverts this change here.

sage/prove_group_implementations.sage
jonasnick: > `b ->infinity = 1`

You mean `r ->infinity = 0`? Anyway, I see what you meant to represent with this branch. Your fixup looks good.

real-or-random: yes... edited my comment to fix.

sipa:
Anything left to do here?

peterdettman:
I may need to followup request from @real-or-random above (unsure if done yet), and rebase.

peterdettman:
OK, fixup applied, squashed, rebased. Ready from my end.

real-or-random (approved):
ACK 2f984ffc45eba89faa9e79da3d5d5bd50a6c1c3d

jonasnick (commented):
ACK 2f984ffc45eba89faa9e79da3d5d5bd50a6c1c3d

real-or-random:
@sipa want to reACK this?

Issue #1051, opened 2021-12-23
Try a non-uniform group law (e.g., for ecmult_gen)?


real-or-random:
The group law in `secp256k1_gej_add_ge` is uniform, i.e., it works for P + Q no matter if P != Q or P = Q. (It still needs to handle infinity and has another special case.) 

It needs 7 fe mul and 5 fe sqr, the best non-uniform group law in EFD needs one fewer sqr (https://www.hyperelliptic.org/EFD/g1p/auto-shortw-jacobian-0.html#addition-madd-2007-bl). Mabye it's faster also in practice. We could simply implement it and benchmark it.

Depending on the specifics of our ecmult_gen algorithm, it may be the case that this algorithm never adds P + Q with P = Q (for example, the ecmult algorithm within the MuSig-DN circuit has a similar property)). Then, we could get away with the non-uniform law in ecmult_gen.

I think we shouldn't really think about this before https://github.com/bitcoin-core/secp256k1/pull/693 has been done.

apoelstra:
ISTR trying "a better" group law from EFD, many years ago, which turned out to be slower because of normalization even though it technically required fewer ops.

May be worth revisiting this after recent Dettman normalization bounds improvements though?

real-or-random:
> ISTR trying "a better" group law from EFD, many years ago, which turned out to be slower because of normalization even though it technically required fewer ops.

Maybe what you're remembering is https://github.com/bitcoin-core/secp256k1/blob/master/src/group_impl.h#L274-L280 ? But this is for `gej_double`. (I had brought this up in https://github.com/bitcoin-core/secp256k1/pull/1033#issuecomment-999520374).

peterdettman:
Doubtless it will be faster. The most obvious use-case to me is for precomputed tables of small multiples.

Back in the long-long-ago I PR'd some CoZ formulae for precomputed tables (https://github.com/bitcoin-core/secp256k1/pull/41 - LOL at "That's the GMP inversion[...]; it will be hard to match it's performance.") and I will be looking to resurrect that as a private implementation for a separated out "precomp_table" (or some such) module.

However @sipa also noted there "I don't like relying on relatively complex reasoning to show that a code path shouldn't be reachable, even if it just contains an assert" and that was in relation to this issue of assuming P != +/-Q.

[...]

PR #1058, opened 2021-12-29
Signed-digit multi-comb ecmult_gen algorithm

sipa:
A third iteration of the signed-digit multi-comb ecmult_gen algorithm (earlier attempts: #693, and #546 by Peter Dettman). Short summary:
* A new constant-time point multiplication algorithm with precomputation (so only used for multiply with G).
* Based on section 3.3 of https://eprint.iacr.org/2012/309 by Mike Hamburg.
* Configurable through two parameters: COMB_BLOCKS and COMB_TEETH
  * Currently only 3 predefined configurations reachable through ./configure. All three are included in precomputed_ecmult_gen.c and tested in CI.
    * `--with-ecmult-gen-kb=2`: 2 kB table with COMB_BLOCKS=2 COMB_TEETH=5
    * `--with-ecmult-gen-kb=22`: 22 kB table with COMB_BLOCKS=11 COMB_TEETH=6
    * `--with-ecmult-gen-kb=86`: 86 kB table with COMB_BLOCKS=43 COMB_TEETH=6
  * Many more configurations can be reached by manually setting the macros. These are not tested.

Compared with the previous PR #693:
* Updated to the new static-precomputation-only model (#893).
* Just 3 curated configurations reachable through configure.
* Removed some optimizations that do not matter (much).
* Do blinding through an final correction add rather than an initial start point, which may later permit usage of incomplete addition formulae (#1051).
* The recoding of the input scalar to signed bit representation is done slightly differently, which needs fewer special cases.


sipa:
Benchmarks:

* Ryzen 5950X, GCC 11.2.0

```
master:       schnorrsig_sign               ,    18.1       ,    18.1       ,    18.1    
pr1058 kb=2:  schnorrsig_sign               ,    17.9       ,    17.9       ,    17.9 
pr1058 kb=22: schnorrsig_sign               ,    15.1       ,    15.1       ,    15.1    
pr1058 kb=86: schnorrsig_sign               ,    14.4       ,    14.4       ,    14.4 
```

sipa:
I occurs to me that we could actually avoid the cost of doing the scalar halving at ecmult_gen time, by instead having precomputed tables with multiples of G/2 instead of G.

peterdettman:
> I occurs to me that we could actually avoid the cost of doing the scalar halving at ecmult_gen time, by instead having precomputed tables with multiples of G/2 instead of G.

The scalar halving also ensures that the low bit (the shifted-away bit) is 0.

sipa:
I don't think that matters? Even/odd has no special meaning when working modulo a prime.

The PR currently uses the bits of scalar `(input + 2^COMB_BITS - 1 - blind) * 2^-1` to select multiples of G to add together.

My suggestion is that instead it could use the bits of `(input + 2^COMC_BITS - 1 - blind)` to select multiples of G/2 to add together.

sipa:
It seems I'm wrong, but I'm confused why!

peterdettman:
I think confusion is creeping in since after the halving the scalar isn't a scalar value anymore; it's in signed-digit form, which can only represent an odd value. In particular the bits of a scalar s in signed-digit form represent the scalar value 2*s+1. I think you should also be careful not to reason modulo the order in signed-digit form.

sipa:
It works, but you need to use the bits of `input - blind + (2^COMB_BITS - 1)/2` instead. That's what you get when you substitute `2*(input-blind)` for `e` in the formula in the paper (because now we're trying to compute `input*G = 2*(input-blind)*(G/2) + blind*G`).

peterdettman:
Oh I see, using double the input and the blind (of G/2) lets the halving be moved to precomputation. Nice.

src/ecmult_gen_impl.h
```diff
@@ -25,41 +25,133 @@ static int secp256k1_ecmult_gen_context_is_built(const secp256k1_ecmult_gen_cont
...
+     * Next, we write (gn-b)*G as a sum of values (2*bit_i-1) * 2^i * G, for i=0..COMB_BITS-1.
+     * The values bit_i can be found as the binary representation of
+     * (gn + 2^COMB_BITS - 1 - b)/2 (mod order), stored in recoded.
+     *
+     * The value (2^COMB_BITS - 1 - b) is precomputed as ctx->scalar_offset, and bG is
+     * precomputed as ctx->final_point_add. Thus recoded can be written as
+     * recoded = (gn + scalar_offset)/2, and R becomes the sum of (2*bit_i-1)*2^i*G
+     * values plus final_point_add. */
+
+    /* Compute (gn + 2^COMB_BITS - 1 - 1)/2 value as a scalar. */
```
peterdettman, line 59: Nit: -1 -b

sipa:
Updated to use the avoid-halving-scalar trick.

src/ecmult_gen_impl.h
sipa: Superseded by new description.

src/ecmult_gen.h
```diff
@@ -10,19 +10,90 @@
...
 typedef struct {
     /* Whether the context has been built. */
     int built;
 
-    /* Blinding values used when computing (n-b)G + bG. */
-    secp256k1_scalar blind; /* -b */
-    secp256k1_gej initial;  /* bG */
+    /* Blinding values used when computing nG as (n-b)G + bG. */
+    secp256k1_scalar scalar_offset; /* -b */
+    secp256k1_ge final_point_add;  /* bG */
```
peterdettman, line 92: These comments are all a bit off. I would suggest a short explanation here and possibly defer to the fuller explanations in ecmult_gen_impl.h. e.g.:

> Values chosen such that n*G == comb(n + scalar_offset, G) + final_point_add. The latter expression lets us use scalar blinding and optimize the comb precomputation.

configure.ac
```diff
@@ -307,20 +306,31 @@ case $set_ecmult_window in
 esac
 
 # Set ecmult gen precision
-if test x"$req_ecmult_gen_precision" = x"auto"; then
-  set_ecmult_gen_precision=4
+if test x"$req_ecmult_gen_kb" = x"auto"; then
+  set_ecmult_gen_kb=86
```
peterdettman, line 310: I'm a little bit of the mind that 22 should be the default, but perhaps that's my old brain stuck in 64kB L1 world. Still, would it be reasonable to default to 22 for 32bit?

src/ecmult_gen.h
```diff
@@ -10,19 +10,90 @@
...
+#    define COMB_BLOCKS 128
+#    define COMB_TEETH 2
+#  else
+#    define COMB_BLOCKS 256
+#    define COMB_TEETH 1
+#  endif
+#else /* !defined(EXHAUSTIVE_TEST_ORDER) */
+/* Use (11, 6) as default configuration, which results in a 22 kB table. */
+#  ifndef COMB_BLOCKS
+#    define COMB_BLOCKS 11
```
peterdettman, line 55: Just noting that this isn't the same as the configure-driven default, not sure that it matters.

src/ecmult_gen.h
```diff
@@ -10,19 +10,90 @@
...
+#  error "COMB_BLOCKS must be in the range [1, 256]"
+#endif
+#if !(1 <= COMB_TEETH && COMB_TEETH <= 8)
+#  error "COMB_TEETH must be in the range [1, 8]"
+#endif
+
+/* The remaining COMB_* parameters are derived values, don't modify these. */
+/* - The distance between the teeth of the comb. */
+#define COMB_SPACING ((255 + COMB_BLOCKS * COMB_TEETH) / (COMB_BLOCKS * COMB_TEETH))
+/* - The number of bits covered by all the combs; must be at least 256. */
```
peterdettman, line 73: "by all the blocks" is maybe better.

src/ecmult_gen.h
```diff
@@ -10,19 +10,90 @@
...
+/* Range checks on the parameters. */
+#if !(1 <= COMB_BLOCKS && COMB_BLOCKS <= 256)
+#  error "COMB_BLOCKS must be in the range [1, 256]"
+#endif
+#if !(1 <= COMB_TEETH && COMB_TEETH <= 8)
+#  error "COMB_TEETH must be in the range [1, 8]"
+#endif
+
+/* The remaining COMB_* parameters are derived values, don't modify these. */
+/* - The distance between the teeth of the comb. */
```
peterdettman, line 71: "in each comb"?

peterdettman:
As an exercise I added a 2 blocks, 4 teeth configuration:

1. add {2,4} to CONFIGS table in precompute_ecmult_gen.c; nice and declarative.
2. make precompute_ecmult_gen and run to get new precomputed_ecmult_gen.c which correctly includes new {2,4} option.
3. modify configure.ac to support new "1" option for ecmult-gen-kb
4. configure --with-ecmult-gen-kb=1, confirm in libsecp256k1-config.h that COMB_BLOCKS, COMB_TEETH are correct.
5. make clean, make, tests to confirm it's working

This all worked without isues and was reasonably minimal effort. This particular example also satisfied me that there is no issue with combs covering exactly 256 bits.

configure.ac
real-or-random: With https://github.com/bitcoin-core/secp256k1/issues/929, it may be good idea to handle the logic in the preprocessor in the case of "auto". 

If you agree, I think it's both ok to do it in this PR, or postpone it to a separate PR that also cares about all other setting/parameters (which we should do anyway).

sipa:
@peterdettman FWIW, the easiest way of achieving the same would be:
1. Modify configure.ac to support a new option
2. Run configure with that option
3. make clean-precomp && make normally

Because precompute_ecmult_gen is automatically built & run when a precomputed_*.c file is missing, and because precompute_ecmult_gen will always include the table for whatever the configuration is.

configure.ac
sipa: I'm not convinced about what the defaults should be, but given that ecmult already defaults to 1 MiB precomputed table, I see little harm in picking an 86 KiB one here.

I agree we can easily revisit defaults later (especially in combination with #929).

src/ecmult_gen.h
sipa: Done.

src/ecmult_gen.h
sipa: Done.

src/ecmult_gen.h
sipa: I changed the configure-driver default to also be 22 kB; we can revisit the defaults later.

configure.ac
sipa: I changed the configure-driver default to also be 22 kB; we can revisit the defaults later.

src/ecmult_gen.h
sipa: Done.

peterdettman:
Having some confusion over possible bad interactions between modular reduction and signed-digit form, I wanted to reason things out in reverse.

For point P of order N, scalar s in [0,N), and an L-bit comb (2^L > N), let C(s,P) be the value calculated by our comb, which considers the bits of s, zero-extended to L bits as signed digits. Then:
```
    C(s,P) == (2.s - (2^L - 1)) * P
```
Therefore in order to use the comb to calculate k * G, we solve for the scalar `t` to use:
```
    k * G == C(t,G) == (2.t - (2^L - 1)) * G 
=>  k == 2.t - (2^L - 1) mod N
=>  t == (k + (2^L - 1))/2 mod N
```
Can we skip that halving and use G/2 instead?
```
    C(2t,G/2) == (4.t - (2^L - 1)) * G/2
              == (2.t - (2^L - 1)/2) * G
              != C(t,G) unless 2^L == 1 mod N
```
So no, but let's back up a step and ask what scalar `u` to use in the comb to calculate k * G as 2.k * (G/2):
```
    2.k * (G/2) == C(u,G/2) == (2.u - (2^L - 1)) * G/2
=>  2.k == 2.u - (2^L - 1) mod N
=>  u == k + (2^L - 1)/2 mod N
```
and since L is constant, the halving is now only needed in the precomputation.

Scalar blinding (using b * G == B):
```
    k * G == (k - b) * G + B == 2.(k - b) * (G/2) + B == C(k - b + (2^L - 1)/2, G/2) + B
```
where `-b + (2^L - 1)/2` can be precomputed.

sipa:
Nice, that's much better explained than my current comments. I'll try to include it.

peterdettman:
Although the above satisfies me mathematically, I would like to see an explicit test case where the comb_offset (2^L - 1)/2 causes a modular reduction (relative to k-b). e.g. arrange for k-b == N + 1 - comb_offset. I hope that's not too painful, but otherwise random testing seems unlikely to hit such a case.

sipa:
@peterdettman Making this change causes instant failure during the tests, at least:

```diff
--- a/src/ecmult_gen_impl.h
+++ b/src/ecmult_gen_impl.h
@@ -78,7 +78,7 @@ static void secp256k1_ecmult_gen(const secp256k1_ecmult_gen_context *ctx, secp25
      */
 
     /* Compute the scalar (gn + ctx->scalar_offset). */
-    secp256k1_scalar_add(&tmp, &ctx->scalar_offset, gn);
+    CHECK(!secp256k1_scalar_add(&tmp, &ctx->scalar_offset, gn));
     /* Convert to recoded array. */
     for (i = 0; i < 8; ++i) {
         recoded[i] = secp256k1_scalar_get_bits(&tmp, 32 * i, 32);
```

Including a test case that results in `tmp=0` or `tmp=1` or so may be useful though.

peterdettman:
scalar_offset includes the randomly-distributed blind, so there will be modular reductions. This isn't concerning at all because we can reason about the blind independently of the comb (just input offset + output offset).

However the comb offset is small: 2^(COMB_BITS-256) * (2^256 - N). So I would like a test that involves the comb_offset itself causing a modular reduction. The math checks out of course, but an explicit test can't hurt.

peterdettman:
Actually, I think (2^L - 1)/2 mod N is only small (128 bits) if L == 256.

sipa:
For the 43/6/1 configuration it is 0x80000000000000000000000000000001e7f9b4a5f9130fa66044722cc7ae9e1e
For the 11/6/4 configuration it is 0x8000000000000000000000000000000987e0873ddd5f4e3fe1563adfe6691698

peterdettman:
> Including a test case that results in `tmp=0` or `tmp=1` or so may be useful though.

So, these and maybe -1 would be enough. I will check a 256-bit config manually if they are added.

sipa:
@peterdettman I've incorporated your derivation in the comments in ecmult_gen_impl.h, and added a test case for recoded={-1,0,1}.

configure.ac
```diff
@@ -177,13 +177,12 @@ AC_ARG_WITH([ecmult-window], [AS_HELP_STRING([--with-ecmult-window=SIZE|auto],
...
 
-AC_ARG_WITH([ecmult-gen-precision], [AS_HELP_STRING([--with-ecmult-gen-precision=2|4|8|auto],
-[Precision bits to tune the precomputed table size for signing.]
-[The size of the table is 32kB for 2 bits, 64kB for 4 bits, 512kB for 8 bits of precision.]
-[A larger table size usually results in possible faster signing.]
-["auto" is a reasonable setting for desktop machines (currently 4). [default=auto]]
+AC_ARG_WITH([ecmult-gen-kb], [AS_HELP_STRING([--with-ecmult-gen-kb=2|22|86|auto],
+[The size of the precomputed table for signing in multiples of 1024 bytes (on typical platforms).]
+[Larger values result in possibly better signing/keygeneration performance at the cost of a larger table.]
+["auto" is a reasonable setting for desktop machines (currently 86). [default=auto]]
```
peterdettman, line 183: 86 -> 22?

src/ecmult_gen.h
```diff
@@ -10,19 +10,94 @@
...
-#define ECMULT_GEN_PREC_N(bits) (256 / bits)
 
 typedef struct {
     /* Whether the context has been built. */
     int built;
 
-    /* Blinding values used when computing (n-b)G + bG. */
-    secp256k1_scalar blind; /* -b */
-    secp256k1_gej initial;  /* bG */
+    /* Values chosen such that n*G == comb(n + scalar_offset, G) + final_point_add.
```
peterdettman, line 92: I guess it should be G/2 in the comb now?

sipa: FIxed. I though of "comb()" here as a more abstract "the comb multiplication algorithm", but I've now just aligned the ecmult_gen.h and ecmult_gen_impl.h descriptions (also switched to other from `c()` to `comb()`).

configure.ac
sipa: Fixed.

sipa:
I've worked on an additional change that introduces a COMB_RANGE which is normally 256, but in exhaustive test mode corresponds to the number of bits in EXHAUSTIVE_TEST_ORDER. Then COMB_BITS only has to cover COMB_RANGE etc, instead of always being at least 256.

And I'm seeing a suspicious failure in certain configurations with that (even making sure that the precomputed table only contains non-infinity points).

I'll retry this approach to make sure I haven't missed something.

sipa:
I added a commit that permits COMB_BITS < 256 in exhaustive test mode. However, it doesn't work in a lot of configurations,  and I don't understand what's causing it.

Here is a list of (blocks teeth spacing) tuples and whether they work (for both order 13 and 199): [groups.txt](https://github.com/bitcoin-core/secp256k1/files/7811096/groups.txt)
.

sipa:
Update: it appears that `EXHAUSTIVE_TEST_ORDER < 2**(BLOCKS * TEETH * (SPACING - 1))` perfectly predicts which configurations work.

sipa:
Final update: I was being very dumb, and precompute_ecmult_gen just had `spacing = (COMB_RANGE + blocks * teeth) / (block * teeth)` hardcoded, leading to an inconsistency between the table and the actual multiplication code.

False alarm.

src/ecmult_gen_compute_table_impl.h
```diff
@@ -1,83 +1,94 @@
...
+         * powers of two times G being negative. */
+        secp256k1_gej_neg(&vs[vs_pos++], &sum);
+        /* And then teeth-1 times "double" the range of i values for which the table
+         * is computed: in each iteration, double the table by taking an existing
+         * table entry and adding ds[tooth]. */
+        for (tooth = 0; tooth < teeth - 1; ++tooth) {
+            size_t stride = ((size_t)1) << tooth;
+            size_t index;
+            for (index = 0; index < stride; ++index, ++vs_pos) {
+                secp256k1_gej_add_var(&vs[vs_pos], &vs[vs_pos - stride], &ds[tooth], NULL);
```
stratospher, line 74: da73fd9: so for vs_pos = [256, 263] - since those bits are always set to 0. their possible combinations are computed and kept but never end up being used? also wondering if those big powers of 2 (maybe something like - 2\*\*263 - 2\*\*259 - 2\*\*255 - 2\*\*251 - 2\*\*247 - 2\*\*243) are safe to add in `secp256k1_gej_add_var`.

(talking about default configuration with blocks=11, teeth=6)

stratospher: oh makes sense.

real-or-random (commented):
More to follow, I'm in the middle of the main commit...

src/ecmult_gen_compute_table_impl.h
```diff
@@ -1,83 +1,95 @@
...
-        r = secp256k1_fe_set_b32_limit(&nums_x, nums_b32);
-        (void)r;
-        VERIFY_CHECK(r);
-        r = secp256k1_ge_set_xo_var(&nums_ge, &nums_x, 0);
-        (void)r;
-        VERIFY_CHECK(r);
-        secp256k1_gej_set_ge(&nums_gej, &nums_ge);
-        /* Add G to make the bits in x uniformly distributed. */
-        secp256k1_gej_add_ge_var(&nums_gej, &nums_gej, gen, NULL);
+    /* u is the running power of two times gen we're working with, initially gen/2. */
```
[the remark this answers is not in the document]
real-or-random: Yeah, okay, it had simply occurred to me that this piece is easy to check without adding a dependency on `ecmult`... But checking that particular condition is not too interesting. So never mind, ignore my comment.

But having said this, do you think it makes sense to add some code in tests.c that does some basic checks on the generated table values? The advantage is that `ecmult`/`ecmult_const` wouldn't add a dependency?

src/ecmult_gen_impl.h
```diff
@@ -29,38 +29,154 @@ static void secp256k1_ecmult_gen_context_clear(secp256k1_ecmult_gen_context *ctx
...
+     *   mask(0)  = 2^0   + 2^4   + 2^8   + 2^12  + 2^16  + 2^20,
+     *   mask(1)  = 2^24  + 2^28  + 2^32  + 2^36  + 2^40  + 2^44,
+     *   mask(2)  = 2^48  + 2^52  + 2^56  + 2^60  + 2^64  + 2^68,
+     *   ...
+     *   mask(10) = 2^240 + 2^244 + 2^248 + 2^252 + 2^256 + 2^260
+     *
+     * Imagine we have a table(b,m) function which can look up, given b and
+     * m=(recoded & mask(b)), the sum of (2*d_i-1)*2^(i-1)*G for all bit positions
+     * i set in mask(b). In our example, table(0, 1 + 2^8 + 2^20) would be equal to
+     * (2^-1 - 2^3 + 2^7 - 2^11 - 2^15 + 2^19)*G.
```
real-or-random, line 126: Hm, this took me a while to follow. Let me explain in my own words. So you assume that the relevant bits of `d` are `1...0...0...1...0...1`, or `1...0...1...0...0...1` in reverse, which is `1 + 2^8 + 2^20`, which determines the positive summands in the G factor (decremented because the formula says so). Is this right? Want to elaborate in the comment?

src/ecmult_gen_impl.h
```diff
@@ -29,38 +29,154 @@ static void secp256k1_ecmult_gen_context_clear(secp256k1_ecmult_gen_context *ctx
...
+     *
+     * With that, we can rewrite R as:
+     *   1*(table(0, recoded & mask(0)) + table(1, recoded & mask(1)) + ...)
+     * + 2*(table(0, (recoded/2) & mask(0)) + table(1, (recoded/2) & mask(1)) + ...)
+     * + 4*(table(0, (recoded/4) & mask(0)) + table(1, (recoded/4) & mask(1)) + ...)
+     * + ...
+     * + 2^(COMB_SPACING-1)*(table(0, (recoded/2^(COMB_SPACING-1)) & mask(0)) + ...)
+     * + ctx->ge_offset.
```
real-or-random, line 134: nit: indent the display formula

src/ecmult_gen_impl.h
```diff
@@ -29,38 +29,154 @@ static void secp256k1_ecmult_gen_context_clear(secp256k1_ecmult_gen_context *ctx
...
+     * of that line and the doubles the result before proceeding to the next line.
+     * In pseudocode:
+     *   R = infinity
+     *   for comb_off in range(COMB_SPACING - 1, -1, -1):
+     *     for block in range(COMB_BLOCKS):
+     *       R += table(block, (recoded >> comb_off) & mask(block))
+     *     if comb_off > 0:
+     *       R = 2*R
+     *   R += ge_offset
+     *   return R
```
real-or-random, line 147: It may be a good idea to explain here why we `ge_offset` is added at the end, instead of using it as initial value for the variable R.

src/ecmult_gen_impl.h
```diff
@@ -29,38 +29,154 @@ static void secp256k1_ecmult_gen_context_clear(secp256k1_ecmult_gen_context *ctx
...
+     *
+     * The last question is how to implement the table(b,m) function. For any value of
+     * b, m=(recoded & mask(b)) can only take on at most 2^COMB_TEETH possible values
+     * (the last one may have fewer as there mask(b) may the curve order). So we could
+     * create COMB_BLOCK tables which contain a value for each such m value.
```
real-or-random, line 152: nit: grammar in parantheses

src/ecmult_gen_impl.h
```diff
@@ -29,38 +29,154 @@ static void secp256k1_ecmult_gen_context_clear(secp256k1_ecmult_gen_context *ctx
...
+     *
+     * Due to the fact that every table entry is a sum of positive and negative powers
+     * of two multiplied by G, every table will contains pairs of negated points:
+     * if all the masked bits in m flip, the table value is negated. We can exploit this
+     * to only store the first half of every table. If an entry from the second half is
+     * needed, we look up its bit-flipped version instead, and conditionally negate it.
```
real-or-random, line 158: nit: remove "conditionally"? (the entire clause is already conditional)

src/ecmult_gen_impl.h
```diff
@@ -29,38 +29,154 @@ static void secp256k1_ecmult_gen_context_clear(secp256k1_ecmult_gen_context *ctx
...
+             * together: bit (tooth) of bits = bit
+             * ((block*COMB_TEETH + tooth)*COMB_SPACING + comb_off) of recoded. */
+            uint32_t bits = 0, sign, abs, index, tooth;
+            for (tooth = 0; tooth < COMB_TEETH && bit_pos < 256; ++tooth) {
+                uint32_t bit = secp256k1_scalar_get_bits(&recoded, bit_pos, 1);
+                bits |= bit << tooth;
+                bit_pos += COMB_SPACING;
+            }
+
+            /* If the top bit of bits is 1, conditionally flip them all (corresponding
```
real-or-random, line 187: nit: same here, remove "conditionally" (or do I have this wrong?)

src/ecmult_gen_impl.h
```diff
@@ -29,38 +29,154 @@ static void secp256k1_ecmult_gen_context_clear(secp256k1_ecmult_gen_context *ctx
...
+        for (block = 0; block < COMB_BLOCKS; ++block) {
+            /* Gather the mask(block)-selected bits of recoded into bits. They're packed
+             * together: bit (tooth) of bits = bit
+             * ((block*COMB_TEETH + tooth)*COMB_SPACING + comb_off) of recoded. */
+            uint32_t bits = 0, sign, abs, index, tooth;
+            for (tooth = 0; tooth < COMB_TEETH && bit_pos < 256; ++tooth) {
+                uint32_t bit = secp256k1_scalar_get_bits(&recoded, bit_pos, 1);
+                bits |= bit << tooth;
+                bit_pos += COMB_SPACING;
+            }
```
real-or-random: IIUC this relies on `COMB_TEETH <= 32`. Want to `VERIFY_CHECK` this?

I guess it will be more natural to declare `tooth` as `int`, it's not involved in all the bit magic/masking. I thought the same is true for `index`, but it's compared against `abs`, so it's probably better to keep it.

On a related note, it may be a good idea to change the return value of `secp256k1_scalar_get_bits to` `uint32_t` ? It's currently `unsigned int`. We kind of assert in `assumptions.h` that this type is at least 32 bits, but I think the implicit conversion to `uint32_t` is unexpected for readers. Plus, only the `_t` types are guaranteed to be "sane" (2's complement, no padding bits, ...).  (Sorry, I haven't checked the other callers...)

real-or-random:
I forgot to say that all the comments I've just posted are in the main commit https://github.com/bitcoin-core/secp256k1/pull/1058/commits/62601920bee3b57456a130d1c10a4e7fc08f3866.

src/precompute_ecmult_gen.c
```diff
@@ -46,16 +47,29 @@ int main(int argc, char **argv) {
...
-        secp256k1_ge_storage* table = checked_malloc(&default_error_callback, blocks * points * sizeof(secp256k1_ge_storage));
-        secp256k1_ecmult_gen_compute_table(table, &secp256k1_ge_const_g, blocks, teeth);
+        if (config < sizeof(CONFIGS) / sizeof(*CONFIGS)) {
+            /* In all but the last iteration, output the configurations in CONFIGS. */
+            blocks = CONFIGS[config][0];
+            teeth = CONFIGS[config][1];
+            if (blocks == COMB_BLOCKS && teeth == COMB_TEETH) did_current_config = 1;
+        } else {
+            /* In the last iteration, output table for (COMB_BLOCKS, COMB_TEETH) if not
+             * already done. */
```
real-or-random, line 66: 84f5ce761ced5da6c9332279ff491b139e1510c7  Always generate tables for current (blocks,teeth) config 

I think this is fine for the table generation code in the end, but I smell a refactor here. This code is kind of abusing the loop. I think what you want is to extract the main part into a function and have something like this in main:
```C
for (config = 0; config < sizeof(CONFIGS) / sizeof(*CONFIGS); ++config) {
    blocks = CONFIGS[config][0];
    teeth = CONFIGS[config][1];
    if (blocks == COMB_BLOCKS && teeth == COMB_TEETH) did_current_config = 1;      
    print_table(fp, blocks, teeth); 
}
if (!did_current_config) {
    print_table(fp, COMB_BLOCKS, COMB_TEETH);
}
``` 
The name `print_table` is consistent with `src/precompute_ecmult.c`.

configure.ac
```diff
@@ -358,19 +357,25 @@ case $set_ecmult_window in
...
-  SECP_CONFIG_DEFINES="$SECP_CONFIG_DEFINES -DECMULT_GEN_PREC_BITS=$set_ecmult_gen_precision"
+case $set_ecmult_gen_kb in
+2)
+  SECP_CONFIG_DEFINES="$SECP_CONFIG_DEFINES -DCOMB_BLOCKS=2 -DCOMB_TEETH=5"
+  ;;
+22)
+  SECP_CONFIG_DEFINES="$SECP_CONFIG_DEFINES -DCOMB_BLOCKS=11 -DCOMB_TEETH=6"
+  ;;
+86)
+  SECP_CONFIG_DEFINES="$SECP_CONFIG_DEFINES -DCOMB_BLOCKS=43 -DCOMB_TEETH=6"
```
real-or-random, line 375: I think in the spirit of trying to make manual builds, i.e., without build system, easy (https://github.com/bitcoin-core/secp256k1/issues/929), it would be better to move the selection logic here to the C code, and have `ECMULT_GEN_KB` as the "exposed" configuration macro. Then configure, cmake, and manual builds will have the same interface. (We could still allow passing `COMB_BLOCKS` and `COMB_TEETH` directly as long as `ECMULT_GEN_KB` is not defined, to retain the possibility to select a non-standard config.)

This shouldn't hold up this PR. It could also be done in a follow-up PR, if you feel that's easier.

src/ecmult_gen_impl.h
real-or-random: Yes!

src/ecmult_gen_impl.h
```diff
@@ -29,38 +29,154 @@ static void secp256k1_ecmult_gen_context_clear(secp256k1_ecmult_gen_context *ctx
...
- * None of the resulting prec group elements have a known scalar, and neither do any of
- * the intermediate sums while computing a*G.
- * The prec values are stored in secp256k1_ecmult_gen_prec_table[i][n_i] = n_i * (PREC_G)^i * G + U_i.
- */
-static void secp256k1_ecmult_gen(const secp256k1_ecmult_gen_context *ctx, secp256k1_gej *r, const secp256k1_scalar *gn) {
-    int bits = ECMULT_GEN_PREC_BITS;
-    int g = ECMULT_GEN_PREC_G(bits);
-    int n = ECMULT_GEN_PREC_N(bits);
+/* Compute the scalar (2^COMB_BITS - 1) / 2. */
+static void secp256k1_ecmult_gen_scalar_diff(secp256k1_scalar* diff) {
```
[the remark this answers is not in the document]
real-or-random: Makes sense, maybe add this explanation to the docstring.

src/ecmult_gen_impl.h
```diff
@@ -200,8 +200,15 @@ static void secp256k1_ecmult_gen(const secp256k1_ecmult_gen_context *ctx, secp25
...
-                bits |= bit << tooth;
+                /* Instead of reading individual bits here to construct bits, build up
+                 * the result by xoring shifted reads together. In every iteration, one
+                 * additional bit is made correct, starting at the bottom. The bits
+                 * above that contain junk. This reduces leakage from single bits. See
+                 * https://www.usenix.org/system/files/conference/usenixsecurity18/sec18-alam.pdf
+                 */
+                uint32_t bitdata = recoded[bit_pos >> 5] >> (bit_pos & 0x1f);
+                bits &= ~(1 << tooth);
+                bits ^= bitdata << tooth;
```
real-or-random: Hm, this is difficult to analyze and review. 

If `(bit_pos & 0x1f)` happens to be 31, we still extract a single bit. Could it help to ensure that there's always some junk in the high bits of `bitdata`? (Not sure what's the best way to achieve this.)

--
Also, compilers are not helpful here: gcc keeps the xor, but clang 17.0.6 manages to move the xor out of the loop and replace by a mov.

To inspect this, try `objdump-nice .libs/libsecp256k1.so | less` and search for `bitdata`, where `objdump-nice` is a shell alias for `objdump --disassemble --wide --debugging --line-numbers --source-comment --visualize-jumps=extended-color --disassembler-options=intel --no-show-raw-insn --no-addresses`

`volatile` helps, but I'm not convinced:
 - `bits = *((volatile uint32_t*)&bits) ^ (bitdata << tooth);` keeps the xor, but moves `bits` from a register to memory. Not sure if this better in terms of leakage.
 - `bits ^= bitdata << *((volatile uint32_t*)&tooth);` also keeps the xor. This moves `tooth` to memory and prevents clang from unrolling the loop. 

--

By the way, the mitigation in OpenSSL is https://github.com/openssl/openssl/pull/6276, but looking at that commit doesn't seem particularly helpful in our case.

src/ecmult_gen_impl.h
real-or-random: > `src/ecmult_gen.h` has an `#error` for `COMB_TEETH > 8`. Should I repeat it?

Never mind, the code is anyway replaced in later commits, so no need to bother.

configure.ac
real-or-random: Makes sense, let me leave a link to  https://github.com/bitcoin-core/secp256k1/issues/189 here, just as a reminder.

src/ecmult_gen_compute_table_impl.h
```diff
@@ -85,6 +84,7 @@ static void secp256k1_ecmult_gen_compute_table(secp256k1_ge_storage* table, cons
     for (block = 0; block < blocks; ++block) {
         size_t index;
         for (index = 0; index < points; ++index) {
+            CHECK(!secp256k1_ge_is_infinity(&prec[block * points + index]));
```
real-or-random, line 87: note: the precompute binaries are built with `-DVERIFY` (it's nevertheless okay to keep this here, I guess)

src/ecmult_gen_impl.h
```diff
@@ -25,41 +25,201 @@ static int secp256k1_ecmult_gen_context_is_built(const secp256k1_ecmult_gen_cont
...
+    /* Outer loop: iterate over comb_off from COMB_SPACING - 1 down to 0. */
+    comb_off = COMB_SPACING - 1;
+    while (1) {
+        uint32_t block;
+        uint32_t bit_pos = comb_off;
+        /* Inner loop: for each block, add table entries to the result. */
+        for (block = 0; block < COMB_BLOCKS; ++block) {
+            /* Gather the mask(block)-selected bits of d into bits. They're packed
+             * together: bit (tooth) of bits = bit
+             * ((block*COMB_TEETH + tooth)*COMB_SPACING + comb_off) of d. */
```
real-or-random, line 201: nit:
```suggestion
            /* Gather the mask(block)-selected bits of d into bits. They're packed:
             * bits[tooth] = d[(block*COMB_TEETH + tooth)*COMB_SPACING + comb_off]. */
```

The took me a while to parse due to the linebreaks.

src/ecmult_gen_impl.h
real-or-random: See https://github.com/real-or-random/secp256k1/commit/50ef2da3df16e7e610098bd50cacb60e5264cf7c for an improved version. I've manually verified that gcc, clang, msvc actually use rots and don't optimize that away.

src/scalar_8x32_impl.h
```diff
@@ -62,23 +62,24 @@ SECP256K1_INLINE static void secp256k1_scalar_set_int(secp256k1_scalar *r, unsig
     SECP256K1_SCALAR_VERIFY(r);
 }
 
-SECP256K1_INLINE static unsigned int secp256k1_scalar_get_bits(const secp256k1_scalar *a, unsigned int offset, unsigned int count) {
+SECP256K1_INLINE static uint32_t secp256k1_scalar_get_bits(const secp256k1_scalar *a, unsigned int offset, unsigned int count) {
     SECP256K1_SCALAR_VERIFY(a);
+    VERIFY_CHECK(count > 0 && count <= 32);
```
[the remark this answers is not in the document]
real-or-random: you're right, I misremembered the surrounding style, sorry

src/ecmult_gen_impl.h
```diff
@@ -25,41 +25,216 @@ static int secp256k1_ecmult_gen_context_is_built(const secp256k1_ecmult_gen_cont
...
+/* Compute the scalar (2^COMB_BITS - 1) / 2, the difference between the gn argument to
+ * secp256k1_ecmult_gen, and the scalar whose encoding the table lookup bits are drawn
+ * from (before applying blinding). */
+static void secp256k1_ecmult_gen_scalar_diff(secp256k1_scalar* diff) {
+    int i;
+
+    /* Compute scalar -1/2. */
+    secp256k1_scalar neghalf = SECP256K1_SCALAR_CONST(0, 0, 0, 0, 0, 0, 0, 2);
+    secp256k1_scalar_inverse_var(&neghalf, &neghalf);
+    secp256k1_scalar_negate(&neghalf, &neghalf);
```
real-or-random, line 42: Optimized this here, if you want to cherry-pick: https://github.com/real-or-random/secp256k1/commit/b71226b71e02ae7c0e67ff9a9546933399032da8 

It may look overkill, but I don't think it is upon a closer look: This saves ~1.1 us on my machine (lazy benchmark with turboboost enabled), which is 40% in context_create with NULL seed (because that has an early return to skip the ecmult_gen). But also ecmult_gen is just ~10 us, so this will also save ~10% in context_randomize, which is not entirely unimportant.

real-or-random: Argh, ok, I see...  I'll try again later...

real-or-random: New attempt: https://github.com/real-or-random/secp256k1/commit/9054994691d9f037146988a6cb41ec7a0d90fe03

src/ecmult_gen_compute_table_impl.h
```diff
@@ -10,74 +10,96 @@
...
-    int i, j;
-
-    VERIFY_CHECK(g > 0);
-    VERIFY_CHECK(n > 0);
-
-    /* get the generator */
-    secp256k1_gej_set_ge(&gj, gen);
-
-    /* Construct a group element with no known corresponding scalar (nothing up my sleeve). */
+    /* u is the running power of two times gen we're working with, initially gen/2. */
```
real-or-random, line 28: Add `VERIFY_CHECK(points_total > 0)` here to tell GCC that the loop is at least executed once.

See https://github.com/real-or-random/secp256k1/commit/854723f1e30d0e7e71967d0c225645275d169883#diff-2d5b5da541c539e4be4fc1506b554895bd68ef20bf8ae17879c8af6c948dbe54 for working CI runs on MacOS GCC.

src/ecmult_gen_impl.h
real-or-random: yeah that's much simpler, I didn't remember `scalar_half`...

src/ecmult_gen_compute_table_impl.h
real-or-random: Yep, and I had to find this [GCC bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100417) to understand that GCC talks about the *memory pointed to by* `ret` (inside `checked_malloc`) being uninitialized, and not `ret` itself...

CHANGELOG.md
real-or-random, line 1: changelog:
 -  I think we should expand a bit on the difference to the old table sizes. (Not sure to which section that belongs then. Your current approach of using three sections is technically correct. But maybe just merge all of these items into one item under [Changed]. Then it's less scattered
 - nit: Want to give also the CMake options (see older changelog items for examples)?

real-or-random:
As part of testing this, I played around with the edge values for the config parameters. This led me to a series of minors nits/fixups in the code that handles the parameters, see https://github.com/real-or-random/secp256k1/tree/202112_sdmc, that you can picks. Most of these commits, if not all, should probably be squashed into your commits (and a regeneration of the precomputed files is necessary, this is why CI fails on my repo.)

Apart from that ACK. I did all the review and testing that I planned to do.

src/util.h
```diff
@@ -170,7 +170,10 @@ static SECP256K1_INLINE void *checked_malloc(const secp256k1_callback* cb, size_
 #define ALIGNMENT 16
 #endif
 
-#define ROUND_TO_ALIGN(size) ((((size) + ALIGNMENT - 1) / ALIGNMENT) * ALIGNMENT)
+/* ceil(x/y) for integers x >= 0 and y > 0. Here, / denotes rational division. */
+#define CEIL_DIV(x, y) (1 + ((x) - 1) / (y))
```
real-or-random, line 174: This is clearly better than my suggestion, but this can now underflow for unsigned x == 0, so the docs need to be changed:
```suggestion
/* ceil(x/y) for integers x > 0 and y > 0. Here, / denotes rational division. */
#define CEIL_DIV(x, y) (1 + ((x) - 1) / (y))
```

It's a bit sad that there's no real way to add a VERIFY_CHECK in a macro (or at least not one which is compatible with the way we use this macro in `WNAF_SIZE_BITS(bits, w)` as a compile time integer constant...). If you think that's better, we could drop that use, and make it a real function `size_t ceil_div(size_t, size_t)` with a VERIFY_CHECK instead, but no strong opinions here.

real-or-random (commented):
ACK mod nit

real-or-random (approved):
ACK 8ab2f5ee3344150e103050f578659592c2e5cc92

src/ecmult_gen.h
```diff
@@ -10,31 +10,122 @@
...
 #include "group.h"
 
-#ifndef ECMULT_GEN_PREC_BITS
-#  define ECMULT_GEN_PREC_BITS 4
+
+/* Configuration parameters for the signed-digit multi-comb algorithm:
+ *
+ * - COMB_BLOCKS is the number of blocks the input is split into. Each
+ *   has a corresponding table.
+ * - COMB_TEETH is the number of bits simultaneously covered by one table.
```
jonasnick, line 18: Not sure if it's helpful to mention separate tables here. In precompute_ecmult_gen.c it's all just in one big table:
```C
secp256k1_ge_storage* table = checked_malloc(&default_error_callback, blocks * points * sizeof(secp256k1_ge_storage));
```

jonasnick (commented):
Concept ACK

Can confirm a 13% speedup with default parameters on my dev laptop.

jonasnick (approved):
ACK 6a3362a1dcb556a84a0876e393f4d2f3fba34a32

real-or-random (approved):
reACK 4c341f89ab704205a89b5c5d404ed60f381f7c48

jonasnick (approved):
ACK 4c341f89ab704205a89b5c5d404ed60f381f7c48

stratospher (commented):
ACK 4c341f8. Did [these benchmarks](https://github.com/bitcoin-core/secp256k1/pull/1058#issuecomment-1002807283) and saw a 12.4% on gcc 13.2.0 and 11.5% on clang 15.0.0. Also summarised how the precomputed table generation works [here](https://github.com/stratospher/blogosphere/blob/main/sdmc.md) for future me :)

Issue #1061, opened 2022-01-03
call `fe_verfiy` for not normalized field element in `fe_set_b32`


siv2r:
https://github.com/bitcoin-core/secp256k1/blob/a1102b12196ea27f44d6201de4d25926a2ae9640/src/field_5x52_impl.h#L331-L339

Shouldn't the call to `secp256k1_fe_verify` come after the else block?

real-or-random:
This was changed in https://github.com/bitcoin-core/secp256k1/commit/34a67c773b0871e5797c7ab506d004e80911f120. I believe this is this intentionally to optimize the tests. Can you check if the tests take noticeably longer if the fe_verify is moved after the else block?

siv2r:
build commands:  `./autogen.sh && ./configure && make -j13`
Execution time (_in seconds_) of [`tests.c`](https://github.com/bitcoin-core/secp256k1/blob/master/src/tests.c) for three iterations (on 64-bit, i7-8750H). I used [`gettime_i64`](https://github.com/bitcoin-core/secp256k1/blob/master/src/bench.h#L15) function to measure the execution time. You can find my complete code [here](https://github.com/siv2r/secp256k1/commit/4c4de09b7c79c4cd2ff36fdaee799757fff93174).
| branch | min | avg | max |
|-----------|------|-------|-------|
| master (`fe_verify` inside if block) | 145.94 |146.26|146.72|
| bench-tests (`fe_verify` after else block)|145.74|146.04|146.27|

There isn't much difference in the execution time.

real-or-random:
Oh I think the `time` command would have been fine-grained enough for the tests. :) 

Yeah, I mean if the difference is below a second, I'd say it's reasonable to always do the verify.

siv2r:
Makes sense. The change required here is relatively small. I will check other parts of the code for similar issues and bundle them up in a PR.

siv2r:
> Oh I think the `time` command would have been fine-grained enough for the tests. :)

I didn't know such a command existed. I initially tried to use `clock_gettime()` for nanosecond precision..xD. It didn't work since `c89` does not seem to support it.

real-or-random:
Maybe this could be solved in https://github.com/bitcoin-core/secp256k1/pull/1062 ? I think it will be similar to some changes proposed there.

real-or-random:
This has been resolved by #1062.

PR #1062, opened 2022-01-04
Removes `_fe_equal_var`, and unwanted `_fe_normalize_weak` calls (in tests)

### LINKED ISSUE #946: Preconditions of fe_equal_var are weaker than documented
From IRC:

`<roconnor>` The documentation for secp256k1_fe_equal_var suggest both inputs have to have magnitude 1; however only the first input needs to have magnitude 1.
`<roconnor>` thus this call to fe_normalize_weak could be eliminated.
`<roconnor>` But it is so very not critical, I'm not sure it is even worth PRing.
`<real_or_random>` roconnor: same thing happens in secp256k1_gej_eq_x_var which is used in ECDSA verification. but ok, the speedup is still negligible 
`<real_or_random>` but I believe it's worth a PR (or an issue) because it would improve the internal docs

siv2r:
Fixes #946 and #1061

Changes:
- removes unwanted `fe_normalize_weak` calls to the second argument of `fe_equal`
- removes `fe_equal_var`

### DIFF
--- a/src/field.h
+++ b/src/field.h
@@ -176,12 +176,6 @@ static int secp256k1_fe_is_odd(const secp256k1_fe *a);
  */
 static int secp256k1_fe_equal(const secp256k1_fe *a, const secp256k1_fe *b);
 
-/** Determine whether two field elements are equal, without constant-time guarantee.
- *
- * Identical in behavior to secp256k1_fe_equal, but not constant time in either a or b.
- */
-static int secp256k1_fe_equal_var(const secp256k1_fe *a, const secp256k1_fe *b);
-
 /** Compare the values represented by 2 field elements, without constant-time guarantee.
  *
  * On input, a and b must be valid normalized field elements.
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -367,7 +367,7 @@ static int secp256k1_gej_eq_x_var(const secp256k1_fe *x, const secp256k1_gej *a)
 #endif
 
     secp256k1_fe_sqr(&r, &a->z); secp256k1_fe_mul(&r, &r, x);
-    return secp256k1_fe_equal_var(&r, &a->x);
+    return secp256k1_fe_equal(&r, &a->x);
 }
 
 static void secp256k1_gej_neg(secp256k1_gej *r, const secp256k1_gej *a) {
@@ -400,7 +400,7 @@ static int secp256k1_ge_is_valid_var(const secp256k1_ge *a) {
     secp256k1_fe_sqr(&y2, &a->y);
     secp256k1_fe_sqr(&x3, &a->x); secp256k1_fe_mul(&x3, &x3, &a->x);
     secp256k1_fe_add_int(&x3, SECP256K1_B);
-    return secp256k1_fe_equal_var(&y2, &x3);
+    return secp256k1_fe_equal(&y2, &x3);
 }
 
 static SECP256K1_INLINE void secp256k1_gej_double(secp256k1_gej *r, const secp256k1_gej *a) {
--- a/src/tests.c
+++ b/src/tests.c
@@ -2991,7 +2991,6 @@ static int check_fe_equal(const secp256k1_fe *a, const secp256k1_fe *b) {
     secp256k1_fe an = *a;
     secp256k1_fe bn = *b;
     secp256k1_fe_normalize_weak(&an);
-    secp256k1_fe_normalize_var(&bn);
     return secp256k1_fe_equal_var(&an, &bn);
 }
 
@@ -3740,9 +3739,9 @@ static void ge_equals_gej(const secp256k1_ge *a, const secp256k1_gej *b) {
     /* Check a.x * b.z^2 == b.x && a.y * b.z^3 == b.y, to avoid inverses. */
     secp256k1_fe_sqr(&z2s, &b->z);
     secp256k1_fe_mul(&u1, &a->x, &z2s);
-    u2 = b->x; secp256k1_fe_normalize_weak(&u2);
+    u2 = b->x;
     secp256k1_fe_mul(&s1, &a->y, &z2s); secp256k1_fe_mul(&s1, &s1, &b->z);
-    s2 = b->y; secp256k1_fe_normalize_weak(&s2);
+    s2 = b->y;
     CHECK(secp256k1_fe_equal_var(&u1, &u2));
     CHECK(secp256k1_fe_equal_var(&s1, &s2));
 }
@@ -4226,8 +4225,8 @@ static void test_pre_g_table(const secp256k1_ge_storage * pre_g, size_t n) {
         secp256k1_ge_from_storage(&q, &pre_g[i]);
         CHECK(secp256k1_ge_is_valid_var(&q));
 
-        secp256k1_fe_negate(&dqx, &q.x, 1); secp256k1_fe_add(&dqx, &gg.x); secp256k1_fe_normalize_weak(&dqx);
-        dqy = q.y; secp256k1_fe_add(&dqy, &gg.y); secp256k1_fe_normalize_weak(&dqy);
+        secp256k1_fe_negate(&dqx, &q.x, 1); secp256k1_fe_add(&dqx, &gg.x);
+        dqy = q.y; secp256k1_fe_add(&dqy, &gg.y);
         /* Check that -q is not equal to gg */
         CHECK(!secp256k1_fe_normalizes_to_zero_var(&dqx) || !secp256k1_fe_normalizes_to_zero_var(&dqy));
 
--- a/src/tests.c
+++ b/src/tests.c
@@ -2991,7 +2991,7 @@ static int check_fe_equal(const secp256k1_fe *a, const secp256k1_fe *b) {
     secp256k1_fe an = *a;
     secp256k1_fe bn = *b;
     secp256k1_fe_normalize_weak(&an);
-    return secp256k1_fe_equal_var(&an, &bn);
+    return secp256k1_fe_equal(&an, &bn);
 }
 
 static void run_field_convert(void) {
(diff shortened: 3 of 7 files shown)


siv2r:
<details>
<summary> make check results </summary>

```bash
make  check-am
make[1]: Entering directory '/home/siv2r/Projects/secp256k1'
make  check-TESTS
make[2]: Entering directory '/home/siv2r/Projects/secp256k1'
make[3]: Entering directory '/home/siv2r/Projects/secp256k1'
PASS: tests
PASS: exhaustive_tests
============================================================================
Testsuite summary for libsecp256k1 0.1.0-pre
============================================================================
# TOTAL: 2
# PASS:  2
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
make[3]: Leaving directory '/home/siv2r/Projects/secp256k1'
make[2]: Leaving directory '/home/siv2r/Projects/secp256k1'
make[1]: Leaving directory '/home/siv2r/Projects/secp256k1'
```

</details>

src/field.h
```diff
@@ -64,10 +64,13 @@ static int secp256k1_fe_is_zero(const secp256k1_fe *a);
 /** Check the "oddness" of a field element. Requires the input to be normalized. */
 static int secp256k1_fe_is_odd(const secp256k1_fe *a);
 
-/** Compare two field elements. Requires magnitude-1 inputs. */
+/** Compare two field elements. Requires both the inputs to have magnitude = 1.
+ *  i.e. a->magnitude = 1 and b->magnitude = 1 */
```
real-or-random, line 68: Just a nit but why not just
```suggestion
/** Compare two field elements. Requires both the inputs to have magnitude 1. */
```
(Hm, in fact it's `<= 1` even but that should go away when  #1001 is solved). 

I don't think it's necessary to phrase it twice (and `x->magnitude` is only there #ifdef VERIFY -- but ok people would understands what is meant. ;)) 

Same below.

real-or-random (commented):
Concept ACK

I think then we should add (#ifdef VERIFY) magnitude checks to secp256k1_fe_equal_var and secp256k1_fe_equal.

If you're up to it, I think the other field function could need an audit for similar checks.

src/field.h
siv2r: Sure, I will remove this additional comment. I wanted to be more explicit, but I forgot that `x->magnitude` is used only in the VERIFY build.

siv2r:
In `fe_sqrt`:
https://github.com/bitcoin-core/secp256k1/blob/a1102b12196ea27f44d6201de4d25926a2ae9640/src/field_impl.h#L132-L135
Here, the `secp256k1_fe_sqr` function ensures that `t1->magnitude = 1` but there is no guarantee for the magnitude of `a` to be one since it is passed into the function. Hence, some `fe_sqrt` tests are failing after adding magnitude checks inside `fe_equal`.

We could do `secp256k1_fe_normalize_weak(a)` but this changes the value of `a` and we don't want that right? Alternatively, we could call `fe_equal_var` instead of `fe_equal`. Is it necessary for `secp256k1_fe_sqrt` need to be constant time?

peterdettman:
@siv2r Regarding `_fe_sqrt`, there is an implicit magnitude restriction on `a`, since it is used as the input to `_fe_mul` and `_fe_sqr` at the top of the method. That magnitude restriction is `<= 8`. 8 is also a kind of global implicit limit for inputs where not otherwise stated (but we should definitely be working towards fully documenting and validating all such assumptions).

I think it might be better to just say that both inputs to the equals methods can be magnitude <= 8 too (indeed tied to the mul/sqr limit whatever it is or changes to). There's no real obstacle to the implementation(s) (just change the third argument to _fe_negate to 8) and no performance hit AFAICT.

peterdettman:
> Is it necessary for `secp256k1_fe_sqrt` to be of constant time?

It's only used in a single var-time caller, so technically it could be changed to sqrt_var, but preferably not.

real-or-random:
Apparently `secp256k1_fe_equal` and `secp256k1_fe_equal_var` are exactly identical and so they both only require that `a` has magnitude 1. So we should just get rid of `secp256k1_fe_equal_var` and `secp256k1_fe_equal` should have the magnitue restriction on a documented.

siv2r:
> It's only used in a single var-time caller

Oh, Okay. Then, creating a new function for `secp256k1_fe_sqrt_var` (while renaming the old one to `_fe_sqrt_const`) might be overkill.

> we should just document that `secp256k1_fe_equal` only requires the magnitude of `a` to be 1.

Yes, I think this might be a better approach since it keeps the functionality of both `fe_equal_var` and `fe_equal` similar. 

After this, I felt we needed to remove the `_fe_normalize_weak` calls before `_fe_equal`, but surprisingly `_fe_equal` is used rarely in the code. The only other place where it is used is:
https://github.com/bitcoin-core/secp256k1/blob/a1102b12196ea27f44d6201de4d25926a2ae9640/src/modules/extrakeys/tests_impl.h#L75-L77
Even here, the magnitude of `y` is two, so modifying the `_fe_sqrt` function (which I suggested) is a bad idea.

real-or-random:
> https://github.com/bitcoin-core/secp256k1/blob/a1102b12196ea27f44d6201de4d25926a2ae9640/src/modules/extrakeys/tests_impl.h#L75-L77

Hehe ok, and here it's basically an oversight. This is just test code, so `fe_equal_var` could be used instead. I mean, not that it matters.

siv2r:
addressed https://github.com/bitcoin-core/secp256k1/pull/1062#discussion_r778917127 and https://github.com/bitcoin-core/secp256k1/pull/1062#pullrequestreview-844742940 (detailed explanation in 08186e8).

siv2r:
I also benchmarked the `tests.c` file (using `time ./tests`) since a lot of `secp256k1_fe_verify` calls were added in this PR. These were the results (on 64-bit, i7-8750H) for three iterations:
| branch | iter1 | iter2 | iter3 |
|-----------|------|-------|-------|
| master |2min 26.21sec|2min 26.44sec|2min 26.68sec|
| this PR|2min 26.99sec|2min 26.91sec|2min 26.83sec|

real-or-random:
@peterdettman Good points. Do you think we should remove `fe_equal_var` entirely and maybe document in `fe_equal` that this is intentional?

peterdettman:
> @peterdettman Good points. Do you think we should remove `fe_equal_var` entirely and maybe document in `fe_equal` that this is intentional?

Yeah, pretty much. Until there's a performance-sensitive caller for whom a != b is the expected result, I don't see any value in `_fe_equal_var`. If such a caller does turn up, then there should be a fast path for inequality that doesn't even need the subtraction, so we'd rewrite it anyway.

siv2r:
addressed #1061

real-or-random:
@siv2r Do you plan to also remove the `fe_equal_var` in a further commit? Or should we do this in a separate PR?

(If you want to add it here, I think it's really ok to add it on top of the existing commits, no need to rewrite the existing commits).

siv2r:
Okay, I will remove `fe_equal_var` in a new commit (this PR).

siv2r:
- Removed `fe_equal_var` and documented the reason in `fe_equal`
- Replaced the `fe_equal_var` calls with `fe_equal`

src/field.h
```diff
@@ -64,12 +64,12 @@ static int secp256k1_fe_is_zero(const secp256k1_fe *a);
 /** Check the "oddness" of a field element. Requires the input to be normalized. */
 static int secp256k1_fe_is_odd(const secp256k1_fe *a);
 
-/** Compare two field elements. Requires magnitude-1 inputs. */
+/** Compare two field elements. Requires the first input to have magnitude equal to 1. 
+ *  The variable time counterpart of this function `secp256k1_fe_equal_var` was removed 
+ *  since it hits fast path only when the inputs are unequal. This unequal inputs
```
sipa, line 69: I think this is something you can document in the commit comment; it's weird to have it in the codebase. As in: there are many other things that aren't implemented because they're not worth it; they aren't exhaustively listed in the source code.

src/group_impl.h
```diff
@@ -241,7 +241,7 @@ static int secp256k1_gej_eq_x_var(const secp256k1_fe *x, const secp256k1_gej *a)
     secp256k1_fe r, r2;
     VERIFY_CHECK(!a->infinity);
     secp256k1_fe_sqr(&r, &a->z); secp256k1_fe_mul(&r, &r, x);
-    r2 = a->x; secp256k1_fe_normalize_weak(&r2);
+    r2 = a->x;
```
sipa, line 244: The `r2` variable isn't needed anymore; you can just call `return secp256k1_fe_equal_var(&r, &a->x);` directly.

sipa:
I missed the "field: add magnitude check and _fe_verify check for internal field APIs" commit being added here before writing #1066 (which duplicates some of its efforts, but also goes a lot further in a number of ways). Feel free to keep it in (and I'll rebase), but if you think #1066 covers everything you're doing here, perhaps it's easier to leave that for that PR.

src/group_impl.h
siv2r: Nice catch, Thanks!

siv2r:
No worries, I will rebase this PR.

real-or-random:
> No worries, I will rebase this PR.

Ok, merging #1066 took a while.... This PR here needs rebase now. :)

siv2r:
Rebased this PR. #1066 took care of all the magnitude and `fe_verify` checks so, this PR currently does:
- removes unwanted `fe_normalize_weak` calls to the second argument of `fe_equal`
- removes `fe_equal_var`

src/tests.c
```diff
@@ -2967,7 +2967,6 @@ static int check_fe_equal(const secp256k1_fe *a, const secp256k1_fe *b) {
     secp256k1_fe an = *a;
     secp256k1_fe bn = *b;
     secp256k1_fe_normalize_weak(&an);
-    secp256k1_fe_normalize_var(&bn);
```
siv2r, line 2970: Not sure why `normalize_var` was used on `bn`. I assumed it intended to set `bn`'s magnitude to 1 for the `fe_equal` call. So, I removed it. Please let me know if that's not the case, and I'll revert this. 

All the tests pass with this change.

real-or-random:
Nice. :)

> #1066 took care of all the magnitude and `fe_verify` checks so, this PR currently does:

Can you update the PR title and the initial comment, so that they reflect the current status?

real-or-random:
Needs rebase again, sorry, but should be trivial. :) 

Thanks for updating the PR title. Can you also update the [initial post/PR description](https://github.com/bitcoin-core/secp256k1/pull/1062#issue-1093823651) (e.g., replace the contents with https://github.com/bitcoin-core/secp256k1/pull/1062#issuecomment-1678833854 )? It's just cosmetic, but the text of will be added to the merge commit, so it makes sense to keep this updated.

siv2r:
Rebased and updated the description. I forgot about the merge commit including the description too. Thanks!

real-or-random (approved):
utACK 54058d16feaa431520029335e2d56252859d3260

jonasnick (commented):
> Until there's a performance-sensitive caller for whom a != b is the expected result, I don't see any value in _fe_equal_var.

I don't think we know what the expected result of the caller is. The difference seems to be negligible though.

ACK 54058d16feaa431520029335e2d56252859d3260

Issue #974, opened 2021-09-13
ECDSA verify accepts all-zero hash


paulmillr:
Hey folks,

ECDSA verify accepts all-zero hash aka (0, 0, 0 ....). Is this a valid behavior? Seems like it could enable fault attacks.

real-or-random:
>     3. `e == 0`, then `u1 == 0`, then `u1 * G` is invalid because you cannot multiply G by 0

Sure you can multiply G by 0. 0*G is the point at infinity.

paulmillr:
Which is invalid point.

elichai:
> Upstream allows it because in Bitcoin Core's use there is no way to get an all-0s hash (though there almost was, and this would've made coins trivial to steal), and preventing this would greatly complicate the C API.

(Here "upstream" meaning this library) by @apoelstra from https://github.com/rust-bitcoin/rust-secp256k1/pull/207#issuecomment-612685695

paulmillr:
@elichai 

As for "no way to get 0": libsecp is used everywhere. Someone will pass all-0 forged hashes which were not generated with sha256.

As for "complicate": just check if the scalar is 0 and early-return false.

real-or-random:
> Which is invalid point.

"invalid" according to what definition? It's a perfectly valid `secp256k1_ge`. What's the attack you have in mind?

sipa:
No, we're not going to change this.

1) It violates the spec. It says nothing about rejecting hash=0, or why the point at infinity would be invalid as an intermediary result. It **is** invalid as a public key, but that doesn't apply here.
2) It shouldn't matter. The hash has to be a hash in correct usage, and thus it won't ever be 0 (as that implies mounting a succesful preimage attack first).

paulmillr:
@sipa so, do all ed25519 libraries violate its spec? By being more strict on signature malleability?

sipa:
ed25519 is not ECDSA.

paulmillr:
@sipa I was talking about ed25519 as an example. See this blog post: https://hdevalence.ca/blog/2020-10-04-its-25519am
RFC8032 is not strict enough and it's pretty bad. Different implementations of ed25519 have different behavior in regard to strictness. So they all deviate from spec? But it's fine for ed25519, why it isn't ok for secp?

Yes, your thinking here matches mine -- it's unsafe API design, and since libsecp is embedded everywhere, in every language, someone will use the api incorrectly. 

[...]

Issue #757, opened 2020-06-10
Nonce functions


real-or-random:
At the moment you can call `secp256k1_ecdsa_sign` with parameters `noncefp == NULL` and `noncedata != NULL`. This basically says "give me whatever function you think you should be the default but feed it this data", which is not really meaningful. This will be relevant if we want to change the default in the future, which is not too unlikely I think.

I see two ways:
 * Either we simply disallow this call (`ARG_CHECK`), or
 * we provide the guarantee that any default function will accept and use some (fixed-length) string of additional entropy.

After the investigation into synthetic nonces in BIP340, I slightly prefer the second option because I don't want to discourage people from adding entropy while leaving the choice of the function to the library. What do you think?

jonasnick:
> I don't want to discourage people from adding entropy while leaving the choice of the function to the library.

I'd be surprised if people don't make use of this already and while it's not explicitly documented I don't think we would ever switch the default to a function that doesn't add noncedata entropy.

elichai:
> I'd be surprised if people don't make use of this already

In rust-secp we don't, if this will be documented I'll add a feature that will seed randomness into the ecdsa_sign nonce function,

bitcoin core actually uses that entropy as counter and does not seed it (https://github.com/bitcoin/bitcoin/blob/195822f1e05e2f36002c906667d4c639663f23b5/src/key.cpp#L215 (I might open a PR to make the counter start at a random point))

real-or-random:
@elichai but both rust-secp and core specify `secp256k1_nonce_function_rfc6979` instead of `null`.

> bitcoin core actually uses that entropy as counter and does not seed it (https://github.com/bitcoin/bitcoin/blob/195822f1e05e2f36002c906667d4c639663f23b5/src/key.cpp#L215 (I might open a PR to make the counter start at a random point))

If we believe in the results of BIP340, we may even want to have randomness there in every call and not only at the start. https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#default-signing

We should also propose to change the derivation function to the one used in BIP340, but that's a larger thing. @jonasnick Is this something we could change in the Schnorr PR? Maybe introduce the nonce derivation function in a separate PR and make it available for ECDSA already now?

jonasnick:
[...]

Issue #286, opened 2015-08-07
Create a release


shazow:
I'm trying to add a homebrew install recipe for secp256k1 (https://github.com/Homebrew/homebrew/pull/42605) but they require a specific tagged version to include in the mainline tap. Any chance we can tag some commit(s) with some version scheme?

sipa:
We're currently still considering this code experimental, and don't have releases or API compatibility yet.

This will hopefully change soon.

shazow:
Fair enough. Please ping the above issue when you're ready and we can update the recipe. :)

sipa:
I'll leave this issue open until then.

yegortimoshenko:
Or you could just cut a v0.x.x tag, as in [semantic versioning](http://semver.org).

dcousens:
@shazow you can't use a commit hash?

shazow:
@dcousens Correct. It's a homebrew requirement.

gmaxwell:
It's a good requirement.  You should not be shipping split out libsecp256k1. The software is not released, the API is highly unstable, etc. We will release when it's ready for that.

afk11:
I noticed this library is being built and shipped in debian/stretch. I've asked in #secp256k1 and #bitcoin-core-dev but no-one seems to be claiming it. Emailed the Debian Bitcoin Team to see what the story is.. https://packages.debian.org/stretch/libsecp256k1-0

sugarjig:
I tried to create a homebrew formula for libbitcoin-explorer, and it has a transitive dependency on this library. I imagine that this issue is blocking lots of other projects from creating homebrew formulae. It's been exactly 2 years since this issue was opened. Are we any closer to a tagged release?

afk11:
My update to the above is that whoever packages groestlecoin & bitcoin into debian (contact details are on the link above) is maintaining it somewhat, but I don't trust the build. 

For now can you specify a git commit when pulling in dependencies?

sugarjig:
@afk11 Unfortunately, no. The official Homebrew taps require a formula be based on a tagged release.

skywinder:
>We're currently still considering this code experimental, and don't have releases or API compatibility yet.
This will hopefully change soon.

For the experimental code is considering as release number = 0.
i.e. `0.0.1` etc. And minor versions and patch version can change according to normal versioning rules. What about this option?

real-or-random:
We really should have releases.

Is there anything that prevent us from doing this?

[...]

PR #1064, opened 2022-01-15
Modulo-reduce msg32 inside RFC6979 nonce fn to match spec. Fixes #1063

### LINKED ISSUE #1063: ECDSA does not conform to RFC6979 for messages > curve_order
[RFC6979 3.2.d](https://datatracker.ietf.org/doc/html/rfc6979#section-3.2) says:

```
K = HMAC_K(V || 0x00 || int2octets(x) || bits2octets(h1))
```

where bits2octets is, as per [RFC6979 2.3.4](https://datatracker.ietf.org/doc/html/rfc6979#section-2.3.4) **curve order modulo-reduced** message:

```
   The bits2octets transform takes as input a sequence of blen bits and
   outputs a sequence of rlen bits.  It consists of the following steps:

   1.  The input sequence b is converted into an integer value z1
       through the bits2int transform:

          z1 = bits2int(b)
   2.  z1 is reduced modulo q, yielding z2 (an integer between 0 and
       q-1, inclusive):

          z2 = z1 mod q

       Note that since z1 is less than 2^qlen, that modular reduction
       can be implemented with a simple conditional subtraction:
       z2 = z1-q if that value is non-negative; otherwise, z2 = z1.

   3.  z2 is transformed into a sequence of octets (a sequence of rlen
       bits) by applying int2octets.
```

The implementation's `sign` takes `msg32` — not modulo-reduced `msg`, and passes it forward.

https://github.com/bitcoin-core/secp256k1/blob/0559fc6e41b65af6e52c32eb9b1286494412a162/src/secp256k1.c#L476

Seems like a bug, which does not exist in [go-btcec](https://github.com/btcsuite/btcd/tree/master/btcec) etc.

paulmillr:


real-or-random (commented):
I think it's good to adhere to the spec.

If we want to change this, the right way will be to change it in `nonce_function_rfc6979` because the change you propose here will also affect callers who provide their custom nonce function. But that means we'll need to do another reduction in  `nonce_function_rfc6979`, change the nonce function API, or introduce strange  special casing. Hm, I don't know, not sure if it's worth the hassle.

Anyway, I think what we should do is to overhaul the ECDSA API (https://github.com/bitcoin-core/secp256k1/issues/974), and then we could provide a nice simple ECDSA function that uses a much faster nonce function than RFC6979.

paulmillr:
Is it okay to cut a new release with the change (or an alternative similar change in nonce function)?

Before the refactoring, which could take a long time.

I develop secp256k1 in js and it seems like there are some assumptions in libraries that test against particular vectors e.g message > curve order. Releasing a new libsecp version would allow to have some "real" vectors to test against, since it's considered a reference implementation.

real-or-random:
> Is it okay to cut a new release with the change (or an alternative similar change in nonce function)?

I'd be totally fine with adding a reduction to `nonce_function_rfc6979`. That's simple and the solution with the cleanest code (right now)  and while it makes signing slightly slowing, you anyway can't expect optimal performance from RFC6979 signing. How does that sound to others?

Well, releases are a separate issue... So far we never had a release (https://github.com/bitcoin-core/secp256k1/issues/286) but there's good progress recently (https://github.com/bitcoin-core/secp256k1/pull/1055).

sipa:
I think we should:
* Make `nonce_function_rfc6979` do the modulo reduction, to comply with the spec.
* We could make `nonce_function_rfc6979` also reduce the output, so that it doesn't cause a second RNG invocation if the result exceed the order. This isn't an observable difference however, if RFC6979 is a PRF.
* Add a new fast, sane, nonce function with the simplest possible semantics around edge cases (something based on the BIP340 nonce functie, I imagine).
* Optionally, make this new nonce function the default, but leave the rfc6979 one in place.

real-or-random:
> Add a new fast, sane, nonce function with the simplest possible semantics around edge cases (something based on the BIP340 nonce function, I imagine).

There's some discussion around this here: https://github.com/bitcoin-core/secp256k1/issues/757#issuecomment-644001840

src/secp256k1.c
```diff
@@ -423,6 +423,10 @@ static int nonce_function_rfc6979(unsigned char *nonce32, const unsigned char *m
    unsigned int offset = 0;
    secp256k1_rfc6979_hmac_sha256 rng;
    unsigned int i;
+   secp256k1_scalar msg;
+   unsigned char msgmod32[32];
+   secp256k1_scalar_set_b32(&msg, msg32, NULL);
+   secp256k1_scalar_get_b32(msgmod32, &msg);
    /* We feed a byte array to the PRNG as input, consisting of:
     * - the private key (32 bytes) and message (32 bytes), see RFC 6979 3.2d.
```
real-or-random, line 431:
```suggestion
    * - the private key (32 bytes) and reduced message (32 bytes), see RFC 6979 3.2d.
```

real-or-random (commented):
Looks good, can you squash?

paulmillr:
@real-or-random done

real-or-random (approved):
ACK 45f37b650635e46865104f37baed26ef8d2cfb97

siv2r:
ACK 45f37b6. The diff looks good. It reduces `msg32` to modulo curve order for rfc6979 nonce generation. All tests passed on my machine with `make check`.
<details>
<summary> make check result </summary>

```
make  check-am
make[1]: Entering directory '/home/siv2r/Projects/secp256k1'
make  check-TESTS
make[2]: Entering directory '/home/siv2r/Projects/secp256k1'
make[3]: Entering directory '/home/siv2r/Projects/secp256k1'
PASS: tests
PASS: exhaustive_tests
============================================================================
Testsuite summary for libsecp256k1 0.1.0-pre
============================================================================
# TOTAL: 2
# PASS:  2
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
make[3]: Leaving directory '/home/siv2r/Projects/secp256k1'
make[2]: Leaving directory '/home/siv2r/Projects/secp256k1'
make[1]: Leaving directory '/home/siv2r/Projects/secp256k1'
```

</details>

sipa:
utACK 45f37b650635e46865104f37baed26ef8d2cfb97

sipa:
Apparently this was detected to cause a divergence with the trezor firmware: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=43997

I haven't verified myself, but I assume it's because they use the unreduced msghash as well.

prusnak:
> I haven't verified myself, but I assume it's because they use the unreduced msghash as well.

We will address this in the upcoming PR (issue https://github.com/trezor/trezor-firmware/issues/2085)

PR #1066, opened 2022-01-29
Abstract out and merge all the magnitude/normalized logic

### LINKED ISSUE #1060: Separate magnitude/normalization/... checking/propagation from implementations
Right now, it is not exactly clear where the definition of magnitude/normalization rules belongs:
* On the one hand, both implementations individually track these values separately (and have their own rules)
* On the other hand, callers of the field code are expected to simultaneously satisfy all of them.

With #967 things may become even more complicated as it adds a "precomputed" dimension (=reduced to 4 64-bit limbs instead of 5), but lacks the need for magnitudes as a concern.

So I think that we should make these rules part of the generic field interface, rather than part of the implementations. To avoid duplication, there could be a field_debug.h that implements these by "wrapping" the underlying field. It could be done by giving all field types/functions a suffix, and if debug is in use, have the real field functions forward to those, and if not, just `#define fn fn_suffix` for all.

One reason for not doing this would be if we want to make group logic dependent on what the used field is (e.g. a field with laxer magnitude requirements could result in the group code emitting fewer normalization calls).

sipa:
Right now, all the logic for propagating/computing the magnitude/normalized fields in `secp256k1_fe` (when `VERIFY` is defined) and the code for checking it, is duplicated across the two field implementations. I believe that is undesirable, as these properties should purely be a function of the performed fe_ functions, and not of the choice of field implementation. This becomes even uglier with #967, which would copy all that, and even needs an additional dimension that would then need to be added to the two other fields. It's also related to #1001, which I think will become easier if it doesn't need to be done/reasoned about separately for every field.

This PR moves all logic around these fields (collectively called field verification) to implementations in field_impl.h, which dispatch to renamed functions in field_*_impl.h for the actual implementation.

Fixes #1060.


src/field_10x26.h
```diff
@@ -9,15 +9,22 @@
...
     uint32_t n[10];
-#ifdef VERIFY
-    int magnitude;
-    int normalized;
-#endif
+    /*
+     * Magnitude m implies n[i] <= 2*m*0x3FFFFFF for i=0..8, and
+     * n[9] <= 2*m*0x3FFFFF (a factor 16 smaller).
+     *
+     * Normalized requires magnitude 1, all n[i] <= 0x3FFFFFF, and
```
jonasnick, line 24: "requires magnitude <= 1"?

src/field_10x26.h
```diff
@@ -9,15 +9,22 @@
...
      * where p = 2^256 - 0x1000003D1
      */
     uint32_t n[10];
-#ifdef VERIFY
-    int magnitude;
-    int normalized;
-#endif
+    /*
+     * Magnitude m implies n[i] <= 2*m*0x3FFFFFF for i=0..8, and
+     * n[9] <= 2*m*0x3FFFFF (a factor 16 smaller).
```
jonasnick, line 22: nit: that wording sounds to me like `0x3FFFFF*16 = 0x3FFFFFF` but that's false.

jonasnick (commented):
PR looks good to me otherwise. Net negative diff :tada:

src/field_10x26_impl.h
```diff
@@ -1081,23 +1081,13 @@ static void secp256k1_fe_from_signed30(secp256k1_fe *r, const secp256k1_modinv32
...
-#endif
 }
 
 static void secp256k1_fe_to_signed30(secp256k1_modinv32_signed30 *r, const secp256k1_fe *a) {
     const uint32_t M30 = UINT32_MAX >> 2;
     const uint64_t a0 = a->n[0], a1 = a->n[1], a2 = a->n[2], a3 = a->n[3], a4 = a->n[4],
                    a5 = a->n[5], a6 = a->n[6], a7 = a->n[7], a8 = a->n[8], a9 = a->n[9];
 
-#ifdef VERIFY
-    VERIFY_CHECK(a->normalized);
```
robot-dreams, line 1225: (No action needed) I agree this can be removed because `fe_to_signed30` is static and only ever called after `fe_normalize{_var}`; similarly with `fe_to_signed62` below.

src/field_impl.h
```diff
@@ -336,4 +336,32 @@ SECP256K1_INLINE static void secp256k1_fe_from_storage(secp256k1_fe *r, const se
...
+
+static void secp256k1_fe_impl_inv_var(secp256k1_fe *r, const secp256k1_fe *x);
+SECP256K1_INLINE static void secp256k1_fe_inv_var(secp256k1_fe *r, const secp256k1_fe *x) {
+#ifdef VERIFY
+    int input_is_zero = secp256k1_fe_normalizes_to_zero(x);
+#endif
+    secp256k1_fe_impl_inv_var(r, x);
+#ifdef VERIFY
+    VERIFY_CHECK(secp256k1_fe_normalizes_to_zero(r) == input_is_zero);
+    r->magnitude = x->magnitude > 0;
```
robot-dreams, line 363: Before this PR, how could `r->magnitude` have ended up as 0?

src/field_impl.h
```diff
@@ -336,4 +336,32 @@ SECP256K1_INLINE static void secp256k1_fe_from_storage(secp256k1_fe *r, const se
 #endif
 }
 
+static void secp256k1_fe_impl_inv(secp256k1_fe *r, const secp256k1_fe *x);
+SECP256K1_INLINE static void secp256k1_fe_inv(secp256k1_fe *r, const secp256k1_fe *x) {
+#ifdef VERIFY
+    int input_is_zero = secp256k1_fe_normalizes_to_zero(x);
```
robot-dreams, line 344: Before this PR, the `secp256k1_fe_normalizes_to_zero` was done on an intermediate result (rather than the input `x`); would moving this check before the start of `impl_inv` change the behavior?

robot-dreams (commented):
Concept ACK

The process you used (rename `fe_` to `fe_impl`, move verification into new `fe_` functions in `field_impl.h`) makes sense.

I don't see any other relevant `#ifdef VERIFY` in `field_10x26_impl.h` or `field_5x52_impl.h` so it seems like you covered everything.

I just have a few clarifying questions, and then I want to do one final pass (maybe there's (1) some easy semi-automated way to check/reproduce the changes or (2) some more macros that can make `field_impl.h` shorter).

sipa:
Made some significant changes here, and addressed some of the comments.

src/field_impl.h
sipa: It couldn't; the new logic is slightly more precise.

src/field.h
```diff
@@ -240,11 +242,18 @@ static void secp256k1_fe_sqr(secp256k1_fe *r, const secp256k1_fe *a);
  */
 static int secp256k1_fe_sqrt(secp256k1_fe *r, const secp256k1_fe *a);
 
-/** Sets a field element to be the (modular) inverse of another. Requires the input's magnitude to be
- *  at most 8. The output magnitude is 1 (but not guaranteed to be normalized). */
+/** Compute the modular inverse of a field element.
+ *
+ * On input, a must be a valid field element; r need not be initialized.
+ * Performs {r = 1/a}.
```
robot-dreams, line 248:
```suggestion
 * Performs {r = 1/a} if a is nonzero, {r = 0} otherwise.
```

src/field_10x26_impl.h
```diff
@@ -331,15 +331,7 @@ static void secp256k1_fe_impl_get_b32(unsigned char *r, const secp256k1_fe *a) {
     r[31] = a->n[0] & 0xff;
 }
 
-SECP256K1_INLINE static void secp256k1_fe_negate(secp256k1_fe *r, const secp256k1_fe *a, int m) {
-#ifdef VERIFY
-    VERIFY_CHECK(a->magnitude <= m);
-    secp256k1_fe_verify(a);
-    VERIFY_CHECK(0x3FFFC2FUL * 2 * (m + 1) >= 0x3FFFFFFUL * 2 * m);
```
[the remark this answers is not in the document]
robot-dreams: Oh, I guess you no longer need these now that you have `secp256k1_fe_verify(a)` together with `a->magnitude <= m <= 31`?

src/field_impl.h
robot-dreams: Makes sense now, I didn't realize before that `secp256k1_fe_impl_inv{_var}` returns zero if the input was zero.

robot-dreams (commented):
New approach makes a lot of sense and commits are very well structured. Looks good overall, I just have a few cosmetic comments (feel free not to apply all of them).

real-or-random:
Concept ACK 

But is there you reason why you prefer to do this before #1001? (@peterdettman  and I argued to in #1060 that it may make sense to solve #1001.)

I haven't looked at the details here but I think this will non-trivially conflict with https://github.com/bitcoin-core/secp256k1/pull/1062. @sipa can you have a look and suggest how we should move forward?

src/field_10x26.h
sipa: This has been changed significantly; the rule that normalized implies magnitude 0 or 1 has now been added to field.h itself (it's in fact a rule that must hold for all users of field.h, independent of which field implementation was used).

src/field_10x26.h
sipa: Done something like this (and now also placed it in the right file).

src/field_10x26_impl.h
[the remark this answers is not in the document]
sipa: I've reinstituted them; they're certainly implied, but still clarify why the logic below is legal.

src/field.h
```diff
@@ -7,23 +7,36 @@
...
+ * fixed or known at compile time, and do not depend on the chosen field
+ * implementation. Despite that, what these properties actually entail for
+ * the field representation values depends on the chosen field implementation.
+ * See the field_*.h files included below for details. They are:
+ * - magnitude: an integer in [0,32]
+ * - normalized: 0 or 1; normalized=1 implies magnitude <= 1.
+ *
+ * In VERIFY mode, they are materialzed explicitly as fields in the struct,
+ * allowing run-time verification of these properties. In that case, the
+ * field implementation also provides a secp256k1_fe_check routine to
```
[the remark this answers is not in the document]
sipa: Done.

src/tests.c
```diff
@@ -6733,7 +6733,7 @@ void fe_cmov_test(void) {
     static const secp256k1_fe one = SECP256K1_FE_CONST(0, 0, 0, 0, 0, 0, 0, 1);
     static const secp256k1_fe max = SECP256K1_FE_CONST(
         0xFFFFFFFFUL, 0xFFFFFFFFUL, 0xFFFFFFFFUL, 0xFFFFFFFFUL,
-        0xFFFFFFFFUL, 0xFFFFFFFFUL, 0xFFFFFFFFUL, 0xFFFFFFFFUL
+        0xFFFFFFFFUL, 0xFFFFFFFFUL, 0xFFFFFFFEUL, 0xFFFFFC2EUL
```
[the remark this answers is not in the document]
sipa: I've instead added a commit that actually sets the magnitude/normalized fields correctly for SECP256K1_FE_CONST, instead of assuming ",1,1".

src/field_impl.h
```diff
@@ -135,4 +136,19 @@ static int secp256k1_fe_sqrt(secp256k1_fe *r, const secp256k1_fe *a) {
...
+static void secp256k1_fe_impl_verify(const secp256k1_fe *a);
+static void secp256k1_fe_verify(const secp256k1_fe *a) {
+    /* Magnitude between 0 and 32. */
+    int r = (a->magnitude >= 0) & (a->magnitude <= 32);
+    /* Normalized is 0 or 1. */
+    r &= (a->normalized == 0) | (a->normalized == 1);
+    /* If normalized, magnitude must be 0 or 1. */
+    if (a->normalized) r &= (a->magnitude <= 1);
+    VERIFY_CHECK(r == 1);
+    /* Invoke field specific checks. */
```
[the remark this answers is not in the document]
sipa: Fixed.

src/field.h
```diff
@@ -7,23 +7,36 @@
...
+ * determine what operations are permitted on it. These are purely a function
+ * of what secp256k1_fe_ operations are applied on it, generally (implicitly)
+ * fixed or known at compile time, and do not depend on the chosen field
+ * implementation. Despite that, what these properties actually entail for
+ * the field representation values depends on the chosen field implementation.
+ * See the field_*.h files included below for details. They are:
+ * - magnitude: an integer in [0,32]
+ * - normalized: 0 or 1; normalized=1 implies magnitude <= 1.
+ *
+ * In VERIFY mode, they are materialzed explicitly as fields in the struct,
```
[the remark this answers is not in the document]
sipa: Done.

src/field.h
```diff
@@ -7,23 +7,36 @@
...
+ * allowing run-time verification of these properties. In that case, the
+ * field implementation also provides a secp256k1_fe_check routine to
+ * verify that these fields match the run-time value and internal consistency
+ * checks. */
```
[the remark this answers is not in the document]
sipa: Done (in next push).

src/field_5x52.h
```diff
@@ -9,15 +9,25 @@
...
+    /*
+     * Magnitude m requires:
+     *     n[i] <= 2 * m * (2^26 - 1) for i=0..8
+     *     n[9] <= 2 * m * (2^22 - 1)
+     *
+     * Normalized requires:
+     *     n[i] <= (2^26 - 1) for i=0..8
+     *     sum(i=0..9, n[i] << (i*26)) < p
+     *     (together these imply n[9] <= 2^22 - 1)
+     */
```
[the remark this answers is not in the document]
sipa: I did. Fixed.

src/field.h
```diff
@@ -59,10 +59,13 @@ static void secp256k1_fe_verify(const secp256k1_fe *a);
...
  * function call (even though presumably inlinable). */
+#  define secp256k1_fe_normalize secp256k1_fe_impl_normalize
 #endif /* defined(VERIFY) */
 
-/** Normalize a field element. This brings the field element to a canonical representation, reduces
- *  its magnitude to 1, and reduces it modulo field size `p`.
+/** Normalize a field element.
+ *
+ * On input, r must be a valid field element.
+ * On output, r represents the same value but has normalized=1 and magnitude=1.
```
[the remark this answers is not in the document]
sipa: So what I'm trying to do in this PR is define a generic interface for field elements, which has well-defined and verified rules at the field layer, independent of the implementation.

Field elements don't have an observable notion of "reduced modulo" - they are elements of the field GF(2^256 - 2^32 - 977). Reducing is a concept that only appears if you're representing those field elements as integers - which is what the field implementations do.

So what's happening here is that the interface only says "makes the field element have the normalized property". What "normalized" entails for the representation is up to the specific field implementation (where you'll find a comment that the sum of the limbs has to be below p).

I've tried to add a few more comments to explain this, but suggestions welcome if you think it isn't clear from the text yet.

src/field.h
```diff
@@ -84,11 +86,17 @@ static void secp256k1_fe_normalize_weak(secp256k1_fe *r);
  */
 static void secp256k1_fe_normalize_var(secp256k1_fe *r);
 
-/** Verify whether a field element represents zero i.e. would normalize to a zero value. */
+/** Determine whether r represents value 0.
```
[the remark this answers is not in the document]
sipa: Done (here and further) in the next push.

src/field.h
```diff
@@ -115,7 +116,14 @@ static void secp256k1_fe_set_int(secp256k1_fe *r, int a);
  */
 static void secp256k1_fe_clear(secp256k1_fe *a);
 
-/** Verify whether a field element is zero. Requires the input to be normalized. */
+/** Determine whether a represents value 0.
+ *
+ * On input a must be a valid normalized field element.
+ * Returns whether a  = 0 (mod p).
```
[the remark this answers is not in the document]
sipa: Done in next push.

src/field.h
```diff
@@ -148,7 +149,13 @@ static int secp256k1_fe_equal(const secp256k1_fe *a, const secp256k1_fe *b);
  */
 static int secp256k1_fe_equal_var(const secp256k1_fe *a, const secp256k1_fe *b);
 
-/** Compare two field elements. Requires both inputs to be normalized */
+/** Compare the values represented by 2 field elements.
```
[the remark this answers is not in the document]
sipa: Done in next push.

src/field_impl.h
```diff
@@ -265,6 +265,17 @@ SECP256K1_INLINE static void secp256k1_fe_negate(secp256k1_fe *r, const secp256k
...
 }
+
+static void secp256k1_fe_impl_mul_int(secp256k1_fe *r, int a);
+SECP256K1_INLINE static void secp256k1_fe_mul_int(secp256k1_fe *r, int a) {
+    secp256k1_fe_verify(r);
+    VERIFY_CHECK(a >= 0 && a <= 32);
+    VERIFY_CHECK(a*r->magnitude <= 32);
+    secp256k1_fe_impl_mul_int(r, a);
+    r->magnitude *= a;
+    r->normalized = 0;
```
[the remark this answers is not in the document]
sipa: I'd instead rather add an `VERIFY_CHECK(a > 1)`, because using this function with a=0 or a=1 is dumb.

src/field.h
```diff
@@ -197,7 +198,13 @@ static void secp256k1_fe_negate(secp256k1_fe *r, const secp256k1_fe *a, int m);
  */
 static void secp256k1_fe_mul_int(secp256k1_fe *r, int a);
 
-/** Adds a field element to another. The result has the sum of the inputs' magnitudes as magnitude. */
+/** Increment a field element by another.
+ *
+ * On input, r and a must be valid field elements. The sum of their magnitudes
+ * may not exceed 32.
```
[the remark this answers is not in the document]
sipa: Done in next push.

src/field.h
sipa: Doing something like this in the next push.

sipa:
@real-or-random 

> But is there you reason why you prefer to do this before #1001? (@peterdettman and I argued to in #1060 that it may make sense to solve #1001.)

The impetus for this is really that I want to bring #967 up-to-date and more production-ready, and doing it properly I think would be very unclear without a change like this (having details from the "old" fields leak into the verify checks of the new one etc).

I'll comment in #1060, but I think this PR makes it at least obvious in what places norm/mag are not compile-time constants.

> I haven't looked at the details here but I think this will non-trivially conflict with #1062. @sipa can you have a look and suggest how we should move forward?

It will. That's one of the reasons why I kept the changes for each function in separate commits. So I'm happy to rebase this after #1062, or otherwise incorporate the commits here.

sipa:
Rebased.

src/field_impl.h
```diff
@@ -131,4 +132,21 @@ static int secp256k1_fe_sqrt(secp256k1_fe *r, const secp256k1_fe *a) {
...
+static void secp256k1_fe_verify(const secp256k1_fe *a) { (void)a; }
+#else
+static void secp256k1_fe_impl_verify(const secp256k1_fe *a);
+static void secp256k1_fe_verify(const secp256k1_fe *a) {
+    /* Magnitude between 0 and 32. */
+    int r = (a->magnitude >= 0) & (a->magnitude <= 32);
+    /* Normalized is 0 or 1. */
+    r &= (a->normalized == 0) | (a->normalized == 1);
+    /* If normalized, magnitude must be 0 or 1. */
+    if (a->normalized) r &= (a->magnitude <= 1);
```
real-or-random, line 161: nit: Maybe make those individual VERIFY_CHECKs. That's better for debugging because then the error message will show you which line failed. 

same is true  for both instances of `secp256k1_fe_impl_verify`

src/field_impl.h
```diff
@@ -131,4 +132,21 @@ static int secp256k1_fe_sqrt(secp256k1_fe *r, const secp256k1_fe *a) {
...
+    VERIFY_CHECK(r == 1);
+    /* Invoke implementation-specific checks. */
+    secp256k1_fe_impl_verify(a);
+}
+#endif /* defined(VERIFY) */
```
real-or-random, line 408: `!defined(VERIFY)` ?! well in this case i don't know :D

src/field.h
```diff
@@ -45,6 +45,20 @@
...
+#define SECP256K1_FE_VERIFY_CONST(d7, d6, d5, d4, d3, d2, d1, d0) \
+    /* Magnitude is 0 for constant 0; 1 otherwise. */ \
+    , (((d7) | (d6) | (d5) | (d4) | (d3) | (d2) | (d1) | (d0)) != 0) \
+    /* Normalized is 1 unless sum(d_i<<(32*i) for i=0..7) exceeds field modulus. */ \
+    , (!(((d7) & (d6) & (d5) & (d4) & (d3) & (d2)) == 0xfffffffful && ((d1) == 0xfffffffful || ((d1) == 0xfffffffe && (d0 >= 0xfffffc2f)))))
+#else
+#define SECP256K1_FE_VERIFY_CONST(d7, d6, d5, d4, d3, d2, d1, d0)
+#endif
+
+/* SECP256K1_FE_CONST_INNER is provided by the field implementation */
```
real-or-random, line 59: nit: this is part of the interface, so it should have a `/**` comment that describes the semantics (incl. mag and normalized).

src/field.h
```diff
@@ -136,7 +137,11 @@ static void secp256k1_fe_clear(secp256k1_fe *a);
  */
 static int secp256k1_fe_is_zero(const secp256k1_fe *a);
 
-/** Check the "oddness" of a field element. Requires the input to be normalized. */
+/** Determine whether a (mod p) is odd.
+ *
+ * On input a must be a valid normalized field element.
```
real-or-random, line 159: nit: "On input, a" makes this much easier to parse for my brain. Maybe it's a good idea add commas after every "On input"/"On output"

src/field_5x52.h
```diff
@@ -29,11 +42,7 @@ typedef struct {
     ((uint64_t)(d6) >> 16) | (((uint64_t)(d7)) << 16) \
 }
 
-#ifdef VERIFY
-#define SECP256K1_FE_CONST(d7, d6, d5, d4, d3, d2, d1, d0) {SECP256K1_FE_CONST_INNER((d7), (d6), (d5), (d4), (d3), (d2), (d1), (d0)), 1, 1}
-#else
-#define SECP256K1_FE_CONST(d7, d6, d5, d4, d3, d2, d1, d0) {SECP256K1_FE_CONST_INNER((d7), (d6), (d5), (d4), (d3), (d2), (d1), (d0))}
-#endif
+#define SECP256K1_FE_CONST(d7, d6, d5, d4, d3, d2, d1, d0) {SECP256K1_FE_CONST_INNER((d7), (d6), (d5), (d4), (d3), (d2), (d1), (d0)) SECP256K1_FE_VERIFY_CONST((d7), (d6), (d5), (d4), (d3), (d2), (d1), (d0)) }
```
jonasnick, line 45: This is redundant with `FE_CONST` definition in field.h

src/field.h
```diff
@@ -28,116 +45,272 @@
...
  */
 static void secp256k1_fe_set_int(secp256k1_fe *r, int a);
 
-/** Sets a field element equal to zero, initializing all fields. */
+/** Set a field element to 0.
+ *
+ * On input, r does not need to be initialized.
+ * On output, r represents 0, is normalized and has magnitude 0.
+ */
 static void secp256k1_fe_clear(secp256k1_fe *a);
```
jonasnick, line 145: The argument is called `a`.

src/field_impl.h
```diff
@@ -131,4 +148,263 @@ static int secp256k1_fe_sqrt(secp256k1_fe *r, const secp256k1_fe *a) {
...
+    r->normalized = 0;
+    secp256k1_fe_verify(r);
+}
+
+static void secp256k1_fe_impl_clear(secp256k1_fe *a);
+SECP256K1_INLINE static void secp256k1_fe_clear(secp256k1_fe *a) {
+    a->magnitude = 0;
+    a->normalized = 1;
+    secp256k1_fe_impl_clear(a);
+}
```
jonasnick, line 229: should verify `a` after clearning?

src/field_impl.h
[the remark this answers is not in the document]
real-or-random: I think it's fine. Making normalization depend on the value may be a step into the wrong direction (it makes magnitude less static, though the `int` argument probably doesn't matter.) 

And I don't see a good reason to VERIFY_CHECK the input. `a` may be determined at run time. For example, we currently call this with `0` (determined at run time) in the tests. If you want to add the VERIFY_CHECK, you'd need to adjust `random_field_element_magnitude`.

src/field.h
```diff
@@ -230,8 +231,13 @@ static void secp256k1_fe_add(secp256k1_fe *r, const secp256k1_fe *a);
  */
 static void secp256k1_fe_mul(secp256k1_fe *r, const secp256k1_fe *a, const secp256k1_fe * SECP256K1_RESTRICT b);
 
-/** Sets a field element to be the square of another. Requires the input's magnitude to be at most 8.
- *  The output magnitude is 1 (but not guaranteed to be normalized). */
+/** Square a field element.
+ *
+ * On input, a must be a valid field element; r does not need to be initialized. The magnitude
+ * of a cannot exceed 8.
```
real-or-random, line 250: nit: s/cannot/must not

src/field.h
```diff
@@ -219,8 +220,14 @@ static void secp256k1_fe_mul_int(secp256k1_fe *r, int a);
  */
 static void secp256k1_fe_add(secp256k1_fe *r, const secp256k1_fe *a);
 
-/** Sets a field element to be the product of two others. Requires the inputs' magnitudes to be at most 8.
- *  The output magnitude is 1 (but not guaranteed to be normalized). */
+/** Multiply two field elements.
+ *
+ * On input, a and b must be valid field elements; r does not need to be initialized.
+ * r and a may point to the same object, but neither can be equal to b. The magnitudes
```
real-or-random, line 240: s/cannot/must not (maybe there are more)

src/field.h
```diff
@@ -240,11 +240,13 @@ static void secp256k1_fe_mul(secp256k1_fe *r, const secp256k1_fe *a, const secp2
...
-/** If a has a square root, it is computed in r and 1 is returned. If a does not
- *  have a square root, the root of its negation is computed and 0 is returned.
- *  The input's magnitude can be at most 8. The output magnitude is 1 (but not
- *  guaranteed to be normalized). The result in r will always be a square
- *  itself. */
+/** Compute a square root of a field element.
+ *
+ * On input, a must be a valid field element with magnitude<=8; r need not be initialized.
+ * Performs {r = sqrt(a)} or {r = sqrt(-a)}, whichever exists. The resulting value
+ * represented by r will be a square itself. R and a cannot point to the same object.
```
real-or-random, line 260: s/R/r

add a `restrict` ? :)

src/field_impl.h
```diff
@@ -341,6 +341,28 @@ SECP256K1_INLINE static void secp256k1_fe_from_storage(secp256k1_fe *r, const se
...
+    secp256k1_fe_impl_inv(r, x);
+    r->magnitude = x->magnitude > 0;
+    r->normalized = 1;
+    VERIFY_CHECK(secp256k1_fe_normalizes_to_zero(r) == input_is_zero);
+    secp256k1_fe_verify(r);
+}
+
+static void secp256k1_fe_impl_inv_var(secp256k1_fe *r, const secp256k1_fe *x);
+SECP256K1_INLINE static void secp256k1_fe_inv_var(secp256k1_fe *r, const secp256k1_fe *x) {
+    int input_is_zero = secp256k1_fe_normalizes_to_zero(x);
```
real-or-random, line 368: `#ifdef VERIFY` here? well ok, I guess dead store elimation can handle it.

src/field.h
```diff
@@ -28,116 +45,278 @@
...
- *  itself. */
-static int secp256k1_fe_sqrt(secp256k1_fe *r, const secp256k1_fe *a);
+/** Compute a square root of a field element.
+ *
+ * On input, a must be a valid field element with magnitude<=8; r need not be initialized.
+ * Performs {r = sqrt(a)} or {r = sqrt(-a)}, whichever exists. The resulting value
+ * represented by r will be a square itself. Variables r and a must not point to the same object.
+ * On output, r will have magnitude 1 but will not be normalized.
+ */
+static int secp256k1_fe_sqrt(secp256k1_fe * SECP256K1_RESTRICT r, const secp256k1_fe * SECP256K1_RESTRICT a);
```
jonasnick, line 269: Did you intentionally skipped adding a postcondition to the `sqrt()` code (fe_verify of the square root, maybe check the magnitude and normalization).

src/field.h
```diff
@@ -28,116 +45,278 @@
...
+ * On input, both r and a must be valid field elements. Flag must be 0 or 1.
+ * Performs {r = flag ? a : r}.
+ * On output, r's magnitude and normalized will equal a's in case of flag=1, unchanged otherwise.
+ */
 static void secp256k1_fe_cmov(secp256k1_fe *r, const secp256k1_fe *a, int flag);
 
 /** Halves the value of a field element modulo the field prime. Constant-time.
  *  For an input magnitude 'm', the output magnitude is set to 'floor(m/2) + 1'.
  *  The output is not guaranteed to be normalized, regardless of the input. */
 static void secp256k1_fe_half(secp256k1_fe *r);
```
jonasnick, line 315: This comment didn't get updated. How about:

```
/** Halve the value of a field element modulo the field prime in constant-time.
 *
 * On input, r must be a valid field element.
 * On output, r will be normalized and have magnitude floor(m/2) + 1 where m is
 * the magnitude of r on input.
 */
 ```

File: src/field.h [hunk 33]
```diff
@@ -28,116 +45,278 @@
...
 
-/** Sets each limb of 'r' to its upper bound at magnitude 'm'. The output will also have its
- *  magnitude set to 'm' and is normalized if (and only if) 'm' is zero. */
+/** Sets r to a field element with magnitude m, normalized if (and only if) m==0.
+ *  The value is chosen so that it is likely to trigger edge cases related to
+ *  internal overflows. */
 static void secp256k1_fe_get_bounds(secp256k1_fe *r, int m);
 
 /** Determine whether a is a square (modulo p). */
```
jonasnick, line 322: This comment didn't get updated: Maybe something like:
```
/** Determine whether a is a square (modulo p).
 *
 * On input, a must be a valid field element.
 */
```
nit: Should this function declaration be moved to the top to the other `_is_` functions?

jonasnick (commented):
ACK 7fc642fa25ad03ebd95cfe237b625dfb6dfdfa94

PR #1068, opened 2022-01-31
sage: Fix incompatibility with sage 9.4

### LINKED ISSUE #1067: TypeError in prove_group_implementations.sage
```
$ sage --version
SageMath version 9.4, Release Date: 2021-08-22
$ sage ./prove_group_implementations.sage
Formula secp256k1_gej_add_var:
Traceback (most recent call last):
  File "sage/rings/polynomial/multi_polynomial_libsingular.pyx", line 4680, in sage.rings.polynomial.multi_polynomial_libsingular.MPolynomial_libsingular.reduce (build/cythonized/sage/rings/polynomial/multi_polynomial_libsingular.cpp:36255)
  File "sage/structure/parent_old.pyx", line 185, in sage.structure.parent_old.Parent._coerce_c (build/cythonized/sage/structure/parent_old.c:3973)
  File "sage/structure/parent.pyx", line 1207, in sage.structure.parent.Parent.coerce (build/cythonized/sage/structure/parent.c:10953)
TypeError: no canonical coercion from Multivariate Polynomial Ring in ax, bx, ay, by, Az, Bz, Ai, Bi over Rational Field to Multivariate Polynomial Ring in ax, bx, ay, by, Az, Bz, Ai, Bi over Integer Ring

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "secp256k1/sage/./prove_group_implementations.sage.py", line 300, in <module>
    check_symbolic_jacobian_weierstrass("secp256k1_gej_add_var", _sage_const_0 , _sage_const_7 , _sage_const_5 , formula_secp256k1_gej_add_var)
  File "<string>", line 255, in check_symbolic_jacobian_weierstrass
  File "<string>", line 220, in check_symbolic_function
  File "<string>", line 306, in check_symbolic
  File "<string>", line 255, in prove_zero
  File "<string>", line 230, in prove_nonzero
  File "/nix/store/fj99mwys7crk5iifzym7q1m4li9ksknc-python3-3.9.9-env/lib/python3.9/site-packages/sage/rings/polynomial/multi_polynomial_ideal.py", line 4530, in reduce
    return f.reduce(gb)
  File "sage/rings/polynomial/multi_polynomial_libsingular.pyx", line 4683, in sage.rings.polynomial.multi_polynomial_libsingular.MPolynomial_libsingular.reduce (build/cythonized/sage/rings/polynomial/multi_polynomial_libsingular.cpp:36314)
TypeError: no canonical coercion from Multivariate Polynomial Ring in ax, bx, ay, by, Az, Bz, Ai, Bi over Rational Field to Multivariate Polynomial Ring in ax, bx, ay, by, Az, Bz, Ai, Bi over Integer Ring
```

For what it's worth, I have only very limited understanding of what the sage script does, but if I change line 230 in group_prover.sage from
```
for (f, n) in zero.reduce(numerator(allexprs)).factor():
```
to
```
for (f, n) in zero.reduce(numerator(allexprs).change_ring(QQ)).factor():
```
the type error disappears, but running `sage ./prove_group_implementations.sage` reports failure for `secp256k1_gej_add_var` (add) and `secp256k1_gej_add_zinv_var` (add).

real-or-random:
`allexprs` is already the product all numerators. Don't take it's
numerator again.

Fixes #1067.

To be honest, I have no idea what I'm doing here.

### DIFF
--- a/sage/group_prover.sage
+++ b/sage/group_prover.sage
@@ -299,22 +299,21 @@ def check_symbolic(R, assumeLaw, assumeAssert, assumeBranch, require):
 
   if conflicts(R, assume):
     # This formula does not apply
-    return None
+    return (True, None)
 
   describe = describe_extra(R, assumeLaw + assumeBranch, assumeAssert)
+  if describe != "":
+    describe = " (assuming " + describe + ")"
 
   ok, msg = prove_zero(R, require.zero, assume)
   if not ok:
-    return "FAIL, %s fails (assuming %s)" % (str(msg), describe)
+    return (False, "FAIL, %s fails%s" % (str(msg), describe))
 
   res, expl = prove_nonzero(R, require.nonzero, assume)
   if not res:
-    return "FAIL, %s fails (assuming %s)" % (str(expl), describe)
+    return (False, "FAIL, %s fails%s" % (str(expl), describe))
 
-  if describe != "":
-    return "OK (assuming %s)" % describe
-  else:
-    return "OK"
+  return (True, "OK%s" % describe)
 
 
 def concrete_verify(c):
--- a/sage/group_prover.sage
+++ b/sage/group_prover.sage
@@ -177,6 +177,30 @@ class constraints:
   def __repr__(self):
     return "%s" % self
 
+def normalize_factor(p):
+  """Normalizes the sign of primitive polynomials (as returned by factor())
+
+  This function ensures that the polynomial has a positive leading coefficient.
+
+  This is necessary because recent sage versions (starting with v9.3 or v9.4,
+  we don't know) are inconsistent about the placement of the minus sign in
+  polynomial factorizations:
+  ```
+  sage: R.<ax,bx,ay,by,Az,Bz,Ai,Bi> = PolynomialRing(QQ,8,order='invlex')
+  sage: R((-2 * (bx - ax)) ^ 1).factor()
+  (-2) * (bx - ax)
+  sage: R((-2 * (bx - ax)) ^ 2).factor()
+  (4) * (-bx + ax)^2
+  sage: R((-2 * (bx - ax)) ^ 3).factor()
+  (8) * (-bx + ax)^3
+  ```
+  """
+  # Assert p is not 0 and that its non-zero coeffients are coprime.
+  # (We could just work with the primitive part p/p.content() but we want to be
+  # aware if factor() does not return a primitive part in future sage versions.)
+  assert p.content() == 1
+  # Ensure that the first non-zero coefficient is positive.
+  return p if p.lc() > 0 else -p
 
 def conflicts(R, con):
   """Check whether any of the passed non-zero assumptions is implied by the zero assumptions"""
@@ -204,10 +228,10 @@ def get_nonzero_set(R, assume):
   nonzero = set()
   for nz in map(numerator, assume.nonzero):
     for (f,n) in nz.factor():
-      nonzero.add(f)
+      nonzero.add(normalize_factor(f))
     rnz = zero.reduce(nz)
     for (f,n) in rnz.factor():
-      nonzero.add(f)
+      nonzero.add(normalize_factor(f))
   return nonzero
 
 
@@ -222,27 +246,27 @@ def prove_nonzero(R, exprs, assume):
       return (False, [exprs[expr]])
   allexprs = reduce(lambda a,b: numerator(a)*numerator(b), exprs, 1)
   for (f, n) in allexprs.factor():
-    if f not in nonzero:
+    if normalize_factor(f) not in nonzero:
       ok = False
   if ok:
     return (True, None)
   ok = True
   for (f, n) in zero.reduce(allexprs).factor():
-    if f not in nonzero:
+    if normalize_factor(f) not in nonzero:
       ok = False
   if ok:
     return (True, None)
   ok = True
   for expr in exprs:
     for (f,n) in numerator(expr).factor():
-      if f not in nonzero:
+      if normalize_factor(f) not in nonzero:
         ok = False
   if ok:
     return (True, None)
   ok = True
   for expr in exprs:
     for (f,n) in zero.reduce(numerator(expr)).factor():
-      if f not in nonzero:
+      if normalize_factor(f) not in nonzero:
         expl.add(exprs[expr])
   if expl:
     return (False, list(expl))
@@ -279,8 +303,8 @@ def describe_extra(R, assume, assumeExtra):
     if base not in zero:
       add = []
       for (f, n) in numerator(base).factor():
-        if f not in nonzero:
-          add += ["%s" % f]
+        if normalize_factor(f) not in nonzero:
+          add += ["%s" % normalize_factor(f)]
       if add:
         ret.add((" * ".join(add)) + " = 0 [%s]" % assumeExtra.zero[base])
   # Iterate over the extra nonzero expressions
@@ -288,8 +312,8 @@ def describe_extra(R, assume, assumeExtra):
     nzr = zeroextra.reduce(numerator(nz))
     if nzr not in zeroextra:
       for (f,n) in nzr.factor():
-        if zeroextra.reduce(f) not in nonzero:
-          ret.add("%s != 0" % zeroextra.reduce(f))
+        if normalize_factor(zeroextra.reduce(f)) not in nonzero:
+          ret.add("%s != 0" % normalize_factor(zeroextra.reduce(f)))
   return ", ".join(x for x in ret)
 
 
--- a/sage/group_prover.sage
+++ b/sage/group_prover.sage
@@ -227,7 +227,7 @@ def prove_nonzero(R, exprs, assume):
   if ok:
     return (True, None)
   ok = True
-  for (f, n) in zero.reduce(numerator(allexprs)).factor():
+  for (f, n) in zero.reduce(allexprs).factor():
     if f not in nonzero:
       ok = False
   if ok:
--- a/sage/group_prover.sage
+++ b/sage/group_prover.sage
@@ -254,7 +254,7 @@ def prove_zero(R, exprs, assume):
   """Check whether all of the passed expressions are provably zero, given assumptions"""
   r, e = prove_nonzero(R, dict(map(lambda x: (fastfrac(R, x.bot, 1), exprs[x]), exprs)), assume)
   if not r:
-    return (False, map(lambda x: "Possibly zero denominator: %s" % x, e))
+    return (False, list(map(lambda x: "Possibly zero denominator: %s" % x, e)))
   zero = R.ideal(list(map(numerator, assume.zero)))
   nonzero = prod(x for x in assume.nonzero)
   expl = []
--- a/sage/group_prover.sage
+++ b/sage/group_prover.sage
@@ -164,6 +164,9 @@ class constraints:
   def negate(self):
     return constraints(zero=self.nonzero, nonzero=self.zero)
 
+  def map(self, fun):
+    return constraints(zero={fun(k): v for k, v in self.zero.items()}, nonzero={fun(k): v for k, v in self.nonzero.items()})
+
   def __add__(self, other):
     zero = self.zero.copy()
     zero.update(other.zero)
--- a/sage/prove_group_implementations.sage
+++ b/sage/prove_group_implementations.sage
@@ -292,15 +292,18 @@ def formula_secp256k1_gej_add_ge_old(branch, a, b):
   return (constraints(zero={b.Z - 1 : 'b.z=1', b.Infinity : 'b_finite'}), constraints(zero=zero, nonzero=nonzero), jacobianpoint(rx, ry, rz))
 
 if __name__ == "__main__":
-  check_symbolic_jacobian_weierstrass("secp256k1_gej_add_var", 0, 7, 5, formula_secp256k1_gej_add_var)
-  check_symbolic_jacobian_weierstrass("secp256k1_gej_add_ge_var", 0, 7, 5, formula_secp256k1_gej_add_ge_var)
-  check_symbolic_jacobian_weierstrass("secp256k1_gej_add_zinv_var", 0, 7, 5, formula_secp256k1_gej_add_zinv_var)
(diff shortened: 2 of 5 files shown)


jonasnick (commented):
I get the same errors here as with the `change_ring` "workaround":

```
Formula secp256k1_gej_add_var:
  add:
    branch 4: FAIL, <map object at 0x7f417d48b820> fails (assuming )

Formula secp256k1_gej_add_zinv_var:
  add:
    branch 4: FAIL, <map object at 0x7f417d67d420> fails (assuming )
```

real-or-random:
@jonasnick Oh I missed this... I believe this a real of case of incompleteness (something in sage changed?) and now our prover isn't complete enough to prove the formula. But the fix is easy. If you prints() to the relevant part of `prove_nonzero` like this...

```
  print("nonzero", nonzero)
  for (f, n) in allexprs.factor():
    print("factor", f)
 ```

... you get

```
nonzero {by, ax, bx, Az, bx - ax, ay, Bz}
factor -bx + ax
```

So it put `bx - ax` in the set of nonzero expressions but then fails to see that then `-bx + ax` is nonzero as well. I added a commit to address this.

Disclaimer: I don't really know what this code is doing in the end, so please take this with a grain of salt. @sipa should check if I'm right.

File: sage/group_prover.sage [hunk 1]
```diff
@@ -222,12 +222,12 @@ def prove_nonzero(R, exprs, assume):
       return (False, [exprs[expr]])
   allexprs = reduce(lambda a,b: numerator(a)*numerator(b), exprs, 1)
   for (f, n) in allexprs.factor():
-    if f not in nonzero:
+    if f not in nonzero and -f not in nonzero:
```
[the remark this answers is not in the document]
real-or-random: Both  the elements of `nonzero` and of `allexprs.factor()` are prime factors obtained by factorization, so I think they can differ at most by their `unit()` (https://doc.sagemath.org/html/en/reference/structure/sage/structure/factorization.html#sage.structure.factorization.Factorization.unit), which is -1 or 1.

So an element of `nonzero` or `allexprs.factor()` cannot have an additional factor c with c != -1 and  c != 1 because then it wouldn't be a prime factor. But there could be a constant prime element c in `allexprs.factor()`, which is then trivially nonzero but the prover fails this.

So if we want to allow for this, I'd suggest (which does not on the `%` operator for which I can't find docs on):
```
  for (f, n) in allexprs.factor():
    if f not in nonzero and -f not in nonzero and not f.is_constant():
      ok = False
```

Docs are here:
https://doc.sagemath.org/html/en/reference/polynomial_rings/sage/rings/polynomial/multi_polynomial_element.html#sage.rings.polynomial.multi_polynomial_element.MPolynomial_polydict.is_constant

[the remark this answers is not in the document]
real-or-random: > Are you sure? These are (multivariate) polynomial rings, and all constant terms are units for that.

I'm not sure at all.

> Specifically, I'm concerned about the case where you'd have factors a + 2*b or b + 1/2*a, depending on which variable is considered the one to make monic.

Hm yeah, I'd imagine this is the same everytime in the same ring class in sage but I don't know.

Would you then change every similar if within this function?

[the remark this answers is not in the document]
sipa: > Hm yeah, I'd imagine this is the same everytime in the same ring class in sage but I don't know.

Indeed, but the same reasoning should IMO apply to the just the factor 1 or -1. If there is a canonical ordering of factors, and e.g. factor() always returns the lexicographically smallest list of factors, there should never be any ambiguity at all. The fact that the sign differs makes me suspicious, so I'd rather be overly cautious (the code still works fine with that change).

> Would you then change every similar if within this function?

Yeah. Perhaps introduce a `constant_ratio_between(a, b)` function or so to abstract it and make it a bit more readable?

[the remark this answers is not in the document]
real-or-random: Ok I played around and I'm 99% convinced that 1 and -1 are complete. The reason is that the unit part is ignored when we convert the `Factorization` to a list in our code (one of the first examples in https://doc.sagemath.org/html/en/reference/structure/sage/structure/factorization.html), which then gives a canonical representation. 

Demonstration: 
(Sorry now even sage 9.5... Arch Linux is pretty quick with its updates...)


```
┌────────────────────────────────────────────────────────────────────┐
│ SageMath version 9.5, Release Date: 2022-01-30                     │
│ Using Python 3.10.2. Type "help()" for help.                       │
└────────────────────────────────────────────────────────────────────┘
sage: R.<ax,bx,ay,by,Az,Bz,Ai,Bi> = PolynomialRing(QQ,8,order='invlex')
sage: R
Multivariate Polynomial Ring in ax, bx, ay, by, Az, Bz, Ai, Bi over Rational Field

sage: R(10 * ax + bx / 2).factor()
(1/2) * (bx + 20*ax)
sage: R(20 * ax + bx).factor()
bx + 20*ax

sage: R(20 * ax + bx).factor().unit()
1
sage: R(10 * ax + bx/2).factor().unit()
1/2

sage: list(R(20 * ax + bx).factor())
[(bx + 20*ax, 1)]
sage: list(R(10 * ax + bx / 2).factor())
[(bx + 20*ax, 1)]
```

Why do 1 and -1 matter then at all, you may ask?
```
sage: R((-2 * (bx - ax)) ^ 1).factor()
(-2) * (bx - ax)
sage: R((-2 * (bx - ax)) ^ 2).factor()
(4) * (-bx + ax)^2
sage: R((-2 * (bx - ax)) ^ 3).factor()
(8) * (-bx + ax)^3
```
So it's not very consistent where they put the minus, not even consistent among odd exponents.

And indeed, I debugged the code on master and we hit `(bx - ax)^34` in `allexprs`, which then gets a sign change. 

For curiosity, can you check what this code does on your Sage 9.2? 

So with all that said, I'm in favor of sticking to current solution in this PR.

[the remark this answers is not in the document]
jonasnick: I did some independent experiments and the only factorizations I was able to get Sage (9.4) to produce had the coefficients of the non-unit factors be integers with gcd 1. This would make them unique if it weren't for this +,- inconsistency.

  I considered that this is a regression in Sage >9.2 with `order="invlex"`, because replacing it with `order="lex"` makes the script work without the `-f not in nonzero` fix. But no.
  ```
  sage: a, b, c = QQ["a,b,c"].gens()
  ....: f = 1 / 2 * a - b
  ....: f.factor()
  ....: 
  (1/2) * (a - 2*b)
  sage: f2 = (1 / 2 * a - b) * (a - c)
  ....: f2.factor()
  (1/2) * (-a + c) * (-a + 2*b)
  ```

  I'd be fine with leaving the PR as is. Another solution would be to wrap the `factor()` method such that it behaves as we would expect, i.e. the smallest variable name in a factor gets a positive coefficient (and our other assumptions are asserted).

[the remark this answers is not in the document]
jonasnick: nit: perhaps note in the commit description that this commit is required to make the script pass with sage 9.4.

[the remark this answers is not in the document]
sipa:
```
$ sage
┌────────────────────────────────────────────────────────────────────┐
│ SageMath version 9.2, Release Date: 2020-10-24                     │
│ Using Python 3.9.7. Type "help()" for help.                        │
└────────────────────────────────────────────────────────────────────┘
sage: R.<ax,bx,ay,by,Az,Bz,Ai,Bi> = PolynomialRing(QQ,8,order='invlex')
sage: R
Multivariate Polynomial Ring in ax, bx, ay, by, Az, Bz, Ai, Bi over Rational Field
sage: R(10 * ax + bx / 2).factor()
(1/2) * (bx + 20*ax)
sage: R(20 * ax + bx).factor()
bx + 20*ax
sage: R(20 * ax + bx).factor().unit()
1
sage: R(10 * ax + bx/2).factor().unit()
1/2
sage: list(R(20 * ax + bx).factor())
[(bx + 20*ax, 1)]
sage: list(R(10 * ax + bx / 2).factor())
[(bx + 20*ax, 1)]
sage: R((-2 * (bx - ax)) ^ 1).factor()
(2) * (-bx + ax)
sage: R((-2 * (bx - ax)) ^ 2).factor()
(4) * (-bx + ax)^2
sage: R((-2 * (bx - ax)) ^ 3).factor()
(8) * (-bx + ax)^3
sage: R((1/10 * ax + 1/15*bx)*(ax/7+bx)).factor()
(1/210) * (2*bx + 3*ax) * (7*bx + ax)
```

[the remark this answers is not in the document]
sipa: Is this `(f not in nonzero) and (-f not in nonzero)` needed in other places where this pattern occurs below?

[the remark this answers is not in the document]
real-or-random: > Another solution would be to wrap the factor() method such that it behaves as we would expect, i.e. the smallest variable name in a factor gets a positive coefficient (and our other assumptions are asserted).

That's a clean idea. I did this.

> Is this `(f not in nonzero) and (-f not in nonzero)` needed in other places where this pattern occurs below?

Not for the current group laws but I added the wrapper to the other places where we call `factor()`.

real-or-random:
I added sage to CI. Ready from my side (if CI passes...)

real-or-random:
Upstream report: https://trac.sagemath.org/ticket/33283

sage/group_prover.sage
```diff
@@ -177,6 +177,30 @@ class constraints:
...
+  sage: R((-2 * (bx - ax)) ^ 3).factor()
+  (8) * (-bx + ax)^3
+  ```
+  """
+  # Assert p is not 0 and that its non-zero coeffients are coprime.
+  # (We could just work with the primitive part p/p.content() but we want to be
+  # aware if factor() does not return a primitive part in future sage versions.)
+  assert p.content() == 1
+  # Ensure that the first non-zero coefficient is positive.
+  return p if p.lc() > 0 else -p
```
sipa, line 203: What about using `p / p.lc()`? That will always guarantee a result with leading coefficient 1, even if there is more than 1/-1 variability.

jonasnick: That seems possible but by asserting that the [content](https://doc.sagemath.org/html/en/reference/polynomial_rings/sage/rings/polynomial/multi_polynomial.html#sage.rings.polynomial.multi_polynomial.MPolynomial.content) is 1 we shouldn't have more than 1/-1 variability.

sipa: Sure, but if you use `p / p.lc()`, there is no need for such an assertion.

jonasnick: Ok. It seemed slightly more appealing to me to just patch the sage bug but \*shrug\*

real-or-random: > What about using `p / p.lc()`? That will always guarantee a result with leading coefficient 1, even if there is more than 1/-1 variability.

That seems to work, though this changes the representation further. `p/p.content() if p.lc() > 0 else -p/p.content()` will be closer to what we have now (and return a primitive polynomial).

But yeah, see the comment. It was an intentional decision to add the assertion. I think if that assertion fails, it may be the case that sage did larger changes and we may want to investigate this then.

But I can change it if you prefer.

jonasnick (commented):
ACK bed969b04fa292346edf5766e681a25d68b7a9d7

sage/prove_group_implementations.sage
```diff
@@ -292,15 +292,18 @@ def formula_secp256k1_gej_add_ge_old(branch, a, b):
...
+  success = success & check_symbolic_jacobian_weierstrass("secp256k1_gej_add_ge", 0, 7, 16, formula_secp256k1_gej_add_ge)
+  success = success & (not check_symbolic_jacobian_weierstrass("secp256k1_gej_add_ge_old [should fail]", 0, 7, 4, formula_secp256k1_gej_add_ge_old))
 
   if len(sys.argv) >= 2 and sys.argv[1] == "--exhaustive":
-    check_exhaustive_jacobian_weierstrass("secp256k1_gej_add_var", 0, 7, 5, formula_secp256k1_gej_add_var, 43)
-    check_exhaustive_jacobian_weierstrass("secp256k1_gej_add_ge_var", 0, 7, 5, formula_secp256k1_gej_add_ge_var, 43)
-    check_exhaustive_jacobian_weierstrass("secp256k1_gej_add_zinv_var", 0, 7, 5, formula_secp256k1_gej_add_zinv_var, 43)
-    check_exhaustive_jacobian_weierstrass("secp256k1_gej_add_ge", 0, 7, 16, formula_secp256k1_gej_add_ge, 43)
-    check_exhaustive_jacobian_weierstrass("secp256k1_gej_add_ge_old [should fail]", 0, 7, 4, formula_secp256k1_gej_add_ge_old, 43)
+    success = success & check_exhaustive_jacobian_weierstrass("secp256k1_gej_add_var", 0, 7, 5, formula_secp256k1_gej_add_var, 43)
```
real-or-random, line 303: This is buggy for `--exhaustive`, let me fix this.

real-or-random: Ok done. Also cherry-picked a commit from https://github.com/real-or-random/secp256k1/commits/202202-gej_add_ge that I discovered when working on #1033.

sage/group_prover.sage
sipa: No problem.

sipa:
ACK ebb1beea7832207e5c8c5112d250fd216259ef41

sage/prove_group_implementations.sage
jonasnick: The commit looks fine but is there a specific reason why this commit was added?

real-or-random: I added it because it fixes a bug in the group prover that prevents us to look into other group formulas. In particular, it stopped me from playing around with the formula here because it may return a literal 1: https://github.com/real-or-random/secp256k1/blob/dc53e24a1b63ec07a9ab086280cab1355cd3a881/sage/prove_group_implementations.sage#L209

jonasnick: Ah, thanks that makes sense.

jonasnick (commented):
ACK ebb1beea7832207e5c8c5112d250fd216259ef41

PR #1072, opened 2022-02-04
ci: Attempt to make macOS builds more reliable

real-or-random:
The macOS CI tasks often error fail when doing `brew update` with
git fetch errors:
```
remote: fatal: packfile /data/repositories/b/nw/b6/07/5c/123272362/network.git/objects/pack/pack-2139bd07361b62a358e380a0e7d58ec35593d191.pack cannot be accessed
fatal: protocol error: bad pack header
Error: Fetching /usr/local/Homebrew/Library/Taps/homebrew/homebrew-core failed!
```
Superficially this seems to be a problem on the GitHub server because
the message shows a "remote" error. But it seems we're the only one in
the world running into this specific issue when doing `brew update`, so
it's more likely that the something else is the culprit, and this error
message is just a symptom.

This commit replaces `brew update` with a complete reinstallation of
brew. This is essentially a shot in the dark but it's worth a try, and
I doubt it's significantly more expensive. If that won't work, we may
consider simply retrying `brew update` a few times.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -135,7 +135,9 @@ task:
   ##   - rm /tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress
   ##
   brew_valgrind_pre_script:
-    - brew update
+    # Reinstall brew. We could do `brew update` instead but that often fails.
+    - /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/uninstall.sh)"
+    - /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
     - brew config
     - brew tap LouisBrunner/valgrind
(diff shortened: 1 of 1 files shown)


real-or-random:
Ok CI passed on first try, and this adds < 10 seconds to the macOS tasks. Let's re-run it. :P

real-or-random:
Ok, all three runs pass. I think it's worth the try and merge it.

jonasnick (commented):
ACK d9396a56da13df97827e193f636a7593fb4d7864

See issue https://github.com/cirruslabs/osx-images/issues/41#issuecomment-1025009317 for more context.

real-or-random:
Great, now  we don't even get proper error messages any more: https://cirrus-ci.com/task/6275416297046016?logs=brew_valgrind_pre#L126

I think we should use the script that Core uses: https://github.com/bitcoin/bitcoin/tree/master/ci/retry

PR #1074, opened 2022-02-08
 ci: Retry brew update a few times to avoid random failures

real-or-random:
Apparently #1072 didn't work out. I should have done this here directly.

Core vendors a script for retrying (https://github.com/bitcoin/bitcoin/tree/master/ci/retry) but I think that's overkill for us.


jonasnick (commented):
ACK e51ad3b737a2f12605d346cf9c0e7eaabf6a39ec

PR #1079, opened 2022-02-23
configure: Add hidden --enable-dev-mode to enable all the stuff

real-or-random:
Rework of #975, which I'm taking over with the permission of @elichai.

See commit messages for some more rationale.

### DIFF
--- a/build-aux/m4/bitcoin_secp.m4
+++ b/build-aux/m4/bitcoin_secp.m4
@@ -38,3 +38,16 @@ AC_DEFUN([SECP_TRY_APPEND_CFLAGS], [
   unset flag_works
   AC_SUBST($2)
 ])
+
+dnl SECP_SET_DEFAULT(VAR, default, default-dev-mode)
+dnl Set VAR to default or default-dev-mode, depending on whether dev mode is enabled
+AC_DEFUN([SECP_SET_DEFAULT], [
+  if test "${enable_dev_mode+set}" != set; then
+    AC_MSG_ERROR([[Set enable_dev_mode before calling SECP_SET_DEFAULT]])
+  fi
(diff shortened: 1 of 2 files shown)


real-or-random:
I was wondering whether we should disable the benchmarks by default. I think normal users won't need them and I suspect these were more interesting in the early days of the library.

sipa:
The benchmarks were actually disabled at some point by default, which led to a lot of wasted time for me because I forgot to compile benchmarks, and then ended up running with an old `bench` binary...

build-aux/m4/bitcoin_secp.m4
```diff
@@ -38,3 +38,16 @@ AC_DEFUN([SECP_TRY_APPEND_CFLAGS], [
   unset flag_works
   AC_SUBST($2)
 ])
+
+dnl SECP_SET_DEFAULT(VAR, default, default-dev-mode)
+dnl Set VAR to default or default-dev-mode, depending on whether dev mode is enabled
+AC_DEFUN([SECP_SET_DEFAULT], [
+  if test "${enable_dev_mode+set}" != set; then
+    AC_MSG_ERROR([[Set enable_dev_mode before calling SECP_SET_DEFAULT]])
+  fi
```
jonasnick, line 47: nit: trailing whitespace

jonasnick (commented):
ACK mod nit

This is useful.

real-or-random:
@elichai Want to review this?

elichai:
tACK e0838d663deba5a9663a0ad02bd7e4c563325c4d

Tested that this is actually hidden, that it enables all the modules as expected, and that it supports disabling specific modules

PR #1084, opened 2022-03-10
ci: Add MSVC builds

real-or-random:
This adds MSVC builds built on Linux using wine. This requires some
settings of tools and flags because the autotools support for MSVC is
naturally somewhat limited.

The advantage of this approach is that it is compatible with our
existing CI scripts, so there's no need to write a Windows CI script
(in PowerShell or similar). If we want to test building and running on
Windows native (e.g., as supported by Cirrus CI) we could still do this
in the future.

Another advantage of this approach is that contributors can simply use
the docker image if they need a MSVC installation in a non-Windows
environment.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -349,7 +349,7 @@ task:
   << : *CAT_LOGS
 
 task:
-  name: "C++ -fpermissive"
+  name: "C++ -fpermissive (entire project)"
   << : *LINUX_CONTAINER
   env:
     CC: g++
@@ -364,6 +364,14 @@ task:
   test_script:
     - ./ci/cirrus.sh
   << : *CAT_LOGS
+
+task:
+  name: "C++ (public headers)"
+  << : *LINUX_CONTAINER
+  test_script:
+    - g++ -Werror include/*.h
+    - clang -Werror -x c++-header include/*.h
+    - /opt/msvc/bin/x64/cl.exe -c -WX -TP include/*.h
 
 task:
   name: "sage prover"
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -312,10 +312,9 @@ task:
   name: "C++ -fpermissive"
   << : *LINUX_CONTAINER
   env:
-    # ./configure correctly errors out when given CC=g++.
-    # We hack around this by passing CC=g++ only to make.
-    CC: gcc
-    MAKEFLAGS: -j4 CC=g++ CFLAGS=-fpermissive\ -g
+    CC: g++
+    CFLAGS: -fpermissive -g
+    CPPFLAGS: -DSECP256K1_CPLUSPLUS_TEST_OVERRIDE
     WERROR_CFLAGS:
     EXPERIMENTAL: yes
     ECDH: yes
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -263,6 +263,42 @@ task:
     - ./ci/cirrus.sh
   << : *CAT_LOGS
 
+task:
+  << : *LINUX_CONTAINER
+  env:
+    WRAPPER_CMD: wine
+    WERROR_CFLAGS: -WX
+    WITH_VALGRIND: no
+    ECDH: yes
+    RECOVERY: yes
+    EXPERIMENTAL: yes
+    SCHNORRSIG: yes
+    CTIMETEST: no
+    # Set non-essential options that affect the CLI messages here.
+    # (They depend on the user's taste, so we don't want to set them automatically in configure.ac.)
+    CFLAGS: -nologo -diagnostics:caret
+    LDFLAGS: -XCClinker -nologo -XCClinker -diagnostics:caret
+  # Use a MinGW-w64 host to tell ./configure we're building for Windows.
+  # This will detect some MinGW-w64 tools but then make will need only
+  # the MSVC tools CC, AR and NM as specified below.
+  matrix:
+    - name: "x86_64 (MSVC): Windows (Debian stable, Wine)"
+      env:
+        HOST: x86_64-w64-mingw32
+        CC: /opt/msvc/bin/x64/cl
+        AR: /opt/msvc/bin/x64/lib
+        NM: /opt/msvc/bin/x64/dumpbin -symbols -headers
+    - name: "i686 (MSVC): Windows (Debian stable, Wine)"
+      env:
+        HOST: i686-w64-mingw32
+        CC: /opt/msvc/bin/x86/cl
+        AR: /opt/msvc/bin/x86/lib
+        NM: /opt/msvc/bin/x86/dumpbin -symbols -headers
+  << : *MERGE_BASE
+  test_script:
+    - ./ci/cirrus.sh
+  << : *CAT_LOGS
+
 # Sanitizers
 task:
   << : *LINUX_CONTAINER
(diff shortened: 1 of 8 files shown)


real-or-random:
by the way, rather for fun:	
```

32-bit MSVC -O2 (over wine):
ecdsa_verify                  ,   226.0       ,   227.0       ,   230.0
ecdsa_sign                    ,   133.0       ,   134.0       ,   134.0
ecdh                          ,   226.0       ,   227.0       ,   229.0
ecdsa_recover                 ,   230.0       ,   231.0       ,   233.0
schnorrsig_sign               ,   120.0       ,   120.0       ,   123.0
schnorrsig_verify             ,   227.0       ,   228.0       ,   232.0

32-bit MinGw -O2 (over wine):
ecdsa_verify                  ,   195.0       ,   196.0       ,   198.0
ecdsa_sign                    ,   115.0       ,   116.0       ,   117.0
ecdh                          ,   199.0       ,   200.0       ,   204.0
ecdsa_recover                 ,   199.0       ,   200.0       ,   204.0
schnorrsig_sign               ,   101.0       ,   102.0       ,   103.0
schnorrsig_verify             ,   196.0       ,   197.0       ,   200.0

32-bit Linux GCC -O2:
ecdsa_verify                  ,    95.6       ,    96.0       ,    96.7    
ecdsa_sign                    ,    64.9       ,    65.6       ,    68.0    
ecdh                          ,    98.7       ,    99.4       ,   102.0    
ecdsa_recover                 ,    97.9       ,    98.8       ,   100.0    
schnorrsig_sign               ,    55.0       ,    55.4       ,    56.6    
schnorrsig_verify             ,    95.9       ,    97.0       ,    99.5   
```

(Not sure why windows binaries are so much slower. Maybe there's something wrong with testing on Linux but wine shouldn't have a large overhead...)

real-or-random:
Ready for review (if CI passes...)

real-or-random:
This will probably conflict with #1088. Let's first try to merge the fix there.

sipa:
utACK c16afd001ac441c7729475fa18bcea60ef2bad2f

src/secp256k1.c
```diff
@@ -4,6 +4,17 @@
...
+/* This is a C project. It should not be compiled with a C++ compiler,
+ * and we error out if we detect one.
+ *
+ * We still want to be able to test the project with a C++ compiler
+ * because it is still good to know if this will lead to real trouble, so
+ * there is a possibility to override the check. But be warned that
+ * compiling with a C++ compiler is not supported. */
+#if defined(__cplusplus) && !defined(SECP256K1_CPLUSPLUS_TEST_OVERRIDE)
+#error Trying to compile a C project with a C++ compiler.
+#endif
```
jonasnick, line 16: Is it intended that I don't see this error with `g++` unless I also compile with `-fpermissive`? This is the case because g++ aborts earlier due to other errors. So if someone would accidentally use g++ (without `-fpermissive`) then they may just be confused vs. having a rather clear error message as before. Not saying that this is terrible, but seems like a change from the previous behavior.

configure.ac
```diff
@@ -89,23 +91,35 @@ esac
...
+      SECP_TRY_APPEND_CFLAGS([-fvisibility=hidden], $1) # GCC >= 4.0
+
+      CFLAGS="$SECP_TRY_APPEND_DEFAULT_CFLAGS_saved_CFLAGS"
+    fi
+
+    # MSVC
+    # Assume MSVC if we're building for Windows but not with GCC or compatible;
+    # libtool makes the same assumption internally.
+    # Note that "/opt" and "-opt" are equivalent for MSVC; we use "-opt" because "/opt" looks like a path.
+    if test x"$GCC" != x"yes" && test x"$build_windows" = x"yes"; then
```
jonasnick, line 119: Instead of
```
if test "x$GCC" = "xyes"; then ...
if test x"$GCC" != x"yes" && test x"$build_windows" = x"yes"; then ...
```
why not set up the conditions like this
```
if test x"$GCC" != x"yes" && test x"$build_windows" = x"yes"; then ...
else ...
```
or
```
if test "x$GCC" = "xyes"; then ...
if test x"$GCC" != x"yes" && test x"$build_windows" = x"yes"; then
else error("something is wrong")
```

.cirrus.yml
```diff
@@ -246,18 +246,58 @@ task:
...
+    WITH_VALGRIND: no
+    ECDH: yes
+    RECOVERY: yes
+    EXPERIMENTAL: yes
+    SCHNORRSIG: yes
+    CTIMETEST: no
+    # Set non-essential options that affect the CLI messages here.
+    # (They depend on the user's taste, so we don't want to set them automatically in configure.ac.)
+    CFLAGS: -nologo -diagnostics:caret
+    LDFLAGS: -XCClinker -nologo -XCClinker -diagnostics:caret
```
jonasnick, line 284: Do you think `-WX` (treat warnings as errors) would be too fragile?

configure.ac
```diff
@@ -32,17 +32,19 @@ AM_INIT_AUTOMAKE([1.11.2 foreign subdir-objects])
 # Make the compilation flags quiet unless V=1 is used.
 m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 
-PKG_PROG_PKG_CONFIG
-
 AC_PROG_CC
-if test x"$ac_cv_prog_cc_c89" = x"no"; then
-  AC_MSG_ERROR([c89 compiler support required])
-fi
```
jonasnick, line 40: Did you try the MSVC `-Za` flag which according to the docs should set `__STDC__`? (https://docs.microsoft.com/en-us/cpp/build/reference/za-ze-disable-language-extensions?view=msvc-170)

.cirrus.yml
jonasnick: Ah sorry I missed this. There's a comment in `configure.ac` which is sufficient imo.

configure.ac
jonasnick: > thought it's a good idea not to fiddle with the CFLAGS at all if this not GCC or compatible

Yeah that makes sense. Just wanted to make sure this wasn't by accident.

> That would always error out on platforms other than Windows 

I believe I intended the second `if` to be an `else if`. Anyway, erroring out is unnecessary.

.cirrus.yml
```diff
@@ -246,18 +246,58 @@ task:
...
+    WRAPPER_CMD: wine
     WITH_VALGRIND: no
     ECDH: yes
     RECOVERY: yes
     EXPERIMENTAL: yes
     SCHNORRSIG: yes
     CTIMETEST: no
+  matrix:
+    - name: "x86_64 (mingw32-w64): Windows (Debian stable, Wine)"
+      env:
```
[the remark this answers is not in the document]
real-or-random: fixed

src/modules/schnorrsig/bench_impl.h
```diff
@@ -91,10 +91,10 @@ void run_schnorrsig_bench(int iters, int argc, char** argv) {
...
         free((void *)data.sigs[i]);
     }
-    free(data.keypairs);
-    free(data.pk);
-    free(data.msgs);
-    free(data.sigs);
+    free((void *)data.keypairs);
+    free((void *)data.pk);
+    free((void *)data.msgs);
+    free((void *)data.sigs);
```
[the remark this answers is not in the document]
real-or-random: done

real-or-random:
addressed all comments

jonasnick (commented):
ACK 49e2acd927ce9eb806cc10f3a1fd89a9ddd081e2

PR #1086, opened 2022-03-12
configure: customizable AR and RANLIB

jb55:
This is needed for emscripten wasm builds

Signed-off-by: William Casarin <jb55@jb55.com>


real-or-random:
It's a nice discovery that the wasm failure is related to `ar`. Can you test #1088 instead? I think that's the modern and recommended to solve the problem.

jb55:
Closing in favor of #1088

PR #1088, opened 2022-03-13
configure: Use modern way to set AR

real-or-random:
This uses AM_PROG_AR to discover ar, which is the recommended way to do
so. Among other advantages, it honors the AR environment variable (as
set from the outside). The macro has been around since automake 1.11.2
(Dec 2011).

This commit also removes code that discovers ranlib and strip. ranlib
has been obsolete for decades (ar does its task now automatically), and
anyway LT_INIT takes care of discovering it. The code we used to set
STRIP was last mentioned in the automake 1.5 manual. Since automake 1.6
(Mar 2002), strip is discovered automatically when necessary (look for
the *private* macro AM_PROG_INSTALL_STRIP in the automake manual).

Alternative to #1086.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -25,24 +25,23 @@ AC_CANONICAL_HOST
 AH_TOP([#ifndef LIBSECP256K1_CONFIG_H])
 AH_TOP([#define LIBSECP256K1_CONFIG_H])
 AH_BOTTOM([#endif /*LIBSECP256K1_CONFIG_H*/])
-AM_INIT_AUTOMAKE([foreign subdir-objects])
 
-LT_INIT([win32-dll])
+# Require Automake 1.11.2 for AM_PROG_AR
+AM_INIT_AUTOMAKE([1.11.2 foreign subdir-objects])
 
 # Make the compilation flags quiet unless V=1 is used.
 m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 
 PKG_PROG_PKG_CONFIG
 
-AC_PATH_TOOL(AR, ar)
-AC_PATH_TOOL(RANLIB, ranlib)
-AC_PATH_TOOL(STRIP, strip)
-
 AC_PROG_CC
 if test x"$ac_cv_prog_cc_c89" = x"no"; then
   AC_MSG_ERROR([c89 compiler support required])
 fi
 AM_PROG_AS
+AM_PROG_AR
+
+LT_INIT([win32-dll])
 
 build_windows=no
 
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ coverage.*.html
 
 src/libsecp256k1-config.h
 src/libsecp256k1-config.h.in
+build-aux/ar-lib
 build-aux/config.guess
 build-aux/config.sub
 build-aux/depcomp


real-or-random:
Force-pushed to improve commit message.

hebasto:
Concept ACK.

configure.ac
```diff
@@ -27,17 +27,14 @@ AH_TOP([#define LIBSECP256K1_CONFIG_H])
 AH_BOTTOM([#endif /*LIBSECP256K1_CONFIG_H*/])
 AM_INIT_AUTOMAKE([foreign subdir-objects])
 
+AM_PROG_AR # must be called before LT_INIT
```
hebasto, line 30: Why the comment does matter? Any docs link?

hebasto (commented):
> The macro has been around since automake 1.11.2 (Dec 2011).

Maybe specify minimum Automake version in the `AM_INIT_AUTOMAKE` macro explicitly?

>This commit also removes code that discovers ranlib and strip.
Slightly prefer to split these changes into 2 commits -- (1) ar-related, and (2) other changes.

configure.ac
real-or-random: ./autogen.sh will warn if you have it the other way around.

The `AM_PROG_AR` felt a little bit out of place here (because all the other tools are detected below), so I thought a comment is good. Another way is to move `AC_PROG_CC` and `AM_PROG_AS` up, maybe I'll do this.

real-or-random: Fixed.

real-or-random:
@hebasto @fanquake I don't think it hurts but one thing to be aware of is that `AM_PROG_AR` explicitly supports the MSVC archiver "lib.exe" and thus `./autogen.sh` will install an additional wrapper script in `build-aux`. This will actually be helpful for my PR #1084, so I believe we even want this here. 

But if that's not desired in Core, sticking with `AC_PATH_TOOL(AR, ar)` is fine for now. (Future versions of autoconf will have `AC_PROG_AR` (note the `AC` instead of `AM`) which essentially does this, see https://github.com/autotools-mirror/autoconf/commit/c48fdb81191c8b7c7c0dde6141b861b178a6a284). But even if this case, I think Core can still remove `AC_PATH_TOOL(RANLIB, ranlib)` and for `AC_PATH_TOOL(STRIP, strip)` for the reasons given above.

hebasto:
> `AM_PROG_AR` explicitly supports the MSVC archiver "lib.exe" and thus `./autogen.sh` will install an additional wrapper script in `build-aux`.

To be precise, it is the `build-aux/ar-lib` script.

configure.ac
hebasto: > ./autogen.sh will warn if you have it the other way around.

`warning: LT_INIT was called before AM_PROG_AR`

hebasto:
> This commit also removes code that discovers ranlib and strip. ranlib
> has been obsolete for decades (ar does its task now automatically), and
> anyway LT_INIT takes care of discovering it. The code we used to set
> STRIP was last mentioned in the automake 1.5 manual. Since automake 1.6
> (Mar 2002), strip is discovered automatically when necessary (look for
> the _private_ macro AM_PROG_INSTALL_STRIP in the automake manual).

Confirming. Here is an excerpt of the `LT_INIT([win32-dll])` output:
```
checking for strip... strip
checking for ranlib... ranlib
```

hebasto (commented):
Approach ACK 67f4569db8be4d96d7016e7acea6ea3f051649c4.

FWIW, here is a diff of `./configure` output:
```diff
--- /home/hebasto/master
+++ /home/hebasto/pr1088
@@ -6,8 +6,9 @@
 checking for gawk... gawk
 checking whether make sets $(MAKE)... yes
 checking whether make supports nested variables... yes
-checking how to print strings... printf
-checking whether make supports the include directive... yes (GNU style)
+checking whether make supports nested variables... (cached) yes
+checking for pkg-config... /usr/bin/pkg-config
+checking pkg-config is at least version 0.9.0... yes
 checking for gcc... gcc
 checking whether the C compiler works... yes
 checking for C compiler default output file name... a.out
@@ -18,7 +19,12 @@
 checking whether gcc accepts -g... yes
 checking for gcc option to enable C11 features... none needed
 checking whether gcc understands -c and -o together... yes
+checking whether make supports the include directive... yes (GNU style)
+checking dependency style of gcc... gcc3
 checking dependency style of gcc... gcc3
+checking for ar... ar
+checking the archiver (ar) interface... ar
+checking how to print strings... printf
 checking for a sed that does not truncate output... /usr/bin/sed
 checking for grep that handles long lines and -e... /usr/bin/grep
 checking for egrep... /usr/bin/grep -E
@@ -36,7 +42,6 @@
 checking how to recognize dependent libraries... pass_all
 checking for dlltool... no
 checking how to associate runtime and link libraries... printf %s\n
-checking for ar... ar
 checking for archiver @FILE support... @
 checking for strip... strip
 checking for ranlib... ranlib
@@ -71,19 +76,6 @@
 checking if libtool supports shared libraries... yes
 checking whether to build shared libraries... yes
 checking whether to build static libraries... yes
-checking whether make supports nested variables... (cached) yes
-checking for pkg-config... /usr/bin/pkg-config
-checking pkg-config is at least version 0.9.0... yes
-checking for ar... /usr/bin/ar
-checking for ranlib... /usr/bin/ranlib
-checking for strip... /usr/bin/strip
-checking for gcc... (cached) gcc
-checking whether the compiler supports GNU C... (cached) yes
-checking whether gcc accepts -g... (cached) yes
-checking for gcc option to enable C11 features... (cached) none needed
-checking whether gcc understands -c and -o together... (cached) yes
-checking dependency style of gcc... (cached) gcc3
-checking dependency style of gcc... gcc3
 checking if gcc supports -Werror=unknown-warning-option... no
 checking if gcc supports -std=c89 -pedantic -Wno-long-long -Wnested-externs -Wshadow -Wstrict-prototypes -Wundef... yes
 checking if gcc supports -Wno-overlength-strings... yes
```

real-or-random:
> ```diff
> +build-aux/ar-lib
> ```

Done, thanks.

hebasto (approved):
ACK 0d253d52e804a5affb0f1c851ec250071e7345d9

jb55:
tACK 0d253d52e804a5affb0f1c851ec250071e7345d9

my [lnsocket](https://github.com/jb55/lnsocket) wasm build is happy with this

roconnor-blockstream:
works for me.

jonasnick (commented):
ACK 0d253d52e804a5affb0f1c851ec250071e7345d9

PR #1089, opened 2022-03-16
Schnorrsig API improvements

real-or-random:
Should be merged before #995 if we want this.

I suspect the only change here which is debatable on a conceptual level is the renaming. I can drop this of course.

### DIFF
--- a/examples/schnorr.c
+++ b/examples/schnorr.c
@@ -18,16 +18,9 @@
 #include "random.h"
 
 int main(void) {
-    /* Instead of signing the message directly, we must sign a 32-byte hash.
-     * Here the message is "Hello, world!" and the hash function was SHA-256.
-     * An actual implementation should just call SHA-256, but this example
-     * hardcodes the output to avoid depending on an additional library. */
-    unsigned char msg_hash[32] = {
-        0x31, 0x5F, 0x5B, 0xDB, 0x76, 0xD0, 0x78, 0xC4,
-        0x3B, 0x8A, 0xC0, 0x06, 0x4E, 0x4A, 0x01, 0x64,
-        0x61, 0x2B, 0x1F, 0xCE, 0x77, 0xC8, 0x69, 0x34,
-        0x5B, 0xFC, 0x94, 0xC7, 0x58, 0x94, 0xED, 0xD3,
-    };
+    unsigned char msg[12] = "Hello World!";
+    unsigned char msg_hash[32];
+    unsigned char tag[17] = "my_fancy_protocol";
     unsigned char seckey[32];
     unsigned char randomize[32];
     unsigned char auxiliary_rand[32];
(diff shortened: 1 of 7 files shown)


examples/schnorr.c
```diff
@@ -96,7 +96,7 @@ int main(void) {
      * improve security against side-channel attacks. Signing with a valid
      * context, verified keypair and the default nonce function should never
      * fail. */
-    return_val = secp256k1_schnorrsig_sign(ctx, signature, msg_hash, &keypair, auxiliary_rand);
+    return_val = secp256k1_schnorrsig_sign32(ctx, signature, msg_hash, &keypair, auxiliary_rand);
```
real-or-random, line 111: If you want to test SECP256K1_DEPRECATED, simply remove the "32" here again and look out for compiler warnings.

michaelfolkson:
Concept ACK

I read the discussion on the renaming and it seems the only downsides are that the BIP needs to edited and a deprecation process of `schnorrsign_sign` is needed. Other projects (e.g. DLCs) seem to want to be able sign variable length messages and we'd like to facilitate other uses of this library if it doesn't present material downsides. I haven't tested yet but thanks for the guidance [comment](https://github.com/bitcoin-core/secp256k1/pull/1089#discussion_r827924114).

michaelfolkson:
Built successfully on MacOS Big Sur with `--enable-dev-mode` flag and ran tests.

> If you want to test SECP256K1_DEPRECATED, simply remove the "32" here again and look out for compiler warnings.

Followed this instruction and indeed got the compiler warnings:

`examples/schnorr.c:111:18: warning: 'secp256k1_schnorrsig_sign' is deprecated: Use secp256k1_schnorrsig_sign32 instead [-Wdeprecated-declarations]
    return_val = secp256k1_schnorrsig_sign(ctx, signature, msg_hash, &keypair, auxiliary_rand);
                 ^
./include/secp256k1_schnorrsig.h:136:3: note: 'secp256k1_schnorrsig_sign' has been explicitly marked deprecated here
  SECP256K1_DEPRECATED("Use secp256k1_schnorrsig_sign32 instead");
  ^
./include/secp256k1.h:179:55: note: expanded from macro 'SECP256K1_DEPRECATED'
#   define SECP256K1_DEPRECATED(_msg) __attribute__ ((__deprecated__(_msg)))
                                                      ^
1 warning generated.
  CCLD     schnorr_example`

include/secp256k1.h
```diff
@@ -800,7 +817,7 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_combine(
  *  implementations optimized for a specific tag can precompute the SHA256 state
  *  after hashing the tag hashes.
  *
- *  Returns 0 if the arguments are invalid and 1 otherwise.
+ *  Returns: 1 always.
```
michaelfolkson, line 820: Wasn't sure why this has changed. Code below hasn't changed.

real-or-random: The function really always returns 1:
https://github.com/bitcoin-core/secp256k1/blob/master/src/secp256k1.c#L750

I think this was just a copy-and-paste mistake when writing the API docs. Strictly speaking, it's not even an API change because there  simply are no invalid arguments for this function (except those catched by `ARG_CHECK` but then we'll die and the function won't return.)

jonasnick: "We" are not guaranteed to die in an `ARG_CHECK` in case custom callbacks are used. Then the return value is unspecified in theory but 0 in practice. This is also how we describe the behavior of other functions (example `keypair_sec`, `keypair_pub`).

include/secp256k1.h
```diff
@@ -169,6 +169,20 @@ typedef int (*secp256k1_nonce_function)(
...
 # endif
 
+/** Attribute for marking functions, types, and variables as deprecated
+  * GCC <4.5 supports the attribute but not with an optional message argument. */
+#if !defined(SECP256K1_BUILD)
+# if defined(__GNUC__) && SECP256K1_GNUC_PREREQ(4, 5)
+#  define SECP256K1_DEPRECATED(_msg) __attribute__ ((__deprecated__(_msg)))
+# elif defined(__has_attribute)
+#  if __has_attribute(__deprecated__)
+#   define SECP256K1_DEPRECATED(_msg) __attribute__ ((__deprecated__(_msg)))
```
jonasnick, line 179: This line seems to be identical to the one in the previous branch. Did you intend to remove the `_msg` argument here?

jonasnick (commented):
concept ACK

For the record, I probably wouldn't rename sign at this point because it's not obviously better. I agree that it's more logical in terms of the spec, but for a user looking at the include file it's not immediately clear what the "32" suffix means and potentially causes confused searches for the "normal" signing function. But I'm ok with the change.

include/secp256k1.h
real-or-random: No this is intentional, I just didn't find a nicer way to write it...

Let me explain. One can't write this as a single condition
```
(defined(__GNUC__) && SECP256K1_GNUC_PREREQ(4, 5)) || (__has_attribute(__deprecated__) && __has_attribute(__deprecated__)
```
because this is illegal if `__has_attribute` is really not defined.

The `__has_attribute` is actually a clang thing (plus very recent GCC) and the clang docs suggest this here:
```
#ifndef __has_attribute
  #define __has_attribute(x) 0
#endif
```

Looks nice but I'm somewhat hesitant to define a reserved `__` identifier in a public include file. It's technically UB and in this case it could really lead to strange issues: Say the user includes our file and then checks `defined(__has_attribute)` in their code... I don't see an immediate problem with this, but it seems just wrong to do this.

real-or-random: Oh indeed, "very recent GCC" is not true. I switched to your suggestion (also added a missing #else case)

include/secp256k1.h
real-or-random: I fixed also the other functions. See the commit message for detailed rationale.

jonasnick (commented):
ACK b8f8b99f0fb3a5cd4c6fb1c9c8dfed881839e19e

include/secp256k1.h
```diff
@@ -169,6 +169,17 @@ typedef int (*secp256k1_nonce_function)(
 #  define SECP256K1_ARG_NONNULL(_x)
 # endif
 
+/** Attribute for marking functions, types, and variables as deprecated */
+#if !defined(SECP256K1_BUILD) && defined(__has_attribute)
```
sipa, line 173: IIRC GCC doesn't have `__has_attribute` until very recent versions, but does have `__attribute__((__deprecated__))` since at least GCC 4.1 (tested with godbolt) and probably earlier. Do we perhaps want this additionally enabled on based on `__GNUC__`?

real-or-random: This is what I had earlier but @jonasnick  pointed out that already GCC >= 5 offers `__has__attribute`. Moreover, `__attribute__((__deprecated__))` indeed works in >= GCC 4.1 but the optional `msg` argument works only in GCC >= 4.5 (in fact you get an error if you pass an argument in GCC < 4.5. 

(See the resolved discussion above).

sipa: Oh wow, I thought __has_attribute was only added much later. Looks good then.

sipa:
utACK b8f8b99f0fb3a5cd4c6fb1c9c8dfed881839e19e

PR #1090, opened 2022-03-16
configure: Remove redundant pkg-config code

real-or-random:
This removes code that detects the pkg-config tool. We used this
back in the days when we had dependencies. ;) It can always be brought
back if we'll need it in the future.

Note that we still deliver a .pc file for this library, and there is
code in Makefile.am to install it. But this does not require the
pkg-config tool; only consumers of the .pc file will need it. This can
be verified by running `make install` (maybe after `mkdir /tmp/pre` and
`./configure --prefix=/tmp/pre` and checking that the .pc file is
installed correctly.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -32,8 +32,6 @@ AM_INIT_AUTOMAKE([1.11.2 foreign subdir-objects])
 # Make the compilation flags quiet unless V=1 is used.
 m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 
-PKG_PROG_PKG_CONFIG
-
 AC_PROG_CC
 if test x"$ac_cv_prog_cc_c89" = x"no"; then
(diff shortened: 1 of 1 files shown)


fanquake:
Nice. Concept ACK

theuni:
ACK 21b2ebaf74222017f85123deb6f30a33c7678513.

Nice to not have dependencies :)

theuni:
Also note that removal may allow Core to get rid of this hack: https://github.com/bitcoin/bitcoin/blob/d6cb4e8ff0d0772fd549240c21cf08e5213259be/configure.ac#L1939-L1945

fanquake (approved):
ACK 21b2ebaf74222017f85123deb6f30a33c7678513

> Also note that removal may allow Core to get rid of this hack:

I've pushed changes up into https://github.com/bitcoin/bitcoin/pull/24341 so we can have a look.

PR #1091, opened 2022-03-16
doc: Update README for --enable-dev-mode

michaelfolkson:
Since #1079 the `--enable-dev-mode` flag turns on all experimental modules so update README to reflect this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ libsecp256k1 is built using autotools:
 Usage examples
 -----------
   Usage examples can be found in the [examples](examples) directory. To compile them you need to configure with `--enable-examples`.
-  For experimental modules, you will also need `--enable-experimental` as well as a flag for each individual module, e.g. `--enable-module-ecdh`.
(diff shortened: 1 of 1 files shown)


real-or-random:
Hm, dev-mode was "hidden" on purpose. I don't think we should encourage people to enable experimental modules, at least not without calling them "experimental" explicitly. Maybe adding the text to the README was actually an oversight.

real-or-random:
In fact,  after #1091 we won't have experimental modules, so #1091 should remove the sentence "For experimental modules, you will also need --enable-experimental as well as a flag for each individual module, e.g. --enable-module-ecdh." This was overlooked because the PR was opened before this text existed.

This will resolve the problem for that part in the README, and I feel the README is the wrong place to mention the option. There's also reason why ./configure --help doesn't show it. I think it would be okay to document the option in some development docs  (but we don't have development docs, and adding them just to point this out seems overkill.)

michaelfolkson:
Ok no worries, closing.

PR #1105, opened 2022-05-04
Don't export symbols in static libraries

theuni:
For context, Bitcoin Core has recently merged [libbitcoin-kernel](https://github.com/bitcoin/bitcoin/pull/24322), a small library that intends to eventually minimally encompass Core's validation engine. This kernel lib includes a static libsecp256k1. Without this change, because libsecp256k1.a ends up with exported symbols, we end up with libsecp256k1 symbols exported by our libbitcoin-kernel library (which causes unrelated problems not worth getting into here).

libtool takes care of building both object versions, and it automatically builds objects for shared libs with -DDLL_EXPORT. We just need to opt-in to its functionality.

I can't imagine this having any negative impact on any current statically-linking applications, if anything they'll just be a tiny bit smaller because they can now strip unused symbols.

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -141,9 +141,13 @@ typedef int (*secp256k1_nonce_function)(
 # define SECP256K1_NO_BUILD
 #endif
 
+/** At secp256k1 build-time DLL_EXPORT is defined when building objects destined
+ *  for a shared library, but not for those intended for static libraries.
+ */
+
 #ifndef SECP256K1_API
 # if defined(_WIN32)
-#  ifdef SECP256K1_BUILD
+#  if defined(SECP256K1_BUILD) && defined(DLL_EXPORT)
 #   define SECP256K1_API __declspec(dllexport)
 #  else
 #   define SECP256K1_API


theuni:
In case it's not obvious because a `git grep DLL_EXPORT` will come up otherwise empty...

It's a libtool convention. `make V=1` shows libtool appending it:

> libtool: compile:  x86_64-w64-mingw32-gcc -DHAVE_CONFIG_H -I. -I./src -I./include -I./src -I/home/cory/dev/bitcoin2/depends/x86_64-w64-mingw32/include/ -O2 -std=c89 -pedantic -Wno-long-long -Wnested-externs -Wshadow -Wstrict-prototypes -Wundef -Wno-overlength-strings -Wall -Wno-unused-function -Wextra -Wcast-align -Wcast-align=strict -fvisibility=hidden -pipe -O2 -MT src/libsecp256k1_la-secp256k1.lo -MD -MP -MF src/.deps/libsecp256k1_la-secp256k1.Tpo -c src/secp256k1.c  **-DDLL_EXPORT** -DPIC -o src/.libs/libsecp256k1_la-secp256k1.o

theuni:
It turns out the same thing was happening for non-windows libraries as well, but I wasn't seeing it because behavior was masked by an unrelated linker flag.

I've gone ahead and added a commit to fixup the non-windows paths as well. Title has been updated accordingly.

I can squash the commits if they make more sense that way.

sipa:
You have angered the CI gods.

theuni:
> You have angered the CI gods.

Yeah, this might take some exploring. I pushed a fix to de-grumpify the gods, but I'm not sure if it's a good idea as the current behavior might be intended.

Converting to a draft while I poke around.

real-or-random:
Not sure what other changes made compilation on CI fail but Concept ACK 6f6cab9989a4d3f4a28e3cdbfacc4e3e1e55c843. This should be entirely harmless.

fanquake:
Concept ACK

hebasto:
Concept ACK.

theuni:
> Not sure what other changes made compilation on CI fail but Concept ACK [6f6cab9](https://github.com/bitcoin-core/secp256k1/commit/6f6cab9989a4d3f4a28e3cdbfacc4e3e1e55c843). This should be entirely harmless.

Thanks for having a look. Agreed the first commit (the one that's there now) is correct and I'll leave the non-Windows side alone for now.

I may follow-up with a PR for an escape-hatch define for static libs if we need one for Core, but only if it turns out to actually be necessary.

Sorry for the noise, marking as ready for review again.

sipa:
utACK 6f6cab9989a4d3f4a28e3cdbfacc4e3e1e55c843

laanwj:
This makes sense (also for non-windows platforms, but i guess we'll get to that later).
utACK 6f6cab9989a4d3f4a28e3cdbfacc4e3e1e55c843

PR #1114, opened 2022-06-29
`_scratch_destroy`: move `VERIFY_CHECK` after invalid scrach space check

siv2r:
If we pass an invalid scratch pointer to `secp256k1_scratch_destroy`, the `VERIFY_CHECK` will probably fail, and libsecp will crash with the message
```bash
src/scratch_impl.h:28: test condition failed: scratch->alloc_size == 0
```

If `VERIFY_CHECK` after the invalid scratch space check, libsecp will call the error callback function with the "invalid scratch space" message instead.


jonasnick (commented):
ACK 1827c9bf2b7d4bf6a78c2f387bdb3e885a71e292

sipa:
utACK 1827c9bf2b7d4bf6a78c2f387bdb3e885a71e292

PR #1116, opened 2022-06-30
build: Fix #include "..." paths to get rid of further -I arguments

real-or-random:
This simplifies building without a build system.

This is in line with #925; the paths fixed here were either forgotten
there or only introduced later. This commit also makes the Makefile
stricter so that further "wrong" #include paths will lead to build
errors even in autotools builds.

This belongs to #929.

### DIFF
--- a/src/modules/extrakeys/tests_exhaustive_impl.h
+++ b/src/modules/extrakeys/tests_exhaustive_impl.h
@@ -7,8 +7,8 @@
 #ifndef SECP256K1_MODULE_EXTRAKEYS_TESTS_EXHAUSTIVE_H
 #define SECP256K1_MODULE_EXTRAKEYS_TESTS_EXHAUSTIVE_H
 
-#include "src/modules/extrakeys/main_impl.h"
 #include "../../../include/secp256k1_extrakeys.h"
+#include "main_impl.h"
 
 static void test_exhaustive_extrakeys(const secp256k1_context *ctx, const secp256k1_ge* group) {
     secp256k1_keypair keypair[EXHAUSTIVE_TEST_ORDER - 1];
--- a/src/tests_exhaustive.c
+++ b/src/tests_exhaustive.c
@@ -342,15 +342,15 @@ void test_exhaustive_sign(const secp256k1_context *ctx, const secp256k1_ge *grou
 }
 
 #ifdef ENABLE_MODULE_RECOVERY
-#include "src/modules/recovery/tests_exhaustive_impl.h"
+#include "modules/recovery/tests_exhaustive_impl.h"
 #endif
 
 #ifdef ENABLE_MODULE_EXTRAKEYS
-#include "src/modules/extrakeys/tests_exhaustive_impl.h"
+#include "modules/extrakeys/tests_exhaustive_impl.h"
 #endif
 
 #ifdef ENABLE_MODULE_SCHNORRSIG
-#include "src/modules/schnorrsig/tests_exhaustive_impl.h"
+#include "modules/schnorrsig/tests_exhaustive_impl.h"
 #endif
 
 int main(int argc, char** argv) {
--- a/Makefile.am
+++ b/Makefile.am
@@ -87,7 +87,7 @@ endif
 endif
 
 libsecp256k1_la_SOURCES = src/secp256k1.c
-libsecp256k1_la_CPPFLAGS = -I$(top_srcdir)/include -I$(top_srcdir)/src $(SECP_INCLUDES)
+libsecp256k1_la_CPPFLAGS = $(SECP_INCLUDES)
 libsecp256k1_la_LIBADD = $(SECP_LIBS) $(COMMON_LIB) $(PRECOMPUTED_LIB)
 libsecp256k1_la_LDFLAGS = -no-undefined -version-info $(LIB_VERSION_CURRENT):$(LIB_VERSION_REVISION):$(LIB_VERSION_AGE)
 
@@ -112,7 +112,7 @@ TESTS =
 if USE_TESTS
 noinst_PROGRAMS += tests
 tests_SOURCES = src/tests.c
-tests_CPPFLAGS = -I$(top_srcdir)/src -I$(top_srcdir)/include $(SECP_INCLUDES) $(SECP_TEST_INCLUDES)
+tests_CPPFLAGS = $(SECP_INCLUDES) $(SECP_TEST_INCLUDES)
 if VALGRIND_ENABLED
 tests_CPPFLAGS += -DVALGRIND
 noinst_PROGRAMS += valgrind_ctime_test
--- a/src/modules/ecdh/bench_impl.h
+++ b/src/modules/ecdh/bench_impl.h
@@ -7,7 +7,7 @@
 #ifndef SECP256K1_MODULE_ECDH_BENCH_H
 #define SECP256K1_MODULE_ECDH_BENCH_H
 
-#include "../include/secp256k1_ecdh.h"
+#include "../../../include/secp256k1_ecdh.h"
 
 typedef struct {
     secp256k1_context *ctx;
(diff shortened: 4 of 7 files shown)


src/tests_exhaustive.c
```diff
@@ -342,15 +342,15 @@ void test_exhaustive_sign(const secp256k1_context *ctx, const secp256k1_ge *grou
 }
 
 #ifdef ENABLE_MODULE_RECOVERY
-#include "src/modules/recovery/tests_exhaustive_impl.h"
+#include "modules/recovery/tests_exhaustive_impl.h"
```
hebasto, line 345: Was about suggesting the same while working on #1113 :)

hebasto (commented):
Concept ACK.

src/modules/extrakeys/tests_exhaustive_impl.h
```diff
@@ -7,8 +7,8 @@
 #ifndef SECP256K1_MODULE_EXTRAKEYS_TESTS_EXHAUSTIVE_H
 #define SECP256K1_MODULE_EXTRAKEYS_TESTS_EXHAUSTIVE_H
 
-#include "src/modules/extrakeys/main_impl.h"
 #include "../../../include/secp256k1_extrakeys.h"
+#include "main_impl.h"
```
hebasto, line 11: Why add this a new header while `#include "../../modules/extrakeys/main_impl.h"` is enough?

hebasto:
Should this diff be added as well
```diff
--- a/src/modules/recovery/tests_exhaustive_impl.h
+++ b/src/modules/recovery/tests_exhaustive_impl.h
@@ -7,7 +7,7 @@
 #ifndef SECP256K1_MODULE_RECOVERY_EXHAUSTIVE_TESTS_H
 #define SECP256K1_MODULE_RECOVERY_EXHAUSTIVE_TESTS_H
 
-#include "src/modules/recovery/main_impl.h"
+#include "main_impl.h"
 #include "../../../include/secp256k1_recovery.h"
 
 void test_exhaustive_recovery_sign(const secp256k1_context *ctx, const secp256k1_ge *group) {
```
?

sipa:
@real-or-random Bitcoin Core uses `#include <...>` style exclusively, even for the project's own files, because they're never relative w.r.t. the source file path. Is that something which would be useful for us too?

real-or-random:
In order to simplify building without build system (#929), we tried to get rid of as many mandatory compiler args as possible. #925 tried to get rid of `-I` but missed a few cases. This is fixed by this PR.

Changing everything to `<>` would be a switch of paradigms. I think in the end there's no definitive answer (see https://stackoverflow.com/a/72065059 for a nice discussion of pros and cons) but I think simple manual builds is a useful goal for us (e.g. for embedded users), so I think we should stick to our `""`-only convention unless we have a good reason to change to `<>`.

sipa:
ACK ac39773ba97f2afd0d84a2b3954e3cdc0a7594f7

hebasto:
Removing of `-I$(top_srcdir)/include` here made me believe that `${PROJECT_SOURCE_DIR}/include` can be [dropped](https://github.com/bitcoin-core/secp256k1/pull/1113#discussion_r911167739) in #1113 as well.

But that is not the case. Two headers cannot be found (when combining this PR and #1113):
https://github.com/bitcoin-core/secp256k1/blob/43756da819a235d813e7ecd53eae6df073b53247/src/modules/ecdh/bench_impl.h#L10 and https://github.com/bitcoin-core/secp256k1/blob/43756da819a235d813e7ecd53eae6df073b53247/src/modules/recovery/bench_impl.h#L10

Wondering why those headers are found using Autotools?

FWIW, all other `#include`s in the `modules` directory use correct path:
```
$ git grep include/secp256k1 -- src/modules/*.h
src/modules/ecdh/bench_impl.h:#include "../include/secp256k1_ecdh.h"
src/modules/ecdh/main_impl.h:#include "../../../include/secp256k1_ecdh.h"
src/modules/extrakeys/main_impl.h:#include "../../../include/secp256k1.h"
src/modules/extrakeys/main_impl.h:#include "../../../include/secp256k1_extrakeys.h"
src/modules/extrakeys/tests_exhaustive_impl.h:#include "../../../include/secp256k1_extrakeys.h"
src/modules/extrakeys/tests_impl.h:#include "../../../include/secp256k1_extrakeys.h"
src/modules/recovery/bench_impl.h:#include "../include/secp256k1_recovery.h"
src/modules/recovery/main_impl.h:#include "../../../include/secp256k1_recovery.h"
src/modules/recovery/tests_exhaustive_impl.h:#include "../../../include/secp256k1_recovery.h"
src/modules/schnorrsig/bench_impl.h:#include "../../../include/secp256k1_schnorrsig.h"
src/modules/schnorrsig/main_impl.h:#include "../../../include/secp256k1.h"
src/modules/schnorrsig/main_impl.h:#include "../../../include/secp256k1_schnorrsig.h"
src/modules/schnorrsig/tests_exhaustive_impl.h:#include "../../../include/secp256k1_schnorrsig.h"
src/modules/schnorrsig/tests_impl.h:#include "../../../include/secp256k1_schnorrsig.h"
```

real-or-random:
@hebasto That's an "interesting" observation. Apparently autotools always add `-I. -I./src` in our case, see https://stackoverflow.com/questions/5641480/automake-overwriting-default-includes... The suggestion of overriding it doesn't work for me. I'll see if there's another way of getting rid of that.

hebasto:
@real-or-random 

As a goal of this PR is to "Fix #include "..." paths", could those [two paths](https://github.com/bitcoin-core/secp256k1/pull/1116#issuecomment-1172141261) be fixed as well?

hebasto:
> That's an "interesting" observation. Apparently autotools always add `-I. -I./src` in our case, see https://stackoverflow.com/questions/5641480/automake-overwriting-default-includes... The suggestion of overriding it doesn't work for me. I'll see if there's another way of getting rid of that.

Another reason to rid of the config header.

real-or-random:
@hebasto Okay I added these additional fixes, I hope that was everything now. 

The issue is that we can't really enforce that policy using autotools. One can add `nostdinc` to the Automake options, and this indeed removes the `-I. -I./src` but this breaks out-of-tree ( yes, this is supported by autotools, called "VPATH" builds) and more impoatantly `make distcheck`, which relies on VPATH builds... The reasons those breaks build is the config header. 

> Another reason to rid of the config header.

Yep. If we do this, then we could enforce the policy... We could also just add a "manual" build to CI. This is already on the task list in #929. But this can be done in a separate PR.

Funnily, the build worked for me locally when setting `stdinc` but without fixing `secp256k1/src/modules/(ecdh|recovery)/bench_impl.h`. This confused me a lot... Guess why? My system installation libsecp256k1: :worried:  (`-H` is really useful flag...)

<details>
<summary>before fix:  <code>gcc -DENABLE_MODULE_ECDH -DENABLE_MODULE_RECOVERY -H -c src/bench.c</code> (snippet) </summary>

```
. src/modules/ecdh/bench_impl.h
.. /usr/include/secp256k1_ecdh.h
... /usr/include/secp256k1.h
. src/modules/recovery/bench_impl.h
.. /usr/include/secp256k1_recovery.h
. src/modules/schnorrsig/bench_impl.h
.. src/modules/schnorrsig/../../../include/secp256k1_schnorrsig.h
... src/modules/schnorrsig/../../../include/secp256k1.h
... src/modules/schnorrsig/../../../include/secp256k1_extrakeys.h
```
</details>

<details>
<summary>after fix:  <code>gcc -DENABLE_MODULE_ECDH -DENABLE_MODULE_RECOVERY -H -c src/bench.c</code> (snippet) </summary>

```
. src/modules/ecdh/bench_impl.h
.. src/modules/ecdh/../../../include/secp256k1_ecdh.h
... src/modules/ecdh/../../../include/secp256k1.h
. src/modules/recovery/bench_impl.h
.. src/modules/recovery/../../../include/secp256k1_recovery.h
... src/modules/recovery/../../../include/secp256k1.h
. src/modules/schnorrsig/bench_impl.h
.. src/modules/schnorrsig/../../../include/secp256k1_schnorrsig.h
... src/modules/schnorrsig/../../../include/secp256k1.h
... src/modules/schnorrsig/../../../include/secp256k1_extrakeys.h
```
</details>

PR #1118, opened 2022-07-03
Add x-only ecmult_const version with x specified as n/d

sipa:
This implements a generalization of Peter Dettman's sqrt-less x-only random-base multiplication algorithm from #262, using the Jacobi symbol algorithm from #979. The generalization is to permit the X coordinate of the base point to be specified as a fraction $n/d$:

To compute $x(q \cdot P)$, where $x(P) = n/d$:
* Compute $g=n^3 + 7d^3$.
* Let $P' = (ng, g^2, 1)$ (the Jacobian coordinates of $P$ mapped to the isomorphic curve $y^2 = x^3 + 7(dg)^3$).
* Compute the Jacobian coordinates $(X',Y',Z') = q \cdot P'$ on the isomorphic curve.
* Return $X'/(dgZ'^2)$, which is the affine x coordinate on the isomorphic curve $X/Z'^2$ mapped back to secp256k1.

This ability to specify the X coordinate as a fraction is useful in the context of x-only [Elligator Swift](https://eprint.iacr.org/2022/759), which can decode to X coordinates on the curve without inversions this way.

### DIFF
--- a/src/ecmult_const.h
+++ b/src/ecmult_const.h
@@ -18,4 +18,25 @@
  */
 static void secp256k1_ecmult_const(secp256k1_gej *r, const secp256k1_ge *a, const secp256k1_scalar *q, int bits);
 
+/**
+ * Same as secp256k1_ecmult_const, but takes in an x coordinate of the base point
+ * only, specified as fraction n/d (numerator/denominator). Only the x coordinate of the result is
+ * returned.
+ *
+ * If known_on_curve is 0, a verification is performed that n/d is a valid X
+ * coordinate, and 0 is returned if not. Otherwise, 1 is returned.
+ *
+ * d being NULL is interpreted as d=1. If non-NULL, d must not be zero. q must not be zero.
+ *
+ * Constant time in the value of q, but not any other inputs.
+ */
+static int secp256k1_ecmult_const_xonly(
+    secp256k1_fe *r,
+    const secp256k1_fe *n,
+    const secp256k1_fe *d,
+    const secp256k1_scalar *q,
+    int bits,
+    int known_on_curve
+);
+
 #endif /* SECP256K1_ECMULT_CONST_H */
--- a/src/ecmult_const_impl.h
+++ b/src/ecmult_const_impl.h
@@ -228,4 +228,139 @@ static void secp256k1_ecmult_const(secp256k1_gej *r, const secp256k1_ge *a, cons
     secp256k1_fe_mul(&r->z, &r->z, &Z);
 }
 
+static int secp256k1_ecmult_const_xonly(secp256k1_fe* r, const secp256k1_fe *n, const secp256k1_fe *d, const secp256k1_scalar *q, int bits, int known_on_curve) {
+
+    /* This algorithm is a generalization of Peter Dettman's technique for
+     * avoiding the square root in a random-basepoint x-only multiplication
+     * on a Weierstrass curve:
+     * https://mailarchive.ietf.org/arch/msg/cfrg/7DyYY6gg32wDgHAhgSb6XxMDlJA/
+     *
+     *
+     * === Background: the effective affine technique ===
+     *
+     * Let phi_u be the isomorphism that maps (x, y) on secp256k1 curve y^2 = x^3 + 7 to
+     * x' = u^2*x, y' = u^3*y on curve y'^2 = x'^3 + u^6*7. This new curve has the same order as
+     * the original (it is isomorphic), but moreover, has the same addition/doubling formulas, as
+     * the curve b=7 coefficient does not appear in those formulas (or at least does not appear in
+     * the formulas implemented in this codebase, both affine and Jacobian). See also Example 9.5.2
+     * in https://www.math.auckland.ac.nz/~sgal018/crypto-book/ch9.pdf.
+     *
+     * This means any linear combination of secp256k1 points can be computed by applying phi_u
+     * (with non-zero u) on all input points (including the generator, if used), computing the
+     * linear combination on the isomorphic curve (using the same group laws), and then applying
+     * phi_u^{-1} to get back to secp256k1.
+     *
+     * Switching to Jacobian coordinates, note that phi_u applied to (X, Y, Z) is simply
+     * (X, Y, Z/u). Thus, if we want to compute (X1, Y1, Z) + (X2, Y2, Z), with identical Z
+     * coordinates, we can use phi_Z to transform it to (X1, Y1, 1) + (X2, Y2, 1) on an isomorphic
+     * curve where the affine addition formula can be used instead.
+     * If (X3, Y3, Z3) = (X1, Y1) + (X2, Y2) on that curve, then our answer on secp256k1 is
+     * (X3, Y3, Z3*Z).
+     *
+     * This is the effective affine technique: if we have a linear combination of group elements
+     * to compute, and all those group elements have the same Z coordinate, we can simply pretend
+     * that all those Z coordinates are 1, perform the computation that way, and then multiply the
+     * original Z coordinate back in.
+     *
+     * The technique works on any a=0 short Weierstrass curve. It is possible to generalize it to
+     * other curves too, but there the isomorphic curves will have different 'a' coefficients,
+     * which typically does affect the group laws.
+     *
+     *
+     * === Avoiding the square root for x-only point multiplication ===
+     *
+     * In this function, we want to compute the X coordinate of q*(n/d, y), for
+     * y = sqrt((n/d)^3 + 7). Its negation would also be a valid Y coordinate, but by convention
+     * we pick whatever sqrt returns (which we assume to be a deterministic function).
+     *
+     * Let g = y^2*d^3 = n^3 + 7*d^3. This also means y = sqrt(g/d^3).
+     * Further let v = sqrt(d*g), which must exist as d*g = y^2*d^4 = (y*d^2)^2.
+     *
+     * The input point (n/d, y) also has Jacobian coordinates:
+     *
+     *     (n/d, y, 1)
+     *   = (n/d * v^2, y * v^3, v)
+     *   = (n/d * d*g, y * sqrt(d^3*g^3), v)
+     *   = (n/d * d*g, sqrt(y^2 * d^3*g^3), v)
+     *   = (n*g, sqrt(g/d^3 * d^3*g^3), v)
+     *   = (n*g, sqrt(g^4), v)
+     *   = (n*g, g^2, v)
+     *
+     * It is easy to verify that both (n*g, g^2, v) and its negation (n*g, -g^2, v) have affine X
+     * coordinate n/d, and this holds even when the square root function doesn't have a
+     * determinstic sign. We choose the (n*g, g^2, v) version.
+     *
+     * Now switch to the effective affine curve using phi_v, where the input point has coordinates
+     * (n*g, g^2). Compute (X, Y, Z) = q * (n*g, g^2) there.
+     *
+     * Back on secp256k1, that means q * (n*g, g^2, v) = (X, Y, v*Z). This last point has affine X
+     * coordinate X / (v^2*Z^2) = X / (d*g*Z^2). Determining the affine Y coordinate would involve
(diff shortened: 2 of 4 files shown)


peterdettman:
Hey @sipa, cool idea. Should that just be sqrt(g) in the second step?

sipa:
@peterdettman I don't think so.

`x = X / Z^2 = n*g / sqrt(d*g)^2 = n*g / (d*g) = n/d`
`y = Y / Z^2 = g^2 / sqrt(d*g)^3 = (n^3 + B*d^3)^2 / sqrt(d^3) / (n^3 + B*d^3)^1.5 = sqrt(n^3 + B*d^3) / sqrt(d^3) = sqrt(n^3/d^3 + B)`

peterdettman:
OK, well it's late here, so I'll go over it when not so tired. A followup question: is there an assumption that d is square? If not, a better starting point might be (y*d^3)^2 =(x*d^2)^3 + B*d^6 i.e. g = (n.d)^3 + B*d^6, so that this will not be a quadratic twist if d is non-square.

sipa:
@peterdettman It works for d square or non-square (obviously not for d=0). The reason is that `g = y^2*d^3`, and thus `d*g = y^2*d^4`, which is always square if `n/d` is an x coordinate on the curve.

peterdettman:
Oh I see you meant that `(n*g, +- g^2, sqrt(d*g))` is on the original curve, not the one implied in step 1.

sipa:
Yeah, so just operationally, the algorithm is:
* Input $n$, $d$, $q$
* Compute $g = n^3 + Bd^3$
* Construct affine point $P = (ng, g^2)$ on isomorphic curve
* Compute Jacobian result point $R = (X, Y, Z) = q \cdot P$ on isomorphic curve
* Return $X / (Z^2 d g)$, which is the affine x coordinate on the original curve

src/ecmult_const_impl.h
```diff
@@ -228,4 +228,58 @@ static void secp256k1_ecmult_const(secp256k1_gej *r, const secp256k1_ge *a, cons
...
+        secp256k1_fe_mul(&b, &b, d);
+        secp256k1_fe_mul(&b, &b, &secp256k1_fe_const_b);
+        secp256k1_fe_add(&g, &b);
+        if (!known_on_curve) {
+            secp256k1_fe c;
+            secp256k1_fe_mul(&c, &g, d);
+            if (secp256k1_fe_jacobi_var(&c) < 0) return 0;
+        }
+    } else {
+        secp256k1_fe_add(&g, &secp256k1_fe_const_b);
```
real-or-random, line 257: not this PR: Now that I see this, I wonder if we should get rid of `secp256k1_fe_const_b` and simply call `secp256k1_fe_set_int`... I don't think it makes a measurable difference in performance (or it may even be faster).

real-or-random (commented):
Concept ACK

I might suggest some additions to the comments that may have helped me understanding. (I can't remember, would this be the first "effective-affine" thing we merge?)

Would it make sense to tests this in exhaustive tests?

Can you rebase on #979?

src/ecmult_const_impl.h
```diff
@@ -228,4 +228,58 @@ static void secp256k1_ecmult_const(secp256k1_gej *r, const secp256k1_ge *a, cons
...
+    secp256k1_gej rj;
+
+    /* Compute g = (n^3 + B*d^3). */
+    secp256k1_fe_sqr(&g, n);
+    secp256k1_fe_mul(&g, &g, n);
+    if (d) {
+        secp256k1_fe b;
+        secp256k1_fe_sqr(&b, d);
+        secp256k1_fe_mul(&b, &b, d);
+        secp256k1_fe_mul(&b, &b, &secp256k1_fe_const_b);
```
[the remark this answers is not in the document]
real-or-random: Oh, I missed that fact.

src/ecmult_const_impl.h
```diff
@@ -228,4 +228,118 @@ static void secp256k1_ecmult_const(secp256k1_gej *r, const secp256k1_ge *a, cons
...
+     *
+     *
+     * Background: the effective affine technique
+     *
+     * Let phi_u be the isomorphism that maps (x, y) on secp256k1 curve y^2 = x^3 + 7 to
+     * x' = u^2*x, y' = u^3*y on curve y'^2 = x'^3 + u^6*7. This new curve has the same order as
+     * the original (it is isomorphic), but moreover, has the same addition/doubling formulas, as
+     * the curve b=7 coefficient does not appear in those formulas.
+     *
+     * This means any computation on secp256k1 can be peformed by applying phi_u for any non-zero
```
real-or-random, line 246: "performed"

real-or-random (commented):
That explanation is very clear indeed!

src/tests.c
```diff
@@ -4283,6 +4351,68 @@ void ecmult_const_mult_zero_one(void) {
     ge_equals_ge(&res2, &point);
 }
 
+void ecmult_const_mult_xonly(void) {
+    int i;
+
+    /* Test correspondence between secp256k1_ecmult_const and secp256k1_ecmult_const_xonly. */
+    for (i = 0; i < 2*count; ++i) {
```
real-or-random, line 4358:
```suggestion
    for (i = 0; i < 2*COUNT; ++i) {
```

src/tests.c
```diff
@@ -4283,6 +4351,68 @@ void ecmult_const_mult_zero_one(void) {
...
+        secp256k1_gej_set_ge(&basej, &base);
+        secp256k1_ecmult(&resj, &basej, &q, NULL);
+        /* Check that resj's X coordinate corresponds with resx. */
+        secp256k1_fe_sqr(&v, &resj.z);
+        secp256k1_fe_mul(&v, &v, &resx);
+        CHECK(check_fe_equal(&v, &resj.x));
+    }
+
+    /* Test that secp256k1_ecmult_const_xonly correctly rejects X coordinates not on curve. */
+    for (i = 0; i < 2*count; ++i) {
```
real-or-random, line 4390:
```suggestion
    for (i = 0; i < 2*COUNT; ++i) {
```

real-or-random (commented):
that should fix the compilation failures on CI

src/tests_exhaustive.c
```diff
@@ -59,6 +59,19 @@ void random_fe(secp256k1_fe *x) {
...
+    int tries = 10;
+    while (--tries >= 0) {
+        random_fe(nz);
+        secp256k1_fe_normalize(nz);
+        if (!secp256k1_fe_is_zero(nz)) {
+            break;
+        }
+    }
+    /* Infinitesimal probability of spurious failure here */
+    CHECK(tries >= 0);
```
real-or-random, line 73: (Is `random_fe` biased towards zero? I don't think so.)

real-or-random (commented):
ACK on the last two commits

src/tests_exhaustive.c
real-or-random: Okay, I was about to say: then just use `random_fe` and don't bother with zero at all? 

but if this is stolen from the other file, then let's not touch it, yes.

real-or-random:
needs rebase :)

src/ecmult_const_impl.h
```diff
@@ -228,4 +228,127 @@ static void secp256k1_ecmult_const(secp256k1_gej *r, const secp256k1_ge *a, cons
...
+    secp256k1_ge p;
+    secp256k1_gej rj;
+
+    /* Compute g = (n^3 + B*d^3). */
+    secp256k1_fe_sqr(&g, n);
+    secp256k1_fe_mul(&g, &g, n);
+    if (d) {
+        secp256k1_fe b;
+        secp256k1_fe_sqr(&b, d);
+        secp256k1_fe_mul_int(&b, SECP256K1_B);
```
real-or-random, line 309: `b` now has magnitude `SECP256K1_B == 7`, which happens to be just below `8`, the maximum for the following `fe_mul` call.

So, we better don't switch to a different curve in the future. ;) On a more serious note, I don't know how much our code depends on B, but I think it's a good idea to avoid introducing further dependencies on the B in the code base. 

If you agree, we could do this: 

```suggestion
        VERIFY_CHECK(SECP256K1_B <= 8); /* magnitude of b will be <= 8 after the next call */
        secp256k1_fe_mul_int(&b, SECP256K1_B);
```

Of course, the same is already checked now in `fe_verify` but this makes it a bit more explicit.

src/ecmult_const_impl.h
```diff
@@ -228,4 +228,127 @@ static void secp256k1_ecmult_const(secp256k1_gej *r, const secp256k1_ge *a, cons
...
+     * sqrt(d*g) must exist if the input was valid, as d*g = y^2*d^4 = (y*d^2)^2.
+     *
+     * The input point (n/d, y) also has Jacobian coordinates:
+     *   (n/d, y, 1)
+     * = (n/d * sqrt^2(d*g), y * sqrt^3(d*g), sqrt(d*g))
+     * = (n/d * d*g, y * +-sqrt(d^3*g^3), sqrt(d*g))
+     * = (n/d * d*g, +-sqrt(g/d^3) * +-sqrt(d^3*g^3), sqrt(d*g))
+     * = (n*g, +-sqrt(g/d^3 * d^3*g^3), sqrt(d*g))
+     * = (n*g, +-sqrt(g^4), sqrt(d*g))
+     * = (n*g, +-g^2, sqrt(d*g))
```
real-or-random, line 285:
```suggestion
     * The input point (n/d, y) also has Jacobian coordinates:
     *   (n/d, y, 1)
     * = (n/d * sqrt^2(d*g), y * sqrt^3(d*g), sqrt(d*g))
     * = (n/d * d*g, y * sqrt(d^3*g^3), sqrt(d*g))
     * = (n/d * d*g, +-sqrt(g/d^3) * sqrt(d^3*g^3), sqrt(d*g))
     * = (n*g, +-sqrt(g/d^3 * d^3*g^3), sqrt(d*g))
     * = (n*g, +-sqrt(g^4), sqrt(d*g))
     * = (n*g, +-g^2, sqrt(d*g))
```

real-or-random (commented):
ACK mod nits

src/ecmult_const_impl.h
real-or-random: Ok, yes... 

(indepedent nit: adding more space left to the equation would make it more readable. The `*` from the comments make it hard to read.)

real-or-random (approved):
ACK c13f877da5f9de008a181df1bfb24342a2799b9b

real-or-random (approved):
ACK a6412d2903d3e057d3806bf0132a4e4a62136f9c

real-or-random (approved):
ACK 9c69617478c780233fe24cd7f9ada65f5c8d8763

src/ecmult_const.h
```diff
@@ -18,4 +18,25 @@
...
+ * Constant time in the value of q, but not any other inputs.
+ */
+static int secp256k1_ecmult_const_xonly(
+    secp256k1_fe* r,
+    const secp256k1_fe *n,
+    const secp256k1_fe *d,
+    const secp256k1_scalar *q,
+    int bits,
+    int known_on_curve
+);
```
jonasnick, line 40: I think we should mention the `*q == 0` edge case. Right now `secp256k1_ecmult_const_xonly` outputs `*r == 0`, which is maybe a reasonable value in that case but not an X coordinate on the curve. We can either disallow `*q == 0` and VERIFY_CHECK it or document that it outputs invalid X-coordinate 0 (and also test this edge case in the regular tests). I'd prefer the former because outputting an X-coordinate that's not on the curve seems confusing and it relies on our definition that `0^-1 = 0` (see `secp256k1_fe_inv(&i, &i);` in the function).

jonasnick (commented):
Concept ACK. Thanks for the detailed description of the technique.

jonasnick (commented):
ACK mod nit

jonasnick (approved):
ACK 0f8642079b0f2e4874393792f5854e3c33742cbd

real-or-random (approved):
ACK 0f8642079b0f2e4874393792f5854e3c33742cbd

alokeutpal (approved):

PR #1121, opened 2022-07-06
config: Set preprocessor defaults for ECMULT_* config values

real-or-random:
This simplifies manual builds and solves one item in #929.

Manual compilation now:
 ```
 gcc -O2 -c src/secp256k1.c src/precomputed_*.c
 ar rcs libsecp256k1.a secp256k1.o precomputed_*.o
```

The second commit will make debugging the config easier when passing `-DCONFIG_DEBUG`. It currently only prints the `ECMULT_*` variables but we could use it for `WIDEMUL` and also synthetic int128 stuff in the future after #1000 is merge. The logic there is more complex than simply falling back to a default value. 

But I'm happy to drop that commit if people don't like this or think it's overkill. 

edit: The third commit removes basic-config.h which is now definitively not necessary anymore because we have default values.

---

My long-term plan is to get rid of the config header "src/libsecp256k1-config.h" currently generated by autotools and simply let autotools pass all config macros via `-D` switches.   Generated header files are annoying, and could then also be avoided in #1113. See https://github.com/bitcoin-core/secp256k1/pull/1113#issuecomment-1171307979 for more context.


sipa:
Concept ACK, and also agree with the intent to get rid of the config files entirely.

One thing I'd like to bring up though: I think it may be useful to have a strict separation between "external" macros (provided in config.h, or through compiler -D flags) flags, and "internal" macros which guide the actual source code conditionals? I can imagine having a single file that does all conversion between the two (and e.g. contains all logic for autodetecting endianness, integer sizes, plus looks at external variables for overriding/guiding them). A downside would be that this file ends up containing logic relating to many aspects of the codebase, while those decisions are now closer to where they're actually used, but having everything together also gives a much better overview.

If we go that direction, it doesn't need to happen now of course, but I wanted to bring up in case we want to start working in that direction already.

sipa:
> A simple way may be to have a prefix for all external/input macros, e.g., CONFIG_. (We could still accept the non-prefixed form for backwards-compability).

Do you mean CONFIG_ as prefix for the internal ones, or the external ones? In either case, perhaps you don't want to use the names `CONFIG_DEBUG`, `CONFIG_DEBUG_MSG`, `CONFIG_DEBUG_DEF` in this PR, as one of those is external and the two others are internal?

real-or-random:
> > A simple way may be to have a prefix for all external/input macros, e.g., CONFIG_. (We could still accept the non-prefixed form for backwards-compability).
> 
> Do you mean CONFIG_ as prefix for the internal ones, or the external ones? 

I meant `CONFIG` for the external ones, because this tells the user "if it starts with `CONFIG`, I'm allowed to set it".  


> In either case, perhaps you don't want to use the names `CONFIG_DEBUG`, `CONFIG_DEBUG_MSG`, `CONFIG_DEBUG_DEF` in this PR, as one of those is external and the two others are internal?

That's a very good point... I'll change it.

sipa:
> I meant `CONFIG` for the external ones, because this tells the user "if it starts with `CONFIG`, I'm allowed to set it".  

SGTM

hebasto (approved):
ACK c27ae451440bdaf68bf8aaa60edb1f4b4614d492, I have reviewed the code and it looks correct.

Tested with `./configure CPPFLAGS=-DDEBUG_CONFIG`.

However,

> The second commit will make debugging the config easier when passing `-DCONFIG_DEBUG`. It currently only prints the `ECMULT_*` variables but we could use it for `WIDEMUL` and also synthetic int128 stuff in the future after #1000 is merge. The logic there is more complex than simply falling back to a default value.

Even now debug messages are noisy as they repeat for every translation unit which includes a relevant header directly or indirectly. Adding more macros will make build logs messier. This approach does not look scalable to me.

sipa:
> Even now debug messages are noisy as they repeat for every translation unit which includes a relevant header directly or indirectly. Adding more macros will make build logs messier. This approach does not look scalable to me.

I don't think that's a concern. You don't need to use it, as it's just for helping debug configuration issues, and isn't enabled by default. How does scalability matter? Do you expect more macros like this?

hebasto:
> Do you expect more macros like this?

> It currently only prints the `ECMULT_*` variables but we could use it for `WIDEMUL` and also synthetic int128 stuff in the future after #1000 is merge.

hebasto:
> That's about extended the output to more configuration macros...

Yes, exactly. Sorry for not being clear.

real-or-random:
> I don't think that's a concern. You don't need to use it, as it's just for helping debug configuration issues, and isn't enabled by default. How does scalability matter? Do you expect more macros like this?

Yep I agree. It's noisy but the point is to be noisy *when we ask for it*. I don't think build log clutter is an issue. I see this as an option for developers and maintainers. You'll only enable this if you want to debug the config log.

jonasnick (commented):
ACK c27ae451440bdaf68bf8aaa60edb1f4b4614d492

Tested various configurations. I didn't find the duplicate output of `DEBUG_CONFIG` too confusing; it's only a debug tool after all.

PR #1122, opened 2022-07-07
tests: Randomize the context with probability 15/16 instead of 1/4

real-or-random:
It's not perfect but I'm convinced it's a useful improvement for now.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -7086,11 +7086,15 @@ int main(int argc, char **argv) {
     run_context_tests(0);
     run_context_tests(1);
     run_scratch_tests();
+
     ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN | SECP256K1_CONTEXT_VERIFY);
-    if (secp256k1_testrand_bits(1)) {
+    /* Randomize the context only with probability 15/16
+       to make sure we test without context randomization from time to time.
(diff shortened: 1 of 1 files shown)


src/tests.c
```diff
@@ -7086,11 +7086,14 @@ int main(int argc, char **argv) {
     run_context_tests(0);
     run_context_tests(1);
     run_scratch_tests();
+
     ctx = secp256k1_context_create(SECP256K1_CONTEXT_SIGN | SECP256K1_CONTEXT_VERIFY);
-    if (secp256k1_testrand_bits(1)) {
+    /* Randomize the context only with probability 15/16
+       to make sure we test without context randomization from time to time.
+       TODO Reconsider this when recalibrating the tests. */
+    if (!secp256k1_testrand_bits(4)) {
```
sipa, line 7094: I think this implements randomization in 1/16 of the cases.

real-or-random: Ha, this happens if you want to do a "quick" change...

Fixed.

> Also, it randomizes with uninitialized memory?

Yeah, don't ask me why I removed this line.

sipa:
ACK 17065f48ae261c6949dab74a7c197ac13b52eb1b

jonasnick (commented):
ACK 17065f48ae261c6949dab74a7c197ac13b52eb1b

PR #1129, opened 2022-07-21
ElligatorSwift + integrated x-only DH

sipa:
Builds on top of #979, #1118. Replaces #982.

This implements encoding of curve points using the ElligatorSwift algorithm, using 4 new API calls:
* `secp256k1_ellswift_encode`, which converts a public key to a 64-byte pseudorandom encoding.
* `secp256k1_ellswift_decode`, the reverse operation to convert back to normal public keys.
* `secp256k1_ellswift_create`, which can be seen as a combination of `secp256k1_ec_pubkey_create` + `secp256k1_ellswift_encode`, but is somewhat safer.
* `secp256k1_ellswift_xdh`, which implements x-only Diffie-Hellman directly on top of 64-byte encoded public keys, and more efficiently than decoding + invoking normal ECDH.

The scheme matches that of the SwiftEC paper (https://eprint.iacr.org/2022/759), with two changes (remapping undefined inputs, and encoding the Y parity in the u/t values themselves rather than in a separate bit). To decode an ElligatorSwift 64-byte encoded public key:
* Interpret the encoding as two field elements $(u, t)$ in big-endian 32-byte encoding each, reducing modulo $p$.
* If $u=0$, let $u=1$ instead.
* If $t=0$, let $t=1$ instead.
* If $u^3 + 7 + t^2 = 0$, let $t = 2t$ instead.
* Let $X = \dfrac{u^3 + 7 - t^2}{2t}$.
* Let $Y = \dfrac{X + t}{\sqrt{-3}u}$.
* Let $x$ be the first of $(u + 4Y^2, \dfrac{-X}{2Y} - \dfrac{u}{2}, \dfrac{X}{2Y} - \dfrac{u}{2})$ which is on the curve (at least one will be).
* Let $y$ be the corresponding Y coordinate, with the same parity as (the original) $t$.
* Return $(x,y)$.

This is significantly faster than the Elligator Squared code in #982.

Relevant benchmark (AMD Ryzen 5950X, GCC 12.2.0, default config options; frequency fixed at 2.80 GHz):

```

ec_keygen                     ,    27.9       ,    28.0       ,    28.0    
ecdh                          ,    53.9       ,    53.9       ,    54.0    
ellswift_encode               ,    24.5       ,    24.5       ,    24.6    
ellswift_decode               ,    11.1       ,    11.1       ,    11.1    
ellswift_keygen               ,    52.5       ,    52.5       ,    52.6    
ellswift_ecdh                 ,    57.9       ,    58.0       ,    58.0    
```


real-or-random (commented):
(just skimming)

Do you think there's a nice way to avoid code duplication between `divsteps` and `posdivsteps`?

sipa:
@real-or-random Re posdivsteps see #979 which this PR is based on.

paulmillr:
Looks awesome! Wanted to clarify this bit, for implementors of EllSwift in other languages / for standardization purpose.

>Inputs (u,t) where u=0, t=0, or u^2+t^3+7=0, are remapped to other finite points, rather than outputting infinity.

To which points exactly? It's not immediately obvious from the code.

sipa:
@paulmillr They are remapped as follows:
* If t=0, set t=1 instead
* If u=0, set u=1 instead
* If u^2+t^3+7=0, set t=2*t instead
* Run the normal algorithm

include/secp256k1_ellswift.h
```diff
@@ -0,0 +1,175 @@
...
+
+/** Given a private key, and ElligatorSwift public keys sent in both directions,
+ *  compute a shared secret using x-only Diffie-Hellman.
+ *
+ *  Returns: 1: shared secret was succesfully computed
+ *           0: secret was invalid or hashfp returned 0
+ *  Args:    ctx:        pointer to a context object.
+ *  Out:     output:     pointer to an array to be filled by hashfp.
+ *  In:      theirs64:   a pointer to the 64-byte ElligatorSquare public key received from the other party.
+ *           ours64:     a pointer to the 64-byte ElligatorSquare public key sent to the other party.
```
stratospher, line 148: s/ElligatorSquare/ElligatorSwift in `theirs64` and `ours64`?

src/modules/ellswift/main_impl.h
```diff
@@ -0,0 +1,491 @@
...
+        /* If s is not square, fail. We have not fully computed s yet, but s is square iff
+         * -(u^3+7)*(u^2+u*x+x^2) is square. */
+        secp256k1_fe_sqr(&g, &u); /* g = u^2 */
+        secp256k1_fe_mul(&g, &g, &u); /* g = u^3 */
+        secp256k1_fe_add_int(&g, SECP256K1_B); /* g = u^3+7 */
+        secp256k1_fe_mul(&m, &s, &g); /* m = -(u^3 + 7)*(u^2 + u*x + x^2) */
+        if (!secp256k1_fe_is_square_var(&m)) return 0;
+
+        /* Let s = -(u^3 + 7)/(u^2 + u*x + x^2) [second part] */
+        secp256k1_fe_inv_var(&s, &s); /* s = -1/(u^2 + u*x + x^2) */
```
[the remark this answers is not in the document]
jonasnick: I don't see why the `if (!secp256k1_fe_is_square_var(&m)) { return 0; }` path can catch this. `secp256k1_fe_is_square_var` returns 1 on input 0.

include/secp256k1_ellswift.h
```diff
@@ -0,0 +1,177 @@
...
+  void *data
+);
+
+/** An implementation of an secp256k1_ellswift_xdh_hash_function which uses
+ *  SHA256(key1 || key2 || x32), where (key1, key2) = sorted([ours64, theirs64]), and
+ *  ignores data. The sorting is lexicographic. */
+SECP256K1_API extern const secp256k1_ellswift_xdh_hash_function secp256k1_ellswift_xdh_hash_function_sha256;
+
+/** A default secp256k1_ellswift_xdh_hash_function, currently secp256k1_ellswift_xdh_hash_function_sha256. */
+SECP256K1_API extern const secp256k1_ellswift_xdh_hash_function secp256k1_ellswift_xdh_hash_function_default;
```
real-or-random, line 76: I know we have this also in the module ECDH, but what's the purpose for exporting a variable for the default? I think in case of ECDH, a hash function that can change due to an API change is exactly not what you want as a user, because it breaks interop. 

Of course, us changing this variable will be a breaking API change. But the way I imagine a change of defaults (if it will ever occur), is not changing `secp256k1_ellswift_xdh_hash_function_default` (and the behavior of `NULL`), because that's just hard breaking. I imagine we'd instead rather deprecate `secp256k1_ellswift_xdh` and add a new function with a different default hash.

So, I don't see much value in adding this default variable as another layer of abstraction. If the user wants to call the hash function explicitly, they could still call `secp256k1_ellswift_xdh_hash_function_sha256` (and be sure what they get!).

(Now that I think about it, having this in the ECDH module may have originated from the way we do this for nonce functions. But in that case, it makes much more sense, and we could swap out the hash function under the hood without any change in interop.)

real-or-random (commented):
Making my way through the commits. 

Here's a branch with fixups: https://github.com/real-or-random/secp256k1/tree/202206_ellswift_dh Currently, the only commit has some nits in the public header.

src/modules/ellswift/main_impl.h
```diff
@@ -0,0 +1,555 @@
...
+static void secp256k1_ellswift_elligatorswift_var(unsigned char *u32, secp256k1_fe *t, const secp256k1_ge *p, const secp256k1_sha256 *hasher) {
+    secp256k1_ellswift_xelligatorswift_var(u32, t, &p->x, hasher);
+    if (secp256k1_fe_is_odd(t) != secp256k1_fe_is_odd(&p->y)) {
+        secp256k1_fe_negate(t, t, 1);
+        secp256k1_fe_normalize_var(t);
+    }
+}
+
+/** Set hash state to the BIP340 tagged hash midstate for "secp256k1_ellswift_encode". */
+static void secp256k1_ellswift_sha256_init_encode(secp256k1_sha256* hash) {
```
jonasnick, line 378: nit: should we add a test for both midstate init functions similar to https://github.com/bitcoin-core/secp256k1/blob/master/src/modules/schnorrsig/tests_impl.h#L50?

src/modules/ellswift/main_impl.h
```diff
@@ -0,0 +1,491 @@
...
+    secp256k1_declassify(ctx, &p, sizeof(p)); /* not constant time in produced pubkey */
+    secp256k1_fe_normalize_var(&p.x);
+    secp256k1_fe_normalize_var(&p.y);
+
+    /* Set up hasher state */
+    secp256k1_sha256_initialize(&hash);
+    secp256k1_sha256_write(&hash, PREFIX, sizeof(PREFIX));
+    secp256k1_sha256_write(&hash, seckey32, 32);
+    secp256k1_sha256_write(&hash, rnd32 ? rnd32 : ZERO, 32);
+    secp256k1_sha256_write(&hash, ZERO, 32 - 9 - 4);
```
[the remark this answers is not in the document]
jonasnick: It's clear now.

include/secp256k1_ellswift.h
```diff
@@ -0,0 +1,170 @@
...
+ *  In:      seckey32:    pointer to a 32-byte secret key.
+ *           auxrand32:   (optional) pointer to 32 bytes of additional randomness
+ *
+ * Constant time in seckey and auxrand32, but not in the resulting public key.
+ *
+ * This function can be used instead of calling secp256k1_ec_pubkey_create followed
+ * by secp256k1_ellswift_encode. It is safer, as it can use the secret key as
+ * entropy for the encoding. That means that if the secret key itself is
+ * unpredictable, no additional auxrand32 is needed to achieve indistinguishability
+ * of the encoding.
```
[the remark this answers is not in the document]
jonasnick: Thanks for the context. The new docs sound good to me.

include/secp256k1_ellswift.h
```diff
@@ -0,0 +1,176 @@
...
+ *  Out:     output:    pointer to an array to be filled by hashfp.
+ *  In:      theirs64:  a pointer to the 64-byte ElligatorSwift public key
+ *                      received from the other party.
+ *           ours64:    a pointer to the 64-byte ElligatorSwift public key
+ *                      sent to the other party.
+ *           seckey32:  a pointer to the 32-byte secret key corresponding to
+ *                      ours64 (the correspondence is not checked)
+ *           hashfp:    pointer to a hash function. If NULL,
+ *                      secp256k1_ellswift_xdh_hash_function_default is used
+ *                      (in which case, 32 bytes will be written to output).
```
theStack, line 166: Since `hashfp` is not allowed to be NULL (as indicated by `SECP256K1_ARG_NONNULL(6)` below and `ARG_CHECK(hashfp != NULL);` in the implementation), the second sentence is obsolete:
```suggestion
 *           hashfp:    pointer to a hash function.
```

include/secp256k1_ellswift.h
```diff
@@ -0,0 +1,176 @@
...
+ *           data:      arbitrary data pointer passed through to hashfp (ignored
+ *                      by secp256k1_ellswift_xdh_hash_function_default).
```
theStack, line 168: Seems like `secp256k1_ellswift_xdh_hash_function_default` doesn't exist anymore:
```suggestion
 *           data:      arbitrary data pointer passed through to hashfp.
```

stratospher (commented):
I've reviewed the first 3 commits so far. And the writeup which was super helpful for reviewing the 3rd commit! (especially for understanding the encoding special cases) The logic described in the writeup and code do match. Also cross checked the constants till now (c1-c4, hash states for encode, create, xdh) by constructing them in python.

include/secp256k1_ellswift.h
```diff
@@ -0,0 +1,186 @@
...
+ *  In:      pubkey:     a pointer to a secp256k1_pubkey containing an
+ *                       initialized public key
+ *           rnd32:      pointer to 32 bytes of randomness
+ *
+ * It is recommended that rnd32 consists of 32 uniformly random bytes, not
+ * known to any adversary trying to detect whether public keys are being
+ * encoded, though 16 bytes of randomness (padded to an array of 32 bytes,
+ * e.g., with zeros) suffice to make the result indistinguishable from
+ * uniform. The randomness in rnd32 must be independent of pubkey. In
+ * particular, it must not be derived deterministically from pubkey.
```
jonasnick, line 97: nit: this confused me because `must` and `must not` sound like necessary requirements. It's fine to use some hardened derivation of the secret key as rnd32 (which is not independent of the pubkey) and it's fine to derive rnd32 deterministically from pubkey and some secret data.

src/modules/ellswift/main_impl.h
```diff
@@ -0,0 +1,589 @@
...
+    hash->s[3] = 0x1037a177ul;
+    hash->s[4] = 0x1b8fcb8bul;
+    hash->s[5] = 0x56023885ul;
+    hash->s[6] = 0x2560ede1ul;
+    hash->s[7] = 0xd626b715ul;
+
+    hash->bytes = 64;
+}
+
+int secp256k1_ellswift_encode(const secp256k1_context *ctx, unsigned char *ell64, const secp256k1_pubkey *pubkey, const unsigned char *ent32) {
```
jonasnick, line 402: The `ent32` arg is called `rnd32` in the include file

src/modules/ellswift/main_impl.h
```diff
@@ -0,0 +1,555 @@
...
+    secp256k1_declassify(ctx, &p, sizeof(p)); /* not constant time in produced pubkey */
+    secp256k1_fe_normalize_var(&p.x);
+    secp256k1_fe_normalize_var(&p.y);
+
+    /* Set up hasher state. The used RNG is H(privkey [|| auxrand32] || cnt++), using BIP340 tagged
+     * hash with tag "secp256k1_ellswift_create". */
+    secp256k1_ellswift_sha256_init_create(&hash);
+    secp256k1_sha256_write(&hash, seckey32, 32);
+    if (auxrand32) secp256k1_sha256_write(&hash, auxrand32, 32);
+    secp256k1_declassify(ctx, &hash, sizeof(hash)); /* hasher gets to declassify private key */
```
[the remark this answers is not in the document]
jonasnick: Looks good!

include/secp256k1_ellswift.h
jonasnick: :+1:

configure.ac
```diff
@@ -185,6 +185,10 @@ AC_ARG_ENABLE(module_schnorrsig,
     AS_HELP_STRING([--enable-module-schnorrsig],[enable schnorrsig module [default=yes]]), [],
     [SECP_SET_DEFAULT([enable_module_schnorrsig], [yes], [yes])])
 
+AC_ARG_ENABLE(module_ellswift,
+    AS_HELP_STRING([--enable-module-ellswift],[enable ElligatorSwift module (experimental)]), [],
```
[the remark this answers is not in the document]
jonasnick: I'm fine with the module being non-experimental. If it happens that we have to change the API, we can make a new (minor) release.

src/ctime_tests.c
```diff
@@ -171,4 +178,28 @@ static void run_tests(secp256k1_context *ctx, unsigned char *key) {
...
+
+    VALGRIND_MAKE_MEM_UNDEFINED(key, 32);
+    VALGRIND_MAKE_MEM_DEFINED(&ellswift, sizeof(ellswift));
+    ret = secp256k1_ellswift_xdh(ctx, msg, ellswift, ellswift, key, secp256k1_ellswift_xdh_hash_function_bip324, &ret);
+    VALGRIND_MAKE_MEM_DEFINED(&ret, sizeof(ret));
+    CHECK(ret == 1);
+
+    VALGRIND_MAKE_MEM_UNDEFINED(key, 32);
+    VALGRIND_MAKE_MEM_DEFINED(&ellswift, sizeof(ellswift));
+    ret = secp256k1_ellswift_xdh(ctx, msg, ellswift, ellswift, key, secp256k1_ellswift_xdh_hash_function_tagged, "bench_test");
```
jonasnick, line 201: CI fails because we need to `s/_tagged/_prefix`.

jonasnick: ... and changing the data to be 64 bytes

CMakeLists.txt
```diff
@@ -260,6 +266,7 @@ message("  ECDH ................................ ${SECP256K1_ENABLE_MODULE_ECDH}
 message("  ECDSA pubkey recovery ............... ${SECP256K1_ENABLE_MODULE_RECOVERY}")
 message("  extrakeys ........................... ${SECP256K1_ENABLE_MODULE_EXTRAKEYS}")
 message("  schnorrsig .......................... ${SECP256K1_ENABLE_MODULE_SCHNORRSIG}")
+message("  ElligatorSwift....................... ${SECP256K1_ENABLE_MODULE_ELLSWIFT}")
```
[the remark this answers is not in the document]
stratospher: doesn't look like it. (there's a single space after schnorrsig/extrakeys etc.. but not after ElligatorSwift.)

src/tests.c
```diff
@@ -3909,6 +3914,29 @@ static void test_ge(void) {
...
+        secp256k1_fe n;
+        secp256k1_ge q;
+        int ret_on_curve, ret_frac_on_curve, ret_set_xo;
+        secp256k1_fe_mul(&n, &zf, &r);
+        ret_on_curve = secp256k1_ge_x_on_curve_var(&r);
+        ret_frac_on_curve = secp256k1_ge_x_frac_on_curve_var(&n, &zf);
+        ret_set_xo = secp256k1_ge_set_xo_var(&q, &r, 0);
+        CHECK(ret_on_curve == ret_frac_on_curve);
+        CHECK(ret_on_curve == ret_set_xo);
+        if (ret_set_xo) CHECK(secp256k1_fe_equal_var(&r, &q.x));
```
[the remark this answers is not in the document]
stratospher: oh! true, it's possible the function behaves unpredictably in a different environment. agree with keeping it as it is.

jonasnick (commented):
code review ACK 3cfa1e033dc7ca55521659bb72b7753053308ac1

I have not reviewed ellswift.md or otherwise verified that the pseudocode is correct.

src/modules/ellswift/main_impl.h
```diff
@@ -462,6 +462,7 @@ int secp256k1_ellswift_create(const secp256k1_context *ctx, unsigned char *ell64
 
     /* Compute (affine) public key */
     ret = secp256k1_ec_pubkey_create_helper(&ctx->ecmult_gen_ctx, &seckey_scalar, &p, seckey32);
+    secp256k1_declassify(ctx, &p, sizeof(p)); /* not constant time in produced pubkey */
```
stratospher, line 465: 70d80ae: why don't variables in the `ellswift_xdh` code which interact with private key not use `secp256k1_declassify`? i don't understand the context for what kind of memory needs to be marked as not secret here.

src/modules/ellswift/tests_impl.h
```diff
@@ -0,0 +1,315 @@
...
+    {0x77, SECP256K1_FE_CONST(0xc17ec69e, 0x665f0fb0, 0xdbab48d9, 0xc2f94d12, 0xec8a9d7e, 0xacb58084, 0x83309180, 0x1eb0b80b), SECP256K1_FE_CONST(0x147756e6, 0x6d96e31c, 0x426d3cc8, 0x5ed0c4cf, 0xbef6341d, 0xd8b28558, 0x5aa574ea, 0x0204b55e), {SECP256K1_FE_CONST(0x6f4aea43, 0x1a0043bd, 0xd03134d6, 0xd9159119, 0xce034b88, 0xc32e50e8, 0xe36c4ee4, 0x5eac7ae9), SECP256K1_FE_CONST(0xfd5be16d, 0x4ffa2690, 0x126c67c3, 0xef7cb9d2, 0x9b74d397, 0xc78b06b3, 0x605fda34, 0xdc9696a6), SECP256K1_FE_CONST(0x5e9c6079, 0x2a2f000e, 0x45c6250f, 0x296f875e, 0x174efc0e, 0x9703e628, 0x706103a9, 0xdd2d82c7), SECP256K1_FE_CONST(0, 0, 0, 0, 0, 0, 0, 0), SECP256K1_FE_CONST(0x90b515bc, 0xe5ffbc42, 0x2fcecb29, 0x26ea6ee6, 0x31fcb477, 0x3cd1af17, 0x1c93b11a, 0xa1538146), SECP256K1_FE_CONST(0x02a41e92, 0xb005d96f, 0xed93983c, 0x1083462d, 0x648b2c68, 0x3874f94c, 0x9fa025ca, 0x23696589), SECP256K1_FE_CONST(0xa1639f86, 0xd5d0fff1, 0xba39daf0, 0xd69078a1, 0xe8b103f1, 0x68fc19d7, 0x8f9efc55, 0x22d27968), SECP256K1_FE_CONST(0, 0, 0, 0, 0, 0, 0, 0)}},
```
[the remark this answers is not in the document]
stratospher: oh, the exactly 0, 4, or 8 t values were before the special cases were factored in! thanks, understood it now.

src/modules/ellswift/tests_impl.h
```diff
@@ -0,0 +1,315 @@
...
+        /* Compute the shared secret both ways and compare with each other. */
+        ret = secp256k1_ellswift_xdh(CTX, share32a, ell64a, ell64b, sec32b, 1, &ellswift_xdh_hash_x32, NULL);
+        CHECK(ret);
+        ret = secp256k1_ellswift_xdh(CTX, share32b, ell64a, ell64b, sec32a, 0, &ellswift_xdh_hash_x32, NULL);
+        CHECK(ret);
+        CHECK(secp256k1_memcmp_var(share32a, share32b, 32) == 0);
+        /* Verify that the shared secret doesn't match if a secret key or remote pubkey changes. */
+        secp256k1_testrand_flip(ell64a, 64);
+        ret = secp256k1_ellswift_xdh(CTX, share32a, ell64a, ell64b, sec32b, 1, &ellswift_xdh_hash_x32, NULL);
+        CHECK(ret);
```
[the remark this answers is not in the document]
stratospher: oh, makes sense! missed the part about both sides being dishonest before.

src/modules/ellswift/main_impl.h
```diff
@@ -495,6 +495,53 @@ int secp256k1_ellswift_decode(const secp256k1_context *ctx, secp256k1_pubkey *pu
     return 1;
 }
 
+static int ellswift_xdh_hash_function_prefix(unsigned char *output, const unsigned char *x32, const unsigned char *ell_a64, const unsigned char *ell_b64, void *data) {
+    secp256k1_sha256 sha;
+
+    (void)data;
```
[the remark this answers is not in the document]
stratospher: interesting! but data is being used 2 lines below?

src/modules/ellswift/main_impl.h
```diff
@@ -0,0 +1,541 @@
...
+}
+
+int secp256k1_ellswift_encode(const secp256k1_context *ctx, unsigned char *ell64, const secp256k1_pubkey *pubkey, const unsigned char *rnd32) {
+    secp256k1_ge p;
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(ell64 != NULL);
+    ARG_CHECK(pubkey != NULL);
+    ARG_CHECK(rnd32 != NULL);
+
+    if (secp256k1_pubkey_load(ctx, &p, pubkey)) {
```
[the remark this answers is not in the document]
stratospher: nice, fascinated by the level of details!

src/modules/ellswift/main_impl.h
stratospher: thanks for the detailed explanation! understood when declassify needs to be used now.

src/modules/ellswift/main_impl.h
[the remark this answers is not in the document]
stratospher: btw, there's another unresolved question in the reply thread here.

CMakeLists.txt
```diff
@@ -71,6 +71,12 @@ if(SECP256K1_ENABLE_MODULE_EXTRAKEYS)
   add_compile_definitions(ENABLE_MODULE_EXTRAKEYS=1)
 endif()
 
+option(SECP256K1_ENABLE_MODULE_ELLSWIFT "Enable ElligatorSwift module." ON)
+if(SECP256K1_ENABLE_MODULE_ELLSWIFT)
+  add_definitions(-DENABLE_MODULE_ELLSWIFT=1)
```
theuni, line 76: I'm assuming this is older than our `add_compile_definitions` switch-over. Mind updating it to match?

theuni (commented):
Quick glimpse at the buildsystem...  changes look good to me.

hebasto (commented):
ACK fe26466f040a6ddc242cea54037c2c4cd85a0bc9 on changes in Autotools, CMake and CI related stuff.

include/secp256k1_ellswift.h
```diff
@@ -0,0 +1,176 @@
...
+  unsigned char *output,
+  const unsigned char *x32,
+  const unsigned char *ell_a64,
+  const unsigned char *ell_b64,
+  void *data
+);
+
+/** Construct a 64-byte ElligatorSwift encoding of a given pubkey.
+ *
+ *  Returns: 1 when pubkey is valid.
```
real-or-random, line 85: always

src/modules/ellswift/main_impl.h
```diff
@@ -0,0 +1,588 @@
...
+        /* Let s = x-u. */
+        secp256k1_fe_negate(&m, &u, 1); /* m = -u */
+        s = m; /* s = -u */
+        secp256k1_fe_add(&s, &x); /* s = x-u */
+
+        /* If s=0, fail. */
+        if (secp256k1_fe_normalizes_to_zero_var(&s)) return 0;
+
+        /* If s is not square, fail. */
+        if (!secp256k1_fe_is_square_var(&s)) return 0;
```
real-or-random, line 263: Could we drop the `s == 0` check here? I assume it happens with negligible probability on honest inputs, and it's subsumed by the `is_square(q)` check below.

If no:
 * Do we want to add an `EXPECT` to the `s == 0` check? (Not sure if it changes much, but I saw you did the same above in the other function.)
 * Do we want to swap these checks for an earlier return?

src/modules/ellswift/main_impl.h
```diff
@@ -0,0 +1,588 @@
...
+        secp256k1_fe_mul(&q, &s, &u2); /* q = 3*s*u^2 */
+        secp256k1_fe_add(&q, &g); /* q = 4*(u^3+7)+3*s*u^2 */
+        secp256k1_fe_mul(&q, &q, &s); /* q = s*(4*(u^3+7)+3*u^2*s) */
+        secp256k1_fe_negate(&q, &q, 1); /* q = -s*(4*(u^3+7)+3*u^2*s) */
+        if (!secp256k1_fe_is_square_var(&q)) return 0;
+        ret = secp256k1_fe_sqrt(&r, &q); /* r = sqrt(-s*(4*(u^3+7)+3*u^2*s)) */
+        VERIFY_CHECK(ret);
+
+        /* If (c & 1) = 1 and r = 0, fail. */
+        if ((c & 1) && secp256k1_fe_normalizes_to_zero_var(&r)) return 0;
```
real-or-random, line 281: Add `EXCEPT` here?

src/modules/ellswift/main_impl.h
```diff
@@ -0,0 +1,588 @@
...
+        secp256k1_fe_set_b32_mod(&u, u32);
+        /* Since u is the output of a hash, it should practically never be 0. We could apply the
+         * u=0 to u=1 correction here too to deal with that case still, but it's such a low
+         * probability that we do not bother. */
+#ifdef VERIFY
+        VERIFY_CHECK(!secp256k1_fe_normalizes_to_zero_var(&u));
+#endif
+        /* Find a remainder t, and return it if found. */
+        if (secp256k1_ellswift_xswiftec_inv_var(t, x, &u, branch)) {
+            secp256k1_fe_normalize_var(t);
```
real-or-random, line 369: This line can be dropped (or better, replaced by an assertion), as the last operation in `secp256k1_ellswift_xswiftec_inv_var` is a mul.

real-or-random (commented):
Here are some minor remarks about performance.

Continuing to go through the main parts of the code. I'll probably also follow up with a fixup commit with some nits in the code comments later this week.

src/modules/ellswift/main_impl.h
real-or-random: > > and it's subsumed by the `is_square(q)` check below.
> 
> That's not correct. The tests fail if you comment it out.

Oh.  When I wrote this, I had assumed, for a moment, that we abort if q is a square. But we abort if it's not a square.

src/modules/ellswift/main_impl.h
real-or-random: Yep, I missed that we need full normalization.

src/modules/ellswift/main_impl.h
```diff
@@ -0,0 +1,588 @@
...
+        secp256k1_fe_normalize_weak(&g);
+        secp256k1_fe_mul_int(&g, 4); /* g = 4*(u^3+7) */
+        secp256k1_fe_mul_int(&u2, 3); /* u2 = 3*u^2 */
+        secp256k1_fe_mul(&q, &s, &u2); /* q = 3*s*u^2 */
+        secp256k1_fe_add(&q, &g); /* q = 4*(u^3+7)+3*s*u^2 */
+        secp256k1_fe_mul(&q, &q, &s); /* q = s*(4*(u^3+7)+3*u^2*s) */
+        secp256k1_fe_negate(&q, &q, 1); /* q = -s*(4*(u^3+7)+3*u^2*s) */
+        if (!secp256k1_fe_is_square_var(&q)) return 0;
+        ret = secp256k1_fe_sqrt(&r, &q); /* r = sqrt(-s*(4*(u^3+7)+3*u^2*s)) */
+        VERIFY_CHECK(ret);
```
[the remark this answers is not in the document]
real-or-random: Makes sense. See https://github.com/bitcoin-core/secp256k1/pull/1347 for the missing docs.

src/modules/ellswift/main_impl.h
```diff
@@ -0,0 +1,586 @@
...
+    if (secp256k1_pubkey_load(ctx, &p, pubkey)) {
+        secp256k1_fe t;
+        unsigned char p64[64] = {0};
+        secp256k1_sha256 hash;
+
+        /* Set up hasher state; the used RNG is H(pubkey || "\x00"*31 || rnd32 || cnt++), using BIP340 tagged
+         * hash with tag "secp256k1_ellswift_encode". */
+        secp256k1_ellswift_sha256_init_encode(&hash);
+        p64[0] = 0x02 ^ secp256k1_fe_is_odd(&p.y);
+        secp256k1_fe_get_b32(p64 + 1, &p.x);
```
real-or-random, line 418: Do you think it's better to call `secp256k1_eckey_pubkey_serialize` here? Or even the API facing `secp256k1_ec_pubkey_serialize`, which also takes care of loading the pubkey?

src/modules/ellswift/main_impl.h
```diff
@@ -0,0 +1,586 @@
...
+        secp256k1_sha256_write(&hash, p64, sizeof(p64));
+        secp256k1_sha256_write(&hash, rnd32, 32);
+
+        /* Compute ElligatorSwift encoding and construct output. */
+        secp256k1_ellswift_elligatorswift_var(ell64, &t, &p, &hash); /* puts u in ell64[0..32] */
+        secp256k1_fe_get_b32(ell64 + 32, &t); /* puts t in ell64[32..64] */
+        return 1;
+    }
+    /* Only returned in case the provided pubkey is invalid. */
```
real-or-random, line 427:
```suggestion
    /* Only reached in case the provided pubkey is invalid. */
    memset(ell64, 0, 64);
```

include/secp256k1_ellswift.h
```diff
@@ -0,0 +1,189 @@
...
+ *  BIP324. It returns H_tag(ell_a64 || ell_b64 || x32), where H_tag is the
+ *  BIP340 tagged hash function with tag "bip324_ellswift_xonly_ecdh". Equivalent
+ *  to secp256k1_ellswift_xdh_hash_function_prefix with prefix64 set to
+ *  SHA256("bip324_ellswift_xonly_ecdh")||SHA256("bip324_ellswift_xonly_ecdh").
+ *  The data argument is ignored. */
+SECP256K1_API_VAR const secp256k1_ellswift_xdh_hash_function secp256k1_ellswift_xdh_hash_function_bip324;
+
+/** Construct a 64-byte ElligatorSwift encoding of a given pubkey.
+ *
+ *  Returns: 1 when pubkey is valid.
```
real-or-random, line 85:
```suggestion
 *  Returns: 1 always.
```
(There are no invalid pubkeys in terms of the API, similar to  https://github.com/bitcoin-core/secp256k1/pull/1341 )

include/secp256k1_ellswift.h
```diff
@@ -0,0 +1,189 @@
...
+ * It is recommended that rnd32 consists of 32 uniformly random bytes, not
+ * known to any adversary trying to detect whether public keys are being
+ * encoded, though 16 bytes of randomness (padded to an array of 32 bytes,
+ * e.g., with zeros) suffice to make the result indistinguishable from
+ * uniform. The randomness in rnd32 must not be a deterministic function of
+ * the pubkey (it can be derived from the private key, though).
+ *
+ * This function runs in variable time.
+ */
+SECP256K1_API int secp256k1_ellswift_encode(
```
real-or-random, line 101: Do we want to add a warning to make sure we can swap the RNG?  Suggestion:

"It is not guaranteed that the computed ell64 is stable across versions of the library, even if all arguments to this function including rnd32 are the same."

If yes, this should be added to secp256k1_ellswift_create too

include/secp256k1_ellswift.h
```diff
@@ -0,0 +1,189 @@
...
+SECP256K1_API int secp256k1_ellswift_decode(
+    const secp256k1_context *ctx,
+    secp256k1_pubkey *pubkey,
+    const unsigned char *ell64
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
+
+/** Compute an ElligatorSwift public key for a secret key.
+ *
+ *  Returns: 1: secret was valid, public key was stored.
+ *           0: secret was invalid, try again.
```
real-or-random, line 126: (Not this PR, I see this is copied from the other docs:) Ha, we should change this, "try again" is terrible advice. Hey, if your RNG outputs 32 zeroes, then don't worry, just call it in a loop. :D

src/modules/ellswift/main_impl.h
real-or-random: > I've made these changes (and moved the `s==0` check even further, so that other more-likely-to-fail check go before it).

Perhaps update the pseudocode in the comment accordingly.

bitcoinfinancier (approved):

src/modules/ellswift/tests_impl.h
```diff
@@ -0,0 +1,402 @@
...
+        /* Construct ElligatorSwift-encoded public keys for those keys. */
+        ret = secp256k1_ellswift_create(CTX, ell64a, sec32a, auxrnd32a);
+        CHECK(ret);
+        ret = secp256k1_ellswift_create(CTX, ell64b, sec32b, auxrnd32b);
+        CHECK(ret);
+
+        /* Compute the shared secret both ways and compare with each other. */
+        ret = secp256k1_ellswift_xdh(CTX, share32a, ell64a, ell64b, sec32b, 1, hash_function, data);
+        CHECK(ret);
+        ret = secp256k1_ellswift_xdh(CTX, share32b, ell64a, ell64b, sec32a, 0, hash_function, data);
```
jonasnick, line 338: Shouldn't the first share be called `share32b` and the second `share32a`? Without this a few of the tests below don't make sense.

If we change the names of the shares as suggested, we would also need to replace `share32b` with `share32a` in
```
        ret = secp256k1_ellswift_xdh(CTX, share32_bad, ell64a, ell64b, sec32a_bad, 0, hash_function, data);
        CHECK(!ret || secp256k1_memcmp_var(share32_bad, share32b, 32) != 0);
 ```

File: src/modules/ellswift/main_impl.h [hunk 33]
```diff
@@ -0,0 +1,542 @@
...
+
+    /* Let w=sqrt(s). */
+    ret = secp256k1_fe_sqrt(&m, &s); /* m = sqrt(s) = w */
+    VERIFY_CHECK(ret);
+
+    /* Return logic. */
+    if ((c & 5) == 0 || (c & 5) == 5) {
+        secp256k1_fe_negate(&m, &m, 1); /* m = -w */
+    }
+    /* Now m = {w if c&5=0 or c&5=5; -w otherwise}. */
```
theStack, line 299: comment nit:
```suggestion
    /* Now m = {-w if c&5=0 or c&5=5; w otherwise}. */
```

theStack (commented):
Noticed that the commit message of 16d251380fb3e619ff7da0a983735ff78d301307 contains some typos/inaccuracies:
```diff
diff --git a/commitmsg.txt b/commitmsg.txt
index b1bd953..fadc3fe 100644
--- a/commitmsg.txt
+++ b/commitmsg.txt
@@ -12,11 +12,11 @@
     interpret them as field elements u and t. Then:
     * Let c=0xa2d2ba93507f1df233770c2a797962cc61f6d15da14ecd47d8d27ae1cd5f852.
     * Let u'=u unless u==0, in which case let u'=1.
-    * Let t'=t unless t==1, in which case let t'=1.
-    * Let t''=t unless u'^3 + + 7 + t'^2 == 0, in which case let t''=2*t'.
+    * Let t'=t unless t==0, in which case let t'=1.
+    * Let t''=t' unless u'^3 + 7 + t'^2 == 0, in which case let t''=2*t'.
     * Let X=(u'^3 + 7 - t''^2)/(2*t'')
     * Let Y=(u'^3 + 7 + t''^2)/(2*c*t''*u')
-    * Let x be the first element out of [u+4*Y^2, (-X/Y-u)/2, (X/Y-u)/2] for which
+    * Let x be the first element out of [u'+4*Y^2, (-X/Y-u')/2, (X/Y-u')/2] for which
       x^3 + 7 is square (there will always be at least one).
     * Let y be the square root of x^3 + 7 which has the same parity as t.
     * Return the point with affine coordinates (x, y)
```
Could probably simply take the version of the decoding algorithm included in `include/secp256k1_ellswift.h` (or even completely remove it from the commit body, as it's added to the header in the same commit anyways?).

theStack (commented):
Verified that the code in the decoding/encoding functions `secp256k1_ellswift_xswiftec_frac_var` and `secp256k1_ellswift_xswiftec_inv_var` match the algorithm written in comments each (https://github.com/sipa/secp256k1/blob/55efb1cb525c1b264c57c35203d1b83109948d5e/src/modules/ellswift/main_impl.h#L74-L84 and https://github.com/sipa/secp256k1/blob/55efb1cb525c1b264c57c35203d1b83109948d5e/src/modules/ellswift/main_impl.h#L170-L189). Still planning to do a pen-and-paper review for the algorithm simplifications steps / substitutions pointed out in `secp256k1_ellswift_xswiftec_frac_var`.

Left two nits below (unrelated to the functions above), regarding API doc and the ellswift benchmark.

src/modules/ellswift/main_impl.h
real-or-random: I think you overlooked this one. My primary suggestion was to add a call to `memset`, the comment change was rather a by-product thereof.  :)

real-or-random (approved):
ACK 55efb1cb525c1b264c57c35203d1b83109948d5e

Here's a branch with some fixup comments, with commits prepared for `git rebase --interactive --autosquash`:  https://github.com/real-or-random/secp256k1/tree/202206_ellswift_dh
You are welcome to apply or reject; ACK anyway. My initial plan was to have only comment nits, but I found two further refinements on the way, see the commits there. 

*edit: It's probably clear to you, but I recommend first taking whatever fixup commits you want from my branch, and then addressing other review commits from @theStack here in order to minimize conflicts.

Davidson-Souza:
tACK 90e360a. Full testing backlog:

- Iterated through the implementation, checking all the math (the verbose comments helped a lot!)
- Indepently generated all constants using python/sage (Python 3.10.10 and SageMath version 9.8, Release Date: 2023-02-11)
- A "manual mutation test" for the valgrind ctime tests, adding some dummy, secret dependent branching, it indeed causes memcheck to complain about uninitialized values. Using `VALGRIND_MAKE_MEM_DEFINED` silences those warnings, as expected.
- Executed all tests, including the memcheck ones (OS and hardware below). Compilation using both gcc and clang, configure script bellow.
- The documentation is sound and very helpful

## Test environments:
Config script: `[CC=clang] ./configure --with-valgrind=yes --enable-module-ecdh --enable-module-ellswift --enable-experimental --no-recursion`

Machine 1: Arch Linux on a AMD Ryze 5 (Linux 6.1.21-hardened1-1-hardened #1 SMP PREEMPT_DYNAMIC x86_64 GNU/Linux )
gcc version 12.2.1 20230201  
clang version 14.0.7
GNU libc stable release version 2.37.
valgrind-3.20.0

Machine 2: Raspbian on a Raspbery Pi 4b (Linux raspberrypi 6.1.21-v8+ #1642 SMP PREEMPT aarch64 GNU/Linux)
gcc version 10.2.1 20210110 (Debian 10.2.1-6)
Debian clang version 11.0.1-2
valgrind-3.16.1
(Debian GLIBC 2.31-13+rpt2+rpi1+deb11u5) stable release version 2.31

## Benchmarks (only for the new module)
Machine 1, clang
```
ellswift_encode               ,    23.8       ,    24.1       ,    24.9    
ellswift_decode               ,    10.9       ,    11.4       ,    11.6    
ellswift_keygen               ,    55.3       ,    58.8       ,    61.7    
ellswift_ecdh                 ,    63.3       ,    64.1       ,    65.3 
```
Machine 1, gcc
```
ellswift_encode               ,    22.6       ,    22.7       ,    22.9    
ellswift_decode               ,    10.7       ,    10.8       ,    11.0    
ellswift_keygen               ,    54.3       ,    54.7       ,    55.9    
ellswift_ecdh                 ,    62.7       ,    63.2       ,    64.1
```
Machine 2, clang
```
ellswift_encode               ,    63.7       ,    63.7       ,    63.8    
ellswift_decode               ,    33.7       ,    33.8       ,    34.0    
ellswift_keygen               ,   177.0       ,   177.0       ,   178.0    
ellswift_ecdh                 ,   233.0       ,   234.0       ,   234.0
```
Machine 2, gcc
```
ellswift_encode               ,    66.4       ,    66.4       ,    66.5    
ellswift_decode               ,    34.9       ,    34.9       ,    34.9    
ellswift_keygen               ,   182.0       ,   182.0       ,   183.0    
ellswift_ecdh                 ,   230.0       ,   231.0       ,   231.0  
```

real-or-random (approved):
ACK 90e360acc2511f313964e394005bafb377b4f191

jonasnick (approved):
ACK 90e360acc2511f313964e394005bafb377b4f191

PR #760, opened 2020-06-18
WIP: Add schnorrsig batch verification

jonasnick:
This was part of #558 (for 20 months) to demonstrate the advantages of batch verification (see [graph](https://github.com/sipa/bips/blob/bip-schnorr/bip-schnorr.mediawiki#design)), but then removed to simplify #558 because there are still ongoing discussions:

- [ ] there's @real-or-random's proposal to add synthetic randomness for batch verification (https://github.com/sipa/bips/issues/204)
- [ ] batch verification fairly well tested, but still wouldn't be comfortable with using this in Bitcoin Core for consensus in its current state because it relies on parts of the lib that are otherwise unused such as scratch spaces and ecmult_multi. Ideally we would have comprehensive fuzz tests for batch verification.
- [ ] adding chacha20 may not be worth it, because it may only provide a negligible speedup over SHA256 (TODO: test this), plus we're planning to allow overriding the SHA256 implementation at compile time (https://github.com/bitcoin-core/secp256k1/pull/558#issuecomment-619579991).

PR #1134, opened 2022-08-21
Add batch verification module

### LINKED ISSUE #1087: General batch verification API context
context: https://github.com/bitcoin-core/secp256k1/pull/760#issuecomment-809242311

I am trying to implement a PoC for the API proposed above. I have the following `batch_verify` object in mind.
```C
typedef struct {
    unsigned char chacha_seed[32];
    secp256k1_scalar randomizer_cache[2];
    schnorrsig_batch_verify sigs_data;
    tweaked_key_batch_verify tweaked_keys_data;
} batch_verify_struct;

typdef struct {
    secp256k1_scratch *sigs_data; /* (sig, msg, pk) */
    size_t len; 
    size_t capacity; /* equals (sigs_data->max_size)/(64 + 32 + sizeof(secp256k1_xonly)) */
    int result;
} schnorrsig_batch_verify;

typdef struct {
    secp256k1_scratch *tweaks_data; /* (pairity, tweaked_key, tweak32, pubkey_key) */
    size_t len;
    size_t capacity;
    int result;
} tweaked_key_batch_verify;
```
----
I plan to use a scratch object to store the data (schnorrsig or tweaks) since it will allow us to keep on adding new data (using `batch_add_sig` and `batch_add_xpubkey_tweak`) and increase the batch object's size accordingly. This batch object doesn't seem compatible with [`ecmult_pippenger_batch`](https://github.com/bitcoin-core/secp256k1/blob/master/src/ecmult_impl.h#L643) or [`ecmult_strauss_batch`](https://github.com/bitcoin-core/secp256k1/blob/master/src/ecmult_impl.h#L355) function call.

Since both Pippenger and Strauss takes the arguments:
- `void *cbdata`  --> contains the required data
- `secp256k1_scratch *scratch` --> newly allocated scratch space where scalars and points are loaded for multi multiplication

But this batch object already has the required data in a scratch space. Maybe use another scratch space for loading scalars and points? Won't this increase memory usage?

siv2r:
### Overview

This PR adds support for batch verifying Schnorr signatures and tweaked x-only public key checks. It is based on the work of @jonasnick in #760. 

### Batch Verification

This implementation does not strictly follow the [BIP340 batch verification spec](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#batch-verification). The API design is loosely based on this suggestion: https://github.com/bitcoin-core/secp256k1/pull/760#issuecomment-809242311. Prior development discussion of this PR can be found in siv2r/secp256k1#2.

### Speed Up
 - batch verifying Schnorr signatures is **_20% faster_** - [graph here](https://github.com/bitcoin-core/secp256k1/pull/1134/files#diff-583efe34fcab54bf7e5296bf65e522adc2345b3580e2a50ad5cccbd91dd82032)
 - batch verifying tweak pubkey checks is **_50% faster_** - [graph here](https://github.com/bitcoin-core/secp256k1/pull/1134/files#diff-1e73547c6cfbb18febec26224f0389d8ed1893ec7883f485a97caa583609b16b)

Fixes #1087


siv2r:
> This implementation does not strictly follow the [BIP340 batch verification spec](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki#batch-verification).

For example,
- the random numbers (or randomizers) aren’t generated by passing a seed (hash of all inputs) to CSPRNG
- allows mixing of schnorrsig and tweak checks
- uses the tag *“BIP0340/batch”* for sha256 (not in the BIP340 specs)

### Alternative Design Options
- Delayed randomizer generation
    - the current code generates a randomizer just after a user enters input (in `_batch_add_*`)
    - we could instead generate them (in `batch_verify`) after the user enters all their input
    - Pros: follows bip340 specs. Cons: consumes more memory.
    - more info: https://github.com/bitcoin-core/secp256k1/issues/1087#issuecomment-1067276801 and [here](https://siv2r.notion.site/siv2r/Batch-Verification-Interface-Proposal-4b41080bd3de4c59a37ad9279d562155#898b8bcc5dc64aa994e5accaa5c3ab2c)
- Use BLAKE256 instead of SHA256 for generating randomizers (to improve speed).
- In `_batch_create`, we could:
    - Get memory size instead of `max_terms`.
    - Provide pre-determined sizes (small, medium, and large) instead of `max_terms`.
        - Better test coverage?


### Questions
- Streaming batch API (this PR) vs Single call batch API?
- Is transparent verification required?
    - the current code implements transparent verification (inside `batch_add_*` APIs)
    - without transparent verification, the user needs to check for space in the batch before calling `_batch_add` APIs:
    ```c
    if (batch_enough_space_for_schnorrsig) {
      batch_add_schnorrsig()
    }
    ```
- On an empty batch, Should `_batch_verify` return 0 or 1?
    - the current code returns 1
        - simple implementation
    - if we want to return 0
        - needs an extra param in batch to avoid ANDing the result with its initial value
        - provides better security?
- `xonly_pubkey_tweak_add_check` recommends `ctx` to be initialized for verification, but it can work even if `ctx` is initialized as none.
    - Since `batch_add_tweak_xonlypub_check` is based on `tweak_add_check`, should it also recommend that `ctx` be initialized for verification?
    - Can it recommend that `ctx` be initialized for none instead?
- A better name for `secp256k1_batch_usable`?
    - the current name is confusing
    - here, *“usable”* means if the batch can be used by the `batch_add_*` functions

siv2r:
### Questions
- Streaming batch API (this PR) vs Single call batch API?
- Is transparent verification required?
    - the current code implements transparent verification (inside `batch_add_*` APIs)
    - without transparent verification, the user needs to check for space in the batch before calling `_batch_add` APIs:
    ```c
    if (batch_enough_space_for_schnorrsig) {
      batch_add_schnorrsig()
    }
    ```
- On an empty batch, Should `_batch_verify` return 0 or 1?
    - the current code returns 1
        - simple implementation
    - if we want to return 0
        - needs an extra param in batch to avoid ANDing the result with its initial value
        - provides better security?
- `xonly_pubkey_tweak_add_check` recommends `ctx` to be initialized for verification, but it can work even if `ctx` is initialized as none.
    - Since `batch_add_tweak_xonlypub_check` is based on `tweak_add_check`, should it also recommend that `ctx` be initialized for verification?
    - Can it recommend that `ctx` be initialized for none instead?
- A better name for `secp256k1_batch_usable`?
    - the current name is confusing
    - here, *“usable”* means if the batch can be used by the `batch_add_*` functions

siv2r:
In `extrakeys/bench_impl.h`, cast pointers to `(void *)` before freeing to avoid MSVC warning.

doc/speedup-batch/bench_output.txt
```diff
@@ -0,0 +1,137 @@
+Benchmark                          ,    Min(us)    ,    Avg(us)    ,    Max(us)    
+
+schnorrsig_sign                    ,    50.4       ,    50.5       ,    50.7    
+schnorrsig_verify                  ,    89.1       ,    89.2       ,    89.3    
+schnorrsig_batch_verify_1          ,   104.0       ,   104.0       ,   104.0
```
sipa, line 5: batch_verify_1 being slower than verify shouldn't be. Is it possible to revert to using non-batch validation logic for this case?

siv2r: The non-batch validation (`secp256k1_schnorr_verify`) logic looks something like this:
 - calc `rj` using `secp256k1_ecmult`: Rj = s*G - e*P
 - convert `rj` (gej) to `r` (ge)
 - check if the `r.x = sig[0:32]` and `r.y = even`

one schnorrsig occupies two points in the batch, and one tweak check occupies one point in the batch. If a batch contains two points, there is no guarantee that they are from a schnorrsig (R, P). It could be from two tweak checks. So, we can't use the `r.y = even` check.

Hence, I tried implementing a slightly modified `schnorr_verify` logic:
 - calc `neg_rj` using `secp256k1_ecmult`: neg_Rj = -s*G + batch.scalar[1]*batch.point[1]
 - check if `neg_rj + batch.point[0] == inf`

This gives somewhat better benchmarks than before:
```bash
Benchmark                          ,    Min(us)    ,    Avg(us)    ,    Max(us)    

schnorrsig_sign                    ,    49.1       ,    50.1       ,    53.4    
schnorrsig_verify                  ,    86.6       ,    87.2       ,    88.4    
schnorrsig_batch_verify_1          ,    94.7       ,    95.0       ,    95.2 
```
But `schnorrsig_batch_verify_1` is still slower than `schnorrsig_vrerify`.

jonasnick (commented):
> This implementation does not strictly follow the BIP340 batch verification spec. [...] the random numbers (or randomizers) aren’t generated by passing a seed (hash of all inputs) to CSPRNG

The [security argument](https://github.com/sipa/bips/issues/219#issuecomment-815575754) for batch verification should relatively easily translate to the approach in this PR.

> Is transparent verification required?

I initially wasn't a fan of transparent verification because developers generally want to know how long a certain function call takes and don't want to be surprised by some `batch_add` calls taking much longer than others. But I changed my mind on this.

The batch verification API right now is used as follows:
```
for (i = 0; i < N_SIGS; i++) {
    if(!batch_usable(batch) || !batch_add_schnorrsig(batch, sig[i], msg[i], sizeof(msg[i]), &pk)) {
        return 0;
    }
}
if(!batch_verify(ctx, batch)) {
    return 0;
}
```

Without TV, then one option is to have the user create a batch that is large enough. Of course that's not alway possible because the batch can get larger than the available memory. Moreover, this would require adding an API that, given a number of schnorrsigs and tweaks (and more in the future) returns the size of the required batch. That seems way more complicated than TV.

If there's no TV and the batch is not guaranteed to be large enough, then users need to essentially reimplement something like transparent verification:
```
for (i = 0; i < N_SIGS; i++) {
    if(!schnorrsig_batch_has_space(batch)) {
        if(!batch_verify(batch)) {
            return 0;
        }
    }
    if (!batch_add_schnorrsig(batch, sig[i], msg[i], sizeof(msg[i]), &pk)) {
        return 0;
    }
}
if(!batch_verify(ctx, batch)) {
    return 0;
}
```

1. This is more code compared to having TV built into `batch_add`.
2. `*_batch_has_space` is specific to whatever you're trying to add to the batch, i.e., we would also have to add `tweak_has_space`, for example.
3. The `batch_usable` function in the current implementation (with TV) only allows for earlier aborts and is not essential (unless you want to determine if a `batch_add` failed because its input is obviously malformed or because the previous `batch_add` triggered a `batch_verify` that failed).
4. If users don't want to use TV (for whatever reason), then they don't have to - even in the current implementation. This requires counting the terms that have been added to the batch and verifying before it's full (perhaps we can make this simpler).

> xonly_pubkey_tweak_add_check recommends ctx to be initialized for verification, but it can work even if ctx is initialized as none. [...] Since batch_add_tweak_xonlypub_check is based on tweak_add_check, should it also recommend that ctx be initialized for verification?

I don't think so. #1126 removes "initialized for verification" from the API docs of `xonly_pubkey_tweak_add_check`. For consistency it would be better if you'd remove "(can be initialized for none)" from the API doc.

real-or-random:
I haven't had a closer look yet but I agree with @jonasnick's comments about TV.

fjahr:
@siv2r Are you still working on this topic? Do you plan to address @jonasnick 's comments?

siv2r:
@fjahr, this PR needs review from other contributors regarding the (batch API) design decision made here. The https://github.com/bitcoin-core/secp256k1/pull/1134#pullrequestreview-1083948472 comment showed its support for the "Transparent Verification" feature, which was implemented.

IIRC, two documentation changes are required (at the time of writing):
- remove "initialized for verification" from the `batch_add_tweak_xonlypub_check` doc.
- document that `batch_verify_1` is slower than `schnorrsig_verify` (see [#1134 (comment)](https://github.com/bitcoin-core/secp256k1/pull/1134#discussion_r952908842)).

I avoided making these changes immediately to keep the commit history clean for reviewers. I would be happy to work on any required changes after it gets enough review.

fjahr (commented):
I have opened a draft PR for using this in bitcoin core: https://github.com/bitcoin/bitcoin/pull/29491

Here is a rebased branch of the code that I am using: https://github.com/fjahr/secp256k1/tree/pr1134-rebase-2024

I hope this can create some new interest and motivate people to review here as well.

include/secp256k1_batch.h
```diff
@@ -0,0 +1,110 @@
...
+ *     (https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki).
+ */
+
+/** Opaque data structure that holds information required for the batch verification.
+ *
+ *  The purpose of this structure is to store elliptic curve points, their scalar
+ *  coefficients, and scalar coefficient of generator point participating in Multi-Scalar
+ *  Point Multiplication computation, which is done by `secp256k1_ecmult_strauss_batch_internal`
+ */
+typedef struct secp256k1_batch_struct secp256k1_batch;
```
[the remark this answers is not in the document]
siv2r: Why do we need this? Currently, users can only create pointers to the `secp256k1_batch` object, which is the intended functionality of opaque objects, right?

src/modules/batch/main_impl.h
```diff
@@ -0,0 +1,206 @@
...
+        return 0;
+    }
+
+    if (batch->len > 0) {
+        int strauss_ret = secp256k1_ecmult_strauss_batch_internal(&ctx->error_callback, batch->data, &resj, batch->scalars, batch->points, &batch->sc_g, batch->len);
+        int mid_res = secp256k1_gej_is_infinity(&resj);
+
+        /* `_strauss_batch_internal` should not fail due to insufficient memory.
+         * `batch_create` will allocate memeory needed by `_strauss_batch_internal`. */
+        VERIFY_CHECK(strauss_ret != 0);
```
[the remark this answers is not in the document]
real-or-random: The macro `VERIFY_CHECK(code)` is defined to be the empty string in production builds, that's why you need to suppress the warning using `(void)strauss_ret;`.

include/secp256k1_batch.h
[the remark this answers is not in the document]
fjahr: Hm, never mind, it seems it's not needed anymore with the latest version of my bitcoin core code.

src/modules/batch/tests_impl.h
```diff
@@ -0,0 +1,213 @@
...
+    secp256k1_context *sign = secp256k1_context_create(SECP256K1_CONTEXT_SIGN);
+    secp256k1_context *vrfy = secp256k1_context_create(SECP256K1_CONTEXT_VERIFY);
+    secp256k1_context *both = secp256k1_context_create(SECP256K1_CONTEXT_SIGN | SECP256K1_CONTEXT_VERIFY);
+    secp256k1_context *sttc = secp256k1_context_clone(secp256k1_context_no_precomp);
+    secp256k1_batch *batch_none;
+    secp256k1_batch *batch_sign;
+    secp256k1_batch *batch_vrfy;
+    secp256k1_batch *batch_both;
+    secp256k1_batch *batch_sttc;
+    unsigned char aux_rand16[32];
```
fjahr, line 71: Getting another warning here when compiling together with Bitcoin Core:

```
src/modules/batch/tests_impl.h:68:19: warning: mixing declarations and code is a C99 extension [-Wdeclaration-after-statement]
    unsigned char aux_rand16[32];
                  ^
```

fjahr:
> Here is a rebased branch of the code that I am using: https://github.com/fjahr/secp256k1/tree/pr1134-rebase-2024

FWIW, I have updated the rebased code recently and also added the changes necessary to build the module with cmake.

src/modules/batch/tests_impl.h
fjahr: Thanks @sipa, yeah, I should have added that I have this fixed now in my rebase branch: https://github.com/fjahr/secp256k1/commit/7c6b9df9ec8fe20fb228ead90d494e95f18c8344#diff-fa3fe2044a5385d080fb813c89c9d0707b50ec7083cef9cd9c114b1f2e5483c3R41

siv2r:
Just updated the PR on top of the current master and added a CMake build option. Planning to work on Pippenger next. Would appreciate any reviews in the meantime. I’ll fix the failed CI checks soon.

README.md
```diff
@@ -22,6 +22,7 @@ Features:
 * Optional module for Schnorr signatures according to [BIP-340](https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki).
 * Optional module for ElligatorSwift key exchange according to [BIP-324](https://github.com/bitcoin/bips/blob/master/bip-0324.mediawiki).
 * Optional module for MuSig2 Schnorr multi-signatures according to [BIP-327](https://github.com/bitcoin/bips/blob/master/bip-0327.mediawiki).
+* Optional module for Batch Verification (experimental).
```
fjahr: Could also mention "according to BIP340" here since batch verification is also specified there.

ci/ci.sh
```diff
@@ -80,6 +80,7 @@ esac
     --enable-module-extrakeys="$EXTRAKEYS" \
     --enable-module-schnorrsig="$SCHNORRSIG" \
     --enable-module-musig="$MUSIG" \
+    --enable-module-batch="$BATCH"
```
fjahr: Missing `\` here?

src/tests.c
```diff
@@ -5026,6 +5003,54 @@ static int test_ecmult_multi_random(secp256k1_scratch *scratch) {
         }
     }
 
+    /* number of (scalars, points) inputs generated */
+    *inp_len = filled;
+    /* number of non-zero (scalars, points) inputs */
+    *nonzero_inp_len = num_nonzero;
+    /* ptr to g_scalar*/
+    g_scalar = g_scalar_ptr;
+    /* is mulciplicand of g nonzero? */
```
fjahr: nit: typo

```suggestion
    /* is multiplicand of g nonzero? */
```

src/modules/schnorrsig/tests_impl.h
```diff
@@ -242,6 +263,9 @@ static void test_schnorrsig_bip_vectors(void) {
         };
         test_schnorrsig_bip_vectors_check_signing(sk, pk, aux_rand, msg, sizeof(msg), sig);
         test_schnorrsig_bip_vectors_check_verify(pk, msg, sizeof(msg), sig, 1);
+        #ifdef ENABLE_MODULE_BATCH
```
fjahr: nit: Not sure if that is something that would be popular but I think you could limit these ifdefs to just the test function itself by just having the function body be enclosed in an batch ifdef. At least that would save a few lines of code. Just an idea though, this doesn't bother me much.

src/modules/batch/main_impl.h
```diff
@@ -0,0 +1,208 @@
...
+ *             respective scalars. To be used in Multi-Scalar Multiplication
+ *             algorithms such as Strauss and Pippenger.
+ *    scalars: pointer to scalars allocated on the scratch space.
+ *     points: pointer to points allocated on the scratch space.
+ *       sc_g: scalar corresponding to the generator point (G) in Multi-Scalar
+ *             Multiplication equation.
+ *     sha256: contains hash of all the inputs (schnorrsig/tweaks) present in
+ *             the batch object, expect the first input. Used for generating a random secp256k1_scalar
+ *             for each term added by secp256k1_batch_add_*.
+ *     sha256: contains hash of all inputs (except the first one) present in the batch.
```
fjahr: `sha256` arg is listed twice here

src/modules/batch/main_impl.h
```diff
@@ -0,0 +1,208 @@
...
+    batch_size = sizeof(secp256k1_batch);
+    batch = (secp256k1_batch *)checked_malloc(&ctx->error_callback, batch_size);
+    batch_scratch_size = secp256k1_batch_scratch_size(max_terms);
+    if (batch != NULL) {
+        /* create scratch space inside batch object, if that fails return NULL*/
+        batch->data = secp256k1_scratch_create(&ctx->error_callback, batch_scratch_size);
+        if (batch->data == NULL) {
+            return NULL;
+        }
+        /* allocate memeory for `max_terms` number of scalars and points on scratch space */
```
fjahr: nit

```suggestion
        /* allocate memory for `max_terms` number of scalars and points on scratch space */
```

include/secp256k1_batch.h
```diff
@@ -0,0 +1,110 @@
...
+ */
+SECP256K1_API int secp256k1_batch_usable(
+    const secp256k1_context *ctx,
+    const secp256k1_batch *batch
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2);
+
+/** Verify the set of schnorr signatures or tweaked pubkeys present in the secp256k1_batch.
+ *
+ *  Returns: 1: every schnorrsig/tweak (in batch) is valid
+ *           0: atleaset one of the schnorrsig/tweak (in batch) is invalid
```
fjahr: nit

```suggestion
 *           0: at least one of the schnorrsig/tweak (in batch) is invalid
```

include/secp256k1_batch.h
```diff
@@ -0,0 +1,110 @@
...
+ *       - case 1: unparsable input (schnorrsig or tweak check)
+ *       - case 2: unusable (or invalid) batch
+ *  Calling `secp256k1_batch_usable` beforehand helps eliminate case 2 if
+ *  `secp256k1_batch_add_*` fails.
+ *
+ *  If you ignore the above advice, all the secp256k1_batch APIs will still
+ *  work correctly. It simply makes it hard to understand the reason behind
+ *  `secp256k1_batch_add_*` failure (if occurs).
+ */
+SECP256K1_API int secp256k1_batch_usable(
```
fjahr: I understand the rationale given here but I am still not sure we really need this function. As far as I understand this only returns what result is set to and only the verify function could set result to false. So this means for this to be useful, the user would first need to have called verify and ignored the result. I am not sure in which context this would really be needed, maybe in some concurrency scenarios? But even then it feels like users should be handling that differently. Happy to be convinced otherwise but currently I would vote to remove it if there isn't some use case I am missing.

fjahr (commented):
Thanks for the update @siv2r! Leaving a few comments from a superficial read-through. I will integrate the latest code into the Bitcoin Core PR and take a deeper look soon.

ci/ci.sh
siv2r: Ah, yes. This is one of the reasons the CI fails. Thanks!

fjahr:
@siv2r I think you can drop the experimental status of the module from this PR. The flag is signalling potential API instability and while that is certainly still the case within this PR now, I think this will not be the case anymore once this PR is in a position to be merged. Once it is merged, we might start using it in Bitcoin Core soon after and dealing with the experimental flag would be annoying in this case and arguably once the code here is used for block validation it should probably not be considered experimental anymore. For comparison: The silent payments module PR also doesn't mark the module experimental at the moment.

siv2r:
I agree. I was wondering whether it should be removed or not, and your points make a lot of sense. Will remove it.

siv2r:
I’m re-thinking some of the earlier design choices. Would appreciate some feedback on them.
<details>
<summary>Some Context on Batch API workflow</summary>

Here’s how it currently works:

1. User creates a batch object with `create`
2. Adds verification checks using `add_schnorrsig` and `add_tweak_check`
    - These store scalar-point pairs from Schnorr signature checks (`s*G = R + e*P`) and tweak checks (`Q = P + t*G`)
3. Runs verification with `batch_verify`

There are a couple features:

- Transparent Verification: When the batch object gets full and user tries to add more schnorrsig/tweak checks, it automatically runs verification, stores the result, and clears the scratch space so they can keep adding more checks.
- Pre defined sizes (suggested by @jonasnick, not implemented yet): Instead of letting users create batch objects with any arbitrary `n` value, we could offer preset sizes like `SMALL`, `MEDIUM`, and `LARGE`.
</details>

Right now the batch module only supports Strauss (it calls `ecmult_strauss_batch` internally), and I'm exploring how to add Pippenger support.  

My first thought was: what if we just call `ecmult_multi_var` directly in the `batch_verify` API? This would let `ecmult_multi_var` handle choosing between Strauss or Pippenger, which seems like it would solve our problem pretty cleanly.

But there's an issue - our current batch object doesn't meet the requirements that `ecmult_multi_var` expects. The function needs an empty scratch space from its callers, but our batch object pre-fills the scratch space with `n` scalar-point pairs, and `ecmult_multi_var` doesn't recognize these pre-filled scalar-point pairs.
<details>
<summary>Some Context on `ecmult_multi_var` Caller Requirements</summary>

When we want to multi-multiply `n` scalar-point pairs using `ecmult_multi_var`, our caller needs to provide three things:

1. **Callback data (`cbdata`)**: A caller-defined data structure that stores the `n` scalar-point pairs (in whatever format works for the caller)
2. **Callback function**: A caller-defined function that extracts a single scalar-point pair from `cbdata` for each index `i < n`
3. **Empty scratch space**: `ecmult_multi_var` partitions the `n` inputs into batches that fit in this scratch space. It then runs either Strauss or Pippenger per batch. These algorithms stream `n_batch_points` scalar-point pairs into the scratch space by repeatedly calling the callback function on `cbdata` with incremental indices.
</details>

---

We have two ways of designing the batch object:

**Design 1**: Have only one scratch space that is pre-filled with scalar-point pairs and call `strauss_ecmult_batch` or `pippenger_ecmult_batch` internally. This pull request attempts this design.

**Design 2**: Have two scratch spaces - one that is pre-filled with scalar-point pairs and another that stays empty. Now we can call `ecmult_multi_var` with the empty scratch space and use the pre-filled scratch space as our `cbdata`.

At first glance, Design 1 seems more memory efficient since it only uses one scratch space. I started implementing this design without much hesitation, but it has some problems.

It's not straightforward to design a pre-filled scratch space that works with both `ecmult_strauss_batch` and `ecmult_pippenger_batch`. Strauss stores the `n` scalar-point pairs as-is in the scratch space, while Pippenger splits each pair (GLV endomorphism), so it stores `2n+2` points in the scratch space. Trying to create a generic scratch space that handles both feels like re-implementing the `ecmult_multi_var` function all over again.

So, are we better off implementing Design 2?

Alternatively, we could refactor `ecmult_multi_var` to recognize a pre-filled scratch space, though I'm not sure about the implementation details. My concern is that this might make the API more complicated.

The issue with Design 2 is that it seems to copy data from one scratch space to another. When we give `ecmult_multi_var` two scratch spaces, it copies the `n` scalar-point pairs from the pre-filled one to the empty one using our callback function. Though with Pippenger, it's not a direct copy since it splits each pair first. But maybe we're okay with this overhead?

**TLDR**: Is Design 2 the way to go? Or should we consider refactoring `ecmult_multi_var`?

include/secp256k1_batch.h
siv2r: Good question! The verify function isn't just called by users - it can also be called internally by the `batch_add_` APIs. We call this transparent verification.

Here's how it works: when a user creates a batch with `batch_create`, they specify a capacity for `n` items (schnorrsig or tweak checks). If they try to add more items than the capacity, the `batch_add_` APIs automatically run verification and clear out the batch object so the user can keep adding items.

Now, if this intermediate verification call fails, there's no point in continuing to add more items to the batch object because the final verify (that the user will call) is going to fail anyway. With this API, users can detect the intermediate verification failure and abort early.

Beyond this scenario, the API doesn't have much practical use. Users who know they'll stay within their batch capacity probably won't need this API. Since we don't provide a `batch_storage_left` API to show how many more items they can add, users might accidentally exceed the capacity without realizing it?

siv2r: I think the documentation doesn't explain this properly. I'll be sure to update them.

README.md
siv2r: Thanks. I'll update this to say "BIP340-compatible batch verification".

The main reason I didn't mention BIP340 initially was that this module doesn't strictly follow the specification. BIP340 requires generating randomizers by seeding ChaCha20 with the hash of all inputs, but with our API design, we don't know all the inputs beforehand. So we simply hash whatever inputs we have at the time.

We could technically defer randomizer generation until the verify function runs, so we'd know all inputs before creating the randomizer. But with transparent verification involved, there's no guarantee that users will stop adding inputs after the verify function is called. So, I don't see any huge benefit.

doc/speedup-batch/bench.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
jonasnick:
```suggestion
#!/usr/bin/env bash
```

src/modules/schnorrsig/bench_impl.h
```diff
@@ -45,6 +54,23 @@ static void bench_schnorrsig_verify(void* arg, int iters) {
...
+static void bench_schnorrsig_verify_n(void* arg, int iters) {
+    bench_schnorrsig_data *data = (bench_schnorrsig_data *)arg;
+    int i, j;
+
+    for (j = 0; j < iters/data->n; j++) {
+        for (i = 0; i < data->n; i++) {
+            secp256k1_xonly_pubkey pk;
+            CHECK(secp256k1_xonly_pubkey_parse(data->ctx, &pk, data->pk[j+i]) == 1);
+            CHECK(secp256k1_batch_usable(data->ctx, data->batch) == 1);
+            CHECK(secp256k1_batch_add_schnorrsig(data->ctx, data->batch, data->sigs[j+i], data->msgs[j+i], MSGLEN, &pk) == 1);
```
jonasnick: Shouldn't the array elements in this block be indexed as `[j*data->n + i]` (same for tweak checks)?

include/secp256k1_batch.h
jonasnick: > Now, if this intermediate verification call fails, there's no point in continuing to add more items to the batch object because the final verify (that the user will call) is going to fail anyway. With this API, users can detect the intermediate verification failure and abort early.

If there is intermediate verification failure, wouldn't users already detect that because the *_add function fails? There is already a branch right at the start of `secp256k1_batch_add_schnorrsig` that makes the function fail if the intermediate result is 0:
https://github.com/bitcoin-core/secp256k1/blob/569c06f2af7f94f452855fe86117c1713a35e104/src/modules/schnorrsig/batch_add_impl.h#L93-L95

What appears to be inconsistent right now is when `secp256k1_batch_add_schnorrsig` does internal verification of the batch, the return value is ignored.
https://github.com/bitcoin-core/secp256k1/blob/569c06f2af7f94f452855fe86117c1713a35e104/src/modules/schnorrsig/batch_add_impl.h#L113-L115

If `secp256k1_batch_add_schnorrsig` would return 0 in that case, I don't see a use for `batch_unusable`.

EDIT: see below for a different API proposal

jonasnick (commented):
I noticed that @fjahr's PR to Core redefines our internal macro `STRAUSS_MAX_TERMS_PER_BATCH` and doesn't use the transparent verification feature, which may indicate that we can improve our API design or documentation.

```
// This is the maximum number of scalar-point pairs on the batch for which
// Strauss' algorithm, which is used in the secp256k1 implementation, is
// still efficient. TODO: This will be changed when Pippenger algorithm is
// being used in the secp implementation too.
const size_t MAX_BATCH_SIZE{106};

BatchSchnorrVerifier::BatchSchnorrVerifier() {
    unsigned char rnd[16];
    GetRandBytes(rnd);
    secp256k1_batch* batch{secp256k1_batch_create(secp256k1_context_static, MAX_BATCH_SIZE, rnd)};
    m_batch = batch;
}
```

jonasnick (commented):
Some of the regular contributors discussed this PR in person and we landed on the following proposal for the API:

```C
/* Change the max_terms argument to max_memory. This does not really work with
 * our current implementation of ecmult_multi, but we came up with a plan
 * to fix this. So, we'd need to refactor ecmult_multi first before switching to
 * the max_memory argument. */
SECP256K1_API secp256k1_batch* secp256k1_batch_create(
    const secp256k1_context* ctx,
    size_t max_memory,
    const unsigned char *aux_rand16
) SECP256K1_ARG_NONNULL(1) SECP256K1_WARN_UNUSED_RESULT;

/* This function allows processing a new batch without reallocating memory for
 * the terms. It would set batch->len = 0 and batch->result = 1. */
SECP256K1_API int secp256k1_batch_reset(
    const secp256k1_context *ctx,
    secp256k1_batch *batch
) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2);

/* batch_destroy and batch_verify are the same as in the current API.
 * batch_verify returns 1 for an empty batch. batch_usable is removed. */
```

```C
/* Change the return type to void. If batch->result == 0, then return
 * immediately. If parsing or transparent verification fails, set batch->result
 * to 0 and return. */
SECP256K1_API void secp256k1_batch_add_schnorrsig(
    const secp256k1_context* ctx,
    secp256k1_batch *batch,
    const unsigned char *sig64,
    const unsigned char *msg,
    size_t msglen,
    const secp256k1_xonly_pubkey *pubkey
) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(6);
```

This API is very simple for the caller. They need to always call verify, which minimizes the number of branches.

```C
for (i = 0; i < N_SIGS; i++) {
    batch_add_schnorrsig(batch, sig[i], msg[i], sizeof(msg[i]), &pk);
}
if(!batch_verify(ctx, batch)) {
    return 0;
}
```

jonasnick:
By the way, I vibe-coded pippenger support into the current PR, which resulted in a 1.75x speedup for 17,000 signatures.

include/secp256k1_batch.h
siv2r: > If there is intermediate verification failure, wouldn't users already detect that because the *_add function fails?

Yes, AFAIR, *_add functions can fail for two reasons: invalid inputs (parsing errors) or transparent verification failures. The `batch_usable` API helps distinguish between these scenarios.

I agree that the current API is confusing. I like the newly proposed APIs. I'll start implementing them.

README.md
siv2r: Fixed

src/tests.c
siv2r: Fixed

src/modules/batch/main_impl.h
siv2r: Fixed

src/modules/batch/main_impl.h
siv2r: Fixed

include/secp256k1_batch.h
siv2r: Fixed

src/modules/schnorrsig/tests_impl.h
siv2r: Ahh, nice idea. Thanks. Implemented it!

doc/speedup-batch/bench.sh
siv2r: fixed

fjahr: Hm, did you forget to push these latest changes or pushed them to a different place? This still seems to be unchanged.

siv2r: The changes are still in my local. Was planning to create a commit & push after address this comment too: https://github.com/bitcoin-core/secp256k1/pull/1134#discussion_r2444894063

src/modules/schnorrsig/bench_impl.h
siv2r: Yes. The current code reuses inputs between different `iter` runs. Will update it. Thank you!

siv2r: Fixed

doc/speedup-batch/bench.sh
siv2r: Pushed the changes now. Sorry, I should've pushed before commenting.

siv2r:
Rebased after the new test framework. The new API design is still to be implemented

siv2r:
> SECP256K1_API int secp256k1_batch_reset(
>     const secp256k1_context *ctx,
>     secp256k1_batch *batch
> ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2);

Shoule we make this a `void` function instead? It always returns 1, never 0.

siv2r:
I've implemented the new API design (https://github.com/bitcoin-core/secp256k1/pull/1134#pullrequestreview-3387343738). Currently, `batch_reset` returns `void` (happy to change it to `int` if needed).

Once the `ecmult_multi_var` refactoring lands, I'll follow up with:
- Update `batch_create` to use `max_memory` parameters
- Remove the `strauss_batch` wrapper and call `ecmult_multi_var` directly in `batch_verify`

w0xlt:
Concept ACK

fjahr:
FYI, I have updated the [Bitcoin Core PR](https://github.com/bitcoin/bitcoin/pull/29491) to use the new API from here and also finally did all the fixups so the CI is green there now as well.

src/CMakeLists.txt
```diff
@@ -7,6 +7,11 @@ set_property(TARGET secp256k1 PROPERTY PUBLIC_HEADER
 
 # Processing must be done in a topological sorting of the dependency graph
 # (dependent module first).
+if(SECP256K1_ENABLE_MODULE_BATCH)
+  add_compile_definitions(ENABLE_MODULE_BATCH=1)
```
fjahr, line 11: Maybe we want to check here if at least one of schnorr or extrakeys is enabled?

siv2r:
Rebased the pull request on the current master

siv2r:
I integrated the ecmult refactor (#1789) into batch module as a PoC (see my [`new-ecmult-batch`](https://github.com/siv2r/secp256k1/tree/new-ecmult-batch) branch), and the results are looking great!

For context, the current PR (using strauss only) gives a ~1.20x speedup for schnorrsig batch verification and ~1.50x speedup for tweaked pubkey check batch verification. With the new ecmult refactor, the batch module can now use pippenger (with auto algorithm selection), and the speedups are significantly better:

| `mem_limit` | Schnorrsig | Tweak check |
|---|---|---|
| 256 KB | ~1.55x (flatlines after ~320 sigs) | ~1.55x (flatlines after ~640 checks) |
| 4 MB | ~1.78x (flatlines after ~5,700 sigs) | ~1.68x (flatlines after ~11,400 checks) |
| 16 MB | ~1.95x (flatlines after ~24,000 sigs) | ~1.73x (flatlines after ~48,000 checks) |

The speedup flatlines because transparent verification kicks in once the batch reaches capacity (determined by `mem_limit` during `batch_create`). With larger `mem_limit`, more points fit into a single batch, allowing pippenger to use higher window sizes (lower per-point cost). At 16 MB, we hit pippenger w12 (the highest window available), so this is close to the global maximum we can reach.

<details>
<summary> Speedup graphs (256 KB mem_limit)</summary>

#### Schnorrsig batch verification (~1.55x)
schnorrsig-256kb


#### Tweaked x-only pubkey check (~1.55x)
tweakcheck-256kb

</details>

<details>
<summary> Speedup graphs (4 MB mem_limit)</summary>

#### Schnorrsig batch verification (~1.78x)
schnorrsig-4mb


#### Tweaked x-only pubkey check (~1.68x)
tweakcheck-4mb

</details>

<details>
<summary>Speedup graphs (16 MB mem_limit)</summary>

#### Schnorrsig batch verification (~1.95x)
schnorrsig-16mb


#### Tweaked x-only pubkey check (~1.73x)
tweakcheck-16mb

</details>

alexgleason:
Hi @siv2r and reviewers — wanted to share some downstream experience integrating this PR into [nostr-wasm](https://github.com/fiatjaf/nostr-wasm), a small WASM build of libsecp256k1 used for verifying BIP-340 Nostr event signatures in JS/browser environments.

### 1. Head of `siv2r:batch-verify-interface` does not compile

Commit [`20af821`](https://github.com/bitcoin-core/secp256k1/pull/1134/commits/20af821dd249bee33a7280451b36cef0f119fcfe) (current PR head) fails to build after the rebase. The rebase picked up [`fdb6a91`](https://github.com/bitcoin-core/secp256k1/commit/fdb6a91) "Introduce hash context to support pluggable SHA256 compression", which made `secp256k1_sha256_write` and `secp256k1_sha256_finalize` take a `const secp256k1_hash_ctx *` first parameter — but the batch module's randomizer-gen functions weren't updated to pass it:

- `src/modules/schnorrsig/batch_add_impl.h` lines 18, 20, 21, 22, 26 — `secp256k1_sha256_write(sha256, …)` (missing `hash_ctx`)
- `src/modules/extrakeys/batch_add_impl.h` lines 18, 20, 21, 22, 23, 27 — same
- `src/modules/schnorrsig/batch_add_impl.h` line 130 — `secp256k1_schnorrsig_challenge(&e, &sig64[0], msg, msglen, buf)` missing the new first arg

Reproduce on a clean checkout:

```sh
git clone --branch batch-verify-interface https://github.com/siv2r/secp256k1
cd secp256k1 && ./autogen.sh
./configure --enable-experimental --enable-module-batch
make
# → 14 errors: "too few arguments to function call, expected 4, have 3" etc.
```

@fjahr's [`pr1134-rebase-2024`](https://github.com/fjahr/secp256k1/tree/pr1134-rebase-2024) branch is based on an earlier master (predating `fdb6a91`) and compiles cleanly, so the integration is still possible — but the canonical PR branch needs the `hash_ctx` plumbing propagated into both `batch_add_impl.h` files and into the `_challenge` call site.

### 2. Integration experience

Wrapping the public API in TypeScript bindings was straightforward — `batch_create` / `batch_add_schnorrsig` / `batch_verify` / `batch_destroy` map cleanly to a `verifyEvents(events): boolean[]` higher-level call. A few small UX observations from doing this work:

- **All-or-nothing verify result is fine in practice.** When a chunk fails we just fall back to per-event `schnorrsig_verify` for that chunk to identify the bad event(s). Bisection wasn't tempting. The "transparent verification fails silently" concern jonasnick raised in the original review wasn't an issue at our layer because we control all `batch_add_*` call sites.
- **`batch_reset` is genuinely useful for our workload.** We verify chunks of up to 53 sigs at a time, repeatedly; without `_reset` we pay a `malloc`+free of the Strauss scratch (~250 KB at WINDOW=4) per chunk. Glad the newer siv2r commits added it. fjahr's older rebase lacks `_reset` and the per-call alloc cost is measurable for many small batches.
- **The cap at `STRAUSS_MAX_TERMS_PER_BATCH = 106` (≈53 schnorrsigs) is a sharp edge.** It silently clamps `max_terms`, which is fine, but a `secp256k1_batch_max_schnorrsigs()` (or a documented constant in `secp256k1_schnorrsig_batch.h`) would let callers correctly size their own input chunking without inspecting `src/modules/batch/main_impl.h:8`. We hardcoded `53` and added a comment pointing at the source line, which is brittle.
- **API uses `xonly_pubkey *` rather than serialized bytes.** Forces JS-side callers to make N round trips through `xonly_pubkey_parse` before adding to the batch. A `_batch_add_schnorrsig_serialized` variant taking 32-byte xonly pubkeys would eliminate this for high-throughput verifiers. (Not blocking, just an ergonomic note — happy to file separately if useful.)

### 3. Benchmarks on a non-Bitcoin workload

Posting these because the PR thread has bench numbers for the upstream `bench_schnorrsig`, but I haven't seen application-level numbers shared. This is Schnorr verification for [BIP-340 Nostr events](https://github.com/nostr-protocol/nips), single-thread, Bun/V8 driving libsecp via WASM, on a modern x86 laptop, built with `--with-ecmult-window=4` (size-optimized for browser delivery; ~285 KB WASM binary):

| N events | loop verify | batch verify | speedup |
|---------:|------------:|-------------:|--------:|
|        5 |      2.84ms |       2.40ms |  1.18× |
|       10 |      5.62ms |       3.92ms |  1.43× |
|       25 |     14.72ms |      10.19ms |  1.44× |
|       50 |     27.73ms |      18.10ms |  1.53× |
|      100 |     54.99ms |      35.33ms |  1.56× |
|      200 |    112.66ms |      69.75ms |  1.62× |
|      500 |    271.57ms |     179.60ms |  1.51× |

Even at the smallest ecmult window, 1.4–1.6× wins from N≈10 onward; better than what `bench_schnorrsig` shows because we also amortize event-id SHA-256 and hex-decoding across the batch in our wrapper. The dip at N=500 is from chunking at the 53-sig cap (10 chunks vs 4 chunks at N=200). Happy to share the test harness if it would help in PR review.

Net: the API surface and behavior are exactly what a downstream high-throughput Schnorr verifier needs. Looking forward to seeing this land.

siv2r:
@alexgleason  Thanks for catching the build failure! It's fixed now, the branch should compile fine.

On the cap: yes, we currently limit each batch to ~53 sigs, but that cap won't exist in the final version of this PR. Once the ecmult_multi refactor (#1789) is merged, I'll update this PR to remove it, so there'll be nothing extra to expose for batch sizing.

If you want to try it now, there's a PoC on my [`new-ecmult-batch`](https://github.com/siv2r/secp256k1/tree/new-ecmult-batch) branch that already drops the cap and gives better speedups, up to ~1.95x. The numbers and graphs are in [this earlier comment](https://github.com/bitcoin-core/secp256k1/pull/1134#issuecomment-4147918886).

src/modules/batch/main_impl.h
```diff
@@ -146,8 +148,18 @@ secp256k1_batch* secp256k1_batch_create(const secp256k1_context* ctx, size_t max
     return batch;
 }
 
+void secp256k1_batch_reset(const secp256k1_context *ctx, secp256k1_batch *batch) {
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK_VOID(batch != NULL);
+
+    secp256k1_batch_scratch_clear(batch);
+    secp256k1_batch_sha256_tagged(&batch->sha256);
```
fjahr, line 156: I have noticed that this drops the `aux_rand16` we may have given this batch when we created it. That surprised me and it seems unintentional but if this is intended I think it should be documented.

src/modules/batch/main_impl.h
```diff
@@ -3,4 +3,139 @@
...
+}
+
+secp256k1_batch* secp256k1_batch_create(const secp256k1_context* ctx, size_t max_terms, const unsigned char *aux_rand16) {
+    size_t batch_size;
+    secp256k1_batch* batch;
+    size_t batch_scratch_size;
+    unsigned char zeros[16] = {0};
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(max_terms != 0);
```
fjahr, line 114: Feels like this should be `ARG_CHECK(max_terms >= 2);` since batching doesn't make sense for 1 term either and maybe that can simplify the code below a little.

fjahr (commented):
I spent a bit more time on my core pull and rebased it and found something maybe worth addressing in the meantime.

I can understand if you want to wait for now but I hope we can get concept acks on https://github.com/bitcoin-core/secp256k1/pull/1789 soon and when that is the case I think it would be great if you base the pull here on it. It would make review and downstream work a bit easier, at least for me.

src/modules/batch/main_impl.h
siv2r: Good idea restricting `max_terms` to at least 2. Batching just one (https://github.com/bitcoin-core/secp256k1/pull/1134#discussion_r952908842) would actually be slower than plain `schnorr_verify`, since we'd be deserializing all the points in the verify equation (regular schnorr doesn't deserialize `R.x`).

real-or-random: > since batching doesn't make sense for 1 term either and maybe that can simplify the code below a little.

Just a general comment:
Questions like these are a double-edged sword. Any restriction that we add in the API may save us code complexity but may simply push this complexity to the callers, including error handling and the like. That's why we often end up buying to the complexity in the library to save callers from it. (And callers include language bindings, wrappers, etc., which often have somewhat special considerations).   

(I don't want to take any side here; I haven't thought this case through. This is really supposed to be a general comment rather.)

PR #1137, opened 2022-08-25
Add vcpkg installation instructions

JonLiu1993:
secp256k1 is available as a port in [vcpkg](https://github.com/microsoft/vcpkg), a C++ library manager that simplifies installation for secp256k1 and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build secp256k1, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and [here is what the port script looks like](https://github.com/microsoft/vcpkg/blob/master/ports/secp256k1/portfile.cmake). We try to keep the library maintained as close as possible to the original library. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,19 @@ To create a CSV file for the benchmark result :
 
     $ ./bench_name | sed '2d;s/ \{1,\}//g' > bench_name.csv
 
+Building secp256k1 - Using vcpkg
+------------
+
+You can download and install secp256k1 using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
+
+    git clone https://github.com/Microsoft/vcpkg.git
+    cd vcpkg
+    ./bootstrap-vcpkg.sh
+    ./vcpkg integrate install
+    ./vcpkg install secp256k1
+
+The secp256k1 port in vcpkg is kept up to date by Microsoft team members and community contributors. If the version is out of date, please [create an issue or pull request](https://github.com/Microsoft/vcpkg) on the vcpkg repository.    
+
 Reporting a vulnerability
 ------------
 


README.md
```diff
@@ -116,6 +116,19 @@ To create a CSV file for the benchmark result :
 
     $ ./bench_name | sed '2d;s/ \{1,\}//g' > bench_name.csv
 
+Building secp256k1 - Using vcpkg
+------------
+
+You can download and install secp256k1 using the [vcpkg](https://github.com/Microsoft/vcpkg) dependency manager:
```
hebasto, line 122: "Building" is not "download and install" :)

This file contains no references to any package manager which provides the `secp256k1` package. Seeing no reasons to make an exception for the vcpkg.

real-or-random:
It's really awesome to see that the library is packaged in various places (also some Linux distros) even though we have been lazy with releases. Thanks for these efforts, and we really hope this will be easier once we have a release out. (#1055) And feel free to ping us here whenever there's an issue. 

On the other hand, the one build method that we currently officially support is the one currently described in the README, and yes, I don't think it would make sense for us to have instructions for other build systems in our repo.

But let me stress that it's great that there is a vcpkg!

real-or-random:
@sipa vcpkg is somewhat agnostic of the compiler. It uses  MSVC on Windows but gcc/g++ on Linux.

hebasto:
> if we add cmake as build system (#1113), which presumably will be MSVC compatible

It is MSVC [compatible](https://github.com/hebasto/secp256k1/blob/220628-cmake/README.md#building-on-windows) :)

real-or-random:
Closing this but please feel free to discuss further.

PR #1147, opened 2022-11-02
ci: print env to allow reproducing the job outside of CI

jonasnick:
Example output:

```
WERROR_CFLAGS="-Werror -pedantic-errors"  MAKEFLAGS="-j4"  BUILD="check"  ECMULTWINDOW="auto"  ECMULTGENPRECISION="auto"  ASM="no"  WIDEMUL="int64"  WITH_VALGRIND="no"  EXTRAFLAGS=""  EXPERIMENTAL="no"  ECDH="no"  RECOVERY="yes"  SCHNORRSIG="no"  SECP256K1_TEST_ITERS=""  BENCH="yes"  SECP256K1_BENCH_ITERS="2"  CTIMETEST="yes"  EXAMPLES="yes"  WRAPPER_CMD=""  CC="gcc"  AR=""  NM=""  HOST=""  ./ci/cirrus.sh
```

### DIFF
--- a/ci/cirrus.sh
+++ b/ci/cirrus.sh
@@ -5,6 +5,26 @@ set -x
 
 export LC_ALL=C
 
+# Print relevant CI environment to allow reproducing the job outside of CI.
+print_environment() {
+    # Turn off -x because it messes up the output
+    set +x
+    # There are many ways to print variable names and their content. This one
+    # does not rely on bash.
+    for i in WERROR_CFLAGS MAKEFLAGS BUILD \
+            ECMULTWINDOW ECMULTGENPRECISION ASM WIDEMUL WITH_VALGRIND EXTRAFLAGS \
+            EXPERIMENTAL ECDH RECOVERY SCHNORRSIG \
+            SECP256K1_TEST_ITERS BENCH SECP256K1_BENCH_ITERS CTIMETEST\
+            EXAMPLES \
+            WRAPPER_CMD CC AR NM HOST
+    do
+        eval 'printf "%s %s " "$i=\"${'"$i"'}\""'
+    done
+    echo "$0"
+    set -x
+}
+print_environment
+
 # Start persistent wineserver if necessary.
 # This speeds up jobs with many invocations of wine (e.g., ./configure with MSVC) tremendously.
 case "$WRAPPER_CMD" in


real-or-random:
Are you aware of this?

https://github.com/bitcoin-core/secp256k1/blob/a43e982bca580f4fba19d7ffaf9b5ee3f51641cb/ci/cirrus.sh#L17

real-or-random:
Yeah, it's skipped for successful tasks due to this line:  https://github.com/bitcoin-core/secp256k1/blame/a43e982bca580f4fba19d7ffaf9b5ee3f51641cb/.cirrus.yml#L39

We should remove the line, I don't know why I included this...

jonasnick:
It doesn't seem straightforward to take the output of `cat_test_env` and run cirrus.sh. The lines are not necessarily shell variable assignments due to the spaces and newlines. Even if they were, it's not clear to me how to easily to run `cirrus.sh` with the same environment (and maybe it's even a bit scary, e.g., env contains `PWD`).

If there's no cleverer solution to this, I think it's fine to maintain the list of variables to make it easy to reproduce a test job.

real-or-random (commented):
ACK https://github.com/bitcoin-core/secp256k1/pull/1147/commits/4e54c03153a307af5cc80c7671ac7eef28138326

jonasnick:
We could make it slightly less fragile by setting the shell flag `-u` ("treat unset variables as an error"). See https://github.com/jonasnick/secp256k1/commit/d45ae832fdee9a9333dcbb2c2d01893fff508bd1 (passes CI). This would probably be worth doing regardless of this PR.

real-or-random:
> We could make it slightly less fragile by setting the shell flag `-u` ("treat unset variables as an error"). See [jonasnick@d45ae83](https://github.com/jonasnick/secp256k1/commit/d45ae832fdee9a9333dcbb2c2d01893fff508bd1) (passes CI). This would probably be worth doing regardless of this PR.

We can do that but in what way would this be less fragile?

jonasnick:
Slightly less fragile because the removal or renaming of any relevant variables that misses the newly added lines would result in an error.

real-or-random:
> Slightly less fragile because the removal or renaming of any relevant variables that misses the newly added lines would result in an error.

Hm yes, though having unused variables actually unset (instead of set to the empty string)  is also a good idea in CI because that's what the user will probably do. I'm fine either way, I think.

Issue #1095, opened 2022-03-28
Don't #include <stdio.h> unconditionally


real-or-random:
We currently include <stdio.h> for `fprintf` used in the a) tests and b) in the default error callbacks ... We should not include the header unconditionally. 
 
https://github.com/bitcoin-core/secp256k1/blob/912b7ccc4473b5c969b01d027b8d5dc515435eb5/src/util.h#L16

This is a problem downstream in https://github.com/rust-bitcoin/rust-secp256k1/pull/421 .

tcharding:
Would it be acceptable to 'if guard' any code that calls `malloc` including the `stdlib.h` includes, or is there a cleaner way to solve this. I had a go but its been 5 years since I wrote C and I got stuck (I do not know the correct syntax and could not set `PREALLOC_INTERFACE_ONLY` in the makefile).

E.g., in `util.h`

```
#if !defined PREALLOC_INTERFACE_ONLY
#include <stdlib.h>
#endif

...

#ifndef PREALLOC_INTERFACE_ONLY
static SECP256K1_INLINE void *checked_malloc(const secp256k1_callback* cb, size_t size) {
    void *ret = malloc(size);
    if (ret == NULL) {
        secp256k1_callback_call(cb, "Out of memory");
    }
    return ret;
}

static SECP256K1_INLINE void *checked_realloc(const secp256k1_callback* cb, void *ptr, size_t size) {
    void *ret = realloc(ptr, size);
    if (ret == NULL) {
        secp256k1_callback_call(cb, "Out of memory");
    }
    return ret;
}
#endif
```

real-or-random:
> Would it be acceptable to 'if guard' any code that calls `malloc` including the `stdlib.h` includes, or is there a cleaner way to solve this.

This is in general the way to go. 

But then, when building the library itself, you would then need 'if guard' also functions like `secp256k1_context_create()`. An issue that arises then is that those functions would still be declared in the normal `secp256k1.h` header. This is not terrible but when building programs that use the library, the build would fail at link time when they call these functions accidentally. It would be cleaner not to declare these functions at all. Perhaps this can be done with another "user" macro `SECP256K1_ONLY_PREALLOC` that the user can set before including `secp256k1.h` but headers that change their behavior based on macros can be nasty.

[...]

PR #1148, opened 2022-11-02
Conditionally include `stdio.h` and `stdlib.h`

tcharding:
Currently in order to use `secp256k1` in WASM builds one must patch into the build system empty header files for `stdlib.h` and `stdio.h` [0].

- Patch 1: Removes the unconditional include of `stdio.h` based on the presence of `USE_EXTERNAL_DEFAULT_CALLBACKS` and `VERIFY`.
- Patch 2: Introduces `PREALLOC_INTERFACE_ONLY` and ifndef guards any calls to malloc and friends as well as the include of `stdio.h`.

Works towards fixing: #1095  

1095 does not mention `string.h` but it is a problem as well. I cannot work out how to resolve it so attempting to push this in without it. Happy to extend this further if anyone has any ideas.

### Notes on testing

I could not work out how to test this in `rust-secp256k1` so the I have not proved that this PR removes the requirement to include empty headers described above.

[0] - https://github.com/rust-bitcoin/rust-secp256k1/blob/master/secp256k1-sys/depend/secp256k1.c.patch

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -13,8 +13,11 @@
 
 #include <stdlib.h>
 #include <stdint.h>
-#include <stdio.h>
 #include <limits.h>
+
+#if defined(VERIFY) || !defined(USE_EXTERNAL_DEFAULT_CALLBACKS)
+#include <stdio.h>
+#endif
 
 #define STR_(x) #x
 #define STR(x) STR_(x)
--- a/src/util.h
+++ b/src/util.h
@@ -11,12 +11,15 @@
 #include "libsecp256k1-config.h"
 #endif
 
-#include <stdlib.h>
 #include <stdint.h>
 #include <limits.h>
 
 #if defined(VERIFY) || !defined(USE_EXTERNAL_DEFAULT_CALLBACKS)
 #include <stdio.h>
+#endif
+
+#ifndef PREALLOC_INTERFACE_ONLY
+#include <stdlib.h>
 #endif
 
 #define STR_(x) #x
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -116,34 +116,42 @@ secp256k1_context* secp256k1_context_preallocated_create(void* prealloc, unsigne
 }
 
 secp256k1_context* secp256k1_context_create(unsigned int flags) {
+    secp256k1_context* ctx = NULL;
+
+#ifndef PREALLOC_INTERFACE_ONLY
     size_t const prealloc_size = secp256k1_context_preallocated_size(flags);
-    secp256k1_context* ctx = (secp256k1_context*)checked_malloc(&default_error_callback, prealloc_size);
+    ctx = (secp256k1_context*)checked_malloc(&default_error_callback, prealloc_size);
     if (EXPECT(secp256k1_context_preallocated_create(ctx, flags) == NULL, 0)) {
         free(ctx);
         return NULL;
     }
-
+#endif
     return ctx;
 }
 
 secp256k1_context* secp256k1_context_preallocated_clone(const secp256k1_context* ctx, void* prealloc) {
-    secp256k1_context* ret;
+    secp256k1_context* ret = NULL;
+
+#ifndef PREALLOC_INTERFACE_ONLY
     VERIFY_CHECK(ctx != NULL);
     ARG_CHECK(prealloc != NULL);
 
     ret = (secp256k1_context*)prealloc;
     *ret = *ctx;
+#endif
     return ret;
 }
 
 secp256k1_context* secp256k1_context_clone(const secp256k1_context* ctx) {
-    secp256k1_context* ret;
+    secp256k1_context* ret = NULL;
     size_t prealloc_size;
 
+#ifndef PREALLOC_INTERFACE_ONLY
     VERIFY_CHECK(ctx != NULL);
     prealloc_size = secp256k1_context_preallocated_clone_size(ctx);
     ret = (secp256k1_context*)checked_malloc(&ctx->error_callback, prealloc_size);
     ret = secp256k1_context_preallocated_clone(ctx, ret);
+#endif
     return ret;
 }
 


real-or-random:
> Fix: #1095

Can you remove the magic word fix? I think this PR solves #1095 only partly (`stdlib.h`)?

src/util.h
```diff
@@ -13,9 +13,12 @@
 
 #include <stdlib.h>
 #include <stdint.h>
-#include <stdio.h>
 #include <limits.h>
 
+#ifndef USE_EXTERNAL_DEFAULT_CALLBACKS
+#include <stdio.h>              /* for fprintf */
```
real-or-random, line 19:
```suggestion
#include <stdio.h>
```

style nit: we usually don't say for what we need includes. (And it's not only needed for `fprintf`, it's also needed for `stderr`. )

real-or-random (commented):
Concept ACK, thanks!

This is a good opportunity to move the definitions of `CHECK` and `TEST_FAILURE` (but probably not `EXPECT` to `tests.c`, if you're willing to add a commit. Otherwise we can do it in a follow-up PR.

src/util.h
tcharding: Sure thing, thanks.

tcharding:
> > Fix: #1095
> 
> Can you remove the magic word fix? I think this PR solves #1095 only partly (`stdlib.h`).

bah, sloppy work by me. I'll try first to fix the `stdlib.h` issue in this PR as well. Not sure off the top of my head `string.h` (referring to `rust-secp256k1/` now) because we have a bunch of function declarations in there as well. Will investigate further, sorry for doing half a job.

tcharding:
> Concept ACK, thanks!
> 
> This is a good opportunity to move the definitions of `CHECK` and `TEST_FAILURE` (but probably not `EXPECT` to `tests.c`, if you're willing to add a commit. Otherwise we can do it in a follow-up PR.

Sure thing, happy to look into it.

tcharding:
> Concept ACK, thanks!
> 
> This is a good opportunity to move the definitions of `CHECK` and `TEST_FAILURE` (but probably not `EXPECT` to `tests.c`, if you're willing to add a commit. Otherwise we can do it in a follow-up PR.

I must be missing something, `CHECK` is used in a whole bunch of files, how can we move the definition to `tests.c`? Is there another header file that it could go in, is that what you meant?

```bash
git grep -l ' CHECK\('
src/bench.c
src/bench_ecmult.c
src/bench_internal.c
src/ecmult_impl.h
src/modules/ecdh/bench_impl.h
src/modules/ecdh/tests_impl.h
src/modules/extrakeys/tests_exhaustive_impl.h
src/modules/extrakeys/tests_impl.h
src/modules/recovery/bench_impl.h
src/modules/recovery/tests_exhaustive_impl.h
src/modules/recovery/tests_impl.h
src/modules/schnorrsig/bench_impl.h
src/modules/schnorrsig/tests_exhaustive_impl.h
src/modules/schnorrsig/tests_impl.h
src/tests.c
src/tests_exhaustive.c
src/util.h
src/valgrind_ctime_test.c
```

tcharding:
> This is a good opportunity to move the definitions of CHECK and TEST_FAILURE

I had a play with this @real-or-random but got stuck, leaving for another day. The investigation did lead me however to the observation, now implemented in patch 1, to conditionally include `stdio.h` based on `VERIFY`.

src/util.h
```diff
@@ -11,11 +11,21 @@
...
-#include <stdio.h>
 #include <limits.h>
 
+#ifndef USE_EXTERNAL_DEFAULT_CALLBACKS
+#include <stdio.h>
+#endif
+
+#ifdef VERIFY
+#include <stdio.h>
+#endif
```
jonasnick, line 23: maybe
```
#if defined(VERIFY) && !defined(USE_EXTERNAL_DEFAULT_CALLBACKS)
#include <stdio.h>
#endif
```

jonasnick (commented):
ACK mod nit for the first commit

As for the second commit, if I compile with `PREALLOC_INTERFACE_ONLY` defined, I get a bunch of errors and warnings. I tried to fix them in [my branch](https://github.com/jonasnick/secp256k1/tree/11-03-stdio-jn) until I realized that stdlib is still needed for `checked_malloc/realloc`, i.e., compilation of my branch with `./configure CPPFLAGS="-DPREALLOC_INTERFACE_ONLY -DUSE_EXTERNAL_DEFAULT_CALLBACKS"` will error out because of that (if we manage to get this to work nonetheless we should add a CI test for `PREALLOC_INTERFACE_ONLY` that at least tests successful compilation).

src/util.h
tcharding: Cheers, rebased and used the suggestion in patch 1.

Can a maintainer please confirm you guys use this process, that this rebasing to keep the PR consisting of only the valid patch series.

jonasnick: Yes, that's the process.

tcharding: Thanks mate.

tcharding:
Any clues on the CI fail crew?

> ./build-aux/test-driver: line 109:  3245 Aborted                 (core dumped) "$@" > $log_file 2>&1
> FAIL: tests

Can I view `$log_file` in any way?

real-or-random:
> [api.cirrus-ci.com/v1/task/6006563169632256/logs/cat_tests_log.log](https://api.cirrus-ci.com/v1/task/6006563169632256/logs/cat_tests_log.log)

Oh that's a failure that's expected to occur with a certain low (but still too high to be annoying) probability, see https://github.com/bitcoin-core/secp256k1/pull/367 if you're interested in the details. I've restarted the job.

tcharding:
Thanks!

real-or-random:
To be honest, I think we should take a step back and first discuss whether we would like to separate the headers as I suggested in https://github.com/bitcoin-core/secp256k1/issues/1095#issuecomment-1303537678. That looks like a very fundamental change but it should be backwards compatible.  That would be cleaner and it would make this PR here obsolete.


---

Let me still comment on the PR. 

> As for the second commit, if I compile with `PREALLOC_INTERFACE_ONLY` defined, I get a bunch of errors and warnings. I tried to fix them in [my branch](https://github.com/jonasnick/secp256k1/tree/11-03-stdio-jn?rgh-link-date=2022-11-18T21%3A53%3A06Z) until I realized that stdlib is still needed for `checked_malloc/realloc`, i.e., compilation of my branch with `./configure CPPFLAGS="-DPREALLOC_INTERFACE_ONLY -DUSE_EXTERNAL_DEFAULT_CALLBACKS"` will error out because of that 

The proper way to do this is to just wrap the `malloc`/`realloc` calls in `checked_malloc/realloc` in `util.h` in `#ifndef PREALLOC_INTERFACE_ONLY` (instead the context creation function). Then all malloc calls (and only these) will be gone if you set `PREALLOC_INTERFACE_ONLY`.

Before I came to this realization, I was about to comment that I think we don't want to change `secp256k1_context_preallocated_clone`. Cloning a context into preallocated memory should be fine even if the users sets `PREALLOC_INTERFACE_ONLY` (and the function doesn't need stdlib.h). But yeah, if we simply wrap inside of `checked_malloc/realloc`  in `util.h`, then `secp256k1_context_preallocated_clone` should work. 


In general, it's a little bit inelegant to make `secp256k1_context_create` simply return NULL when `PREALLOC_INTERFACE_ONLY` is set.  Can we instead call a callback, e.g., 
```
            secp256k1_callback_call(&default_illegal_callback, "Use secp256k1_context_preallocated_create instead when compiled with PREALLOC_INTERFACE_ONLY");
```

Arguably that's maybe not terribly useful if you also set `USE_EXTERNAL_DEFAULT_CALLBACKS` and don't have a proper way to abort, but it's certainly better than returning just NULL.

> (if we manage to get this to work nonetheless we should add a CI test for `PREALLOC_INTERFACE_ONLY` that at least tests successful compilation).

Well testing is another problem. Many of the tests currently error out when they can't call `secp256k1_context_create`...

tcharding:
Thanks @real-or-random, I don't think chopping up headers is a good task for someone new to a project, I'm going to leave this one for now. Will move the PR to draft for the discussion to indicate its not a merge candidate.

tcharding:
No forward path, closing.

real-or-random:
> No forward path, closing.

Indeed. It's tracked in #1095, I hope I can come back to this soon. This is a major pain point for users.

PR #1150, opened 2022-11-07
ci: always cat test_env.log

jonasnick:
See https://github.com/bitcoin-core/secp256k1/pull/1147#issuecomment-1303277502


real-or-random:
Concept ACK maybe move all of these commands to `always`? All of them are very cheap, so we can afford to run them.

real-or-random (approved):
ACK 5c9f1a5c3774192f77c97474431bb70a124c7b0c

Issue #1151, opened 2022-11-14
Constant-time tests on macOS


sipa:
I noticed this in the CI output:

[We are sunsetting support for Intel-based macOS instances! Please migrate before January 1st 2023.](https://cirrus-ci.org/guide/macOS/)

MarcoFalke:
Is this fixed now?

real-or-random:
from #1152:

> It drops valgrind and the ctime tests on macOS, as valgrind simply isn't supported anymore there. We may recover some of that functionality by adding MSan (https://github.com/bitcoin-core/secp256k1/pull/1155) for memory-checking there, and adding ctime-using-mtime (TODO).

I think we should leave it open until this has been fixed.

real-or-random:
Oh, the valgrind fork now works on Ventura https://github.com/LouisBrunner/valgrind-macos/issues/54#issuecomment-1377334552 (except for memory leak tracking, which is still WIP but that shouldn't be required for our purposes.) So we should probably https://github.com/bitcoin-core/secp256k1/pull/1152/commits/c0ae48c9950a908b637bff27791fabbe2833c4a5 partly.

hebasto:
>    * [ ]  Enable MSan for MacOS

https://clang.llvm.org/docs/MemorySanitizer.html#supported-platforms:
> MemorySanitizer is supported on the following OS:
>    * Linux
>    * NetBSD
>    * FreeBSD

real-or-random:
Oh indeed, MSan doesn't work on Clang...

>   * [x]  Make the constant-time tests work on MSan

And this has been solved by https://github.com/bitcoin-core/secp256k1/pull/1169.

This means that there's nothing to do for us here except to wait for https://github.com/LouisBrunner/valgrind-macos/issues/56. This will make it possible to run Valgrind on macOS again. 

Leaving this open to track this.

PR #1152, opened 2022-11-14
Update macOS image for CI

sipa:
This upgrades the macOS image from Catalina to Ventura, which also means it's switching from x86_64 to arm64 (addressing #1151).

It drops valgrind and the ctime tests on macOS, as valgrind simply isn't supported anymore there. We may recover some of that functionality by adding MSan (#1155) for memory-checking there, and adding ctime-using-mtime (TODO).

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -107,61 +107,27 @@ task:
   << : *CAT_LOGS
 
 task:
-  name: "x86_64: macOS Catalina"
+  name: "arm64: macOS Ventura"
   macos_instance:
-    image: catalina-base
+    image: ghcr.io/cirruslabs/macos-ventura-base:latest
   env:
     HOMEBREW_NO_AUTO_UPDATE: 1
     HOMEBREW_NO_INSTALL_CLEANUP: 1
-    # Cirrus gives us a fixed number of 12 virtual CPUs. Not that we even have that many jobs at the moment...
-    MAKEFLAGS: -j13
+    # Cirrus gives us a fixed number of 4 virtual CPUs. Not that we even have that many jobs at the moment...
+    MAKEFLAGS: -j5
   matrix:
     << : *ENV_MATRIX
+  env:
+    ASM: no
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -106,62 +111,29 @@ task:
...
   macos_instance:
-    image: catalina-base
+    image: ghcr.io/cirruslabs/macos-ventura-base:latest
   env:
     HOMEBREW_NO_AUTO_UPDATE: 1
     HOMEBREW_NO_INSTALL_CLEANUP: 1
-    # Cirrus gives us a fixed number of 12 virtual CPUs. Not that we even have that many jobs at the moment...
-    MAKEFLAGS: -j13
+    # Cirrus gives us a fixed number of 4 virtual CPUs. Not that we even have that many jobs at the moment...
+    MAKEFLAGS: -j4
```
real-or-random, line 121:
```suggestion
    MAKEFLAGS: -j5
```
maybe that's even better :shrug: but I don't know.

real-or-random (commented):
Concept ACK

Sad that we have to give up on valgrind but this was foreseeable. There's really not much we can do as long as recent version don't support MacOS.

It may be interesting to have MSan as an alternative, both for normal memory issues but also for constant-time tests, see https://www.amongbytes.com/post/20210709-testing-constant-time/ . Its guarantees are somewhat weaker than valgrind (e.g., it does not run on the actual uninstrumented binary) but if it's supported on more platforms, that will be helpful.

real-or-random:
> Also Cirrus supports running on AWS Graviton (native ARM) now. We can probably get some of our ARM64 testing done there (including valgrind).

see #1153

real-or-random (approved):
ACK a8494b02bfe7578ffb76e66924e76c83556a802d

jonasnick (commented):
ACK a8494b02bfe7578ffb76e66924e76c83556a802d

Issue #707, opened 2020-01-06
General CI discussion 


real-or-random:
When thinking about #705 I realized it's maybe a good idea to take a step back and look at this in the context of a general CI revamp. There are multiple things that would be nice and I think some of these probable have a higher priority than coverage reports:

 - ARM tests https://github.com/bitcoin-core/secp256k1/issues/598
 - Big-endian tests https://github.com/bitcoin-core/secp256k1/pull/696
 - Windows and/or MSVC builds?
 - MacOS builds? (MacOS relies on clang anyway, so not this is not as interesting as MSVC.)
 - Coverage reports #705 
 - Reducing configuration options to keep the test matrix manageable, see e.g., https://github.com/bitcoin-core/secp256k1/pull/596#issuecomment-495908309

Is there anything else you would like to see?

If we want to work on this, this is a good time to think about the CI platform. I don't mind Travis and it's reasonably stable in the past months but we think that a different (or additional) provider is a better fit, it's better to discuss this before we start working on these issues.

By the way, Bitcoin is currently looking into GitHub Actions for Windows at least:
https://github.com/bitcoin/bitcoin/issues/17803

gmaxwell:
I think BE is basically solved in travis since there is s390 (lol).  It sounds like the arm is arm64 which is useful but less interesting than 32bit arm.

The big gaps I see in travis:

*  You cannot usefully benchmark, so you can't watch for performance regressions-- the systems it runs on are wildly different in performance and wildly inconsistent in performance because they're VPSes which are concurrently running other loads.

* Getting work product output not supported (seems like a workaround would be possible), meaning you can't get CI [coverage analysis](https://mf4.xiph.org/jenkins/job/opus-coverage/ws/coverage/index.html), or [static analysis reports](https://mf4.xiph.org/jenkins/job/opus-scan-build/ws/scan-build/current/index.html)... except via integrating with other services which creates an unnecessary disparity between the free software tools the developers can use locally and the (usually commercial) services that the project ends up beholden to.

* Travis CPU time is extremely limited (the fact that the travis executors are slow disguises how limited it is, e.g. it takes minutes for what would be tens of seconds locally).  This means that stuff like automated fuzz testing at a scale which is likely to find interesting bugs is essentially off the table, at least as part of CI tests.  This also clearly distorts testing priorities, with stuff like not using valgrind or cutting down tests just to fit under travis' limits.

[...]

PR #1155, opened 2022-11-16
Add MSan CI jobs

sipa:
Add an MSan based sanitizer job, for a number of reasons:
* Does something memcheck (valgrind) like, but can also run on macOS (even post #1152), which can be added as a follow-up.
* We could extend to ctime testing to work on this instead (meaning against ctime testing, with some limitations, can be re-enabled on macOS post #1152).

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -340,6 +340,32 @@ task:
     - ./ci/cirrus.sh
   << : *CAT_LOGS
 
+# Memory sanitizers
+task:
+  << : *LINUX_CONTAINER
+  name: "MSan"
+  env:
+    ECDH: yes
+    RECOVERY: yes
+    SCHNORRSIG: yes
+    CTIMETEST: no
+    CC: clang
+    SECP256K1_TEST_ITERS: 32
+    ASM: no
+  container:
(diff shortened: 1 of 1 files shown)


.cirrus.yml
```diff
@@ -340,6 +340,32 @@ task:
...
+    ASM: no
+  container:
+    memory: 2G
+  matrix:
+    - env:
+        CFLAGS: "-fsanitize=memory -g"
+    - env:
+        ECMULTGENPRECISION: 2
+        ECMULTWINDOW: 2
+        CFLAGS: "-fsanitize=memory -g -O3"
```
real-or-random, line 363: O3 is a nice idea! we should do this for some other jobs too. 

(In another PR -- I added a note in https://github.com/bitcoin-core/secp256k1/issues/707#issuecomment-1317575905)

real-or-random (approved):
utACK 00a42b91b3477f63d6f9b6fe0df02bda8b09cddd

jonasnick (commented):
ACK 00a42b91b3477f63d6f9b6fe0df02bda8b09cddd

PR #1156, opened 2022-11-16
Followups to int128_struct arithmetic

sipa:
This is a follow-up to #1000:
* Add randomized unit tests for int128 logic.
* Add CI for the `_(u)mulh` code path (on non-ARM64 MSVC).
* Add heuristic logic to enable int128_struct based arithmetic on 64-bit MSVC, or systems with pointers wider than 32 bits.
* Fix signed overflow in ARM64 MSVC code.

### DIFF
--- a/src/int128_native_impl.h
+++ b/src/int128_native_impl.h
@@ -3,6 +3,10 @@
 
 #include "int128.h"
 
+static SECP256K1_INLINE void secp256k1_u128_load(secp256k1_uint128 *r, uint64_t hi, uint64_t lo) {
+    *r = (((uint128_t)hi) << 64) + lo;
+}
+
 static SECP256K1_INLINE void secp256k1_u128_mul(secp256k1_uint128 *r, uint64_t a, uint64_t b) {
    *r = (uint128_t)a * b;
 }
@@ -35,6 +39,10 @@ static SECP256K1_INLINE void secp256k1_u128_from_u64(secp256k1_uint128 *r, uint6
 static SECP256K1_INLINE int secp256k1_u128_check_bits(const secp256k1_uint128 *r, unsigned int n) {
    VERIFY_CHECK(n < 128);
    return (*r >> n == 0);
+}
+
+static SECP256K1_INLINE void secp256k1_i128_load(secp256k1_int128 *r, int64_t hi, uint64_t lo) {
+    *r = (((uint128_t)(uint64_t)hi) << 64) + lo;
 }
 
 static SECP256K1_INLINE void secp256k1_i128_mul(secp256k1_int128 *r, int64_t a, int64_t b) {
--- a/src/tests.c
+++ b/src/tests.c
@@ -1977,8 +1977,7 @@ void run_int128_tests(void) {
         secp256k1_uint128 res;
 
         /* Check secp256k1_u128_accum_mul overflow */
-        secp256k1_u128_from_u64(&res, 0);
-        secp256k1_u128_accum_mul(&res, UINT64_MAX, UINT64_MAX);
+        secp256k1_u128_mul(&res, UINT64_MAX, UINT64_MAX);
         secp256k1_u128_accum_mul(&res, UINT64_MAX, UINT64_MAX);
         CHECK(secp256k1_u128_to_u64(&res) == 2);
         CHECK(secp256k1_u128_hi_u64(&res) == 18446744073709551612U);
@@ -1987,8 +1986,7 @@ void run_int128_tests(void) {
         secp256k1_int128 res;
 
         /* Compute INT128_MAX = 2^127 - 1 with secp256k1_i128_accum_mul */
-        secp256k1_i128_from_i64(&res, 0);
-        secp256k1_i128_accum_mul(&res, INT64_MAX, INT64_MAX);
+        secp256k1_i128_mul(&res, INT64_MAX, INT64_MAX);
         secp256k1_i128_accum_mul(&res, INT64_MAX, INT64_MAX);
         CHECK(secp256k1_i128_to_i64(&res) == 2);
         secp256k1_i128_accum_mul(&res, 4, 9223372036854775807);
@@ -1998,8 +1996,7 @@ void run_int128_tests(void) {
         CHECK(secp256k1_i128_to_i64(&res) == INT64_MAX);
 
         /* Compute INT128_MIN = - 2^127 with secp256k1_i128_accum_mul */
-        secp256k1_i128_from_i64(&res, 0);
-        secp256k1_i128_accum_mul(&res, INT64_MAX, INT64_MIN);
+        secp256k1_i128_mul(&res, INT64_MAX, INT64_MIN);
         CHECK(secp256k1_i128_to_i64(&res) == INT64_MIN);
         secp256k1_i128_accum_mul(&res, INT64_MAX, INT64_MIN);
         CHECK(secp256k1_i128_to_i64(&res) == 0);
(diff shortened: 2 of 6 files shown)


src/util.h
```diff
@@ -230,21 +230,34 @@ static SECP256K1_INLINE void secp256k1_int_cmov(int *r, const int *a, int flag)
...
 # define SECP256K1_INT128_NATIVE 1
+#elif defined(_MSC_VER) && (defined(_M_X64) || defined(_M_ARM64))
+/* On 64-bit MSVC targets (x86_64 and arm64), use int128_struct
+ * (which has special logic to implement using intrinsics on those systems). */
+# define SECP256K1_WIDEMUL_INT128 1
+# define SECP256K1_INT128_STRUCT 1
+#elif UINTPTR_MAX > 0xffffffff
+/* Systems with 64-bit pointers (and thus registers) very likely benefit from
+ * using 64-bit based arithmetic (even if we need to fall back to 32x32->64 based
+ * multiplication logic. */
```
real-or-random, line 256:
```suggestion
/* Systems with 64-bit pointers (and thus registers) very likely benefit from
 * using 64-bit based arithmetic (even if we need to fall back to 32x32->64 based
 * multiplication logic). */
```

src/util.h
```diff
@@ -230,21 +230,34 @@ static SECP256K1_INLINE void secp256k1_int_cmov(int *r, const int *a, int flag)
...
 # define SECP256K1_INT128_STRUCT 1
 #elif defined(USE_FORCE_WIDEMUL_INT128)
+/* If USE_FORCE_WIDEMUL_INT128 is set, use int128. */
 # define SECP256K1_WIDEMUL_INT128 1
 # define SECP256K1_INT128_NATIVE 1
 #elif defined(USE_FORCE_WIDEMUL_INT64)
+/* If USE_FORCE_WIDEMUL_INT64 is set, use int64. */
 # define SECP256K1_WIDEMUL_INT64 1
 #elif defined(UINT128_MAX) || defined(__SIZEOF_INT128__)
+/* If __int128 exists, use int128. */
```
real-or-random, line 245:
```suggestion
/* If native 128-bit types exist, use int128. */
```

(because our logic will also happily use `int128_t`, not only `__int128`.)

src/int128_native_impl.h
```diff
@@ -37,6 +41,10 @@ static SECP256K1_INLINE int secp256k1_u128_check_bits(const secp256k1_uint128 *r
    return (*r >> n == 0);
 }
 
+static SECP256K1_INLINE void secp256k1_i128_load(secp256k1_int128 *r, int64_t hi, uint64_t lo) {
+    *r = (((uint128_t)(uint64_t)hi) << 64) + lo;
```
roconnor-blockstream, line 45: why not `((int128_t)hi << 64) + lo`?

roconnor-blockstream: Somehow I missed the fact the left-shift of negative values is UB.  In retrospect, that makes a bit of sense if you cannot count on two's complement, even if the formal definition of multiplying by 2^n still makes sense for negative numbers.

src/util.h
```diff
@@ -230,21 +230,34 @@ static SECP256K1_INLINE void secp256k1_int_cmov(int *r, const int *a, int flag)
...
 # define SECP256K1_WIDEMUL_INT128 1
 # define SECP256K1_INT128_NATIVE 1
+#elif defined(_MSC_VER) && (defined(_M_X64) || defined(_M_ARM64))
+/* On 64-bit MSVC targets (x86_64 and arm64), use int128_struct
+ * (which has special logic to implement using intrinsics on those systems). */
+# define SECP256K1_WIDEMUL_INT128 1
+# define SECP256K1_INT128_STRUCT 1
+#elif UINTPTR_MAX > 0xffffffff
```
real-or-random, line 253:
```suggestion
#elif SIZE_MAX > 0xffffffff
```

maybe this is a better heuristic because `uintptr_t` is an optional type. Though even if it's not defined, your code at least won't error because then `UINTPTR_MAX` will be treated as `0` in the preprocessor`. I don't know.

real-or-random (approved):
utACK 3afce0af7c00eb4c5ca6d303e36a48c91a800459

src/tests.c
```diff
@@ -1752,8 +1715,308 @@ void run_modinv_tests(void) {
...
+    }
+}
+
+void run_int128_tests(void) {
+    {   /* secp256k1_u128_accum_mul */
+        secp256k1_uint128 res;
+
+        /* Check secp256k1_u128_accum_mul overflow */
+        secp256k1_u128_from_u64(&res, 0);
+        secp256k1_u128_accum_mul(&res, UINT64_MAX, UINT64_MAX);
```
roconnor-blockstream, line 1981: maybe just `secp256k1_u128_mul(&res, UINT64_MAX, UINT64_MAX);` without the `secp256k1_u128_from_u64`?

src/tests.c
```diff
@@ -1752,8 +1715,308 @@ void run_modinv_tests(void) {
...
+    }
+    /* test secp256k1_i128_mul */
+    mulmod256(rswr, rsb, rsc, NULL);
+    secp256k1_i128_mul(&swz, sb, sc);
+    load256i128(rswz, &swz);
+    CHECK(secp256k1_memcmp_var(rswr, rswz, 16) == 0);
+    /* test secp256k1_i128_accum_mul */
+    mulmod256(rswr, rsb, rsc, NULL);
+    add256(rswr, rswr, rswa);
+    if (int256is127(rswr)) {
```
roconnor-blockstream, line 1900: how often is this true?

src/tests.c
roconnor-blockstream: I believe what I'm proposing is a small change. I'm proposing changing the logic from

    res = 0
    res += UINT64_MAX*UINT64_MAX
    res += UINT64_MAX*UINT64_MAX

to

    res = UINT64_MAX*UINT64_MAX
    res += UINT64_MAX*UINT64_MAX

It still tests accumulating multiplication overflow.

But maybe it isn't worth changing.

real-or-random (approved):
utACK 99bd3355994a436e25d148c68e097cca11f3c63e

src/util.h
```diff
@@ -230,21 +230,34 @@ static SECP256K1_INLINE void secp256k1_int_cmov(int *r, const int *a, int flag)
...
+/* On 64-bit MSVC targets (x86_64 and arm64), use int128_struct
+ * (which has special logic to implement using intrinsics on those systems). */
+# define SECP256K1_WIDEMUL_INT128 1
+# define SECP256K1_INT128_STRUCT 1
+#elif SIZE_MAX > 0xffffffff
+/* Systems with 64-bit pointers (and thus registers) very likely benefit from
+ * using 64-bit based arithmetic (even if we need to fall back to 32x32->64 based
+ * multiplication logic). */
+# define SECP256K1_WIDEMUL_INT128 1
+# define SECP256K1_INT128_STRUCT 1
```
jonasnick, line 258: Do we have a benchmark for a system where this is the case? @sipa's [benchmarks](https://github.com/bitcoin-core/secp256k1/pull/1000#issuecomment-1306385135) suggest that on 64 bit aarch64, int64 is faster than int128_struct.

jonasnick: Ok, if you think that's a significant improvement on these systems, fine with me.

jonasnick (commented):
utACK 99bd3355994a436e25d148c68e097cca11f3c63e

real-or-random (approved):
ACK 99bd3355994a436e25d148c68e097cca11f3c63e tested this also on MSVC locally with the override, including all the benchmark binaries

PR #1158, opened 2022-11-17
Add a secp256k1_i128_to_u64 function.

roconnor-blockstream:
I wanted to experiment with what would be required to split up `secp256k1_i128_to_i64` between those cases when a signed 64 bit value is being demoted, versus an unsigned 64 bit value is being extracted from the lower bits, and this is the result.

I'm not sure this is a useful PR, so feel free to close it.  However, since it is already written, I figured it is worth at least discussing.

### DIFF
--- a/src/int128_native_impl.h
+++ b/src/int128_native_impl.h
@@ -67,7 +67,12 @@ static SECP256K1_INLINE void secp256k1_i128_rshift(secp256k1_int128 *r, unsigned
    *r >>= n;
 }
 
+static SECP256K1_INLINE uint64_t secp256k1_i128_to_u64(const secp256k1_int128 *a) {
+   return (uint64_t)*a;
+}
+
 static SECP256K1_INLINE int64_t secp256k1_i128_to_i64(const secp256k1_int128 *a) {
+   VERIFY_CHECK(INT64_MIN <= *a && *a <= INT64_MAX);
    return *a;
 }
 
--- a/src/int128_struct_impl.h
+++ b/src/int128_struct_impl.h
@@ -170,8 +170,14 @@ static SECP256K1_INLINE void secp256k1_i128_rshift(secp256k1_int128 *r, unsigned
    }
 }
 
+static SECP256K1_INLINE uint64_t secp256k1_i128_to_u64(const secp256k1_int128 *a) {
+   return a->lo;
+}
+
 static SECP256K1_INLINE int64_t secp256k1_i128_to_i64(const secp256k1_int128 *a) {
-   return (int64_t)a->lo;
+   /* Verify that a represents a 64 bit signed value by checking that the high bits are a sign extension of the low bits. */
+   VERIFY_CHECK(a->hi == -(a->lo >> 63));
+   return (int64_t)secp256k1_i128_to_u64(a);
 }
 
 static SECP256K1_INLINE void secp256k1_i128_from_i64(secp256k1_int128 *r, int64_t a) {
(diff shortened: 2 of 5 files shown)


roconnor-blockstream:
As far as C-semantics is goes, these two functions correspond to the `(uint64_t)(...)` and `(int64_t)(...)` cast operations, which are distinct.

roconnor-blockstream:
Yep I agree with you.  I only meant to add on top of what you were saying.

peterdettman:
I have to object to calling the centered modulo (cmod 2^64) mathematical nonsense.

src/int128_native_impl.h
```diff
@@ -59,7 +59,12 @@ static SECP256K1_INLINE void secp256k1_i128_rshift(secp256k1_int128 *r, unsigned
    *r >>= n;
 }
 
+static SECP256K1_INLINE uint64_t secp256k1_i128_to_u64(const secp256k1_int128 *a) {
+   return *a;
```
real-or-random, line 63: nit, just for clarity:
```suggestion
   return (uint64_t)*a;
```

src/int128_struct_impl.h
```diff
@@ -155,8 +155,14 @@ static SECP256K1_INLINE void secp256k1_i128_rshift(secp256k1_int128 *r, unsigned
...
 }
 
+static SECP256K1_INLINE uint64_t secp256k1_i128_to_u64(const secp256k1_int128 *a) {
+   return a->lo;
+}
+
 static SECP256K1_INLINE int64_t secp256k1_i128_to_i64(const secp256k1_int128 *a) {
-   return (int64_t)a->lo;
+   /* Verify that a represents a 64 bit signed value by checking that the high bits are a sign extension of the low bits. */
+   VERIFY_CHECK(a->hi == -(uint64_t)(0x8000000000000000u == (0x8000000000000000u & a->lo)));
```
real-or-random, line 164:
```suggestion
   VERIFY_CHECK(a->hi == -(a->lo >> 63));
```
I find this easier to read (but it may just be me)

real-or-random (commented):
Concept ACK

src/int128_struct_impl.h
roconnor-blockstream: Thank you.  I was fishing around for a good way to write this.

real-or-random:
I think it would make sense to change the `M62` to unsigned also in `secp256k1_modinv64_normalize_62`, and then let the `VERIFY` part of that function deal with the consequences of this. (It feels cleaner to optimize the cleanness of the real code instead of the test code.)

roconnor-blockstream:
@real-or-random `M62` is also used in `r0 &= M62`.  Are you sure you want to get into the logic of mixing signed and unsigned bitwise operations?

real-or-random:
Ok, you're right, that's not a good idea. I didn't think that through entirely.

roconnor-blockstream:
I still need to add a `/* test secp256k1_i128_to_i64 */` section to `tests.c`.

roconnor-blockstream:
I added a test.

real-or-random (approved):
ACK d21647520532957a78027be1ab606b814a2ec720

roconnor-blockstream:
I've updated my proofs to cover the updates in this PR and the previous  PR #1156. You can "browse" the proof at <https://htmlpreview.github.io/?https://github.com/ElementsProject/simplicity/blob/a3c5213a81ee4cdd51b6b39080258d51f987f199/alectryon/verif_int128_impl.v.html>.

PR #1160, opened 2022-11-22
Makefile: add `-I$(top_srcdir)/{include,src}` to `CPPFLAGS` for precomputed

whitslack:
When performing an out-of-source-tree build, regenerating the source files for the precomputed ecmult tables places them outside the source tree. Then, when they are to be compiled, they cannot find the headers they need because the source tree is absent from their include search path. This appears to have been an oversight, as the relevant `-I` options are present in `libsecp256k1_la_CPPFLAGS` but were missing from `libsecp256k1_precomputed_la_CPPFLAGS`. This PR adds them.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -73,7 +73,9 @@ noinst_HEADERS += examples/random.h
 PRECOMPUTED_LIB = libsecp256k1_precomputed.la
 noinst_LTLIBRARIES = $(PRECOMPUTED_LIB)
 libsecp256k1_precomputed_la_SOURCES =  src/precomputed_ecmult.c src/precomputed_ecmult_gen.c
-libsecp256k1_precomputed_la_CPPFLAGS = $(SECP_INCLUDES) $(SECP_CONFIG_DEFINES)
+# We need `-I$(top_srcdir)/src` in VPATH builds if libsecp256k1_precomputed_la_SOURCES have been recreated in the build tree.
+# This helps users and packagers who insist on recreating the precomputed files (e.g., Gentoo).
+libsecp256k1_precomputed_la_CPPFLAGS = -I$(top_srcdir)/src $(SECP_INCLUDES) $(SECP_CONFIG_DEFINES)
 
 if USE_EXTERNAL_ASM
 COMMON_LIB = libsecp256k1_common.la


Makefile.am
```diff
@@ -73,7 +73,7 @@ noinst_HEADERS += examples/random.h
 PRECOMPUTED_LIB = libsecp256k1_precomputed.la
 noinst_LTLIBRARIES = $(PRECOMPUTED_LIB)
 libsecp256k1_precomputed_la_SOURCES =  src/precomputed_ecmult.c src/precomputed_ecmult_gen.c
-libsecp256k1_precomputed_la_CPPFLAGS = $(SECP_INCLUDES)
+libsecp256k1_precomputed_la_CPPFLAGS = -I$(top_srcdir)/include -I$(top_srcdir)/src $(SECP_INCLUDES)
```
real-or-random, line 76:
```suggestion
# We need `-I$(top_srcdir)/src` in VPATH builds if libsecp256k1_precomputed_la_SOURCES have been cleaned from the source tree and then recreated in the build tree.
# This helps users and packagers who insist on recreating the precomputed files (e.g., Gentoo).
libsecp256k1_precomputed_la_CPPFLAGS = -I$(top_srcdir)/src $(SECP_INCLUDES)
```

`-I$(top_srcdir)/src` suffices. We use relative includes throughout the library, and since `src/precomputed_ecmult*.c` believe they live in `-I$(top_srcdir)/src`, everything they include is reachable from there.

real-or-random (commented):
Okay, I think this is not how we intend users to build the library, but if this one-line patch makes it work for your use case, there's probably no reason not to accept this patch.

hebasto:
I can reproduce the bug when building out-of-source:
```
src/precomputed_ecmult.c:8:10: fatal error: ../include/secp256k1.h: No such file or directory
    8 | #include "../include/secp256k1.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```
```
src/precomputed_ecmult_gen.c:6:10: fatal error: ../include/secp256k1.h: No such file or directory
    6 | #include "../include/secp256k1.h"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

Tested that adding `-I$(top_srcdir)/src` is enough to fix it, as @real-or-random [suggested](https://github.com/bitcoin-core/secp256k1/pull/1160#discussion_r1029197729).

---

> as the relevant `-I` options are present in `libsecp256k1_la_CPPFLAGS` but were missing from `libsecp256k1_precomputed_la_CPPFLAGS`.

I'm not sure about it:

https://github.com/bitcoin-core/secp256k1/blob/751c4354d51fb5b10a80764df627b84e1a5ccd4f/Makefile.am#L95

https://github.com/bitcoin-core/secp256k1/blob/751c4354d51fb5b10a80764df627b84e1a5ccd4f/Makefile.am#L76

Makefile.am
hebasto: > _libsecp256k1_precomputed_la_SOURCES have been cleaned from the source tree_

This is not required to trigger the error. Having them in the build tree is enough for it.

Therefore, the comment could be adjusted.

real-or-random: > This is not required to trigger the error. Having them in the build tree is enough for it.

Hm strange. What are the exact steps you've run to reproduce this?

hebasto: > Hm strange. What are the exact steps you've run to reproduce this?

```
$ ./autogen.sh
$ mkdir ../build && cd ../build
$ ../secp256k1/configure
$ make precompute_ecmult precompute_ecmult_gen
$ ./precompute_ecmult
$ ./precompute_ecmult_gen
$ make
```

real-or-random: Oh indeed, I edited my suggestion.

real-or-random:
@whitslack Friendly ping. :)

whitslack:
> Okay, I think this is not how we intend users to build the library

The reason is to build for multiple ABIs (e.g., x86 and x86_64) without needing to make an extra copy of the sources.

> `-I$(top_srcdir)/src` suffices.

Confirmed.

real-or-random:
@whitslack 

Nice! Can you also add the comments I suggested above? https://github.com/bitcoin-core/secp256k1/pull/1160#discussion_r1029197729

real-or-random (approved):
ACK e862c4af0c5a7300129700d38eff499a836a108d

hebasto (commented):
Post-merge ACK e862c4af0c5a7300129700d38eff499a836a108d.

Issue #1153, opened 2022-11-15
More ARM Linux builds, fewer MacOS builds 


real-or-random:
Cirrus has native ARM64 Linux jobs, so we should use them (see also https://github.com/bitcoin-core/secp256k1/issues/707#issuecomment-902639436)

And if we have that, and since MacOS jobs will run on ARM / M1 after #1152, maybe it makes sense to reduce the number of MacOS jobs a little bit. They really need a long time to start.

real-or-random:
This "half-depends" on the support of Docker as a CI environment on ARM, see https://github.com/cirruslabs/cirrus-ci-docs/pull/905#issuecomment-1330488930

MarcoFalke:
If the Cirrus ARM is a full vm, you could also install docker and then do the extra step of loading the dockerfile in the cirrus yml?

real-or-random:
> This "half-depends" on the support of Docker as a CI environment on ARM, see [cirruslabs/cirrus-ci-docs#905 (comment)](https://github.com/cirruslabs/cirrus-ci-docs/pull/905#issuecomment-1330488930)

This should work now according to https://github.com/cirruslabs/cirrus-ci-docs/issues/1154.

PR #1163, opened 2022-11-29
ci: Native ARM64 jobs

real-or-random:


real-or-random:
CI even fails to start:
```
     Error while loading tasks: UNIMPLEMENTED: Dockerfile as CI environment is not yet supported for Arm containers!!
```
(https://cirrus-ci.com/build/4859893270708224).

So this is going to be much more difficult, closing for now...

real-or-random:
Let's see what CI is doing here after support for Dockerfile CI on ARM has been added. https://github.com/bitcoin-core/secp256k1/issues/1153#issuecomment-1545702240

real-or-random:
Okay, so ARM64 builds work!

But ARM32 builds do not run. I tried to play around with this a bit using the Cirrus CI terminal, and I can't make it work. Giving up for now, but here are some notes:
 - Error message was `Cannot execute binary file: Exec format error`
 - This indicates that the CPU simply does not support AArch32.
 - Indeed, not all ARM64 CPUs are backwards-compatible, but the one the task was running on is a Neoverse N1 and that one has AArch32 support. (https://developer.arm.com/documentation/100616/0301/functional-description/introduction/features)
 - We need `CONFIG_COMPAT` in the kernel, but this was set.
 - I've seen suggestions to use `setarch linux32 -B [cmd]` to run binaries, but this not possible from within unprivileged Docker...

PR #1169, opened 2022-12-06
Add support for msan instead of valgrind (for memcheck and ctime test)

sipa:
This introduces an abstraction layer `src/checkmem.h`, which defines macros for interacting with memory checking tools. Depending on the environment, they're mapped to MemorySanitizer builtins, Valgrind integration macros, or nothing at all.

This means that msan builds immediately benefit from existing undefined memory checks in the tests. It also means those builds result in a `ctime_tests` (new name for `valgrind_ctime_test`) binary that can usefully test constant-timeness (not inside Valgrind, and with the downside that it's not running against a production library build, but it's faster and available on more platforms).

Such an msan-ctime test is added to the Linux x86_64 msan CI job, as an example. More CI cases could be added (e.g. for MacOs or ARM Linux) later.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -330,10 +330,11 @@ task:
     ECDH: yes
     RECOVERY: yes
     SCHNORRSIG: yes
-    CTIMETEST: no
+    CTIMETEST: yes
     CC: clang
     SECP256K1_TEST_ITERS: 32
     ASM: no
+    WITH_VALGRIND: no
   container:
     memory: 2G
   matrix:
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -23,7 +23,7 @@ env:
   SECP256K1_TEST_ITERS:
   BENCH: yes
   SECP256K1_BENCH_ITERS: 2
-  CTIMETEST: yes
+  CTIMETESTS: yes
   # Compile and run the tests
   EXAMPLES: yes
 
@@ -81,9 +81,9 @@ task:
     - env: {WIDEMUL: int128,                 ECDH: yes, SCHNORRSIG: yes}
     - env: {WIDEMUL: int128,  ASM: x86_64}
     - env: {                  RECOVERY: yes,            SCHNORRSIG: yes}
-    - env: {BUILD: distcheck, WITH_VALGRIND: no, CTIMETEST: no, BENCH: no}
+    - env: {BUILD: distcheck, WITH_VALGRIND: no, CTIMETESTS: no, BENCH: no}
     - env: {CPPFLAGS: -DDETERMINISTIC}
-    - env: {CFLAGS: -O0, CTIMETEST: no}
+    - env: {CFLAGS: -O0, CTIMETESTS: no}
     - env: { ECMULTGENPRECISION: 2, ECMULTWINDOW: 2 }
     - env: { ECMULTGENPRECISION: 8, ECMULTWINDOW: 4 }
   matrix:
@@ -128,7 +128,7 @@ task:
   env:
     ASM: no
     WITH_VALGRIND: no
-    CTIMETEST: no
+    CTIMETESTS: no
   matrix:
     - env:
         CC: gcc
@@ -153,7 +153,7 @@ task:
     ECDH: yes
     RECOVERY: yes
     SCHNORRSIG: yes
-    CTIMETEST: no
+    CTIMETESTS: no
   << : *MERGE_BASE
   test_script:
     # https://sourceware.org/bugzilla/show_bug.cgi?id=27008
@@ -172,7 +172,7 @@ task:
     ECDH: yes
     RECOVERY: yes
     SCHNORRSIG: yes
-    CTIMETEST: no
+    CTIMETESTS: no
   matrix:
     - env: {}
     - env: {EXPERIMENTAL: yes, ASM: arm}
@@ -192,7 +192,7 @@ task:
     ECDH: yes
     RECOVERY: yes
     SCHNORRSIG: yes
-    CTIMETEST: no
+    CTIMETESTS: no
   << : *MERGE_BASE
   test_script:
     - ./ci/cirrus.sh
@@ -209,7 +209,7 @@ task:
     ECDH: yes
     RECOVERY: yes
     SCHNORRSIG: yes
-    CTIMETEST: no
+    CTIMETESTS: no
   << : *MERGE_BASE
   test_script:
     - ./ci/cirrus.sh
@@ -223,7 +223,7 @@ task:
     ECDH: yes
     RECOVERY: yes
     SCHNORRSIG: yes
-    CTIMETEST: no
+    CTIMETESTS: no
   matrix:
     - name: "x86_64 (mingw32-w64): Windows (Debian stable, Wine)"
       env:
@@ -246,7 +246,7 @@ task:
     RECOVERY: yes
     EXPERIMENTAL: yes
     SCHNORRSIG: yes
-    CTIMETEST: no
+    CTIMETESTS: no
     # Use a MinGW-w64 host to tell ./configure we're building for Windows.
     # This will detect some MinGW-w64 tools but then make will need only
     # the MSVC tools CC, AR and NM as specified below.
@@ -285,7 +285,7 @@ task:
     ECDH: yes
     RECOVERY: yes
     SCHNORRSIG: yes
-    CTIMETEST: no
+    CTIMETESTS: no
   matrix:
     - name: "Valgrind (memcheck)"
       container:
@@ -330,7 +330,7 @@ task:
     ECDH: yes
     RECOVERY: yes
     SCHNORRSIG: yes
-    CTIMETEST: yes
+    CTIMETESTS: yes
     CC: clang
     SECP256K1_TEST_ITERS: 32
     ASM: no
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -40,8 +40,8 @@ cat_logs_snippet: &CAT_LOGS
       - cat noverify_tests.log || true
     cat_exhaustive_tests_log_script:
       - cat exhaustive_tests.log || true
-    cat_valgrind_ctime_test_log_script:
-      - cat valgrind_ctime_test.log || true
+    cat_ctime_tests_log_script:
+      - cat ctime_tests.log || true
     cat_bench_log_script:
       - cat bench.log || true
     cat_config_log_script:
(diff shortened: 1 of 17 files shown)


src/secp256k1.c
```diff
@@ -116,7 +120,9 @@ secp256k1_context* secp256k1_context_preallocated_create(void* prealloc, unsigne
     /* Flags have been checked by secp256k1_context_preallocated_size. */
     VERIFY_CHECK((flags & SECP256K1_FLAGS_TYPE_MASK) == SECP256K1_FLAGS_TYPE_CONTEXT);
     secp256k1_ecmult_gen_context_build(&ret->ecmult_gen_ctx);
+#if SECP256K1_CHECKMEM_ENABLED
     ret->declassify = !!(flags & SECP256K1_FLAGS_BIT_CONTEXT_DECLASSIFY);
+#endif
```
real-or-random, line 125: Should this raise the illegal callback if `SECP256K1_FLAGS_BIT_CONTEXT_DECLASSIFY && !SECP256K1_CHECKMEM_ENABLED`?

.cirrus.yml
```diff
@@ -328,10 +328,11 @@ task:
     ECDH: yes
     RECOVERY: yes
     SCHNORRSIG: yes
-    CTIMETEST: no
+    CTIMETEST: yes
```
real-or-random, line 331: nit: maybe rename this variable to plural, too

src/secp256k1.c
```diff
@@ -63,14 +60,18 @@ struct secp256k1_context_struct {
...
 
 static const secp256k1_context secp256k1_context_static_ = {
     { 0 },
     { secp256k1_default_illegal_callback_fn, 0 },
-    { secp256k1_default_error_callback_fn, 0 },
-    0
+    { secp256k1_default_error_callback_fn, 0 }
+#if SECP256K1_CHECKMEM_ENABLED
+    , 0
+#endif
```
real-or-random, line 74: (mostly educational) nit. Trailing commas are allowed in initializer lists, even back in C89:  
```suggestion
    { secp256k1_default_error_callback_fn, 0 },
#if SECP256K1_CHECKMEM_ENABLED
    0,
#endif
```

real-or-random (commented):
I'm not entirely sure about dropping the `declassify` member from the struct. I don't see anything immediately wrong about dropping it, but I think the reason why we currently always have it is to make sure the compiled code with valgrind support is as close as possible to the code without valgrind support. This makes testing easier because it (attempts to) avoid yet another config option (plus we save a few preprocessing directives).

src/checkmem.h
```diff
@@ -0,0 +1,85 @@
...
+/* If valgrind integration is desired (through the VALGRIND define), implement the
+ * SECP256K1_CHECKMEM_* macros using valgrind. */
+#if !defined SECP256K1_CHECKMEM_ENABLED
+#  if defined VALGRIND
+#    include <valgrind/memcheck.h>
+#    define SECP256K1_CHECKMEM_ENABLED 1
+#    define SECP256K1_CHECKMEM_UNDEFINE(p, len) VALGRIND_MAKE_MEM_UNDEFINED((p), (len))
+#    define SECP256K1_CHECKMEM_DEFINE(p, len) VALGRIND_MAKE_MEM_DEFINED((p), (len))
+#    define SECP256K1_CHECKMEM_CHECK(p, len) VALGRIND_CHECK_MEM_IS_DEFINED((p), (len))
+#    define SECP256K1_CHECKMEM_RUNNING() (RUNNING_ON_VALGRIND)
```
real-or-random, line 66: When playing around with this on valgrind and reading the [memcheck docs](https://valgrind.org/docs/manual/mc-manual.html#mc-manual.clientreqs), I figured out this trick, which you may or may not take. 
```suggestion
     /* VALGRIND_MAKE_MEM_DEFINED returns 0 iff not running on memcheck.
      * This is more precise than the RUNNING_ON_VALGRIND macro, which
      * checks for valgrind in general instead of memcheck specifically. */
#    define VALGRIND_MAKE_MEM_DEFINED(NULL, 0) != 0)
```

real-or-random (commented):
Changes look good. I still want to test this locally on msan. I'll probably do this tomorrow and ACK it then.

real-or-random (approved):
ACK 2788ae70c2bb444365881ce4727e5325b132b0f3 I also tested that the ctimetest fails on valgrind and msan when I introduce variable-time code

Next step (after this PR) would then to run this on macos.
 
Fwiw, if anyone wonders: Unfortunately, this doesn't work on mingw, which doesn't support sanitizers.

real-or-random:
> @real-or-random There appears to be some work on a clang/mingw crossover ([mstorsjo/llvm-mingw](https://github.com/mstorsjo/llvm-mingw)), but I'm not sure it's mature enough to try (or whether it support msan; only ubsan and asan are listed).

It's probably not worth the effort. I'd expect that the differences between different LLVM OS targets are rather small, or at least in areas such as binary formats, linkage, symbols, etc. and not in actual code generation. This is even more true in our case because we don't really interact with the OS except.

src/checkmem.h
```diff
@@ -0,0 +1,88 @@
...
+
+/* Define a statement-like macro that ignores the arguments. */
+#define SECP256K1_CHECKMEM_NOOP(p, len) do { (void)(p); (void)(len); } while(0)
+
+/* If compiling under msan, map the SECP256K1_CHECKMEM_* functionality to msan.
+ * Choose this preferentially, even when VALGRIND is defined, as msan-compiled
+ * binaries can't be run under valgrind anyway. */
+#if defined(__has_feature)
+#  if __has_feature(memory_sanitizer)
+#    include <stddef.h>
```
real-or-random, line 47: nit: Do we need this here? I think no. But we need it below where we use `NULL`.

real-or-random (approved):
ACK 6ef1afa3ff1437d47ce21aace4409a51ebf81aa5 I also tested that the ctimetest fails on valgrind and msan when I introduce variable-time code

Makefile.am
```diff
@@ -119,18 +116,19 @@ if USE_TESTS
 noinst_PROGRAMS += tests
 tests_SOURCES = src/tests.c
 tests_CPPFLAGS = $(SECP_INCLUDES) $(SECP_TEST_INCLUDES) $(SECP_CONFIG_DEFINES)
-if VALGRIND_ENABLED
```
hebasto, line 122: fc2ff99cfdc7005598826d0485aabe3abb349525 removes the last usage of the `VALGRIND_ENABLED`. The following line can be removed now as well https://github.com/bitcoin-core/secp256k1/blob/31ed5386e8450845c2cdbf1af9dd257937c46e48/configure.ac#L228

hebasto:
Testing 6ef1afa3ff1437d47ce21aace4409a51ebf81aa5 on Ubuntu 22.04:
```
$ ./autogen.sh
$ ./configure --enable-ctime-tests CFLAGS="-fsanitize=memory -O1" CC=clang
...
Build Options:
  with external callbacks = no
  with benchmarks         = yes
  with tests              = yes
  with ctime tests        = yes
  with coverage           = no
  with examples           = no
  module ecdh             = yes
  module recovery         = no
  module extrakeys        = yes
  module schnorrsig       = yes

  asm                     = x86_64
  ecmult window size      = 15
  ecmult gen prec. bits   = 4

  valgrind                = no
  CC                      = clang
  CPPFLAGS                =  
  SECP_CFLAGS             = -O2  -std=c89 -pedantic -Wno-long-long -Wnested-externs -Wshadow -Wstrict-prototypes -Wundef -Wno-overlength-strings -Wall -Wno-unused-function -Wextra -Wcast-align -Wconditional-uninitialized -fvisibility=hidden 
  CFLAGS                  = -fsanitize=memory -O1
  LDFLAGS 
$ make clean
$ make
$ ./libtool --mode=execute ./ctime_tests 
==346524==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x7f87d91c34e2 in secp256k1_scalar_mul secp256k1.c
    #1 0x7f87d9193de5 in secp256k1_ecdsa_sign (/home/hebasto/git/secp256k1/.libs/libsecp256k1.so.1+0xbde5) (BuildId: 8a99d7711aeabba7b1dc311e74fedcef767eb76b)
    #2 0x5618435418c7 in run_tests (/home/hebasto/git/secp256k1/.libs/ctime_tests+0xa48c7) (BuildId: 013cc8d2ba72c4e385d011449343a9ea11a16616)
    #3 0x56184354145b in main (/home/hebasto/git/secp256k1/.libs/ctime_tests+0xa445b) (BuildId: 013cc8d2ba72c4e385d011449343a9ea11a16616)
    #4 0x7f87d8e66d8f  (/lib/x86_64-linux-gnu/libc.so.6+0x29d8f) (BuildId: 69389d485a9793dbe873f0ea2c93e02efaa9aa3d)
    #5 0x7f87d8e66e3f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x29e3f) (BuildId: 69389d485a9793dbe873f0ea2c93e02efaa9aa3d)
    #6 0x5618434bb3a4 in _start (/home/hebasto/git/secp256k1/.libs/ctime_tests+0x1e3a4) (BuildId: 013cc8d2ba72c4e385d011449343a9ea11a16616)

SUMMARY: MemorySanitizer: use-of-uninitialized-value secp256k1.c in secp256k1_scalar_mul
Exiting
```

Is it expected?

real-or-random:
@hebasto Yes, msan is not clever enough to work with our asm. Try disabling asm.

real-or-random (approved):
ACK 39c35e727719536878521fdb0e41e1df8e466d80 I also tested that the ctimetest fails on valgrind and msan when I introduce variable-time code

real-or-random (approved):
ACK 472baae02b5ade196fdc1d229d3d52783bbc3b6f

the third time :)

real-or-random (approved):
ACK 0f088ec11263261497661215c110a4c395acc0ac

hebasto (approved):
ACK 0f088ec11263261497661215c110a4c395acc0ac, I have reviewed the code and it looks OK. Able to build `ctime_tests` using MSan.

real-or-random:
Valgrind is also better for other reasons: Because it works on the normal binary without instrumentation, it also covers ASM, and checked code is exactly what people will run in production. 

We currently don't have any docs for constant-time tests. Creating them will be the first step. And yes, then it probably makes sense to mention limitations.

PR #1171, opened 2022-12-08
Change ARG_CHECK_NO_RETURN to ARG_CHECK_VOID which returns (void) 

real-or-random:
And a few other commits cherry-picked from #1170. In particular I included the first commit because the ARG_CHECK_VOID commit doesn't cleanly apply without this first commit. The new helper function is currently unused but that's not a problem.

I also added a commit to fix the typo that was pointed out in #1126 (which #1170 would solved differently -- by removing the text entirely).


jonasnick (commented):
ACK a49e0940ad671f96533d5a79f2ca1fa4020abc0a

PR #1173, opened 2022-12-12
Don't use compute credits for now

sipa:


real-or-random (approved):
ACK 7e5b22684f4f3e53fa94af84286d21a40dd95525

PR #1183, opened 2022-12-29
Bugfix: pass SECP_CONFIG_DEFINES to bench compilation

sipa:


real-or-random (approved):
utACK c0a555b2ae35d5cca0481ecdd4cff5a1ae314283

We could also add an `AM_CPPFLAGS` that will be used as a default for binaries that don't have a corresponding `_CPPFLAGS` variable. But I'm not convinced that this is a good idea, I don't think there's a very good default... It's probably better to keep setting the CPPFLAGS explicitly for every binary.

apoelstra (approved):
ACK c0a555b2ae35d5cca0481ecdd4cff5a1ae314283

PR #1184, opened 2022-12-30
Signed-digit based ecmult_const algorithm

sipa:
Using some insights learned from #1058, this replaces the fixed-wnaf ecmult_const algorithm with a signed-digit based one. Conceptually both algorithms are very similar, in that they boil down to summing precomputed odd multiples of the input points. Practically however, the new algorithm is simpler because it's just using scalar operations, rather than relying on wnaf machinery with skew terms to guarantee odd multipliers.

The idea is that we can compute $q \cdot A$ as follows:
* Let $s = f(q)$, for some function $f()$.
* Compute $(s_1, s_2)$ such that $s = s_1 + \lambda s_2$, using `secp256k1_scalar_lambda_split`.
* Let $v_1 = s_1 + 2^{128}$ and $v_2 = s_2 + 2^{128}$ (such that the $v_i$ are positive and $n$ bits long).
* Computing the result as $$\sum_{i=0}^{n-1} (2v_1[i]-1) 2^i A + \sum_{i=0}^{n-1} (2v_2[i]-1) 2^i \lambda A$$ where $x[i]$ stands for the *i*'th bit of $x$, so summing positive and negative powers of two times $A$, based on the bits of $v_1.$

The comments in `ecmult_const_impl.h` show that if $f(q) = (q + (1+\lambda)(2^n - 2^{129} - 1))/2 \mod n$, the result will equal $q \cdot A$.

This last step can be performed in groups of multiple bits at once, by looking up entries in a precomputed table of odd multiples of $A$ and $\lambda A$, and then multiplying by a power of two before proceeding to the next group.

The result is slightly faster (I measure ~2% speedup), but significantly simpler as it only uses scalar arithmetic to determine the table lookup values. The speedup is due to the fact that no skew corrections at the end are needed, and less overhead to determine table indices. The precomputed table sizes are also made independent from the `ecmult` ones, after observing that the optimal table size is bigger here (which also gives a small speedup).


src/tests.c
```diff
@@ -2195,6 +2202,31 @@ void run_scalar_tests(void) {
...
+            /* 1 */
+            SECP256K1_SCALAR_CONST(0, 0, 0, 0, 0, 0, 0, 1),
+            /* -1 */
+            SECP256K1_SCALAR_CONST(0xfffffffful, 0xfffffffful, 0xfffffffful, 0xfffffffeul, 0xbaaedce6ul, 0xaf48a03bul, 0xbfd25e8cul, 0xd0364140ul),
+            /* 2^255 */
+            SECP256K1_SCALAR_CONST(0x80000000ul, 0, 0, 0, 0, 0, 0, 0),
+            /* 2^255 - 1 */
+            SECP256K1_SCALAR_CONST(0x7ffffffful, 0xfffffffful, 0xfffffffful, 0xfffffffful, 0xfffffffful, 0xfffffffful, 0xfffffffful, 0xfffffffful),
+        };
+        for (i = 0; (unsigned)i < sizeof(HALF_TESTS) / sizeof(HALF_TESTS[0]); ++i) {
```
real-or-random, line 2219: nit: Maybe introduce a `size_t` instead of reusing `i` and casting it.. Easy given that you anyway open a new scope.  (You could also scope the other uses of `i` properly as a clean up.)

real-or-random (commented):
Whoa nice! The skew correction is really annoying to reason about... 

I just don't know where to get all the review power from.

apoelstra:
Very cool! I wonder if the explanation might be clearer if you said something like "without any further optimizations, k0 would be 2^n - 1 with n = 256; however we are going to combine this scheme with the GLV endomorphism and do a windowed rather than bitwise approach. So instead we solve for k0" rather than "for some constant k0".

In the end I was able to understand what you were doing without much trouble, so what you've written is fine, but it was a bit intimidating to see k0, k1, k2 all introduced at once just as unknowns.

apoelstra (approved):
ACK f16c5008ba0975b6111f4c1f1b5a2e3005c00c8f

apoelstra:
@sipa this looks way better, thanks!

apoelstra (approved):
ACK b4ff4ebbb6486756f33c3f2119bc897d1c75179a

src/ecmult_const_impl.h
```diff
@@ -12,220 +12,242 @@
...
+     * and s2 in range (-2^128, 2^128) mod n. This does work, but is uninteresting:
+     *
+     *   To compute q*A:
+     *   - Let s1, s2 = split_lambda(q)
+     *   - Let R1 = C_256((s1 + 2^256 - 1) / 2, A)
+     *   - Let R2 = C_256((s2 + 2^256 - 1) / 2, lambda*A)
+     *   - Return R1 + R2
+     *
+     * The issue is that while q1 and q2 are small-range numbers, (q1 + 2^256 - 1) / 2 mod n
+     * and (q2 + 2^256 - 1) / 2 mod n are not, undoing the benefit of the splitting.
```
peterdettman, line 116: ~~q1, q2~~ s1, s2

src/ecmult_const_impl.h
```diff
@@ -12,220 +12,242 @@
...
+     * equals ECMULT_CONST_BITS.
+     *
+     * All together, that means k = (2^ECMULT_CONST_BITS + 1 - 2^129) * (1 + lambda) (mod order).
+     */
 
-static void secp256k1_ecmult_const(secp256k1_gej *r, const secp256k1_ge *a, const secp256k1_scalar *scalar, int size) {
-    secp256k1_ge pre_a[ECMULT_TABLE_SIZE(WINDOW_A)];
-    secp256k1_ge tmpa;
+    /* The offset to add to s1 and s2 to make them positive. Equal to 2^128 - 1. */
+    static const secp256k1_scalar S_OFFSET = SECP256K1_SCALAR_CONST(0, 0, 0, 0, 0xfffffffful, 0xfffffffful, 0xfffffffful, 0xfffffffful);
```
peterdettman, line 151: It seems 2^128 would work just as well for the offset. Is it easier to just add 1 to bit 128 than keep this offset around?

src/ecmult_const_impl.h
```diff
@@ -12,220 +12,242 @@
...
+     *   To compute q*A:
+     *   - Let s1, s2 = split_lambda(q)
+     *   - Let R1 = C_256((s1 + 2^256 - 1) / 2, A)
+     *   - Let R2 = C_256((s2 + 2^256 - 1) / 2, lambda*A)
+     *   - Return R1 + R2
+     *
+     * The issue is that while q1 and q2 are small-range numbers, (q1 + 2^256 - 1) / 2 mod n
+     * and (q2 + 2^256 - 1) / 2 mod n are not, undoing the benefit of the splitting.
+     *
+     * The make it work, we apply the corrections to the input scalar, including the division by
```
peterdettman, line 118: ~The~ To make it work...

src/ecmult_const_impl.h
```diff
@@ -12,220 +12,242 @@
...
+     * two, before calling split_lambda. Let's introduce an offset k, and solve for it. We also
+     * add 2^128-1 to s1 and s2 after splitting to make sure we end up with positive scalars.
```
peterdettman, line 120: I guess we technically want "non-negative" scalars (or rephrase as "to make sure neither are negative").

src/ecmult_const_impl.h
```diff
@@ -12,220 +12,242 @@
...
+     *     q*A = R1 + R2
+     * <=> q*A = C_l(s1 + 2^128 - 1, A) + C_l(s2 + 2^128 - 1, lambda*A)
+     * <=> q*A = (2*(s1 + 2^128 - 1) + 1 - 2^l) * A + (2*(s2 + 2^128 - 1) + 1 - 2^l) * lambda*A
+     * <=> q*A = (2*(s1 + s2*lambda) + (2^129 - 1 - 2^l)*(1 + lambda)) * A
+     * <=> q   = 2*(s1 + s2*lambda) + (2^129 - 1 - 2^l)*(1 + lambda) mod n
+     * <=> q   = 2*(q + k) / 2 + (2^129 - 1 - 2^l)*(1 + lambda) mod n
+     * <=> q   = q + k + (2^129 - 1 - 2^l)*(1 + lambda) mod n
+     * <=> k   = (2^l + 1 - 2^129)*(1 + lambda) mod n
+     *
+     * We will process the computation of C_l and C_l in groups of ECMULT_CONST_GROUP_SIZE, so we
```
peterdettman, line 143: Should this say "C_l(v1, A) and C_l(v2, lambda*A)"?

src/ecmult_const_impl.h
peterdettman: Also ~~positive~~ non-negative here.

peterdettman:
> ~Mental note: instead of adding 2128 to the split scalars, I believe it's also possible to (a) conditionally negate the scalar (if it's negative) and then (b) swap the meaning of positive/negate in the table (alternatively, bitwise negate the integers read from the table). This would mean we only need enough table coverage for 128 bits rather than 129, which for window=4 means one fewer addition.~
> 
> EDIT: no, doesn't work

Are you able to explain why that doesn't work?

peterdettman:
Nice idea, simple and clear.

- I would prefer to use 2^128 as the offset, allowing the split to use the full two's complement range of [-2^128, 2^128). That generalizes better to other split schemes (I have in mind the basis reduction of https://ia.cr/2020/454).
- The Straus ladder has the same need to deal with negative split outputs; should the mechanism be unified (one way or the other?)
- I don't have a build to test the performance of the current _scalar_split_lambda, but it's doing quite a bit of unnecessary work compared to the "original" non-modular formula, which it seems could be done cmod 2^129 (i.e. 129-bit two's complement representation) - and e.g. avoid calculating the 384 bits that are shifted away. Not to belabor the point, but if the split outputs were in 129-bit two's complement, adding a 2^128 offset is just a bit flip (or maybe a single limb complement).

src/ecmult_const_impl.h
```diff
@@ -12,220 +12,244 @@
...
+     * and (s2 + 2^256 - 1) / 2 mod n are not, undoing the benefit of the splitting.
+     *
+     * To make it work, we apply the corrections to the input scalar, including the division by
+     * two, *before* calling split_lambda. Let's introduce an offset k, and solve for it. We also
+     * add 2^128 to s1 and s2 after splitting to make sure we end up with non-negative scalars
+     * (a slightly smaller offset would work due to the bounds on the split, but we pick 2^128
+     * for simplicity).
+     *
+     *   To compute q*A:
+     *   - Let s1, s2 = split_lambda((q + k) / 2 mod n)
```
peterdettman, line 125: It's a bit jarring to see the form of the solution magically appear here as (q + k)/2. It could just be 't' (or 'f(q)') and the derivation below works out fine, but without the leap.

src/ecmult_const_impl.h
```diff
@@ -12,220 +12,244 @@
...
+     *   - Let v1 = s1 + 2^128
+     *   - Let v2 = s2 + 2^128
```
peterdettman, line 127: I kinda feel like these should say "mod n", but it's not quite clear. `Let R1 = C_256((s1 + 2^256 - 1) / 2, A)` implies that s1 is a 256-bit value so for v1 to become small enough for the following C_l presumably the mod is needed.

peterdettman:
I think if _scalar_split_lambda just output 129-bit two's complement values (and then sign-extend to ECMULT_CONST_BITS), then they would be directly usable in the comb, no offset needed. Something for the future, perhaps.

src/scalar_8x32_impl.h
```diff
@@ -218,6 +218,37 @@ static void secp256k1_scalar_negate(secp256k1_scalar *r, const secp256k1_scalar
     r->d[7] = t & nonzero;
 }
 
+static void secp256k1_scalar_half(secp256k1_scalar *r, const secp256k1_scalar *a) {
```
real-or-random, line 221: Notes for myself (that could be turned into a comment):
 
```
Interpreting `a` as a bit array, we compute
  a[255..0] / 2 = (a[255..0] - a[0]) / 2 + a[0]/2
                = (a[255..0] >> 1) + (a[0]/2 ?   1/2 : 0)
                = (a[255..0] >> 1) + (a[0]/2 ? n/2+1 : 0),
where n is the group order and in the last equality we have used 1/2 = n/2+1 (mod n).
For n/2, we have constants SECP256K1_N_H_7, ..., SECP256K1_N_H_0.

File: src/scalar_8x32_impl.h [hunk 10]
```diff
@@ -218,6 +218,37 @@ static void secp256k1_scalar_negate(secp256k1_scalar *r, const secp256k1_scalar
...
+    t += SECP256K1_N_H_1 & mask;
+    r->d[1] = t; t >>= 32;
+    t += (uint32_t)((a->d[2] >> 1) | (a->d[3] << 31));
+    t += SECP256K1_N_H_2 & mask;
+    r->d[2] = t; t >>= 32;
+    t += (uint32_t)((a->d[3] >> 1) | (a->d[4] << 31));
+    t += SECP256K1_N_H_3 & mask;
+    r->d[3] = t; t >>= 32;
+    t += (uint32_t)((a->d[4] >> 1) | (a->d[5] << 31));
+    t += mask;
```
real-or-random, line 236:
```suggestion
    t += SECP256K1_N_H_4 & mask;
```
and same below is easier to read and more general.

src/scalar_8x32_impl.h
```diff
@@ -218,6 +218,37 @@ static void secp256k1_scalar_negate(secp256k1_scalar *r, const secp256k1_scalar
...
+    r->d[5] = t; t >>= 32;
+    t += (uint32_t)((a->d[6] >> 1) | (a->d[7] << 31));
+    t += mask;
+    r->d[6] = t; t >>= 32;
+    r->d[7] = (uint32_t)t + (uint32_t)(a->d[7] >> 1) + (SECP256K1_N_H_7 & mask);
+#ifdef VERIFY
+    /* The line above only computed the bottom 32 bits of r->d[7]. Redo the computation
+     * in full 64 bits to make sure no overflow occured. */
+    VERIFY_CHECK((t + (a->d[7] >> 1) + (SECP256K1_N_H_7 & mask)) >> 32 == 0);
+#endif
```
real-or-random, line 249: Add a `VERIFY(secp256k1_scalar_check_overflow(r)) == 0)`? 

(We should probably add a `secp256k1_scalar_verify` and check elements in every function, like for field and group. But that's a separate PR.)

src/scalar_8x32_impl.h
real-or-random: Yeah, I edited the text to remove it. It was a relic of an earlier draft.

jonasnick (commented):
> I didn't include the macro -> function change, as I measure a slight slowdown from it.

Oh, ok. I did not run benchmarks.

1a26896eff6bc2f7b2edbcd0313c3e2e574b08a6 looks good

real-or-random:
needs rebase (and I plan to have a look soon :))

src/ecmult_const_impl.h
```diff
@@ -12,208 +12,251 @@
...
+     * point multiplication in that fashion. Let v be an n-bit non-negative integer (0 <= v < 2^n),
+     * and v[i] its i'th bit (so v = sum(v[i] * 2^i, i=0..n-1)). Then define:
+     *
+     *   C_l(v, A) = sum((2*v[i] - 1) * 2^i*A, i=0..l-1)
+     *
+     * Then it holds that C_l(v, A) = sum((2*v[i] - 1) * 2^i*A, i=0..l-1)
+     *                              = (2*sum(v[i] * 2^i, i=0..l-1) + 1 - 2^l) * A
+     *                              = (2*v + 1 - 2^l) * A
+     *
+     * Thus, one can compute q*A as C_256((q + 2^l - 1) / 2, A). This is the basis for the
```
siv2r, line 119:
```suggestion
     * Thus, one can compute q*A as C_256((q + 2^256 - 1) / 2, A). This is the basis for the
```

jonasnick (approved):
ACK mod nits c8eb787cd96995aefa79bf4bab18903c409db026

I added a couple of (micro-)nit fixups to [my branch](https://github.com/jonasnick/secp256k1/commits/202212_sd_ecmult_const-jn). Most notably, it tries to unify the first iteration in ecmult_const with the loop and adds an explanation for the index computation in `ECMULT_CONST_TABLE_GET_GE`.

real-or-random (commented):
utACK mod these nits

also utACK on jonas' fixups (except a typo and one further comment here: https://github.com/jonasnick/secp256k1/commit/cdf545a9d30cf3805b6e0c34d1baeacefdbf9323 )

I believe I've done a thorough review, but I really have only nits.

stratospher (commented):
ACK c8eb787.

was able to follow the useful comments and code! also observed a 2% speedup on my machine.

jonasnick (approved):
ACK 355bbdf38a2f932daadd02325a0d90d902cb2af4

siv2r (commented):
ACK 355bbdf

real-or-random (approved):
ACK 355bbdf38a2f932daadd02325a0d90d902cb2af4

PR #1178, opened 2022-12-14
Drop `src/libsecp256k1-config.h`

hebasto:
Cherry-picked the first commit from #1142 and addressed a [comment](https://github.com/bitcoin-core/secp256k1/pull/1142#issuecomment-1295099597).

PR #1185, opened 2023-01-03
Drop `SECP_CONFIG_DEFINES` from examples

hebasto:
User applications shouldn't need or rely on `SECP_CONFIG_DEFINES`.

See https://github.com/bitcoin-core/secp256k1/pull/1178#discussion_r1059457252.


real-or-random (approved):
utACK 2f9ca284e2af0c738e701b3dc3e348b87f26a7a2

Issue #1167, opened 2022-12-01
Setting counting_illegal_callback may hide failing tests


jonasnick:
The following code snippet very roughly resembles what happens in the libsecp test harness and illustrates the issue:

```C
/* in the tests we have access to a global context */
static secp256k1_context *ctx = NULL;

void test_foo() {
  int ecount;
  context_set_error_callback(ctx, counting_illegal_callback_fn, &ecount);
  /* do some tests with ecount */
}
void test_bar() {
  /* we don't set the counting_illegal_callback here because we don't want to test that here */
  some_function(ctx);
}
void main() {
  test_foo();
  test_bar();
}
```

The code is fine, until one day `some_function(ctx)` results in the illegal callback being called.
Then we'd want the test to fail but instead what happens is that some stack region formerly known as `ecount` is modified, which does not necessarily result in a crash.

One solution would be to never add the `counting_illegal_callback` to the global context and instead create a local context for counting.

real-or-random:
> One solution would be to never add the `counting_illegal_callback` to the global context and instead create a local context for counting.

Or make the ecount global. :shrug: I guess that's good enough for test code.

apoelstra:
We could write an `EXPECT_ARG_CHECK_FAIL` macro (or maybe something with a nicer name :)) which creates a fresh `ecount` variable at 0, sets the context object's error callbacks, checks the `ecount` is 1, then restores the old callbacks.

This would also get rid of all the ecount-accounting code, which is quite a PITA to maintain and sometimes results in huge diffs.

real-or-random:
We have `CHECK_ILLEGAL` and `CHECK_ILLEGAL_VOID` since e39d954f118a29db2c33e9a9a507053fff5573ed:

https://github.com/bitcoin-core/secp256k1/blob/c545fdc374964424683d9dac31a828adedabe860/src/tests.c#L55-L72

The "only" thing left to do is to use these everywhere (and maybe introduce a similar macro for the error callback).

PR #1186, opened 2023-01-04
tests: Tidy context tests

real-or-random:
This is an improved version of some of the tidying/refactoring in #1170. 

I think it's enough to deserve a separate PR. Once this is merged, I'll get back to the actual goal of #1170 (namely, forbidding cloning and randomizing static contexts.)

This PR is a general clean up of the context tests. A notable change is that this avoids a code smell where `run_context_tests()` would use the global `ctx` variable like a local one (i.e., create a context in it and destroy it afterwards).  After this PR, the global `ctx` is properly initialized for all the other tests, and they can decide whether they want to use it or not. Same for a global `sttc`, which is a memcpy of the static context (we need a writable copy in order to be able to set callbacks).

Note that this touches code which is also affected by #1167 but I refrained from trying to solve this issue. The goal of this PR is simply not to worsen the situation w.r.t. #1167. We should really introduce a macro to solve #1167 but that's another PR.

### DIFF
--- a/src/modules/extrakeys/tests_impl.h
+++ b/src/modules/extrakeys/tests_impl.h
@@ -336,7 +336,6 @@ void test_keypair(void) {
     secp256k1_xonly_pubkey xonly_pk, xonly_pk_tmp;
     int pk_parity, pk_parity_tmp;
     int ecount;
-    secp256k1_context *sttc = secp256k1_context_clone(secp256k1_context_static);
 
     set_counting_callbacks(ctx, &ecount);
     set_counting_callbacks(sttc, &ecount);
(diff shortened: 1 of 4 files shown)


apoelstra:
valgrind does not pass on fc16b698. The next commit, 3c54c4adda8d79cdb574604757ffddc481096c9e, passes.

I don't know if we have it as a goal that every commit passes valgrind, but there you go.

Otherwise ACK 82d6a5a5a090cc920ff92842656e78889a25d0fd

real-or-random:
> I don't know if we have it as a goal that every commit passes valgrind, but there you go.

Not strictly but I was certainly my intention here. Fixed.

I tested every commit with valgrind now. (https://github.com/mhagger/git-test or https://github.com/spotify/git-test are really nice for this!)

apoelstra (approved):
ACK 39e8f0e3d7ba7924e9cc5f9e0c56747e942f1eab

PR #1187, opened 2023-01-05
refactor: Rename global variables in tests

real-or-random:
On top of #1186 . 

I feel that this is an improvement, but it touches a lot of lines and so it deserves a separate discussion.


apoelstra (approved):
ACK 5e03c7c3365e610983e7ee8c4e136b5374774cff

Issue #1037, opened 2021-12-07
Split tests into public and internal ones


sipa:
The current unit tests are all built against the library's internal source code directly. That means they have the advantage of being to test internal functions that aren't exposed from the library, but it also means the tests cannot actually run against the normal release build of the library.

To improve this:
* Split the tests into separate public API tests, and internal tests.
* A public test binary: public API tests, linked against the release build of the library (no -DVERIFY).
* An internal test binary: #include's the library code directly, plus public API tests, plus internal test (compiled with -DVERIFY).

real-or-random:
I also see some value in running the internal tests without -DVERIFY. This would exercise a lot of functions with their production source.

That doesn't mean that we'll get the same compiler output as in production (but probably the compiler will be closer to production at least). And it will catch cases in our source code, in which VERIFY code introduces real differences (which it should not).

So it's not perfect but tests are never perfect. I think this is a low hanging fruit, given that it should not be a lot of effort.

PR #1188, opened 2023-01-05
tests: Add noverify_tests which is like tests but without VERIFY

real-or-random:
mentioned in https://github.com/bitcoin-core/secp256k1/issues/1037#issuecomment-1371870423

Let's see how this affects CI time


real-or-random:
> Let's see how this affects CI time

Less than I expected:

master: https://cirrus-ci.com/build/6730654054875136
this PR: https://cirrus-ci.com/build/4723750860488704


What do you think? Is this useful? If yes, should we just keep it as is or reduce the test iterations?

real-or-random:
> I think we can leave it as-is. I think only the duration of the longest-running CI job really matters, and that seems barely affected.

Indeed. Ready for review then.

apoelstra (approved):
ACK 9953ce8d030d62f0e5ae4a4ae4f25a151d602992

real-or-random:
> Instead of disabling `-DVERIFY` for `tests` if coverage is enabled, maybe just disable `tests` (and leave `noverify_tests`)?

done

apoelstra (approved):
ACK 203760023c60d250cb5937e27bcf29e7a829096c

PR #1190, opened 2023-01-09
Make all non-API functions (except main) static

sipa:


real-or-random (commented):
Totally not opposed to this but do we really care about this in bench/tests? And if yes, do you see any way to ensure we don't miss functions (in the future)?

real-or-random (approved):
Okay, agreed on both counts. 

utACK e03ef8655933d3e2b4830e7f8fe86deba820a073

hebasto (approved):
ACK e03ef8655933d3e2b4830e7f8fe86deba820a073.

On my Ubuntu 22.04 (x86_64), the resulted test and bench binaries are a bit smaller in size, except for the `bench_internal`, which is a bit larger.

real-or-random:
(Fwiw, github indicated a merge conflict here (due to the renamed `src/cttime_tests.c`) but my local git auto-merged this cleanly without conflicts, and the merge results looks clean...)

PR #1192, opened 2023-01-10
Switch to exhaustive groups with small B coefficient

sipa:
This has the advantage that in the future, multiplication with B can be done using `secp256k1_fe_mul_int` rather than the slower `secp256k1_fe_mul`.


sage/gen_exhaustive_groups.sage
```diff
@@ -1,120 +1,121 @@
...
-    print(");")
+    print(f"static const secp256k1_ge secp256k1_ge_const_g = SECP256K1_G_ORDER_{f};")
+    output_b(b)
 print("#  else")
 print("#    error No known generator for the specified exhaustive test group order.")
 print("#  endif")
+print("#else")
+print("static const secp256k1_ge secp256k1_ge_const_g = SECP256K1_G;")
+output_b(7)
+print("#endif")
```
real-or-random, line 111:
```suggestion
print("#else")
print()
print("static const secp256k1_ge secp256k1_ge_const_g = SECP256K1_G;")
output_b(7)
print()
print("#endif")
```

I think this improves readability of the output a lot. Same for the other `#else`/`#elif` branches.

sage/gen_exhaustive_groups.sage
```diff
@@ -1,120 +1,121 @@
...
+def output_generator(g, name):
+    print(f"#define {name} SECP256K1_GE_CONST(\\")
+    print("    0x%08x, 0x%08x, 0x%08x, 0x%08x,\\" % tuple((int(g[0]) >> (32 * (7 - i))) & 0xffffffff for i in range(4)))
+    print("    0x%08x, 0x%08x, 0x%08x, 0x%08x,\\" % tuple((int(g[0]) >> (32 * (7 - i))) & 0xffffffff for i in range(4, 8)))
+    print("    0x%08x, 0x%08x, 0x%08x, 0x%08x,\\" % tuple((int(g[1]) >> (32 * (7 - i))) & 0xffffffff for i in range(4)))
+    print("    0x%08x, 0x%08x, 0x%08x, 0x%08x\\" % tuple((int(g[1]) >> (32 * (7 - i))) & 0xffffffff for i in range(4, 8)))
+    print(")")
+
+def output_b(b):
+    print(f"#define SECP256K1_B {int(b)}U")
```
real-or-random, line 75: It won't matter but is there a specific reason you put the `U`? `secp256k1_fe_mul_int` takes a signed int.

sage/gen_exhaustive_groups.sage
```diff
@@ -1,120 +1,121 @@
...
+
 print("")
 print("")
 print("/* To be put in src/group_impl.h: */")
```
real-or-random, line 79: Maybe add another line with `Generated by  ...`

real-or-random (commented):
Great :)

This should touch also touch the exhaustive tests code touched in https://github.com/bitcoin-core/secp256k1/commit/b110c106fa9704e30f6b0c2ffa6a2697031e89a8, at least the comments.

apoelstra (approved):
ACK 28cba925eb8d6b434604595fad92bd9e1da54913

sage/gen_exhaustive_groups.sage
```diff
@@ -1,120 +1,121 @@
...
+for f in sorted(gens.keys()):
+    # Use as generator/2 the one with lowest b, and lowest (x, y) generator (interpreted as integers).
```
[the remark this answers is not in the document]
real-or-random: nevermind,

apoelstra (approved):
ACK d4cf3f2da1e35118ede5b682082827f7d0393334

real-or-random:
When running the sage script locally, I noticed that it outputs a group of order 7, too. Did you omit this on purpose when copying the output of the script the C files?

Anyway, the exhaustive tests break don't compile for this order of 7. Here's a branch that fixes this problem and also has fixups (please squash) to your commits which include the 7 group (plus further formatting nits, what I actually had in mind here https://github.com/bitcoin-core/secp256k1/pull/1192#discussion_r1066739496): https://github.com/real-or-random/secp256k1/commits/202301_small_b

This comment of mine is still unaddressed:

> This should touch also touch the exhaustive tests code touched in [b110c10](https://github.com/bitcoin-core/secp256k1/commit/b110c106fa9704e30f6b0c2ffa6a2697031e89a8), at least the comments.


Sorry, I didn't foresee that my comment about using `mul_int` in #1118 gets us sidetracked so much... ^^

real-or-random (approved):
ACK ce60785b2654e60b43577dd75996b7020afbfec8

I still think that https://github.com/real-or-random/secp256k1/commit/7eef8f4a305a72567db0b563260081712304b5fb is a good cleanup but we can also add it after this PR

apoelstra (approved):
ACK ce60785b2654e60b43577dd75996b7020afbfec8

real-or-random:
For our future curious minds:

```
00:44 <sipa> Let G1 and G2 be generators of the two endomorphism-compatible order-7 subgroups.
00:46 <sipa> Let C_i be the subgroup consisting of itG1 + tG2 for t=0..6 (with i=infinity for the group consisting of tG1).
00:47 <sipa> Then beta*C_i lands you in C_{2i}.
00:47 <sipa> If i=0 or i=inf, it's an endomorphism.
01:20 <sipa> Ah! And there is a much simpler explanation.
01:21 <sipa> With the same two generators G1,G2 above, beta-multiplying a*G1 + b*G2 gives 2*a*G1 + 4*b*G2.
01:21 <sipa> If you started off with a*G1, you get twice your input, so lambda=2.
01:22 <sipa> If you started off with b*G2, you get 4 times your input, so lambda=4.
01:22 <sipa> But if you started off with a linear combination with non-zero contributions from both, you get a different linear combination that's not a pure multiple back.
01:24 <sipa> Put otherwise, for any independent generators G1,G2 of the order-49 subgroup, if you think of the elements as vectors with basis G1,G2, then beta-multiplication is effectively a linear transformation on those vectors.
01:24 <sipa> That linear transformation has eigenvalues 2 and 4.
01:25 <sipa> The subgroup consisting of eigenvectors corresponding to eigenvalue 2 is the subgroup in which the endomorphism works, with lambda=2.
01:25 <sipa> Same with eigenvalue 4.
```

https://gnusha.org/secp256k1/2023-01-16.log

PR #1194, opened 2023-01-17
Ensure safety of ctz_debruijn implementation.

roconnor-blockstream:
Adding `U` to the magic constants ensures that we are not mixing unsigned and signed value during multiplication, and ensures that the multiplication will not be subject to integer promotion.

The `(uint32_t)`/`(uint64_t)` casts ensure the values are properly truncated no matter the size of an int.

Prior to this commit, if `secp256k1_ctz32_var_debruijn` were some how managed to be built on a platform with 64-bit ints, (though this function is specifically only intended to be used on 32-bit platforms) it would perform an out-of-bounds array access.


real-or-random (approved):
utACK d6ff738d5bbdf965590fc07efca23b13c0ea3082

> a platform with 64-bit ints

be aware that our code base will most likely explode in a lot of ways on such a platform

apoelstra (approved):
ACK d6ff738d5bbdf965590fc07efca23b13c0ea3082

PR #1195, opened 2023-01-18
ctime_tests: improve output when CHECKMEM_RUNNING is not defined

jonasnick:
When seeing the output
```
Unless compiled under msan, this test can only usefully be run inside valgrind.
```
I thought that I would have to go back to the `configure` output to manually check if it was compiled under memsan to determine whether this test can be usefully run outside valgrind. But when we go into this branch then it was definitely not compiled under msan, which means that we can make the output clearer.

### DIFF
--- a/src/ctime_tests.c
+++ b/src/ctime_tests.c
@@ -38,7 +38,7 @@ int main(void) {
     int ret, i;
 
     if (!SECP256K1_CHECKMEM_RUNNING()) {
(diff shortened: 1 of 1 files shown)


real-or-random (approved):
utACK https://github.com/bitcoin-core/secp256k1/pull/1195/commits/8f51229e0348a1524fed541f334cd4f1726d2685

PR #1196, opened 2023-01-19
Drop no longer used variables from the build system

hebasto:
The last usage of the `SECP_TEST_{LIBS,INCLUDE}` variables was removed in https://github.com/bitcoin-core/secp256k1/pull/983.

The last usage of the `SECP_INCLUDE` variable was removed in https://github.com/bitcoin-core/secp256k1/pull/1169.


real-or-random (approved):
utACK 2cd4e3c0a976c78c1619fc7456fcc4eaa92897a9

PR #1203, opened 2023-01-30
Do not link `bench` and `ctime_tests` to `COMMON_LIB`

hebasto:
The `bench` and `ctime_tests` binaries are users of the library, they should only be linked to the library, not the objects it was built from.


real-or-random (approved):
utACK ef39721cccec344983f09180bcf9c443d491f7cb

PR #1204, opened 2023-01-31
[POC] cmake: Add option to run `include-what-you-use` with compiler

hebasto:
[IWYU](https://include-what-you-use.org/) could be useful for developers, e.g., #924 etc.

Example of usage on Ubuntu 22.04:
```
$ cmake -S . -B build -DSECP256K1_ENABLE_IWYU=ON -DCMAKE_C_COMPILER=clang
$ cmake --build build
```


real-or-random:
Oh, this may indeed be helpful. I haven't tested anything yet, but if this helps towards https://github.com/bitcoin-core/secp256k1/issues/1039, I'm interested.

hebasto:
Rebased on top of the #1113.

real-or-random:
I fear `include-what-you-use` is too dump for our file organization and won't help towards making files self-contained #924.

For example, it suggests:
```
The full include-list for /home/tim/bs/dev/secp256k1/src/secp256k1.c:
(...)
#include "field_5x52_impl.h"                    // for secp256k1_fe_get_b32
(...)
```

We don't want to include `field_5x52_impl.h`  because we include `field_impl.h` (which handles selection of the right full file)


Also, it suggests includes only for .c files and not for .h files... If we want to have self-contained files, it may be much better to simply check that every file complies individually as suggested in #924.

hebasto:
> Also, it suggests includes only for .c files and not for .h files...

For example:
```
/home/hebasto/git/secp256k1/src/precomputed_ecmult.h should add these lines:
#include "ecmult.h"  // for ECMULT_TABLE_SIZE
```

hebasto:
> I fear `include-what-you-use` is too dump for our file organization...

Sure, it is not a panacea. However, it can be useful. For instance, consider the [demo](https://github.com/hebasto/secp256k1/commits/230309-iwyu.DEMO) branch (a mix with #1231):
```
$ cmake -S . -B build -DSECP256K1_ENABLE_IWYU=ON -DCMAKE_C_COMPILER=clang
$ cmake --build build --target precomputed  # no IWYU diagnostic
```

fanquake:
Concept ~0. Cores experience with IWYU has been mixed. We've found plenty of bugs/nonsensical include suggestions, and it's not ideal if by default you have to add comments all over the code. We should probably look at migrating to something else, maybe LLVMs `clang-include-fixer` tool.

Generally I'm not sure that "convenience option for running an include fixer" is something that needs to exists in libsecp256k1s build system. Running `include-what-you-use` yourself is pretty straight forward.

hebasto:
> Running `include-what-you-use` yourself is pretty straight forward.

I agree with that. For example,
```
cmake -S . -B build -DCMAKE_C_COMPILER=clang -DCMAKE_C_INCLUDE_WHAT_YOU_USE=$(command -v include-what-you-use)
cmake --build build
```

Closing.

PR #1205, opened 2023-02-01
field: Improve docs +tests of secp256k1_fe_set_b32

real-or-random:
We could also change the interface to accept an explicit `overflow` output variable, like for the scalar function. I think this makes handling (and ignoring!) overflow much more explicit in the code. What do you think?

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -44,6 +44,25 @@ static int all_bytes_equal(const void* s, unsigned char value, size_t n) {
     return 1;
 }
 
+/* Debug helper for printing arrays of unsigned char. */
+#define PRINT_BUF(buf, len) do { \
+    printf("%s[%lu] = ", #buf, (unsigned long)len); \
+    print_buf_plain(buf, len); \
+} while(0);
+static void print_buf_plain(const unsigned char *buf, size_t len) {
+    size_t i;
+    printf("{");
+    for (i = 0; i < len; i++) {
+        if (i % 8 == 0) {
+            printf("\n    ");
+        } else {
+            printf(" ");
+        }
+        printf("0x%02X,", buf[i]);
+    }
+    printf("\n}\n");
+}
+
 /* TODO Use CHECK_ILLEGAL(_VOID) everywhere and get rid of the uncounting callback */
 /* CHECK that expr_or_stmt calls the illegal callback of ctx exactly once
  *
--- a/src/tests.c
+++ b/src/tests.c
@@ -2961,6 +2961,69 @@ static void run_field_convert(void) {
     CHECK(secp256k1_memcmp_var(&fes2, &fes, sizeof(fes)) == 0);
 }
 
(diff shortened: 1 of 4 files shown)


real-or-random:
> Concept ACK, though is the inclusion of the last commit intentional?

Yes, but I'm happy to remove it, of course. The story behind this is that I had forgotten to remove a debug printer first when I opened this PR. I mentioned this to @jonasnick who told me that he also had to write these functions multiple times in the past, so we thought it's a good idea to have one in the code base. Then I made it a bit nicer and added a commit. And I felt it's not worth a separate PR (but maybe I was just being lazy there).

real-or-random:
Ok yes, understood now.

src/tests.c
```diff
@@ -2961,6 +2976,74 @@ static void run_field_convert(void) {
...
+            0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+            0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,
+            0xFF, 0xFF, 0xFF, 0xFE, 0xFF, 0xFF, 0xFC, 0x2F,
+        };
+        static const unsigned char zero[32] = {
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+        };
```
jonasnick, line 2992: nit: fewer loc, easier to check
```suggestion
        static const unsigned char zero[32] = { 0x00 };
```

src/tests.c
```diff
@@ -44,6 +44,21 @@ static int all_bytes_equal(const void* s, unsigned char value, size_t n) {
     return 1;
 }
 
+/* Debug helper for printing arrays of unsigned char. */
+static void print_buf(const unsigned char *buf, size_t len) {
+    size_t i;
+    printf("{");
```
jonasnick, line 50: This would be extra useful if this helper would also output the name of the printed variable. If `print_buf` was a  macro we could do it like this:
```suggestion
    printf("%s = {", #buf);
```

jonasnick, line 50: This would be extra useful if this helper would also output the name of the printed variable. If `print_buf` was a  macro we could do it like this:
```suggestion
    printf("%s = {", #buf);
```

jonasnick (commented):
ACK mod nits

real-or-random:
force-pushed to address @jonasnick's nits

jonasnick (commented):
ACK 162da73e9a48875aab1ee6ca1c14f86ca4646946

PR #1206, opened 2023-02-02
build: Add -Wreserved-identifier supported by clang

real-or-random:
This warns on certain identifiers reserved by the C standard, namely
 * identifiers that begin with an underscore followed by an uppercase letter, and
 * identifiers in the global namespace that begin with an underscore.

We had used such identifiers in the past for macros in include guards, and we should make sure that we don't reintroduce such identifiers going forward.

Note that C reserves more identifiers for "future library directions", e.g., identifiers that begin with "str" followed by a lowercase letter. But even the C standards committee has decided that this is somewhat silly and adopted a proposal [1] for C23 that removes the restriction that programs using these identifiers have UB. Instead, these identifiers are now "potentially reserved", which is not a normative restriction but simply an informative warning that the identifiers may become fully reserved in the future.

[1] https://www.open-std.org/jtc1/sc22/WG14/www/docs/n2625.pdf


jonasnick (commented):
tested ACK 8c7e0fc1de048be98a1f1bc75557671afc14beaa

PR #1207, opened 2023-02-05
Split fe_set_b32 into reducing and normalizing variants

sipa:
Follow-up to #1205.

This splits the `secp256k1_fe_set_b32` function into two variants:
* `secp256k1_fe_set_b32_mod`, which returns `void`, reduces modulo the curve order, and only promises weakly normalized output.
* `secp256k1_fe_set_b32_limit`, which returns `int` indicating success/failure, and only promises valid output in case the input is in range (but guarantees it's strongly normalized in this case).

This removes one of the few cases in the codebase where normalization status depends on runtime values, making it fixed at compile-time instead.

### DIFF
--- a/src/field.h
+++ b/src/field.h
@@ -85,7 +85,8 @@ static const secp256k1_fe secp256k1_const_beta = SECP256K1_FE_CONST(
 #  define secp256k1_fe_is_zero secp256k1_fe_impl_is_zero
 #  define secp256k1_fe_is_odd secp256k1_fe_impl_is_odd
 #  define secp256k1_fe_cmp_var secp256k1_fe_impl_cmp_var
-#  define secp256k1_fe_set_b32 secp256k1_fe_impl_set_b32
+#  define secp256k1_fe_set_b32_mod secp256k1_fe_impl_set_b32_mod
+#  define secp256k1_fe_set_b32_limit secp256k1_fe_impl_set_b32_limit
 #  define secp256k1_fe_get_b32 secp256k1_fe_impl_get_b32
 #  define secp256k1_fe_negate secp256k1_fe_impl_negate
 #  define secp256k1_fe_mul_int secp256k1_fe_impl_mul_int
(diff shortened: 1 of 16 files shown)


src/field_10x26_impl.h
```diff
@@ -364,12 +363,29 @@ static int secp256k1_fe_set_b32(secp256k1_fe *r, const unsigned char *a) {
...
+    ret = !((r->n[9] == 0x3FFFFFUL) & ((r->n[8] & r->n[7] & r->n[6] & r->n[5] & r->n[4] & r->n[3] & r->n[2]) == 0x3FFFFFFUL) & ((r->n[1] + 0x40UL + ((r->n[0] + 0x3D1UL) >> 26)) > 0x3FFFFFFUL));
+
+#ifdef VERIFY
+    if (ret) {
+        r->magnitude = 1;
+        r->normalized = 1;
+        secp256k1_fe_verify(r);
+    } else {
+        r->n[4] ^= 1; /* change value, hopefully triggering errors if the value is used. */
+    }
```
real-or-random, line 386: Perhaps we should instead make sure that we call `secp256k1_fe_verify` on every entry of a function that reads a `fe`? We do it in some functions but not in all. For instance, t's missing in `secp256k1_fe_mul_int`.

src/field.h
```diff
@@ -75,10 +75,14 @@ static int secp256k1_fe_equal_var(const secp256k1_fe *a, const secp256k1_fe *b);
 /** Compare two field elements. Requires both inputs to be normalized */
 static int secp256k1_fe_cmp_var(const secp256k1_fe *a, const secp256k1_fe *b);
 
-/** Set a field element equal to 32-byte big endian value.
- *  Returns 1 if no overflow occurred, and then the output is normalized.
- *  Returns 0 if overflow occurred, and then the output is only weakly normalized. */
-static int secp256k1_fe_set_b32(secp256k1_fe *r, const unsigned char *a);
+/** Set a field element equal to 32-byte big endian value, reducing modulo the curve
+ *  order if need be. The output is weakly normalized. */
```
real-or-random, line 79: nit: s/if need be/if necessary is simpler English

real-or-random (commented):
Concept ACK

Not sure what's wrong with CI. Some tasks show red (exit code 134) though all tests have passed. I restarted those.

real-or-random:
> Not sure what's wrong with CI. Some tasks show red (exit code 134) though all tests have passed. I restarted those.

Oh, these are the new `-DVERIFY` tests, so they fail in the benchmark. It's hard to see from the CI output. We should perhabs reconsider that `cat` to `.log` thing.

real-or-random:
A variant of the commit has been merged as part of https://github.com/bitcoin-core/secp256k1/pull/1205, so it should be removed.

real-or-random (approved):
ACK 47bfc0a934c36e9fe1446f583789de6399bdd81d

src/field.h
```diff
@@ -189,16 +190,20 @@ static int secp256k1_fe_equal_var(const secp256k1_fe *a, const secp256k1_fe *b);
...
+ */
+static void secp256k1_fe_set_b32_mod(secp256k1_fe *r, const unsigned char *a);
+
+/** Set a field element equal to a provided 32-byte big endian value.
  *
- * Note that this function is unusual in that the normalization of the output depends on the
- * run-time value of a.
+ * On input, r does not need to be initalized. a must be a pointer to an initialized 32-byte array.
+ * On output, r = a if (a < p), it will be normalized with magnitude 1, and 1 is returned.
+ * If a >= 0, 0 is returned, and r will be made invalid (and cannot be used without overwriting).
```
jonasnick, line 204: Should be "if a >= p"?

jonasnick (commented):
Concept ACK

jonasnick (approved):
ACK 5b32602295ff7ad9e1973f96b8ee8344b82f4af0

Issue #1181, opened 2022-12-19
Symbol visibility


real-or-random:
In certain cases, we export too many symbols:
 - [ ] With `CFLAGS=-fvisibility=default`, there's a few exported _variables_ : `secp256k1_ecmult_gen_prec_table`, `secp256k1_pre_g`, `secp256k1_pre_g_128`
 - [x] on ARM, all the functions in the `.s` assembly files are visible but none should be visible (solved by https://github.com/bitcoin-core/secp256k1/pull/1242)
 - [x] Should we make all test functions static? (solved by https://github.com/bitcoin-core/secp256k1/pull/1190)
 - [ ] How to ensure correct visibility going forward (e.g., #1135)

real-or-random:
> * With `CFLAGS=-fvisibility=default`, there's a few exported _variables_ : `secp256k1_ecmult_gen_prec_table`, `secp256k1_pre_g`, `secp256k1_pre_g_128`

The reason is that these three variables are not in the main translation unit, in which they're merely declared `extern`. (We cannot simply add `static`, because that would imply internal linkage which conflicts which the external linkage that we need due to the use of multiple compilation units.)  

There's no generic C language feature to fix this. But we can fix this in both GCC and clang by any of the following (in descending order of my personal preference).
1. Add `__attribute__ ((visibility ("hidden")))` to the `extern` declarations
2. Wrap the `extern` declarations in `#pragma GCC visibility push(hidden)` ... `#pragma GCC visibility pop`. 
    1. We could in fact do this with the entire code base (excluding where we include foreign headers). :shrug: Not entirely sure if it doesn't create further problems...
3. Add `-Wl,--exclude-libs,libsecp256k1_precomputed.a` to `LDFLAGS` but that's kind of pointless to cover the case of default visibility, which typically occurs when the user invokes the compiler manually without even caring about `-fvisibility`.

In any case, `-fvisibility=hidden` would not be strictly necessary on the command line anymore (except for rather exotic compilers such as `icc` which understand it but not the attribute or the `#pragma`. This would then help [manual builds](https://github.com/bitcoin-core/secp256k1/issues/929).

hebasto:
>    * [ ]  on ARM, all the functions in the `.s` assembly files are visible but none should be visible

Solved by @theuni in #1242.

theuni:
[...]

PR #1209, opened 2023-02-06
build: Add SECP256K1_API_VAR to fix importing variables from DLLs

real-or-random:
... and more Windows fixes, please see the individual commits.

The fixed issues were discovered in https://github.com/bitcoin-core/secp256k1/pull/1198.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -115,6 +115,12 @@ AC_DEFUN([SECP_TRY_APPEND_DEFAULT_CFLAGS], [
     if test x"$GCC" != x"yes" && test x"$build_windows" = x"yes"; then
       SECP_TRY_APPEND_CFLAGS([-W2 -wd4146], $1) # Moderate warning level, disable warning C4146 "unary minus operator applied to unsigned type, result still unsigned"
       SECP_TRY_APPEND_CFLAGS([-external:anglebrackets -external:W0], $1) # Suppress warnings from #include <...> files
+      # We pass -ignore:4217 to the MSVC linker to suppress warning 4217 when
+      # importing variables from a statically linked secp256k1.
+      # (See the libtool manual, section "Windows DLLs" for background.)
+      # Unfortunately, libtool tries to be too clever and strips "-Xlinker arg"
+      # into "arg", so this will be " -Xlinker -ignore:4217" after stripping.
+      LDFLAGS="-Xlinker -Xlinker -Xlinker -ignore:4217 $LDFLAGS" 
     fi
 ])
 SECP_TRY_APPEND_DEFAULT_CFLAGS(SECP_CFLAGS)
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -258,7 +258,7 @@ task:
     # Set non-essential options that affect the CLI messages here.
     # (They depend on the user's taste, so we don't want to set them automatically in configure.ac.)
     CFLAGS: -nologo -diagnostics:caret
-    LDFLAGS: -XCClinker -nologo -XCClinker -diagnostics:caret
+    LDFLAGS: -Xlinker -Xlinker -Xlinker -nologo
   matrix:
     - name: "x86_64 (MSVC): Windows (Debian stable, Wine)"
     - name: "x86_64 (MSVC): Windows (Debian stable, Wine, int128_struct)"
--- a/ci/cirrus.sh
+++ b/ci/cirrus.sh
@@ -34,6 +34,8 @@ print_environment
 # This speeds up jobs with many invocations of wine (e.g., ./configure with MSVC) tremendously.
 case "$WRAPPER_CMD" in
     *wine*)
+        # Make sure to shutdown wineserver whenever we exit.
+        trap "wineserver -k || true" EXIT INT HUP
         # This is apparently only reliable when we run a dummy command such as "hh.exe" afterwards.
         wineserver -p && wine hh.exe
         ;;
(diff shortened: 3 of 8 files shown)


real-or-random:
Note that users will have the same issue that they see linker warning LNK4217 when they try to use MSVC to link statically against libsecp256k1 (but better a warning than a failing build.)

I think we should note that somewhere and recommend ignoring that warning by passing `-ignore:4217' (but where should we note that?). We could alternatively offer the possibility suppress the dllimport() stuff when a macro `SECP256K1_STATIC_LINKAGE` is defined, but I don't think that's better than just documenting the issue. (The macro will also need documentation.)

configure.ac
```diff
@@ -115,6 +115,12 @@ AC_DEFUN([SECP_TRY_APPEND_DEFAULT_CFLAGS], [
     if test x"$GCC" != x"yes" && test x"$build_windows" = x"yes"; then
       SECP_TRY_APPEND_CFLAGS([-W2 -wd4146], $1) # Moderate warning level, disable warning C4146 "unary minus operator applied to unsigned type, result still unsigned"
       SECP_TRY_APPEND_CFLAGS([-external:anglebrackets -external:W0], $1) # Suppress warnings from #include <...> files
+      # We pass -ignore:4217 to the MSVC linker to suppress warning 4217 when
+      # importing variables from a statically linked secp256k1.
+      # (See the libtool manual, section "Windows DLLs" for background.)
+      # Unfortunately, libtool tries to be too clever and strips "-Xlinker arg"
```
hebasto, line 121: Libtool vs CMake -- 0:1 :smile:

hebasto (approved):
ACK e4330341bd648e93b60fe70c631e311a98bce549, tested on Windows using [CMake](https://github.com/bitcoin-core/secp256k1/pull/1113) (which means that the 3rd commit is reviewed only, but not tested). FWIW, `LNK4217` warnings have been indeed observed.

configure.ac
hebasto: This patch has been [mirrored](https://github.com/bitcoin-core/secp256k1/pull/1113#issuecomment-1439689959) in the [CMake-based](https://github.com/bitcoin-core/secp256k1/pull/1113) build system.

hebasto:
Sorry for touching this PR after 2 weeks, but I need to clarify a few things, at least for myself.

After @theuni's #1105, when cross-compiling a library for Windows, the `SECP256K1_API` macro is defined to `__declspec(dllexport)` because Libtool uses ` -DDLL_EXPORT `.

After this PR, that is no longer the case.

Are we skipping `__declspec(dllexport)` for cross-compiled DLL altogether?

real-or-random:
> After @theuni's #1105, when cross-compiling a library for Windows, the `SECP256K1_API` macro is defined to `__declspec(dllexport)` because Libtool uses `-DDLL_EXPORT`.
> 
> After this PR, that is no longer the case.


I can't follow. What's the exact combination of defines under which this PR has changed something and which bothers you?

hebasto:
> I can't follow. What's the exact combination of defines under which this PR has changed something and which bothers you?

Nothing special. Just `./configure --host=x86_64-w64-mingw32 --disable-static`. The `SECP256K1_API` macro is defined as follows:
- before this PR:
```c
#   define SECP256K1_API __declspec(dllexport)
```

- after this PR:
```c
#  define SECP256K1_API             __attribute__ ((visibility ("default")))
```

real-or-random:
Ok, I see.

> Are we skipping `__declspec(dllexport)` for cross-compiled DLL altogether?

Yes, at least when using MinGW. (I assume you call it cross-compiling when GNU tools are used to build a DLL?) As I understand https://www.gnu.org/software/libtool/manual/libtool.html#Windows-DLLs, this should work. I guess MinGW counts as GNU tool. I simply took the code from there because they seem to have thought that through.

hebasto:
> Yes, at least when using MinGW. (I assume you call it cross-compiling when GNU tools are used to build a DLL?)

Correct, I'm using MinGW.

> As I understand https://www.gnu.org/software/libtool/manual/libtool.html#Windows-DLLs, this should work.

Apparently, it works :)

hebasto:
Thanks to @theuni it is clear now that this PR did [introduce](https://github.com/bitcoin-core/secp256k1/pull/1209#issuecomment-1464946813) a regression that listed in https://github.com/bitcoin-core/secp256k1/issues/1181:
>    * [ ]  With `CFLAGS=-fvisibility=default`, there's a few exported _variables_ : `secp256k1_ecmult_gen_prec_table`, `secp256k1_pre_g`, `secp256k1_pre_g_128`

I've verified that @rheuni's [suggestion](https://github.com/bitcoin-core/secp256k1/issues/1181#issuecomment-1472496663) fixes this regression.

hebasto:
Sorry for the late question :)

I'm curios about rationale of putting the `extern` keyword into the macro?

real-or-random:
> I'm curios about rationale of putting the `extern` keyword into the macro?

Ha, the simple answer is that I took this from the libtool manual and it just worked.

If you're asking why we need `extern` at all, then this is a bit ugly in C: The rules are different for functions are variables: https://port70.net/~nsz/c/c11/n1570.html#6.2.2p5 In my own words: `extern` is always the default for functions, so you don't have to write it. For variables, `extern` is only the default if they appear in "file scope"/top level.  

So if some user code had `#include <secp256k1.h>` within a function (for whatever reason), our variable declarations wouldn't work without `extern`. (See https://gcc.godbolt.org/z/9h1qqEjT7, also try enabling linking and look at the linker errors).

hebasto:
> > I'm curios about rationale of putting the `extern` keyword into the macro?
> 
> Ha, the simple answer is that I took this from the libtool manual and it just worked.
> 
> If you're asking why we need `extern` at all...

My point is if we keep `extern` in their place in the source code, then the `SECP256K1_API_VAR` macro becomes equal to `SECP256K1_API`.

PR #1210, opened 2023-02-07
Convert type of eta to int

roconnor-blockstream:
Almost everywhere eta is cast to or otherwise treated as an int.

I suspect that the existing `int64_t` treatment of eta is a carryover from the constant time code where the specific type is perhaps more important.


roconnor-blockstream:
Oh well.

PR #1212, opened 2023-02-17
Prevent dead-store elimination when clearing secrets in examples

Harshil-Jani:
Signed-off-by: Harshil Jani <harshiljani2002@gmail.com>

### DIFF
--- a/examples/schnorr.c
+++ b/examples/schnorr.c
@@ -15,7 +15,7 @@
 #include <secp256k1_extrakeys.h>
 #include <secp256k1_schnorrsig.h>
 
-#include "random.h"
+#include "examples_util.h"
 
 int main(void) {
     unsigned char msg[12] = "Hello World!";
@@ -140,9 +140,8 @@ int main(void) {
      * example through "out of bounds" array access (see Heartbleed), Or the OS
      * swapping them to disk. Hence, we overwrite the secret key buffer with zeros.
      *
-     * TODO: Prevent these writes from being optimized out, as any good compiler
+     * Here we are preventing these writes from being optimized out, as any good compiler
      * will remove any writes that aren't used. */
-    memset(seckey, 0, sizeof(seckey));
-
+    secure_erase(seckey, sizeof(seckey));
     return 0;
 }
--- a/Makefile.am
+++ b/Makefile.am
@@ -69,7 +69,7 @@ noinst_HEADERS += contrib/lax_der_parsing.h
 noinst_HEADERS += contrib/lax_der_parsing.c
 noinst_HEADERS += contrib/lax_der_privatekey_parsing.h
 noinst_HEADERS += contrib/lax_der_privatekey_parsing.c
-noinst_HEADERS += examples/random.h
+noinst_HEADERS += examples/examples_util.h
 
 PRECOMPUTED_LIB = libsecp256k1_precomputed.la
 noinst_LTLIBRARIES = $(PRECOMPUTED_LIB)
--- a/examples/ecdh.c
+++ b/examples/ecdh.c
@@ -14,8 +14,7 @@
 #include <secp256k1.h>
 #include <secp256k1_ecdh.h>
 
-#include "random.h"
-
+#include "examples_util.h"
 
 int main(void) {
     unsigned char seckey1[32];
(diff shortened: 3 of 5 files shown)


real-or-random:
It's probably a good idea to add this to the examples, but I think an approach that has a separate function like in https://github.com/bitcoin-core/secp256k1/pull/636/commits/494965880b47c3f2b752391b0c757614d6cb9f75 is cleaner. `secure_erase` sounds like a good function name, for example.

alokeutpal:
Please help me a little. I want to get it in phantom wallet. One more thing
I want to know is my address in my phantom wallet where can I get it?

alokeutpal:
Really I am non technical person only understand mobile off on help me

Harshil-Jani:
> It's probably a good idea to add this to the examples, but I think an approach that has a separate function like in https://github.com/bitcoin-core/secp256k1/commit/494965880b47c3f2b752391b0c757614d6cb9f75 is cleaner. secure_erase sounds like a good function name, for example.

Thanks @real-or-random  for this !! Will look at this soon.

> Please help me a little. I want to get it in phantom wallet. One more thing
I want to know is my address in my phantom wallet where can I get it?

Sorry ?? I guess, This would be a wrong place for the question.

Harshil-Jani:
@real-or-random Let me know your feedback on this. Also, I was wondering is there any header where we can club this ? or how about a seperate header for `secure_erase` instead of writing it for each file.

real-or-random:
Yes, looks better now. Now if we include such a function, I still feel we should recommend state-of-the art practices, which is what I implemented in https://github.com/bitcoin-core/secp256k1/commit/494965880b47c3f2b752391b0c757614d6cb9f75. Note that essentially the same code was added to Bitcoin Core, see https://github.com/bitcoin-core/secp256k1/commit/494965880b47c3f2b752391b0c757614d6cb9f75.

> Also, I was wondering is there any header where we can club this ? or how about a seperate header for `secure_erase` instead of writing it for each file.

I think we could put it in `random.h` and rename this to `util.h` or `examples_util.h` to better distinguish it from our own `util.h`.

Harshil-Jani:
@real-or-random I tried renaming the file. But, I think that would need to change few other things from `secp256k1.h` which could be part of another patch.

real-or-random:
> @real-or-random I tried renaming the file. But, I think that would need to change few other things from `secp256k1.h` which could be part of another patch.

Why do you think so? This should be unrelated?

Harshil-Jani:
> > @real-or-random I tried renaming the file. But, I think that would need to change few other things from `secp256k1.h` which could be part of another patch.
> 
> Why do you think so? This should be unrelated?

My bad. I just renamed it and `SECP256K1_INLINE`  instantly throwed an error of `explicit type is missing ('int' assumed)` (And I thought how would renaming a header affect the same file altough I knew we need to change it's name in other files as well)so I assumed we need to fix this from secp.h But on further inspection I got to know that after renaming the header wherever it was used, It can find the inline and things would fix up.

real-or-random:
Awesome. Can you also squash the first three commits together?

Harshil-Jani:
> Awesome. Can you also squash the first three commits together?

Done !! Thanks for reviewing.

examples/schnorr.c
```diff
@@ -140,9 +140,8 @@ int main(void) {
      * example through "out of bounds" array access (see Heartbleed), Or the OS
      * swapping them to disk. Hence, we overwrite the secret key buffer with zeros.
      *
-     * TODO: Prevent these writes from being optimized out, as any good compiler
+     * Here we are preventing these writes from being optimized out, as any good compiler
      * will remove any writes that aren't used. */
-    memset(seckey, 0, sizeof(seckey));
-
+    secure_erase(seckey,sizeof(seckey));
```
real-or-random, line 145:
```suggestion
    secure_erase(seckey, sizeof(seckey));
```

And the same for the other calls to `secure_erase`.

real-or-random (commented):
ACK with the comment addressed

That was really the last comment I have.

examples/schnorr.c
Harshil-Jani: > That was really the last comment I have.

Oh that's totally fine. Even I am sceptical about the coding standards for established projects :) 
Done. Thanks for suggesting!!

real-or-random (approved):
utACK 5660c137552c657da5265691dea0fb10faae6a76

PR #1215, opened 2023-02-27
Use arc4random_buf() as the best practice method for obtaining randomness on OpenBSD (examples)

alpn:
The method above is suggested instead of the current `getentropy()`, which is not intended to be used by user code. 
See OpenBSD's [man page](https://man.openbsd.org/getentropy.2):

> [getentropy](https://man.openbsd.org/getentropy.2#getentropy~2)() is not intended for regular code; use the [arc4random(3)](https://man.openbsd.org/arc4random.3) family of functions instead.

### DIFF
--- a/examples/random.h
+++ b/examples/random.h
@@ -12,7 +12,7 @@
  * Linux   -> `getrandom(2)`(`sys/random.h`), if not available `/dev/urandom` should be used. http://man7.org/linux/man-pages/man2/getrandom.2.html, https://linux.die.net/man/4/urandom
  * macOS   -> `getentropy(2)`(`sys/random.h`), if not available `/dev/urandom` should be used. https://www.unix.com/man-page/mojave/2/getentropy, https://opensource.apple.com/source/xnu/xnu-517.12.7/bsd/man/man4/random.4.auto.html
  * FreeBSD -> `getrandom(2)`(`sys/random.h`), if not available `kern.arandom` should be used. https://www.freebsd.org/cgi/man.cgi?query=getrandom, https://www.freebsd.org/cgi/man.cgi?query=random&sektion=4
- * OpenBSD -> `getentropy(2)`(`unistd.h`), if not available `/dev/urandom` should be used. https://man.openbsd.org/getentropy, https://man.openbsd.org/urandom
+ * OpenBSD -> `arc4random_buf(3)`(`stdlib.h`), if not available `/dev/urandom` should be used. https://man.openbsd.org/arc4random.3, https://man.openbsd.org/urandom
  * Windows -> `BCryptGenRandom`(`bcrypt.h`). https://docs.microsoft.com/en-us/windows/win32/api/bcrypt/nf-bcrypt-bcryptgenrandom
  */
 
@@ -23,7 +23,7 @@
 #elif defined(__linux__) || defined(__APPLE__) || defined(__FreeBSD__)
 #include <sys/random.h>
 #elif defined(__OpenBSD__)
-#include <unistd.h>
+#include <stdlib.h>
 #else
 #error "Couldn't identify the OS"
 #endif
@@ -50,7 +50,7 @@ static int fill_random(unsigned char* data, size_t size) {
     } else {
         return 1;
     }
-#elif defined(__APPLE__) || defined(__OpenBSD__)
+#elif defined(__APPLE__)
     /* If `getentropy(2)` is not available you should fallback to either
      * `SecRandomCopyBytes` or /dev/urandom */
     int res = getentropy(data, size);
@@ -59,6 +59,11 @@ static int fill_random(unsigned char* data, size_t size) {
     } else {
         return 0;
     }
+#elif defined(__OpenBSD__)
+    /* This function is always successful, and no return value is reserved to
+     indicate an error.  */
+    arc4random_buf(data, size);
+    return 1;
 #endif
     return 0;
 }


real-or-random:
`getentropy()` was chosen deliberately over `arc4random()` when this was introduced, see  https://github.com/bitcoin-core/secp256k1/pull/748#discussion_r603295709 for the discussion.

> `getentropy()`, which is not intended to be used by user code. See OpenBSD's [man page](https://man.openbsd.org/getentropy.2):
> 
> > [getentropy](https://man.openbsd.org/getentropy.2#getentropy~2)() is not intended for regular code; use the [arc4random(3)](https://man.openbsd.org/arc4random.3) family of functions instead.

Hm, I don't think "regular code" means "user code". This is a syscall after all, if it is not supposed to be called by user code who else should call it?
 
> [getentropy](https://man.openbsd.org/getentropy.2#getentropy)() fills a buffer with high-quality entropy, which can be used as input for process-context pseudorandom generators like [arc4random(3)](https://man.openbsd.org/arc4random.3).

Together with the above quote, I read this as "Use `getentropy()` for seeding PRGs because it's high-quality, but if you want PRG output, then use `arc4random()`. But this distinction is a bit arbitrary, to be honest. And it's wrong because it contradicts the implementation of `getentropy()`, which has [always used `arc4random()`](https://github.com/openbsd/src/commit/eacbc5b32cfe3ca4de47833fa1b4b0efc9b3080a).

alpn:
Hey Tim, thank you for the links and the quick response.

OpenBSD actually has two functions named `arc4random_buf()`.
The one you linked to is the kernel side one, which as you saw is used in the implementation of the `getentropy()` syscall.
The one in this PR (`arc4random_buf(3)`) is part of `libc`, and somewhat confusingly, is itself implemented using that same `getentropy()` syscall, but additionally applies some `ChaCha` based per-process stirring:

https://github.com/openbsd/src/blob/7c9f822481c4d6823c0a1b1741f55500f10f1341/lib/libc/crypt/arc4random.c#L198

The linked discussion mentions that **OpenSSL** uses `getentropy()`, but [**LibreSSL**](https://www.libressl.org), OpenBSD’s fork thereof, is using `arc4random_buf()` to generate its private keys, e.g:

https://github.com/libressl/openbsd/blob/c5e95c4ee6991643e587f9e4b8f23c7dbb5a66fc/src/lib/libcrypto/curve25519/curve25519.c#L4905

https://github.com/openbsd/src/blob/7c9f822481c4d6823c0a1b1741f55500f10f1341/lib/libcrypto/bn/bn_rand.c#L153

real-or-random:
Sigh ok, all of this is a huge mess.

> OpenBSD actually has two functions named arc4random_buf().
> The one you linked to is the kernel side one, which as you saw is used in the implementation of the getentropy() syscall.
> The one in this PR (`arc4random_buf(3)`) is part of `libc`, and somewhat confusingly, is itself implemented using that same `getentropy()` syscall, but additionally applies some `ChaCha` based per-process stirring:
> 
> [openbsd/src@`7c9f822`/lib/libc/crypt/arc4random.c#L198](https://github.com/openbsd/src/blob/7c9f822481c4d6823c0a1b1741f55500f10f1341/lib/libc/crypt/arc4random.c#L198)


Oh, you're right. The kernel one is here: 
https://github.com/openbsd/src/blob/deb51e2ea407a9dc0b2340231f789bb752bc00ed/sys/dev/rnd.c#L538

Internally, it performs the same  ChaCha "stirring" as the libc one.  

So IIUC the entire stack is as follows. There's 
1. an `entropy_pool` in the kernel which is used to seed
2. the kernel `arc4random_buf` which produces randomness for
3. the `getentropy` syscall which seeds
4. the libc `arc4random_buf` which produces randomness (and hopefully handles process forks) for
5. the user without a need for a syscall

I think this confirms my theory that randomness output by `getentropy` is considered "scarcer" and is recommended to seed software RNGs.  `arc4random_get` is then recommended for users that just want random data, but without syscalls. 

By the way: The way they get rid of error statues is as follows: If libc's call to `getentropy` fails, they [try to kill the process](https://github.com/openbsd/src/blob/7c9f822481c4d6823c0a1b1741f55500f10f1341/lib/libc/crypt/arc4random.h#L32).
 
Another interesting question one can ask: What does Bitcoin Core do?  They [have changed to `arc4random_buf`](https://github.com/bitcoin/bitcoin/pull/24238) as a result of [our previous discussion ](https://github.com/bitcoin-core/secp256k1/pull/748#discussion_r524605472), even though we ended up with a different decision. /cc @thestack
It also contradicts the above theory. If `getentropy` is supposed to seed software RNGs, then Bitcoin Core should use that perhabs.

Okay, what does all of this mean for us? 
 - What we currently (`getentropy`) do is fine.
 - What this PR proposes (`arc4random_buf`) is what Bitcoin Core does, and it is fine. It is probably a bit more appropriate for the use case.

But currently all methods in `random.h` (except maybe on Windows) use syscalls, and it's fine to recommend this. Syscalls are more direct than relying on libc (with its fork handling etc.) and the performance penalty is not huge and certainly okay for example code... So to be honest, I'm a bit reluctant to support this PR because it seems arbitrary. It's changes good code to code which is not clearly better. 

I think a better approach would be to sync our code with Bitcoin Core. There could be a library/header file which is used by both, and then we don't need to maintain two versions of this mess. (Of course, requirements are bit different, e.g., getting randomness vs seeding RNGs vs just getting randomness, but in the end, none of this really matters.)

PR #1217, opened 2023-03-01
Add secp256k1_fe_add_int function

sipa:
Suggested here: https://github.com/bitcoin-core/secp256k1/pull/1118#discussion_r1067131562

This adds a new field function secp256k1_fe_add_int which is equivalent to secp256k1_fe_set_int + secp256k1_fe_add, which is a pattern that occurs a few times, and uses a needlessly convoluted computation for an absolutely trivial effect (incrementing one limb variable). Existing PRs like #1118 and #1129 add more ways this function would be useful.

### DIFF
--- a/src/field.h
+++ b/src/field.h
@@ -85,6 +85,9 @@ static void secp256k1_fe_get_b32(unsigned char *r, const secp256k1_fe *a);
  *  as an argument. The magnitude of the output is one higher. */
 static void secp256k1_fe_negate(secp256k1_fe *r, const secp256k1_fe *a, int m);
 
+/** Adds a small integer (up to 0x7FFF) to r. The resulting magnitude increases by one. */
+static void secp256k1_fe_add_int(secp256k1_fe *r, int a);
+
 /** Multiplies the passed field element with a small integer constant. Multiplies the magnitude by that
  *  small integer. */
 static void secp256k1_fe_mul_int(secp256k1_fe *r, int a);
--- a/src/field_5x52_impl.h
+++ b/src/field_5x52_impl.h
@@ -425,6 +425,20 @@ SECP256K1_INLINE static void secp256k1_fe_mul_int(secp256k1_fe *r, int a) {
 #endif
 }
 
+SECP256K1_INLINE static void secp256k1_fe_add_int(secp256k1_fe *r, int a) {
+#ifdef VERIFY
+    secp256k1_fe_verify(r);
+    VERIFY_CHECK(a >= 0);
+    VERIFY_CHECK(a <= 0x7FFF);
+#endif
+    r->n[0] += a;
+#ifdef VERIFY
+    r->magnitude += 1;
+    r->normalized = 0;
+    secp256k1_fe_verify(r);
+#endif
+}
+
 SECP256K1_INLINE static void secp256k1_fe_add(secp256k1_fe *r, const secp256k1_fe *a) {
 #ifdef VERIFY
     secp256k1_fe_verify(a);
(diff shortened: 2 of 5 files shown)


src/group_impl.h
```diff
@@ -376,7 +376,7 @@ static void secp256k1_gej_add_var(secp256k1_gej *r, const secp256k1_gej *a, cons
             secp256k1_gej_double_var(r, a, rzr);
         } else {
             if (rzr != NULL) {
-                secp256k1_fe_set_int(rzr, 0);
+                secp256k1_fe_clear(rzr);
```
real-or-random, line 379: This reminds me of https://github.com/bitcoin-core/secp256k1/pull/636/commits/b33a8e49e89f5a1ea02a9b9b9b208cb4f3d59e44 Okay, the PR is almost 4 years old now... :/ But I should continue the work at some point.

Currently, `clear` is mostly used to kill secrets (though that doesn't work due to dead-store elimination), and this should be separated from setting to 0 in the future for the purposes of the aforementioned PR. Now, the advantage of `clear` over `set_int(0)` is that the former sets the magnitude to 0 instead of 1. In the mentioned PR, [I change `set_int` to do the same for input 0](https://github.com/bitcoin-core/secp256k1/pull/636/commits/b17a7df8145a6a86d49c354c7e7b59a432ea5346) but that's a bad idea because we want [statically implied magnitude](https://github.com/bitcoin-core/secp256k1/issues/1001) in the long term. The proper solution to all of this is a `set_zero` function that guarantees to set the value and magnitude to 0 (and won't just clear out memory).

What does this mean for this PR? To avoid getting side tracked again, I'd say don't introduce that new function here, but maybe keep that line and the instances below just how it is. That avoids introducing more code locations where `clear` is not used for clearing memory. The current behavior of setting magnitude of 1 is apparently good enough with the current code.

src/field_5x52_impl.h
```diff
@@ -425,6 +425,20 @@ SECP256K1_INLINE static void secp256k1_fe_mul_int(secp256k1_fe *r, int a) {
 #endif
 }
 
+SECP256K1_INLINE static void secp256k1_fe_add_int(secp256k1_fe *r, int a) {
+#ifdef VERIFY
+    secp256k1_fe_verify(r);
+    VERIFY_CHECK(a >= 0);
+    VERIFY_CHECK(a <= 0xFFFF);
```
real-or-random, line 432: Is there a reason to keep it in these bounds? Sorry if it's obvious, but I don't see it.

src/field.h
```diff
@@ -85,6 +85,9 @@ static void secp256k1_fe_get_b32(unsigned char *r, const secp256k1_fe *a);
  *  as an argument. The magnitude of the output is one higher. */
 static void secp256k1_fe_negate(secp256k1_fe *r, const secp256k1_fe *a, int m);
 
+/** Adds a small integer (up to 0x7FFF) to r. The resulting magnitude increases by one. */
```
real-or-random, line 88:
```suggestion
/** Adds a small integer (up to 0xFFFF) to r. The resulting magnitude increases by one. */
```

At least that's what you VERIFY_CHECK.

real-or-random:
Nice! Concept ACK

src/field_5x52_impl.h
real-or-random: 0x7FFF is the minimum portable MAX_INT, see https://github.com/bitcoin-core/secp256k1/pull/943#discussion_r729889981

jonasnick:
ACK b081f7e4cbfd27edc36e823dcd93537a46f7d2a6

src/group_impl.h
```diff
@@ -227,7 +227,7 @@ static int secp256k1_ge_set_xo_var(secp256k1_ge *r, const secp256k1_fe *x, int o
     secp256k1_fe_sqr(&x2, x);
     secp256k1_fe_mul(&x3, x, &x2);
     r->infinity = 0;
-    secp256k1_fe_add(&x3, &secp256k1_fe_const_b);
```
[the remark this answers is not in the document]
roconnor-blockstream: I suppose I can do that.

PR #1221, opened 2023-03-07
Update Changelog

### LINKED ISSUE #1220: Update Changelog
Looking at the [PRs since the last release](https://github.com/bitcoin-core/secp256k1/pulls?q=is%3Apr+is%3Aclosed+merged%3A%3E%3D2022-12-12+base%3Amaster+sort%3Aupdated-asc), the following ones seem to deserve a changelog entry:

- https://github.com/bitcoin-core/secp256k1/pull/1178
- https://github.com/bitcoin-core/secp256k1/pull/1209
- maybe https://github.com/bitcoin-core/secp256k1/pull/1212

real-or-random:
Fixes #1220.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+#### Added
+ - Usage examples: Added a recommended method for securely clearing sensitive data, e.g., secret keys, from memory.
+ - Tests: Added a new test binary `noverify_tests`. This binary runs the tests without some additional checks present in the ordinary `tests` binary and is thereby closer to production binaries. The `noverify_tests` binary is automatically run as part of the `make check` target. 
+
+#### Fixed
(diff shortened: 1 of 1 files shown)


jonasnick (commented):
ACK d232112fa7e17fbbed7927bf02e8b1a0c54aeb53

jonasnick:
The release process also [reminds us to add a section on ABI compatibility](https://github.com/bitcoin-core/secp256k1/blob/master/doc/release-process.md#regular-release). As far as I am aware, there has been no changes that affect the ABI but not the API.  The ABI is backward incompatible for the same reason the API is backward incompatible (changed handling of static contexts). But this doesn't need to be added in this PR.

real-or-random:
> As far as I am aware, there has been no changes that affect the ABI but not the API. The ABI is backward incompatible for the same reason the API is backward incompatible (changed handling of static contexts).

This looks correct to me. (Indeed, if an old consumer program is linked against our new library, the new library could break call the illegal callback if the consumer tries to (for example) randomize the static context.)

> . But this doesn't need to be added in this PR.

Do you want to give it a try and also increase the ABI numbers?

PR #1218, opened 2023-03-06
Update overflow check

roconnor-blockstream:
One does not simply check for integer overlow.

PR #1222, opened 2023-03-07
Remove redundant checks.

roconnor-blockstream:
These abs checks are implied by the subsequent line, and with the subsequent line written as it is, no underflow is possible with signed integers.

Follows up on https://github.com/bitcoin-core/secp256k1/pull/1218.


real-or-random (approved):
utACK 5d8f53e31293c582fb3fe02157bc67d2eeccea77

real-or-random (approved):
ACK 5d8f53e31293c582fb3fe02157bc67d2eeccea77

PR #1227, opened 2023-03-08
readme: Use correct build type in CMake/Windows build instructions

real-or-random:


hebasto (approved):
ACK 756b61d451d2e00c357a6b55863717519164cdb2, it is correct to provide the "RelWithDebInfo" configuration in multi-config setup, as the same build type is the default in single-config setups.

Issue #1224, opened 2023-03-08
build: Meta-issue for follow-ups to initial CMake merge (#1113)


real-or-random:
Issues that have been (re)discovered in #1113 but have not been addressed there: 

Affecting both build systems:
 - [ ] Unified docs for configuration autotools/CMake/manual builds (https://github.com/bitcoin-core/secp256k1/pull/1113#issuecomment-1171307979 and  https://github.com/bitcoin-core/secp256k1/pull/1142)
 - [ ] Improve MSVC warning options (https://github.com/bitcoin-core/secp256k1/pull/1113#discussion_r1124276576)
 - [ ] Set `-fvisibility-inlines-hidden`? (https://github.com/bitcoin-core/secp256k1/pull/1113#discussion_r1124273356)
 - [ ] Consider changing `DLL_EXPORT` TO `SECP256k1_DLL_EXPORT` (https://github.com/bitcoin-core/secp256k1/pull/1113#discussion_r1126797102)
 - [ ] Consider setting CMake's default of /MD for MSVC also in autotools builds. Perhaps make `CMAKE_MSVC_RUNTIME_LIBRARY` a CMake cache variable. (https://github.com/bitcoin-core/secp256k1/pull/1113#issuecomment-1458005771)
 - [ ] Print out a warning in the tests if both `VERIFY` and `COVERAGE` are defined. For clarity, this should be done in the C source, not in the build system. (https://github.com/bitcoin-core/secp256k1/pull/1113#discussion_r1127856040)
 - [ ] Consider dropping support for Valgrind on MacOS (https://github.com/bitcoin-core/secp256k1/pull/1113#discussion_r1128588031)
 - [ ] Actually check if ARM assembly is supported (https://github.com/bitcoin-core/secp256k1/pull/1113#issuecomment-1460298292 and also tracked in https://github.com/bitcoin-core/secp256k1/issues/1034) 

Affecting only CMake:
 - [x] Changelog entry (https://github.com/bitcoin-core/secp256k1/pull/1113#issuecomment-1460298292; solved by #1125)
 - [ ] Add CMake instructions to release process (solved by #1226)
 - [ ] Compute string `secp256k1` (e.g., in `target_link_libraries`) as substring of `PROJECT_NAME`. That will help with distinguishing forks. (https://github.com/bitcoin-core/secp256k1/pull/1229#discussion_r1130064718).
[...]

PR #1230, opened 2023-03-08
Build: allow static or shared but not both

theuni:
Continuing from here: https://github.com/bitcoin-core/secp256k1/issues/1224#issuecomment-1460438227

Unfortunately it wasn't really possible to keep a clean diff here because of the nature of the change. I suggest reviewing the lib creation stuff in its entirety, sorry about that :\

Rather than allowing for shared and static libs to be built at the same time like autotools, this PR switches to the CMake convention of allowing only 1.

A new `BUILD_SHARED_LIBS` option is added to match CMake convention, as well as a `SECP256K1_DISABLE_SHARED` option which overrides it. That way even projects which have `BUILD_SHARED_LIBS=1` can opt-into a static libsecp in particular.

Details:

Two object libraries are created: `secp_common_objs` and `secp_precomputed_objs`. All tests and benchmarks use these objects directly, that way they continue to be linked statically even if `BUILD_SHARED_LIBS` is true.

`secp_interface` allows flags/properties to be set for the object libs.

Because the objs don't know what they're going to be linked into, they need to be told how to deal with PIC and DLL_EXPORT.


theuni:
Hmm, this blows up when tested with CMake 3.1. Going to convert to draft for now until that works.

theuni:
I have this working locally but it's pretty ugly due to sacrifices made for such an ancient CMake version.

It seems like such a shame to introduce a new buildsystem that's already saddled with technical debt. What was the reasoning behind choosing compatibility for CMake 3.1. Is that up for discussion still?

hebasto:
> I have this working locally but it's pretty ugly due to sacrifices made for such an ancient CMake version.
> 
> 
> 
> It seems like such a shame to introduce a new buildsystem that's already saddled with technical debt. What was the reasoning behind choosing compatibility for CMake 3.1. Is that up for discussion still?

This project docs do not stipulate any minimum supported versions for compilers, which implicitly assumes that they can be old enough. In the same spirit, the minimum required CMake version was chosen as oldest as possible. Particularly, CMake 3.1 supports C_STANDARD property.

On the recent IRC discussion, it was established that we need at least CMake 3.11 to support SameMinorVersion.

Speaking of bumping minimum required CMake version, I'd prefer 3.13 which has improved syntax for object libraries and available on Debian 10.

theuni:
Ok, great. I think 3.13 is what I needed for objects for my desired approach as well.

I'll go back to my non-ancient approach and push something up tomorrow.

hebasto:
I apologies if my question looks out of this PR scope, but what are actual drawbacks of using global `set(CMAKE_POSITION_INDEPENDENT_CODE ON)` _nowadays_? It restores PIC-consistency across all build artifacts while keeping ability to build dual shared/static libraries.

hebasto:
> Ok, great. I think 3.13 is what I needed for objects for my desired approach as well.

> But I'd also be okay with a version newer than 3.13 if it unlocks more features that will really help. Again, we still have autotools for people on older distros.

The current WIP for implementing a CMake-based build system in Bitcoin Core [contains](https://github.com/hebasto/bitcoin/tree/cmake-staging):
```cmake
cmake_minimum_required(VERSION 3.13...3.24)
```
which is based mostly on availability of CMake across current Linux distros, as macOS (via Homebrew) and Windows (via Visual Studio) provide quite recent versions.

theuni:
> I apologies if my question looks out of this PR scope, but what are actual drawbacks of using global `set(CMAKE_POSITION_INDEPENDENT_CODE ON)` _nowadays_? It restores PIC-consistency across all build artifacts while keeping ability to build dual shared/static libraries.

It wouldn't be the end of the world, it's just a pretty clear sign to me that we're not defining our targets explicitly enough if we need to set a global and let it trickle down. Also, it _is_ slower for architectures that care. This means that on some architectures and to some very small degree, libtool-built libsecp would be faster than cmake-built.

But besides, there are other reasons (like dllimport/export) to want to compile shared objects differently that static. So I don't see why we'd use a global hammer for all downstreams rather than just setting the correct properties.

Note that the CMake docs actually [recommend this approach](https://cmake.org/cmake/help/latest/guide/tutorial/Selecting%20Static%20or%20Shared%20Libraries.html) (very bottom). They use `${BUILD_SHARED_LIBS}` as opposed to querying the property itself which is about the same thing I guess, but a bit less explicit imo.


For Core, however, sure. Let's set the global :)

hebasto:
@theuni 
> It seems like such a shame to introduce a new buildsystem that's already saddled with technical debt.

@real-or-random 
> But I'd also be okay with a version newer than 3.13 if it unlocks more features that will really help. Again, we still have autotools for people on older distros.

I [agree](https://github.com/bitcoin-core/secp256k1/pull/1230#issuecomment-1461752437) to bump the minimum required CMake version up to 3.13.

---

@fanquake Your opinion regarding the minimum required CMake version will be much appreciated.

hebasto:
This PR changes the way how examples are linked. Aren't they [supposed](https://github.com/bitcoin-core/secp256k1/pull/1113#discussion_r1072776703) to be linked as if for actual users (with no using of internal objects)?

CMakeLists.txt
```diff
@@ -230,8 +229,7 @@ message("\n")
 message("secp256k1 configure summary")
 message("===========================")
 message("Build artifacts:")
-message("  shared library ...................... ${SECP256K1_BUILD_SHARED}")
-message("  static library ...................... ${SECP256K1_BUILD_STATIC}")
+message("  shared library ...................... ${BUILD_SHARED_LIBS}")
```
hebasto, line 232: nit: When `BUILD_SHARED_LIBS` is FALSE/OFF, at first glance it is not obvious that the build system is configured to build a static library.

hebasto:
Concept ACK.

theuni:
> [f96bd51](https://github.com/bitcoin-core/secp256k1/commit/f96bd5184b6534de7c20efba1a35248d940bebb5)
> 
> This PR changes the way how examples are linked. Aren't they [supposed](https://github.com/bitcoin-core/secp256k1/pull/1113#discussion_r1072776703) to be linked as if for actual users (with no using of internal objects)?

Ah thanks, that's very helpful context. This PR indeed changes to link everything against objects which I thought would be easier than dealing with tests linked against local shared libs, but that's a really good point that the api/abi really isn't getting tested this way.

I will re-work and see what breaks when linked against actual shared libs.

theuni:
Just to reiterate: I threw this PR up for discussion. I wanted to see what the simplification would look like, but it's not my intention to be a bully about it.

I'm not wedded to it (though I do prefer it ;) and I agree there are still some outstanding questions (@hebasto has mentioned downstream packaging).

theuni:
> I will re-work and see what breaks when linked against actual shared libs.

Huh, it just worked. Neat.

examples/CMakeLists.txt
```diff
@@ -8,27 +8,22 @@ target_compile_options(example INTERFACE
...
-  if(MSVC)
-    target_link_options(example INTERFACE /IGNORE:4217)
-  endif()
+if(NOT BUILD_SHARED_LIBS AND MSVC)
+  target_link_options(example INTERFACE /IGNORE:4217)
 endif()
 
 add_executable(ecdsa_example ecdsa.c)
-target_link_libraries(ecdsa_example example)
+target_link_libraries(ecdsa_example example secp256k1)
```
hebasto, line 16: `secp256k1` can be mentioned in the `example` INTERFACE library.

src/CMakeLists.txt
```diff
@@ -1,107 +1,85 @@
...
+add_library(secp_interface INTERFACE)
+target_link_libraries(secp_common_objs PRIVATE secp_interface)
+target_link_libraries(secp_precomputed_objs PRIVATE secp_interface)
+
+# Add objects explicitly rather than linking to the object libs to keep them
+# from being exported.
+add_library(secp256k1 $<TARGET_OBJECTS:secp_precomputed_objs> $<TARGET_OBJECTS:secp_common_objs>)
+
+target_compile_definitions(secp_interface INTERFACE
+  $<$<C_COMPILER_ID:MSVC>:_CRT_SECURE_NO_WARNINGS>
```
theuni, line 26: @hebasto What is this btw? I don't see it in the autotools build and I suspect I didn't handle it correctly here.

hebasto: During working on #1113, at some point in the past, I had a ["unsafe CRT Library function"](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-3-c4996?view=msvc-170#unsafe-crt-library-functions) warning. However, I'm not able to reproduce it now.

hebasto: FWIW, a similar suppression is used in [Bitcoin Core](https://github.com/bitcoin/bitcoin/blob/c7f1d95f52883d7087b74f45f5e4ce1100d51149/build_msvc/common.init.vcxproj.in#L93):
```XML
      <PreprocessorDefinitions>_SILENCE_CXX17_CODECVT_HEADER_DEPRECATION_WARNING;ZMQ_STATIC;NOMINMAX;WIN32;HAVE_CONFIG_H;_CRT_SECURE_NO_WARNINGS;_SCL_SECURE_NO_WARNINGS;_CONSOLE;_WIN32_WINNT=0x0601;_WIN32_IE=0x0501;WIN32_LEAN_AND_MEAN;%(PreprocessorDefinitions)</PreprocessorDefinitions>
```

theuni: I realize this is kinda unrelated here, but it snuck in in #1113 so we may as well address it.

Due to the nature of that suppression, I think we'd rather know about it and address it. I'm tempted to remove this as part of this PR and ask you to submit a new one to rea-add it if you're ever able to reproduce again. That way we can debate the merit/alternatives.

Thoughts on that?

hebasto: > I'm tempted to remove this as part of this PR...

Agree.

theuni:
Looking at this again... now that we're (correctly) linking against the real shared lib rather than object files, I think this could be simplified even further to a simple helper static lib. I'm going to continue to work on this a bit.

examples/CMakeLists.txt
```diff
@@ -8,14 +8,10 @@ target_compile_options(example INTERFACE
...
-  target_link_libraries(example INTERFACE secp256k1)
-elseif(SECP256K1_BUILD_STATIC)
-  target_link_libraries(example INTERFACE secp256k1_static)
-  if(MSVC)
-    target_link_options(example INTERFACE /IGNORE:4217)
-  endif()
+if(NOT BUILD_SHARED_LIBS AND MSVC)
+  target_link_options(example INTERFACE /IGNORE:4217)
 endif()
+target_link_libraries(example INTERFACE secp256k1)
```
hebasto, line 14: Maybe reuse `target_link_libraries` command:

```diff
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -6,12 +6,12 @@ target_compile_options(example INTERFACE
   $<$<C_COMPILER_ID:MSVC>:/wd4005>
 )
 target_link_libraries(example INTERFACE
+  secp256k1
   $<$<PLATFORM_ID:Windows>:bcrypt>
 )
 if(NOT BUILD_SHARED_LIBS AND MSVC)
   target_link_options(example INTERFACE /IGNORE:4217)
 endif()
-target_link_libraries(example INTERFACE secp256k1)
 
 add_executable(ecdsa_example ecdsa.c)
 target_link_libraries(ecdsa_example example)
```
?

src/CMakeLists.txt
```diff
@@ -1,107 +1,81 @@
...
   precomputed_ecmult_gen.c
 )
-set(internal_obj "$<TARGET_OBJECTS:precomputed>" "${common_obj}")
 
-add_library(secp256k1 SHARED EXCLUDE_FROM_ALL
-  secp256k1.c
-  ${internal_obj}
-)
-target_include_directories(secp256k1 INTERFACE
+add_library(secp_common_objs OBJECT EXCLUDE_FROM_ALL secp256k1.c)
```
hebasto, line 15: Why is `secp256k1.c` the source of the `secp_common_objs` library, not the `secp256k1` one?

theuni: Yeah, I'm going to do away with common entirely as it's not needed now that I understand that we're linking benches/tests directly to shared libs. Will push up a cleaner version next week.

hebasto:
@theuni 

What do you think about splitting minimum required CMake version bumping into a separate PR, as it is a blocker for other [issues](https://github.com/bitcoin-core/secp256k1/issues/1235)?

theuni:
Updated, fixed nits, squashed, and rebased. Now tested with external arm asm which was broken but now works.

Should be ready for review.

src/CMakeLists.txt
```diff
@@ -1,97 +1,78 @@
...
-endif()
 
 if(SECP256K1_BUILD_BENCHMARK)
   add_executable(bench bench.c)
-  target_link_libraries(bench link_library)
-  add_executable(bench_internal bench_internal.c ${internal_obj})
-  add_executable(bench_ecmult bench_ecmult.c ${internal_obj})
+  target_link_libraries(bench secp256k1)
+  add_executable(bench_internal bench_internal.c)
+  target_link_libraries(bench_internal secp_precomputed_objs secp_asm)
```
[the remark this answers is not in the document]
hebasto: > Would it make sense (if it's possible) to define a library/target/whatever that combines secp256k1_precomputed_objs and secp_asm?

I was thinking about the same :)
The diff as follows can help:
```diff
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -10,14 +10,12 @@ add_library(secp_precomputed_objs OBJECT EXCLUDE_FROM_ALL
 # from being exported.
 add_library(secp256k1 secp256k1.c $<TARGET_OBJECTS:secp_precomputed_objs>)
 
-add_library(secp_asm INTERFACE)
 if(SECP256K1_ASM STREQUAL "arm")
   add_library(secp_asm_arm OBJECT EXCLUDE_FROM_ALL)
   target_sources(secp_asm_arm PUBLIC
     asm/field_10x26_arm.s
   )
   target_sources(secp256k1 PRIVATE $<TARGET_OBJECTS:secp_asm_arm>)
-  target_link_libraries(secp_asm INTERFACE secp_asm_arm)
 endif()
 
 #Define our export symbol only for Win32 and only for shared libs.
@@ -38,23 +36,29 @@ set_target_properties(secp256k1 PROPERTIES
   SOVERSION "${${PROJECT_NAME}_LIB_VERSION_CURRENT}"
 )
 
+add_library(secp256k1_internal INTERFACE)
+set_property(TARGET secp256k1_internal PROPERTY INTERFACE_SOURCES
+  $<TARGET_OBJECTS:secp_precomputed_objs>
+  $<$<TARGET_EXISTS:secp_asm_arm>:$<TARGET_OBJECTS:secp_asm_arm>>
+)
+
 if(SECP256K1_BUILD_BENCHMARK)
   add_executable(bench bench.c)
   target_link_libraries(bench secp256k1)
   add_executable(bench_internal bench_internal.c)
-  target_link_libraries(bench_internal secp_precomputed_objs secp_asm)
+  target_link_libraries(bench_internal secp256k1_internal)
   add_executable(bench_ecmult bench_ecmult.c)
-  target_link_libraries(bench_ecmult secp_precomputed_objs secp_asm)
+  target_link_libraries(bench_ecmult secp256k1_internal)
 endif()
 
 if(SECP256K1_BUILD_TESTS)
   add_executable(noverify_tests tests.c)
-  target_link_libraries(noverify_tests secp_precomputed_objs secp_asm)
+  target_link_libraries(noverify_tests secp256k1_internal)
   add_test(noverify_tests noverify_tests)
   if(NOT CMAKE_BUILD_TYPE STREQUAL "Coverage")
     add_executable(tests tests.c)
     target_compile_definitions(tests PRIVATE VERIFY)
-    target_link_libraries(tests secp_precomputed_objs secp_asm)
+    target_link_libraries(tests secp256k1_internal)
     add_test(tests tests)
   endif()
 endif()
@@ -62,7 +66,7 @@ endif()
 if(SECP256K1_BUILD_EXHAUSTIVE_TESTS)
   # Note: do not include secp_precomputed_objs in exhaustive_tests (it uses runtime-generated tables).
   add_executable(exhaustive_tests tests_exhaustive.c)
-  target_link_libraries(exhaustive_tests secp_asm)
+  target_link_libraries(exhaustive_tests $<TARGET_NAME_IF_EXISTS:secp_asm_arm>)
   target_compile_definitions(exhaustive_tests PRIVATE $<$<NOT:$<CONFIG:Coverage>>:VERIFY>)
   add_test(exhaustive_tests exhaustive_tests)
 endif()
```

CMakeLists.txt
```diff
@@ -26,11 +26,10 @@ set(CMAKE_C_EXTENSIONS OFF)
...
 list(APPEND CMAKE_MODULE_PATH ${PROJECT_SOURCE_DIR}/cmake)
 
-# We do not use CMake's BUILD_SHARED_LIBS option.
-option(SECP256K1_BUILD_SHARED "Build shared library." ON)
-option(SECP256K1_BUILD_STATIC "Build static library." ON)
-if(NOT SECP256K1_BUILD_SHARED AND NOT SECP256K1_BUILD_STATIC)
-  message(FATAL_ERROR "At least one of SECP256K1_BUILD_SHARED and SECP256K1_BUILD_STATIC must be enabled.")
+option(BUILD_SHARED_LIBS "Build shared libraries" ON)
+option(SECP256K1_DISABLE_SHARED "Disable shared library. Overrides BUILD_SHARED_LIBS." OFF)
+if (SECP256K1_DISABLE_SHARED)
```
hebasto, line 31: style nit:
```suggestion
if(SECP256K1_DISABLE_SHARED)
```

src/CMakeLists.txt
```diff
@@ -1,107 +1,78 @@
...
+  target_link_libraries(secp_asm INTERFACE secp_asm_arm)
 endif()
 
-add_library(precomputed OBJECT
-  precomputed_ecmult.c
-  precomputed_ecmult_gen.c
-)
-set(internal_obj "$<TARGET_OBJECTS:precomputed>" "${common_obj}")
+#Define our export symbol only for Win32 and only for shared libs.
+if (WIN32)
```
hebasto, line 24: style nit:
```suggestion
if(WIN32)
```

src/CMakeLists.txt
```diff
@@ -1,107 +1,78 @@
...
+
+# Add objects explicitly rather than linking to the object libs to keep them
+# from being exported.
+add_library(secp256k1 secp256k1.c $<TARGET_OBJECTS:secp_precomputed_objs>)
+
+add_library(secp_asm INTERFACE)
 if(SECP256K1_ASM STREQUAL "arm")
-  add_library(common OBJECT
+  add_library(secp_asm_arm OBJECT EXCLUDE_FROM_ALL)
+  target_sources(secp_asm_arm PUBLIC
```
hebasto, line 16: Why `PUBLIC`? If it was unintentional then adding of the `target_sources` seems unneeded, no?

hebasto (commented):
Approach ACK f5d117d9df400eb7cfa5fd0caf2af84f176d4a4e.

src/CMakeLists.txt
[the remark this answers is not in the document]
theuni: > Would it make sense (if it's possible) to define a library/target/whatever that combines secp256k1_precomputed_objs and secp_asm?

I didn't combine them here because I wasn't sure if the combo of `secp256k1_precomputed_objs` and `secp_asm` made any logical sense worth defining, as opposed to just listing them explicitly where needed.

Happy to take a patch like @hebasto's above.

> Also naming... can we be more consistent than having secp256k1 in one and secp in the other?

Roger.

src/CMakeLists.txt
theuni: Otherwise it ends up missing as part of the interface and link fails. Is there some trick I'm missing?

src/CMakeLists.txt
```diff
@@ -1,107 +1,78 @@
...
     asm/field_10x26_arm.s
```
hebasto, line 17: Unrelated to this PR, just a note (for myself).

It will be more CMake-ish to move the `-DUSE_ASM_X86_64=1` definition from globals to a usage requirement of a library which compiles this source file.

src/CMakeLists.txt
```diff
@@ -1,107 +1,78 @@
...
 include(GNUInstallDirs)
-set(${PROJECT_NAME}_installables "")
 
+add_library(secp_precomputed_objs OBJECT EXCLUDE_FROM_ALL
+  precomputed_ecmult.c
+  precomputed_ecmult_gen.c
+)
+
+# Add objects explicitly rather than linking to the object libs to keep them
+# from being exported.
```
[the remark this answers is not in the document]
theuni: If you link against it, CMake will complain that it's not added to the export list. If you then add it to that list, it gets [exported as an interface library in the installed `libsecp256k1-*.cmake` files](https://gitlab.kitware.com/cmake/cmake/-/issues/18935#note_521599). I'm unsure of the consequences of that, but because it's so easily avoidable I decided to just avoid finding out. If nothing else it feels a little icky to expose our internal structure when it's not necessary.

I can change it if there's any reason to though.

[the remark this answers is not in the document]
hebasto: > If you link against it, CMake will complain that it's not added to the export list.

Right. I can reproduced it with `-DBUILD_SHARED_LIBS=OFF`.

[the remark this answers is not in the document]
theuni: Ah yes sorry, forgot to mention that requirement.

hebasto (approved):
ACK 71648cf19fdd986e005327736bc7b0b5439fe67a, tested on Ubuntu 22.04 (both shared and static libs).

Although some style nits still remain unresolved:
- [naming consistency](https://github.com/bitcoin-core/secp256k1/pull/1230#discussion_r1137778372)
- spaces ([here](https://github.com/bitcoin-core/secp256k1/pull/1230#discussion_r1138721938) and [here](https://github.com/bitcoin-core/secp256k1/pull/1230#discussion_r1138723174))

src/CMakeLists.txt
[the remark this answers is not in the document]
theuni: Imo the complexity here isn't worth it and I'd rather not take this change unless anyone feels strongly about it.

theuni:
Thanks, nits addressed. Will squash after deciding on https://github.com/bitcoin-core/secp256k1/pull/1230#discussion_r1143896402 .

hebasto:
> ... after deciding on [#1230 (comment)](https://github.com/bitcoin-core/secp256k1/pull/1230#discussion_r1143896402) .

I'm OK with the current state for a reason mentioned [here](https://github.com/bitcoin-core/secp256k1/pull/1230#discussion_r1139951225).

theuni:
> I'm OK with the current state for a reason mentioned [here](https://github.com/bitcoin-core/secp256k1/pull/1230#discussion_r1139951225).

Thanks! Squashed.

hebasto (approved):
ACK 79fed0ca8c0a2f0f073c3735f6c5e50b23c7ef5a, tested on Ubuntu 22.04 and Windows 11 22H2 (using VS Build Tools).

theuni:
Ping @real-or-random . Bumping for visibility in parallel to #1272.

If we're going to have a bumped CMake requirement, we may as well take advantage :)

real-or-random:
I'll have a look next week!

real-or-random (commented):
ACK mod nits

src/CMakeLists.txt
```diff
@@ -1,107 +1,78 @@
 # Must be included before CMAKE_INSTALL_INCLUDEDIR is used.
 include(GNUInstallDirs)
-set(${PROJECT_NAME}_installables "")
 
+add_library(secp256k1_precomputed_objs OBJECT EXCLUDE_FROM_ALL
```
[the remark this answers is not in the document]
theuni: I just don't see it necessary to add the helper libs to `all`. As far as I can tell, it does the reasonable thing of rebuilding as needed.

theuni:
Addressed @real-or-random's feedback:
- Renamed `secp256k1_precomputed_objs` to `secp256k1_precomputed`
- Added comment about libtool

See diff with `git diff 79fed0ca8c..ef49a11d296`

hebasto (approved):
re-ACK ef49a11d29601e09e94134975c968e92c0214102, only [suggested](https://github.com/bitcoin-core/secp256k1/pull/1230#pullrequestreview-1388191165) changes since my recent [review](https://github.com/bitcoin-core/secp256k1/pull/1230#pullrequestreview-1352125381).

real-or-random (approved):
ACK ef49a11d29601e09e94134975c968e92c0214102

PR #1233, opened 2023-03-10
cmake: Move runtime output artifacts to `${PROJECT_BINARY_DIR}`

hebasto:
From [IRC](https://gnusha.org/secp256k1/2023-03-08.log):
> 12:32 \< sipa\> With cmake, the build artefacts seem to end up in build/src/, while I'd expect them in build/.

Implemented in this PR.

On Windows, this also simplifies running examples, because the DLL must resides either in the same folder where the executable is or somewhere in `PATH`.


real-or-random (commented):
Hm, I don't know. This seems a bit arbitrary.

 If we now deviate from putting the artifacts in their respective source directory anyway, and setting the artifact path is just setting a variable, would it make sense to also just put anything in `${PROJECT_BINARY_DIR}` (or in a dedicated subdir of that)?

hebasto:
> Hm, I don't know. This seems a bit arbitrary.

Agree. But adding each path for every configuration to `PATH`, which is the case for multi-config, is really annoying.

>  If we now deviate from putting the artifacts in their respective source directory anyway, and setting the artifact path is just setting a variable, would it make sense to also just put anything in `${PROJECT_BINARY_DIR}` (or in a dedicated subdir of that)?

Like `${PROJECT_BINARY_DIR}/out`?

real-or-random:
> Agree. But adding each path for every configuration to `PATH`, which is the case for multi-config, is really annoying.

Indeed. It's also annoying for devs on Windows (though noone uses Windows right now...)

> Like `${PROJECT_BINARY_DIR}/out`? Or `${PROJECT_BINARY_DIR}/bin`?

Yeah, I had something like this in mind. `out` is probably a bit nicer because `bin` looks wrong for `.so` etc. Just not convinced it's worth it. Having a subdir is a cheap and feels tiny bit cleaner because it avoids name collisions with existing files (e.g., what if we have a binary `examples`). Though in that case, autotools would break anyway... I guess there's no clear right or wrong here, but if we want to move all artifacts to a single location, then `${PROJECT_BINARY_DIR}` is probably good enough, and then it matches the autotools behavior at least. As discussed in IRC, we don't care much and should do whatever makes things easier.

hebasto:
> ... if we want to move all artifacts to a single location, then `${PROJECT_BINARY_DIR}` is probably good enough, and then it matches the autotools behavior at least.

Done.

The PR description has been updated (friendly ping @sipa :D ).

hebasto:
Converted to draft for now.

theuni:
Rather than setting the PATH in CI, can we set it in CMake for `make check`?

real-or-random:
> Rather than setting the PATH in CI, can we set it in CMake for `make check`?

I think this makes sense, and then Concept NACK on moving the binaries

hebasto:
@theuni @real-or-random 
> Rather than setting the PATH in CI, can we set it in CMake for `make check`?

Implemented in #1290.

Closing in favour of #1290.

PR #1234, opened 2023-03-10
cmake: Add dev-mode

real-or-random:
To use, invoke `cmake` with argument `--preset dev-mode`.

One disadvantage over `./configure --enable-dev-mode` is that CMake does not provide a way to "hide" presets from users. That is, `cmake --list-presets` will list dev-mode, and it will also appear in `cmake-gui`, even though it's not selectable there due to a bug in cmake-gui.

Solves one item in #1224.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -59,5 +59,7 @@ build-aux/compile
 build-aux/test-driver
 libsecp256k1.pc
 
+### CMake
+/CMakeUserPresets.json
 # Default CMake build directory.
 /build
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,19 @@
+{
+  "cmakeMinimumRequired": {"major": 3, "minor": 21, "patch": 0}, 
+  "version": 3,
+  "configurePresets": [
+    {
+      "name": "dev-mode",
+      "displayName": "Development mode (intended only for developers of the library)",
+      "cacheVariables": {
+        "SECP256K1_EXPERIMENTAL": "ON",
+        "SECP256K1_ENABLE_MODULE_RECOVERY": "ON",
+        "SECP256K1_BUILD_EXAMPLES": "ON"
+      },
+      "warnings": {
+        "dev": true,
+        "uninitialized": true
+      }
+    }
+  ]
+}


CMakePresets.json
```diff
@@ -0,0 +1,23 @@
+{
+  "version": 3,
+  "configurePresets": [
+    {
+      // Don't set "generator" here because developers should still be able to
+      // select their preferred generators. As a result we're affected by CMake
+      // bug https://gitlab.kitware.com/cmake/cmake/-/issues/23341, which means
+      // that dev-mode is not selectable in cmake-gui. But in our case, that's
+      // probably rather a feature than a bug.
```
real-or-random, line 9: GitHub highlights these lines because JSON doesn't have comments. But some parsers are very forgiving, and CMake's parser seems to be of that kind.

hebasto:
Concept ACK.

real-or-random:
nb: This is not urgent, but anyway my availability in the next weeks will be restricted. If this requires changes, feel free to take over and make changes in a separate branch / PR.

hebasto:
> To use, invoke `cmake` with argument `--preset dev-mode`.

Presets are available in CMake 3.19+ only. Should it be documented?

real-or-random:
> Presets are available in CMake 3.19+ only. Should it be documented?

I tend to say it's not necessary. It's for devs only and if you invoke `cmake --preset` on a lower cmake version, I assume it will just error out with `CMake Error: Unknown argument` as for other unknown arguments.

CMakePresets.json
```diff
@@ -0,0 +1,23 @@
+{
+  "version": 3,
```
hebasto, line 2: That makes in usable only with CMake 3.21+.

And CMake 3.19 does fails _silently_  when `cmake --preset dev-mode`.

hebasto (approved):
ACK d12195836d91b6d00423c3c42fd43cdcec6543b7, tested on Ubuntu 22.04.

CMakePresets.json
real-or-random: Hm, I see. Do you have a suggestion? I don't think we can do much about it, except hope that noone tries to use this with 3.19 or 3.20.

hebasto: As it is a developer-specific feature, I'm OK with the current patch.

hebasto:
Maybe add `/CMakeUserPresets.json` to the `.gitignore`?

hebasto:
Well, `cmake --preset dev-mode` will place the build artifacts in the source tree, which is discouraged.

There are two options to avoid such behavior:
- `cmake --preset dev-mode -B build` or
- add `"binaryDir": "${sourceDir}/build"` to the `CMakePresets.json`

real-or-random:
> Maybe add `/CMakeUserPresets.json` to the `.gitignore`?

Done. I took the version without the leading slash because this file could in principle exist in `src/` or `examples/` too.

---

> Well, cmake --preset dev-mode will place the build artifacts in the source tree, which is discouraged.

Okay, yes, this command was just intended as an example.

> There are two options to avoid such behavior:
> *    cmake --preset dev-mode -B build or
> *   add "binaryDir": "${sourceDir}/build" to the CMakePresets.json

My thinking is that, while `${sourceDir}/build` is typically used, this should remain a local decision of the user/developer and not be the responsibility of the template. For example, if someone wants to maintain multiple for testing etc. ... I guess the `binaryDir` is rather useful for automated builds, e.g, to facilitate CI configs.

.gitignore
```diff
@@ -59,5 +59,7 @@ build-aux/compile
 build-aux/test-driver
 libsecp256k1.pc
 
+### CMake
+CMakeUserPresets.json
```
hebasto, line 63: > > Maybe add `/CMakeUserPresets.json` to the `.gitignore`?
> 
> Done. I took the version without the leading slash because this file could in principle exist in `src/` or `examples/` too.

https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html:
> `CMakePresets.json` and `CMakeUserPresets.json` live in the project's root directory.

hebasto (approved):
ACK e8a4a94e4b70cc9bc7c6942204a211bfdb492622

hebasto:
> My thinking is that, while `${sourceDir}/build` is typically used, this should remain a local decision of the user/developer and not be the responsibility of the template.

Command line options allow the user/developer to override a template.

.gitignore
real-or-random: Ok, you're right, fixed.

hebasto (approved):
re-ACK 712090f417b440da62c73db0603eea58edce4cda

real-or-random:
> > My thinking is that, while `${sourceDir}/build` is typically used, this should remain a local decision of the user/developer and not be the responsibility of the template.
> 
> Command line options allow the user/developer to override a template.

Would you prefer this to be added? I'm fine with either, but I still tend to think that this should not the concern of the template. Pragmatically speaking, it doesn't make a big difference, of course.

hebasto:
> > > My thinking is that, while `${sourceDir}/build` is typically used, this should remain a local decision of the user/developer and not be the responsibility of the template.
> > 
> > 
> > Command line options allow the user/developer to override a template.
> 
> Would you prefer this to be added? I'm fine with either, but I still tend to think that this should not be the concern of the template. Pragmatically speaking, it doesn't make a big difference, of course.

Let's merge it in its current state :)

real-or-random:
@theuni Would you like to review this? :)

theuni (approved):
ACK modulo comment nit.

CMakePresets.json
```diff
@@ -0,0 +1,33 @@
+{
+  // Pass `--preset <name>` to CMake to use a preset defined here.
+  //
+  // We curently use version 3 presets which require CMake 3.21+.
+  // Unfortunately, CMake versions before 3.19 may ignore the `--preset`
+  // argument silently. So if the preset is not picked up, make sure you
+  // have a recent enough CMake version.
+  //
+  // Note: JSON does not support officially, but CMake apparently
+  // ignores lines starting with whitespace followed by "//" silently.
```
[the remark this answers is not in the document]
real-or-random: Urghs, thanks for doing the research here. So they first enabled comment support intentionally, then disabled it intentionally.  Currently, [it just works by accident](https://gitlab.kitware.com/cmake/cmake/-/issues/21858#note_908543). 

I'm not sure how we should deal with this. Having the comments is risky because future versions may not support them. (This PR here uses presets only for dev mode, but we may add other presets, e.g., for coverage....)

It seems that CMake's parser defeats all the hacks I could imagine or find in https://stackoverflow.com/questions/244777/can-comments-be-used-in-json. Things we could do:
 - Have a template file and strip `//` lines using grep, generating the actual `.json` from it. But this needs some mechanism to make sure that the file is up to date etc... All doable but seems overkill for this tiny issue.
 - (https://hjson.github.io/ is an entire off-the-shelf solution for this, but that adds a maintainer dependency and is hard to remember.)   
 - We could just drop the comments. CMake templates support a `cmakeMinimumRequired` field which I think is enough to document the need for CMake 3.21. A better warning should then be added to build docs (to be written :D). The comment about the missing generator thing can be in the commit message (not optimal but okayish).

I lean towards dropping comments. Sad that we need to spend time on this...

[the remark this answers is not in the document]
hebasto: > I lean towards dropping comments.

Agree.

CMakePresets.json
```diff
@@ -0,0 +1,19 @@
+{
+  "cmakeMinimumRequired": {"major": 3, "minor": 21, "patch": 0}, 
+  "version": 3,
```
hebasto, line 3: A note for reviewers: version 3 is required as we omit the `generator` entry.

hebasto (approved):
ACK ce5ba9e24dfcceb49ed7f83a87548fd8b3b0cab2

theuni (approved):
ACK ce5ba9e24dfcceb49ed7f83a87548fd8b3b0cab2

PR #1238, opened 2023-03-12
build: bump CMake minimum requirement to 3.13

theuni:
As requested here: https://github.com/bitcoin-core/secp256k1/pull/1230#issuecomment-1464730218 . Ping @hebasto 

Among other things this allows us to link against object libraries.

3.13 has been mentioned several times as a good overlap between newish features and widespread Linux availability.


CMakeLists.txt
```diff
@@ -203,11 +203,9 @@ else()
...
 endif()
 
-if(CMAKE_VERSION VERSION_GREATER 3.2)
-  # Honor visibility properties for all target types.
-  # See: https://cmake.org/cmake/help/latest/policy/CMP0063.html
-  cmake_policy(SET CMP0063 NEW)
-endif()
+# Honor visibility properties for all target types.
+# See: https://cmake.org/cmake/help/latest/policy/CMP0063.html
+cmake_policy(SET CMP0063 NEW)
```
hebasto, line 208: These lines can be removed, as [CMP0063](https://cmake.org/cmake/help/latest/policy/CMP0063.html) was introduced in CMake 3.3, and `cmake_minimum_required(VERSION 3.13)` will set it to `NEW` by default.

hebasto (commented):
Concept ACK.

CMakeLists.txt
theuni: I was going by the docs:

> This policy was introduced in CMake version 3.3. CMake version 3.26.0-rc6 warns when the policy is not set and uses OLD behavior. Use the [cmake_policy()](https://cmake.org/cmake/help/latest/command/cmake_policy.html#command:cmake_policy) command to set it to OLD or NEW explicitly.

Is that not what happens in practice?

hebasto: > > CMake version 3.26.0-rc6 warns when the policy is not set and uses OLD behavior.
> 
> Is that not what happens in practice?

CMake _will_ emit such a warning.

https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html#policy-settings:
> The `cmake_minimum_required(VERSION)` command implicitly invokes the [`cmake_policy(VERSION)`](https://cmake.org/cmake/help/latest/command/cmake_policy.html#command:cmake_policy) command...

> All policies known to the running version of CMake and introduced in the `<min>` ... version or earlier will be set to use `NEW` behavior.

hebasto: Oh, I see...

"use the `cmake_policy()` _explicitly_" vs "_implicitly_ invokes the `cmake_policy()`" -- indeed confusing :D

theuni: Thanks for the explanation, fixed.

hebasto (approved):
ACK 96dd0625112672e841eea723398cc2a1c3489a30

hebasto:
cc @fanquake

real-or-random (approved):
utACK 96dd0625112672e841eea723398cc2a1c3489a30

fanquake:
> cc @fanquake

Seems fine to me. IIRC we are also going to set our minimum to 3.13.

Issue #1235, opened 2023-03-10
build: Meta-issue for follow-ups to initial CMake merge (#1113)


hebasto:
Issues that have been (re)discovered in #1113 but have not been addressed there: 

Affecting both build systems:
 - [ ] Unified docs for configuration autotools/CMake/manual builds (https://github.com/bitcoin-core/secp256k1/pull/1113#issuecomment-1171307979 and  https://github.com/bitcoin-core/secp256k1/pull/1142); WIP -- https://github.com/bitcoin-core/secp256k1/pull/1372
 - [x] Improve MSVC warning options (https://github.com/bitcoin-core/secp256k1/pull/1113#discussion_r1124276576); solved by https://github.com/bitcoin-core/secp256k1/pull/1328
 - [x] Consider changing `DLL_EXPORT` TO `SECP256k1_DLL_EXPORT` (https://github.com/bitcoin-core/secp256k1/pull/1113#discussion_r1126797102); solved by https://github.com/bitcoin-core/secp256k1/pull/1367
 - ~Consider setting CMake's default of /MD for MSVC also in autotools builds. Perhaps make `CMAKE_MSVC_RUNTIME_LIBRARY` a CMake cache variable. (https://github.com/bitcoin-core/secp256k1/pull/1113#issuecomment-1458005771)~ See https://github.com/bitcoin-core/secp256k1/issues/1235#issuecomment-1609373112 and https://github.com/bitcoin-core/secp256k1/issues/1235#issuecomment-1609503130
 - [x] Print out a warning in the tests if both `VERIFY` and `COVERAGE` are defined. For clarity, this should be done in the C source, not in the build system. (https://github.com/bitcoin-core/secp256k1/pull/1113#discussion_r1127856040); solved by https://github.com/bitcoin-core/secp256k1/pull/1333
 - [x] Actually check if ARM assembly is supported (https://github.com/bitcoin-core/secp256k1/pull/1113#issuecomment-1460298292 and also tracked in https://github.com/bitcoin-core/secp256k1/issues/1034); solved by https://github.com/bitcoin-core/secp256k1/pull/1304 
 - ~Set `-fvisibility-inlines-hidden`? (https://github.com/bitcoin-core/secp256k1/pull/1113#discussion_r1124273356).~ See https://github.com/bitcoin-core/secp256k1/issues/1235#issuecomment-1466477275.
[...]

PR #1239, opened 2023-03-12
cmake: Bugfix and other improvements after bumping CMake up to 3.13

hebasto:
This PR:
- resolves two items from #1235, including a bugfix with package version compatibility
- includes other improvements which have become available for CMake 3.13+.

To test the `GENERATOR_IS_MULTI_CONFIG` property on Linux, one can use the "[Ninja Multi-Config](https://cmake.org/cmake/help/latest/generator/Ninja%20Multi-Config.html)" generator:
```sh
cmake -S . -B build -G "Ninja Multi-Config"
```


hebasto:
> ... assuming #1238 goes in.

Rebased on top of the merged #1238 and un-drafted.

CMakeLists.txt
```diff
@@ -177,7 +182,7 @@ if(cached_cmake_build_type)
 endif()
 
 set(default_build_type "RelWithDebInfo")
-if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
+if(NOT CMAKE_BUILD_TYPE AND NOT GENERATOR_IS_MULTI_CONFIG)
```
real-or-random, line 185:
```suggestion
if(NOT GENERATOR_IS_MULTI_CONFIG)
```
https://cmake.org/cmake/help/latest/prop_gbl/GENERATOR_IS_MULTI_CONFIG.html says "Multi-config generators (...) ignore `CMAKE_BUILD_TYPE`.", so I guess we should do the same?

hebasto: The `NOT GENERATOR_IS_MULTI_CONFIG` condition ensures that we use a single-config generator.

The `NOT CMAKE_BUILD_TYPE` condition checks whether we need to set the default build type. This is actually is the case when the user does not provide `-DCMAKE_BUILD_TYPE=...` on the command line.

hebasto: Going to update this branch shortly with a more readable implementation.

hebasto:
The commit "_cmake: Use dedicated `GENERATOR_IS_MULTI_CONFIG` property_" has been reworked entirely. Now the `GENERATOR_IS_MULTI_CONFIG` is treated correctly as a property, not as CMake variable.

CMakeLists.txt
hebasto: [Updated](https://github.com/bitcoin-core/secp256k1/pull/1239#issuecomment-1484070941).

hebasto:
Added commit "_cmake: Use `if(... IN_LIST ...)` command_".

hebasto:
Rebased 1341b4ba8c00ef7a4d182c0c5fa94d5e693f02e8 -> bf5a373e473d03821effc02c338cb0a425989074 ([pr1239.07](https://github.com/hebasto/secp256k1/commits/pr1239.07) -> [pr1239.08](https://github.com/hebasto/secp256k1/commits/pr1239.08)) due to the conflict with #1268.

hebasto:
Updated bf5a373e473d03821effc02c338cb0a425989074 -> a273d74b2ea1ef115a7e40fe89a64a6c744018c6 ([pr1239.08](https://github.com/hebasto/secp256k1/commits/pr1239.08) -> [pr1239.09](https://github.com/hebasto/secp256k1/commits/pr1239.09)):

- rebased due to the conflict with #1263
- add a new "cmake: Improve version comparison" commit suggested by @real-or-random on IRC

src/CMakeLists.txt
```diff
@@ -114,7 +114,7 @@ if(SECP256K1_INSTALL)
     NO_SET_AND_CHECK_MACRO
   )
   write_basic_package_version_file(${PROJECT_NAME}-config-version.cmake
-    COMPATIBILITY SameMajorVersion
+    COMPATIBILITY SameMinorVersion
```
real-or-random, line 117: note: If we'd follow semver strictly, then that's not true for versions pre 1.0.0. 

But I think it's good to add "SameMinorVersion" here. It seems that we have agreed on bumping minor for incompatible API changes even pre 1.0.0. (We bumped 0.2.0 to 0.3.0 for example because of this.)

(The rationale in semver is that you should release 1.0.0 as soon as people rely on the API. We don't do that either, so I think what we do makes sense in the end, even though we deviate from semver.)

real-or-random (approved):
ACK a273d74b2ea1ef115a7e40fe89a64a6c744018c6

real-or-random:
(Not this PR:) This is not rejected: `cmake -B build -DCMAKE_BUILD_TYPE=yolo`. Maybe it should be rejected.

theuni (approved):
ACK a273d74b2ea1ef115a7e40fe89a64a6c744018c6

src/CMakeLists.txt
```diff
@@ -141,7 +141,7 @@ configure_package_config_file(
   NO_SET_AND_CHECK_MACRO
 )
 write_basic_package_version_file(${PROJECT_NAME}-config-version.cmake
-  COMPATIBILITY SameMajorVersion
+  COMPATIBILITY SameMinorVersion
```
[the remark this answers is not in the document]
real-or-random: see https://github.com/bitcoin-core/secp256k1/pull/1239#discussion_r1173643595

PR #1241, opened 2023-03-13
build: Improve `SECP_TRY_APPEND_DEFAULT_CFLAGS` macro

hebasto:
While the `SECP_TRY_APPEND_DEFAULT_CFLAGS` macro works with the current compiler flag set, it is not perfect.

For example, it will accept `-fvisibility-inlines-hidden` flag, which actually causes:
```
cc1: warning: command-line option ‘-fvisibility-inlines-hidden’ is valid for C++/ObjC++ but not for C
```

This PR improves robustness of the `SECP_TRY_APPEND_DEFAULT_CFLAGS` macro.

FWIW, the same approach is used in our [CMake-based](https://github.com/bitcoin-core/secp256k1/blob/427bc3cdcfbc74778070494daab1ae5108c71368/cmake/TryAddCompileOption.cmake#L9) build system, and in [Bitcoin Core](https://github.com/bitcoin/bitcoin/blob/f088949fcfe6ba5000caa2f6adc6803e81925afb/configure.ac#L437-L472).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -88,11 +88,14 @@ esac
 AC_DEFUN([SECP_TRY_APPEND_DEFAULT_CFLAGS], [
     # GCC and compatible (incl. clang)
     if test "x$GCC" = "xyes"; then
-      # Try to append -Werror=unknown-warning-option to CFLAGS temporarily. Otherwise clang will
-      # not error out if it gets unknown warning flags and the checks here will always succeed
-      # no matter if clang knows the flag or not.
+      # Try to append -Werror to CFLAGS temporarily. Otherwise checks for some unsupported
+      # flags will succeed.
+      # Note that failure to append -Werror does not necessarily mean that -Werror is not
+      # supported. The compiler may already be warning about something unrelated, for example
+      # about some path issue. If that is the case, -Werror cannot be used because all
+      # of those warnings would be turned into errors.
       SECP_TRY_APPEND_DEFAULT_CFLAGS_saved_CFLAGS="$CFLAGS"
-      SECP_TRY_APPEND_CFLAGS([-Werror=unknown-warning-option], CFLAGS)
+      SECP_TRY_APPEND_CFLAGS([-Werror], CFLAGS)
 
       SECP_TRY_APPEND_CFLAGS([-std=c89 -pedantic -Wno-long-long -Wnested-externs -Wshadow -Wstrict-prototypes -Wundef], $1) # GCC >= 3.0, -Wlong-long is implied by -pedantic.
       SECP_TRY_APPEND_CFLAGS([-Wno-overlength-strings], $1) # GCC >= 4.2, -Woverlength-strings is implied by -pedantic.


fanquake:
Looks like the comment needs updating as well? 
What happens after this change, with (supported versions of) Clang?

hebasto:
> What happens after this change, with (supported versions of) Clang?

I've tested clang-8 and clang-14. The only diff in `configure` script output looks like that:
```diff
--- /home/hebasto/config-master
+++ /home/hebasto/config-pr1241
@@ -74,7 +74,7 @@
 checking if libtool supports shared libraries... yes
 checking whether to build shared libraries... yes
 checking whether to build static libraries... yes
-checking if clang-14 supports -Werror=unknown-warning-option... yes
+checking if clang-14 supports -Werror... yes
 checking if clang-14 supports -std=c89 -pedantic -Wno-long-long -Wnested-externs -Wshadow -Wstrict-prototypes -Wundef... yes
 checking if clang-14 supports -Wno-overlength-strings... yes
 checking if clang-14 supports -Wall... yes
```

hebasto:
Updated cac8fa4cd52c6086949a84cd6916ec15f111214a -> 686fa5d9ebbafdfc8f7fe41fdbbb0f8e46d96d65 ([pr1241.01](https://github.com/hebasto/secp256k1/commits/pr1241.01) -> [pr1241.02](https://github.com/hebasto/secp256k1/commits/pr1241.02), [diff](https://github.com/hebasto/secp256k1/compare/pr1241.01..pr1241.02)).

Addressed @fanquake's [comment](https://github.com/bitcoin-core/secp256k1/pull/1241#issuecomment-1466624736):
> Looks like the comment needs updating as well?

real-or-random (commented):
Concept ACK

I think the comment could also point out that `-Werror` may in principle be supported but warnings are already present (
https://github.com/bitcoin/bitcoin/blob/f088949fcfe6ba5000caa2f6adc6803e81925afb/configure.ac#L344-L347). That's rather subtle.

hebasto:
@real-or-random 
> I think the comment could also point out that `-Werror` may in principle be supported but warnings are already present

Mind suggesting a wording for the comment, please?

real-or-random:
> Mind suggesting a wording for the comment, please?

"Note that failure to append -Werror does not necessarily mean that -Werror is not supported. The compiler may already be warning about something unrelated, for example about some path issue. If that is the case, -Werror cannot be used because all of those warnings would be turned into errors."

real-or-random (approved):
utACK 3addb4c1e8a50df7dcf4465a7f149f78bf5af78b

jonasnick (commented):
ACK 3addb4c1e8a50df7dcf4465a7f149f78bf5af78b

theuni:
Following up on this, I didn't realize this existed: [CMAKE_COMPILE_WARNING_AS_ERROR](https://cmake.org/cmake/help/latest/variable/CMAKE_COMPILE_WARNING_AS_ERROR.html). It'd be nice to opt-in to that behavior for v3.24+, or at least make note of it.

real-or-random:
> Following up on this, I didn't realize this existed: [CMAKE_COMPILE_WARNING_AS_ERROR](https://cmake.org/cmake/help/latest/variable/CMAKE_COMPILE_WARNING_AS_ERROR.html). It'd be nice to opt-in to that behavior for v3.24+, or at least make note of it.

Added to the list in #1224.

alokeutpal (approved):

PR #1242, opened 2023-03-13
Set ARM ASM symbol visibility to `hidden`

hebasto:
Solves one item in #1181.

To test on arm-32bit hardware, run:
```
$ ./autogen.sh && ./configure --enable-experimental --with-asm=arm && make
```

On master branch (427bc3cdcfbc74778070494daab1ae5108c71368):
```
$ nm -D .libs/libsecp256k1.so | grep secp256k1_fe
0000e2bc T secp256k1_fe_mul_inner
0000e8dc T secp256k1_fe_sqr_inner
```

With this PR:
```
$ nm -D .libs/libsecp256k1.so | grep secp256k1_fe | wc -l
0
```

For reference, see https://sourceware.org/binutils/docs/as/Hidden.html.

### DIFF
--- a/src/asm/field_10x26_arm.s
+++ b/src/asm/field_10x26_arm.s
@@ -29,6 +29,7 @@ Note:
 	.align	2
 	.global secp256k1_fe_mul_inner
 	.type	secp256k1_fe_mul_inner, %function
+	.hidden secp256k1_fe_mul_inner
 	@ Arguments:
 	@  r0  r      Restrict: can overlap with a, not with b
 	@  r1  a
@@ -516,6 +517,7 @@ secp256k1_fe_mul_inner:
 	.align	2
 	.global secp256k1_fe_sqr_inner
 	.type	secp256k1_fe_sqr_inner, %function
+	.hidden secp256k1_fe_sqr_inner
 	@ Arguments:
 	@  r0  r	 Can overlap with a
 	@  r1  a


theuni:
Mmm, concept NACK. This is relying on a side-effect of linker behavior rather than avoiding exporting them in the first place. The problem is that the visibility flag doesn't get set for asm.

 This should really be:
```cmake
set(CMAKE_ASM_VISIBILITY_PRESET hidden)
```

However cmake doesn't understand that :(

Instead I guess we should add the flag ourselves after testing that it'll work?

theuni:
Proposed alternative: just always make it hidden.

Since this is already a special-cased platform using gas syntax, I don't see any problem with adding another directive.

```patch
diff --git a/src/asm/field_10x26_arm.s b/src/asm/field_10x26_arm.s
index 5f68cefc..f7cc1f26 100644
--- a/src/asm/field_10x26_arm.s
+++ b/src/asm/field_10x26_arm.s
@@ -29,6 +29,7 @@ Note:
        .align  2
        .global secp256k1_fe_mul_inner
        .type   secp256k1_fe_mul_inner, %function
+       .hidden secp256k1_fe_mul_inner
        @ Arguments:
        @  r0  r      Restrict: can overlap with a, not with b
        @  r1  a
@@ -516,6 +517,7 @@ secp256k1_fe_mul_inner:
        .align  2
        .global secp256k1_fe_sqr_inner
        .type   secp256k1_fe_sqr_inner, %function
+       .hidden secp256k1_fe_sqr_inner
        @ Arguments:
        @  r0  r         Can overlap with a
        @  r1  a
```

hebasto:
Implemented @theuni's [suggestion](https://github.com/bitcoin-core/secp256k1/pull/1242#issuecomment-1468901366).

Thank you!

theuni:
Thanks! I've tested this locally and it indeed gets the symbols hidden.

Since all the other tools seem to consider it non-standard to pass visibility flags to an asm file, rather than trying to force `-fvisibility=hidden` in tooling that doesn't want it, I think it's perfectly reasonable to embed the property in the file instead.

ACK fd2a408647ba0f999b7b217977cc68773fa35257.

theuni (approved):

PR #1243, opened 2023-03-14
build: Ensure no optimization when building for coverage analysis

hebasto:
#944 introduced a regression when building for coverage analysis. The `-O2` flag from the default Autoconf's `CFLAGS` overrides the coverage-specific `-O0` one, which makes coverage analysis results [less reliable](https://gcc.gnu.org/onlinedocs/gcc/Gcov-and-Optimization.html).

This PR restores the pre-#944 behaviour.

In contrast to an alternative smaller diff:
```diff
--- a/configure.ac
+++ b/configure.ac
@@ -240,7 +240,7 @@ fi
 
 if test x"$enable_coverage" = x"yes"; then
     SECP_CONFIG_DEFINES="$SECP_CONFIG_DEFINES -DCOVERAGE=1"
-    SECP_CFLAGS="-O0 --coverage $SECP_CFLAGS"
+    CFLAGS="$CFLAGS -O0 --coverage "
     LDFLAGS="--coverage $LDFLAGS"
 else
     # Most likely the CFLAGS already contain -O2 because that is autoconf's default.
```

this PR ensures that the user always has the last word.

FWIW, Bitcoin Core uses a similar [approach](https://github.com/bitcoin/bitcoin/blob/460e394625fab2942748aaeec9be31f460f91c58/configure.ac#L879-L884).

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -29,6 +29,11 @@ AM_INIT_AUTOMAKE([1.11.2 foreign subdir-objects])
 # Make the compilation flags quiet unless V=1 is used.
 m4_ifdef([AM_SILENT_RULES], [AM_SILENT_RULES([yes])])
 
+if test "${CFLAGS+set}" = "set"; then
+  CFLAGS_overridden=yes
+else
+  CFLAGS_overridden=no
+fi
 AC_PROG_CC
 AM_PROG_AS
 AM_PROG_AR
@@ -241,6 +246,12 @@ fi
 if test x"$enable_coverage" = x"yes"; then
     SECP_CONFIG_DEFINES="$SECP_CONFIG_DEFINES -DCOVERAGE=1"
     SECP_CFLAGS="-O0 --coverage $SECP_CFLAGS"
+    # If coverage is enabled, and the user has not overridden CFLAGS,
+    # override Autoconf's value "-g -O2" with "-g". Otherwise we'd end up
+    # with "-O0 --coverage -g -O2".
+    if test "$CFLAGS_overridden" = "no"; then
+      CFLAGS="-g"
+    fi
     LDFLAGS="--coverage $LDFLAGS"
 else
     # Most likely the CFLAGS already contain -O2 because that is autoconf's default.


real-or-random (commented):
Concept ACK

Good catch!

I first wondered if simply setting `-g` will result in the same problem as [here](https://github.com/bitcoin-core/secp256k1/issues/896#issuecomment-780747007). To be honest, I don't remember exactly how this caused a problem on mac.

But since you use a Mac, and since Core does the same, I assume your PR here does the right thing?

hebasto:
> I first wondered if simply setting `-g` will result in the same problem as [here](https://github.com/bitcoin-core/secp256k1/issues/896#issuecomment-780747007). To be honest, I don't remember exactly how this caused a problem on mac.

The mentioned problem can be easily inducted by `./configure CFLAGS=-g` on macOS. But that is the user's responsibility :)

https://lists.nongnu.org/archive/html/bug-autoconf/2007-11/msg00032.html:
>     So the Autoconf test that checks for -g support creates this directory
>     which is not removed by most rm -f conftest.% because most of the time
>     the % is explicitly given (e.g., that test runs "rm -f core *.core
>     core.conftest.* gmon.out bb.out conftest$ac_exeext conftest.$ac_objext
>     conftest.$ac_ext").  Other tests simply go for "rm -f conftest.*" and
>     fail on conftest.dSYM being a directory.


> But since you use a Mac

Not my main machine :)

> and since Core does the same

Not "the same", but similar. Bitcoin Core clears `CFLAGS`. This PR restores flags as they were before #944.

jonasnick (commented):
I remember getting quite unexpected output in a coverage report recently. This may have very well been the issue. Thanks!

tested ACK 8e79c7ed11fa50bd6b8a3d3203b2fc330a0c37ea

real-or-random (approved):
utACK 8e79c7ed11fa50bd6b8a3d3203b2fc330a0c37ea

alokeutpal (approved):

Issue #1106, opened 2022-05-11
Wycheproof tests vectors


real-or-random:
Though we have good tests, the [Wycheproof](https://github.com/google/wycheproof) test vectors will be a good addition.

- The ECDSA test vectors are probably most interesting but some test vectors are incompatible with Bitcoin (see https://github.com/google/wycheproof/issues/70). We asked for other vectors there but they don't respond currently. But even if we disable the incompatible vectors (after careful checking), the other vectors will be a good addition 
- Not sure if we can ECDH to work but I expect it will work at least with a custom derivation function
- SHA256 and HMAC should not be an issue

The test cases are in a simple JSON format: https://github.com/google/wycheproof/tree/master/testvectors We could either parse them (e.g., https://github.com/zserge/jsmn is very thin, MIT single-header parser that we ), or have a small script (in whatever language) that converts the vectors to C code.

andozw:
We opened a PR that adds coverage for ECDSA Wycheproof test vectors at https://github.com/bitcoin-core/secp256k1/pull/1245 . We wrote a small Python script that parses the JSON and outputs C data structure, and embedded it in the source file (much like it is done elsewhere). All the tests for ECDSA pass from the latest drop of Wycheproof (released about 3 weeks ago).

jonasnick:
Done in #1245 and #1492.

PR #1133, opened 2022-08-11
schnorrsig: Add test vectors for variable-length messages 

real-or-random:
Related: https://github.com/sipa/bips/pull/221

PR #1245, opened 2023-03-22
tests: Add Wycheproof ECDSA vectors

RandomLattice:
This PR adds a test using the Wycheproof vectors as outlined in #1106. We add all 463 ECDSA test vectors. These vectors cover:
- edge cases in arithmetic operations
- signatures with special values for (r,s) that should be rejected
- special cases of public keys

The vectors are pulled from the Wycheproof project using a python script to emit C code.

All the new ECDSA Wycheproof vectors pass.


Sajjon:
IMO it is a mistake to have the unit test consume test vectors as C code, because we miss out on 50% of the real value add - universally portable test vectors.

This is such a fundamental library that so many DLT project rely on, many of which through wrappers in higher abstraction languages, e.g. my own [Swift Wrapper _K1_](https://github.com/sajjon/K1). Thus it will be an enormous value add for all these wrappers and other secp256k1 libraries to be able to copy over a JSON file in full. 

Thus I think it is a mistake to not focus on the portability benefits of putting test vectors in a file, most preferably JSON.

Second best option is putting the test vector in a file, but on another format than JSON, eg raw binary or CBOR, because we still get the portability advantage, but less easy accessible for other language.

What so you all think?

apoelstra:
I don't have any opinion on one file vs many, but regarding high-level approach, my vote is

* Check in both the source .json and output .c files
* Check in CI (not in the normal build process!) that the checked-in .c files can be regenerated and are identical to the checked-in ones.

apoelstra:
> What do you think it’s best for this? Would a line in ci/cirrus.sh work?

That sounds great to me. Though bear in mind that I'm not a maintainer of this library (anymore) and I'm not the one to make such calls.

real-or-random:
> What do you think it’s best for this? Would a line in `ci/cirrus.sh` work?

We already have that line because we check the same with the "precomputed" C files, see:
https://github.com/bitcoin-core/secp256k1/blob/9c8c4f443c0027c3ed933a3883497131071b1c7e/ci/cirrus.sh#L109-L118

We could add make targets `testvectors` and `clean-testvectors` (or with a different name) similar to the existing ones for "precomputed" (see the Makefile) that create the C files from JSON. If we then call `make clean-testvectors` and `make testvectors` in the CI script, the already existing `git diff --exit-code` will catch any need to regenerate.

CyonAlexRDX (approved):
(Used wrong GitHub user)

real-or-random (commented):
Awesome. I have only looked at the diff so far, and I just have a bunch of minor comments, mostly related to consistent naming.

---

One major issue is licensing: It's not entirely clear if the test vectors are copyrightable, so let's be conservative and assume they are. Then Wycheproof's Apache license applies, and we have to adhere to it, see https://www.apache.org/licenses/LICENSE-2.0 Section 4. This means we should probably create a subdir `src/wycheproof` with
 - the `.json` file
 - the generated `.h` file
 - a `COPYING` file that says that the `.json` files in this directory have been taken from Google's project Wycheproof with commit id (you can then remove the comment in your python file), that the C files are generated from the .json files using the script. And the file should include the Apache 2.0 license text (i.e., the contents of https://github.com/google/wycheproof/blob/master/LICENSE )

The script can then go to a new directory `tools` in the root of the repo (as suggested in other unmerged PRs too): https://github.com/bitcoin-core/secp256k1/pull/1135#issuecomment-1335398197

tools/tests_wycheproof_generate.py
```diff
@@ -0,0 +1,66 @@
...
+    doc = json.load(f)
+
+num_groups = len(doc['testGroups'])
+
+def to_c_array(x):
+    if x == "": return "{0}"
+    s = ',0x'.join(a+b for a,b in zip(x[::2], x[1::2]))
+    return "{ 0x" + s + " }"
+
+num_cases, max_msg_size, max_sig_size = 0, 0, 0
```
real-or-random, line 18:
```suggestion
num_vectors, max_msg_size, max_sig_size = 0, 0, 0
```

tools/tests_wycheproof_generate.py
```diff
@@ -0,0 +1,66 @@
...
+} wycheproof_ecdsa_testvector;
+"""
+
+
+print("/* Note: this file was autogenerated using tests_wycheproof_generate.py. Do not edit. */")
+print("#define SECP256K1_ECDSA_WYCHEPROOF_NUMBER_TESTVECTORS ({})".format(num_cases))
+print("#define SECP256K1_ECDSA_WYCHEPROOF_MAX_MSG_SIZE ({})".format(max_msg_size))
+print("#define SECP256K1_ECDSA_WYCHEPROOF_MAX_SIG_SIZE ({})".format(max_sig_size))
+print(struct_definition)
+print("static const wycheproof_ecdsa_testvector testcases[SECP256K1_ECDSA_WYCHEPROOF_NUMBER_TESTVECTORS] = {")
```
real-or-random, line 64:
```suggestion
print("static const wycheproof_ecdsa_testvector testvectors[SECP256K1_ECDSA_WYCHEPROOF_NUMBER_TESTVECTORS] = {")
```

tools/tests_wycheproof_generate.py
```diff
@@ -0,0 +1,66 @@
...
+
+num_cases, max_msg_size, max_sig_size = 0, 0, 0
+out = ""
+
+for i in range(num_groups):
+    group = doc['testGroups'][i]
+    num_tests = len(group['tests'])
+    public_key = group['publicKey']
+    for j in range(num_tests):
+        test_case = group['tests'][j]
```
real-or-random, line 26:
```suggestion
        test_vector = group['tests'][j]
```

src/wycheproof/WYCHEPROOF_COPYING
```diff
@@ -0,0 +1,212 @@
+* The file `ecdsa_secp256k1_sha256_bitcoin_test.json` in this directory
+  comes from Google's project Wycheproof with git commit
+  `b063b4aedae951c69df014cd25fa6d69ae9e8cb9`.
+  The Wycheproof project can be found at https://github.com/google/wycheproof
```
real-or-random, line 4: In the spirit of @Sajjon 's suggestion:

```suggestion
* The file `ecdsa_secp256k1_sha256_bitcoin_test.json` in this directory
  comes from Google's project Wycheproof with git commit
  `b063b4aedae951c69df014cd25fa6d69ae9e8cb9`, see
  https://github.com/google/wycheproof/blob/b063b4aedae951c69df014cd25fa6d69ae9e8cb9/testvectors_v1/ecdsa_secp256k1_sha256_bitcoin_test.json .
```

I think this is useful, I first looked in the wrong directory (`testvectors` instead of the new `testvectors_v1`...).

real-or-random (approved):
Thanks for the quick fixups!

> I put the license file in `src/wycheproof/WYCHEPROOF_COPYING`. Like this, the name does not clash with the top level `COPYING` file (which is a different license). This is the same approach as with `examples/EXAMPLES_COPYING`.

Good catch!


Here are some more comments, but this is good to go from my side, with or without these addressed.

ACK 04c70ffacc04ef120d8c700a9e1e650bde0e2570

real-or-random (approved):
ACK f47c00812afa9a0a7297fb9f381863ab7cd4c413

Bezzo100-lab:
@Bezzoxoxo

Sajjon:
> This encoding using maximum-length padded arrays for all inputs does result in about 2 MiB of redundant zeroes in the binaries (and even more in the checked-in version of the .h file).
> 
> 
> 
> What do you think about an encoding using separately declared array constants for all the messages/signatures, pointed to by the main `wycheproof_ecdsa_testvector` struct elements?

Then we lose the other important property which is portability. 

Fine for the wycheproof vectors since we are not the original source, the Wycheproof github repo is. 

But if we generate and vendor our own test vectors for other libraries to use, it would be great if those we portable, eg in JSON format to be used as is.

What are your thoughts on that?

I would eg love to see a standardization around serialization of recoverable signatures. Some libraries chose (using `V` for `recid`) format: `V || R || S` and some use `R || S || V`, and there exist few test vectors for recovery on interwebz. So would be great if libsecp256k1 could vendor some, and on a portable format.

tools/tests_wycheproof_generate.py
```diff
@@ -0,0 +1,91 @@
...
+    size_t sig_len;
+    int expected_verify;
+} wycheproof_ecdsa_testvector;
+"""
+
+
+print("/* Note: this file was autogenerated using tests_wycheproof_generate.py. Do not edit. */")
+print("#define SECP256K1_ECDSA_WYCHEPROOF_NUMBER_TESTVECTORS ({})".format(num_vectors))
+print("#define SECP256K1_ECDSA_WYCHEPROOF_MAX_MSG_SIZE ({})".format(max_msg_size))
+print("#define SECP256K1_ECDSA_WYCHEPROOF_MAX_SIG_SIZE ({})".format(max_sig_size))
```
real-or-random, line 82: these two are unused now

tools/tests_wycheproof_generate.py
```diff
@@ -0,0 +1,91 @@
...
+        msg_size = len(test_vector['msg']) // 2
+        max_sig_size = max(sig_size, max_sig_size)
+        max_msg_size = max(msg_size, max_msg_size)
+
+        if test_vector['result'] == "invalid": expected_verify = 0
+        elif test_vector['result'] == "valid": expected_verify = 1
+        else: raise ValueError("invalid result field")
+
+        if num_vectors != 0 and msg_size != 0: messages += ", "
+        if num_vectors != 0 and sig_size != 0: signatures += ", "
```
real-or-random, line 50:
```suggestion
        if num_vectors != 0 and msg_size != 0: messages += ",\n"
        if num_vectors != 0 and sig_size != 0: signatures += ",\n"
```
This will result in much more readable output. And we should avoid very long lines. Some tools don't like them, e.g., some editors become very slow.

real-or-random:
> @Sajjon That seems unrelated to this PR, but I see no reason why any such portable test vectors couldn't use a similar approach. 

Indeed. If we want to create portable test vectors, we could still create them in JSON and add have another conversion script. 

> Following this PR, I think perhaps we may want to also add Python code to e.g. convert the BIP340 CSV test vectors to C source code, rather than just hardcoding.

It's cleaner, but now that we have them in C, I don't think it's worth the effort. They won't change often. (Well, https://github.com/bitcoin-core/secp256k1/pull/1133 adds some but even this PR is ready already...)

tools/tests_wycheproof_generate.py
real-or-random, line 1: nit: maybe make that file executable

real-or-random (approved):
ACK 2d6f94bc5eb4d8ed9ff72c3ef3d98dc10e86f414

real-or-random (approved):
ACK e5de45460953c8ae16521b1928ac14de218998a3

alokeutpal (approved):

PR #1247, opened 2023-03-23
Apply Checks only in VERIFY mode.

roconnor-blockstream:
This is already done in `field_5x52_impl.h`.


jonasnick:
If the goal is to make the two files consistent, wouldn't it be better to remove the `ifdef` from `field_5x52_impl.h`?

roconnor-blockstream:
We generally do wrap non-trivial `VERIFY_CHECK`s, particularly when they call functions.

jonasnick (commented):
You're right.

ACK 4ebd82852d3ad00ab579b26173575a4f4642ea76

roconnor-blockstream:
I do admit, with so many exceptions, the whole idea of enabling `VERIY_CHECK`s in release builds feels of questionable value.

alokeutpal (approved):

PR #1250, opened 2023-03-27
No need to subtract 1 before doing a right shift

roconnor-blockstream:


jonasnick (commented):
ACK 3e43041be68c1288ad9897525a15e21945fb3eb9

PR #1251, opened 2023-03-28
cmake: Use "coverage" preset instead of "Coverage" build type and document it

hebasto:
Using `-DCMAKE_BUILD_TYPE=Coverage` to configure the build for coverage analysis has a flow as build types are not defined at the configuration stage for multi-config generators.

This PR implements the idea suggested [here](https://github.com/bitcoin-core/secp256k1/issues/1235#issuecomment-1484045062).

Please note that both added to the `README.md` commands, `cmake --build build` and `ctest --test-dir build`, accept the `-j $(nproc)` option to parallelize work.

Drafted as it is based on https://github.com/bitcoin-core/secp256k1/pull/1234.


CMakePresets.json
```diff
@@ -18,6 +18,17 @@
         "dev": true,
         "uninitialized": true
       }
+    },
+    {
+      "name": "coverage",
+      "displayName": "Build for coverage analysis",
+      "generator": "Unix Makefiles",
+      "cacheVariables": {
+        "CMAKE_C_COMPILER": "gcc",
```
real-or-random, line 27: https://clang.llvm.org/docs/SourceBasedCodeCoverage.html clang supports this too, so no need to set a default here, I think.

hebasto: Doesn't clang require the use of the llvm-cov tool, which is currently undocumented in the README.md?

theuni:
Concept ACK assuming one can stack multiple presets at once.

theuni:
After a discussion about this in-person, we've found drawbacks in the build type approach as well as the preset one.

Instead, let's just do the simple thing and hook up a new option.

hebasto:
> After a discussion about this in-person, we've found drawbacks in the build type approach as well as the preset one.
> 
> Instead, let's just do the simple thing and hook up a new option.

Closing in favour of #1291.

hebasto:
> > After a discussion about this in-person, we've found drawbacks in the build type approach as well as the preset one.
> > Instead, let's just do the simple thing and hook up a new option.
> 
> Closing in favour of #1291.

Another reincarnation of this approach is [here](https://github.com/hebasto/secp256k1/tree/250810-coverage-preset).

It works with both single-config and multi-config generators.

The following are examples workflow on Ubuntu 24.04:
- a single-config generator:
```
cmake --preset coverage -DCMAKE_GENERATOR="Ninja"
cmake --build --preset coverage
ctest --preset coverage
```
- a multi-config generator:
```
cmake --preset coverage -DCMAKE_GENERATOR="Ninja Multi-Config"
cmake --build --preset coverage
ctest --preset coverage
```

In the future, with CMake 3.25 or newer, we can use a single invocation:
```
cmake --workflow --preset coverage
```

PR #1252, opened 2023-03-28
Make position of * in pointer declarations in include/ consistent

jonasnick:


apoelstra:
I think we should move it the other direction, so the `*` is part of the type and not part of the name.

jonasnick:
I went with the traditional convention as used in the C standard. I'd agree with you if the position of `*` was entirely arbitrary. But making `*` part of the name makes more sense in C because it applies to the variable and not the type. For example, `char* foo, bar;` declares `foo` to be a pointer to char and `bar` to be a char.

real-or-random (approved):
I think they're arguments for either side. But yeah, we should make it consistent.

utACK https://github.com/bitcoin-core/secp256k1/pull/1252/commits/3d1f430f9f32d45885b0a10b448c0f15386c423d

I would ACK the other variant as well.

hebasto:
> I think we should move it the other direction, so the `*` is part of the type and not part of the name.

+1 on that. Especially in the case when this repo is a subtree in other C++ project (like Bitcoin Core), which uses the same convention.

apoelstra:
utACK 3d1f430 from me too. I also value consistency more than either specific choice.'

PR #1263, opened 2023-04-09
cmake: Make installation optional

CyberTailor:
Useful for embedding secp256k1 in a subproject.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,8 @@ if(SECP256K1_DISABLE_SHARED)
   set(BUILD_SHARED_LIBS OFF)
 endif()
 
+option(SECP256K1_INSTALL "Enable installation" ON)
+
 option(SECP256K1_ENABLE_MODULE_ECDH "Enable ECDH module." ON)
 if(SECP256K1_ENABLE_MODULE_ECDH)
   add_definitions(-DENABLE_MODULE_ECDH=1)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -73,51 +73,54 @@ if(SECP256K1_BUILD_CTIME_TESTS)
   target_link_libraries(ctime_tests secp256k1)
 endif()
 
-install(TARGETS secp256k1
-  EXPORT ${PROJECT_NAME}-targets
-  RUNTIME DESTINATION ${CMAKE_INSTALL_BINDIR}
-  LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
-  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
-)
-set(${PROJECT_NAME}_headers
-  "${PROJECT_SOURCE_DIR}/include/secp256k1.h"
-  "${PROJECT_SOURCE_DIR}/include/secp256k1_preallocated.h"
-)
-if(SECP256K1_ENABLE_MODULE_ECDH)
-  list(APPEND ${PROJECT_NAME}_headers "${PROJECT_SOURCE_DIR}/include/secp256k1_ecdh.h")
-endif()
-if(SECP256K1_ENABLE_MODULE_RECOVERY)
-  list(APPEND ${PROJECT_NAME}_headers "${PROJECT_SOURCE_DIR}/include/secp256k1_recovery.h")
-endif()
-if(SECP256K1_ENABLE_MODULE_EXTRAKEYS)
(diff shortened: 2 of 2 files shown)


real-or-random:
cc @hebasto @theuni

hebasto:
> Useful for embedding secp256k1 in a subproject.

I've been thinking on the same. Just waiting with a PR as "cmake"-PR pile looks quite big for now :)

real-or-random:
What's the rationale behind this? I guess if you're not using it as a subproject, you can simply omit `make install`, but that doesn't work in for subproject because there's only a single `make` invocation for the parent project?

CyberTailor:
> What's the rationale behind this?

To call `add_subdirectory()` just for the static library, not installing it after.

> So if you `make install` the parent project, you'd automatically get `make install` here?

Yes, unless you use `EXCLUDE_FROM_ALL` option, but in that case you wouldn't also get tests.

hebasto:
FWIW, I've been maintaining a [repo](https://github.com/hebasto/secp256k1-CMake-example) with usage examples since reviewing of https://github.com/bitcoin-core/secp256k1/pull/1113.

There is a ["subtree"](https://github.com/hebasto/secp256k1-CMake-example/tree/subtree) branch there.

There are two options to include a subtree:

- `add_subdirectory(src/secp256k1)` includes all subtree's targets into a project's default ones, including tests and an install target.
- `add_subdirectory(src/secp256k1 EXCLUDE_FROM_ALL)` does not include a subtree install target into a project's default ones. However, the parent project can call `install(TARGETS secp256k1)`.

Here is an [example](https://github.com/google/leveldb/blob/77d66aaf3edd6db0f1bfb4308fdc68c62fcc7bf6/CMakeLists.txt#L35) of a similar option in Google's LevelDB project:
```CMake
option(LEVELDB_INSTALL "Install LevelDB's header and library" ON)
```

theuni:
I believe [PROJECT_IS_TOP_LEVEL](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html) and friends were introduced to help with this type of logic?

> A boolean variable indicating whether the most recently called [project()](https://cmake.org/cmake/help/latest/command/project.html#command:project) command in the current scope or above was in the top level CMakeLists.txt file.

hebasto:
> I believe [PROJECT_IS_TOP_LEVEL](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html) and friends were introduced to help with this type of logic?

Indeed. For CMake 3.21+.

real-or-random:
Concept ACK

> > I believe [PROJECT_IS_TOP_LEVEL](https://cmake.org/cmake/help/latest/variable/PROJECT_IS_TOP_LEVEL.html) and friends were introduced to help with this type of logic?
> 
> Indeed. For CMake 3.21+.

Would it be a good idea to make the default depend on `PROJECT_IS_TOP_LEVEL` if the cmake version is at least 3.21?

real-or-random:
and needs rebase :)

real-or-random (approved):
utACK 47ac3d63cd5e00a2d50cb489461c8bc349d37912

hebasto (approved):
ACK 47ac3d63cd5e00a2d50cb489461c8bc349d37912, tested on Ubuntu 23.04.

real-or-random:
> > Would it be a good idea to make the default depend on PROJECT_IS_TOP_LEVEL if the cmake version is at least 3.21?
> 
> Would prefer to see this done here, but no need to block this PR I guess.


To be honest, my current policy is to merge everything that has enough ACKs. We see stalled PRs way too often in this repo (I don't know if it's worse than in Core, but it's annoying.)  I'll add this as an item to #1235.

@CyberTailor Are you willing to work on this? :)

CyberTailor:
Later, when 3.21 could be safely set as the minimum supported version

hebasto:
> > Later, when 3.21 could be safely set as the minimum supported version
> 
> Oh, I think the suggestion is to wrap it in `if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.21)`

See #1284.

theuni:
> To be honest, my current policy is to merge everything that has enough ACKs. We see stalled PRs way too often in this repo (I don't know if it's worse than in Core, but it's annoying.) I'll add this as an item to #1235.

Thanks, this is really helpful. Will keep in mind while reviewing

hebasto:
FWIW, another option that could be considered for such a use case is [`CMAKE_SKIP_INSTALL_RULES`](https://cmake.org/cmake/help/latest/variable/CMAKE_SKIP_INSTALL_RULES.html).

PR #1265, opened 2023-04-10
Remove bits argument from secp256k1_wnaf_const{_xonly}

sipa:
There is little reason for having the number of bits in the scalar as a parameter, as I don't think there are any (current) use cases for non-256-bit scalars.

### DIFF
--- a/src/bench_ecmult.c
+++ b/src/bench_ecmult.c
@@ -113,7 +113,7 @@ static void bench_ecmult_const(void* arg, int iters) {
     int i;
 
     for (i = 0; i < iters; ++i) {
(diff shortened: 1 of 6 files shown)


real-or-random:
Concept ACK

> I don't think there are any (current) use cases for non-256-bit scalars.

Indeed, it seems unlikely that we'll ever have non-256-bit *private* scalars.

jonasnick (approved):
ACK a575339c0282ba49a4f46c9c660a4cc3b6bfc703

real-or-random (approved):
utACK a575339c0282ba49a4f46c9c660a4cc3b6bfc703

PR #1027, opened 2021-12-03
build: Add a check that Valgrind actually supports a host platform

hebasto:
This PR adds a check that Valgrind actually supports a host platform.

On master (49f608de474a89e5c742830c8ff5ac8978b5796a):
```
$ ./autogen.sh &> /dev/null && ./configure -q --host=riscv64-linux-gnu 2>&1 | grep valgrind
  valgrind                = yes
```

With this PR:
```
$ ./autogen.sh &> /dev/null && ./configure -q --host=riscv64-linux-gnu 2>&1 | grep valgrind
  valgrind                = no
```

Closes #1023.

PR #1273, opened 2023-04-11
build: Make `SECP_VALGRIND_CHECK` preserve `CPPFLAGS`

hebasto:
It was overlooked in #862 and #1027.


real-or-random (approved):
utACK 1ecb94ebe9800900c7dd3a4f9883c600e25eecf7

PR #1274, opened 2023-04-11
test: Silent noisy clang warnings about Valgrind code on macOS x86_64

hebasto:
Since #1206, on macOS x86_64 with Valgrind installed, clang emits a massive amount of `-Wreserved-identifier` and `-Wreserved-macro-identifier` warnings from the `valgrind/valgrind.h` and `valgrind/memcheck.h` headers.

This PR prevents warnings emitted for the Valgrind code.

### DIFF
--- a/src/checkmem.h
+++ b/src/checkmem.h
@@ -58,7 +58,14 @@
 #if !defined SECP256K1_CHECKMEM_ENABLED
 #  if defined VALGRIND
 #    include <stddef.h>
+#  if defined(__clang__) && defined(__APPLE__)
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wreserved-identifier"
+#  endif
 #    include <valgrind/memcheck.h>
+#  if defined(__clang__) && defined(__APPLE__)
+#    pragma clang diagnostic pop
+#  endif
 #    define SECP256K1_CHECKMEM_ENABLED 1
 #    define SECP256K1_CHECKMEM_UNDEFINE(p, len) VALGRIND_MAKE_MEM_UNDEFINED((p), (len))
 #    define SECP256K1_CHECKMEM_DEFINE(p, len) VALGRIND_MAKE_MEM_DEFINED((p), (len))


configure.ac
```diff
@@ -57,7 +57,7 @@ case $host_os in
          # Homebrew where each one is located, then adjust paths accordingly.
          if $BREW list --versions valgrind >/dev/null; then
            valgrind_prefix=$($BREW --prefix valgrind 2>/dev/null)
-           VALGRIND_CPPFLAGS="-I$valgrind_prefix/include"
+           VALGRIND_CPPFLAGS="-nostdinc -isystem $valgrind_prefix/include"
```
real-or-random, line 60: I guess there's a reason why you also add `-nostdinc` but I don't see it.

real-or-random (commented):
I wonder if anyone out there uses a compiler that doesn't understand `-isystem`. But writing a check seems overkill too... Maybe just check if this is gcc or clang (`"x$GCC" = "xyes"` should do it)?

Yet another alternative is to wrap the `#include` in pragmas.

configure.ac
hebasto: On macOS x86_64, the /usr/local/include is included by default. Packages been installed via Homebrew have alternative softlinks in it, in addition to $(brew --prefix package)/include.

fanquake:
Is it worth making further changes to Valgrind (via brew) on macOS? You haven't been able tp `brew install valgrind` on macOS for more than 2 years (https://github.com/Homebrew/homebrew-core/pull/67341).

real-or-random:
Good point. However, the fork https://github.com/LouisBrunner/valgrind-macos works (even on Ventura, see also https://github.com/bitcoin-core/secp256k1/issues/1151#issuecomment-1378782870) and can also be installed with brew. We did this a while on CI, so I think we should keep the code. And we should really bring c0ae48c9950a908b637bff27791fabbe2833c4a5 back, at least the valgrind part.

real-or-random:
Oh, indeed... Well, at least they're working on it. The maintainer is not very responsive, but he regularly pushes out updates. I think that's a reason to keep the macOS Valgrind support in the build system (plus it works on x86_64).

hebasto:
> Yet another alternative is to wrap the `#include` in pragmas.

I've taken this approach for the following reasons:
- It is self-documented.
- It is build system agnostic.

Here is an example of Valgrind usage on macOS Monterey 12.6.6 (x86_64):
```
% valgrind ./build/src/ctime_tests
==18400== Memcheck, a memory error detector
==18400== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==18400== Using Valgrind-3.21.0.GIT-lbmacos and LibVEX; rerun with -h for copyright info
==18400== Command: ./build/src/ctime_tests
==18400== 
--18400-- run: /usr/bin/dsymutil "./build/src/ctime_tests"
--18400-- run: /usr/bin/dsymutil "/Users/hebasto/secp256k1/build/src/libsecp256k1.2.0.2.dylib"
==18400== 
==18400== HEAP SUMMARY:
==18400==     in use at exit: 7,982 bytes in 165 blocks
==18400==   total heap usage: 169 allocs, 4 frees, 8,262 bytes allocated
==18400== 
==18400== LEAK SUMMARY:
==18400==    definitely lost: 4,160 bytes in 130 blocks
==18400==    indirectly lost: 0 bytes in 0 blocks
==18400==      possibly lost: 576 bytes in 2 blocks
==18400==    still reachable: 3,246 bytes in 33 blocks
==18400==         suppressed: 0 bytes in 0 blocks
==18400== Rerun with --leak-check=full to see details of leaked memory
==18400== 
==18400== For lists of detected and suppressed errors, rerun with: -s
==18400== ERROR SUMMARY: 0 errors from 0 contexts (suppressed: 122 from 21)
```

real-or-random (approved):
utACK 747ada35877d4392c453b7c7249465fb382125ea

real-or-random:
Can we get more at least one more Concept (N)ACK on maintaining Valgrind for MacOS further? (See the brief discussion above)

real-or-random:
Merging... I think it's obvious now that we want to keep Valgrind for macOS with the switch to GitHub Actions, where we'll have x86_64 macOS runners (see #1394).

fanquake:
> where we'll have x86_64 macOS runners (see https://github.com/bitcoin-core/secp256k1/pull/1394).

I might be missing something, but that combination (x86_64 & macos 13) isn't supported by Valgrind, or the fork mentioned above?

Also, have we followed up upstream about actually fixing the cause of the warnigns, if possible?

real-or-random:
> I might be missing something, but that combination ([x86_64 & macos 13](https://github.com/bitcoin-core/secp256k1/pull/1394/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR44)) isn't supported by Valgrind, or the fork mentioned above?

It is supported by the fork mentioned above, see the table at https://github.com/LouisBrunner/valgrind-macos#status, column amd64.


> Also, have we followed up upstream about actually fixing the cause of the warnings, if possible?

No, and I suspect it won't be fruitful in that case. I think the usual convention is that warnings should be off when compiling system headers. Or at least that only essential warnings should be enabled, and our `-Wreserved-identifier` and `-Wreserved-macro-identifier` are rather uncommon. This is why we have `-isystem` and the like.

 In this case here, the issue is that with the homebrew on macOS, the headers end up at a normal include path (`-I`) and not at a system include path (`-isystem`). So this is just about convincing clang to treat these includes as system headers. I think reporting to Valgrind (or the fork) won't make much sense. In theory, we could report to Apple clang and complain about their decision to treating `/usr/local/bin` as a normal include path, but I don't think anyone will care.

fanquake:
It's hard to know, given the PR doesn't actually outline what the warnings where, but I'm pretty sure the issue is that the valgrind/memcheck headers have identifiers starting with `__`, and any warnings are correct, regardless of anything (Apple) Clang, or macOS, or system include related.

i.e compiling this code:
```c
#include <valgrind.h>
#include <memcheck.h>

int main() { return 0;}
```
with clang on ubuntu: `clang test.c -I/usr/include/valgrind -Wreserved-macro-identifier -Wreserved-identifier`, outputs warnings as expected:
```bash
In file included from test.c:1:
/usr/include/valgrind/valgrind.h:74:9: warning: macro name is a reserved identifier [-Wreserved-macro-identifier]
#define __VALGRIND_H
        ^
/usr/include/valgrind/valgrind.h:91:9: warning: macro name is a reserved identifier [-Wreserved-macro-identifier]
#define __VALGRIND_MAJOR__    3
        ^
/usr/include/valgrind/valgrind.h:92:9: warning: macro name is a reserved identifier [-Wreserved-macro-identifier]
#define __VALGRIND_MINOR__    19
        ^
/usr/include/valgrind/valgrind.h:416:9: warning: macro name is a reserved identifier [-Wreserved-macro-identifier]
#define __SPECIAL_INSTRUCTION_PREAMBLE                            \
        ^
/usr/include/valgrind/valgrind.h:1666:9: warning: macro name is a reserved identifier [-Wreserved-macro-identifier]
#define __CALLER_SAVED_REGS /*"rax",*/ "rcx", "rdx", "rsi",       \
        ^
/usr/include/valgrind/valgrind.h:1724:11: warning: macro name is a reserved identifier [-Wreserved-macro-identifier]
#  define __FRAME_POINTER                                         \
          ^
In file included from test.c:2:
/usr/include/valgrind/memcheck.h:61:9: warning: macro name is a reserved identifier [-Wreserved-macro-identifier]
#define __MEMCHECK_H
        ^
/usr/include/valgrind/memcheck.h:103:7: warning: identifier '_VG_USERREQ__MEMCHECK_RECORD_OVERLAP_ERROR' is reserved because it starts with '_' followed by a capital letter [-Wreserved-identifier]
      _VG_USERREQ__MEMCHECK_RECORD_OVERLAP_ERROR 
      ^
8 warnings generated.
```

real-or-random:
> but I'm pretty sure the issue is that the valgrind/memcheck headers have identifiers starting with `__`,

Right.

> any warnings are correct, regardless of anything (Apple) Clang, or macOS, or system include related.

Okay, what do you suggest doing? Reporting it to Valgrind?

real-or-random:
Makes sense.

Note that while we try to be strict about this, even the C committee itself went a step back and reworked the "reserved identifier" stuff for C23. Now identifiers like `__VALGRIND_H` are only "potentially reserved", which just means "don't use them because future C version may make them reserved". This at least solves the (rather theoretical) problem that any program declaring `__VALGRIND_H` simply had UB by any literal reading of the standard. Now with C23, this is not UB anymore, but programmers are warned that their programs may have UB in future versions of C.

PR #1276, opened 2023-04-14
autotools: Don't regenerate Wycheproof header automatically

real-or-random:
This is a hot fix for https://github.com/bitcoin/bitcoin/pull/27445 .

--- 

Pregenerated files that we distribute should not have dependencies in Makefile.am. For rationale, see the comments about the precomputed table files.

See also https://github.com/bitcoin/bitcoin/pull/27445#issuecomment-1502994264 .

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -255,8 +255,8 @@ EXTRA_DIST += tools/tests_wycheproof_generate.py
 
 TESTVECTORS = src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.h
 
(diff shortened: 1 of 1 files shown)


hebasto (approved):
ACK 06c67dea9f6d46d3e24e810900fbb03045eae641

That's OK as we have dedicated targets `clean-testvectors` and `testvectors`.

Good for a hotfix, but it is still broken for out-of-source builds.

rkhadem4 (approved):

PR #1277, opened 2023-04-14
autotools: Clean up after adding Wycheproof

real-or-random:
Follow-up to https://github.com/bitcoin-core/secp256k1/pull/1245.

This builds on top of https://github.com/bitcoin-core/secp256k1/pull/1276. Let's only merge https://github.com/bitcoin-core/secp256k1/pull/1276 as a hotfix for the Core build.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -1,5 +1,3 @@
-.PHONY: clean-precomp precomp
-
 ACLOCAL_AMFLAGS = -I build-aux/m4
 
 # AM_CFLAGS will be automatically prepended to CFLAGS by Automake when compiling some foo
@@ -218,10 +216,10 @@ precomp: $(PRECOMP)
 # e.g., after `make maintainer-clean`).
 BUILT_SOURCES = $(PRECOMP)
 
-maintainer-clean-local: clean-precomp
-
+.PHONY: clean-precomp
 clean-precomp:
 	rm -f $(PRECOMP)
+maintainer-clean-local: clean-precomp
 
 ### Pregenerated test vectors
 ### (see the comments in the previous section for detailed rationale)
@@ -234,10 +232,10 @@ testvectors: $(TESTVECTORS)
 
 BUILT_SOURCES += $(TESTVECTORS)
 
-maintainer-clean-local: clean-testvectors
-
+.PHONY: clean-testvectors
 clean-testvectors:
 	rm -f $(TESTVECTORS)
+maintainer-clean-local: clean-testvectors
 
 ### Additional files to distribute
 EXTRA_DIST = autogen.sh CHANGELOG.md SECURITY.md
--- a/Makefile.am
+++ b/Makefile.am
@@ -203,7 +203,7 @@ precompute_ecmult_gen_LDADD = $(COMMON_LIB)
 # otherwise make's decision whether to rebuild them (even in the first
 # build by a normal user) depends on mtimes, and thus is very fragile.
 # This means that rebuilds of the prebuilt files always need to be
-# forced by deleting them, e.g., by invoking `make clean-precomp`.
+# forced by deleting them.
 src/precomputed_ecmult.c:
 	$(MAKE) $(AM_MAKEFLAGS) precompute_ecmult$(EXEEXT)
 	./precompute_ecmult$(EXEEXT)
@@ -224,6 +224,7 @@ clean-precomp:
 	rm -f $(PRECOMP)
 
 ### Pregenerated test vectors
+### (see the comments in the previous section for detailed rationale)
 TESTVECTORS = src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.h
 
 src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.h:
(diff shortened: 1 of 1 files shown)


Makefile.am
```diff
@@ -217,11 +218,29 @@ precomp: $(PRECOMP)
...
+
+### Pregenerated test vectors
+### (see the comments in the previous section for detailed rationale)
+TESTVECTORS = src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.h
+
+src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.h:
+	mkdir -p $(@D)
+	python3 tools/tests_wycheproof_generate.py src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.json > $@
+
+testvectors: $(TESTVECTORS)
```
hebasto, line 234:
```suggestion
.PHONY: testvectors
testvectors: $(TESTVECTORS)
```

hebasto (commented):
In 3955f8f9940f35a37faa01c08d17795adf8ceb43
> autotools: Rename "pregenerated" clean targets and make them .PHONY
> 
> This follows the automake conventions, see:
> https://www.gnu.org/software/automake/manual/html_node/Clean.html

The `clean-precomp` was `.PHONY`: https://github.com/bitcoin-core/secp256k1/blob/3bab71cf0534fd5948386de2b6ac31a7044ad76b/Makefile.am#L1

A comment needs to be updated: https://github.com/bitcoin-core/secp256k1/blob/3bab71cf0534fd5948386de2b6ac31a7044ad76b/Makefile.am#L205

Why not just `MAINTAINERCLEANFILES = $(PRECOMP) $(TESTVECTORS)`?

hebasto:
```diff
--- a/ci/cirrus.sh
+++ b/ci/cirrus.sh
@@ -109,7 +109,7 @@ fi
 # Rebuild precomputed files (if not cross-compiling).
 if [ -z "$HOST" ]
 then
-    make clean-precomp clean-testvectors
+    make maintainer-clean
     make precomp testvectors
 fi
 
```

to fix CI?

real-or-random:
> Why not just `MAINTAINERCLEANFILES = $(PRECOMP) $(TESTVECTORS)`?

Yeah, that would work, too. But I think it's nice to have the ability to clean `precomp` or `testvectors` separately. I'll address the other comments.

(Apparently it was a good idea that I opened a separate PR for the first commit.)

real-or-random:
I changed my mind and removed the renaming. It's not clear that a "maintainer"-prefix is better... Sorry for the back and forth.

Independently, `make maintainer-clean` in the Cirrus script won't work because that target also removes the Makefile.

Makefile.am
```diff
@@ -217,11 +216,29 @@ precomp: $(PRECOMP)
...
 clean-precomp:
 	rm -f $(PRECOMP)
+maintainer-clean-local: clean-precomp
+
+### Pregenerated test vectors
+### (see the comments in the previous section for detailed rationale)
+TESTVECTORS = src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.h
+
+src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.h:
+	mkdir -p $(@D)
```
hebasto, line 229: For more robustness, could consider
```suggestion
	$(MKDIR_P) $(@D)
```

with the [`AC_PROG_MKDIR_P`](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/autoconf.html#index-AC_005fPROG_005fMKDIR_005fP-1) macro in the `configure.ac`?

Makefile.am
```diff
@@ -217,11 +216,29 @@ precomp: $(PRECOMP)
...
+	python3 tools/tests_wycheproof_generate.py src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.json > $@
```
hebasto, line 230: 2dd672a287a0c9c4382ffbf56f690727f2e74696

To make it work in a VPATH build, the diff required as follows:
```suggestion
	python3 $(top_srcdir)/tools/tests_wycheproof_generate.py $(top_srcdir)/src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.json > $@
```

Makefile.am
```diff
@@ -202,7 +201,7 @@ precompute_ecmult_gen_LDADD = $(COMMON_LIB)
 # otherwise make's decision whether to rebuild them (even in the first
 # build by a normal user) depends on mtimes, and thus is very fragile.
 # This means that rebuilds of the prebuilt files always need to be
-# forced by deleting them, e.g., by invoking `make clean-precomp`.
+# forced by deleting them, e.g., using `make maintainer-clean`.
```
hebasto, line 204: Considering that `make maintainer-clean` [removes](https://github.com/bitcoin-core/secp256k1/pull/1277#issuecomment-1514815590) `Makefile`, could this comment be confusing if someone follows it blindly?

Makefile.am
hebasto: nm, ignore this ^

Makefile.am
real-or-random: As I understand the history of this, the reason why this macro exists is that some implementations of `mkdir -p` on some old systems are not thread safe. And that's a problem in a parallel make, where multiple make processes try to create the same directory (see https://lists.gnu.org/r/automake/2006-04/msg00084.html).

I don't see how this could be a problem here, given that there's only a single recipe in the Makefile that creates this directory. That's why I think it's better to stick to `mkdir -p` here for simplicity.

Makefile.am
real-or-random: I just removed that part.

Makefile.am
real-or-random: added a commit

hebasto (approved):
ACK 7e977b3c5071fc17575ff88ebbc9db7b17c70497

PR #1279, opened 2023-04-14
tests: lint wycheproof's python script

RandomLattice:
This PR lints tests_wycheproof_generate.py according to bitcoin's python linting scripts. This is a follow-up to PR #1245.


real-or-random:
@sipa Makes sense.

But if your "disable linting" in https://github.com/bitcoin/bitcoin/pull/27445 is just a WIP commit you don't want to have there right now, I'm also happy to merge this PR here quickly. We know that the current changes make Core's linter happy. We should then address `pylint` stuff in a separate PR.

real-or-random:
Ok, just sounds good! Either approach is fine with me.

real-or-random (approved):
utACK 35ada3b954ccc6c54628fb3bcc0365d176297019

PR #1282, opened 2023-04-20
Remove secp256k1_fe_const_b

roconnor-blockstream:
It's last use was eliminated in #1217.


real-or-random (approved):
utACK d6be470ee574b58bbd222a5c14cd867fc0a7a363

real-or-random:
> ```
> src/tests.c: In function 'ecmult_const_mult_xonly':
> src/tests.c:4501:35: error: 'secp256k1_fe_const_b' undeclared (first use in this function); did you mean 'secp256k1_ge_const_g'?
>  4501 |             secp256k1_fe_add(&c, &secp256k1_fe_const_b);
>       |                                   ^~~~~~~~~~~~~~~~~~~~
>       |                                   secp256k1_ge_const_g
> ```
> 
> It would appear that #1118 added another use back (which can easily be converted to `secp256k1_fe_add_int()` instead.)

By the way, this shows that it's a good idea to remove `secp256k1_fe_const_b` to make sure that noone will use it in the future.

PR #1283, opened 2023-04-20
Get rid of secp256k1_fe_const_b

sipa:
Replaces #1282.

Its only remaining use is in a test introduced in #1118, and it is easily replaced by the new `secp256k1_fe_add_int` from #1217.


real-or-random (approved):
utACK 69e1ec033120497b83dd95d92166fa05c54b8a06

PR #1285, opened 2023-04-21
bench: Make sys/time.h a system include

real-or-random:
just because it is minimally more correct


hebasto (approved):
ACK 68b16a1662af2db801a87d6f1afedca93ec2501c, I've skimmed through the whole codebase and did not find any more similar cases.

PR #1286, opened 2023-04-21
tests: remove extra semicolon in macro

jonasnick:


hebasto:
Where the `PRINT_BUF` macro is used?

jonasnick:
@hebasto it was just added with #1205 and not used anywhere as it's intended for development and debugging (see https://github.com/bitcoin-core/secp256k1/pull/1205#issuecomment-1413364303 for the rationale).

real-or-random:
Now I also realized that the current location of the macro and the function are not great because you can use them only in `tests.c` but not in the actual library... Sorry for that.

@jonasnick Can you move the macro and the function? I propose to move it to the top of `util.h`, which is included very early. This should do the trick.

jonasnick:
@real-or-random good point. Moved to `util.h`.

real-or-random (approved):
utACK c4062d6b5d83572c1932f32003a7c0e901fffc23

hebasto (approved):
ACK c4062d6b5d83572c1932f32003a7c0e901fffc23, I have reviewed the code and it looks OK.

PR #1289, opened 2023-04-29
cmake: Use full signature of `add_test()` command

hebasto:
This PR fixes tests for Windows binaries using Wine:
```
$ cmake -S . -B ../mingw -DCMAKE_TOOLCHAIN_FILE=cmake/x86_64-w64-mingw32.toolchain.cmake
$ cmake --build ../mingw
$ cmake --build ../mingw -t check
Test project /home/hebasto/git/secp256k1/mingw
    Start 1: noverify_tests
Could not find executable noverify_tests
...
```


real-or-random:
Concept ACK

Mostly because I'm curious: Do you know why this fixes tests? After reading the docs in https://cmake.org/cmake/help/latest/command/add_test.html, I don't see how the full syntax should make a difference here.

.cirrus.yml
```diff
@@ -403,7 +403,7 @@ task:
     - cmake --build build --config RelWithDebInfo -- -property:UseMultiToolTask=true;CL_MPcount=5
   check_script:
     - '%x64_NATIVE_TOOLS%'
-    - ctest --test-dir build -j 5
+    - ctest -C RelWithDebInfo --test-dir build -j 5
```
real-or-random, line 406: Is this change necessary? Shouldn't that be the default already?

Coins2payroll (approved):

.cirrus.yml
hebasto: > Is this change necessary? Shouldn't that be the default already?

It does. Otherwise, on Windows, `ctest` complains:
```
Test not available without configuration.  (Missing "-C <config>"?)
```

real-or-random (approved):
ACK 755629bc032af155ff68fa820244d04cb80547c5

PR #1293, opened 2023-04-30
refactor: Avoid unary minus on unsigned integers

hebasto:
The assembly code remains unchanged.

In addition, this PR enables MSVC warning [C4146](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4146) for the entire codebase.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -397,7 +397,7 @@ task:
     - PowerShell -NoLogo -Command if ($env:CIRRUS_PR -ne $null) { git fetch $env:CIRRUS_REPO_CLONE_URL pull/$env:CIRRUS_PR/merge; git reset --hard FETCH_HEAD; }
   configure_script:
     - '%x64_NATIVE_TOOLS%'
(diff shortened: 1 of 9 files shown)


real-or-random:
I think rewriting `-x` to `(0 - x)` is a bit nicer than `(~x + 1). 

With that change applied, I think I'm literally ~0 on that PR: On the one hand, changing the code to make the compiler happy is not great... But perhaps having a clean build with just /W2 is good.

hebasto:
> I think rewriting `-x` to `(0 - x)` is a bit nicer than `(~x + 1).

Done.

> On the one hand, changing the code to make the compiler happy is not great...

I agree with you.

> But perhaps having a clean build with just /W2 is good.

That is my point.

jonasnick (commented):
Having to change perfectly fine `-x` to `0-x` seems a bit annoying. I can imagine that this will confuse potential future reviewers. I don't think that's worth it.

PR #1294, opened 2023-05-04
cmake: Emulate GNU Autotools 'make check -jN'

hebasto:
It seems parallelism is requested for `make check` target when building with CMake and the "Unix Makefiles" generator which is the default on Linux and macOS. See:
- https://gnusha.org/secp256k1/2023-03-08.log:
> 13:53 \< sipa\> `make -j check` doesn't actually parallellize in the cmake build, which is a bit unfortunate
- https://github.com/hebasto/bitcoin/pull/15#issuecomment-1531155633:
> Is `make check` being run with multiple jobs? I assume not

With this PR, it is possible to build like that:
```
cmake -S . -B  build
make -C build -j $(nproc)
make check -C build -j $(nproc)
```

Making this PR a draft as it seems low priority and the diff seems a bit hackish.

My personal preference is to use the CMake's native `ctest` command :)

Anyway, this PR fixes an item in https://github.com/bitcoin-core/secp256k1/issues/1235.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -228,15 +228,18 @@ endif()
 
 set(CMAKE_C_VISIBILITY_PRESET hidden)
 
-# Ask CTest to create a "check" target (e.g., make check) as alias for the "test" target.
-# CTEST_TEST_TARGET_ALIAS is not documented but supposed to be user-facing.
-# See: https://gitlab.kitware.com/cmake/cmake/-/commit/816c9d1aa1f2b42d40c81a991b68c96eb12b6d2
-set(CTEST_TEST_TARGET_ALIAS check)
 include(CTest)
 # We do not use CTest's BUILD_TESTING because a single toggle for all tests is too coarse for our needs.
 mark_as_advanced(BUILD_TESTING)
 if(SECP256K1_BUILD_BENCHMARK OR SECP256K1_BUILD_TESTS OR SECP256K1_BUILD_EXHAUSTIVE_TESTS OR SECP256K1_BUILD_CTIME_TESTS OR SECP256K1_BUILD_EXAMPLES)
   enable_testing()
+  if(CMAKE_GENERATOR STREQUAL "Unix Makefiles")
+    # Emulate GNU Autotools 'make check -jN'.
+    add_custom_target(check
+      COMMAND @$(CMAKE_COMMAND) -E cmake_echo_color --switch=$(COLOR) --cyan \"Running tests...\"
+      COMMAND ${CMAKE_CTEST_COMMAND} --force-new-ctest-process -$(MAKEFLAGS)
+    )
+  endif()
 endif()
 
 add_subdirectory(src)


CMakeLists.txt
```diff
@@ -228,15 +228,18 @@ endif()
...
 include(CTest)
 # We do not use CTest's BUILD_TESTING because a single toggle for all tests is too coarse for our needs.
 mark_as_advanced(BUILD_TESTING)
 if(SECP256K1_BUILD_BENCHMARK OR SECP256K1_BUILD_TESTS OR SECP256K1_BUILD_EXHAUSTIVE_TESTS OR SECP256K1_BUILD_CTIME_TESTS OR SECP256K1_BUILD_EXAMPLES)
   enable_testing()
+  if(CMAKE_GENERATOR STREQUAL "Unix Makefiles")
+    # Emulate GNU Autotools 'make check -jN'.
+    add_custom_target(check
+      COMMAND @$(CMAKE_COMMAND) -E cmake_echo_color --switch=$(COLOR) --cyan \"Running tests...\"
+      COMMAND ${CMAKE_CTEST_COMMAND} --force-new-ctest-process -$(MAKEFLAGS)
```
real-or-random, line 240: Passing `$MAKEFLAGS` to ctest is indeed pretty hackish (and ctest doesn't like `-j` without an integer and some users have more `MAKEFLAGS` set in their env...)

real-or-random:
> My personal preference is to use the CMake's native `ctest` command :)

Yep, we should just update the CMake build docs in the README. We could even remove the "make check" emulation entirely if this is more CMake-ish. What do you think?

hebasto:
> > My personal preference is to use the CMake's native `ctest` command :)
> 
> Yep, we should just update the CMake build docs in the README. We could even remove the "make check" emulation entirely if this is more CMake-ish. What do you think?

As we discussed at CoreDev, it would be better for the common good to use a unified approach for both projects, Bitcoin Core and libsecp256k1. A similar discussion is currently taking place in Bitcoin Core ([here](https://github.com/hebasto/bitcoin/pull/15#issuecomment-1531155633), [here](https://github.com/hebasto/bitcoin/pull/15#issuecomment-1531440684) and [here](https://github.com/hebasto/bitcoin/pull/15#issuecomment-1531450728)). Regarding your question, I think it would be advisable to ask for the opinions of developers who expressed concerns about `make check`, namely @sipa and @fanquake.

hebasto:
I've updated the PR description with a couple of alternatives.

real-or-random:
> As we discussed at CoreDev, it would be better for the common good to use a unified approach for both projects, Bitcoin Core and libsecp256k1

@hebasto Has been any progress on this point in Core?

hebasto:
> > As we discussed at CoreDev, it would be better for the common good to use a unified approach for both projects, Bitcoin Core and libsecp256k1
> 
> @hebasto Has been any progress on this point in Core?

In the https://github.com/hebasto/bitcoin/tree/cmake-staging, we lean to use the native CMake's invocations in all scripts (CI, Guix etc). This approach is agnostic to the underlying build system, and `make check` parallelism issue is no longer a concern.

However, it might change in the future :)

real-or-random:
> In the [hebasto/bitcoin@`cmake-staging`](https://github.com/hebasto/bitcoin/tree/cmake-staging?rgh-link-date=2024-01-17T12%3A50%3A58Z), we lean to use the native CMake's invocations in all scripts (CI, Guix etc). This approach is agnostic to the underlying build system, and `make check` parallelism issue is no longer a concern.

Let's stick with this for now. Then people can use `ctest -j3` or `ctest -j$(nproc)` or they can also set `CTEST_PARALLEL_LEVEL="$(nproc)"` similar to `MAKEFLAGS="-j($proc)"`. Just be aware that `ctest -j` (without an explicit int) doesn't have an effect.

PR #1295, opened 2023-05-04
abi: Use dllexport for mingw builds

theuni:
Addresses the first part of #1181. See the discussion there for more context and history.

After this, all that remains is a (platform-independent) exports checker for c-i. Or perhaps a linker script or .def file could be tricked into testing as a side-effect.

This should fix mingw exports, specifically hiding the following:
`secp256k1_pre_g_128`
`secp256k1_pre_g`
`secp256k1_ecmult_gen_prec_table`

This changes our visibility macros to look more like [gcc's recommendation](https://gcc.gnu.org/wiki/Visibility#How_to_use_the_new_C.2B-.2B-_visibility_support).

Edit:
Note that we could further complicate this by supporting `__attribute__ ((dllexport))` as well, though I didn't bother as I'm not sure what compiler combo would accept that but not the bare dllexport syntax.

Edit2:
As the title implies, this affects this ABI and could affect downstream libs/apps in unintended ways (though it's hard to imagine any real downside). Though because it's win32 only, I'm imagining very little real-world impact at all.

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -133,8 +133,9 @@ typedef int (*secp256k1_nonce_function)(
 # define SECP256K1_NO_BUILD
 #endif
 
-/* Symbol visibility. See libtool manual, section "Windows DLLs". */
-#if defined(_WIN32) && !defined(__GNUC__)
+/* Symbol visibility. See https://gcc.gnu.org/wiki/Visibility */
+/* DLL_EXPORT is defined internally for shared builds */
+#if defined(_WIN32)
 # ifdef SECP256K1_BUILD
 #  ifdef DLL_EXPORT
 #   define SECP256K1_API            __declspec (dllexport)


theuni:
Ping @hebasto

theuni:
@real-or-random We should probably add a changelog entry for this (if you agree with it, and once we have a test for it): 

> Symbol visibility is now believed to be handled properly on supported platforms and is now considered to be part of the ABI. Please report any improperly exported symbols as a bug.

include/secp256k1.h
```diff
@@ -134,7 +134,7 @@ typedef int (*secp256k1_nonce_function)(
 #endif
 
 /* Symbol visibility. See libtool manual, section "Windows DLLs". */
```
real-or-random, line 136: Can you update this comment to say that we slightly deviate from the suggestion in the libtool manual because the mingw linker apparently respects `_declspec (dllexport)` but not `_attribute__ ((visibility ("default")))`?

real-or-random (commented):
Concept ACK

real-or-random:
> As the title implies, this affects this ABI and could affect downstream libs/apps in unintended ways (though it's hard to imagine any real downside). Though because it's win32 only, I'm imagining very little real-world impact at all.

Yeah, I wouldn't consider it an ABI or an API change. IIUC the worst that can happen is that someone (wrongly) used one of these symbols in the past and now the dynamic linker can't find it anymore.

hebasto:
I've verified the cross-compiled DLLs on Windows using the `dumpbin /exports` command. I can confirm that this PR indeed hides the mentioned symbols.

khademreza688 (approved):

hebasto:
When cross-compiling for Windows, in which cases is the `SECP256K1_BUILD` macro defined, and `DLL_EXPORT` is not defined?

hebasto:
FWIW, I've rebased https://github.com/bitcoin-core/secp256k1/pull/1135 on top of this PR and integrated it into CI: https://github.com/hebasto/secp256k1/tree/pr1295/0505.test.2

Only "{x86_64|i686} (MSVC): Windows (Debian stable, Wine...)" jobs [fail](https://cirrus-ci.com/build/6466894450917376).

theuni:
> When cross-compiling for Windows, in which cases is the `SECP256K1_BUILD` macro defined, and `DLL_EXPORT` is not defined?

When building a static libsecp256k1.

hebasto (commented):
ACK ac3c47bac37d2d51092897d6a663cf3989de590c

theuni:
> FWIW, I've rebased #1135 on top of this PR and integrated it into CI: https://github.com/hebasto/secp256k1/tree/pr1295/0505.test.2
> 
> Only "{x86_64|i686} (MSVC): Windows (Debian stable, Wine...)" jobs [fail](https://cirrus-ci.com/build/6390887589085184).

It's not clear to me what the error is here?

hebasto:
> > FWIW, I've rebased #1135 on top of this PR and integrated it into CI: [hebasto/secp256k1@`pr1295`/0505.test.2](https://github.com/hebasto/secp256k1/tree/pr1295/0505.test.2?rgh-link-date=2023-05-05T14%3A42%3A01Z)
> > Only "{x86_64|i686} (MSVC): Windows (Debian stable, Wine...)" jobs [fail](https://cirrus-ci.com/build/6390887589085184).
> 
> It's not clear to me what the error is here?

Symbols are still [exported](https://cirrus-ci.com/task/5455323922694144):
```
python3 ./tools/symbol-check.py .libs/secp256k1-2.dll
.libs/secp256k1-2.dll: export of function ?_OptionsStorage@?1??__local_stdio_printf_options@@9@9 not expected
.libs/secp256k1-2.dll: export of function secp256k1_ecmult_gen_prec_table not expected
.libs/secp256k1-2.dll: export of function secp256k1_pre_g not expected
.libs/secp256k1-2.dll: export of function secp256k1_pre_g_128 not expected
.libs/secp256k1-2.dll: failed EXPORTED_FUNCTIONS
```

But it is not mingw compiler.

include/secp256k1.h
real-or-random: https://www.gnu.org/software/libtool/manual/html_node/Windows-DLLs.html

theuni: Updated to include our reference as opposed to what we _don't_ use. This ok?

theuni:
Ping @real-or-random. Anything left to do here?

real-or-random (approved):
utACK bc7c8db179a56cf7273f3c4c0decd10543a10521

PR #1299, opened 2023-05-09
Infinity handling: ecmult_const(infinity) works, and group verification

sipa:
Rebase of #791.

* Clean up infinity handling, make x/y/z always initialized for infinity.
* Make secp256k1_ecmult_const handle infinity. 
  * Infinity isn't currently needed here, but correctly handling it is a little more safe against future changes.
  * Update docs for it to make it clear that it is not constant time in Q. It never was constant time in Q (and would be a little complicated to make constant time in Q: needs a constant time addition function that tracks RZR). It isn't typical for ECDH to be constant time in terms of the pubkey. If it was later made constant time in Q infinity support would be easy to preserve, e.g. by running it on a dummy value and cmoving infinity into the output.
* Add group verification (`secp256k1_ge_verify` and `secp256k1_gej_verify`, mimicking `secp256k1_fe_verify`).
* Make the `secp256k1_{fe,ge,gej}_verify` functions also defined (as no-ops) in non-VERIFY mode.

### DIFF
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -92,12 +92,26 @@ static void secp256k1_gej_verify(const secp256k1_gej *a) {
     (void)a;
 }
 
+/* Set r to the affine coordinates of Jacobian point (a.x, a.y, 1/zi). */
 static void secp256k1_ge_set_gej_zinv(secp256k1_ge *r, const secp256k1_gej *a, const secp256k1_fe *zi) {
     secp256k1_fe zi2;
     secp256k1_fe zi3;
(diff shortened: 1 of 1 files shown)


real-or-random (approved):
ACK 2d3b35077f98bea0a136b7cabd7592dd3b246827

jonasnick (commented):
ACK bbc834467c5d14e3e53744211e7c4fa9d8fabe41

real-or-random (approved):
ACK bbc834467c5d14e3e53744211e7c4fa9d8fabe41

PR #1300, opened 2023-05-09
Avoid normalize conditional on VERIFY

sipa:
In the old code, `secp256k1_gej_rescale` requires a normalized input in VERIFY mode, but not otherwise. Its requirements shouldn't depend on this mode being enabled or not.

### DIFF
--- a/src/ecmult_impl.h
+++ b/src/ecmult_impl.h
@@ -279,9 +279,6 @@ static void secp256k1_ecmult_strauss_wnaf(const struct secp256k1_strauss_state *
          */
         tmp = a[np];
         if (no) {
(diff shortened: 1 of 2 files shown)


real-or-random (approved):
utACK I've also verified that the loop in `secp256k1_ecmult_strauss_wnaf` holds up the invariant that the magnitude of `Z` is 1, even with the normalization removed

real-or-random (approved):
utACK 97c63b90390b0b11a5d3530b03855ec9cc0de343 I've also verified that the loop in secp256k1_ecmult_strauss_wnaf holds up the invariant that the magnitude of Z is 1, even with the normalization removed

PR #1301, opened 2023-05-09
Avoid using bench_verify_data as bench_sign_data; merge them

sipa:
The existing bench.c code defines `bench_verify_data data` variable, but some of the benchmarks then use it as `bench_sign`. Fix this by merging the two types into one.

### DIFF
--- a/src/bench.c
+++ b/src/bench.c
@@ -64,11 +64,11 @@ typedef struct {
     size_t siglen;
     unsigned char pubkey[33];
     size_t pubkeylen;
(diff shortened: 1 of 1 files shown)


stratospher:
ACK 2e65f1f.

real-or-random (approved):
utACK https://github.com/bitcoin-core/secp256k1/pull/1301/commits/2e65f1fdbcc87e2ef8c0baf4abc8ee0f56daf7fe

Issue #1302, opened 2023-05-10
Rework or get rid of scratch space 


real-or-random:
Our confidence in the scratch space code isn't particularly high. It reinvents bump allocation, but it had a few issues in the past. All the code that uses scratch space is currently unreachable from the public API (except that we have `secp256k1_scratch_create` and `secp256k1_scratch_destroy` themselves in the public API.

A much simpler alternative is to get rid of scratch spaces and just assume the existence of `malloc`/`free` and use these directly. The disadvantage of this is that it's a bit harder for platforms that don't have `malloc`. 

Another alternative is to rework the scratch space code. It may be possible to simply it and improve its usability.  

I think our future directions on this should be guided by whatever we feel is best for our cases:
  - batch validation (see https://github.com/bitcoin-core/secp256k1/issues/1087 and https://github.com/bitcoin-core/secp256k1/pull/1134)
  - possibly musig2 key aggregation (maybe this is something people want to do run on hardware wallets?)

apoelstra:
In secp256k1-zkp the scratch space gets actual usage, and I find it has a much nicer API than malloc/free. The API lets us:

* Allocate a slab up-front (or rather, check that the user passed in enough memory)
* As we are writing code, bump-allocate from the slab, asserting that we have enough, which sanity checks our up-front calculation
* Before doing something branch-heavy, checkpointing the allocator, then in the end just resetting to that, rather than having to wrory about correct cleanup logic in every branch

So I think it has more benefit than just replacing `malloc` and `free` for freestanding systems. Though it may still make sense to drop the code here and have it exclusively live in secp-zkp for a while.

apoelstra:
@sipa these are all good points. concept ACK from me on removing the API from this library.

jonasnick:
> The fact that the scratch space API forces us to write code which has predictable (and settable) memory usage, is both very useful (but only to some users) and a large source of the complexity of accomodating the current scratch space API. However, nothing prevents us from having that property too in different ways (e.g. by passing a "max memory usage" argument to functions).

In the case of `ecmult_multi`, it can be argued that we could achieve similar predictability by adding a `max_batch_size` argument to `ecmult_multi` and the user-facing API (e.g., `schnorrsig_batch_verify` and `musig_keyagg`).
This would have the advantage of freeing us from the complexity of dealing with memory usage.

[...]

PR #1305, opened 2023-05-11
Remove unused scratch space from API

jonasnick:
Not sure if we want the typedef and `secp256k1_scratch_space_{create,destroy}` but if we don't keep them then this PR will be a rather large diff.

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -49,19 +49,6 @@ extern "C" {
  */
 typedef struct secp256k1_context_struct secp256k1_context;
 
-/** Opaque data structure that holds rewritable "scratch space"
- *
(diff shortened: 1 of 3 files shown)


real-or-random (approved):
utACK 712e7f8722eba5dec2bc6b37d75aadeb6f6e633b

jonasnick:
I should have mentioned the rationale. The idea is to remove the scratch space from the API is because it's unused right now. Maybe we bring it back (in some form) at some point (see #1302).

real-or-random:
cc @apoelstra

apoelstra (approved):
ACK 712e7f8722eba5dec2bc6b37d75aadeb6f6e633b

PR #1306, opened 2023-05-11
build: Make tests work with external default callbacks

real-or-random:
Supersedes #971.

After that, we still want to fix the build systems:
benchmarks and examples should be disabled when external callbacks are enabled.


jonasnick (commented):
ACK 1907f0f1664e3a966daa58be956af18e48834ffd

PR #1307, opened 2023-05-12
Mark more assembly outputs as early clobber

### LINKED ISSUE #766: Missing early clobber

sipa:
This fixes #766.

In the existing code, the compiler is allowed to allocate the RSI register for outputs m0, m1, or m2, which are written to before the input in RSI is read from. Fix this by marking them as early clobber.

Make a similar change in the field logic, even though not triggerable there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+#### Fixed
+ - Fixed an old bug that permitted compilers to potentially output bad assembly code on x86_64. In theory, it could lead to a crash or a read of unrelated memory, but this has never been observed on any compilers so far.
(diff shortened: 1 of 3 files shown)


real-or-random (approved):
ACK 8c9ae37a5a26cdeb6365624fee43f41b238830e4

jonasnick (commented):
ACK 8c9ae37a5a26cdeb6365624fee43f41b238830e4

I'm reasonably sure that we with this PR we found all outputs that needed to be marked as early clobbers.

PR #1316, opened 2023-05-15
Do not invoke fe_is_zero on failed set_b32_limit

sipa:
Noticed in the CI output of #1313 (https://cirrus-ci.com/task/5117786435878912)

The code violates the field element contract that states that a field element that comes out of a failed `secp256k1_fe_set_b32_limit` call cannot be used before overwriting it. This is not an issue in practice, as such failure can only occur with negligible probability, but the experimental compiler in that CI setting is technically correct in detecting this possibility.

Fix it by setting it to 1 based on a `secp256k1_fe_normalizes_to_zero` test rather than a `secp256k1_fe_is_zero` one (which does not require normalization).

### DIFF
--- a/src/ecmult_gen_impl.h
+++ b/src/ecmult_gen_impl.h
@@ -87,7 +87,6 @@ static void secp256k1_ecmult_gen_blind(secp256k1_ecmult_gen_context *ctx, const 
     secp256k1_fe s;
     unsigned char nonce32[32];
     secp256k1_rfc6979_hmac_sha256 rng;
-    int overflow;
     unsigned char keydata[64];
     if (seed32 == NULL) {
         /* When seed is NULL, reset the initial point and blinding value. */
@@ -106,11 +105,9 @@ static void secp256k1_ecmult_gen_blind(secp256k1_ecmult_gen_context *ctx, const 
     memcpy(keydata + 32, seed32, 32);
     secp256k1_rfc6979_hmac_sha256_initialize(&rng, keydata, 64);
     memset(keydata, 0, sizeof(keydata));
-    /* Accept unobservably small non-uniformity. */
     secp256k1_rfc6979_hmac_sha256_generate(&rng, nonce32, 32);
-    overflow = !secp256k1_fe_set_b32_limit(&s, nonce32);
-    overflow |= secp256k1_fe_is_zero(&s);
-    secp256k1_fe_cmov(&s, &secp256k1_fe_one, overflow);
+    secp256k1_fe_set_b32_mod(&s, nonce32);
+    secp256k1_fe_cmov(&s, &secp256k1_fe_one, secp256k1_fe_normalizes_to_zero(&s));
     /* Randomize the projection to defend against multiplier sidechannels.
        Do this before our own call to secp256k1_ecmult_gen below. */
     secp256k1_gej_rescale(&ctx->initial, &s);


real-or-random (approved):
utACK 9fb5f2d3b0fdd00a9f5b51499a8f7d694725a8a5

I was about to open the same PR... :) And yes, ignoring the issue seems to be the simplest and most reasonable solution in this case.

stratospher (commented):
ACK 9fb5f2d.

stratospher:
>The new code violates the precondition of secp256k1_gej_rescale with negligible probability. 

which precondition? did i miss something?

the preconditions for `s` in `secp256k1_gej_rescale` are that it should pass the checks for:
- `secp256k1_fe_verify` (magnitude = [0, 32] and normalized = 0 or 1)
- `secp256k1_fe_sqr` (magnitude <= 8)

these would be satisfied because `secp256k1_fe_set_b32_mod` sets magnitude = 1 and  normalized=0 for `s`.

src/ecmult_gen_impl.h
```diff
@@ -106,11 +105,13 @@ static void secp256k1_ecmult_gen_blind(secp256k1_ecmult_gen_context *ctx, const
...
     secp256k1_rfc6979_hmac_sha256_generate(&rng, nonce32, 32);
-    overflow = !secp256k1_fe_set_b32_limit(&s, nonce32);
-    overflow |= secp256k1_fe_is_zero(&s);
-    secp256k1_fe_cmov(&s, &secp256k1_fe_one, overflow);
+    secp256k1_fe_set_b32_mod(&s, nonce32);
+#ifdef VERIFY
+    /* Rescaling with s=0 would be invalid, but can only happen with negligible probability as it
+     * is the output of a CSPRNG. */
+    VERIFY_CHECK(!secp256k1_fe_normalizes_to_zero_var(&s));
+#endif
```
real-or-random, line 114:
```suggestion
    secp256k1_fe_set_b32_mod(&s, nonce32);
    secp256k1_fe_cmov(&s, &secp256k1_fe_one, secp256k1_fe_normalizes_to_zero(&s);
```

or just this.

real-or-random:
> which precondition?

Ok yes, I didn't mean magnitude or normalization. `secp256k1_gej_rescale`  requires that `s` is not zero, or more technically, that `!secp256k1_fe_normalizes_to_zero(&s)` (which is why the current code is checks for this).

stratospher:
> Ok yes, I didn't mean magnitude or normalization. secp256k1_gej_rescale requires that s is not zero, or more technically, that !secp256k1_fe_normalizes_to_zero(&s) (which is why the current code checks for this).

oh! makes sense.

real-or-random (approved):
utACK 6433175ffe2435bcee7333e21480e4194083caae

real-or-random:
> @real-or-random This makes me wonder if a "nonzero" property (like magnitude/normalized) could be added...

That would be true if the fe is guaranteed to be nonzero? (Rust has this in the type system: https://doc.rust-lang.org/std/num/struct.NonZeroU64.html)

Hm, not sure if it's worth the hassle. It's pretty hard to guarantee that numbers are nonzero. I mean not mean operations keep the guarantees (mostly just neg and mul). Do we need the property that things nonzero in many other places? Inversions perhaps?

PR #1323, opened 2023-05-23
tweak_add: fix API doc for tweak=0

### LINKED ISSUE #1322: Wrong documentation for secp256k1_ec_seckey_tweak_add
The documentation for `secp256k1_ec_seckey_tweak_add` states that `If the tweak is invalid according to secp256k1_ec_seckey_verify, this function returns 0`. However if `tweak32` is zero, it will not produce an error unlike the tests in `secp256k1_ec_seckey_verify` so this is incorrect. It only produces an error on an overflow.

Either it should test for zero, or the documentation should be changed. A zero scalar results in the same private key. I'm not sure if allowing a zero scalar is sensible or not.

jonasnick:
Fixes #1322 (@MatthewLM). We actually test that a 0 tweak is fine for all `tweak_add` functions.


include/secp256k1.h
```diff
@@ -733,10 +733,11 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ec_pubkey_negate(
  *                  invalid according to secp256k1_ec_seckey_verify, this
  *                  function returns 0. seckey will be set to some unspecified
  *                  value if this function returns 0.
- *  In:    tweak32: pointer to a 32-byte tweak. If the tweak is invalid according to
- *                  secp256k1_ec_seckey_verify, this function returns 0. For
- *                  uniformly random 32-byte arrays the chance of being invalid
- *                  is negligible (around 1 in 2^128).
+ *  In:    tweak32: pointer to a 32-byte tweak. For non-zero tweaks, the function
```
real-or-random, line 736:
```suggestion
 *  In:    tweak32: pointer to a 32-byte tweak. For non-zero tweaks, the function
```

real-or-random (approved):
ACK 05873bb6b1041227f413626717b7af69fb6176ef

PR #1320, opened 2023-05-23
ci: Remove quirk that runs dummy command after wineserver

real-or-random:
The underlying issue is now worked around in upstream, see https://github.com/mstorsjo/msvc-wine/issues/47 for details.

PR #1327, opened 2023-05-24
ci: Move wine prefix to /tmp to avoid error D8037 in cl.exe

### LINKED ISSUE #1326: ci: MSVC fails after rebuild of Docker image
As part of testing https://github.com/bitcoin-core/secp256k1/pull/1320, I forced a rebuild of the Docker image (after the successful CI run on the old image, i.e., after @hebasto's ACK.)  I think we should anyway do this from time to time...

Now MSVC refuses to run because it can't write some temp files. See for example:
https://github.com/bitcoin-core/secp256k1/pull/1320/checks?check_run_id=13686337961

Some resources:
 - https://www.winehq.org/pipermail/wine-users/2007-June/027383.html
 - https://stackoverflow.com/questions/10560779/cl-exe-when-launched-via-createprocess-does-not-seem-to-have-write-permissions

real-or-random:
Don't ask me why this makes a difference. It may be some permission problem even though everything in Cirrus CI runs as root anyway. In any case, I'll probably get mad if I investigate this further.

Fixes #1326.

### DIFF
--- a/ci/linux-debian.Dockerfile
+++ b/ci/linux-debian.Dockerfile
@@ -31,7 +31,9 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     python3 msvc-wine/vsdownload.py --accept-license --dest /opt/msvc Microsoft.VisualStudio.Workload.VCTools && \
     msvc-wine/install.sh /opt/msvc
 
-# Initialize the wine environment. Wait until the wineserver process has
+# Moving the wine prefix to /tmp avoids error D8037 when invoking cl.exe.
+ENV WINEPREFIX=/tmp/wineprefix
+# Initialize the wine prefix. Wait until the wineserver process has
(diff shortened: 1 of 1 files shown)


hebasto (approved):
ACK 27504d5c941df89bc828067248270179c9dcb04b, tested in my personal Cirrus account.

> Don't ask me why this makes a difference.

I won't :)

real-or-random:
> Apparently this change does not work -- see CI in https://github.com/bitcoin-core/secp256k1/pull/1320.

I'm pretty sure that I could reliably reproduce D8037 in Docker locally, and moving WINEPREFIX to /tmp made this go away reliably. So, I'm rather convinced that there's some truth in this PR here, even if it fails now in #1320. (But yeah, we'll need to investigate further. Not sure at the moment).

PR #1328, opened 2023-05-24
build: Bump MSVC warning level up to W3

hebasto:
Solves one item in https://github.com/bitcoin-core/secp256k1/issues/1235.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,8 +212,12 @@ endif()
 include(TryAppendCFlags)
 if(MSVC)
   # Keep the following commands ordered lexicographically.
-  try_append_c_flags(/W2) # Moderate warning level.
+  try_append_c_flags(/W3) # Production quality warning level.
   try_append_c_flags(/wd4146) # Disable warning C4146 "unary minus operator applied to unsigned type, result still unsigned".
+  try_append_c_flags(/wd4244) # Disable warning C4244 "'conversion' conversion from 'type1' to 'type2', possible loss of data".
+  try_append_c_flags(/wd4267) # Disable warning C4267 "'var' : conversion from 'size_t' to 'type', possible loss of data".
+  # Eliminate deprecation warnings for the older, less secure functions.
+  add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
 else()
   # Keep the following commands ordered lexicographically.
   try_append_c_flags(-pedantic)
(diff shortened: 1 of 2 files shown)


src/CMakeLists.txt
```diff
@@ -75,31 +75,37 @@ elseif(CMAKE_SYSTEM_NAME STREQUAL "Windows")
 endif()
 unset(${PROJECT_NAME}_soversion)
 
+add_library(secp256k1_warning_suppression INTERFACE)
+target_compile_options(secp256k1_warning_suppression INTERFACE
+  # Disable warning C4996 for the usage of a function, class member, variable, or typedef that's marked deprecated.
+  $<$<C_COMPILER_ID:MSVC>:/wd4996>
+)
```
real-or-random, line 82: Why treat this special and not add it using `try_append_c_flags`?

real-or-random (commented):
Concept ACK

src/CMakeLists.txt
hebasto: Isn't it better to catch such warnings in non-test/bench code?

real-or-random: I'm not convinced that this justifies making things more complicated in the build system.

Hm, I just checked. The deprecated warnings here are all the same kind:

```
src/tests.c(7695,27): warning C4996: 'getenv': This function or variable may be unsafe. Consider using _dupenv_s instead. To disable deprecation, use _CRT_SECURE_NO_WARNINGS. See online help for details.
```

And we really don't want *these* warnings. We can't use `_dupenv_s` because it exists only on Windows. We can't use `getenv_s` because it requires C11 Annex K. But can we just pass `/D_CRT_SECURE_NO_WARNINGS` instead of `/wd4996`?  That's narrower, and then I think there's really no need to restrict it to test/bench/etc.

hebasto:
Addressed @real-or-random's https://github.com/bitcoin-core/secp256k1/pull/1328#discussion_r1205148502.

src/CMakeLists.txt
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/secp256k1/pull/1328#issuecomment-1562534860).

real-or-random (approved):
ACK 1549db0ca5193b8ba5d8f7478d54af2ca4b36c7e

Do you believe it's worth making the `-` vs `/` style of passing options consistent? I prefer dashes slightly, but that's just my taste. It doesn't matter at all in the end. Maybe if CMake internally uses `/` (e.g., in `add_compile_definitions`), we should just switch to that one.

hebasto:
> ... if CMake internally uses `/` ...

It does.

> Do you believe it's worth making the `-` vs `/` style of passing options consistent?

Consistency within a particular build system does matter. Not sure about "cross build systems" consistency although.

PR #1331, opened 2023-05-30
secp256k1_frost_sign(): use free_binding_factors() instead of manually rolled code

muxator:
In this way we have a single way of clearing binding factors across the code base. Before this change, `secp256k1_frost_sign()` was the only function that rolled its own code.


muxator:
sorry

PR #1332, opened 2023-05-30
cmake: Fix checking compiler flags like `-Wno-some-warning`

hebasto:
Some compilers ([GCC](https://gcc.gnu.org/onlinedocs/gcc/Warning-Options.html)) produce no diagnostic for `-Wno-unknown-warning` unless other diagnostics are being produced:

```
$ git diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 240557f..f976824 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,6 +220,7 @@ if(MSVC)
   add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
 else()
   # Keep the following commands ordered lexicographically.
+  try_append_c_flags(-Wno-unknown-warning)
   try_append_c_flags(-pedantic)
   try_append_c_flags(-Wall) # GCC >= 2.95 and probably many other compilers.
   try_append_c_flags(-Wcast-align) # GCC >= 2.95.
$ cmake -B build -DCMAKE_C_COMPILER=gcc 2>&1 | grep -i unknown
-- Performing Test C_SUPPORTS__WNO_UNKNOWN_WARNING
-- Performing Test C_SUPPORTS__WNO_UNKNOWN_WARNING - Success
Compile options ....................... -Wno-unknown-warning -pedantic -Wall -Wcast-align -Wcast-align=strict -Wextra -Wnested-externs -Wno-long-long -Wno-overlength-strings -Wno-unused-function -Wshadow -Wstrict-prototypes -Wundef
```

This PR fixes compiler flag check in CMake-based build system.

Let me know, if the same fix is desirable in Autotools-based build system.

### DIFF
--- a/cmake/TryAppendCFlags.cmake
+++ b/cmake/TryAppendCFlags.cmake
@@ -10,7 +10,13 @@ function(secp256k1_check_c_flags_internal flags output)
 
   # This avoids running a linker.
   set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
-  check_c_compiler_flag("${flags}" ${result})
+
+  # Some compilers (GCC) produce no diagnostic for -Wno-unknown-warning
+  # unless other diagnostics are being produced. Therefore, test the
+  # -Wsome-warning case instead of the -Wno-some-warning one.
+  string(REPLACE "-Wno-" "-W" non_negated_flags "${flags}")
+
+  check_c_compiler_flag("${non_negated_flags}" ${result})
 
   set(${output} ${${result}} PARENT_SCOPE)
 endfunction()


fanquake:
Is this currently an issue for anything? If the check is broken, then it should be fixed before we start using it.

real-or-random:
> Is this currently an issue for anything? If the check is broken, then it should be fixed before we start using it.

I think noone has run into this so far, but yeah, we should fix it.

Concept ACK

cmake/TryAppendCFlags.cmake
```diff
@@ -10,7 +10,13 @@ function(secp256k1_check_c_flags_internal flags output)
 
   # This avoids running a linker.
   set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
-  check_c_compiler_flag("${flags}" ${result})
+
+  # Some compilers (GCC) produce no diagnostic for -Wno-unknown-warning
+  # unless other diagnostics are being produced. Therefore, test the
+  # -Wsome-warning case instead of the -Wno-some-warning one.
```
real-or-random, line 16:
```suggestion
  # Some compilers (GCC) produce no diagnostic for -Wno-some-warning,
  # if "some-warning" is unknown to the compiler and no other diagnostics are being produced.
  #  Therefore, test the
  # -Wsome-warning case instead of the -Wno-some-warning one.
```

Modulo paragraph formatting.

This confused me a lot because I thought `-Wno-unknown-warning` is a real flag that says "Don't warn on unknown warnings". (Clang has a real flag `-Wno-unknown-warning-option` for this!)

hebasto:
> Is this currently an issue for anything?

The newest current option--`-Wno-overlength-strings`--is supported for GCC >=4.2. Only someone using an older compiler would be affected.

> If the check is broken, then it should be fixed before we start using it.

The PR description mentions two checks, one per build system. The check in the Autotools-based build system: https://github.com/bitcoin-core/secp256k1/blob/a5269373fa13ff845f654d81b90629dd78495641/build-aux/m4/bitcoin_secp.m4#L64-L78 is broken.

The check in the CMake-based build system mirrored the behaviour of its Autotools' counterpart.

Therefore, both checks need to be fixed.

cmake/TryAppendCFlags.cmake
hebasto: Thanks! Reworked.

PR #1333, opened 2023-05-30
test: Warn if both `VERIFY` and `COVERAGE` are defined

hebasto:
Solves one item in https://github.com/bitcoin-core/secp256k1/issues/1235.

Also see: https://github.com/bitcoin-core/secp256k1/pull/1113#discussion_r1127856040.


real-or-random (approved):
ACK e83801f5db2cb63e6343eff0ecfa40b7cd4a9090

Issue #1324, opened 2023-05-24
secp256k1_u128_rshift should use __shiftright128 intrinsic on MSVC


real-or-random:
https://learn.microsoft.com/en-us/cpp/intrinsics/shiftright128?view=msvc-170

PR #1336, opened 2023-06-01
Use `__shiftright128` intrinsic in `secp256k1_u128_rshift` on MSVC

hebasto:
Closes https://github.com/bitcoin-core/secp256k1/issues/1324.

As the `__shiftright128` [docs](https://learn.microsoft.com/en-us/cpp/intrinsics/shiftright128) state:
> The `Shift` value is always modulo 64...

it is not applicable for the `n >= 64` branch.

### DIFF
--- a/src/int128_struct_impl.h
+++ b/src/int128_struct_impl.h
@@ -80,7 +80,12 @@ static SECP256K1_INLINE void secp256k1_u128_rshift(secp256k1_uint128 *r, unsigne
      r->lo = r->hi >> (n-64);
      r->hi = 0;
    } else if (n > 0) {
+#if defined(_MSC_VER) && defined(_M_X64)
+     VERIFY_CHECK(n < 64);
+     r->lo = __shiftright128(r->lo, r->hi, n);
+#else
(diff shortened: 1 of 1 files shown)


hebasto:
> > it is not applicable for the `n >= 64` branch.
> 
> Good point, can you add a `VERIFY_CHECK(n < 64);` over the intrinsic line to document this?

Added.

> This #ifdef should be active on CI, right?

Yes.

> So this is tested on CI?

It is.

> It may make sense to add a few simple (fixed-input) tests for that function.

In addition to https://github.com/bitcoin-core/secp256k1/blob/60556c9f49a9384efd7f16b734820ae19108f053/src/tests.c#L1990-L1995 ?

hebasto:
Rebased to avoid CI failures.

real-or-random (approved):
utACK 5b7bf2e9d4ee02cbec1105ad6e890c34a4da1beb

> In addition to ...

Sorry, I missed the existing test.

real-or-random (approved):
ACK 5b7bf2e9d4ee02cbec1105ad6e890c34a4da1beb tested with MSVC x64

PR #1337, opened 2023-06-02
ci: Fix error D8037 in `cl.exe` (attempt 2)

hebasto:
Since the https://github.com/mstorsjo/msvc-wine/commit/2146cbfaf037e21de56c7157ec40bb6372860f51, the `msvc-wine` effectively initializes the WINE prefix when running the `install.sh` script. See [`install.sh`#L143](https://github.com/mstorsjo/msvc-wine/blob/2146cbfaf037e21de56c7157ec40bb6372860f51/install.sh#L143): 
```sh
    WINEDEBUG=-all wine64 wineboot &>/dev/null
```

Our following `wine64 wineboot --init` just messes up with the prefix.

This PR fixes this issue.

Also https://github.com/bitcoin-core/secp256k1/pull/1327 has been reverted as apparently it does not work. And https://github.com/bitcoin-core/secp256k1/pull/1320 has been combined into this one.

### DIFF
--- a/ci/linux-debian.Dockerfile
+++ b/ci/linux-debian.Dockerfile
@@ -29,9 +29,10 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     git clone https://github.com/mstorsjo/msvc-wine && \
     mkdir /opt/msvc && \
     python3 msvc-wine/vsdownload.py --accept-license --dest /opt/msvc Microsoft.VisualStudio.Workload.VCTools && \
-    msvc-wine/install.sh /opt/msvc
-
-# Initialize the wine environment. Wait until the wineserver process has
-# exited before closing the session, to avoid corrupting the wine prefix.
-RUN wine64 wineboot --init && \
+# Since commit 2146cbfaf037e21de56c7157ec40bb6372860f51, the
+# msvc-wine effectively initializes the wine prefix when running
+# the install.sh script.
+    msvc-wine/install.sh /opt/msvc && \
+# Wait until the wineserver process has exited before closing the session,
+# to avoid corrupting the wine prefix.
     while (ps -A | grep wineserver) > /dev/null; do sleep 1; done
--- a/ci/linux-debian.Dockerfile
+++ b/ci/linux-debian.Dockerfile
@@ -31,9 +31,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     python3 msvc-wine/vsdownload.py --accept-license --dest /opt/msvc Microsoft.VisualStudio.Workload.VCTools && \
     msvc-wine/install.sh /opt/msvc
 
-# Moving the wine prefix to /tmp avoids error D8037 when invoking cl.exe.
-ENV WINEPREFIX=/tmp/wineprefix
-# Initialize the wine prefix. Wait until the wineserver process has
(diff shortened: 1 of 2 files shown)


real-or-random:
Given this error message (see https://cirrus-ci.com/task/4836636333178880?logs=test#L7)
```
wineserver: chdir to /tmp/wineprefix: No such file or directory
```
I assume, the problem is rather that `/tmp` is not persistent in Docker on Cirrus CI. So the Docker build creates a new wineprefix in `/tmp/wineprefix`, but it's not persistent, so when CI runs on docker, the directory is gone already.

When I run `WINEPREFIX=/tmp/wineprefix wineserver -p` locally without having created this directory, it fails with the same error message. So maybe a proper solution is to move WINEPREFIX to some other directory than the default `~/wine`, but not on `/tmp`. (Or recreate the wineprefix in every CI run, but that's slower.) 


> To work around the current issue and to prevent similar issues in the future, this PR introduces the CI_MSVC_WINE_COMMIT Docker argument that binds to the known working commit in the [mstorsjo/msvc-wine](https://github.com/mstorsjo/msvc-wine) repo.

I think we should avoid pinning unless it's really really necessary. We want to get MSVC updates.

hebasto:
@real-or-random 

I'll post a real fix in a second :)

real-or-random:
Can you try moving WINEPREFIX to `/var/lib/wine` for example?

hebasto:
Reworked.

> I think we should avoid pinning unless it's really really necessary. We want to get MSVC updates.

Done.

real-or-random:
> Reworked.

Ok, let's see if this does the job. :)

ci/linux-debian.Dockerfile
```diff
@@ -28,12 +28,11 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
...
-# Moving the wine prefix to /tmp avoids error D8037 when invoking cl.exe.
-ENV WINEPREFIX=/tmp/wineprefix
-# Initialize the wine prefix. Wait until the wineserver process has
+# Since commit 2146cbfaf037e21de56c7157ec40bb6372860f51, the
+# msvc-wine effectively initializes the wine prefix when running
+# the install.sh script. Wait until the wineserver process has
 # exited before closing the session, to avoid corrupting the wine prefix.
-RUN wine64 wineboot --init && \
+RUN msvc-wine/install.sh /opt/msvc && \
     while (ps -A | grep wineserver) > /dev/null; do sleep 1; done
```
real-or-random, line 38: Nit: maybe this should be merged with the `RUN` command above.

(but let's first see if it works at all :) nice catch!)

real-or-random (commented):
Concept ACK

hebasto:
Addressed @real-or-random's https://github.com/bitcoin-core/secp256k1/pull/1337#discussion_r1214891014.

ci/linux-debian.Dockerfile
hebasto: > edit: This is my suggestion

Thanks! [Updated](https://github.com/bitcoin-core/secp256k1/pull/1337#issuecomment-1574769542).

real-or-random (approved):
ACK db29bf220c73bc71b9d48c262ba2e88c1e741f92

PR #1338, opened 2023-06-06
Drop no longer needed `#include "../include/secp256k1.h"`

hebasto:
The removed header includes have not been needed since https://github.com/bitcoin-core/secp256k1/pull/1231.

Test suggestions:
1. Using Autotools-based build system:
```
./autogen.sh
./configure
make clean-precomp
make
```
2. Using CMake-based build system:
```
cmake -B build -DCMAKE_C_INCLUDE_WHAT_YOU_USE="include-what-you-use"
cmake --build build --target secp256k1_precomputed
```


real-or-random (approved):
utACK e449af6872445d33a0796224fcb733be6476ad36

PR #1093, opened 2022-03-25
hash: Make code agnostic of endianness

real-or-random:
Recent compilers compile the two new functions to very efficient code
on various platforms. In particular, already GCC >= 5 and clang >= 4
understand do this for the read function, which is the one critical
for performance (called 16 times per SHA256 transform).

Fixes #1080.

PR #1339, opened 2023-06-10
scalar: refactor: use `secp256k1_{read,write}_be32` helpers

theStack:
This refactoring PR takes use of the `secp256k1_{read,write}_be32` helpers (introduced in PR #1093, commit 8d89b9e6e562000cdb91a70a85fae5e4817cec8a) in the scalar <-> byte array conversion functions, for both the 8x32 and 4x64 implementations. (An alternative for the latter would be to introduce special helpers for reading/writing uint64_t in big endian `secp256k1_{read,write}_be64`).

Verified via `objdump -D libsecp256k1.a` that `secp256k1_scalar_set_b32` for 4x64 compiles to the same code on master and the PR (`secp256k1_scalar_get_b32` is apparently always inlined) on amd64 with clang 13.0.0.


real-or-random (approved):
utACK 52b84238de27889f2517c845757bf5c2c65ed1d6

PR #1340, opened 2023-06-11
clean up in-comment Sage code (refer to secp256k1_params.sage, update to Python3)

theStack:
Some of the C source files contain contain in-comment Sage code calculating secp256k1 parameters that are already defined in the file secp256k1_params.sage.  Replace that by a corresponding load instruction and access the necessary variables. In ecdsa_impl.h, update the comment to use a one-line shell command calling sage to get the values.

The remaining code (test `test_add_neg_y_diff_x` in tests.c) is updated to work with a current version based on Python3 (Sage 9.0+, see https://wiki.sagemath.org/Python3-Switch).

The latter can be seen as a small follow-up to PR #849 (commit 13c88efed0005eb6745a222963ee74564054eafb).

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -4009,22 +4009,15 @@ static void test_add_neg_y_diff_x(void) {
      * which this test is a regression test for.
      *
      * These points were generated in sage as
-     * # secp256k1 params
-     * F = FiniteField (0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F)
-     * C = EllipticCurve ([F (0), F (7)])
-     * G = C.lift_x(0x79BE667EF9DCBBAC55A06295CE870B07029BFCDB2DCE28D959F2815B16F81798)
-     * N = FiniteField(G.order())
      *
-     * # endomorphism values (lambda is 1^{1/3} in N, beta is 1^{1/3} in F)
-     * x = polygen(N)
-     * lam  = (1 - x^3).roots()[1][0]
+     * load("secp256k1_params.sage")
      *
      * # random "bad pair"
      * P = C.random_element()
-     * Q = -int(lam) * P
-     * print "    P: %x %x" % P.xy()
-     * print "    Q: %x %x" % Q.xy()
-     * print "P + Q: %x %x" % (P + Q).xy()
+     * Q = -int(LAMBDA) * P
+     * print("    P: %x %x" % P.xy())
+     * print("    Q: %x %x" % Q.xy())
+     * print("P + Q: %x %x" % (P + Q).xy())
      */
     secp256k1_gej aj = SECP256K1_GEJ_CONST(
         0x8d24cd95, 0x0a355af1, 0x3c543505, 0x44238d30,
--- a/src/ecdsa_impl.h
+++ b/src/ecdsa_impl.h
@@ -16,17 +16,8 @@
 #include "ecdsa.h"
 
 /** Group order for secp256k1 defined as 'n' in "Standards for Efficient Cryptography" (SEC2) 2.7.1
- *  sage: for t in xrange(1023, -1, -1):
- *     ..   p = 2**256 - 2**32 - t
- *     ..   if p.is_prime():
- *     ..     print '%x'%p
- *     ..     break
- *   'fffffffffffffffffffffffffffffffffffffffffffffffffffffffefffffc2f'
- *  sage: a = 0
- *  sage: b = 7
- *  sage: F = FiniteField (p)
- *  sage: '%x' % (EllipticCurve ([F (a), F (b)]).order())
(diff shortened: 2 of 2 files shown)


real-or-random (commented):
Thanks for caring about this!

I think the sage code for `secp256k1_ecdsa_const_order_as_fe` and also `secp256k1_ecdsa_const_p_minus_order` could be dropped entirely. They're easy to write given https://github.com/bitcoin-core/secp256k1/blob/master/sage/secp256k1_params.sage#L17. Or even better, those values can just be defined in the `secp256k1_params.sage` file, which is run on CI to avoid code rot.

Not sure what do to with the code in the test. I haven't checked. Are these the only instances in the entire code base?

theStack:
> I think the sage code for `secp256k1_ecdsa_const_order_as_fe` and also `secp256k1_ecdsa_const_p_minus_order` could be dropped entirely. They're easy to write given https://github.com/bitcoin-core/secp256k1/blob/master/sage/secp256k1_params.sage#L17. Or even better, those values can just be defined in the `secp256k1_params.sage` file, which is run on CI to avoid code rot.
 
Sounds like a good idea. By the way, I found this approach for calculating p interesting (i.e. "find the first number equal or lower than 2^256 - 2^32 - 1023 which is prime"):
https://github.com/bitcoin-core/secp256k1/blob/67214f5f7d8e0bdf193ceb1f47ba8277d1a0871a/src/ecdsa_impl.h#L18-L23

But I guess there's no good reason to keep that code and having the fixed p number (as it is now in `secp256k1_params.sage`) is just fine? The referenced SEC2 document also doesn't mention anything about the above approach. 🤷‍♂️ 

> Not sure what do to with the code in the test. I haven't checked. Are these the only instances in the entire code base?

I think that's the only instances. My approach was manually looking through all the results of `$ git grep -i sage` (which included many false-positives due to the word "message") -- of course, it could still be possible that I either missed something, or there is sage code around without the word "sage" in a description close.

real-or-random:
> But I guess there's no good reason to keep that code and having the fixed p number (as it is now in `secp256k1_params.sage`) is just fine?

Yeah. If anything, the definition in SEC2 is canonical, and it just gives the number as you point out. 

---

For the code in the tests, perhaps that can at least be shortened to assume that the constants from https://github.com/bitcoin-core/secp256k1/blob/master/sage/secp256k1_params.sage are already defined.

theStack:
I've removed all in-comment sage code for calculating secp256k1 parameters and referred to `secp256k1_params.sage` instead. For ecdsa_impl.h, the comment is now a shell command which calls sage to emit the necessary values. There's still a small chance for code rot, though I'd argue it's neglible -- the chance that the filename `secp256k1_params.sage` ever changes or that the basic parameters `P` and `N` are renamed is rather small I'd say. Still open for alternative approaches.

src/tests.c
```diff
@@ -4009,22 +4009,19 @@ static void test_add_neg_y_diff_x(void) {
...
      * These points were generated in sage as
-     * # secp256k1 params
-     * F = FiniteField (0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEFFFFFC2F)
-     * C = EllipticCurve ([F (0), F (7)])
-     * G = C.lift_x(0x79BE667EF9DCBBAC55A06295CE870B07029BFCDB2DCE28D959F2815B16F81798)
-     * N = FiniteField(G.order())
+     *
+     * load("secp256k1_params.sage")
      *
      * # endomorphism values (lambda is 1^{1/3} in N, beta is 1^{1/3} in F)
```
real-or-random, line 4015: comment needs to be updated, I think

real-or-random:
> There's still a small chance for code rot, though I'd argue it's negligible -- the chance that the filename `secp256k1_params.sage` ever changes or that the basic parameters `P` and `N` are renamed is rather small I'd say. Still open for alternative approaches.

certainly good enough :)

src/tests.c
theStack: Done (assumed that you only meant the rename from N to Z).

theStack:
I only realized now that the code for calculating the lambda value in `test_add_neg_y_diff_x` can also be removed, as it's equivalent with `LAMBDA` from secp256k1_params.sage (i.e. given order N and field Z for secp256k1, the sage expressions `Z(3)^((N-1)/3)` and `(1 - (polygen(Z))^3).roots()[1][0]` lead to the same value):
```sage
sage: load("secp256k1_params.sage")
sage: Z(3)^((N-1)/3)
37718080363155996902926221483475020450927657555482586988616620542887997980018
sage: (1 - (polygen(Z))^3).roots()[1][0]
37718080363155996902926221483475020450927657555482586988616620542887997980018
sage: LAMBDA
37718080363155996902926221483475020450927657555482586988616620542887997980018
```
Force-pushed accordingly.
(For a rough explanation why these two expression are equivalent, I'd be more than thankful :))

real-or-random (approved):
ACK 600c5adcd59240305e22918943f45dceeabb7e93

PR #1341, opened 2023-06-11
docs: correct `pubkey` param descriptions for `secp256k1_keypair_{xonly_,}pub`

theStack:
From an API perspective, the functions `secp256k1_keypair_pub` and `secp256k1_keypair_xonly_pub` always succeed (i.e. return the value 1), so the other cases in the `pubkey` parameter descriptions never happen and can hence be removed.

Note that the "1 always" return value description was previously done in commit https://github.com/bitcoin-core/secp256k1/commit/b8f8b99f0fb3a5cd4c6fb1c9c8dfed881839e19e (PR https://github.com/bitcoin-core/secp256k1/pull/1089), which also explains why invalid inputs for the affected functions are in practice only possible in violation of the type system.

### DIFF
--- a/include/secp256k1_extrakeys.h
+++ b/include/secp256k1_extrakeys.h
@@ -185,9 +185,8 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_keypair_sec(
 /** Get the public key from a keypair.
  *
  *  Returns: 1 always.
- *  Args:    ctx: pointer to a context object.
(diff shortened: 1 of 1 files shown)


real-or-random (approved):
ACK https://github.com/bitcoin-core/secp256k1/pull/1341/commits/f3644287b10cd7339fbb0e35e953235351f5efb6 because it's consistent with the other docs

jonasnick (commented):
ACK f3644287b10cd7339fbb0e35e953235351f5efb6

PR #1347, opened 2023-06-15
field: Document return value of fe_sqrt()

real-or-random:


real-or-random:
@stratospher Interested in reviewing this small PR?

src/field.h
```diff
@@ -267,8 +267,10 @@ static void secp256k1_fe_sqr(secp256k1_fe *r, const secp256k1_fe *a);
 /** Compute a square root of a field element.
  *
  * On input, a must be a valid field element with magnitude<=8; r need not be initialized.
- * Performs {r = sqrt(a)} or {r = sqrt(-a)}, whichever exists. The resulting value
- * represented by r will be a square itself. Variables r and a must not point to the same object.
+ * Performs {r = sqrt(a)} and returns 1 if sqrt(a) exists.
+ * Performs {r = sqrt(-a)} and returns 0 otherwise; then sqrt(-a) exists.
```
jonasnick, line 271: nit: maybe case distinction like this is clearer?
```
 * If sqrt(a) exists, performs {r = sqrt(a)} and returns 1.
 * Otherwise, sqrt(-a) exists. The function performs {r = sqrt(-a)} and returns 0.
```

real-or-random: Done.

theStack (approved):
ACK 57791374579595bbc5d8d2e55f7fa0bc40d861a4

PR #1348, opened 2023-06-15
tighten group magnitude limits, save normalize_weak calls in group add methods (revival of #1032)

theStack:
This PR picks up #1032 by peterdettman. It's essentially a rebase on master; the original first commit (09dbba561fdb9d57a2cc9842ce041d9ba29a6189) which introduced group verification methods has mostly been replaced by PR #1299 (commit f20266722ac93ca66d1beb0d2f2d2469b95aafea) and what remains now is only adding a few missing checks at some places. The remaining commits are unchanged, though some (easy-to-solve) conflicts appeared through cherry-picking. The last commit which actually removes the `normalize_weak` calls is obviously the critical one and needs the most attention for review.

### DIFF
--- a/src/field.h
+++ b/src/field.h
@@ -352,4 +352,7 @@ static int secp256k1_fe_is_square_var(const secp256k1_fe *a);
 /** Check invariants on a field element (no-op unless VERIFY is enabled). */
 static void secp256k1_fe_verify(const secp256k1_fe *a);
 
+/** Check that magnitude of a is at most m (no-op unless VERIFY is enabled). */
+static void secp256k1_fe_verify_magnitude(const secp256k1_fe *a, int m);
+
 #endif /* SECP256K1_FIELD_H */
--- a/src/group.h
+++ b/src/group.h
@@ -46,11 +46,11 @@ typedef struct {
 
 /** Maximum allowed magnitudes for group element coordinates
  *  in affine (x, y) and jacobian (x, y, z) representation. */
-#define SECP256K1_GE_X_MAGNITUDE_MAX  8
-#define SECP256K1_GE_Y_MAGNITUDE_MAX  8
-#define SECP256K1_GEJ_X_MAGNITUDE_MAX 8
-#define SECP256K1_GEJ_Y_MAGNITUDE_MAX 8
-#define SECP256K1_GEJ_Z_MAGNITUDE_MAX 8
+#define SECP256K1_GE_X_MAGNITUDE_MAX  4
+#define SECP256K1_GE_Y_MAGNITUDE_MAX  3
+#define SECP256K1_GEJ_X_MAGNITUDE_MAX 4
+#define SECP256K1_GEJ_Y_MAGNITUDE_MAX 4
+#define SECP256K1_GEJ_Z_MAGNITUDE_MAX 1
 
 /** Set a group element equal to the point with given X and Y coordinates */
 static void secp256k1_ge_set_xy(secp256k1_ge *r, const secp256k1_fe *x, const secp256k1_fe *y);
--- a/src/group.h
+++ b/src/group.h
@@ -44,6 +44,14 @@ typedef struct {
 
 #define SECP256K1_GE_STORAGE_CONST_GET(t) SECP256K1_FE_STORAGE_CONST_GET(t.x), SECP256K1_FE_STORAGE_CONST_GET(t.y)
 
+/** Maximum allowed magnitudes for group element coordinates
+ *  in affine (x, y) and jacobian (x, y, z) representation. */
+#define SECP256K1_GE_X_MAGNITUDE_MAX  8
+#define SECP256K1_GE_Y_MAGNITUDE_MAX  8
+#define SECP256K1_GEJ_X_MAGNITUDE_MAX 8
+#define SECP256K1_GEJ_Y_MAGNITUDE_MAX 8
+#define SECP256K1_GEJ_Z_MAGNITUDE_MAX 8
+
 /** Set a group element equal to the point with given X and Y coordinates */
 static void secp256k1_ge_set_xy(secp256k1_ge *r, const secp256k1_fe *x, const secp256k1_fe *y);
 
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -77,6 +77,8 @@ static void secp256k1_ge_verify(const secp256k1_ge *a) {
 #ifdef VERIFY
     secp256k1_fe_verify(&a->x);
     secp256k1_fe_verify(&a->y);
+    secp256k1_fe_verify_magnitude(&a->x, SECP256K1_GE_X_MAGNITUDE_MAX);
+    secp256k1_fe_verify_magnitude(&a->y, SECP256K1_GE_Y_MAGNITUDE_MAX);
     VERIFY_CHECK(a->infinity == 0 || a->infinity == 1);
 #endif
     (void)a;
@@ -87,6 +89,9 @@ static void secp256k1_gej_verify(const secp256k1_gej *a) {
     secp256k1_fe_verify(&a->x);
     secp256k1_fe_verify(&a->y);
     secp256k1_fe_verify(&a->z);
+    secp256k1_fe_verify_magnitude(&a->x, SECP256K1_GEJ_X_MAGNITUDE_MAX);
+    secp256k1_fe_verify_magnitude(&a->y, SECP256K1_GEJ_Y_MAGNITUDE_MAX);
+    secp256k1_fe_verify_magnitude(&a->z, SECP256K1_GEJ_Z_MAGNITUDE_MAX);
     VERIFY_CHECK(a->infinity == 0 || a->infinity == 1);
 #endif
     (void)a;
@@ -358,7 +363,6 @@ static int secp256k1_gej_eq_x_var(const secp256k1_fe *x, const secp256k1_gej *a)
     secp256k1_fe_verify(x);
     secp256k1_gej_verify(a);
 #ifdef VERIFY
-    VERIFY_CHECK(a->x.magnitude <= 31);
     VERIFY_CHECK(!a->infinity);
 #endif
 
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -99,11 +99,13 @@ static void secp256k1_ge_set_gej_zinv(secp256k1_ge *r, const secp256k1_gej *a, c
     secp256k1_gej_verify(a);
     secp256k1_fe_verify(zi);
     VERIFY_CHECK(!a->infinity);
+
     secp256k1_fe_sqr(&zi2, zi);
     secp256k1_fe_mul(&zi3, &zi2, zi);
     secp256k1_fe_mul(&r->x, &a->x, &zi2);
     secp256k1_fe_mul(&r->y, &a->y, &zi3);
     r->infinity = a->infinity;
+
     secp256k1_ge_verify(r);
 }
 
@@ -114,39 +116,47 @@ static void secp256k1_ge_set_ge_zinv(secp256k1_ge *r, const secp256k1_ge *a, con
     secp256k1_ge_verify(a);
     secp256k1_fe_verify(zi);
     VERIFY_CHECK(!a->infinity);
+
     secp256k1_fe_sqr(&zi2, zi);
     secp256k1_fe_mul(&zi3, &zi2, zi);
     secp256k1_fe_mul(&r->x, &a->x, &zi2);
     secp256k1_fe_mul(&r->y, &a->y, &zi3);
     r->infinity = a->infinity;
+
     secp256k1_ge_verify(r);
 }
 
 static void secp256k1_ge_set_xy(secp256k1_ge *r, const secp256k1_fe *x, const secp256k1_fe *y) {
     secp256k1_fe_verify(x);
     secp256k1_fe_verify(y);
+
     r->infinity = 0;
     r->x = *x;
     r->y = *y;
+
     secp256k1_ge_verify(r);
 }
 
 static int secp256k1_ge_is_infinity(const secp256k1_ge *a) {
     secp256k1_ge_verify(a);
+
     return a->infinity;
 }
 
 static void secp256k1_ge_neg(secp256k1_ge *r, const secp256k1_ge *a) {
     secp256k1_ge_verify(a);
+
     *r = *a;
     secp256k1_fe_normalize_weak(&r->y);
     secp256k1_fe_negate(&r->y, &r->y, 1);
+
     secp256k1_ge_verify(r);
 }
 
 static void secp256k1_ge_set_gej(secp256k1_ge *r, secp256k1_gej *a) {
     secp256k1_fe z2, z3;
     secp256k1_gej_verify(a);
+
     r->infinity = a->infinity;
     secp256k1_fe_inv(&a->z, &a->z);
     secp256k1_fe_sqr(&z2, &a->z);
@@ -156,12 +166,15 @@ static void secp256k1_ge_set_gej(secp256k1_ge *r, secp256k1_gej *a) {
     secp256k1_fe_set_int(&a->z, 1);
     r->x = a->x;
     r->y = a->y;
+
+    secp256k1_gej_verify(a);
     secp256k1_ge_verify(r);
 }
 
 static void secp256k1_ge_set_gej_var(secp256k1_ge *r, secp256k1_gej *a) {
     secp256k1_fe z2, z3;
     secp256k1_gej_verify(a);
+
     if (secp256k1_gej_is_infinity(a)) {
         secp256k1_ge_set_infinity(r);
         return;
@@ -174,6 +187,8 @@ static void secp256k1_ge_set_gej_var(secp256k1_ge *r, secp256k1_gej *a) {
     secp256k1_fe_mul(&a->y, &a->y, &z3);
     secp256k1_fe_set_int(&a->z, 1);
     secp256k1_ge_set_xy(r, &a->x, &a->y);
+
+    secp256k1_gej_verify(a);
     secp256k1_ge_verify(r);
 }
 
@@ -181,9 +196,13 @@ static void secp256k1_ge_set_all_gej_var(secp256k1_ge *r, const secp256k1_gej *a
     secp256k1_fe u;
     size_t i;
     size_t last_i = SIZE_MAX;
-
+#ifdef VERIFY
     for (i = 0; i < len; i++) {
         secp256k1_gej_verify(&a[i]);
+    }
+#endif
+
+    for (i = 0; i < len; i++) {
         if (a[i].infinity) {
             secp256k1_ge_set_infinity(&r[i]);
         } else {
@@ -217,36 +236,46 @@ static void secp256k1_ge_set_all_gej_var(secp256k1_ge *r, const secp256k1_gej *a
         if (!a[i].infinity) {
             secp256k1_ge_set_gej_zinv(&r[i], &a[i], &r[i].x);
         }
+    }
+
+#ifdef VERIFY
+    for (i = 0; i < len; i++) {
         secp256k1_ge_verify(&r[i]);
     }
+#endif
 }
 
 static void secp256k1_ge_table_set_globalz(size_t len, secp256k1_ge *a, const secp256k1_fe *zr) {
-    size_t i = len - 1;
+    size_t i;
     secp256k1_fe zs;
-
-    if (len > 0) {
-        /* Verify inputs a[len-1] and zr[len-1]. */
+#ifdef VERIFY
+    for (i = 0; i < len; i++) {
         secp256k1_ge_verify(&a[i]);
         secp256k1_fe_verify(&zr[i]);
+    }
+#endif
+
+    if (len > 0) {
+        i = len - 1;
         /* Ensure all y values are in weak normal form for fast negation of points */
         secp256k1_fe_normalize_weak(&a[i].y);
         zs = zr[i];
 
         /* Work our way backwards, using the z-ratios to scale the x/y values. */
         while (i > 0) {
-            /* Verify all inputs a[i] and zr[i]. */
-            secp256k1_fe_verify(&zr[i]);
-            secp256k1_ge_verify(&a[i]);
             if (i != len - 1) {
                 secp256k1_fe_mul(&zs, &zs, &zr[i]);
             }
             i--;
             secp256k1_ge_set_ge_zinv(&a[i], &a[i], &zs);
-            /* Verify the output a[i]. */
-            secp256k1_ge_verify(&a[i]);
         }
     }
+
+#ifdef VERIFY
+    for (i = 0; i < len; i++) {
+        secp256k1_ge_verify(&a[i]);
+    }
+#endif
 }
 
 static void secp256k1_gej_set_infinity(secp256k1_gej *r) {
@@ -254,6 +283,7 @@ static void secp256k1_gej_set_infinity(secp256k1_gej *r) {
     secp256k1_fe_clear(&r->x);
     secp256k1_fe_clear(&r->y);
     secp256k1_fe_clear(&r->z);
+
     secp256k1_gej_verify(r);
 }
 
@@ -261,6 +291,7 @@ static void secp256k1_ge_set_infinity(secp256k1_ge *r) {
     r->infinity = 1;
     secp256k1_fe_clear(&r->x);
     secp256k1_fe_clear(&r->y);
+
     secp256k1_ge_verify(r);
 }
 
@@ -269,18 +300,23 @@ static void secp256k1_gej_clear(secp256k1_gej *r) {
     secp256k1_fe_clear(&r->x);
     secp256k1_fe_clear(&r->y);
     secp256k1_fe_clear(&r->z);
+
+    secp256k1_gej_verify(r);
 }
 
 static void secp256k1_ge_clear(secp256k1_ge *r) {
     r->infinity = 0;
     secp256k1_fe_clear(&r->x);
     secp256k1_fe_clear(&r->y);
+
+    secp256k1_ge_verify(r);
 }
 
 static int secp256k1_ge_set_xo_var(secp256k1_ge *r, const secp256k1_fe *x, int odd) {
     secp256k1_fe x2, x3;
     int ret;
     secp256k1_fe_verify(x);
+
     r->x = *x;
     secp256k1_fe_sqr(&x2, x);
     secp256k1_fe_mul(&x3, x, &x2);
@@ -291,16 +327,19 @@ static int secp256k1_ge_set_xo_var(secp256k1_ge *r, const secp256k1_fe *x, int o
     if (secp256k1_fe_is_odd(&r->y) != odd) {
         secp256k1_fe_negate(&r->y, &r->y, 1);
     }
+
     secp256k1_ge_verify(r);
     return ret;
 }
 
 static void secp256k1_gej_set_ge(secp256k1_gej *r, const secp256k1_ge *a) {
    secp256k1_ge_verify(a);
+
    r->infinity = a->infinity;
    r->x = a->x;
    r->y = a->y;
    secp256k1_fe_set_int(&r->z, 1);
+
    secp256k1_gej_verify(r);
 }
 
@@ -308,6 +347,7 @@ static int secp256k1_gej_eq_var(const secp256k1_gej *a, const secp256k1_gej *b) 
     secp256k1_gej tmp;
     secp256k1_gej_verify(b);
     secp256k1_gej_verify(a);
+
     secp256k1_gej_neg(&tmp, a);
     secp256k1_gej_add_var(&tmp, &tmp, b, NULL);
     return secp256k1_gej_is_infinity(&tmp);
@@ -315,11 +355,10 @@ static int secp256k1_gej_eq_var(const secp256k1_gej *a, const secp256k1_gej *b) 
 
 static int secp256k1_gej_eq_x_var(const secp256k1_fe *x, const secp256k1_gej *a) {
     secp256k1_fe r;
-
-#ifdef VERIFY
     secp256k1_fe_verify(x);
-    VERIFY_CHECK(a->x.magnitude <= 31);
     secp256k1_gej_verify(a);
+#ifdef VERIFY
+    VERIFY_CHECK(a->x.magnitude <= 31);
     VERIFY_CHECK(!a->infinity);
 #endif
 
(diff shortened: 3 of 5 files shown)


theStack:
Rebased on master (there was a conflict in tests.c due to #1358) and also updated the operations count comments on each of the optimized functions (i.e. `secp256k1_gej_add_ge_var`, `secp256k1_gej_add_zinv_var` and `secp256k1_gej_add_ge`).

src/field.h
```diff
@@ -352,4 +352,12 @@ static int secp256k1_fe_is_square_var(const secp256k1_fe *a);
 /** Check invariants on a field element (no-op unless VERIFY is enabled). */
 static void secp256k1_fe_verify(const secp256k1_fe *a);
 
+#ifdef VERIFY
+
+/** Assert a context-specific limit m on the magnitude of a. m cannot be outside
```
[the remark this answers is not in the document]
theStack: Moved `secp256k1_fe_verify_magnitude` to a (single) implementation in field_impl.h, also removed the VERIFY pre-processor condition from field.h and add a no-op implementation for the case that VERIFY isn't set instead (even though the no-op case isn't currently used, it seems to be more consistent to _fe_verify that way).

src/group_impl.h
```diff
@@ -231,6 +235,7 @@ static void secp256k1_ge_table_set_globalz(size_t len, secp256k1_ge *a, const se
         secp256k1_fe_verify(&zr[i]);
         /* Ensure all y values are in weak normal form for fast negation of points */
         secp256k1_fe_normalize_weak(&a[i].y);
+        secp256k1_ge_verify(&a[i]);
```
real-or-random, line 243: nit: I think the functions would be easier to read, if they all follow the same pattern (in particular the functions with loops:)
 * on entry, verify all input ge, gej
 * (possibly an empty line if you're willing to add this)
 * actual function body
 * (possibly an empty line if you're willing to add this)
 * on exit, verify all output gej, possibly separated by a newline

src/field.h
```diff
@@ -352,4 +352,8 @@ static int secp256k1_fe_is_square_var(const secp256k1_fe *a);
 /** Check invariants on a field element (no-op unless VERIFY is enabled). */
 static void secp256k1_fe_verify(const secp256k1_fe *a);
 
+/** Assert a context-specific limit m on the magnitude of a, where
+ *  m must be in the supported range [0,32] (no-op unless VERIFY is enabled). */
+static void secp256k1_fe_verify_magnitude(const secp256k1_fe *a, int m);
```
real-or-random, line 357: nits in the comment: 
 - s/Assert/Check (Assert is actually the better term, but we write check everywhere else for this. Or change it in the entire code base ;))
 - "context-specific" can be a bit unclear. And I don't think it adds much.
 - Maybe just: "Check that the magnitude of a is at most m. (...) "?

src/group_impl.h
```diff
@@ -89,9 +89,9 @@ static void secp256k1_gej_verify(const secp256k1_gej *a) {
     secp256k1_fe_verify(&a->x);
     secp256k1_fe_verify(&a->y);
     secp256k1_fe_verify(&a->z);
-    secp256k1_fe_verify_magnitude(&a->x, 8);
-    secp256k1_fe_verify_magnitude(&a->y, 8);
-    secp256k1_fe_verify_magnitude(&a->z, 8);
+    secp256k1_fe_verify_magnitude(&a->x, 6);
+    secp256k1_fe_verify_magnitude(&a->y, 4);
+    secp256k1_fe_verify_magnitude(&a->z, 2);
```
real-or-random, line 94: I wonder if it's better to have constants for this. If we change these values here, one needs to change the tests, too, and this is rather easy to miss.

Alternatively (and slightly worse if you ask me), you could add comments here and in the tests that just state that the values need to match these in the other file.

real-or-random:
I think some of the existing `VERIFY_CHECK`s in `field_impl.h` can be replaced by calls to `secp256k1_fe_verify_magnitude` (If that's desirable but I guess so?)

src/group_impl.h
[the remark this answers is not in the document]
theStack: Done, included the removal of this line in the last commit with a mention in the commit message.

src/group_impl.h
theStack: Agree! I've introduced #defines SECP256K1_{GE,GEJ}_{X,Y,Z}_MAGNITUDE_MAX in group.h and used that in the following commit, indeed much nicer now. Happy to receive improvement suggestions for the new constants naming and the place where they are defined.

src/field.h
theStack: Thanks, went with your suggestion.

src/group_impl.h
theStack: Started working on that but got doubt, as I think that would need additional verification _loops_ at the start and end of a function if the input is an array of ge(j) instances (like in secp256k1_ge_table_set_globalz). Is that what you meant? (If yes, maybe good follow-up material?)

theStack:
Force-pushed with most suggestions tackled.

> I think some of the existing `VERIFY_CHECK`s in `field_impl.h` can be replaced by calls to `secp256k1_fe_verify_magnitude` (If that's desirable but I guess so?)

Added another commit for doing that after the "secp256k1_ge_table_set_globalz" one ([203e41b86280d5ea5ed9ab599ddbc21970529601](https://github.com/bitcoin-core/secp256k1/pull/1348/commits/203e41b86280d5ea5ed9ab599ddbc21970529601)).

Bezzo100-lab (approved):

Bezzo100-lab (approved):

src/group_impl.h
real-or-random: > Is that what you meant?

Yes, this would decouple the verification from the actual code. (And loops could be in `#ifdef VERIFY`, though any sane compiler would remove the empty loop.)

> (If yes, maybe good follow-up material?)

Okay, fair. I'm happy with any way in the end.

My thinking was that the advantage of doing it in this PR is that it will be easier to check for reviewers that verification is present everywhere. (The diff of that commit will be large, but noone will need to read the diff; they can just check if the result catches everything.) But I'm convinced of this already because I did the work to go through the `group_impl.h`.

If you think a separate PR for this refactor is better, then yet another way to do this would be to have a separate PR *before this PR*. But this may require more rebasing work.

I believe you can decide best now that you looked into this.

theStack: > My thinking was that the advantage of doing it in this PR is that it will be easier to check for reviewers that verification is present everywhere. (The diff of that commit will be large, but noone will need to read the diff; they can just check if the result catches everything.)

That makes sense! Thanks for elaborating, the motivation is now clear to me. Added the suggestion in the first commit. Should be quite trivial to review, only the changes in `secp256k1_ge_table_set_globalz` are a bit more invasive, as the initialization of `i` for the backwards loop is now done later, due to `i` also being used for the verification loop of inputs before. But I think that's actually even a slight improvement because we avoid a potential integer underflow for `i` in the case of `len == 0` (even though that's more theoretical, in practice no-one would call that function with a zero-length I guess).

I decided to put the input/output verification loops in `#ifdef VERIFY`, mostly for visual reasons (I'd also assume that even without those, most compilers would optimize the loops away anyway), but also happy to remove them if wished.

real-or-random (commented):
Although that could be done in a separate PR, I think this ideally should also remove the restriction to magnitude 31 introduced here: https://github.com/bitcoin-core/secp256k1/pull/1344/commits/07c0e8b82e2cea87f85263512945fed7adffea18#diff-28d9b92418cbddbe487fe84847993cc5da0b096363c83631a525cff25189752cR104 This is anyway now implied by the new limits.

Possible follow-up PRs/issues: 
 - Some (but not all) of the group functions verify their input `fe`s.  That doesn't hurt, but it's also not necessary because all field functions do that anyway. Anyway, we should either always do it or never.
 - We should really think about `#ifdef VERIFY` now. It seems that the current convention is to put all new VERIFY_CHECKs in ifdefs.  If we do this anyway, then we may as well redefine VERIFY_CHECK to be a noop (and give up on the idea that this macro is "side-effect safe". Another consideration is readability the uppercase VERIFY stands out. With more functions like `ge_verify`, it's a bit hard to spot them, so maybe we should actually keep the ifdefs. Or rename the functions to fe_VERIFY.

src/field.h
```diff
@@ -352,4 +352,7 @@ static int secp256k1_fe_is_square_var(const secp256k1_fe *a);
 /** Check invariants on a field element (no-op unless VERIFY is enabled). */
 static void secp256k1_fe_verify(const secp256k1_fe *a);
 
+/** Check that magnitude of a ist at most m (no-op unless VERIFY is enabled). */
```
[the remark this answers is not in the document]
theStack: Oops, probably my mind was switching over to german for a moment when I typed this :sweat_smile: Fixed.

src/group_impl.h
```diff
@@ -750,34 +749,34 @@ static void secp256k1_gej_add_ge(secp256k1_gej *r, const secp256k1_gej *a, const
...
     rr_alt = s1;
-    secp256k1_fe_mul_int(&rr_alt, 2);       /* rr = Y1*Z2^3 - Y2*Z1^3 (2) */
-    secp256k1_fe_add(&m_alt, &u1);          /* Malt = X1*Z2^2 - X2*Z1^2 */
+    secp256k1_fe_mul_int(&rr_alt, 2);       /* rr_alt = Y1*Z2^3 - Y2*Z1^3 (8) */
+    secp256k1_fe_add(&m_alt, &u1);          /* Malt = X1*Z2^2 - X2*Z1^2 (8) */
 
-    secp256k1_fe_cmov(&rr_alt, &rr, !degenerate);
-    secp256k1_fe_cmov(&m_alt, &m, !degenerate);
+    secp256k1_fe_cmov(&rr_alt, &rr, !degenerate);       /* rr_alt (8) */
+    secp256k1_fe_cmov(&m_alt, &m, !degenerate);         /* m_alt (5) */
```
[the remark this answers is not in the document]
theStack: Fixed.

theStack:
Thanks for the review. Force-pushed, removed the <= 31 magnitude check in `secp256k1_gej_eq_x_var` (included in the commit which introduces the tighter group magnitude limits) and also tackled the two other smaller review comments above.

Some comments to the follow-up ideas:

> Some (but not all) of the group functions verify their input fes. That doesn't hurt, but it's also not necessary because all field functions do that anyway. Anyway, we should either always do it or never.

Any preference between "always" and "never"? Since the checks are not necessary, I would personally prefer "never", leading to less code, even though the diff would be a bit larger (there are more functions taking fe inputs in group_impl.h that already verify the fe than those that miss the check), but that shouldn't matter.

As for the `#ifdef VERIFY` subject, what exactly does "side-effect safe" mean in this context, is the point here to _allow_ side-effects, and that's why the expression also needs to be executed in non-verify mode? I fail to see why having side-effects in checks would ever be useful or generally desirable (quickly glancing over `git grep VERIFY_CHECK` results also didn't show me any expressions with side-effects, but maybe I'm missing something), so your proposal of redefining VERIFY_CHECK to a no-op in non-verify mode sounds reasonable to me.

real-or-random (approved):
ACK 31bc9a434811d9d2c0a8058c93957c365b39089b

real-or-random:
> > Some (but not all) of the group functions verify their input fes. That doesn't hurt, but it's also not necessary because all field functions do that anyway. Anyway, we should either always do it or never.
> 
> Any preference between "always" and "never"? Since the checks are not necessary, I would personally prefer "never", leading to less code, even though the diff would be a bit larger (there are more functions taking fe inputs in group_impl.h that already verify the fe than those that miss the check), but that shouldn't matter.
> 

I tend to agree. This will also separate concerns more clearly: The field code verifies the field element, the group code verifies the group elements. And if the group code accesses `fe` internals directly (e.g. `magnitude`), that should be considered a layer violation (and this PR here will fix this by adding `secp256k1_fe_verify_magnitude`).

> As for the `#ifdef VERIFY` subject, [...]

See #1381.

src/group_impl.h
```diff
@@ -501,11 +553,11 @@ static void secp256k1_gej_add_ge_var(secp256k1_gej *r, const secp256k1_gej *a, c
...
 
     secp256k1_fe_sqr(&z12, &a->z);
-    u1 = a->x; secp256k1_fe_normalize_weak(&u1);
+    u1 = a->x;
     secp256k1_fe_mul(&u2, &b->x, &z12);
-    s1 = a->y; secp256k1_fe_normalize_weak(&s1);
+    s1 = a->y;
     secp256k1_fe_mul(&s2, &b->y, &z12); secp256k1_fe_mul(&s2, &s2, &a->z);
-    secp256k1_fe_negate(&h, &u1, 1); secp256k1_fe_add(&h, &u2);
+    secp256k1_fe_negate(&h, &u1, 6); secp256k1_fe_add(&h, &u2);
```
jonasnick, line 560: Better to replace `6` with `SECP256K1_GE_X_MAGNITUDE_MAX`? Same for the `fe_negate` in `secp256k1_gej_add_zinv_var`.

src/group.h
```diff
@@ -44,6 +44,14 @@ typedef struct {
 
 #define SECP256K1_GE_STORAGE_CONST_GET(t) SECP256K1_FE_STORAGE_CONST_GET(t.x), SECP256K1_FE_STORAGE_CONST_GET(t.y)
 
+/** Maximum allowed magnitudes for group element coordinates
+ *  in affine (x, y) and jacobian (x, y, z) representation. */
+#define SECP256K1_GE_X_MAGNITUDE_MAX  6
+#define SECP256K1_GE_Y_MAGNITUDE_MAX  4
+#define SECP256K1_GEJ_X_MAGNITUDE_MAX 6
+#define SECP256K1_GEJ_Y_MAGNITUDE_MAX 4
+#define SECP256K1_GEJ_Z_MAGNITUDE_MAX 2
```
jonasnick, line 53: How did you arrive at these values? The tightest limits that let the tests pass are
```
#define SECP256K1_GE_X_MAGNITUDE_MAX  4
#define SECP256K1_GE_Y_MAGNITUDE_MAX  3
#define SECP256K1_GEJ_X_MAGNITUDE_MAX 4
#define SECP256K1_GEJ_Y_MAGNITUDE_MAX 4
#define SECP256K1_GEJ_Z_MAGNITUDE_MAX 1
```

theStack: Good question. I took these limits over from the original PR by @peterdettman (#1032) without further evaluation. Even on that PR, I verified that with the limits (4,3) and (4,4,1) the tests still pass, so no idea where the (6,4)/(6,4,2) limits were originally coming from. (My best guess would be that in the earliest version of the PR, those were the tightest limits where test would pass?!).

src/group_impl.h
```diff
@@ -672,17 +727,17 @@ static void secp256k1_gej_add_ge(secp256k1_gej *r, const secp256k1_gej *a, const
      */
 
     secp256k1_fe_sqr(&zz, &a->z);                       /* z = Z1^2 */
-    u1 = a->x; secp256k1_fe_normalize_weak(&u1);        /* u1 = U1 = X1*Z2^2 (1) */
+    u1 = a->x;                                          /* u1 = U1 = X1*Z2^2 (6) */
```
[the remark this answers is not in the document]
theStack: This sounds like a good idea, as it would avoid the tedious work of adapting the comments if the magnitude limits ever change in the future. On the other hand, the expressions get quite long at the end, e.g. `((GEJ_Y_M+3)/2 + 1)` for the last annotation in `secp256k1_gej_add_ge` (instruction `secp256k1_fe_half(&r->y);`), not sure if this is still that helpful for the reader in constrast to a concrete magnitude number.

theStack:
@jonasnick: Thanks for the review! I've tried to tackle your suggestions in a separate branch for review of how the proposed changes would look like (https://github.com/theStack/secp256k1/tree/revival_of_1032_magnitude_comments, in particular the last commit https://github.com/theStack/secp256k1/commit/b7c685e74adbd83937990e90f076600fabf8ccf0). This is the range-diff from the current version (which already has one ACK) to the one with tighter limits and magnitude annotations expressed with limit variables (`git range-diff 31bc9...b7c68`):

<details>
  <summary>Diff</summary>

```diff
1:  d4527d6 ! 1:  c83afa6 Tighten group magnitude limits
    @@ Commit message
         - adjust test methods that randomize magnitudes
     
         Co-authored-by: Sebastian Falbesoner <sebastian.falbesoner@gmail.com>
    +    Co-authored-by: Jonas Nick <jonasd.nick@gmail.com>
     
      ## src/group.h ##
     @@ src/group.h: typedef struct {
    @@ src/group.h: typedef struct {
     -#define SECP256K1_GEJ_X_MAGNITUDE_MAX 8
     -#define SECP256K1_GEJ_Y_MAGNITUDE_MAX 8
     -#define SECP256K1_GEJ_Z_MAGNITUDE_MAX 8
    -+#define SECP256K1_GE_X_MAGNITUDE_MAX  6
    -+#define SECP256K1_GE_Y_MAGNITUDE_MAX  4
    -+#define SECP256K1_GEJ_X_MAGNITUDE_MAX 6
    ++#define SECP256K1_GE_X_MAGNITUDE_MAX  4
    ++#define SECP256K1_GE_Y_MAGNITUDE_MAX  3
    ++#define SECP256K1_GEJ_X_MAGNITUDE_MAX 4
     +#define SECP256K1_GEJ_Y_MAGNITUDE_MAX 4
    -+#define SECP256K1_GEJ_Z_MAGNITUDE_MAX 2
    ++#define SECP256K1_GEJ_Z_MAGNITUDE_MAX 1
      
      /** Set a group element equal to the point with given X and Y coordinates */
      static void secp256k1_ge_set_xy(secp256k1_ge *r, const secp256k1_fe *x, const secp256k1_fe *y);
2:  31bc9a4 ! 2:  b7c685e Save _normalize_weak calls in group add methods
    @@ Commit message
     
         Co-authored-by: Sebastian Falbesoner <sebastian.falbesoner@gmail.com>
         Co-authored-by: Tim Ruffing <crypto@timruffing.de>
    +    Co-authored-by: Jonas Nick <jonasd.nick@gmail.com>
     
      ## src/group_impl.h ##
     @@ src/group_impl.h: static void secp256k1_gej_add_var(secp256k1_gej *r, const secp256k1_gej *a, cons
    @@ src/group_impl.h: static void secp256k1_gej_add_ge_var(secp256k1_gej *r, const s
     +    s1 = a->y;
          secp256k1_fe_mul(&s2, &b->y, &z12); secp256k1_fe_mul(&s2, &s2, &a->z);
     -    secp256k1_fe_negate(&h, &u1, 1); secp256k1_fe_add(&h, &u2);
    -+    secp256k1_fe_negate(&h, &u1, 6); secp256k1_fe_add(&h, &u2);
    ++    secp256k1_fe_negate(&h, &u1, SECP256K1_GEJ_X_MAGNITUDE_MAX); secp256k1_fe_add(&h, &u2);
          secp256k1_fe_negate(&i, &s2, 1); secp256k1_fe_add(&i, &s1);
          if (secp256k1_fe_normalizes_to_zero_var(&h)) {
              if (secp256k1_fe_normalizes_to_zero_var(&i)) {
    @@ src/group_impl.h: static void secp256k1_gej_add_zinv_var(secp256k1_gej *r, const
     +    s1 = a->y;
          secp256k1_fe_mul(&s2, &b->y, &z12); secp256k1_fe_mul(&s2, &s2, &az);
     -    secp256k1_fe_negate(&h, &u1, 1); secp256k1_fe_add(&h, &u2);
    -+    secp256k1_fe_negate(&h, &u1, 6); secp256k1_fe_add(&h, &u2);
    ++    secp256k1_fe_negate(&h, &u1, SECP256K1_GEJ_X_MAGNITUDE_MAX); secp256k1_fe_add(&h, &u2);
          secp256k1_fe_negate(&i, &s2, 1); secp256k1_fe_add(&i, &s1);
          if (secp256k1_fe_normalizes_to_zero_var(&h)) {
              if (secp256k1_fe_normalizes_to_zero_var(&i)) {
    @@ src/group_impl.h: static void secp256k1_gej_add_ge(secp256k1_gej *r, const secp2
      
          secp256k1_fe_sqr(&zz, &a->z);                       /* z = Z1^2 */
     -    u1 = a->x; secp256k1_fe_normalize_weak(&u1);        /* u1 = U1 = X1*Z2^2 (1) */
    -+    u1 = a->x;                                          /* u1 = U1 = X1*Z2^2 (6) */
    ++    u1 = a->x;                                          /* u1 = U1 = X1*Z2^2 (GEJ_X_M) */
          secp256k1_fe_mul(&u2, &b->x, &zz);                  /* u2 = U2 = X2*Z1^2 (1) */
     -    s1 = a->y; secp256k1_fe_normalize_weak(&s1);        /* s1 = S1 = Y1*Z2^3 (1) */
    -+    s1 = a->y;                                          /* s1 = S1 = Y1*Z2^3 (4) */
    ++    s1 = a->y;                                          /* s1 = S1 = Y1*Z2^3 (GEJ_Y_M) */
          secp256k1_fe_mul(&s2, &b->y, &zz);                  /* s2 = Y2*Z1^2 (1) */
          secp256k1_fe_mul(&s2, &s2, &a->z);                  /* s2 = S2 = Y2*Z1^3 (1) */
     -    t = u1; secp256k1_fe_add(&t, &u2);                  /* t = T = U1+U2 (2) */
     -    m = s1; secp256k1_fe_add(&m, &s2);                  /* m = M = S1+S2 (2) */
    -+    t = u1; secp256k1_fe_add(&t, &u2);                  /* t = T = U1+U2 (7) */
    -+    m = s1; secp256k1_fe_add(&m, &s2);                  /* m = M = S1+S2 (5) */
    ++    t = u1; secp256k1_fe_add(&t, &u2);                  /* t = T = U1+U2 (GEJ_X_M+1) */
    ++    m = s1; secp256k1_fe_add(&m, &s2);                  /* m = M = S1+S2 (GEJ_Y_M+1) */
          secp256k1_fe_sqr(&rr, &t);                          /* rr = T^2 (1) */
     -    secp256k1_fe_negate(&m_alt, &u2, 1);                /* Malt = -X2*Z1^2 */
     -    secp256k1_fe_mul(&tt, &u1, &m_alt);                 /* tt = -U1*U2 (2) */
    @@ src/group_impl.h: static void secp256k1_gej_add_ge(secp256k1_gej *r, const secp2
          rr_alt = s1;
     -    secp256k1_fe_mul_int(&rr_alt, 2);       /* rr = Y1*Z2^3 - Y2*Z1^3 (2) */
     -    secp256k1_fe_add(&m_alt, &u1);          /* Malt = X1*Z2^2 - X2*Z1^2 */
    -+    secp256k1_fe_mul_int(&rr_alt, 2);       /* rr_alt = Y1*Z2^3 - Y2*Z1^3 (8) */
    -+    secp256k1_fe_add(&m_alt, &u1);          /* Malt = X1*Z2^2 - X2*Z1^2 (8) */
    ++    secp256k1_fe_mul_int(&rr_alt, 2);       /* rr_alt = Y1*Z2^3 - Y2*Z1^3 (GEJ_Y_M*2) */
    ++    secp256k1_fe_add(&m_alt, &u1);          /* Malt = X1*Z2^2 - X2*Z1^2 (GEJ_X_M+2) */
      
     -    secp256k1_fe_cmov(&rr_alt, &rr, !degenerate);
     -    secp256k1_fe_cmov(&m_alt, &m, !degenerate);
    -+    secp256k1_fe_cmov(&rr_alt, &rr, !degenerate);       /* rr_alt (8) */
    -+    secp256k1_fe_cmov(&m_alt, &m, !degenerate);         /* m_alt (8) */
    ++    secp256k1_fe_cmov(&rr_alt, &rr, !degenerate);       /* rr_alt (GEJ_Y_M*2) */
    ++    secp256k1_fe_cmov(&m_alt, &m, !degenerate);         /* m_alt (GEJ_X_M+2) */
          /* Now Ralt / Malt = lambda and is guaranteed not to be Ralt / 0.
           * From here on out Ralt and Malt represent the numerator
           * and denominator of lambda; R and M represent the explicit
           * expressions x1^2 + x2^2 + x1x2 and y1 + y2. */
          secp256k1_fe_sqr(&n, &m_alt);                       /* n = Malt^2 (1) */
     -    secp256k1_fe_negate(&q, &t, 2);                     /* q = -T (3) */
    -+    secp256k1_fe_negate(&q, &t, 7);                     /* q = -T (8) */
    ++    secp256k1_fe_negate(&q, &t,
    ++        SECP256K1_GEJ_X_MAGNITUDE_MAX + 1);             /* q = -T (GEJ_X_M+2) */
          secp256k1_fe_mul(&q, &q, &n);                       /* q = Q = -T*Malt^2 (1) */
          /* These two lines use the observation that either M == Malt or M == 0,
           * so M^3 * Malt is either Malt^4 (which is computed by squaring), or
    @@ src/group_impl.h: static void secp256k1_gej_add_ge(secp256k1_gej *r, const secp2
     -    secp256k1_fe_sqr(&n, &n);
     -    secp256k1_fe_cmov(&n, &m, degenerate);              /* n = M^3 * Malt (2) */
     +    secp256k1_fe_sqr(&n, &n);                           /* n = Malt^4 (1) */
    -+    secp256k1_fe_cmov(&n, &m, degenerate);              /* n = M^3 * Malt (5) */
    ++    secp256k1_fe_cmov(&n, &m, degenerate);              /* n = M^3 * Malt (GEJ_Y_M+1) */
          secp256k1_fe_sqr(&t, &rr_alt);                      /* t = Ralt^2 (1) */
          secp256k1_fe_mul(&r->z, &a->z, &m_alt);             /* r->z = Z3 = Malt*Z (1) */
          secp256k1_fe_add(&t, &q);                           /* t = Ralt^2 + Q (2) */
    @@ src/group_impl.h: static void secp256k1_gej_add_ge(secp256k1_gej *r, const secp2
     -    secp256k1_fe_add(&t, &n);                           /* t = Ralt*(2*X3 + Q) + M^3*Malt (3) */
     -    secp256k1_fe_negate(&r->y, &t, 3);                  /* r->y = -(Ralt*(2*X3 + Q) + M^3*Malt) (4) */
     -    secp256k1_fe_half(&r->y);                           /* r->y = Y3 = -(Ralt*(2*X3 + Q) + M^3*Malt)/2 (3) */
    -+    secp256k1_fe_add(&t, &n);                           /* t = Ralt*(2*X3 + Q) + M^3*Malt (6) */
    -+    secp256k1_fe_negate(&r->y, &t, 6);                  /* r->y = -(Ralt*(2*X3 + Q) + M^3*Malt) (7) */
    -+    secp256k1_fe_half(&r->y);                           /* r->y = Y3 = -(Ralt*(2*X3 + Q) + M^3*Malt)/2 (4) */
    ++    secp256k1_fe_add(&t, &n);                           /* t = Ralt*(2*X3 + Q) + M^3*Malt (GEJ_Y_M+2) */
    ++    secp256k1_fe_negate(&r->y, &t,
    ++        SECP256K1_GEJ_Y_MAGNITUDE_MAX + 2);             /* r->y = -(Ralt*(2*X3 + Q) + M^3*Malt) (GEJ_Y_M+3) */
    ++    secp256k1_fe_half(&r->y);                           /* r->y = Y3 = -(Ralt*(2*X3 + Q) + M^3*Malt)/2 ((GEJ_Y_M+3)/2 + 1) */
      
          /* In case a->infinity == 1, replace r with (b->x, b->y, 1). */
          secp256k1_fe_cmov(&r->x, &b->x, a->infinity);
```

</details>

Any opinions on what of those changes should be adapted? On one hand I think it's nice to express the magnitudes in a generic way in the comments, on the other hand it probably causes more confusion and is not that helpful (especially for longer expressions). Also, is there any drawback in tightening the limits too much, e.g. increased risk that they have to be lifted again soon, causing extra work? Happy to receive input and change the PR either way.

src/group.h
peterdettman: > Good question. I took these limits over from the original PR by @peterdettman (#1032) without further evaluation. Even on that PR, I verified that with the limits (4,3) and (4,4,1) the tests still pass, so no idea where the (6,4)/(6,4,2) limits were originally coming from. (My best guess would be that in the earliest version of the PR, those were the tightest limits where test would pass?!).

There is an inherent tension between tight limits, which might enable optimizations at method entry, and more generous limits, which might enable optimizations at method exit.

The true initial choice of limit was 8, to reflect the tacit assumption throughout the existing group code that it was safe to immediately (i.e. without normalization) mul/sqr any coordinate(s). Abstractly the intention was then to conservatively tighten the limits as necessary to enable specific method-entry optimizations without adding method-exit costs.

The only actual optimization applied was the one in the original PR removing the need to _normalize_weak as part of u1, u2 calculations in group addition. This leads to the choice of 6 for x, y.

My memory fails me as to the choice of 4 for z, but it might make sense that I was contemplating a possible implementation that calculates the output z using the form (a+b)^2 - a^2 - b^2 (instead of 2.a.b).

src/group_impl.h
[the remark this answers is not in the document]
real-or-random: I think the version with constants is slightly better.

src/group.h
peterdettman: Oops, not sure why I thought the original PR had z magnitude limit of 4. Anyway the choice was just based on the various extant group formulae at the time.

jonasnick: Thanks @peterdettman. The argument that it's a trade-off makes sense. The tighter limits don't add any method-exit costs though. Did you have any specific change in mind that would introduce method-exit costs when using the tighter limits?

jonasnick:
@theStack I slightly prefer this variant and I don't think the resulting expressions are particularly confusing. For the code it makes more sense to use the `SECP256K1_GE{,J}_{X,Y,Z}_MAGNITUDE_MAX` constants instead of magic numbers. And then it's easier to follow the comments if they use the same constants.

theStack:
Force-pushed the prepared variant which uses the magnitude limit constants both in the code and in the comments. Right now the tightest limits that still pass the tests are set ((4,3) for ge, (4,4,1) for gej), but this can be changed easily if there are concerns raised and more relaxed limits would be preferred (following the open discussion at https://github.com/bitcoin-core/secp256k1/pull/1348#discussion_r1279232666).

src/group.h
peterdettman: I think the introduction of _fe_half is what allowed the z max magnitude to be lowered to 1. We have flirted in the past with formulae that utilize an output z calculation of the form (a+b)^2 - a^2 - b^2 (where a^2 and b^2 are already used in the formula, so that overall 1M is saved at 1S cost - plus linear ops), which suggests it might like z max mag >= 4, but there's no strong reason to expect we will move to such formulae.

My main focus was really just getting the limits 8,8(,8) in place and verified. I don't see any harm in setting the tightest limits that work; future changes contemplating changes to those limits can then be argued case-by-case.

real-or-random: > I don't see any harm in setting the tightest limits that work; future changes contemplating changes to those limits can then be argued case-by-case.

Agreed. The important thing is that valid limits *exist*. Which among all valid limits we verify doesn't matter for correctness, so why not verify the tightest?

real-or-random (approved):
ACK b7c685e74adbd83937990e90f076600fabf8ccf0

jonasnick (commented):
ACK b7c685e74adbd83937990e90f076600fabf8ccf0

real-or-random:
@sipa Want to re-review this? If no, I think it's also ready for merge without another ACK.

PR #1350, opened 2023-06-17
scalar: introduce and use `secp256k1_{read,write}_be64` helpers

theStack:
This is a simple follow-up to #1339, as suggested in comment https://github.com/bitcoin-core/secp256k1/pull/1339#issuecomment-1587508040.


real-or-random (approved):
utACK 7067ee54b4206c26b382980f3c20b5fa0262a23a

PR #1355, opened 2023-06-24
Fix a typo in the error message

hebasto:
The code has been copy-pasted from the `precompute_ecmult_gen.c` source file.


real-or-random (approved):
ACK 67887ae65cf11d02c7055709082acd0e5d86db9b 

It took me a while to spot that the typo is `.c` vs `.h`.

PR #1356, opened 2023-06-25
ci: Adjust Docker image to Debian 12 "bookworm"

hebasto:
Since the [release](https://www.debian.org/News/2023/20230610.html) of Debian 12 "bookworm", it has become the "stable" one that our `ci/linux-debian.Dockerfile` relies on.

Last time the Docker image was built basing on Debian Bullseye.

Changes in packages are significant, for instance:
- `gcc` 10.2. --> 12.2
- `clang` 11.0 --> 14.0
- `wine` 5.0 --> 8.0

which requires certain adjustments provided in this PR.

The first commit has been cherry-picked from https://github.com/bitcoin-core/secp256k1/pull/1313.

### DIFF
--- a/ci/cirrus.sh
+++ b/ci/cirrus.sh
@@ -53,6 +53,22 @@ if [ -n "$WRAPPER_CMD" ]; then
     $WRAPPER_CMD --version
 fi
 
+# Workaround for https://bugs.kde.org/show_bug.cgi?id=452758 (fixed in valgrind 3.20.0).
+case "${CC:-undefined}" in
+    clang*)
+        if [ "$CTIMETESTS" = "yes" ] && [ "$WITH_VALGRIND" = "yes" ]
+        then
+            export CFLAGS="${CFLAGS:+$CFLAGS }-gdwarf-4"
+        else
+            case "$WRAPPER_CMD" in
+                valgrind*)
+                    export CFLAGS="${CFLAGS:+$CFLAGS }-gdwarf-4"
+                    ;;
+            esac
+        fi
+        ;;
+esac
+
 ./autogen.sh
 
 ./configure \
--- a/ci/linux-debian.Dockerfile
+++ b/ci/linux-debian.Dockerfile
@@ -11,9 +11,9 @@ RUN dpkg --add-architecture i386 && \
 RUN apt-get update && apt-get install --no-install-recommends -y \
         git ca-certificates \
         make automake libtool pkg-config dpkg-dev valgrind qemu-user \
-        gcc clang llvm libc6-dbg \
+        gcc clang llvm libclang-rt-dev libc6-dbg \
         g++ \
-        gcc-i686-linux-gnu libc6-dev-i386-cross libc6-dbg:i386 libubsan1:i386 libasan6:i386 \
+        gcc-i686-linux-gnu libc6-dev-i386-cross libc6-dbg:i386 libubsan1:i386 libasan8:i386 \
         gcc-s390x-linux-gnu libc6-dev-s390x-cross libc6-dbg:s390x \
         gcc-arm-linux-gnueabihf libc6-dev-armhf-cross libc6-dbg:armhf \
         gcc-aarch64-linux-gnu libc6-dev-arm64-cross libc6-dbg:arm64 \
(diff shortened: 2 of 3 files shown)


ci/cirrus.sh
```diff
@@ -53,6 +53,21 @@ if [ -n "$WRAPPER_CMD" ]; then
     $WRAPPER_CMD --version
 fi
 
+case "${CC:-undefined}" in
```
real-or-random, line 56:
```suggestion
# Workaround for https://bugs.kde.org/show_bug.cgi?id=452758 (fixed in valgrind 3.20.0)
case "${CC:-undefined}" in
```

ci/linux-debian.Dockerfile
```diff
@@ -11,21 +11,25 @@ RUN dpkg --add-architecture i386 && \
...
         gcc-powerpc64le-linux-gnu libc6-dev-ppc64el-cross libc6-dbg:ppc64el \
-        gcc-mingw-w64-x86-64-win32 wine64 wine \
+        gcc-mingw-w64-x86-64-win32 wine64 \
         gcc-mingw-w64-i686-win32 wine32 \
         sagemath
 
 WORKDIR /root
+
+# A workaround for `wine` package failure to employ the Debian alternatives system properly.
+RUN ln -s /usr/lib/wine/wine64 /usr/bin/wine64
```
real-or-random, line 28: What happens if we don't include this?

ci/linux-debian.Dockerfile
```diff
@@ -11,21 +11,25 @@ RUN dpkg --add-architecture i386 && \
...
+        gcc clang llvm libclang-rt-dev libc6-dbg \
         g++ \
-        gcc-i686-linux-gnu libc6-dev-i386-cross libc6-dbg:i386 libubsan1:i386 libasan6:i386 \
+        gcc-i686-linux-gnu libc6-dev-i386-cross libc6-dbg:i386 libubsan1:i386 libasan8:i386 \
         gcc-s390x-linux-gnu libc6-dev-s390x-cross libc6-dbg:s390x \
         gcc-arm-linux-gnueabihf libc6-dev-armhf-cross libc6-dbg:armhf \
         gcc-aarch64-linux-gnu libc6-dev-arm64-cross libc6-dbg:arm64 \
         gcc-powerpc64le-linux-gnu libc6-dev-ppc64el-cross libc6-dbg:ppc64el \
-        gcc-mingw-w64-x86-64-win32 wine64 wine \
+        gcc-mingw-w64-x86-64-win32 wine64 \
```
real-or-random, line 21: When I added this, the duplication was intentional. If every step in the dockerfile lists its dependencies, it's easier to remove steps. For example, if we ever want to get rid of `msvc-wine`, we can just drop that `RUN` step and the rest of the image is unaffected.

real-or-random (commented):
Concept ACK, thanks for taking care of this! :)

ci/linux-debian.Dockerfile
hebasto: The `wine64` won't be available in `PATH`, and the [`wine-msvc.sh`](https://github.com/mstorsjo/msvc-wine/blob/master/wrappers/wine-msvc.sh) script will fail.

ci/linux-debian.Dockerfile
hebasto: I see. It seems the `wine` package is not necessary for cross-compiling tasks anyway. But I need to test this assumption once more.

real-or-random: It installs the `wine` symlink. I think that's convenient.

ci/linux-debian.Dockerfile
real-or-random: Ok, I see, wine-msvc.sh has `wine64` hardcoded. Then I suggest including this `ln` in the `RUN` command of wine-msvc below. (Unless we really want to get rid of the `wine` convenience package, then we'll always need some manual symlink.)

hebasto:
Addressed @real-or-random's comments.

ci/linux-debian.Dockerfile
hebasto: True. Reverted back.

ci/linux-debian.Dockerfile
hebasto: > Then I suggest including this `ln` in the `RUN` command of wine-msvc below.

Done.

ci/cirrus.sh
hebasto: Updated.

real-or-random (approved):
ACK c862a9fb49e885dcafb42d4e21e05a244248aab0

PR #1357, opened 2023-06-25
tests: refactor: take use of `secp256k1_ge_x_on_curve_var`

theStack:
The recently merged ellswift PR (#1129) introduced a helper `secp256k1_ge_x_on_curve_var` to check if a given X coordinate is on the curve (i.e. the expression x^3 + 7 is square, see commit 79e5b2a8b80f507e2c9936ff1c4e2fb39bc66a4e). This can be used for code deduplication in the `ecmult_const_mult_xonly` test.

(Found this instance via `$ git grep add_int.*SECP256K1_B`, I think it's the only one where the helper can be used.)


real-or-random (approved):
utACK 7d8d5c86df8b27b45e80ed50341dd0ce64546c0f

PR #1358, opened 2023-06-26
tests: introduce helper for non-zero `random_field_element_test()` results 

theStack:
There are several instances in the tests where random non-zero field elements are generated by calling `random_field_element_test` in a do/while-loop with is-zero condition. This PR deduplicates all these by introducing a `random_field_element_non_zero_test` helper. Note that some instances checked the is-zero condition via `secp256k1_fe_normalizes_to_zero_var`, which is unnecessary, as the result of `random_field_element_test` is already normalized (so strictly speaking, this is not a pure refactor, and there could be tiny run-time improvements, though I doubt that's measurable).

Additionally, the first commit removes the function `random_fe_test` as it is logically a duplicate of `random_field_element_test`.


theStack:
I guess an alternative would be to not care about the zero-case at all (in the sense of not looping, but only asserting/verify-checking) due to the incredibly low probability of that ever happening? (I just remembered the discussion at https://github.com/bitcoin/bitcoin/pull/27479#discussion_r1171972333, though that was about invalid private keys; the probability to create a zero field-element seems to be even many orders of magnitudes lower with 1/p)

real-or-random (commented):
Concept ACK

> Additionally, the first commit removes the function `random_fe_test` as it is logically a duplicate of `random_field_element_test`.

I think it's better to use the name `random_fe_test` because we have other `random_fe_*` functions in the tests.

real-or-random:
needs rebase :/

theStack:
> needs rebase :/

Done.

real-or-random (approved):
ACK 5a95a268b944ffe64b7857e58f5b3b44aba514da

PR #1359, opened 2023-06-26
Fix symbol visibility issues, add test for it

hebasto:
Closes https://github.com/bitcoin-core/secp256k1/issues/1181.

[Catches](https://github.com/bitcoin-core/secp256k1/pull/1359#issuecomment-2417892235) the actual symbol visibility issue.

Replaces https://github.com/bitcoin-core/secp256k1/pull/1135.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -40,6 +40,7 @@ env:
   BENCH: 'yes'
   SECP256K1_BENCH_ITERS: 2
   CTIMETESTS: 'yes'
+  SYMBOL_CHECK: 'yes'
   # Compile and run the examples.
   EXAMPLES: 'yes'
 
@@ -365,6 +366,7 @@ jobs:
       ASAN_OPTIONS: 'strict_string_checks=1:detect_stack_use_after_return=1:detect_leaks=1'
       LSAN_OPTIONS: 'use_unaligned=1'
       SECP256K1_TEST_ITERS: 32
+      SYMBOL_CHECK: 'no'
 
     steps:
       - name: Checkout
@@ -415,6 +417,7 @@ jobs:
       SECP256K1_TEST_ITERS: 32
       ASM: 'no'
       WITH_VALGRIND: 'no'
+      SYMBOL_CHECK: 'no'
 
     steps:
       - name: Checkout
@@ -483,6 +486,7 @@ jobs:
       CC: 'clang'
       HOMEBREW_NO_AUTO_UPDATE: 1
       HOMEBREW_NO_INSTALL_CLEANUP: 1
+      SYMBOL_CHECK: 'no'
 
     strategy:
       fail-fast: false
@@ -515,6 +519,12 @@ jobs:
         env: ${{ matrix.env_vars }}
         run: ./ci/ci.sh
 
+      - name: Symbol check
+        run: |
+          python3 --version
+          python3 -m pip install lief
+          python3 ./tools/symbol-check.py .libs/libsecp256k1.dylib
+
       - name: Print logs
         uses: ./.github/actions/print-logs
         if: ${{ !cancelled() }}
@@ -530,6 +540,7 @@ jobs:
       HOMEBREW_NO_INSTALL_CLEANUP: 1
       WITH_VALGRIND: 'no'
       CTIMETESTS: 'no'
+      SYMBOL_CHECK: 'no'
 
     strategy:
       fail-fast: false
@@ -557,6 +568,16 @@ jobs:
         env: ${{ matrix.env_vars }}
         run: ./ci/ci.sh
 
+      - name: Symbol check
+        env:
+          VIRTUAL_ENV: '${{ github.workspace }}/venv'
+        run: |
+          python3 --version
+          python3 -m venv $VIRTUAL_ENV
+          export PATH="$VIRTUAL_ENV/bin:$PATH"
+          python3 -m pip install lief
+          python3 ./tools/symbol-check.py .libs/libsecp256k1.dylib
+
       - name: Print logs
         uses: ./.github/actions/print-logs
         if: ${{ !cancelled() }}
@@ -573,6 +594,7 @@ jobs:
         configuration:
           - job_name: 'x64 (MSVC): Windows (VS 2022, shared)'
             cmake_options: '-A x64 -DBUILD_SHARED_LIBS=ON'
+            symbol_check: 'true'
           - job_name: 'x64 (MSVC): Windows (VS 2022, static)'
             cmake_options: '-A x64 -DBUILD_SHARED_LIBS=OFF'
           - job_name: 'x64 (MSVC): Windows (VS 2022, int128_struct)'
@@ -600,6 +622,13 @@ jobs:
         shell: bash
         run: |
           cd build/bin/RelWithDebInfo && file *tests.exe bench*.exe libsecp256k1-*.dll || true
+
+      - name: Symbol check
+        if: ${{ matrix.configuration.symbol_check }}
+        run: |
+          py -3 --version
+          py -3 -m pip install lief
+          py -3 .\tools\symbol-check.py build\bin\RelWithDebInfo\libsecp256k1-5.dll
 
       - name: Check
         run: |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -271,8 +271,6 @@ else()
   try_append_c_flags(-Wundef)
 endif()
 
-set(CMAKE_C_VISIBILITY_PRESET hidden)
-
 set(print_msan_notice)
 if(SECP256K1_BUILD_CTIME_TESTS)
   include(CheckMemorySanitizer)
--- a/src/precomputed_ecmult.h
+++ b/src/precomputed_ecmult.h
@@ -13,6 +13,8 @@ extern "C" {
 
 #include "ecmult.h"
 #include "group.h"
+#include "util_local_visibility.h"
+
 #if defined(EXHAUSTIVE_TEST_ORDER)
 #    if EXHAUSTIVE_TEST_ORDER == 7
 #        define WINDOW_G 3
@@ -27,8 +29,8 @@ static secp256k1_ge_storage secp256k1_pre_g[ECMULT_TABLE_SIZE(WINDOW_G)];
 static secp256k1_ge_storage secp256k1_pre_g_128[ECMULT_TABLE_SIZE(WINDOW_G)];
 #else /* !defined(EXHAUSTIVE_TEST_ORDER) */
 #    define WINDOW_G ECMULT_WINDOW_SIZE
-extern const secp256k1_ge_storage secp256k1_pre_g[ECMULT_TABLE_SIZE(WINDOW_G)];
-extern const secp256k1_ge_storage secp256k1_pre_g_128[ECMULT_TABLE_SIZE(WINDOW_G)];
+SECP256K1_LOCAL_VAR const secp256k1_ge_storage secp256k1_pre_g[ECMULT_TABLE_SIZE(WINDOW_G)];
+SECP256K1_LOCAL_VAR const secp256k1_ge_storage secp256k1_pre_g_128[ECMULT_TABLE_SIZE(WINDOW_G)];
 #endif /* defined(EXHAUSTIVE_TEST_ORDER) */
 
 #ifdef __cplusplus
--- a/src/util_local_visibility.h
+++ b/src/util_local_visibility.h
@@ -1,0 +1,12 @@
+#ifndef SECP256K1_LOCAL_VISIBILITY_H
+#define SECP256K1_LOCAL_VISIBILITY_H
+
+/* Global variable visibility */
+/* See: https://github.com/bitcoin-core/secp256k1/issues/1181 */
+#if !defined(_WIN32) && defined(__GNUC__) && (__GNUC__ >= 4)
+# define SECP256K1_LOCAL_VAR extern __attribute__ ((visibility ("hidden")))
+#else
+# define SECP256K1_LOCAL_VAR extern
+#endif
+
+#endif /* SECP256K1_LOCAL_VISIBILITY_H */
--- a/tools/symbol-check.py
+++ b/tools/symbol-check.py
@@ -1,0 +1,72 @@
+#!/usr/bin/env python3
+"""Check that a libsecp256k1 shared library exports only expected symbols.
+
+Usage examples:
+  - When building with Autotools:
+    ./tools/symbol-check.py .libs/libsecp256k1.so
+    ./tools/symbol-check.py .libs/libsecp256k1-<V>.dll
+    ./tools/symbol-check.py .libs/libsecp256k1.dylib
+
+  - When building with CMake:
+    ./tools/symbol-check.py build/lib/libsecp256k1.so
+    ./tools/symbol-check.py build/bin/libsecp256k1-<V>.dll
+    ./tools/symbol-check.py build/lib/libsecp256k1.dylib"""
+
+import re
+import sys
+import subprocess
+
+import lief
+
+
+class UnexpectedExport(RuntimeError):
+    pass
+
+
+def get_exported_exports(library) -> list[str]:
+    """Adapter function to get exported symbols based on the library format."""
+    if library.format == lief.Binary.FORMATS.ELF:
+        return [symbol.name for symbol in library.exported_symbols]
+    elif library.format == lief.Binary.FORMATS.PE:
+        return [entry.name for entry in library.get_export().entries]
+    elif library.format == lief.Binary.FORMATS.MACHO:
+        return [symbol.name[1:] for symbol in library.exported_symbols]
+    raise NotImplementedError(f"Unsupported format: {library.format}")
+
+
+def grep_expected_symbols() -> list[str]:
+    """Guess the list of expected exported symbols from the source code."""
+    grep_output = subprocess.check_output(
+        ["git", "grep", r"^\s*SECP256K1_API", "--", "include"],
+        universal_newlines=True,
+        encoding="utf-8"
+    )
+    lines = grep_output.split("\n")
+    pattern = re.compile(r'\bsecp256k1_\w+')
+    exported: list[str] = [pattern.findall(line)[-1] for line in lines if line.strip()]
+    return exported
+
+
+def check_symbols(library, expected_exports) -> None:
+    """Check that the library exports only the expected symbols."""
+    actual_exports = get_exported_exports(library)
+    unexpected_exports = set(actual_exports) - set(expected_exports)
+    if unexpected_exports != set():
+        raise UnexpectedExport(f"Unexpected exported symbols: {unexpected_exports}")
+
+def main():
+    if len(sys.argv) != 2:
+        print(__doc__)
+        return 1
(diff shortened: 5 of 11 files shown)


hebasto:
Rebased on top of the merged #1356.

hebasto:
Rebased f8167296935958e1c7a65654de8c6606959e9f3f -> d7b1ceaa834d00709dc1206be6dbd65428513ef2 ([pr1359.02](https://github.com/hebasto/secp256k1/commits/pr1359.02) -> [pr1359.03](https://github.com/hebasto/secp256k1/commits/pr1359.03)) due to the conflict with #1313.

hebasto:
@maflcko 

Is it possible for a Cirrus persistent worker to do not override environment variables set with the `ENV` command in the Dockerfile?

hebasto:
Rebased to resolve conflicts.

src/util_defs.h
```diff
@@ -0,0 +1,12 @@
+#ifndef SECP256K1_UTIL_DEFS_H
+#define SECP256K1_UTIL_DEFS_H
```
theuni, line 2: Nit: `util_defs.h` is pretty ambiguous. How about `util_local_visibility.h` or so?

CMakeLists.txt
```diff
@@ -267,8 +267,6 @@ else()
   try_append_c_flags(-Wundef)
 endif()
 
-set(CMAKE_C_VISIBILITY_PRESET hidden)
```
theuni, line 270: In case it's not obvious to other reviewers, this is no longer needed because every non-static function should be externally visible.

If we ever need to share private symbols between compilation units this won't work. But I agree that it makes sense to take advantage of that if we can.

theuni (approved):
Code review and approach ACK.

Cool that this caught a real (new) symbol leak.

@fanquake suggested that the test could potentially be simplified for c-i.

src/util_defs.h
hebasto: Thanks! Reworked.

hebasto:
> Cool that this caught a real (new) symbol leak.

To clarify, the fix is in the first commit. The excerpt from the [CI log](https://github.com/bitcoin-core/secp256k1/actions/runs/11364808398/job/31612568439) with an error message from the previous iteration of this branch looks as follows:
```
+ python3 ./tools/symbol-check.py .libs/libsecp256k1.so
.libs/libsecp256k1.so: export of symbol secp256k1_musig_nonce_gen_internal not expected
.libs/libsecp256k1.so: export of symbol secp256k1_musig_nonce_gen_internal not expected
.libs/libsecp256k1.so: failed EXPORTED_SYMBOLS
Error: Process completed with exit code 1.
```

> @fanquake suggested that the test could potentially be simplified for c-i.

Thanks! The `tools/symbol-check.py` script has been simplified.

theuni (approved):
Code review ACK 44a9392f8b124aab44bd38c7aab6fbf8c4ac3c11. No opinion on the tests/ci.

src/precomputed_ecmult.h
```diff
@@ -26,9 +26,10 @@ extern "C" {
 static secp256k1_ge_storage secp256k1_pre_g[ECMULT_TABLE_SIZE(WINDOW_G)];
 static secp256k1_ge_storage secp256k1_pre_g_128[ECMULT_TABLE_SIZE(WINDOW_G)];
 #else /* !defined(EXHAUSTIVE_TEST_ORDER) */
+#include "util_local_visibility.h"
 #    define WINDOW_G ECMULT_WINDOW_SIZE
-extern const secp256k1_ge_storage secp256k1_pre_g[ECMULT_TABLE_SIZE(WINDOW_G)];
-extern const secp256k1_ge_storage secp256k1_pre_g_128[ECMULT_TABLE_SIZE(WINDOW_G)];
+SECP256K1_LOCAL_VAR const secp256k1_ge_storage secp256k1_pre_g[ECMULT_TABLE_SIZE(WINDOW_G)];
+SECP256K1_LOCAL_VAR const secp256k1_ge_storage secp256k1_pre_g_128[ECMULT_TABLE_SIZE(WINDOW_G)];
 #endif /* defined(EXHAUSTIVE_TEST_ORDER) */
```
real-or-random, line 33: Is there a specific reason to keep the `#include` in a specific branch (instead of putting it at the top of file)?

tools/symbol-check.py
```diff
@@ -0,0 +1,98 @@
...
+import re
+import sys
+import subprocess
+from typing import List
+
+import lief
+
+
+def grep_exported_symbols() -> List[str]:
+    grep_output = subprocess.check_output(["git", "grep", "^SECP256K1_API", "--", "include"], universal_newlines=True, encoding="utf8")
```
real-or-random, line 33:
```suggestion
    grep_output = subprocess.check_output(["git", "grep", "^[\s*]SECP256K1_API", "--", "include"], universal_newlines=True, encoding="utf8")
```
That's perhaps a bit more robust

src/util_local_visibility.h
real-or-random, line 1: I think this is small enough to keep it in `util.h`. Or is there a reason not to keep it here?

hebasto: The reason is to facilitate the further attempts to make all headers [self-contained](https://github.com/bitcoin-core/secp256k1/issues/1039).

hebasto:
@real-or-random's comments have been addressed.

.github/workflows/ci.yml
```diff
@@ -763,6 +782,12 @@ jobs:
         run: |
           cd build/bin/RelWithDebInfo && file *tests.exe bench*.exe libsecp256k1-*.dll || true
 
+      - name: Symbol check
+        if: ${{ matrix.configuration.cmake_options != '-A x64 -DBUILD_SHARED_LIBS=OFF' }}
```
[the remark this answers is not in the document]
hebasto: Thanks! [Reworked](https://github.com/bitcoin-core/secp256k1/pull/1359#issuecomment-2434852632).

tools/symbol-check.py
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python3
+'''
+A script to check that a secp256k1 shared library
```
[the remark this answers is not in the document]
hebasto: Thanks! [Amended](https://github.com/bitcoin-core/secp256k1/pull/1359#issuecomment-2434852632).

tools/symbol-check.py
```diff
@@ -0,0 +1,98 @@
...
+    ./tools/symbol-check.py build/lib/libsecp256k1.so
+or
+    ./tools/symbol-check.py build/bin/libsecp256k1-<V>.dll
+or
+    ./tools/symbol-check.py build/lib/libsecp256k1.dylib
+'''
+import re
+import sys
+import subprocess
+from typing import List
```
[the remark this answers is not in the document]
hebasto: Thanks! [Done](https://github.com/bitcoin-core/secp256k1/pull/1359#issuecomment-2434852632).

src/precomputed_ecmult.h
hebasto: Nope. [Reworked](https://github.com/bitcoin-core/secp256k1/pull/1359#issuecomment-2434852632).

tools/symbol-check.py
```diff
@@ -0,0 +1,97 @@
...
+        if name in expected_exports:
+            continue
+        print(f'{filename}: export of function {name} not expected')
+        ok = False
+    return ok
+
+
+def check_MACHO_exported_functions(library, expected_exports) -> bool:
+    ok: bool = True
+    macho_lib: lief.MACHO.Binary = library.concrete
```
fanquake, line 71: What are the calls to `.concrete` for?

tools/symbol-check.py
```diff
@@ -0,0 +1,97 @@
...
+    library: lief.Binary = lief.parse(filename)
+    exe_format: lief.Binary.FORMATS = library.format
+    if exe_format == lief.Binary.FORMATS.ELF:
+        success = check_ELF_exported_symbols(library, grep_exported_symbols())
+    elif exe_format == lief.Binary.FORMATS.PE:
+        success = check_PE_exported_functions(library, grep_exported_symbols())
+    elif exe_format == lief.Binary.FORMATS.MACHO:
+        success = check_MACHO_exported_functions(library, grep_exported_symbols())
+    else:
+        print(f'{filename}: unknown executable format')
```
fanquake, line 93: I don't think this code can be reached, because if it is an unknown format, `.parse()` will have errored (printing `Unknown format`), and then `lief.Binary.FORMATS = library.format` will fail:
```bash
./tools/symbol-check.py not_a_file
Unknown format
Traceback (most recent call last):
  File "/Users/michael/secp256k1/./tools/symbol-check.py", line 81, in <module>
    exe_format: lief.Binary.FORMATS = library.format
                                      ^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'format'
```

hebasto:
@fanquake's feedback has been addressed.

tools/symbol-check.py
hebasto: Thanks! [Reworked](https://github.com/bitcoin-core/secp256k1/pull/1359#issuecomment-2435037570).

tools/symbol-check.py
hebasto: [Dropped](https://github.com/bitcoin-core/secp256k1/pull/1359#issuecomment-2435037570).

real-or-random (commented):
This script can still be simplified. Let me suggest a refactored version later.

tools/symbol-check.py
real-or-random: > I don't think this code can be reached,

This will be reachable once LIEF adds support for a new binary format, so I think it was useful to have it.

real-or-random:
Here's a refactored version, which is more readable IMHO (haven't tested on Mac or Windows):

```python
#!/usr/bin/env python3
"""Check that a libsecp256k1 shared library exports only expected symbols.

Usage examples:
  - When building with Autotools:
    ./tools/symbol-check.py .libs/libsecp256k1.so
    ./tools/symbol-check.py .libs/libsecp256k1-<V>.dll
    ./tools/symbol-check.py .libs/libsecp256k1.dylib

  - When building with CMake:
    ./tools/symbol-check.py build/lib/libsecp256k1.so
    ./tools/symbol-check.py build/bin/libsecp256k1-<V>.dll
    ./tools/symbol-check.py build/lib/libsecp256k1.dylib"""

import re
import sys
import subprocess

import lief


class UnexpectedExport(RuntimeError):
    pass


def get_exported_exports(library) -> list[str]:
    """Adapter function to get exported symbols based on the library format."""
    if library.format == lief.Binary.FORMATS.ELF:
        return [symbol.name for symbol in library.exported_symbols]
    elif library.format == lief.Binary.FORMATS.PE:
        return [function.name for function in library.exported_functions]
    elif library.format == lief.Binary.FORMATS.MACHO:
        return [function.name[1:] for function in library.exported_functions]
    raise NotImplementedError(f"Unsupported format: {library.format}")


def grep_expected_symbols() -> list[str]:
    """Guess the list of expected exported symbols from the source code."""
    grep_output = subprocess.check_output(
        ["git", "grep", "^SECP256K1_API", "--", "include"], # TODO WHITESPACE
        universal_newlines=True,
        encoding="utf-8"
    )
    lines = grep_output.split("\n")
    pattern = re.compile(r'\bsecp256k1_\w+')
    exported: list[str] = [pattern.findall(line)[-1] for line in lines if line.strip()]
    return exported


def check_symbols(library, expected_exports) -> None:
    """Check that the library exports only the expected symbols."""
    actual_exports = list(get_exported_exports(library))
    unexpected_exports = set(actual_exports) - set(expected_exports)
    if unexpected_exports != set():
        raise UnexpectedExport(f"Unexpected exported symbols: {unexpected_exports}")

def main():
    if len(sys.argv) != 2:
        print(__doc__)
        return 1
    library = lief.parse(sys.argv[1])
    expected_exports = grep_expected_symbols()
    try:
        check_symbols(library, expected_exports)
    except UnexpectedExport as e:
        print(f"ERROR: In {sys.argv[1]}: {e}")
        return 1
    return 0


if __name__ == "__main__":
    sys.exit(main())
```

tools/symbol-check.py
```diff
@@ -0,0 +1,88 @@
...
+        if name in expected_exports:
+            continue
+        print(f'{filename}: export of symbol {name} not expected')
+        ok = False
+    return ok
+
+
+def check_PE_exported_functions(library, expected_exports) -> bool:
+    ok: bool = True
+    for function in library.exported_functions:
```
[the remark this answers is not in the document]
hebasto: The suggested approach returns the same list of symbols on the master branch. Incorporated as a potentially superior one.

tools/symbol-check.py
```diff
@@ -0,0 +1,88 @@
...
+        if name in expected_exports:
+            continue
+        print(f'{filename}: export of function {name} not expected')
+        ok = False
+    return ok
+
+
+def check_MACHO_exported_functions(library, expected_exports) -> bool:
+    ok: bool = True
+    for function in library.exported_functions:
```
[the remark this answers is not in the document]
hebasto: The suggested approach returns the same list of symbols on the master branch. Incorporated as a potentially superior one.

hebasto:
> Here's a refactored version, which is more readable IMHO (haven't tested on Mac or Windows):

Tested on macOS and on Windows. Incorporated. Thank you!

tools/symbol-check.py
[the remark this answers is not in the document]
real-or-random: Hm, just checking: You're saying that this change didn't make a difference, right? So it still finds only exported functions and not exported variables?

(And I have the same question for PE.)

tools/symbol-check.py
```diff
@@ -0,0 +1,72 @@
...
+        return [entry.name for entry in library.get_export().entries]
+    elif library.format == lief.Binary.FORMATS.MACHO:
+        return [symbol.name[1:] for symbol in library.exported_symbols]
+    raise NotImplementedError(f"Unsupported format: {library.format}")
+
+
+def grep_expected_symbols() -> list[str]:
+    """Guess the list of expected exported symbols from the source code."""
+    grep_output = subprocess.check_output(
+        ["git", "grep", "^SECP256K1_API", "--", "include"], # TODO WHITESPACE
```
real-or-random, line 40:
```suggestion
        ["git", "grep", "^\s*SECP256K1_API", "--", "include"],
```

Oops, I think I had this in mind here, to make the matching a bit more flexible.

src/util_local_visibility.h
real-or-random: I guess I'm just missing it. Can you explain how this will help with #1039?

tools/symbol-check.py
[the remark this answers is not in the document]
hebasto: > Hm, just checking: You're saying that this change didn't make a difference, right?

Correct.

> So it still finds only exported functions and not exported variables?

With the reverted second commit, both variants catch the same local variables:
```
./tools/symbol-check.py: In .libs/libsecp256k1.dylib: Unexpected exported symbols: {'secp256k1_pre_g', 'secp256k1_pre_g_128', 'secp256k1_ecmult_gen_prec_table'}
```

> (And I have the same question for PE.)

With the reverted second commit, both variants do _not_ catch local variables.

For more details, please refer to:
- https://github.com/hebasto/secp256k1/actions/runs/11640943476
- https://github.com/hebasto/secp256k1/actions/runs/11640955643

src/util_local_visibility.h
hebasto: My intention was to avoid bringing the high-level `../include/secp256k1.h` header into lower-level code via `util.h`. However, at this point, `util.h` is already being included indirectly through other headers.

tools/symbol-check.py
hebasto: Thanks! Done.

tools/symbol-check.py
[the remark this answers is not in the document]
real-or-random: Nevermind, my analysis of variables is wrong. LIEF shows them. I'll revisit this later.

tools/symbol-check.py
```diff
@@ -0,0 +1,72 @@
...
+        return [entry.name for entry in library.get_export().entries]
+    elif library.format == lief.Binary.FORMATS.MACHO:
+        return [symbol.name[1:] for symbol in library.exported_symbols]
+    raise NotImplementedError(f"Unsupported format: {library.format}")
+
+
+def grep_expected_symbols() -> list[str]:
+    """Guess the list of expected exported symbols from the source code."""
+    grep_output = subprocess.check_output(
+        ["git", "grep", "^\s*SECP256K1_API", "--", "include"],
```
jonasnick, line 40: This gives me error `symbol-check.backup.py:40: SyntaxWarning: invalid escape sequence '\s'`.  I think it should be 
```
r"^\s*SECP256K1_API"
```
or 
```
"^\\s*SECP256K1_API"
```

hebasto:
Rebased and addressed @jonasnick's https://github.com/bitcoin-core/secp256k1/pull/1359#discussion_r1827744514.

theuni (approved):
Code-review reACK 7ac46cf93c322193e428effeac3f024f4ab6a05b.

Still no opinion on the .py/ci.

Being able to drop `-fvisibility=hidden` is a nice improvement (and also a testament to the code structure!)

tools/symbol-check.py
```diff
@@ -0,0 +1,72 @@
...
+    )
+    lines = grep_output.split("\n")
+    pattern = re.compile(r'\bsecp256k1_\w+')
+    exported: list[str] = [pattern.findall(line)[-1] for line in lines if line.strip()]
+    return exported
+
+
+def check_symbols(library, expected_exports) -> None:
+    """Check that the library exports only the expected symbols."""
+    actual_exports = list(get_exported_exports(library))
```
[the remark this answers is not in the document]
hebasto: Thanks! Fixed.

hebasto:
Addressed the feedback from @real-or-random's and rebased.

real-or-random:
needs rebase, sorry :/

hebasto:
> needs rebase, sorry :/

Rebased :)

real-or-random (approved):
reACK d1478763a5f400fafb42af5911db4a9460dd4a5d

tools/symbol-check.py
```diff
@@ -0,0 +1,72 @@
...
+def check_symbols(library, expected_exports) -> None:
+    """Check that the library exports only the expected symbols."""
+    actual_exports = get_exported_exports(library)
+    unexpected_exports = set(actual_exports) - set(expected_exports)
+    if unexpected_exports != set():
+        raise UnexpectedExport(f"Unexpected exported symbols: {unexpected_exports}")
+
+def main():
+    if len(sys.argv) != 2:
+        print(__doc__)
```
fanquake, line 59: > print(__doc__)

Don't think this does anything?

fanquake: macOS & Python 3.13.2

real-or-random: > ```shell
> ./tools/symbol-check.py                                
> Traceback (most recent call last):
>   File "/Users/michael/secp256k1/./tools/symbol-check.py", line 83, in <module>
>     filename: str = sys.argv[1]
>                     ~~~~~~~~^^^
> IndexError: list index out of range
> ```

Are you on the most recent revision of this PR? The current `./tools/symbol-check.py` has no line 83.

Issue #1361, opened 2023-06-27
`-Wmaybe-uninitialized` when compiling with `-O1`


hebasto:
```
$ ./autogen.sh && ./configure CFLAGS=-O1
$ make clean > /dev/null && make > /dev/null 
In file included from src/secp256k1.c:30:
In function 'secp256k1_ecmult_strauss_wnaf',
    inlined from 'secp256k1_ecmult' at src/ecmult_impl.h:353:5:
src/ecmult_impl.h:291:5: warning: 'aux' may be used uninitialized [-Wmaybe-uninitialized]
  291 |     secp256k1_ge_table_set_globalz(ECMULT_TABLE_SIZE(WINDOW_A) * no, state->pre_a, state->aux);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from src/secp256k1.c:29:
src/group_impl.h: In function 'secp256k1_ecmult':
src/group_impl.h:224:13: note: by argument 3 of type 'const secp256k1_fe *' to 'secp256k1_ge_table_set_globalz' declared here
  224 | static void secp256k1_ge_table_set_globalz(size_t len, secp256k1_ge *a, const secp256k1_fe *zr) {
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/ecmult_impl.h:345:18: note: 'aux' declared here
  345 |     secp256k1_fe aux[ECMULT_TABLE_SIZE(WINDOW_A)];
      |                  ^~~
In file included from src/secp256k1.c:30,
                 from src/bench_internal.c:8:
In function ‘secp256k1_ecmult_strauss_wnaf’,
    inlined from ‘secp256k1_ecmult’ at src/ecmult_impl.h:353:5:
src/ecmult_impl.h:291:5: warning: ‘aux’ may be used uninitialized [-Wmaybe-uninitialized]
  291 |     secp256k1_ge_table_set_globalz(ECMULT_TABLE_SIZE(WINDOW_A) * no, state->pre_a, state->aux);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from src/secp256k1.c:29:
[...]

PR #1364, opened 2023-06-28
Avoid `-Wmaybe-uninitialized` when compiling with `gcc -O1`

hebasto:
Fixes https://github.com/bitcoin-core/secp256k1/issues/1361.

CI tasks have been adjusted to catch similar issues in the future.


real-or-random (approved):
utACK 5b9f37f136620b9c61cd66439904b2db266fba70

jonasnick (commented):
tACK 5b9f37f136620b9c61cd66439904b2db266fba70

PR #1366, opened 2023-06-29
field: Use `restrict` consistently in fe_sqrt

real-or-random:
That is, use it also in the definition and not only the declaration.

I believe this was the intention of commit
https://github.com/bitcoin-core/secp256k1/commit/be82bd8e0347e090037ff1d30a22a9d614db8c9f, but it was omitted there.

edit: Changed the description. I'm not entirely sure but after looking at the standard, I tend to think this is more than a cosmetic change, and only this change actually makes the parameters `restrict`. Anyway, I believe making them `restrict` was simply forgotten in be82bd8e0347e090037ff1d30a22a9d614db8c9f.

### DIFF
--- a/src/field_impl.h
+++ b/src/field_impl.h
@@ -44,7 +44,7 @@ SECP256K1_INLINE static int secp256k1_fe_equal_var(const secp256k1_fe *a, const 
     return secp256k1_fe_normalizes_to_zero_var(&na);
 }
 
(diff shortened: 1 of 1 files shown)


real-or-random:
I double-checked all the call sites manually, and we check aliasing in VERIFY mode, so I think it's safe to merge this.

PR #1371, opened 2023-07-04
Add exhaustive tests for ellswift (with create+decode roundtrip)

theStack:
This PR adds the basic structure for ellswift exhaustive tests. Right now only a `secp256k1_ellswift_create` + `secp256k1_ellswift_decode` indirect roundtrip (exhaustive loop scalar -> ellswift pubkey -> decoded pubkey -> decoded group element, compared with exhaustive precomputed group element) is included.

The exhaustive tests passes locally with all currently supported orders (n=13 [default] and n=199). Note that for n=7, the test is skipped, as the used curve in this case is even-ordered and ellswift only supports odd-ordered curves.

### DIFF
--- a/src/tests_exhaustive.c
+++ b/src/tests_exhaustive.c
@@ -13,6 +13,9 @@
 #define EXHAUSTIVE_TEST_ORDER 13
 #endif
 
+/* These values of B are all values in [1, 8] that result in a curve with even order. */
+#define EXHAUSTIVE_TEST_CURVE_HAS_EVEN_ORDER (SECP256K1_B == 1 || SECP256K1_B == 6 || SECP256K1_B == 8)
+
 #ifdef USE_EXTERNAL_DEFAULT_CALLBACKS
     #pragma message("Ignoring USE_EXTERNAL_CALLBACKS in exhaustive_tests.")
     #undef USE_EXTERNAL_DEFAULT_CALLBACKS
@@ -395,6 +398,10 @@ static void test_exhaustive_sign(const secp256k1_context *ctx, const secp256k1_g
 #include "modules/schnorrsig/tests_exhaustive_impl.h"
 #endif
 
+#ifdef ENABLE_MODULE_ELLSWIFT
+#include "modules/ellswift/tests_exhaustive_impl.h"
+#endif
+
 int main(int argc, char** argv) {
     int i;
     secp256k1_gej groupj[EXHAUSTIVE_TEST_ORDER];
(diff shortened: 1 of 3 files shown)


alokeutpal (approved):

real-or-random:
I think it's worth including these restrictions as a comment (now that @sipa spelled them out explicitly).

Concept ACK.

As @sipa says, in terms of code complexity, etc. it probably won't hurt to add tests. Also, those here should be quick to run.

It's reasonable to ask if it's worth our time working on this, but now that the code is already there, why not? :) But not sure either if it makes sense to spend addtional time on `secp256k1_ellswift_xdh`.

theStack:
@sipa @real-or-random Thanks for the valuable feedback!

Added both (shortened) comments about the limitations. As for the second one, IIUC for the n=7 exhaustive group we are currently just lucky that this passes and it would be cleaner to not run the test for even-order curves at all, as they are not supported. I've included skipping to run the test in this case with an additional pre-processor definition. (This is obviously a bit ugly. I though rather than checking for a concrete value, it would be nice if `gen_exhaustive_groups.sage` could emit another #define that tells whether the a curve is even- or odd-ordered and we could then check on that instead. But that seems to be overkill?)

real-or-random (approved):
ACK f4527f443685730a919d9125c4590b21d4f400f0

src/tests_exhaustive.c
```diff
@@ -13,6 +13,8 @@
 #define EXHAUSTIVE_TEST_ORDER 13
 #endif
 
+#define EXHAUSTIVE_TEST_CURVE_IS_EVEN_ORDERED (SECP256K1_B == 1 || SECP256K1_B == 6 || SECP256K1_B == 8)
```
real-or-random, line 16:
```suggestion
/* These values of B are all values in [1, 8] that result in a curve with even order. */ 
#define EXHAUSTIVE_TEST_CURVE_HAS_EVEN_ORDER (SECP256K1_B == 1 || SECP256K1_B == 6 || SECP256K1_B == 8)
```

theStack: 👍 Done.

real-or-random (approved):
utACK 883d64a492d4aec7e2dd0987e0d8300495267e8a

src/tests_exhaustive.c
theStack: Oops, missed the second part of your recommendation. Agree that `_HAS_EVEN_ORDER` sounds better than `_IS_EVEN_ORDERED`. Force-pushed once again.

real-or-random (approved):
utACK 2792119278bcb2a0befce3fbc64c83578df54953

PR #1291, opened 2023-04-29
cmake: Use `SECP256K1_COVERAGE` option instead of `CMAKE_BUILD_TYPE=Coverage`

hebasto:
Providing `-DCMAKE_BUILD_TYPE=Coverage` does not work for multi-config generators.

For example:
```
$ cmake -S . -B ../build -G "Ninja Multi-Config"
$ cmake --build ../build --config Coverage
```
still builds the `tests` binary.

This PR:
- fixes this issue
- is based on the in-person discussion during the recent CoreDev
- is an alternative to #1251

The last commit addresses [that](https://github.com/bitcoin-core/secp256k1/pull/1113#discussion_r916977981) comment.

PR #1372, opened 2023-07-04
doc: Add unified docs for configuration autotools/CMake/manual builds

hebasto:
Closes the last item in the "Affecting both build systems" group in https://github.com/bitcoin-core/secp256k1/issues/1235.

Draft for now, as it is based on https://github.com/bitcoin-core/secp256k1/pull/1291.


real-or-random:
Hm, I tend to think that I like the current layout, with the top-level sections being the different options (and not the build systems).  It makes it easy to have larger description texts, e.g., the explanation of the window size. 


But yeah, it's not very elegant when the tables all have different column widths (if this is what you dislike @sipa). In principle, it seems possible to set the column widths:
```
<table>
<tr><td width="500">foo</td><td>bar</td>
<table>
```
But that's hard to read as plain text, and it's deprecated HTML (GitHub doesn't like `style="width:500px"`)... Not convinced.

fanquake:
> Draft for now, as it is based on https://github.com/bitcoin-core/secp256k1/pull/1291.

If this is just docs, does it need to be based on #1291 (which seems to have stalled)?

Issue #1360, opened 2023-06-27
scalar: Verify invariants on every entry


real-or-random:
We have `secp256k1_ge_verify`, `secp256k1_gej_verify`, and `secp256k1_fe_verify` functions to the invariants of the respective type. We call them on every entry/exit of a function that operates on a respective element.  

We should add a similar function for scalars. I think the only invariant is that scalars `a` are reduced mod the group order, i.e., `secp256k1_scalar_check_overflow(a) == 0`.

(see https://github.com/bitcoin-core/secp256k1/pull/1184#discussion_r1239566294)

@stratospher Are you interested in working on this?

PR #1373, opened 2023-07-06
Add invariant checking for scalars

stratospher:
From #1360. This PR:
1. adds `secp256k1_scalar_verify` to make sure scalars are reduced mod the group order in VERIFY mode
2. uses `secp256k1_scalar_verify` in all the scalar functions except `secp256k1_scalar_clear`, `secp256k1_scalar_reduce_512`, `secp256k1_scalar_mul_512` and `secp256k1_scalar_*_var` functions in `scalar_low_impl.h`


src/scalar.h
```diff
@@ -19,6 +19,34 @@
 #error "Please select wide multiplication implementation"
 #endif
 
+#ifndef VERIFY
+/* In non-VERIFY mode, we #define the scalar operations to be identical to their
+ * internal scalar implementation, to avoid the potential overhead of a
+ * function call (even though presumably inlinable). */
+#  define secp256k1_scalar_clear secp256k1_scalar_impl_clear
```
real-or-random, line 26: Hm, when I opened #1360 I didn't except that we want this indirection also here, but I think it makes sense.

On the one hand, the situation here is similar to the field code in the sense that we have multiple scalar implementations. On the other hand, the invariant here (no overflow) is inherently a property of the implementation, and does not leak into the scalar interface. So, you could argue that checking this is a concern of the implementation. But hm, what you do here is simply introduce a generic way for implementations to provide a verify function (which happens to be very similar here in all three implementations).  And this makes sense.

To be clear: What you propose here makes sense and solves the problem, I'm just wrapping my around whether it's the cleanest/most elegant solution. cc @sipa who worked on this in the field code.

real-or-random:
By the way, CI fails with:

```
==3276==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x55d56ff0370a in secp256k1_scalar_verify /tmp/cirrus-ci-build/src/scalar_impl.h:40:5
    #1 0x55d56ff0370a in secp256k1_scalar_clear /tmp/cirrus-ci-build/src/scalar_impl.h:44:5
    #2 0x55d56ff0370a in secp256k1_ecdsa_sign_inner /tmp/cirrus-ci-build/src/secp256k1.c:547:5
    #3 0x55d57005097f in secp256k1_ecdsa_sign /tmp/cirrus-ci-build/src/secp256k1.c:567:11
    #4 0x55d57005097f in test_ecdsa_edge_cases /tmp/cirrus-ci-build/src/tests.c:7378:9
    #5 0x55d56ff8cea2 in run_ecdsa_edge_cases /tmp/cirrus-ci-build/src/tests.c:7462:5
    #6 0x55d56ff8cea2 in main /tmp/cirrus-ci-build/src/tests.c:7854:5
    #7 0x7f34b1aaa189 in __libc_start_call_main csu/../sysdeps/nptl/libc_start_call_main.h:58:16
    #8 0x7f34b1aaa244 in __libc_start_main csu/../csu/libc-start.c:381:3
    #9 0x55d56fe732b0 in _start (/tmp/cirrus-ci-build/tests+0x212b0) (BuildId: 8839613b681f146479fe0450db93bdb91ebc1e2b)
SUMMARY: MemorySanitizer: use-of-uninitialized-value /tmp/cirrus-ci-build/src/scalar_impl.h:40:5 in secp256k1_scalar_verify
```
(https://cirrus-ci.com/task/4600089314852864?logs=cat_tests_log#L16)

real-or-random:
> should we just skip the check in `secp256k1_scalar_set_int`, `secp256k1_scalar_get_b32` instead in `scalar_low_impl.h`?

This means whether we check or not depends on the specifics of the tests, which sounds wrong. (Or in other words: In principle, the code should work for all valid tests, independently of whether we happen to have them in our test suite.)


And skipping the check it in some functions is somewhat inconsequential.  I think we should either decide that we want to enforce the invariant in `low` (and then do it everywhere because otherwise it's not an invariant) or we don't want to enforce it.  And I still think having the invariant is a good idea. 

Now that I think about this more, I tend to believe that the test is flawed. Or at least, I don't understand the test. Why would the test except two different representations of the *same* scalar yield different results? I think it should either not test `EXHAUSTIVE_TEST_ORDER` or if it does, it should expect that the result is the same as for `0` (but if we want to test the latter, then it should probably happen outside this loop, which is supposed to find exactly one valid sig).

src/modules/schnorrsig/tests_exhaustive_impl.h
```diff
@@ -116,7 +116,8 @@ static void test_exhaustive_schnorrsig_verify(const secp256k1_context *ctx, cons
                         if (s <= EXHAUSTIVE_TEST_ORDER) {
                             secp256k1_scalar s_s;
                             secp256k1_scalar_set_int(&s_s, s);
-                            secp256k1_scalar_get_b32(sig64 + 32, &s_s);
+                            memset(sig64 + 32, 0, 32);
+                            sig64[63] = s;
```
real-or-random, line 120: I think this is equivalent and shorter:

```suggestion
                            secp256k1_write_be32(sig64 + 32, s);
```

(also need to change `s_s` to `s` below)

real-or-random:
> i've updated the test so that `sig64` stores `EXHAUSTIVE_TEST_ORDER` without reducing it mod the group order. would that work?

Makes sense to me.

src/scalar_4x64_impl.h
real-or-random, line 1: I think it will be a good idea to add empty lines between the "verification" blocks (e.g., containing `secp256k1_scalar_verify` and other `VERIFY_CHECK`s) and the actual function body. I suggested the same in #1348, see https://github.com/bitcoin-core/secp256k1/pull/1348/commits/690b0fc05abd76cb7f6bd87e88bf7b8b0fd1ab70.

src/scalar_low_impl.h
```diff
@@ -60,79 +69,102 @@ static void secp256k1_scalar_set_b32(secp256k1_scalar *r, const unsigned char *b
...
     for (i = 0; i < EXHAUSTIVE_TEST_ORDER; i++)
         if ((i * *x) % EXHAUSTIVE_TEST_ORDER == 1)
             *r = i;
     /* If this VERIFY_CHECK triggers we were given a noninvertible scalar (and thus
      * have a composite group order; fix it in exhaustive_tests.c). */
     VERIFY_CHECK(*r != 0);
+    secp256k1_scalar_verify(r);
 }
 
 static void secp256k1_scalar_inverse_var(secp256k1_scalar *r, const secp256k1_scalar *x) {
```
real-or-random, line 170: Did you skip this intentionally because this function just delegates to `secp256k1_scalar_inverse`? I think it's preferable to have the checks in every function. This makes the code easier to change, in the sense that we don't need to bother with adding checks when change functions. (And even if we care about the runtime overhead of this in the tests, the compiler will most likely figure out that the duplicate call be dropped.)

real-or-random:
> i've updated the test so that `sig64` stores `EXHAUSTIVE_TEST_ORDER` without reducing it mod the group order.

One more nit: It's a good idea to reorder the commits such that this commit comes first. Then the tests will pass even on intermediate commits. (We have no strict rule on this, but it's a good idea. For example, it helps with `git bisect`.)

real-or-random (approved):
utACK d23da6d55714271c720fee58fbff5e5ef2fe193f

src/modules/schnorrsig/tests_exhaustive_impl.h
```diff
@@ -110,15 +110,15 @@ static void test_exhaustive_schnorrsig_verify(const secp256k1_context *ctx, cons
...
                     for (s = 0; s <= EXHAUSTIVE_TEST_ORDER + 1; ++s) {
                         int expect_valid, valid;
                         if (s <= EXHAUSTIVE_TEST_ORDER) {
-                            secp256k1_scalar s_s;
-                            secp256k1_scalar_set_int(&s_s, s);
-                            secp256k1_scalar_get_b32(sig64 + 32, &s_s);
+                            memset(sig64 + 32, 0, 32);
+                            secp256k1_write_be32(sig64 + 60, s);
                             expect_valid = actual_k != -1 && s != EXHAUSTIVE_TEST_ORDER &&
-                                           (s_s == (actual_k + actual_d * e) % EXHAUSTIVE_TEST_ORDER);
```
theStack, line 121: Unrelated note: looking at that replaced line, I was confused for a moment that comparing `secp256k1_scalar` directly with an integer is possible, until I realized that it's simply a uint32_t typedef in the scalar_low implementation (would have expected that it's still wrapped in a struct), and that's of course the only one used in exhaustive tests.

theStack (approved):
Code-review ACK d23da6d55714271c720fee58fbff5e5ef2fe193f

src/modules/schnorrsig/tests_exhaustive_impl.h
real-or-random: Yep, indeed. I guess that's okay for test code, introducing another function here will be overkill.

PR #1378, opened 2023-07-17
ellswift: fix probabilistic test failure when swapping sides

jonasnick:
Reported by jonatack in https://github.com/bitcoin/bitcoin/issues/28079.

When configured with `--disable-module-ecdh --enable-module-recovery`, then `./tests  64 81af32fd7ab8c9cbc2e62a689f642106` fails with
```
src/modules/ellswift/tests_impl.h:396: test condition failed: secp256k1_memcmp_var(share32_bad, share32a, 32) != 0
```

This tests verifies that changing the `party` bit of the `secp256k1_ellswift_xdh` function results in a different share. However, that's not the case when the secret keys of both parties are the same and this is actually what happens in the observed test failure. The keys can be equal in this test case because they are created by the `random_scalar_order_test` function whose output is not uniformly random (it's biased towards 0).

This commit restores the assumption that the secret keys differ.

### DIFF
--- a/src/modules/ellswift/tests_impl.h
+++ b/src/modules/ellswift/tests_impl.h
@@ -322,7 +322,9 @@ void run_ellswift_tests(void) {
         secp256k1_testrand256_test(auxrnd32a);
         secp256k1_testrand256_test(auxrnd32b);
         random_scalar_order_test(&seca);
-        random_scalar_order_test(&secb);
+        /* Draw secb uniformly at random to make sure that the secret keys
+         * differ */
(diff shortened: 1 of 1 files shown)


real-or-random (approved):
utACK c424e2fb43c8ed959b2af7b2216028ce2a023488

real-or-random:
> `random_scalar_order_test` function whose output is not uniformly random (it's biased towards 0).

Or more precisely, the individual bytes are biased towards 0x00 but `random_scalar_order_test` will ensure that the scalar is not zero.

Issue #1136, opened 2022-08-24
release: Checking binary compatibility


real-or-random:
When looking into binary compatibility (see #1055, #964), I found this nice website which describes best practices and open source tools for checking binary compatibility: https://abi-laboratory.pro This may be something to look into *after* the first release (but maybe before the second release).

See for example [the report for Botan](https://abi-laboratory.pro/index.php?view=timeline&l=botan), [which by the way doesn't provide binary compatibility across versions at all](https://botan.randombit.net/handbook/abi.html).

hebasto:
FWIW, here is what I've got for the recent releases:
- [API compatibility report for the libsecp256k1.so object between 0.2.0 and 0.3.0 versions on x86_64](https://htmlpreview.github.io/?https://github.com/hebasto/artefacts/blob/master/libsecp256k1%20ABI%20compatibility/compat_reports/libsecp256k1/0.2.0_to_0.3.0/compat_report.html)
- [API compatibility report for the libsecp256k1.so object between 0.3.0 and 0.3.1 versions on x86_64](https://htmlpreview.github.io/?https://github.com/hebasto/artefacts/blob/master/libsecp256k1%20ABI%20compatibility/compat_reports/libsecp256k1/0.3.0_to_0.3.1/compat_report.html)

hebasto:
> @hebasto I don't understand how that commit can have any effect on the public ABI.

Neither do I. That was reported by the tool. Maybe a false positive report?

hebasto:
No issues with gcc-10:
- [0.2.0 --> 0.3.0](https://htmlpreview.github.io/?https://github.com/hebasto/artefacts/blob/gcc-10/libsecp256k1%20ABI%20compatibility/compat_reports/libsecp256k1/0.2.0_to_0.3.0/compat_report.html)
- [0.3.0 --> 0.3.1](https://htmlpreview.github.io/?https://github.com/hebasto/artefacts/blob/gcc-10/libsecp256k1%20ABI%20compatibility/compat_reports/libsecp256k1/0.3.0_to_0.3.1/compat_report.html)

real-or-random:
Interestingly, the dumps generated with gcc 10 don't have any "Reg"(ister) entries in their json file, whereas the dumps generated with gcc 11 list registers for some arguments...  

Maybe open an issue with the tool?

PR #1380, opened 2023-07-18
Add ABI checking tool for release process

hebasto:
Closes https://github.com/bitcoin-core/secp256k1/issues/1136.

An example of using the `check-abi.sh` tool against changes from https://github.com/bitcoin-core/secp256k1/pull/1140:

image

### DIFF
--- a/doc/release-process.md
+++ b/doc/release-process.md
@@ -34,6 +34,11 @@ build=$(mktemp -d)
 cmake -B $build -DCMAKE_INSTALL_PREFIX=$dir && cmake --build $build --target install && ls -RlAh $dir
 gcc -o ecdsa examples/ecdsa.c -I $dir/include -L $dir/lib*/ -l secp256k1 -Wl,-rpath,"$dir/lib",-rpath,"$dir/lib64" && ./ecdsa
 ```
+4. Use the [`check-abi.sh`](/tools/check-abi.sh) tool to ensure there are no unexpected ABI incompatibilities and that the version number and release notes accurately reflect all potential ABI changes. To run this tool, the `abi-dumper` and `abi-compliance-checker` packages are required.
+
+```shell
+tools/check-abi.sh
+```
 
 ## Regular release
 
--- a/tools/check-abi.sh
+++ b/tools/check-abi.sh
@@ -1,0 +1,64 @@
+#!/bin/sh
+
+set -eu
+
+default_base_version="$(git describe --match "v*.*.*" --abbrev=0)"
+default_new_version="master"
+
+display_help_and_exit() {
+    echo "Usage: $0 <base_ver> <new_ver>"
+    echo ""
+    echo "Description: This script uses the ABI Compliance Checker tool to determine if the ABI"
+    echo "             of a new version of libsecp256k1 has changed in a backward-incompatible way."
+    echo ""
+    echo "Options:"
+    echo "  base_ver      Specify the base version (default: $default_base_version)"
+    echo "  new_ver       Specify the new version (default: $default_new_version)"
+    echo "  -h, --help    Display this help message"
+    exit 0
+}
+
+if [ "$#" -eq 0 ]; then
+    base_version="$default_base_version"
+    new_version="$default_new_version"
+elif [ "$#" -eq 1 ] && { [ "$1" = "-h" ] || [ "$1" = "--help" ]; }; then
+    display_help_and_exit
+elif [ "$#" -eq 2 ]; then
+    base_version="$1"
+    new_version="$2"
+else
+    echo "Invalid usage. See help:"
+    echo ""
+    display_help_and_exit
+fi
+
+checkout_and_build() {
+    git worktree add -d "$1" "$2"
+    cd "$1"
+    mkdir build && cd build
+    cmake -S .. --preset dev-mode \
+        -DCMAKE_C_COMPILER=gcc -DCMAKE_BUILD_TYPE=None -DCMAKE_C_FLAGS="-g -Og -gdwarf-4" \
+        -DSECP256K1_BUILD_BENCHMARK=OFF \
+        -DSECP256K1_BUILD_TESTS=OFF \
+        -DSECP256K1_BUILD_EXHAUSTIVE_TESTS=OFF \
+        -DSECP256K1_BUILD_CTIME_TESTS=OFF \
+        -DSECP256K1_BUILD_EXAMPLES=OFF
+    cmake --build . -j "$(nproc)"
+    abi-dumper src/libsecp256k1.so -o ABI.dump -lver "$2"
+}
+
+echo "Comparing $base_version (base version) to $new_version (new version)"
+echo
+
+original_dir="$(pwd)"
+
+base_source_dir=$(mktemp -d)
+checkout_and_build "$base_source_dir" "$base_version"
+
+new_source_dir=$(mktemp -d)
+checkout_and_build "$new_source_dir" "$new_version"
+
+cd "$original_dir"
+abi-compliance-checker -lib libsecp256k1 -old "${base_source_dir}/build/ABI.dump" -new "${new_source_dir}/build/ABI.dump"
+git worktree remove "$base_source_dir"
+git worktree remove "$new_source_dir"


fanquake:
Concept ~0. Looks like this tool [is essentially unmaintained](https://github.com/lvc/abi-dumper) (and only currently works with workarounds (-gdwarf-4)). There have been [requests for new releases](https://github.com/lvc/abi-dumper/issues/34) / [dwarf-5 support](https://github.com/lvc/abi-dumper/issues/33) but they haven't gotten any response.

Are current/previous users migrating to something else?

hebasto:
> Looks like this tool [is essentially unmaintained](https://github.com/lvc/abi-dumper)

Correct. I've noticed the same. But both packages still provided in Debian repositories. AND this approach is still mentioned in https://sourceware.org/glibc/wiki/Testing/ABI_checker.

tools/check-abi.sh
```diff
@@ -0,0 +1,37 @@
...
+set -eux
+
+configure_and_build() {
+    cmake -S "$1" \
+        -DCMAKE_C_COMPILER=gcc -DCMAKE_BUILD_TYPE=None -DCMAKE_C_FLAGS="-g -Og -gdwarf-4" \
+        -DSECP256K1_ENABLE_MODULE_ECDH=ON \
+        -DSECP256K1_ENABLE_MODULE_RECOVERY=ON \
+        -DSECP256K1_ENABLE_MODULE_EXTRAKEYS=ON \
+        -DSECP256K1_ENABLE_MODULE_SCHNORRSIG=ON \
+        -DSECP256K1_ENABLE_MODULE_ELLSWIFT=ON \
```
real-or-random, line 12: I bet that we'll forget to update this as soon as we add the next module. I suggest using dev mode instead. Yes, it's intended for developers, but I think it's acceptable in this case. Or what do you think?

doc/release-process.md
```diff
@@ -46,6 +46,13 @@ gcc -o ecdsa examples/ecdsa.c -I $dir/include -L $dir/lib*/ -l secp256k1 -Wl,-rp
    * if this is not a patch release
        * updates `_PKG_VERSION_*` and `_LIB_VERSION_*`  in `configure.ac` and
        * updates `project(libsecp256k1 VERSION ...)` and `${PROJECT_NAME}_LIB_VERSION_*` in `CMakeLists.txt`.
+
+ABI Compatibility must be checked with the [`check-abi.sh`](/tools/check-abi.sh) tool (installing packages `abi-dumper` and
+ `abi-compliance-checker` might be required):
+ ```shell
+ tools/check-abi.sh $(git describe --abbrev=0) master
```
real-or-random, line 53: On which platform(s) do we want to check?

real-or-random (commented):
Weak Concept ACK

Yeah, it's a bit sad to introduce an unmaintained tool. But this PR is just 44 LoCs, I think that's acceptable...

tools/check-abi.sh
hebasto: > I bet that we'll forget to update this as soon as we add the next module. I suggest using dev mode instead.

We might forget to update [`CMakePresets.json`](https://github.com/bitcoin-core/secp256k1/blob/master/CMakePresets.json) as well :D

For example, it lacks `SECP256K1_ENABLE_MODULE_ELLSWIFT`: https://github.com/bitcoin-core/secp256k1/blob/c545fdc374964424683d9dac31a828adedabe860/CMakePresets.json#L8-L12

doc/release-process.md
hebasto: Both packages are widely available among Linux distros, including Debian and Fedora:
- https://repology.org/project/abi-dumper/versions
- https://repology.org/project/abi-compliance-checker/versions

I have no preferences regarding the platforms to use for the ABI check. Debian works.

hebasto:
My point is that checking the ABI compatibility with a dedicated tool is strictly better than an educated guess.

The suggested set of tools:
- works fine
- is used by other projects (for example, [glibc](https://sourceware.org/glibc/wiki/Testing/ABI_checker))
- is widely adopted in Linux distros repositories

> Are current/previous users migrating to something else?

I've made searching twice, and did not find any alternatives users are migrating to. However, I did not ask people directly, for example glibc developers.

> Looks like this tool [is essentially unmaintained](https://github.com/lvc/abi-dumper) (and only currently works with workarounds (-gdwarf-4)). There have been [requests for new releases](https://github.com/lvc/abi-dumper/issues/34) / [dwarf-5 support](https://github.com/lvc/abi-dumper/issues/33) but they haven't gotten any response.

As this is a tool, not a dependency library, I think we can use it as long as it works regardless of its maintenance status. As far as I'm aware, there are no bugs in the tool that might affect our use case.

DWARF v4 is still well supported by the latest gcc. It is a legitimate option, not a "workaround".

fanquake:
> is used by other projects (for example, [glibc](https://sourceware.org/glibc/wiki/Testing/ABI_checker))

The wiki you link too describes using ABI dumper on a version of glibc that is more than a decade old. Is it actually being used *now*? (I can't find any use in the glibc repos). If so, I guess they aren't maintaining those docs, as the ones listed in that wiki page wouldn't work.

> It is a legitimate option, not a "workaround".

You are working around the fact that the tool doesn't support DWARF 5, which is the default for modern compilers.

hebasto:
> > It is a legitimate option, not a "workaround".
> 
> You are working around the fact that the tool doesn't support DWARF 5, which is the default for modern compilers.

How can using DWARF v5 affect our use case?

tools/check-abi.sh
real-or-random: ellswift is enabled by default, so we don't need to enable it here explicitly.

hebasto: I mean, that "we'll forget to update this as soon as we add the next module" might be applied to the `CMakePresets.json` as well.

And using dev mode implies building examples, which is redundant for the ABI check.

real-or-random:
The wiki page is indeed obsolete, so we should not count it as an argument for the tool.

But yeah, I also think `-gdwarf-4` is not an issue. It's just not the default.

tools/check-abi.sh
real-or-random: > I mean, that "we'll forget to update this as soon as we add the next module" might be applied to the `CMakePresets.json` as well.

Sure, but that doesn't make the point invalid. I think fewer code locations that we can forget to update is still better, even if it's not zero.  

> And using dev mode implies building examples, which is redundant for the ABI check.

Ok, what I had in mind is that it could still be disabled manually. But even if we don't, it doesn't really hurt to build the examples.

hebasto: > I suggest using dev mode instead.

Done.

hebasto:
@real-or-random's https://github.com/bitcoin-core/secp256k1/pull/1380#discussion_r1270002064 has been addressed.

real-or-random (commented):
Concept ACK, I'll have another close look soon

doc/release-process.md
real-or-random, line 53: Worth adding `--match "v*.*.*"` or  `--first-parent` to the git command?

doc/release-process.md
```diff
@@ -46,6 +46,13 @@ gcc -o ecdsa examples/ecdsa.c -I $dir/include -L $dir/lib*/ -l secp256k1 -Wl,-rp
    * if this is not a patch release
        * updates `_PKG_VERSION_*` and `_LIB_VERSION_*`  in `configure.ac` and
        * updates `project(libsecp256k1 VERSION ...)` and `${PROJECT_NAME}_LIB_VERSION_*` in `CMakeLists.txt`.
+
+ABI Compatibility must be checked with the [`check-abi.sh`](/tools/check-abi.sh) tool (installing packages `abi-dumper` and
```
jonasnick, line 50: Why is the ABI compatibility checked between step 1 and 2? Shouldn't this paragraph be part of the enumeration as well (and get number 2)?

But actually shouldn't this be done before updating the version number and writing the release notes?

jonasnick (commented):
Concept ACK. The tool & `check-abi.sh` appear to work as intended. `abi-dumper` and `abi-compliance-checker` are available on NixOS. I think this can be a helpful addition to the release process at the cost of adding a small shell script.

hebasto:
> Rebasing this will be good for testing. In the current branch, `git describe` says that the last tag is v0.3.2. :)

Rebased.

@jonasnick's comment has been addressed.

doc/release-process.md
hebasto: Thanks! [Reworked](https://github.com/bitcoin-core/secp256k1/pull/1380#issuecomment-1855841984).

hebasto:
@real-or-random's comments have been addressed. I apologize for any delay.

doc/release-process.md
hebasto: Thanks! [Done](https://github.com/bitcoin-core/secp256k1/pull/1380#issuecomment-1858802055).

tools/check-abi.sh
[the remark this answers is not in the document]
hebasto: Thanks! [Reworked](https://github.com/bitcoin-core/secp256k1/pull/1380#issuecomment-1858802055).

jonasnick (commented):
Can confirm that the tool works according to the help output and interprets the arguments correctly.

doc/release-process.md
```diff
@@ -37,7 +37,11 @@ gcc -o ecdsa examples/ecdsa.c -I $dir/include -L $dir/lib*/ -l secp256k1 -Wl,-rp
 
 ## Regular release
 
-1. Open a PR to the master branch with a commit (using message `"release: prepare for $MAJOR.$MINOR.$PATCH"`, for example) that
+1. Check the ABI compatibility with the [`check-abi.sh`](/tools/check-abi.sh) tool (installing the `abi-dumper` and `abi-compliance-checker` packages might be required):
+```shell
+tools/check-abi.sh
+```
```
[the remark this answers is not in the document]
hebasto: Thanks! Done.

hebasto:
> I'm happy to reACK quickly this with my suggestion applied...

Do you mean this [one](https://github.com/bitcoin-core/secp256k1/pull/1380#discussion_r1418651847)?

tools/check-abi.sh
```diff
@@ -0,0 +1,62 @@
...
+cd "$source_dir"
+
+git -c advice.detachedHead=false checkout "$new_version"
+current_build_dir=$(mktemp -d)
+cd "$current_build_dir"
+configure_and_build "$source_dir"
+abi-dumper src/libsecp256k1.so -o ABI.dump -lver "$new_version"
+cd "$source_dir"
+
+abi-compliance-checker -lib libsecp256k1 -old "${base_build_dir}/ABI.dump" -new "${current_build_dir}/ABI.dump"
```
real-or-random, line 62:
```suggestion
checkout_and_build() {
    git worktree add -d "$1"
    cd "$1"
    mkdir build && cd build
    cmake -S .. --preset dev-mode \
        -DCMAKE_C_COMPILER=gcc -DCMAKE_BUILD_TYPE=None -DCMAKE_C_FLAGS="-g -Og -gdwarf-4" \
        -DSECP256K1_BUILD_BENCHMARK=OFF \
        -DSECP256K1_BUILD_TESTS=OFF \
        -DSECP256K1_BUILD_EXHAUSTIVE_TESTS=OFF \
        -DSECP256K1_BUILD_CTIME_TESTS=OFF \
        -DSECP256K1_BUILD_EXAMPLES=OFF
    cmake --build . -j "$(nproc)"
}

echo "Comparing $base_version (base version) to $new_version (new version)"
echo

base_source_dir=$(mktemp -d)
checkout_and_build "$base_source_dir"
abi-dumper src/libsecp256k1.so -o ABI.dump -lver "$base_version"

current_source_dir=$(mktemp -d)
checkout_and_build "$current_source_dir"
abi-dumper src/libsecp256k1.so -o ABI.dump -lver "$new_version"

abi-compliance-checker -lib libsecp256k1 -old "${base_source_dir}/build/ABI.dump" -new "${current_source_dir}/build/ABI.dump"
```

real-or-random:
> Do you mean this [one](https://github.com/bitcoin-core/secp256k1/pull/1380#discussion_r1418651847)?

Sorry, I meant the one I've just posted now... :D (Sorry, not sure if this got lost in transit, or if I just forgot to click the "Comment" button...)

hebasto:
> > Do you mean this [one](https://github.com/bitcoin-core/secp256k1/pull/1380#discussion_r1418651847)?
> 
> Sorry, I meant the one I've just posted now... :D (Sorry, not sure if this got lost in transit, or if I just forgot to click the "Comment" button...)

Thanks! Done.

hebasto:
> > Do you mean this [one](https://github.com/bitcoin-core/secp256k1/pull/1380#discussion_r1418651847)?
> 
> Sorry, I meant the one I've just posted now... :D (Sorry, not sure if this got lost in transit, or if I just forgot to click the "Comment" button...)

Sorry. The new script builds the same commit twice.

real-or-random:
Argh, sorry -.-

I think we'll need to add the git revision as an arg to `git worktree add -d "$1"`, like `git worktree add -d "$1" <tag>`.. And then move the `git worktree` command out of the function again... I can look into it later, but I don't have the time right now.

hebasto:
Fixed minor issues in the recent implementation.

Suggesting to test by running `tools/check-abi.sh v0.3.2 v0.4.0`. The resulted reported must be the same as in https://github.com/bitcoin-core/secp256k1/pull/1415#issuecomment-1705504989.

real-or-random (approved):
ACK 74a4d974d5c81fbc437287dffc453028509682ab it compares the right commits now

jonasnick (commented):
re-Concept ACK 74a4d974d5c81fbc437287dffc453028509682ab

The tool continues to work for me.

PR #1382, opened 2023-07-24
refactor: Drop unused cast

hebasto:
The resulted code treats `memcpy`'s arguments exactly the same way as it does a few lines above: https://github.com/bitcoin-core/secp256k1/blob/c545fdc374964424683d9dac31a828adedabe860/src/hash_impl.h#L133


real-or-random (approved):
utACK 4f8c5bd76132a0d0242c7a29b666153f927800bb

PR #1383, opened 2023-07-25
util: remove unused checked_realloc

theuni:
Usage was removed in 6fe50439 . This should be a NOOP.

Noticed when analyzing for zenbleed exposure: stdlib calls that aren't optimized away.

In this case realloc isn't making it into the final binary, but as far as I can tell this is completely dead code and should be dropped.

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -152,14 +152,6 @@ static SECP256K1_INLINE void *checked_malloc(const secp256k1_callback* cb, size_
     return ret;
 }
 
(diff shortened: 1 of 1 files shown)


jonasnick (approved):
ACK b097a466c168dcdb3fde435ec4a1e0b63609f55d

real-or-random (approved):
ACK b097a466c168dcdb3fde435ec4a1e0b63609f55d

PR #1384, opened 2023-07-26
build: enable ellswift module via SECP_CONFIG_DEFINES

jonasnick:
...like the other modules.


real-or-random (approved):
utACK 78ca8807880def68a575b2487f374fdf4b49962d

PR #1290, opened 2023-04-29
cmake: Set `ENVIRONMENT` property for examples on Windows 

hebasto:
This PR simplifies running examples on Windows, because the DLL must resides either in the same folder where the executable is or somewhere in PATH.

It is an alternative to #1233.

Drafted for now as it is based on #1289.

PR #1389, opened 2023-07-28
ci: Run "Windows (VS 2022)" job on GitHub Actions

hebasto:
This PR solves one item in https://github.com/bitcoin-core/secp256k1/issues/1392.

In response to upcoming [limiting free usage of Cirrus CI](https://cirrus-ci.org/blog/2023/07/17/limiting-free-usage-of-cirrus-ci/), suggesting to move (partially?) CI tasks/jobs from Cirrus CI to [GitHub Actions](https://docs.github.com/actions) (GHA).

Here is example from my personal repo: https://github.com/hebasto/secp256k1/actions/runs/5806269046.

For security concerns, see:
- https://github.com/bitcoin/bitcoin/issues/28098#issuecomment-1651432106
- https://github.com/bitcoin/bitcoin/issues/28098#issuecomment-1651688197

I'm suggesting the repository "Actions permissions" as follows:

image

image

---

See build logs in my personal repo: https://github.com/hebasto/secp256k1/actions/runs/5692587475.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,39 @@
+name: CI
+on:
+  pull_request:
+  push:
+    branches:
+      - '**'
+    tags-ignore:
+      - '**'
+
+env:
+  SECP256K1_BENCH_ITERS: 2
+
+jobs:
+  win64-native:
+    name: "x86_64: Windows (VS 2022)"
+    # See: https://github.com/actions/runner-images#available-images.
+    runs-on: windows-2022
+
+    strategy:
+      fail-fast: false
+      matrix:
+        build_shared_libs: ['ON', 'OFF']
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Generate buildsystem
+        run: cmake -E env CFLAGS="/WX" cmake -B build -A x64 -DSECP256K1_ENABLE_MODULE_RECOVERY=ON -DSECP256K1_BUILD_EXAMPLES=ON -DBUILD_SHARED_LIBS=${{ matrix.build_shared_libs }}
+
+      - name: Build
+        run: cmake --build build --config RelWithDebInfo -- /p:UseMultiToolTask=true /maxCpuCount
+
+      - name: Check
+        run: |
+          ctest -C RelWithDebInfo --test-dir build -j ([int]$env:NUMBER_OF_PROCESSORS + 1)
+          build\src\RelWithDebInfo\bench_ecmult.exe
+          build\src\RelWithDebInfo\bench_internal.exe
+          build\src\RelWithDebInfo\bench.exe


hebasto:
> Could delete the Windows task on Cirrus in this pull...

Deleted.

hebasto:
Hmm... GH's not showing my recent push for some reasons...

hebasto:
Rebased on top of the merged #1290.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,34 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Generate buildsystem
+        run: cmake -E env CFLAGS="/WX" cmake -B build -A x64 -DSECP256K1_ENABLE_MODULE_RECOVERY=ON -DSECP256K1_BUILD_EXAMPLES=ON -DBUILD_SHARED_LIBS=${{ matrix.build_shared_libs }}
+
+      - name: Build
+        run: cmake --build build --config RelWithDebInfo -- /p:UseMultiToolTask=true /p:CL_MPcount=3
```
real-or-random, line 27: optional nit: The 3 could be computed as n + 1, where n is read from the system automatically, same below.

real-or-random (commented):
Approach ACK

Is there a better way to test PRs that change workflows (other than deploying them in your fork?) Or is the problem simply that we haven't enabled GitHub Actions at all? AFAIU, `on [pull_request]` should run the workflow from the merge commit, see for example https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/ . Is this correct?

hebasto:
> Is there a better way to test PRs that change workflows (other than deploying them in your fork?) Or is the problem simply that we haven't enabled GitHub Actions at all?

Enabled GitHub Actions in this repo should help.

> AFAIU, `on [pull_request]` should run the workflow from the merge commit, see for example [github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/) . Is this correct?

According to the [docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request), for `on.pull_request`:
> a workflow only runs when a `pull_request` event's activity type is `opened`, `synchronize`, or `reopened`.

hebasto:
Updated 72b8d875b21bfbd50dd95ec625d95dddcea37614 -> 5a3d45baa00057d6e27f2119ccdeea80108373f1 ([pr1389.03](https://github.com/hebasto/secp256k1/commits/pr1389.03) -> [pr1389.04](https://github.com/hebasto/secp256k1/commits/pr1389.04), [diff](https://github.com/hebasto/secp256k1/compare/pr1389.03..pr1389.04)):

- addressed @real-or-random's [comment](https://github.com/bitcoin-core/secp256k1/pull/1389#discussion_r1287427328)
- a few other improvements:
	- the workflow run name is set as the [commit message](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#run-name)
	- push event for tags are skipped
	- [`strategy.fail-fast`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast) is set to `false`

.github/workflows/ci.yml
hebasto: [Done](https://github.com/bitcoin-core/secp256k1/pull/1389#issuecomment-1670843751l).

In the MSBuild option list, it has been replaced with the high level [`-maxCpuCount`](https://learn.microsoft.com/en-us/visualstudio/msbuild/building-multiple-projects-in-parallel-with-msbuild?view=vs-2022#-maxcpucount-switch) switch.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,40 @@
...
+  pull_request:
+  push:
+    branches:
+      - '**'
+    tags-ignore:
+      - '**'
+
+env:
+  SECP256K1_BENCH_ITERS: 2
+  SECP256K1_TEST_ITERS: 16
```
real-or-random, line 12: Any reason not to keep the default of 64 for `TEST_ITERS`? We typically set lower values only in special environments such as QEMU.

real-or-random (commented):
GitHub Actions should be enabled now. I tried to close and re-open to trigger the workflow, but it didn't work. But yeah, happy to merge this (with the comment addressed) even if it doesn't run on this PR. We can always adjust.

hebasto:
@real-or-random 
> GitHub Actions should be enabled now.

Just to double check, are "Actions permissions" and "Workflow permissions" set to safe values?

hebasto:
> I tried to close and re-open to trigger the workflow, but it didn't work.

It should though...

Let me try once more.

.github/workflows/ci.yml
hebasto: I cannot recall what exactly caused my decision. Going to address your comment and make a new push after addressing https://github.com/bitcoin-core/secp256k1/pull/1389#issuecomment-1671120227.

achow101:
The suggested permissions were applied.

hebasto:
Hmm... I'm struggling to figure out the reason for no activity in https://github.com/bitcoin-core/secp256k1/actions.

.github/workflows/ci.yml
hebasto: Fixed.

real-or-random (approved):
utACK 7d05f78fab1c930abdc19da64cc5d576d3523cb7

Let's try merging it...

real-or-random:
Or actually... @hebasto Can you remove the second commit for now, if we're unsure whether this works at all.

hebasto:
> Or actually... @hebasto Can you remove the second commit for now, if we're unsure whether this works at all.

Removed.

real-or-random (approved):
utACK a2f7ccdecc4721d972f36d6aacc5f0c85ce0557d

hebasto:
image

PR #1390, opened 2023-07-28
tests: Replace counting_illegal_callbacks with CHECK_ILLEGAL_VOID

### LINKED ISSUE #1167: Setting counting_illegal_callback may hide failing tests

jonasnick:
Fixes #1167

### DIFF
--- a/src/modules/ecdh/tests_impl.h
+++ b/src/modules/ecdh/tests_impl.h
@@ -25,32 +25,19 @@ static int ecdh_hash_function_custom(unsigned char *output, const unsigned char 
 }
 
 static void test_ecdh_api(void) {
-    /* Setup context that just counts errors */
-    secp256k1_context *tctx = secp256k1_context_create(SECP256K1_CONTEXT_NONE);
     secp256k1_pubkey point;
     unsigned char res[32];
     unsigned char s_one[32] = { 0 };
-    int32_t ecount = 0;
     s_one[31] = 1;
 
-    secp256k1_context_set_error_callback(tctx, counting_illegal_callback_fn, &ecount);
-    secp256k1_context_set_illegal_callback(tctx, counting_illegal_callback_fn, &ecount);
-    CHECK(secp256k1_ec_pubkey_create(tctx, &point, s_one) == 1);
+    CHECK(secp256k1_ec_pubkey_create(CTX, &point, s_one) == 1);
 
     /* Check all NULLs are detected */
-    CHECK(secp256k1_ecdh(tctx, res, &point, s_one, NULL, NULL) == 1);
-    CHECK(ecount == 0);
-    CHECK(secp256k1_ecdh(tctx, NULL, &point, s_one, NULL, NULL) == 0);
-    CHECK(ecount == 1);
-    CHECK(secp256k1_ecdh(tctx, res, NULL, s_one, NULL, NULL) == 0);
-    CHECK(ecount == 2);
-    CHECK(secp256k1_ecdh(tctx, res, &point, NULL, NULL, NULL) == 0);
-    CHECK(ecount == 3);
-    CHECK(secp256k1_ecdh(tctx, res, &point, s_one, NULL, NULL) == 1);
-    CHECK(ecount == 3);
-
-    /* Cleanup */
-    secp256k1_context_destroy(tctx);
+    CHECK(secp256k1_ecdh(CTX, res, &point, s_one, NULL, NULL) == 1);
+    CHECK_ILLEGAL(CTX, secp256k1_ecdh(CTX, NULL, &point, s_one, NULL, NULL));
+    CHECK_ILLEGAL(CTX, secp256k1_ecdh(CTX, res, NULL, s_one, NULL, NULL));
+    CHECK_ILLEGAL(CTX, secp256k1_ecdh(CTX, res, &point, NULL, NULL, NULL));
+    CHECK(secp256k1_ecdh(CTX, res, &point, s_one, NULL, NULL) == 1);
 }
 
 static void test_ecdh_generator_basepoint(void) {
--- a/src/modules/extrakeys/tests_impl.h
+++ b/src/modules/extrakeys/tests_impl.h
@@ -132,7 +132,7 @@ static void test_xonly_pubkey_comparison(void) {
     CHECK_ILLEGAL_VOID(CTX, CHECK(secp256k1_xonly_pubkey_cmp(CTX, &pk1, &pk2) < 0));
     {
         int32_t ecount = 0;
-        secp256k1_context_set_illegal_callback(CTX, counting_illegal_callback_fn, &ecount);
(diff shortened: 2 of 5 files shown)


src/tests.c
```diff
@@ -6712,34 +6590,30 @@ static void run_pubkey_comparison(void) {
...
-        CHECK(ecount == 5);
-        CHECK(secp256k1_ec_pubkey_cmp(CTX, &pk2, &pk_tmp) > 0);
-        CHECK(ecount == 6);
+        CHECK_ILLEGAL_VOID(CTX, CHECK(secp256k1_ec_pubkey_cmp(CTX, &pk_tmp, &pk2) < 0));
+        {
+            int32_t ecount = 0;
+            secp256k1_context_set_illegal_callback(CTX, counting_illegal_callback_fn, &ecount);
+            CHECK(secp256k1_ec_pubkey_cmp(CTX, &pk_tmp, &pk_tmp) == 0);
+            CHECK(ecount == 2);
+            secp256k1_context_set_illegal_callback(CTX, NULL, NULL);
```
real-or-random, line 6612: Is it a bug that `secp256k1_ec_pubkey_cmp` calls the callback twice?

I mean, we never do this anywhere else, even if multiple arguments are illegal. In particular, `ARG_CHECK` makes the function return immediately on failure.

src/modules/extrakeys/tests_impl.h
```diff
@@ -139,29 +118,26 @@ static void test_xonly_pubkey_comparison(void) {
...
-    CHECK(ecount == 5);
-    CHECK(secp256k1_xonly_pubkey_cmp(CTX, &pk2, &pk1) > 0);
-    CHECK(ecount == 6);
+    CHECK_ILLEGAL_VOID(CTX, CHECK(secp256k1_xonly_pubkey_cmp(CTX, &pk1, &pk2) < 0));
+    {
+        int32_t ecount = 0;
+        secp256k1_context_set_illegal_callback(CTX, counting_illegal_callback_fn, &ecount);
+        CHECK(secp256k1_xonly_pubkey_cmp(CTX, &pk1, &pk1) == 0);
+        CHECK(ecount == 2);
+        secp256k1_context_set_illegal_callback(CTX, NULL, NULL);
```
real-or-random, line 138: (if you change it, also change it here)

src/modules/ecdh/tests_impl.h
```diff
@@ -26,31 +26,19 @@ static int ecdh_hash_function_custom(unsigned char *output, const unsigned char
 
 static void test_ecdh_api(void) {
     /* Setup context that just counts errors */
```
real-or-random, line 28: nit: Comment needs to be removed (or changed)

src/tests.c
```diff
@@ -5962,142 +5916,99 @@ static void run_ec_pubkey_parse_test(void) {
         0xB8, 0x00
     };
     unsigned char sout[65];
-    unsigned char shortkey[2];
+    unsigned char shortkey[2] = { 0 };
```
real-or-random, line 5919: nit:

```suggestion
    unsigned char shortkey[2] = { 0x03, 0x05 };
```

0x0500000000000000000000000000000000000000000000000000000000000000 is a valid x-coord with odd y-coord

real-or-random (commented):
utACK mod nits


This makes the test code much cleaner.

I suggest (this PR or followup)
 * changing `secp256k1_ec_pubkey_cmp` to call a callback at most once (see review)
 * adding a macro `CHECK_ERROR` for the `error_callback` (which should probably involve renaming `counting_illegal_callback` to just `counting_callback` -- the callback doesn't care what it's used for)

src/tests.c
jonasnick: I'm not sure how this would improves the tests.

src/tests.c
jonasnick: It's intentional, see the comment in `pubkey_cmp`:

```
        /* If the public key is NULL or invalid, ec_pubkey_serialize will call
         * the illegal_callback and return 0. In that case we will serialize the
         * key as all zeros which is less than any valid public key. This
         * results in consistent comparisons even if NULL or invalid pubkeys are
         * involved and prevents edge cases such as sorting algorithms that use
         * this function and do not terminate as a result. */
```

In particular, returning 0 immediately in pubkey_cmp if parsing fails does not signal an error to the color, but that both public keys are identical.

src/modules/ecdh/tests_impl.h
jonasnick: fixed

jonasnick:
> - adding a macro CHECK_ERROR for the error_callback (which should probably involve renaming counting_illegal_callback to just counting_callback -- the callback doesn't care what it's used for)

done

src/tests.c
real-or-random: Yeah sorry, I didn't think it through entirely.

src/tests.c
real-or-random: Oh, crazy. I still think we could get the best of both worlds and retain the current all-zeros behavior and at the same time call the callback only once. But that's better done in a separate PR then.

real-or-random (approved):
utACK fb5eb75a89598394f615eb55f0757e7c35e420ba

siv2r (commented):
The `error_callback` and `illegal_callback` are being set to `NULL` at the end of `test_keypair` fn (in `extrakeys/tests_impl.h`). Shouldn't we remove these too?

Github doesn't allow me to comment on these lines. So, here's the link: https://github.com/bitcoin-core/secp256k1/pull/1390/files#diff-07d311190e50e344466bc33ed56c60d2a15f4e8221bfc97792e1cd441b8f3e19L443-L444

siv2r (commented):
ACK fb5eb75, the tests locally pass on my machine (for all four commits).

real-or-random:
needs rebase :)

jonasnick:
rebased

jonasnick:
@siv2r 

> The error_callback and illegal_callback are being set to NULL at the end of test_keypair fn (in extrakeys/tests_impl.h).

Good catch, fixed.

siv2r:
reACK 7030364 (tests pass locally)

PR #1391, opened 2023-08-01
refactor: take use of `secp256k1_scalar_{zero,one}` constants (part 2)

theStack:
This is a follow-up to PR #1330, which replaced scalar variables that were set to zero and one via `secp256k1_scalar_set_int`, but missed instances which were directly initialized via SECP256K1_SCALAR_CONST (found via `git grep "SECP256K1_SCALAR_CONST(0,"`).


real-or-random (approved):
utACK https://github.com/bitcoin-core/secp256k1/pull/1391/commits/a1bd4971d6c6113f140143275141adb336fc3702

jonasnick (commented):
ACK a1bd4971d6c6113f140143275141adb336fc3702

Issue #1381, opened 2023-07-24
Policy for VERIFY_CHECK and #ifdef VERIFY  


real-or-random:
### Current situation

The purpose of `VERIFY_CHECK(cond)` is to assert `cond` in test builds, i.e., when `VERIFY` is defined. While `cond` is supposed to have no side effects, `VERIFY_CHECK` will evaluate it even in production builds, to make sure that no side effects are masked in production builds.

In the case of simple `cond`, we can rely on the ability of the compiler to detect that `cond` has no side effects and to omit its evaluation entirely. However, many recently introduced `cond` are neither simple nor have negligible computation costs. In this case, we usually wrap `VERIFY_CHECK(cond)` in an `#ifdef VERIFY` block, which forces the evaluation to be omitted in production, at the cost of foregoing the aforementioned "side-effect safety" of `VERIFY_CHECK`. [1]

On top of this, we now have (full) functions like `secp256k1_fe_verify`, which internally perform `VERIFY_CHECK`s.

I don't think anything we currently do is wrong, but sometimes the borders are a bit arbitrary, and with a mix of `#ifdef VERIFY` and no `#ifdef VERIFY`, and a mix of uppercase and lowercase, also readability and consistency suffer.

### Options

When it comes to forcing evaluations to be omitted, we can either keep the current convention and force omitting on a case-by-case basis. Alternatively, we could just always force omitting, e.g., by defining `VERIFY_CHECK` to be empty in production. That will keep things simple and save us the decision of whether to force omitting. The comment in https://github.com/bitcoin-core/secp256k1/pull/904#issuecomment-1540601651 appears to agree with this.  

When it comes to readability, I think that `#ifdef` blocks tend to clutter the code. If we redefine `VERIFY_CHECK` to be empty in production, we could omit many `#ifdef` lines. Even if we keep two types of `VERIFY_CHECK`s, I suggest introducing a separate macro instead to replace `#ifdef`.

One remaining purpose of the `#ifdef`s is to highlight the test code blocks, e.g., [when we call `secp256k1_fe_verify` in `field_impl.h`](https://github.com/bitcoin-core/secp256k1/blob/c545fdc374964424683d9dac31a828adedabe860/src/field_impl.h#L34-L45). But then we could just make these function names uppercase. And if we want blocks/grouping for better readability, we can simply use empty lines instead of `#ifdef` lines.

So I suggest 
 - redefining `VERIFY_CHECK` to be empty in production
[...]

PR #1393, opened 2023-08-04
Implement new policy for VERIFY_CHECK and #ifdef VERIFY (issue #1381)

theStack:
As suggested in #1381, this PR reworks the policy for VERIFY_CHECK and when to use #ifdef VERIFY, by:
- redefining VERIFY_CHECK to empty in production (non-VERIFY) mode
- removing many then superflous #ifdef VERIFY blocks (if they exclusively contained VERIFY_CHECKs)
- introducing uppercase macros around verify_ functions and using them for better readabiliy

What is _not_ included yet is the proposed renaming from "_check" to "_assert":
> And while we're touching this anyway, we could consider renaming "check" to "assert", which is a more precise term. (In fact, if we redefine VERIFY_CHECK to be empty in production, we have almost reimplemented assert.h...)

This should be easy to achieve with simple search-and-replace (e.g. using sed), but I was hesitant as this would probably case annoying merge conflicts on some of the open PRs. Happy to add this if the rename if desired (#1381 didn't get any feedback about the renaming idea yet).


theStack:
Rebased on master (since #1348 has been merged) and updated the PR description accordingly.

theStack:
Rebased on master again and introduced an uppercase macro for `secp256k1_verify_scalar` as well (introduced in recently merged #1373).

theStack:
Rebased on master (CI was failing previously).

src/util.h
```diff
@@ -132,15 +132,12 @@ static const secp256k1_callback default_error_callback = {
...
-#define VERIFY_SETUP(stmt)
-#elif defined(VERIFY)
+/* Like assert(), but when VERIFY is defined. */
+#if defined(VERIFY)
 #define VERIFY_CHECK CHECK
 #define VERIFY_SETUP(stmt) do { stmt; } while(0)
 #else
-#define VERIFY_CHECK(cond) do { (void)(cond); } while(0)
+#define VERIFY_CHECK(cond)
 #define VERIFY_SETUP(stmt)
```
stratospher, line 141: c723b85: what does `VERIFY_SETUP` do? it looks really similar to `VERIFY_CHECK` and used in only 2 places in the codebase.

src/scalar_8x32_impl.h
```diff
@@ -179,9 +179,8 @@ static void secp256k1_scalar_cadd_bit(secp256k1_scalar *r, unsigned int bit, int
     r->d[7] = t & 0xFFFFFFFFULL;
 
     secp256k1_scalar_verify(r);
-#ifdef VERIFY
     VERIFY_CHECK((t >> 32) == 0);
-#endif
+    VERIFY_CHECK(secp256k1_scalar_check_overflow(r) == 0);
```
stratospher, line 183: 33314ca: can we remove this since it's already done inside `secp256k1_scalar_verify`?

real-or-random: This will be resolved by a rebase because it has already been done in #1184, see https://github.com/bitcoin-core/secp256k1/pull/1184#discussion_r1371803481.

src/util.h
real-or-random: The only two uses trace back to 72ae443afb536221a8523646ea4a88162341c3df where the macro was introduced.

My thinking is that the purpose of `VERIFY_SETUP` is to wrap code whose only purpose is to perform side effects ("setup") for further `VERIFY_CHECK`s. But assuming this interpretation, I don't really understand 72ae443afb536221a8523646ea4a88162341c3df. 

The effect of the two existing `VERIFY_SETUP` uses is that they initialize `r->x` and `r->y` but that's done anyway in the current code, immediately afterward: 
https://github.com/bitcoin-core/secp256k1/blob/5814d8485cbb074115c03b587b6799a812e0f267/src/ecmult_const_impl.h#L90-L95

@sipa Is there a reason why you kept these in #1184?

Anyway, I suggest:
 - Drop the two uses of `VERIFY_SETUP`.
 - Drop `VERIFY_SETUP`. (Our implicit convention for such code is to nest it in an `#ifdef VERIFY` block.)

theStack:
Rebased on master (taking the new function `_scalar_half` into account, introduced in #1184), fixed https://github.com/bitcoin-core/secp256k1/pull/1393#discussion_r1410107068 and added commits to drop `VERIFY_SETUP` as suggested.

src/scalar_8x32_impl.h
theStack: > [33314ca](https://github.com/bitcoin-core/secp256k1/commit/33314ca3b1dc55cdc4acf8f7327cfb57ba700101): can we remove this since it's already done inside `secp256k1_scalar_verify`?

Oops, no idea why I added this line in a commit which is supposed to only remove lines (and only add blank ones for better readability). Removed now.

> This will be resolved by a rebase because it has already been done in #1184, see [#1184 (comment)](https://github.com/bitcoin-core/secp256k1/pull/1184#discussion_r1371803481).



@real-or-random: The comment link refers to the same expression, but in a different function. The one in the review comment (being in `secp256k1_scalar_cadd_bit`) was indeed still there after the rebase.

stratospher (commented):
ACK bb46723.

real-or-random (approved):
utACK bb4672342efce7fae1cfd30e007c6835a25286a7

It looks a bit strange that the new uppercase macros have a `SECP256K1_` prefix, but our existing macros don't. If you ask me, we should probably add a prefix to everything, but that could be controversial since it touches many lines. Let's not hold this PR up.

PR #1395, opened 2023-08-07
tests: simplify `random_fe_non_zero` (remove loop limit and unneeded normalize)

theStack:
`random_fe_non_zero` contains a loop iteration limit that ensures that we abort if `random_fe` ever yielded zero more than ten times in a row. This construct was first introduced in PR #19 (commit 09ca4f32) for random non-square field elements and was later refactored into the non-zero helper in PR #25 (commit 6d6102fe). The copy-over to the exhaustive tests happened recently in PR #1118 (commit 0f864207).

This case seems to be practically irrelevant and I'd argue for keeping things simple and removing it (which was already suggested in https://github.com/bitcoin-core/secp256k1/pull/1118#discussion_r1067259954); if there's really a worry that the test's random generator is heavily biased towards certain values or value ranges then there should consequently be checks at other places too (e.g. directly in `random_fe` for 256-bit values that repeatedly overflow, i.e. >= p).

Also, the _fe_normalize call is not needed and can be removed, as the result of `random_fe` is already normalized.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -46,6 +46,7 @@ noinst_HEADERS += src/precomputed_ecmult.h
 noinst_HEADERS += src/precomputed_ecmult_gen.h
 noinst_HEADERS += src/assumptions.h
 noinst_HEADERS += src/checkmem.h
+noinst_HEADERS += src/testutil.h
 noinst_HEADERS += src/util.h
 noinst_HEADERS += src/int128.h
 noinst_HEADERS += src/int128_impl.h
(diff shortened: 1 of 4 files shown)


real-or-random (approved):
utACK 37974c5e007ba70056dc41e61f6c226075305c4f

real-or-random:
To get rid of the duplication, we could move the `ge_equals_*` functions to `group` (I guess it doesn't hurt to have them in the code) and the `random_fe` to `testrand`. Or we could just move everything to a new `testutil.h`.

real-or-random (approved):
utACK c45b7c4fbbf41b011f138c465a58322a36664fd3

siv2r (commented):
ACK `c45b7c4` (reviewed the changes and tests for both the commits passed locally).

PR #1397, opened 2023-08-09
ci: Remove "Windows (VS 2022)" task from Cirrus CI

hebasto:
A follow-up for https://github.com/bitcoin-core/secp256k1/pull/1389.

https://github.com/bitcoin-core/secp256k1/pull/1389#issuecomment-1671345100:
> Or actually... hebasto Can you remove the second commit for now, if we're unsure whether this works at all.

---

Second commit effect:
- [before (master branch)](https://github.com/bitcoin-core/secp256k1/actions/runs/5809860925):
image

- [after (this PR)](https://github.com/bitcoin-core/secp256k1/actions/runs/5810140851):
image


hebasto:
> Nice, the second commit here clearly shows that the GHA runs its config from the merge commit (and not from master).

FWIW, from the "Checkout" step log:
```
"C:\Program Files\Git\bin\git.exe" checkout --progress --force refs/remotes/pull/1397/merge
```

Issue #1392, opened 2023-08-04
ci: Future of CI after Cirrus pricing change 


real-or-random:
Roadmap (keeping this up to date):

I think the natural way forward for us is:
 - [x] Move native Windows tasks to GitHub Actions (#1389 and #1397)
 - [x] Move SageMath task to GitHub Actions (#1399)
 - [x] Move native macOS tasks to GitHub Actions, this will convert them to x86_64 unfortunately (#1394, #1404)
 - Move Linux tasks to the Bitcoin Core persistent workers or alternatively to GitHub Actions
   - [x] wine/msvc tasks (converting them to native windows in #1401)
   - [x] wine/mingw tasks (#1398)
   - [x] actual normal Linux tasks (#1396)
   - [x] special tasks like qemu/sanitizers (#1406, #1409)

Possible follow-ups:
 - ~Consider using [artifacts](https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) to move the Docker image from the Docker build job to the actual CI job (for Linux tasks).~ (Not worth the hassle, the current approach seems to work well.)<details><summary>Details</summary>
 This should be a [cleaner solution](https://github.com/bitcoin-core/secp256k1/pull/1398#issuecomment-1677074743), but it adds some complexity. It's also worth checking if this avoids [network issues](https://github.com/bitcoin-core/secp256k1/pull/1396#issuecomment-1684108729). In terms of delay, [this adds about 12 min uploading time to the Docker build job, but avoids about 1 min delay in the actual CI jobs](https://github.com/bitcoin-core/secp256k1/pull/1396#issuecomment-1686387785) as compared to the current solution that relies purely on the GHA cache (#1398).  So this will speed up CI if we could avoid re-uploading existing artifacts, e.g., have another digest file that just stores the SHA256, and re-upload only if the SHA does not match. But all of this is probably not worth the complexity if the current approach with the cache turns out to be good enough.</details>
 - [x] Enable Valgrind on macOS again now that the macOS tasks run on x86_64 (#1151, #1412)
 - Bring back some ARM testing (see for details: https://github.com/bitcoin-core/secp256k1/pull/1394#issuecomment-1682321329)
   - [x] Add a task for ctimetest on ARM64/Linux/Valgrind on Cirrus CI using free minutes or the self-hosted runner  (#1426)
   - [x] Add more ARM64 tests on QEMU (#1414)
[...]

PR #1398, opened 2023-08-10
ci, gha: Add Windows jobs based on Linux image

hebasto:
Solves one item in https://github.com/bitcoin-core/secp256k1/issues/1392 partially.


.github/workflows/ci.yml
```diff
@@ -8,9 +8,236 @@ on:
...
+              ECDH="${{ env.ECDH }}" \
+              RECOVERY="${{ env.RECOVERY }}" \
+              SCHNORRSIG="${{ env.SCHNORRSIG }}" \
+              ELLSWIFT="${{ env.ELLSWIFT }}" \
+              SECP256K1_TEST_ITERS="${{ env.SECP256K1_TEST_ITERS }}" \
+              BENCH="${{ env.BENCH }}" \
+              SECP256K1_BENCH_ITERS="${{ env.SECP256K1_BENCH_ITERS }}" \
+              CTIMETESTS="${{ env.CTIMETESTS }}" \
+              EXAMPLES="${{ env.EXAMPLES }}" \
+              ${{ matrix.configuration.env_var }}
```
real-or-random, line 125: I suspect this is necessary because `addnab/docker-run-action@v3`. I wonder if there's a better way or a different action we could use, or if we could just run docker manually.

But okay, yes, I am starting to see that this approach of using the GHA cache also has its disadvantages.

real-or-random: > run docker manually

We could perhaps do something like this (assuming we add a unique prefix our env variables) https://stackoverflow.com/a/61955515

real-or-random:
> In the meantime, reviewers can view the builds in my personal repo: [hebasto/secp256k1/actions/runs/5818492202](https://github.com/hebasto/secp256k1/actions/runs/5818492202).

Caching doesn't seem to work properly. I suggest trying to remove the `mode=min`.

hebasto:
> > In the meantime, reviewers can view the builds in my personal repo: [hebasto/secp256k1/actions/runs/5818492202](https://github.com/hebasto/secp256k1/actions/runs/5818492202).
> 
> Caching doesn't seem to work properly.

Which job?

real-or-random:
> > > In the meantime, reviewers can view the builds in my personal repo: [hebasto/secp256k1/actions/runs/5818492202](https://github.com/hebasto/secp256k1/actions/runs/5818492202).
> > 
> > 
> > Caching doesn't seem to work properly.
> 
> Which job?

Sorry, all jobs, but I meant to include a link to the job. But I think I was wrong. The "Building container" stages takes ~5min, but all of this time seems to be spent downloading from the cache?! I didn't expect that it's that slow.

hebasto:
> The "Building container" stages takes ~5min, but all of this time seems to be spent downloading from the cache?!

Correct.

> I didn't expect that it's that slow.

That is what I meant in https://github.com/bitcoin-core/secp256k1/pull/1396#issuecomment-1669712120 :)

real-or-random:
Okay! But do you agree with my observation that GHA cache vs. GHCR doesn't make a significant difference here? Then we can ignore that piece when deciding what to use for Linux tasks. 

(Heck, for Linux native tasks with standard gcc and clang, using Docker image at all may be faster... But I'd say let's worry about optimizations later.)

hebasto:
Reworked. The third-party `addnab/docker-run-action` has been replaced with own reusable workflow.

.github/workflows/ci.yml
hebasto: [Reworked](https://github.com/bitcoin-core/secp256k1/pull/1398#issuecomment-1674068461).

real-or-random:
Hm, sad to see that it's already starting to get flaky: https://github.com/bitcoin-core/secp256k1/actions/runs/5827144103/job/15802793632?pr=1398 Perhaps this helps: https://github.com/moby/buildkit/issues/3969

Anyway, I like the changes.

hebasto:
> Hm, sad to see that it's already starting to get flaky: [bitcoin-core/secp256k1/actions/runs/5827144103/job/15802793632?pr=1398](https://github.com/bitcoin-core/secp256k1/actions/runs/5827144103/job/15802793632?pr=1398)

From experience, I would estimate the rate of such kind of failure about 1 (one) per cent of total runs.

.github/workflows/ci.yml
```diff
@@ -8,9 +8,153 @@ on:
...
+      RECOVERY: 'yes'
+      SCHNORRSIG: 'yes'
+      CTIMETESTS: 'no'
+    outputs:
+      json: ${{ steps.result.outputs.json }}
+    steps:
+      - name: Put environment variables into a standard job output
+        id: result
+        run: |
+          echo "json=$(echo '${{ toJSON(env) }}' | jq -c '.')" >> "$GITHUB_OUTPUT"
```
real-or-random, line 75: This still seems a bit overkill to me. Couldn't we list the env variables just in the `mingw_debian` job?

What I imagine is something like this:
```
mingw_debian:
  env:
    - WRAPPER_CMD: 'wine'
```
Am I right that this doesn't work, because it actually sets these environment variables, but then it's not clear to `run-in-docker` which variables should be passed to the inside of the container?

But what about setting `common_env_vars` only inside `mingw_debian`? Something like  
```
mingw_debian:
  with:
      common_env_vars: |
        WRAPPER_CMD=wine
        [...]
```

like this, with a more consistent syntax   (using an array of size 1 for `common_configuration`)

```
mingw_debian:
  strategy:
    matrix:
      common_configuration:
        - common_env_vars:
           WRAPPER_CMD: 'wine'
           [...]
      configuration:
[...]
```

hebasto: https://docs.github.com/en/actions/using-workflows/reusing-workflows#limitations:
>  - Any environment variables set in an `env` context defined at the workflow level in the caller workflow are not propagated to the called workflow.
>
>  - Similarly, environment variables set in the env context, defined in the called workflow, are not accessible in the env context of the caller workflow. Instead, you must use outputs of the reusable workflow.

hebasto: > But what about setting `common_env_vars` only inside `mingw_debian`?

It seems working .Going to give it a try.

hebasto: > But what about setting `common_env_vars` only inside `mingw_debian`?

This works for  `common_env_vars`, but we still need a way to pass global environment variables to the called workflow or duplicate them inside `mingw_debian`.

hebasto:
Reworked with using a composite action to avoid environment variables hassles.

Using `compression=estargz,force-compression=true` for caching as suggested by @real-or-random offline.

hebasto:
The only thing I'm not sure about is that whether PR branches will hit image cache from the master branch.

real-or-random:
> Using `compression=estargz,force-compression=true` for caching as suggested by @real-or-random offline.

Okay, as @hebasto figured out that doesn't work with the GHA cache backend for now (see https://docs.docker.com/build/cache/backends/#cache-compression). Well fine.

hebasto:
> > Using `compression=estargz,force-compression=true` for caching as suggested by @real-or-random offline.
> 
> Okay, as @hebasto figured out that doesn't work with the GHA cache backend for now (see [docs.docker.com/build/cache/backends/#cache-compression](https://docs.docker.com/build/cache/backends/#cache-compression)). Well fine.

Dropped.

real-or-random:
This looks pretty clean now. Will perform a closer review soon. 

One nit: Perhaps it's now time to rename `cirrus.ci` to `ci.sh`.

Some observations:
 - All the Docker stuff is a bit more complex on GHA than on Cirrus CI, but I still believe Docker is a good idea: It makes it easier to reproduce failures locally, and in principle, it makes it easy to switch between different CI providers. For example, we can easily switch back to Cirrus (self-hosted) runners if we keep the Docker images.
 - Store/restoring the Docker build cache is still a bit slow. It's a bit sad that downloading the cache takes almost as long as running the actual CI script. But we may be able to improve this later. And even if we don't manage to improve, I think the build times are acceptable. (Moreover, the native Windows builds don't use Docker, so we still get some immediate feedback for the most obvious problems.)

hebasto:
> One nit: Perhaps it's now time to rename `cirrus.ci` to `ci.sh`.

Done.

.github/actions/run-in-docker-action/action.yml
```diff
@@ -0,0 +1,40 @@
...
+        context: .
+        file: ${{ inputs.dockerfile }}
+        tags: ${{ inputs.tag }}
+        push: false
+        load: true
+        cache-from: type=gha
+
+    - run: >
+        docker run \
+          $(echo '${{ toJSON(env) }}' | jq -r 'keys[] | "-e \(.)"' | paste -sd " ") \
```
real-or-random, line 36:
```suggestion
          # Tell Docker to pass environment variables in `env` (which includes those set via `GITHUB_ENV`) into the container.
          $(echo '${{ toJSON(env) }}' | jq -r 'keys[] | "--env \(.)"' | paste -sd " ") \
```

.github/actions/run-in-docker-action/action.yml
```diff
@@ -0,0 +1,40 @@
...
+    description: 'A command to run in a container'
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        for var in "$(echo '${{ toJSON(matrix.configuration.env_vars) }}' | jq -r 'to_entries | .[] | "\(.key)=\(.value)"')"; do
+          echo "$var" >> "$GITHUB_ENV"
+        done
+        echo "MAKEFLAGS=-j$(($(nproc) + 1))" >> "$GITHUB_ENV"
```
real-or-random, line 20: nit: This will probably override `MAKEFLAGS` if in `env_vars`.  Perhaps it's a bit cleaner to simply *prepend* `MAKEFLAGS` with `$(($(nproc) + 1))`, and do this inside `ci.sh`?

.github/actions/run-in-docker-action/action.yml
```diff
@@ -0,0 +1,40 @@
+name: 'CI script in Docker'
+description: 'Run a CI script in a Docker container'
```
real-or-random, line 2:
```suggestion
name: 'Run in Docker with environment'
description: 'Run a command in a Docker container, while passing explicitly set environment variables into the container.'
```

.github/workflows/ci.yml
```diff
@@ -8,9 +8,203 @@ on:
...
+        if: ${{ always() }}
+      - run: cat bench.log || true
+        if: ${{ always() }}
+      - run: cat config.log || true
+        if: ${{ always() }}
+      - run: cat test_env.log || true
+        if: ${{ always() }}
+      - name: CI env
+        run: env
+        if: ${{ always() }}
```
real-or-random, line 111: Could this be refactored into an action? Or is this overkill?

real-or-random (commented):
ACK mod nits

Do you think it makes sense to retain the Cirrus config commented out, like sometimes done in Core? This would require a bit more work here to make sure it still works now that we split the image.

real-or-random:
@MarcoFalke if you want to have a look

.github/workflows/ci.yml
hebasto: > Could this be refactored into an action? Or is this overkill?

In that case it won't show separated entries in the log.

hebasto:
> Do you think it makes sense to retain the Cirrus config commented out, like sometimes done in Core?

We have the commit history for that, don't we?

.github/actions/run-in-docker-action/action.yml
hebasto: To keep the scope of this PR limited, is it OK to just set `mingw_debian.env.MAKEFLAGS: '-j3'`?

hebasto:
Updated 1212e9a556c3e58582f8677ebc2bb8585495acc9 -> 67d6118606c95f2e92e42bd492eed14700c9edfb ([pr1398.06](https://github.com/hebasto/secp256k1/commits/pr1398.06) -> [pr1398.07](https://github.com/hebasto/secp256k1/commits/pr1398.07), [diff](https://github.com/hebasto/secp256k1/compare/pr1398.06..pr1398.07)):

- addressed @real-or-random's comments

.github/actions/run-in-docker-action/action.yml
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/secp256k1/pull/1398#issuecomment-1675222830).

.github/actions/run-in-docker-action/action.yml
hebasto: > Or rely on the global `env.MAKEFLAGS: '-j4'`?

[Implemented](https://github.com/bitcoin-core/secp256k1/pull/1398#issuecomment-1675222830).

.github/actions/run-in-docker-action/action.yml
hebasto: Thanks! [Updated](https://github.com/bitcoin-core/secp256k1/pull/1398#issuecomment-1675222830).

Bezzo100-lab:
Help

Sent from my Metro By T-Mobile 4G LTE Android Device
Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

hebasto:
Being rebased on top of this PR, https://github.com/bitcoin-core/secp256k1/pull/1396 got a very small diff, btw.

hebasto:
Updated 67d6118606c95f2e92e42bd492eed14700c9edfb -> a2a7fc7aa3fb8be68efa126848348daf13bd3bb6 ([pr1398.07](https://github.com/hebasto/secp256k1/commits/pr1398.07) -> [pr1398.08](https://github.com/hebasto/secp256k1/commits/pr1398.08), [diff](https://github.com/hebasto/secp256k1/compare/pr1398.07..pr1398.08)).

Environment variables from `strategy.matrix` are passed to `steps.*.env` directly now.

.github/workflows/ci.yml
```diff
@@ -8,9 +8,207 @@ on:
...
+      CFLAGS: '-nologo -diagnostics:caret'
+      LDFLAGS: '-Xlinker -Xlinker -Xlinker -nologo'
+
+    strategy:
+      fail-fast: false
+      matrix:
+        configuration:
+          - job_name: 'x86_64 (MSVC): Windows (Debian stable, Wine)'
+            env_vars:
+              HOST: 'x86_64-w64-mingw32'
```
real-or-random, line 146: Is there a reason why you moved this down to `configurations` as compared to the previous version of the PR?

.github/workflows/ci.yml
```diff
@@ -8,9 +8,207 @@ on:
...
+              CC: '/opt/msvc/bin/x86/cl'
+              AR: '/opt/msvc/bin/x86/lib'
+              NM: '/opt/msvc/bin/x86/dumpbin -symbols -headers'
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: CI script
+        env: ${{ matrix.configuration.env_vars }}
```
real-or-random, line 168: Nice! So simple once you see it. :)

.github/workflows/ci.yml
real-or-random: Ok yes, let's leave it like this for now. 

(I'm not happy with this and it should be reworked in a separate PR. I had introduced this printing of files to make sure that we always have all logs, and I think that's a good idea in general. But it's not a good idea to pipe it *only* to files. We should pipe it to files but also keep stdout/stderr, e.g., using `tee`. People get confused about this over and over because they can't find the error messages in the failing tasks etc.)

.github/workflows/ci.yml
hebasto: > Is there a reason why you moved this down to `configurations` as compared to the previous version of the PR?

Yes. The reason is to do not leave `matrix.configuration.env_vars` empty. Otherwise, `env: ${{ matrix.configuration.env_vars }}` fails.

hebasto:
Addressed @real-or-random's [comment]( https://github.com/bitcoin-core/secp256k1/pull/1398#issuecomment-1674495293).

.github/workflows/ci.yml
real-or-random: Oh okay, hm. What's the error message? (Just curious, perhaps I have an idea ...)

real-or-random: Have you tried `env_vars: {}` ? It's the empty YAML dict, see https://stackoverflow.com/questions/33510094/syntax-for-empty-dictionary-in-yaml

hebasto:
Updated f4347ce35ea434c031e4a3eae70b36587f39ba77 -> 9ade5743f19f91a3da7ef2b1fe38be0bc6914362 ([pr1398.09](https://github.com/hebasto/secp256k1/commits/pr1398.09) -> [pr1398.10](https://github.com/hebasto/secp256k1/commits/pr1398.10), [diff](https://github.com/hebasto/secp256k1/compare/pr1398.09..pr1398.10)):

- addressed @real-or-random's [comment](https://github.com/bitcoin-core/secp256k1/pull/1398#discussion_r1292275706)

.github/workflows/ci.yml
hebasto: > Have you tried `env_vars: {}` ? It's the empty YAML dict, see [stackoverflow.com/questions/33510094/syntax-for-empty-dictionary-in-yaml](https://stackoverflow.com/questions/33510094/syntax-for-empty-dictionary-in-yaml)

Thank you! It [works](https://github.com/bitcoin-core/secp256k1/pull/1398#issuecomment-1675902686).

hebasto:
> > This PR relies on GitHub Actions cache rather on GitHub Container Registry (in comparison to #1396).
> 
> This works because this repo doesn't use ccache and thus, the docker images are the only thing in the cache and won't exceed the 10GB limit?

Yes, it does.

As it was pointed by @real-or-random offline, there are alternatives for the GitHub's cache, for example, the [buildjet/cache](https://buildjet.com/for-github-actions/docs/guides/migrating-to-buildjet-cache) action.

hebasto:
The actual use of the [cache storage](https://github.com/bitcoin-core/secp256k1/actions/caches) is as follows:
image

real-or-random:
> This works because this repo doesn't use ccache and thus, the docker images are the only thing in the cache and won't exceed the 10GB limit?

Indeed, but we already ran into problems here a few days ago with the 10GB limit.

The docker images in the cache are quite large (see https://github.com/bitcoin-core/secp256k1/actions/caches), so we can fit at most builds in the limit, e.g., one PR + master. When we have multiple PRs updating the Dockerfiles, we see cache misses. I don't think that's a big issue. We usually don't have multiple PRs updating CI. In the worst case, the images needs to be rebuilt more often than necessary, which just adds <30 min of delay to the CI. Moreover, we can reduce the size of the images a bit. Just these the following two should get us multiple GBs back:
 - [ ] removing GCC build files after the build (currently testing this locally, I'll have a patch soon)
 - [ ] simply using the official https://hub.docker.com/r/sagemath/sagemath for the sage job instead of installing sagemath in our Linux image

However, a conceptual problem is that we rely on the images not being evicted between the "build docker image" job and the actual CI jobs. That contradicts the philosophy of a cache a bit. When we have multiple concurrent CI runs going on, this may fail:
 - PR1 stores docker image A in cache
 - PR2 stores docker image B in cache, will evict image A
 - PR1 tries to run the CI script, but A has been evicted: So A is rebuilt as part of every single job. This is slow, may fail due to insufficient disk space, and may evict B again 
 - ...

There's a possible third way to move the docker image from the build job to the actual CI job: Use artifacts: https://github.com/actions/upload-artifact https://github.com/actions/download-artifact This seems to be the proper way to move data between jobs. Artifacts won't be evicted automatically, but they'll be cleaned up after 90 days, so this is probably the sweet spot between GHCR and the GHA cache. We'd still use the GHA cache for caching the docker build, but just the way we transfer the image from one job to another (in the same CI run) would use artifacts.

I'm not entirely sure how to proceed. I think I'd like to see at least one of the two aforementioned issues addressed. If we can reduce the size of the images for now, that will probably good enough. Even if we then have issues with concurrent CI runs, that's not the end of the world. (One can just re-run manually...) But it's certainly not convenient, and it's a bit hacky, so having a reliable way to move images from one job to another (e.g., artifacts) will be better.

.github/workflows/ci.yml
```diff
@@ -8,9 +8,209 @@ on:
...
+        if: ${{ always() }}
+      - run: cat bench.log || true
+        if: ${{ always() }}
+      - run: cat config.log || true
+        if: ${{ always() }}
+      - run: cat test_env.log || true
+        if: ${{ always() }}
+      - name: CI env
+        run: env
+        if: ${{ always() }}
```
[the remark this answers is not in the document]
hebasto: It is not supported, unfortunately.

See: https://github.com/actions/runner/issues/1182.

ci/linux-debian-wine.Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM debian:stable-slim
+
+SHELL ["/bin/bash", "-c"]
+
+WORKDIR /root
+
+# The "wine" package provides a convenience wrapper that we need
+RUN dpkg --add-architecture i386 && apt-get update && apt-get install --no-install-recommends -y \
+        autoconf automake libtool make \
```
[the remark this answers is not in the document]
real-or-random: Hm, I guess it's okay in this case.

real-or-random:
Just the following two should get us multiple GBs back:
> 
>    * [ ]  removing GCC build files after the build (currently testing this locally, I'll have a patch soon)
>
>    * [ ]  simply using the official [hub.docker.com/r/sagemath/sagemath](https://hub.docker.com/r/sagemath/sagemath) for the sage job instead of installing sagemath in our Linux image

[This commit](https://github.com/real-or-random/secp256k1/commit/20b208df0ab22a2afd80f3cf172f100365766e7b) reduces the *uncompressed* size from 9.37GB to 3.55GB. The GHA cache uses compression, but this should still make a large difference.  

Cirrus CI will fail now due to Sagemath being unavailable. @hebasto If you want to keep the scope of this PR small, perhaps we should move the Sagemath job to GHA in a separate PR first.

real-or-random:
> As it was pointed by @real-or-random offline, there are alternatives for the GitHub's cache, for example, the [buildjet/cache](https://buildjet.com/for-github-actions/docs/guides/migrating-to-buildjet-cache) action.

Yep, this would give us 20 GB, but after I thought about it, I'd like to avoid it if possible. It adds another infrastructure dependency, we can't use the nice GHA interface to look at the cache sizes, and it's not directly compatible with the docker build action. So while we could cache the docker build directory, we'd need to write our own cache logic then (which version to restore etc...).

hebasto:
>    * [ ]  simply using the official [hub.docker.com/r/sagemath/sagemath](https://hub.docker.com/r/sagemath/sagemath) for the sage job instead of installing sagemath in our Linux image

Done in https://github.com/bitcoin-core/secp256k1/pull/1399.

real-or-random:
> I'm not entirely sure how to proceed. I think I'd like to see at least one of the two aforementioned issues addressed. If we can reduce the size of the images for now, that will probably good enough. Even if we then have issues with concurrent CI runs, that's not the end of the world. (One can just re-run manually...)

I'd say let's first try and see how much space we can gain. If we reduce the images far enough, then this may indeed be good enough, and at least prevent the issues of caused by the actual CI job having to rebuild and then running out of disk space.

.github/workflows/ci.yml
```diff
@@ -8,9 +8,209 @@ on:
...
+      # This will detect some MinGW-w64 tools but then make will need only
+      # the MSVC tools CC, AR and NM as specified below.
+      HOST: 'x86_64-w64-mingw32'
+      CC: '/opt/msvc/bin/x64/cl'
+      AR: '/opt/msvc/bin/x64/lib'
+      NM: '/opt/msvc/bin/x64/dumpbin -symbols -headers'
+      # Set non-essential options that affect the CLI messages here.
+      # (They depend on the user's taste, so we don't want to set them automatically in configure.ac.)
+      CFLAGS: '-nologo -diagnostics:caret'
+      LDFLAGS: '-Xlinker -Xlinker -Xlinker -nologo'
```
real-or-random, line 143: Using MSVC on Linux is still pretty hacky. The reason why I introduced this was simply that we didn't have Windows runners. We could consider moving it to native Windows (later, in a separate PR). 

Advantages:
 -  That would get us rid of the entire business of "installing MSVC" and it would make this Docker image much smaller. 

Disadvantages:
 -  We won't have Dockerfile anymore that people can use locally, e.g., when working on problems that appear only with MSVC.
 - We'll probably need some separate CI script for Windows / CMake. Currently, we simply reuse the Linux / autotools script, which is neat.

hebasto: > * We'll probably need some separate CI script for Windows / CMake.

In addition to https://github.com/bitcoin-core/secp256k1/blob/ce765a5b8eb93af589bc7dad2f34da0b8b51557c/.github/workflows/ci.yml#L14-L17?

real-or-random: I mean we'll at least need a way to pass config different config options from a matrix to the build systems. (Yes, it shouldn't be a lot of work. ^^)

jonasnick (commented):
Concept ACK, thanks @hebasto!

After @real-or-random explained this approach to me I was able to understand what happens without doing too much research into the github actions semantics. In general, it would be nice to have some comments that document what happens (e.g. what exactly is the purpose of `docker_wine_cache` and what it does if the image already exists), but we can add them over time in follow-up PRs.

.github/workflows/ci.yml
hebasto: Please see https://github.com/bitcoin-core/secp256k1/pull/1401.

hebasto:
Converting this PR to a draft for now.

Please review https://github.com/bitcoin-core/secp256k1/pull/1401 first.

hebasto:
Rebased.

Ready for review.

hebasto:
Rebased.

real-or-random:
I believe it's possible to shave off several minutes from the tasks that run in Docker by setting the docker build driver to `docker`, see https://github.com/docker/buildx/issues/107#issuecomment-1307929258 ff . This should avoid the long "sending tarball" phases. It can be done by setting `driver: docker`, see https://github.com/docker/setup-buildx-action#inputs . But let's experiment with this after this PR. :)

hebasto:
> I believe it's possible to shave off several minutes from the tasks that run in Docker by setting the docker build driver to `docker`, see [docker/buildx#107 (comment)](https://github.com/docker/buildx/issues/107#issuecomment-1307929258) ff . This should avoid the long "sending tarball" phases. It can be done by setting `driver: docker`, see [docker/setup-buildx-action#inputs](https://github.com/docker/setup-buildx-action#inputs) . But let's experiment with this after this PR. :)

Unfortunately, [GitHub Actions cache](https://docs.docker.com/build/cache/backends/gha/) does not work with the `docker` build driver.

jonasnick (commented):
ACK 87d35f30c0a322e9b4bc5ee1addc1d0cd463562a

PR #1399, opened 2023-08-14
ci, gha: Run "SageMath prover" job on GitHub Actions

hebasto:
Solves one item in https://github.com/bitcoin-core/secp256k1/issues/1392 partially.

Requested in https://github.com/bitcoin-core/secp256k1/pull/1398#issuecomment-1677074743.


real-or-random (approved):
utACK 8408dfdc4cfcc5cf87ac0a7d229f8a1c1d63128a

Let's also wait and see what cirrus says. (I mean, we just removed a task, so nothing should be broken but you never know...)

jonasnick (commented):
ACK 8408dfdc4cfcc5cf87ac0a7d229f8a1c1d63128a

PR #1401, opened 2023-08-15
ci, gha: Run all MSVC tests on Windows natively

hebasto:
This PR addresses the https://github.com/bitcoin-core/secp256k1/pull/1398#discussion_r1294608154 and is an alternative to https://github.com/bitcoin-core/secp256k1/pull/1398 in some part.

From https://github.com/bitcoin-core/secp256k1/pull/1398#discussion_r1294608154:
> Advantages:
> 
>    * That would get us rid of the entire business of "installing MSVC" and it would make this Docker image much smaller.
> 
> 
> Disadvantages:
> 
>    * We won't have Dockerfile anymore that people can use locally, e.g., when working on problems that appear only with MSVC.
> 
>    * We'll probably need some separate CI script for Windows / CMake. Currently, we simply reuse the Linux / autotools script, which is neat.

I do prefer this approach because it is natural.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,24 +12,41 @@ env:
 
 jobs:
   win64-native:
-    name: "x86_64: Windows, VS 2022"
+    name: ${{ matrix.configuration.job_name }}
     # See: https://github.com/actions/runner-images#available-images.
     runs-on: windows-2022
 
     strategy:
       fail-fast: false
       matrix:
-        lib_type: ['shared', 'static']
+        configuration:
+          - job_name: 'x64 (MSVC): Windows (VS 2022, shared)'
+            cmake_options: '-A x64 -DBUILD_SHARED_LIBS=ON'
+          - job_name: 'x64 (MSVC): Windows (VS 2022, static)'
+            cmake_options: '-A x64 -DBUILD_SHARED_LIBS=OFF'
+          - job_name: 'x64 (MSVC): Windows (VS 2022, int128_struct)'
+            cmake_options: '-A x64 -DSECP256K1_TEST_OVERRIDE_WIDE_MULTIPLY=int128_struct'
+          - job_name: 'x64 (MSVC): Windows (VS 2022, int128_struct with __(u)mulh)'
+            cmake_options: '-A x64 -DSECP256K1_TEST_OVERRIDE_WIDE_MULTIPLY=int128_struct'
+            cpp_flags: '/DSECP256K1_MSVC_MULH_TEST_OVERRIDE'
+          - job_name: 'x86 (MSVC): Windows (VS 2022)'
+            cmake_options: '-A Win32'
 
     steps:
       - name: Checkout
         uses: actions/checkout@v3
 
       - name: Generate buildsystem
-        run: cmake -E env CFLAGS="/WX" cmake -B build -A x64 -DSECP256K1_ENABLE_MODULE_RECOVERY=ON -DSECP256K1_BUILD_EXAMPLES=ON -DBUILD_SHARED_LIBS=${{ matrix.lib_type == 'shared' && 'ON' || 'OFF' }}
+        run: cmake -E env CFLAGS="/WX ${{ matrix.configuration.cpp_flags }}" cmake -B build -DSECP256K1_ENABLE_MODULE_RECOVERY=ON -DSECP256K1_BUILD_EXAMPLES=ON ${{ matrix.configuration.cmake_options }}
 
       - name: Build
         run: cmake --build build --config RelWithDebInfo -- /p:UseMultiToolTask=true /maxCpuCount
+
+      - name: Binaries info
+        # Use the bash shell included with Git for Windows.
+        shell: bash
+        run: |
+          cd build/src/RelWithDebInfo && file *tests.exe bench*.exe libsecp256k1-*.dll || true
 
       - name: Check
         run: |
@@ -37,6 +54,22 @@ jobs:
           build\src\RelWithDebInfo\bench_ecmult.exe
           build\src\RelWithDebInfo\bench_internal.exe
           build\src\RelWithDebInfo\bench.exe
+
+  win64-native-headers:
+    name: "x64 (MSVC): C++ (public headers)"
+    # See: https://github.com/actions/runner-images#available-images.
+    runs-on: windows-2022
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Add cl.exe to PATH
+        uses: ilammy/msvc-dev-cmd@v1
+
+      - name: C++ (public headers)
+        run: |
+          cl.exe -c -WX -TP include/*.h
 
   sage:
     name: "SageMath prover"
--- a/ci/cirrus.sh
+++ b/ci/cirrus.sh
@@ -31,16 +31,6 @@ print_environment() {
 }
 print_environment
 
-# Start persistent wineserver if necessary.
-# This speeds up jobs with many invocations of wine (e.g., ./configure with MSVC) tremendously.
-case "$WRAPPER_CMD" in
-    *wine*)
-        # Make sure to shutdown wineserver whenever we exit.
-        trap "wineserver -k || true" EXIT INT HUP
-        wineserver -p
-        ;;
-esac
-
 env >> test_env.log
 
 # If gcc is requested, assert that it's in fact gcc (and not some symlinked Apple clang).
--- a/ci/linux-debian.Dockerfile
+++ b/ci/linux-debian.Dockerfile
@@ -21,7 +21,8 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
         gcc-aarch64-linux-gnu libc6-dev-arm64-cross libc6-dbg:arm64 \
         gcc-powerpc64le-linux-gnu libc6-dev-ppc64el-cross libc6-dbg:ppc64el \
         gcc-mingw-w64-x86-64-win32 wine64 wine \
-        gcc-mingw-w64-i686-win32 wine32
+        gcc-mingw-w64-i686-win32 wine32 \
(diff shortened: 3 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -12,32 +12,53 @@ env:
...
       - name: Check
         run: |
           ctest -C RelWithDebInfo --test-dir build -j ([int]$env:NUMBER_OF_PROCESSORS + 1)
           build\src\RelWithDebInfo\bench_ecmult.exe
           build\src\RelWithDebInfo\bench_internal.exe
           build\src\RelWithDebInfo\bench.exe
 
+      - uses: ilammy/msvc-dev-cmd@v1
+      - name: C++ (public headers)
+        run: cl.exe -c -WX -TP include/*.h
```
real-or-random, line 60: Is the `ilammy/msvc-dev-cmd@v1` required only for the C++ compiler? To be honest, we could drop that part then. But do you really need this? Can't we use the already installed MSVC? (I mean, we use it above to compile the library, so it must be somewhere on these machines, right?)

ci/linux-debian.Dockerfile
```diff
@@ -21,7 +21,8 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
         gcc-aarch64-linux-gnu libc6-dev-arm64-cross libc6-dbg:arm64 \
         gcc-powerpc64le-linux-gnu libc6-dev-ppc64el-cross libc6-dbg:ppc64el \
         gcc-mingw-w64-x86-64-win32 wine64 wine \
-        gcc-mingw-w64-i686-win32 wine32
+        gcc-mingw-w64-i686-win32 wine32 \
+        python3
```
real-or-random, line 25: Where do we need python?

.github/workflows/ci.yml
hebasto: We can use `C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.35.32215/bin/HostX64/x64/cl.exe`, but it needs to be updated every time GH Windows image is updated due to the version part in the path.

ci/linux-debian.Dockerfile
hebasto:
```
make  precompute_ecmult_gen
make  precompute_ecmult
mkdir -p src/wycheproof
python3 ./tools/tests_wycheproof_generate.py ./src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.json > src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.h
/bin/bash: line 1: python3: command not found
```

real-or-random: Oh indeed, I forgot.

.github/workflows/ci.yml
real-or-random: Couldn't we  just invoke it with a glob like `C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/*/bin/HostX64/x64/cl.exe`? (or whatever the windows equivalent of this is, or use bash).

real-or-random:
> * We won't have Dockerfile anymore that people can use locally, e.g., when working on problems that appear only with MSVC.

I don't think that's a big deal. This feature is rarely needed, and if it's needed, people can use the one here: https://github.com/mstorsjo/msvc-wine/blob/master/Dockerfile 

So yes, Concept ACK

.github/workflows/ci.yml
hebasto: > Couldn't we just invoke it with a glob...

I guess not:

```
  dir "C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/"
  shell: C:\Program Files\PowerShell\7\pwsh.EXE -command ". '{0}'"

    Directory: C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----            8/4/2023 10:41 PM                14.16.27023
d----            8/4/2023 10:39 PM                14.29.30133
d----            8/4/2023 11:07 PM                14.35.32215
d----            8/4/2023 10:28 PM                14.36.32532
```

hebasto: > To be honest, we could drop that part then.

It has been dropped.

real-or-random: Okay, after some digging, I'm actually convinced that https://github.com/ilammy/msvc-dev-cmd is the commonly accepted way of doing this. I also found https://github.com/actions/runner-images/issues/5986#issuecomment-1201012941, but not sure if it's better.

So I'm leaning towards bringing https://github.com/ilammy/msvc-dev-cmd back. Then it will be even cleaner to do this step before the regular CMake invocations, too? And If this ever breaks in the future, it should still be easy to get rid of it. Sorry for the back and forth.

real-or-random:
We'll need to bring the `wineboot` back now (https://github.com/bitcoin-core/secp256k1/commit/c7db4942b34acd2a34e6249112f6c1db6cf5681d) to resolve the failures.

We could add it to the CI script (instead of the Docker image). It runs very fast, but it will add 1.3 GB to the image, so having it in the script may be the better trade-off. (wine prefixes seem to be pretty big these days. I heard this is because wine will copy a lot of DLLs there now, instead of symlinking them... This improves the emulation for anti-cheat detection for games... ^^)

.github/workflows/ci.yml
hebasto: > It seems to have been added already.

It works!

ci/cirrus.sh
```diff
@@ -31,12 +31,20 @@ print_environment() {
 }
 print_environment
 
-# Start persistent wineserver if necessary.
-# This speeds up jobs with many invocations of wine (e.g., ./configure with MSVC) tremendously.
 case "$WRAPPER_CMD" in
     *wine*)
+        # Initialize the wine environment.
+        wine64 wineboot --init
```
real-or-random, line 37:
```suggestion
        wineboot --init
```
That's enough, and this should fix the failure.

nit: I wonder if it makes sense to do this in the yaml file. People tend to run this script locally on their machine, and it's probably a tiny bit better to let it only do "non-intrusive" stuff.  `wineboot --init` shouldn't do any harm, but it touches the user's wine prefix.

real-or-random (commented):
Pretty happy with this now, modulo the nits

.github/workflows/ci.yml
hebasto: > ... it seems that CMake works here more or less by accident.

The Windows image has [CMake installed](https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md#tools) separately from Visual Studio tools, which is nice.

In turn, CMake itself is smart enough to find paths to all tools it requires.

Only direct invocation of the `cl.exe` compiler requires having it in the `%PATH%`.

That's why I do prefer keeping these lines of code together.

real-or-random: Ok, makes sense. Keeping them together also self-documents the fact that the external action is only required for that single line of code.

ci/cirrus.sh
```diff
@@ -31,13 +31,10 @@ print_environment() {
...
 
-# Start persistent wineserver if necessary.
-# This speeds up jobs with many invocations of wine (e.g., ./configure with MSVC) tremendously.
 case "$WRAPPER_CMD" in
     *wine*)
         # Make sure to shutdown wineserver whenever we exit.
         trap "wineserver -k || true" EXIT INT HUP
-        wineserver -p
         ;;
 esac
```
real-or-random, line 39: Oh, I didn't mean to remove this code. Starting the wineserver is still useful, because it speeds up the tests.  I just meant that we can drop these lines:
```
        # Wait until the wineserver process has
        # exited, to avoid corrupting the wine prefix.
        while (ps -A | grep wineserver) > /dev/null; do sleep 1; done
```

real-or-random: Sorry for constantly offering tiny suggestions that break stuff... -.- 

What happens here is that `wineserver -p` fails because there's a server running already (from the `wineboot` still, but it's not a persistent one...) While the "wait until the wineserver process has exited" code was still there, this didn't happen.

I took a step back and did some testing on my local machine:
 - `wineserver -p` is at most useful with MSVC builds. With mingw builds, there won't be many wine invocations. And even with MSVC builds, it's questionable: The server stays around for 3 seconds and waits for another wine process to appear before it exits, so this should be necessary only when things are really slow. This has indeed speed up builds for us in the past, but maybe this was because some other part didn't work... No idea. 
 - `wineboot --init` doesn't seem to be required. The first wine invocation should init stuff automatically (and this is how I know wine...)

So I suggest to just remove all of this and pushed it as a fixup here: https://github.com/real-or-random/secp256k1/tree/230815-gha-msvc . CI still in my repo running, but I hope that it just works now...

hebasto: https://github.com/real-or-random/secp256k1/commit/76ecd47c42e320cab53ae30dce18710ad3a45ff1 has been cherry-picked and squashed with the last commit.

hebasto:
I've just realised that "C++ (public headers)" is a step in every MSVC job. Shouldn't it be a separated job?

real-or-random:
> I've just realized that "C++ (public headers)" is a step in every matrix-generated MSVC job. Shouldn't it be a separated job?

Yes, I had the same thought earlier. I think it makes sense to move it to a separate job if you're willing to make that change.

hebasto:
> > I've just realized that "C++ (public headers)" is a step in every matrix-generated MSVC job. Shouldn't it be a separated job?
> 
> Yes, I had the same thought earlier. I think it makes sense to move it to a separate job if you're willing to make that change.

Fixed.

real-or-random (commented):
ACK d78bec7001fe6f5ed8d5b215bf61e7b74e3369ca

PR #1405, opened 2023-08-20
ci: Drop no longer needed workaround

hebasto:
The https://sourceware.org/bugzilla/show_bug.cgi?id=27008 bug has been resolved since libc 2.33.

Debian Bookworm has [libc](https://packages.debian.org/bookworm/libc6) 2.36.

I've separated this change from moving CI tasks to GitHub Actions intentionally.


real-or-random (commented):
ACK ef9fe959deb638228bca8f1068f078e87e271b02

PR #1408, opened 2023-08-21
ci, doc: Document required actions

hebasto:
Suggested in https://github.com/bitcoin/bitcoin/pull/28173#pullrequestreview-1587098202:

> I wonder if it makes sense to document the required permissions inside the file?


real-or-random:
Hm, isn't this clear from reading the files? I think that means that adding a comment is mostly an additional maintenance burden because we need to keep it up-to-date (and no tool complains if we forget). 

Also, the required permissions are not canonical. You could also allow any action, or pin the actions to specific commit ids. All of these options would work.

PR #1409, opened 2023-08-22
ci: Move remained task from Cirrus to GitHub Actions

hebasto:
This PR concludes the migration from Cirrus CI to GitHub Actions.


.github/workflows/ci.yml
```diff
@@ -373,6 +373,138 @@ jobs:
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Set environment variables
+        run: |
+          for var in "$(jq -s 'add' <<< '${{ toJSON(matrix.configuration.env_vars) }} ${{ toJSON(matrix.asm.env_vars) }} ${{ toJSON(matrix.cc.env_vars) }}' | jq -r 'to_entries | .[] | "\(.key)=\(.value)"')"; do
+            echo "$var" >> "$GITHUB_ENV"
+          done
```
real-or-random, line 420: A bit ugly, hm... There's no better way to merge this, I guess.

I think it makes sense to split this job into two jobs "Valgrind (memcheck)" and "UBSan, ASan, LSan". Then for each of these jobs, we can expand the remaining 2x2 combinations, i.e., list the 4 resulting combinations explicitly. This is probably a good idea anyway, it will also make it a bit easier to add variety to the jobs.

What do you think?

hebasto: > I think it makes sense to split this job into two jobs "Valgrind (memcheck)" and "UBSan, ASan, LSan".

I was thinking about the same. But it won't help.

Only one `env` property is allowed job-wide and step-wide. Job-wide `env` is already busy with global variables. Therefore, even 2 sets won't fit one step-wide `env`.

hebasto: > list the 4 resulting combinations explicitly

... but that is completely feasible.

hebasto: > > list the 4 resulting combinations explicitly
> 
> ... but that is completely feasible.

Done.

.github/workflows/ci.yml
```diff
@@ -373,6 +373,138 @@ jobs:
...
+        uses: actions/checkout@v3
+
+      - name: CI script
+        env: ${{ matrix.configuration.env_vars }}
+        uses: ./.github/actions/run-in-docker-action
+        with:
+          dockerfile: ./ci/linux-debian.Dockerfile
+          tag: linux-debian-image
+          command: >
+            git config --global --add safe.directory ${{ github.workspace }} &&
```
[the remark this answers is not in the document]
real-or-random: Ok, sure, I remember now. I still think this should not be in `ci.sh`, but yeah, we could consider moving it to `run-in-docker-action` (in a separate PR probably).

.github/workflows/ci.yml
```diff
@@ -548,6 +722,70 @@ jobs:
...
+        uses: actions/checkout@v3
+
+      - name: CI script
+        uses: ./.github/actions/run-in-docker-action
+        with:
+          dockerfile: ./ci/linux-debian.Dockerfile
+          tag: linux-debian-image
+          command: >
+            g++ -Werror include/*.h &&
+            clang -Werror -x c++-header include/*.h
```
real-or-random, line 787: nit: it's probably more convenient to run `clang` even if `g++` fails
```suggestion
            g++ -Werror include/*.h
            clang -Werror -x c++-header include/*.h
```
(I think that linebreak should work. Or do we need a `;` here?)

hebasto: Thanks! Reworked.

.github/workflows/ci.yml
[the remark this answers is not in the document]
hebasto: Done in https://github.com/bitcoin-core/secp256k1/pull/1411.

PR #1411, opened 2023-08-23
ci: Make repetitive command the default one

hebasto:
This PR addresses the https://github.com/bitcoin-core/secp256k1/pull/1409#discussion_r1301767281:
> couldn't we add this to `run-in-docker-action` to avoid duplication?

### DIFF
--- a/.github/actions/run-in-docker-action/action.yml
+++ b/.github/actions/run-in-docker-action/action.yml
@@ -9,7 +9,8 @@ inputs:
     required: true
   command:
     description: 'A command to run in a container'
-    required: true
+    required: false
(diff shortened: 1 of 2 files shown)


real-or-random:
Thanks. I think it's even cleaner to make `./ci/ci.sh` the default, and *always* prepend the `git config`?

hebasto:
> I think it's even cleaner to make `./ci/ci.sh` the default, and _always_ prepend the `git config`?

Thank you! Implemented.

real-or-random (approved):
ACK cce045630407e662c90279abb2b0e71161f24e74

PR #1412, opened 2023-08-26
ci: Switch macOS from Ventura to Monterey and add Valgrind

hebasto:
This PR switches the macOS native job from Ventura to Monterey, which allows to support Valgrind.

Both runners--`macos-12` and `macos-13`--have the same clang compilers installed:
- https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md
- https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md

But Valgrind works fine on macOS Monterey, but not on Ventura.

See: https://github.com/bitcoin-core/secp256k1/issues/1392#issuecomment-1693685610.

The Homebrew's Valgrind package is cached once it has been built (as it was before https://github.com/bitcoin-core/secp256k1/pull/1152). Therefore, the `actions/cache@*` action is needed to be added to the list of the allowed actions.

https://github.com/bitcoin-core/secp256k1/pull/1412#issuecomment-1695716350:
> By the way, this solves #1151.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -575,27 +575,28 @@ jobs:
         if: ${{ always() }}
 
   macos-native:
-    name: "x86_64: macOS Ventura"
+    name: "x86_64: macOS Monterey"
     # See: https://github.com/actions/runner-images#available-images.
-    runs-on: macos-13 # Use M1 once available https://github.com/github/roadmap/issues/528
+    runs-on: macos-12 # Use M1 once available https://github.com/github/roadmap/issues/528
 
     env:
-      ASM: 'no'
-      WITH_VALGRIND: 'no'
-      CTIMETESTS: 'no'
       CC: 'clang'
+      HOMEBREW_NO_AUTO_UPDATE: 1
+      HOMEBREW_NO_INSTALL_CLEANUP: 1
 
     strategy:
       fail-fast: false
       matrix:
         env_vars:
           - { WIDEMUL: 'int64',  RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes' }
-          - { WIDEMUL: 'int64',  RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes', CC: 'gcc' }
           - { WIDEMUL: 'int128_struct', ECMULTGENPRECISION: 2, ECMULTWINDOW: 4 }
           - { WIDEMUL: 'int128',                  ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes' }
-          - { WIDEMUL: 'int128', RECOVERY: 'yes',              SCHNORRSIG: 'yes' }
+          - { WIDEMUL: 'int128', RECOVERY: 'yes' }
+          - { WIDEMUL: 'int128', RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes' }
           - { WIDEMUL: 'int128', RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes', CC: 'gcc' }
-          - { WIDEMUL: 'int128', RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes', CPPFLAGS: '-DVERIFY' }
+          - { WIDEMUL: 'int128', RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes',            WRAPPER_CMD: 'valgrind --error-exitcode=42', SECP256K1_TEST_ITERS: 2 }
+          - { WIDEMUL: 'int128', RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes', CC: 'gcc', WRAPPER_CMD: 'valgrind --error-exitcode=42', SECP256K1_TEST_ITERS: 2 }
+          - { WIDEMUL: 'int128', RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes', CPPFLAGS: '-DVERIFY', CTIMETESTS: 'no' }
           - BUILD: 'distcheck'
 
     steps:
@@ -603,12 +604,12 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install Homebrew packages
-        env:
-          HOMEBREW_NO_AUTO_UPDATE: 1
-          HOMEBREW_NO_INSTALL_CLEANUP: 1
         run: |
           brew install automake libtool gcc
           ln -s $(brew --prefix gcc)/bin/gcc-?? /usr/local/bin/gcc
+
+      - name: Install and cache Valgrind
+        uses: ./.github/actions/install-homebrew-valgrind
 
       - name: CI script
         env: ${{ matrix.env_vars }}
--- a/.github/actions/install-homebrew-valgrind/action.yml
+++ b/.github/actions/install-homebrew-valgrind/action.yml
@@ -1,0 +1,33 @@
+name: "Install Valgrind"
+description: "Install Homebrew's Valgrind package and cache it."
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        brew tap LouisBrunner/valgrind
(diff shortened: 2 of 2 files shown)


real-or-random (commented):
nit: Perhaps rename the PR + commit to include Valgrind. We mostly add Valgrind here, which is useful in itself. Adding the ctime tests is just a side benefit of Valgrind, although an important one).

What I had in mind was initially to run CI for macOS only on Monterey, but yeah, maybe it's not a bad idea to test on multiple Apple clangs...

hebasto:
> What I had in mind was initially to run CI for macOS only on Monterey, but yeah, maybe it's not a bad idea to test on multiple Apple clangs...

I guess, we are keeping the other macOS job for upcoming switching to arm64:https://github.com/bitcoin-core/secp256k1/blob/ea26b71c3a3e1a5c39a4fbdaf2f0b787c5139f56/.github/workflows/ci.yml#L580

Right?

hebasto:
> nit: Perhaps rename the PR + commit to include Valgrind. We mostly add Valgrind here, which is useful in itself. Adding the ctime tests is just a side benefit of Valgrind, although an important one).

Done.

.github/workflows/ci.yml
```diff
@@ -632,6 +632,61 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        env_vars:
+          - { WIDEMUL: 'int64',  RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes' }
+          - { WIDEMUL: 'int64',  RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes', CC: 'gcc' }
+          - { WIDEMUL: 'int128_struct', ECMULTGENPRECISION: 2, ECMULTWINDOW: 4 }
+          - { WIDEMUL: 'int128',                  ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes' }
+          - { WIDEMUL: 'int128', RECOVERY: 'yes',              SCHNORRSIG: 'yes' }
+          - { WIDEMUL: 'int128', RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes', CC: 'gcc' }
```
real-or-random, line 654: I suggest this merged matrix, which also includes real Valgrind jobs (see the main comment in my review for details): 
```suggestion
        env_vars:
          - { WIDEMUL: 'int64',  RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes' }
          - { WIDEMUL: 'int128_struct', ECMULTGENPRECISION: 2, ECMULTWINDOW: 4 }
          - { WIDEMUL: 'int128',                  ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes' }
          - { WIDEMUL: 'int128', RECOVERY: 'yes' }
          - { WIDEMUL: 'int128', RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes' }
          - { WIDEMUL: 'int128', RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes', CC: 'gcc' }
          - { WIDEMUL: 'int128', RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes',            WRAPPER_CMD: 'valgrind --exit-code=42' }
          - { WIDEMUL: 'int128', RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes', CC: 'gcc', WRAPPER_CMD: 'valgrind --exit-code=42' }
          - { WIDEMUL: 'int128', RECOVERY: 'yes', ECDH: 'yes', SCHNORRSIG: 'yes', ELLSWIFT: 'yes', CPPFLAGS: '-DVERIFY' }
          - BUILD: 'distcheck'
```

.github/workflows/ci.yml
```diff
@@ -632,6 +632,61 @@ jobs:
         run: env
         if: ${{ always() }}
 
+  macos-monterey:
+    name: "x86_64: macOS Monterey, ctime_tests"
+    # See: https://github.com/actions/runner-images#available-images.
+    runs-on: macos-12
```
real-or-random, line 638: If we remove the Ventura jobs: 
```suggestion
  macos-native:
    name: "x86_64: macOS Monterey"
    # See: https://github.com/actions/runner-images#available-images.
    runs-on: macos-12 # Use M1 once available https://github.com/github/roadmap/issues/528
```

real-or-random (commented):
> What I had in mind was initially to run CI for macOS only on Monterey, but yeah, maybe it's not a bad idea to test on multiple Apple clangs...

Okay, I was wrong. The runners have the same Apple clang. And they also have the same gcc, so this is not a valid argument. 


> I guess, we are keeping the other macOS job for upcoming switching to arm64:

Hm, not sure then. I think we should switch to arm64 once it's available. But when we do this, we'll probably anyway want to rearrange the matrix again (e.g., this will depend on the other ARM jobs we have etc.)

So I suggest keeping only Monterey for now, and moving the "# Use M1 once available https://github.com/github/roadmap/issues/528" comment there.

--------

> nit: Perhaps rename the PR + commit to include Valgrind. We mostly add Valgrind here, which is useful in itself. Adding the ctime tests is just a side benefit of Valgrind, although an important one).

Okay, sorry, nevermind. I was wrong here, too... -.- When I wrote this, I assumed this PR adds also Valgrind jobs. But it's a good idea to do this, and my suggested matrix has some Valgrind jobs...

.github/workflows/ci.yml
real-or-random: By the way, I don't like this style of having the same job with and without valgrind. Perhaps we should reintroduce a `RUN_ON_VALGRIND` variable (removed in fcfcb97e74b55a107290d44c81c049d6168e954f), but that should be a separate PR, and I don't think it should hold up this PR here.

hebasto:
Updated 97d0cd422cf26d789b5652689585c6b06380fff6 -> 7247231ff1bc12127acbcdef51e6c482f7d057b9 ([pr1412.02](https://github.com/hebasto/secp256k1/commits/pr1412.02) -> [pr1412.03](https://github.com/hebasto/secp256k1/commits/pr1412.03), [diff](https://github.com/hebasto/secp256k1/compare/pr1412.02..pr1412.03)):

- addressed @real-or-random's comments
- the PR description and title have been updated

hebasto:
Updated 97d0cd422cf26d789b5652689585c6b06380fff6 -> 2ad93e81f845fa959514b8fb0cfd5df9f69d40da ([pr1412.02](https://github.com/hebasto/secp256k1/commits/pr1412.02) -> [pr1412.04](https://github.com/hebasto/secp256k1/commits/pr1412.04), [diff](https://github.com/hebasto/secp256k1/compare/pr1412.02..pr1412.04)):

- addressed @real-or-random's comments
- the PR description and title have been updated

.github/workflows/ci.yml
hebasto: > I suggest this merged matrix, which also includes real Valgrind jobs (see the main comment in my review for details):

Thanks! Implemented in https://github.com/bitcoin-core/secp256k1/pull/1412#issuecomment-1695937918.

.github/workflows/ci.yml
hebasto: Thanks! [Done](https://github.com/bitcoin-core/secp256k1/pull/1412#issuecomment-1695937918).

.github/workflows/ci.yml
real-or-random: Ah, I think we should set `SECP256K1_TEST_ITERS=2` like for the other Valgrind jobs....

Perhaps it's really cleaner to have a `RUN_ON_VALGRIND`, and the script then reduces the number of iterations automatically by a factor of 32 or 16...  

> but that should be a separate PR, and I don't think it should hold up this PR here.

I stand by this, so I'd say do whatever you prefer.

hebasto: > Ah, I think we should set `SECP256K1_TEST_ITERS=2` like for the other Valgrind jobs....

Sure. Done.

hebasto:
For testing purposes, I set this pull request branch as the default one in my fork and manually triggered the new workflow: https://github.com/hebasto/secp256k1/actions/runs/6010693824.

real-or-random (approved):
ACK c223d7e33d50b1da0b3ba617c83534185428b3d5 I tested that a cttest failure makes CI fail: https://github.com/real-or-random/secp256k1/actions/runs/6010365844

PR #1413, opened 2023-08-28
ci: Add `release` job

hebasto:
This PR introduces a new "Release" job that conducts sanity checks as defined in [`doc/release-process.md`](https://github.com/bitcoin-core/secp256k1/blob/master/doc/release-process.md#sanity-checks).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -780,3 +780,27 @@ jobs:
         run: |
           cd sage
           sage prove_group_implementations.sage
+
+  release:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - run: ./autogen.sh && ./configure --enable-dev-mode && make distcheck
+
+      - name: Check installation with Autotools
+        env:
+          CI_INSTALL: ${{ runner.temp }}/${{ github.run_id }}${{ github.action }}
+        run: |
+          ./autogen.sh && ./configure --prefix=${{ env.CI_INSTALL }} && make clean && make install && ls -RlAh ${{ env.CI_INSTALL }}
+          gcc -o ecdsa examples/ecdsa.c $(PKG_CONFIG_PATH=${{ env.CI_INSTALL }}/lib/pkgconfig pkg-config --cflags --libs libsecp256k1) -Wl,-rpath,"${{ env.CI_INSTALL }}/lib" && ./ecdsa
+
+      - name: Check installation with CMake
+        env:
+          CI_BUILD: ${{ runner.temp }}/${{ github.run_id }}${{ github.action }}/build
+          CI_INSTALL: ${{ runner.temp }}/${{ github.run_id }}${{ github.action }}/install
+        run: |
+          cmake -B ${{ env.CI_BUILD }} -DCMAKE_INSTALL_PREFIX=${{ env.CI_INSTALL }} && cmake --build ${{ env.CI_BUILD }} --target install && ls -RlAh ${{ env.CI_INSTALL }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,42 @@
+# See: https://github.com/bitcoin-core/secp256k1/blob/master/doc/release-process.md.
```
real-or-random, line 1:
```suggestion
# See: https://github.com/bitcoin-core/secp256k1/blob/master/doc/release-process.md
```
nit: link is broken otherwise

real-or-random:
This is neat! 

We could also print the file listings for both installation methods (something like `cd ${{ env.CI_INSTALL }} && find`) and/or create tarball artifacts of the installation directories.

hebasto:
> We could also print the file listings for both installation methods (something like `cd ${{ env.CI_INSTALL }} && find`)

Instead of the current `ls -l ${{ env.CI_INSTALL }}/include ${{ env.CI_INSTALL }}/lib`, right?

> ... and/or create tarball artifacts of the installation directories.

What for?

real-or-random:
> > We could also print the file listings for both installation methods (something like `cd ${{ env.CI_INSTALL }} && find`)
> 
> Instead of the current `ls -l ${{ env.CI_INSTALL }}/include ${{ env.CI_INSTALL }}/lib`, right?
> 

Yep, it's good to see if too many files are installed.

> > ... and/or create tarball artifacts of the installation directories.
> 
> What for?

This would allow for an even closer inspection of the installation, including file sizes etc. But yeah, it's probably overkill. If people want to see this, they should check locally. We could also use `ls -RlAh` instead of using `find`, and this prints sizes and has also a nice format.

hebasto:
Updated 94dd59a525e22221190a7a18e5d96f509666daf8 -> f9ba3f200a5e67eb1ba454ee478c2bc023d2c2a2 ([pr1413.03](https://github.com/hebasto/secp256k1/commits/pr1413.03) -> [pr1413.05](https://github.com/hebasto/secp256k1/commits/pr1413.05), [diff](https://github.com/hebasto/secp256k1/compare/pr1413.03..pr1413.05)).

The following @real-or-random's comments have been addressed:
- https://github.com/bitcoin-core/secp256k1/pull/1413#discussion_r1308458759
- https://github.com/bitcoin-core/secp256k1/pull/1413#issuecomment-1697097054

.github/workflows/release.yml
hebasto: Thanks! [Fixed](https://github.com/bitcoin-core/secp256k1/pull/1413#issuecomment-1697108649).

hebasto:
For testing purposes, I set this pull request branch as the default one in my fork and manually triggered the new workflow: https://github.com/hebasto/secp256k1/actions/runs/6010823279.

real-or-random (approved):
ACK 0ae58afcedfe24babd7d58b69b3b69f5f43aa79b

real-or-random:
> If your concerns are about running the workflow conditionally only on release PRs, than that condition might be dropped from the script.

I think that was his concern. I agree, we could just run these checks every time. They should always succeed.

hebasto:
> But all of the tests currently included in this workflow appear to me to be things we want always. So why not add them to normal CI instead?

Done.

.github/workflows/ci.yml
```diff
@@ -774,9 +774,38 @@ jobs:
...
 
       - name: CI script
         run: |
           cd sage
           sage prove_group_implementations.sage
+
+  release:
+    runs-on: ubuntu-latest
+    # The `startsWith` and `contains` functions are not case sensitive.
+    if: >
```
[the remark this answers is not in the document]
hebasto: Thanks! Fixed.

hebasto:
The PR description has been updated.

real-or-random (approved):
ACK 8659a01714c1b4fcd349ee1a7d733f6934c5d184

PR #1414, opened 2023-08-30
ci/gha: Add ARM64 QEMU jobs for clang and clang-snapshot

real-or-random:
Solves one item in https://github.com/bitcoin-core/secp256k1/issues/1392.

This PR also has a few tweaks to the Dockerfile, see individual commits.

---

I'll follow up soon with a PR for ARM64/gcc. This will rely on Cirrus CI.

### DIFF
--- a/ci/linux-debian.Dockerfile
+++ b/ci/linux-debian.Dockerfile
@@ -46,17 +46,21 @@ RUN mkdir gcc && cd gcc && \
     cd ../.. && rm -rf gcc && \
     ln -s /opt/gcc-snapshot/bin/gcc /usr/bin/gcc-snapshot
 
-# Install clang snapshot
-RUN wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc && \
+# Install clang snapshot, see https://apt.llvm.org/
+RUN \
+    # Setup GPG keys of LLVM repository
+    apt-get update && apt-get install --no-install-recommends -y wget && \
+    wget -qO- https://apt.llvm.org/llvm-snapshot.gpg.key | tee /etc/apt/trusted.gpg.d/apt.llvm.org.asc && \
     # Add repository for this Debian release
     . /etc/os-release && echo "deb http://apt.llvm.org/${VERSION_CODENAME} llvm-toolchain-${VERSION_CODENAME} main" >> /etc/apt/sources.list && \
-    # Install clang snapshot
-    apt-get update && apt-get install --no-install-recommends -y clang && \
-    # Remove just the "clang" symlink again
-    apt-get remove -y clang && \
-    # We should have exactly two clang versions now
-    ls /usr/bin/clang* && \
-    [[    $(ls /usr/bin/clang-?? | sort | wc -l) -eq "2" ]] && \
-    # Create symlinks for them
-    ln -s $(ls /usr/bin/clang-?? | sort | tail -1) /usr/bin/clang-snapshot && \
-    ln -s $(ls /usr/bin/clang-?? | sort | head -1) /usr/bin/clang
+    apt-get update && \
+    # Determine the version number of the LLVM development branch
+    LLVM_VERSION=$(apt-cache search --names-only '^clang-[0-9]+$' | sort -V | tail -1 | cut -f1 -d" " | cut -f2 -d"-" ) && \
+    # Install
+    apt-get install --no-install-recommends -y "clang-${LLVM_VERSION}" && \
+    # Create symlink
+    ln -s "/usr/bin/clang-${LLVM_VERSION}" /usr/bin/clang-snapshot && \
+    # Clean up
+    apt-get autoremove -y wget && \
+    apt-get clean
+
--- a/ci/linux-debian.Dockerfile
+++ b/ci/linux-debian.Dockerfile
@@ -2,6 +2,17 @@ FROM debian:stable-slim
 
 SHELL ["/bin/bash", "-c"]
 
+WORKDIR /root
+
+# A too high maximum number of file descriptors (with the default value
+# inherited from the docker host) can cause issues with some of our tools:
+#  - sanitizers hanging: https://github.com/google/sanitizers/issues/1662 
+#  - valgrind crashing: https://stackoverflow.com/a/75293014
+# This is not be a problem on our CI hosts, but developers who run the image
+# on their machines may run into this (e.g., on Arch Linux), so warn them.
+# (Note that .bashrc is only executed in interactive bash shells.)
+RUN echo 'if [[ $(ulimit -n) -gt 200000 ]]; then echo "WARNING: Very high value reported by \"ulimit -n\". Consider passing \"--ulimit nofile=32768\" to \"docker run\"."; fi' >> /root/.bashrc
+
 RUN dpkg --add-architecture i386 && \
     dpkg --add-architecture s390x && \
     dpkg --add-architecture armhf && \
@@ -23,8 +34,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
         gcc-mingw-w64-x86-64-win32 wine64 wine \
         gcc-mingw-w64-i686-win32 wine32 \
         python3
-
-WORKDIR /root
 
 # Build and install gcc snapshot
 ARG GCC_SNAPSHOT_MAJOR=14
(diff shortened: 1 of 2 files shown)


hebasto:
Concept ACK.

hebasto:
https://github.com/bitcoin-core/secp256k1/actions/runs/6027879043/job/16355370189:
```
Run cat tests.log || true
MemorySanitizer: CHECK failed: sanitizer_allocator_primary64.h:133 "((kSpaceBeg)) == ((address_range.Init(TotalSpaceSize, PrimaryAllocatorName, kSpaceBeg)))" (0xe00000000000, 0xfffffffffffffff4) (tid=3244)
    <empty stack>

FAIL tests (exit status: 1)
```

ci/linux-debian.Dockerfile
```diff
@@ -37,26 +47,31 @@ RUN mkdir gcc && cd gcc && \
...
-    # Create symlinks for them
-    ln -s $(ls /usr/bin/clang-?? | sort | tail -1) /usr/bin/clang-snapshot && \
-    ln -s $(ls /usr/bin/clang-?? | sort | head -1) /usr/bin/clang
+    apt-get update && \
+    # Determine the version number of the LLVM development branch
+    LLVM_VERSION=$(apt-cache search --names-only '^clang-[0-9]+$' | sort | tail -1 | cut -f1 -d" " | cut -f2 -d"-" ) && \
+    # Install packages
+    apt-get install --no-install-recommends -y "clang-${LLVM_VERSION}" "libclang-rt-${LLVM_VERSION}-dev:arm64" && \
+    # Assert that we have exactly two clang versions now
+    ls /usr/bin/clang* && [[ $(ls /usr/bin/clang-?? | sort | wc -l) -eq "2" ]] && \
```
hebasto, line 71: Why this code is needed now?

hebasto:
https://github.com/bitcoin-core/secp256k1/actions/runs/6027879043/job/16355369997:
>     /usr/bin/aarch64-linux-gnu-ld: cannot find /usr/lib/llvm-14/lib/clang/14.0.6/lib/linux/libclang_rt.msan-aarch64.a: No such file or directory

Should be provided by the [`libclang-rt-14-dev:arm64`](https://packages.debian.org/bookworm/amd64/libclang-rt-14-dev/filelist) package.

ci/linux-debian.Dockerfile
real-or-random: Okay, it's just an assertion. So strictly speaking, it was never needed. My thinking is that it still doesn't hurt to check for this, and the code was already there.

hebasto: In the current master branch, it is useful because we are processing two files using `tail -1` and `head -1`. That is not the case anymore.

hebasto:
> I think the second option doesn't sound too bad. I could rework this PR to drop MSan. This would resolve all the "ugly" issues mentioned above. And if it turns out that free minutes on Cirrus are not enough, we can still easily move to the Bitcoin Core runner. What do you think?

Looks good to me.

bitcoinfinancier (approved):

real-or-random:
> > I think the second option doesn't sound too bad. I could rework this PR to drop MSan. This would resolve all the "ugly" issues mentioned above. And if it turns out that free minutes on Cirrus are not enough, we can still easily move to the Bitcoin Core runner. What do you think?
> 
> Looks good to me.

Ok, did this. Ready for review.

For the follow-up PR that uses Cirrus: I forgot that the free budged is shared between all repos in the bitcoin-core org. I guess it will be less hassle to use the persistent runner. (For example, the HWI repo hasn't converted yet, so we'd compete with them for the free runners etc.) So I'll probably do this.

ci/linux-debian.Dockerfile
real-or-random: Yeah, ok, dropped.

ci/linux-debian.Dockerfile
```diff
@@ -37,26 +47,29 @@ RUN mkdir gcc && cd gcc && \
...
-    apt-get remove -y clang && \
-    # We should have exactly two clang versions now
-    ls /usr/bin/clang* && \
-    [[    $(ls /usr/bin/clang-?? | sort | wc -l) -eq "2" ]] && \
-    # Create symlinks for them
-    ln -s $(ls /usr/bin/clang-?? | sort | tail -1) /usr/bin/clang-snapshot && \
-    ln -s $(ls /usr/bin/clang-?? | sort | head -1) /usr/bin/clang
+    apt-get update && \
+    # Determine the version number of the LLVM development branch
+    LLVM_VERSION=$(apt-cache search --names-only '^clang-[0-9]+$' | sort | tail -1 | cut -f1 -d" " | cut -f2 -d"-" ) && \
```
hebasto, line 67: e1ce5667749445c025a3225ff487b31e38195653

This line works only accidentally because the current stable Debian has no one-digit versions of clang.

Perhaps,
```suggestion
    LLVM_VERSION=$(apt-cache search --names-only '^clang-[0-9]+$' | sort -V | tail -1 | cut -f1 -d" " | cut -f2 -d"-" ) && \
```
?

ci/linux-debian.Dockerfile
```diff
@@ -2,6 +2,17 @@ FROM debian:stable-slim
...
+WORKDIR /root
+
+# A too high maximum number of file descriptors (with the default value
+# inherited from the docker host) can cause issues with some of our tools:
+#  - sanitizers hanging: https://github.com/google/sanitizers/issues/1662 
+#  - valgrind crashing: https://stackoverflow.com/a/75293014
+# This is not be a problem on our CI hosts, but developers who run the image
+# on their machines may run into this (e.g., on Arch Linux), so warn them.
+# (Note that .bashrc is only executed in interactive bash shells.)
+RUN echo 'if [[ $(ulimit -n) -gt 200000 ]]; then echo "WARNING: Very high value reported by \"ulimit -n\". Consider passing \"--ulimit nofile=32768\" to \"docker run\"."; fi' >> /root/.bashrc
```
hebasto, line 14: eff42a02c22dcce1a097ab6f122f0fbe5189538a

The downloadable size of the added Docker layer is 595 bytes, which is OK.

hebasto:
I've compare downloadable Docker layers sizes:
- master branch @ 65c79fe2d05d05258eb95a9c51a306841c8febf2:
```
sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1 32B
sha256:dca4ea09548975c9d3eb83c5d6d12b942b499a4d2bd7421366172fe45c8d8788 201B
sha256:3a95c245563807a61f1929525812d0cd7a97e1af9a29bd1e340f782d0ac030d5 29.12MB
sha256:8c5c787d0228d1c83d0dd6bb53ae4dbd78fca356d6ded7231feb2ec5e5874f8d 106.06MB
sha256:1ecdca4ed1cd6f5998d1ecc18886973818b254fce924bb48abbf1478a65c7fc7 491.72MB
sha256:0ff9fabc2d015c9b9ed4a6693d02d19d1b11ce9edf9cf86e85af639eeefc09b2 1.43GB
```

- this PR @ c35e655337e0937d475a0a3fc98458c9c731cfb8:
```
sha256:4f4fb700ef54461cfa02571ae0db9a0dc1e0cdb5577484a6d75e68dc38e8acc1 32B
sha256:e14a3809ad13fbf2d8a6baa3412d21b97038009f65e7fcc6f7a43e6c0a2ed629 201B
sha256:dd221a06e6ff8bfca340a07201c9c1591bfbd04f878c5d87de77d5fd65563442 595B
sha256:3a95c245563807a61f1929525812d0cd7a97e1af9a29bd1e340f782d0ac030d5 29.12MB 
sha256:b92eb44f78d023d7c830186e49b783965f4c2c8048c4b9007cfb5e0610f41eb4 70.68MB
sha256:146e8ec787b6e508eb96dbf8e9659c9ed52ccfcf1763a0f6af3684ba9656f125 492.50MB
sha256:bec59d901529dae853a627ccff960bbe3a2cf9e6a35c1776af625f9d98c3374a 1.43GB
```

It looks like there's a decrease of approximately 35MB.

ci/linux-debian.Dockerfile
real-or-random: thanks, updated!

real-or-random:
> It looks like there's a decrease of approximately 35MB.

Ok, thanks for checking. Not sure if 35 was worth it, but I think removing the build deps after each stage is just cleaner.

hebasto:
ACK 2635068abf93ebcf7f200eef14acafc3300e32f7.

PR #1416, opened 2023-09-04
doc: Align documented scripts with CI ones

hebasto:
Using `ls -RlAh ...` was suggested in https://github.com/bitcoin-core/secp256k1/pull/1413#issuecomment-1697097054.


hebasto:
> While you're at it, can you also adjust this here (if you agree with it)?

Done.

real-or-random (approved):
ACK b0f7bfedc9fcb4f49fe965b16ae49d0e36c1d32b

PR #1422, opened 2023-09-07
cmake: Install `libsecp256k1.pc` file

hebasto:
This PR allows downstream projects to use pkg-config to search for the libsecp256k1 library that is built with CMake.

Addressed https://github.com/bitcoin-core/secp256k1/discussions/1419#discussioncomment-6922896:
> We could just ship the pkg-config file also in CMake builds.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -792,7 +792,7 @@ jobs:
 
       - name: Check installation with Autotools
         env:
-          CI_INSTALL: ${{ runner.temp }}/${{ github.run_id }}${{ github.action }}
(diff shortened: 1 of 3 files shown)


real-or-random (approved):
ACK https://github.com/bitcoin-core/secp256k1/pull/1422/commits/421d84855aeb9324b4f131c3e6c7ee8cf4f5f4e5 I compared the generated pc files and they match in autotools and CMake

(Well, they match except for the `-dev` suffix we have in autotools on non-release commits... but there's not really support for this in CMake, so it seems we have to live with this for now. https://gitlab.kitware.com/cmake/cmake/-/issues/16716 )

Issue #1421, opened 2023-09-06
mingw-w64: "visibility attribute not supported in this configuration; ignored" under `-flto`


fanquake:
Building master (b10ddd2bd2bdce9ca8f2d4733636a9d9e7ac3da1):
```bash
# x86_64-w64-mingw32-gcc (GCC) 12-win32
./autogen.sh
./configure --host=x86_64-w64-mingw32 CFLAGS="-flto" LDFLAGS="-flto"
make -j9
<snip>
src/secp256k1.c: In function 'secp256k1_context_create':
src/secp256k1.c:149:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
  149 | }
      | ^
src/secp256k1.c: In function 'secp256k1_context_preallocated_clone':
src/secp256k1.c:160:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
  160 | }
      | ^
src/secp256k1.c: In function 'secp256k1_context_clone':
src/secp256k1.c:173:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
  173 | }
      | ^
src/secp256k1.c: In function 'secp256k1_context_preallocated_destroy':
src/secp256k1.c:184:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
  184 | }
      | ^
src/secp256k1.c: In function 'secp256k1_context_destroy':
src/secp256k1.c:196:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
  196 | }
      | ^
src/secp256k1.c: In function 'secp256k1_ec_pubkey_parse':
src/secp256k1.c:288:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
  288 | }
      | ^
src/secp256k1.c: In function 'secp256k1_ec_pubkey_serialize':
src/secp256k1.c:311:1: warning: visibility attribute not supported in this configuration; ignored [-Wattributes]
  311 | }
      | ^
< similar output continues>
```

Wanted to raise this, as Core may eventually go down the LTO route (https://github.com/bitcoin/bitcoin/pull/25391), and `-flto` will be getting passed down.

real-or-random:
Hm, thanks. It's a bit mysterious that this happens only with `-flto` and only when compiling tests or noverify_tests, but for example not in the exhaustive_tests. But okay, whatever, apparently we hit that line here, which is supposed to be active only when not on Windows:
[...]

PR #1309, opened 2023-05-12
changelog: Catch up

real-or-random:

PR #1423, opened 2023-09-11
POC: Self-contained headers

hebasto:
This PR adds a tool that verifies every header whether it is self-contained.

As an example, the `field_5x52.h` and `field_10x26.h` headers have been refactored to get self-contained.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -775,7 +775,7 @@ jobs:
 
       - name: Test script
         run: |
-          ./tools/check_header.sh src/field.h
+          ./tools/check_header.sh src/field.h src/field_5x52.h src/field_10x26.h
 
   sage:
     name: "SageMath prover"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -765,6 +765,18 @@ jobs:
             g++ -Werror include/*.h
             clang -Werror -x c++-header include/*.h
 
+  headers:
+    name: "Self-contained headers"
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Test script
+        run: |
+          ./tools/check_header.sh src/field.h
+
   sage:
     name: "SageMath prover"
     runs-on: ubuntu-latest
--- a/src/field_10x26.h
+++ b/src/field_10x26.h
@@ -30,7 +30,10 @@ typedef struct {
      *     sum(i=0..9, n[i] << (i*26)) < p
      *     (together these imply n[9] <= 2^22 - 1)
      */
-    SECP256K1_FE_VERIFY_FIELDS
+#ifdef VERIFY
+    int magnitude;
+    int normalized;
+#endif
 } secp256k1_fe;
 
 /* Unpacks a constant into a overlapping multi-limbed FE element. */
--- a/src/field.h
+++ b/src/field.h
@@ -29,13 +29,6 @@
  * implementation also provides a secp256k1_fe_verify routine to verify that
  * these fields match the run-time value and perform internal consistency
  * checks. */
-#ifdef VERIFY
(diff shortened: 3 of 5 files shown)


real-or-random:
Interesting! 

This raises the question: Should self-containedness be tested in VERIFY mode, or non-VERIFY mode, or both?  I believe at least in VERIFY mode because the goal here will be to help development tooling (e.g., clangd), and development is typically more convenient in VERIFY mode. Otherwise all `#ifdef VERIFY` blocks don't exist, i.e., you don't even get syntax highlighting for them...

Of course, we could manage to make this happen in both modes without too much hassle, that will be ideal.    


This is related to #1309.

hebasto:
> This is related to #1039, where I suggested something like this.

I've been working on this branch a few months but decided to PR it now as I feel it might help with https://github.com/bitcoin-core/secp256k1/issues/1421#issuecomment-1714079714.

src/field_10x26.h
```diff
@@ -30,7 +30,10 @@ typedef struct {
      *     sum(i=0..9, n[i] << (i*26)) < p
      *     (together these imply n[9] <= 2^22 - 1)
      */
-    SECP256K1_FE_VERIFY_FIELDS
+#ifdef VERIFY
+    int magnitude;
+    int normalized;
+#endif
 } secp256k1_fe;
```
real-or-random, line 37: Have you tried/considered including `field.h` instead?

hebasto: Doesn't it create a circular dependency?

real-or-random: In some sense, yes, but it's not a problem because the include guards should break any loops.

If you take that to the extreme, we could also have a single `headers.h` that includes *all* internal headers and include `headers.h` (almost) everywhere. That's a rather simple solution. I'm not convinced that this is very elegant, but it gets the job done...

hebasto: > it's not a problem because the include guards should break any loops.

Right. But not for renaming a header to a source file. It is not clear (at least for me) how to make it meaningful. FWIW, I've already tried the testing script that keeps an original header unmodified.

real-or-random: > Right. But not for renaming a header to a source file. It is not clear (at least for me) how to make it meaningful.

You mean problems will occur if we move code from a header to a source file?

> FWIW, I've already tried the testing script that keeps an original header unmodified.

I don't understand. Can you rephrase?

hebasto: > > Right. But not for renaming a header to a source file. It is not clear (at least for me) how to make it meaningful.
> 
> You mean problems will occur if we move code from a header to a source file?

Yes. I mean:
```sh
mv src/field_10x26.h src/field_10x26.c
gcc -c src/field_10x26.c -o src/field_10x26.o
```

> > FWIW, I've already tried the testing script that keeps an original header unmodified.
> 
> I don't understand. Can you rephrase?

A modified script looks like that:
```sh
cp src/field_10x26.h src/field_10x26.c
gcc -c src/field_10x26.c -o src/field_10x26.o
```

hebasto:
Rebased.

.github/workflows/ci.yml
```diff
@@ -738,6 +738,18 @@ jobs:
             g++ -Werror include/*.h
             clang -Werror -x c++-header include/*.h
 
+  headers:
+    name: "Self-contained headers"
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
fanquake:
```suggestion
        uses: actions/checkout@v5
```

hebasto: Thanks!. Fixed.

hebasto:
Addressed [feedback](https://github.com/bitcoin-core/secp256k1/pull/1423#discussion_r2348405846) from @fanquake.

Additionally, the check script now compiles in a temporary directory rather than in the source tree.

PR #1426, opened 2023-09-20
ci/cirrus: Add native ARM64 jobs

MarcoFalke:


ci/linux-debian.Dockerfile
```diff
@@ -29,11 +29,14 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
...
-        gcc-aarch64-linux-gnu libc6-dev-arm64-cross libc6-dbg:arm64 \
         gcc-powerpc64le-linux-gnu libc6-dev-ppc64el-cross libc6-dbg:ppc64el \
         gcc-mingw-w64-x86-64-win32 wine64 wine \
         gcc-mingw-w64-i686-win32 wine32 \
-        python3
+        python3 && \
+        if ! ( dpkg --print-architecture | grep --quiet "arm64" ) ; then \
+         apt-get install --no-install-recommends -y \
+         gcc-aarch64-linux-gnu libc6-dev-arm64-cross libc6-dbg:arm64 ;\
+        fi
```
real-or-random, line 39: Strictly speaking, this is unrelated to this PR, but I think it's good practice/idiom to clean the cache after every `apt-get` to make the cached layer smaller, so we should add it here

```suggestion
        fi && \
        apt-get clean && rm -rf /var/lib/apt/lists/*
```

real-or-random (commented):
Concept ACK

There are a few lines added in the second commit and removed again in the third commit. I suggest just squashing the two commits.

hebasto:
Where is the docker image cache located?

Is it possible to demonstrate the `build_script` with unpopulated cache?

ci/linux-debian.Dockerfile
```diff
@@ -29,11 +29,14 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
...
         gcc-arm-linux-gnueabihf libc6-dev-armhf-cross libc6-dbg:armhf \
-        gcc-aarch64-linux-gnu libc6-dev-arm64-cross libc6-dbg:arm64 \
         gcc-powerpc64le-linux-gnu libc6-dev-ppc64el-cross libc6-dbg:ppc64el \
         gcc-mingw-w64-x86-64-win32 wine64 wine \
         gcc-mingw-w64-i686-win32 wine32 \
-        python3
+        python3 && \
+        if ! ( dpkg --print-architecture | grep --quiet "arm64" ) ; then \
+         apt-get install --no-install-recommends -y \
+         gcc-aarch64-linux-gnu libc6-dev-arm64-cross libc6-dbg:arm64 ;\
```
hebasto, line 38: On an `arm64` worker, other cross-toolchains are also not used. Maybe move them into this `if` block?

hebasto (approved):
ACK 2f31c66cfb1faf34397ddb3850adbfdb6836586c.

ci/linux-debian.Dockerfile
real-or-random: The reason to keep these packages is that the Docker image can then be used by developers locally and has all the tools available for testing (if they have ARM machines, but this is not unlikely with recent Mac hardware). And okay, yeah, it costs a bit of time on CI, but I assume it's in the range of seconds?!

real-or-random:
> Who should be the author of the commit then?

You and add me in `Co-authored-by:`? Or the other way around, I don't care. Happy to flip a coin for you if you can't decide. :)

real-or-random:
> > You and add me in `Co-authored-by:`?
> 
> Thx, done

Now it's authored *and* co-authored by me. :D

hebasto (approved):
re-ACK fa4d6c76b6dc249d02c926ad75725556614fdddd, only last two commits have been squashed since my recent [review](https://github.com/bitcoin-core/secp256k1/pull/1426#pullrequestreview-1636119941).

real-or-random:
@MarcoFalke Follow-up question: Can the persistent runner run ARM32 binaries natively? From what I remember, most ARM64 machines can do this, but it depends on the CPU. If yes, we could add jobs tests for this, too.

.cirrus.yml
```diff
@@ -0,0 +1,95 @@
...
+      - env
+
+linux_arm64_container_snippet: &LINUX_ARM64_CONTAINER
+  env_script:
+    - env | tee /tmp/env
+  build_script:
+    - DOCKER_BUILDKIT=1 docker build --file "ci/linux-debian.Dockerfile" --tag="ci_secp256k1_arm"
+    - docker image prune --force  # Cleanup stale layers
+  test_script:
+    - docker run --rm --mount "type=bind,src=./,dst=/ci_secp256k1" --env-file /tmp/env --replace --name "ci_secp256k1_arm" "ci_secp256k1_arm" bash -c "cd /ci_secp256k1/ && ./ci/ci.sh"
```
real-or-random, line 59: @maflcko  
I think `--replace` is a podman-only argument. `docker run` doesn't seem to have it. Should we just rewrite the command to use the "native" podman syntax? I guess there's no point in having an invalid docker command here.

real-or-random: On a second thought, we could also prepend `docker rm --force ci_secp256k1_arm`, which has the same effect as `--replace`.

PR #1430, opened 2023-10-17
README: remove CI badge

jonasnick:


real-or-random:
Oh, good catch! The thing is that we now use both Cirrus and CI... Unfortunately, there's no general "badge" for all GitHub "Checks". I guess just keeping the GHA badge is okay because it's where most tasks run. Or we could keep both badges. 

Should we just drop the CI entirely? ^^ GitHub shows the status anyway (with a green checkmark / red cross) on the main repo site. Dropping the badge avoids all the bikesheddy questions above, it saves maintenance work in the future, and it also makes forking easier.

jonasnick:
> Should we just drop the CI entirely?

sgtm

real-or-random (approved):
utACK 5dab0baa80d763b1de599e64307c090d1a3fb04d

PR #1433, opened 2023-10-26
ci/cirrus: Add ARM32 valgrind tasks

real-or-random:
Solves one item in https://github.com/bitcoin-core/secp256k1/issues/1392.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -58,28 +58,59 @@ linux_arm64_container_snippet: &LINUX_ARM64_CONTAINER
   test_script:
     - docker run --rm --mount "type=bind,src=./,dst=/ci_secp256k1" --env-file /tmp/env --replace --name "ci_secp256k1_arm" "ci_secp256k1_arm" bash -c "cd /ci_secp256k1/ && ./ci/ci.sh"
 
-task:
-  name: "ARM64: Linux (Debian stable)"
-  persistent_worker:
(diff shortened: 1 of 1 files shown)


real-or-random:
cc @maflcko

real-or-random:
This is blocked on https://github.com/bitcoin-core/secp256k1/issues/776.

real-or-random:
Closing because we don't use Cirrus anymore.

maflcko:
For reference, GitHub offers `ubuntu-24.04-arm` runners, but I haven't tried if they work for this purpose here.

real-or-random:
> For reference, GitHub offers `ubuntu-24.04-arm` runners, but I haven't tried if they work for this purpose here.

We use these heavily in CI already, but I don't think you can compile our ARM32 assembly for these. We currently test the ARM32 assembly only inside QEMU, and the issue is that this is a somewhat limited (e.g., we can't use Valgrind).

real-or-random:
> > For reference, GitHub offers `ubuntu-24.04-arm` runners, but I haven't tried if they work for this purpose here.
> 
> [...] but I don't think you can compile our ARM32 assembly for these.

Sorry, I'm probably misremembering here. The intent of this PR was exactly this. I opened https://github.com/bitcoin-core/secp256k1/issues/1856 to track this.

PR #1437, opened 2023-11-06
ci: Ignore internal errors of snapshot compilers

hebasto:
It was discussed on today's IRC meeting.

### DIFF
--- a/ci/ci.sh
+++ b/ci/ci.sh
@@ -83,7 +83,21 @@ esac
     --host="$HOST" $EXTRAFLAGS
 
 # We have set "-j<n>" in MAKEFLAGS.
-make
+build_exit_code=0
+make > make.log 2>&1 || build_exit_code=$?
+cat make.log
+if [ $build_exit_code -ne 0 ]; then
+    case "${CC:-undefined}" in
+        *snapshot*)
+            # Ignore internal compiler errors in gcc-snapshot and clang-snapshot
+            grep -e "internal compiler error:" -e "PLEASE submit a bug report" make.log
+            return $?;
+            ;;
+        *)
+            return 1;
+            ;;
+    esac
+fi
 
 # Print information about binaries so that we can see that the architecture is correct
 file *tests* || true


ci/ci.sh
```diff
@@ -83,7 +83,19 @@ esac
     --host="$HOST" $EXTRAFLAGS
 
 # We have set "-j<n>" in MAKEFLAGS.
-make
+build_exit_code=0
+make || build_exit_code=$?
```
real-or-random, line 87: nit: This should be simpler
```suggestion
make
build_exit_code=$?
```

ci/ci.sh
```diff
@@ -83,7 +83,19 @@ esac
...
+if [ $build_exit_code -ne 0 ]; then
+    case $CC in
+        *snapshot*)
+            make 2>&1 >/dev/null | grep "internal compiler error:"
```
real-or-random, line 91: Regarding `make 2>&1 >/dev/null`:

I think it would be more reliable to run `make` only once. (Perhaps gcc-snapshot crashes in one run but not in another.)

One way could be: Change the first `make` to `make | tee make.log` and then use `grep` on `make.log`.

real-or-random, line 91: Regarding `grep "internal compiler error:"`:

Could also additionally grep for "PLEASE submit a bug report". This is what clang will output when it crashes.

real-or-random (commented):
Concept ACK, thanks for the quick response

ci/ci.sh
hebasto: > nit: This should be simpler

Doesn't `set -eux` in the top of the script force exiting when `make` fails?

hebasto:
Updated d14484f81a0799917ce5799a78b2a9fdb92711f8 -> 5a8db10c59974a4a9b0a00d2954e6a5cd3fd055f ([pr1437.01](https://github.com/hebasto/secp256k1/commits/pr1437.01) -> [pr1437.03](https://github.com/hebasto/secp256k1/commits/pr1437.03), [diff](https://github.com/hebasto/secp256k1/compare/pr1437.01..pr1437.03)).

Addressed @real-or-random's comments:
- https://github.com/bitcoin-core/secp256k1/pull/1437#discussion_r1384697101
- https://github.com/bitcoin-core/secp256k1/pull/1437#discussion_r1384698745

ci/ci.sh
hebasto: Thanks! [Done](https://github.com/bitcoin-core/secp256k1/pull/1437#issuecomment-1798325434).

hebasto: Sure. [Done](https://github.com/bitcoin-core/secp256k1/pull/1437#issuecomment-1798325434).

real-or-random (approved):
ACK mod nit and CI results

hebasto:
Updated 5a8db10c59974a4a9b0a00d2954e6a5cd3fd055f -> 61575fd5a647ff6ca0b0b4111b21dc1e0a12d5ae ([pr1437.03](https://github.com/hebasto/secp256k1/commits/pr1437.03) -> [pr1437.04](https://github.com/hebasto/secp256k1/commits/pr1437.04), [diff](https://github.com/hebasto/secp256k1/compare/pr1437.03..pr1437.04)).

Addressed @real-or-random's [comment](https://github.com/bitcoin-core/secp256k1/pull/1437#discussion_r1384788153).

ci/ci.sh
```diff
@@ -83,7 +83,20 @@ esac
...
 
 # We have set "-j<n>" in MAKEFLAGS.
-make
+build_exit_code=0
+make > make.log 2>&1 || build_exit_code=$?
+cat make.log
+if [ $build_exit_code -ne 0 ]; then
+    case $CC in
+        *snapshot*)
+            grep -e "internal compiler error:" -e "PLEASE submit a bug report" make.log
```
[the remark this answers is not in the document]
hebasto: [Done](https://github.com/bitcoin-core/secp256k1/pull/1437#issuecomment-1798404095).

real-or-random:
Real failures are still reported as real failures:
 https://github.com/bitcoin-core/secp256k1/actions/runs/6791773753/job/18463903460?pr=1440

But for some of them, `CC` is not set (and we let configure detect the right `CC`)
https://github.com/bitcoin-core/secp256k1/actions/runs/6791773753/job/18463894482?pr=1440#step:3:780

Perhaps just add a default substitution in case CC is not set. This should be good enough. The snapshots are only used if we ask for them explicitly by setting CC. We could also read `CC` from the `CC=` line in config.log, but that sounds like overkill.

hebasto:
> But for some of them, `CC` is not set...

Solved by using the same approach as used in other places of the `ci.sh` script.

real-or-random (approved):
ACK 8185e72d299bc77de9c06cc92fd1988676df3bc1

See https://github.com/real-or-random/secp256k1/actions/runs/6802798679/job/18496698030 for a real build failure on top of this PR

real-or-random:
Merging this, we shouldn't hold this up because a broken CI is very annoying.

Issue #1439, opened 2023-11-07
ci: Random Valgrind MacOS build failures


real-or-random:
I had to restart this job because make failed for Valgrind MacOS:
https://github.com/bitcoin-core/secp256k1/actions/runs/6777069897/job/18419944841?pr=1437

We've seen other (different) random build issues before: https://github.com/LouisBrunner/valgrind-macos/issues/26

We could print the logs The make log is in `/Users/runner/Library/Logs/Homebrew/valgrind/04.make`. Perhaps cat just all of ``/Users/runner/Library/Logs/Homebrew/valgrind/*`

But I suspect these are concurrency issues. We could prefix the command with `MAKEFLAGS="-j1"` when we call brew. Or just retry make if it fails... cc @hebasto

real-or-random:
Hm, this is another one, but this one is a linker error and it looks very similar to https://github.com/LouisBrunner/valgrind-macos/issues/26: 
https://github.com/bitcoin-core/secp256k1/actions/runs/6756875722/job/18366814430?pr=1184#step:4:245

I don't know to be honest...

hebasto:
Is this still an issue?

real-or-random:
Hm, not entirely sure, but I haven't seen this in a while. Let's just close the issue, it's anyway not very actionable. And we can always reopen.

PR #1440, opened 2023-11-07
CI TEST

real-or-random:


real-or-random:
Checking if real compilation failures are still reported as failing in the CI for #1439. And yes, this is working. :)

PR #1442, opened 2023-11-14
Return temporaries to being unsigned in secp256k1_fe_sqr_inner

roconnor-blockstream:
These temporaries seem to been inadvertently changed to signed during a refactoring.  Generally, bit shifting is frowned upon for signed values.


real-or-random:
> Generally, bit shifting is frowned upon for signed values.

True, at least for left shifting. In this case, the only relevant shift is `u0 << 4`, and here we know that u0 is only 52 bits, so this is fine. But yes, this should be fixed, unsigned is just safer. Thanks for the PR.

real-or-random (approved):
utACK 10271356c8fc34395850ac70df5902571945fbea

real-or-random:
> The stack layout change is strange, [...]

I guess my conclusion is simply that compilers are strange?

roconnor-blockstream:
Maybe stack layout is arbitrary and the compiler sorts variables by type?

Issue #726, opened 2020-03-24
Consider removing x86_64 field assembly


jonasnick:
I noticed that turning off x86_64 assembly on my laptop actually speeds up `ecdsa_verify`. The internal benchmarks show that `--without-asm` scalar operations are slower, but field operations are faster. In order to investigate this I created a branch that includes the configurable benchmark iterations from #722, a test_matrix.sh script and allows turning on field assembly individually (https://github.com/jonasnick/secp256k1/tree/eval-asm).

Here are the results with gcc 9.3.0 (got similar results with clang 9.0.1):
```
SECP256K1_BENCH_ITERS=200000

bench config CFLAGS=-DUSE_ASM_X86_64_FIELD ./configure --disable-openssl-tests --with-asm=x86_64
scalar_sqr: min 0.0331us / avg 0.0332us / max 0.0337us
scalar_mul: min 0.0342us / avg 0.0343us / max 0.0345us
field_sqr: min 0.0165us / avg 0.0165us / max 0.0167us
field_mul: min 0.0204us / avg 0.0205us / max 0.0209us
ecdsa_sign: min 40.3us / avg 40.3us / max 40.4us
ecdsa_verify: min 56.9us / avg 56.9us / max 56.9us

bench config CFLAGS= ./configure --disable-openssl-tests --without-asm
scalar_sqr: min 0.0375us / avg 0.0376us / max 0.0383us
scalar_mul: min 0.0362us / avg 0.0366us / max 0.0396us
field_sqr: min 0.0152us / avg 0.0152us / max 0.0152us
field_mul: min 0.0177us / avg 0.0178us / max 0.0178us
ecdsa_sign: min 41.8us / avg 41.8us / max 41.9us
ecdsa_verify: min 54.6us / avg 54.7us / max 54.7us

bench config CFLAGS= ./configure --disable-openssl-tests --with-asm=x86_64
scalar_sqr: min 0.0331us / avg 0.0331us / max 0.0333us
scalar_mul: min 0.0342us / avg 0.0343us / max 0.0347us
field_sqr: min 0.0152us / avg 0.0153us / max 0.0154us
field_mul: min 0.0178us / avg 0.0178us / max 0.0180us
ecdsa_sign: min 40.3us / avg 40.3us / max 40.4us
ecdsa_verify: min 53.2us / avg 53.2us / max 53.2us
```

Note the 6.5% ecdsa_verify speedup. However, I don't fully understand this:

[...]

PR #1445, opened 2023-11-23
bench: add --help option to bench_internal

theStack:
While coming up with commands for running the benchmarks for issue https://github.com/bitcoin-core/secp256k1/issues/726#issuecomment-1824625653, I noticed that in contrast to `bench{_ecmult}`, `bench_internal` doesn't have a help option yet and figured it would be nice to have one. A comparable past PR is https://github.com/bitcoin-core/secp256k1/pull/1008. Benchmark categories appear in the same order as they are executed, the concrete benchmark names in parantheses per category are listed in alphabetical order.

### DIFF
--- a/src/bench_internal.c
+++ b/src/bench_internal.c
@@ -17,6 +17,25 @@
 #include "ecmult_impl.h"
 #include "bench.h"
 
+static void help(int default_iters) {
+    printf("Benchmarks various internal routines.\n");
+    printf("\n");
+    printf("The default number of iterations for each benchmark is %d. This can be\n", default_iters);
+    printf("customized using the SECP256K1_BENCH_ITERS environment variable.\n");
+    printf("\n");
+    printf("Usage: ./bench_internal [args]\n");
+    printf("By default, all benchmarks will be run.\n");
+    printf("args:\n");
+    printf("    help       : display this help and exit\n");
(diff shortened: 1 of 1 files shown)


src/bench_internal.c
```diff
@@ -17,6 +17,31 @@
...
 #include "bench.h"
 
+static void help(int default_iters) {
+    printf("Benchmarks internal routines for the following areas:\n");
+    printf("    - Scalar operations (modulo the curve's order)\n");
+    printf("    - Field operations (modulo the curve's field size)\n");
+    printf("    - Group operations (both in Jacobian and affine coordinates)\n");
+    printf("    - Point multiplication\n");
+    printf("    - Hash algorithms\n");
+    printf("    - Context object handling\n");
```
real-or-random, line 27: Perhaps it suffices to say "Benchmarks various internal routines". This list is mostly redundant to the list below. If we omit it, we won't need to (remember to) touch two lists every time we add or remove operations.

real-or-random (commented):
Concept ACK

src/bench_internal.c
theStack: Agree, done.

real-or-random (approved):
utACK 1ddd76af0a735b7fcbec7f37c0d99a7db9893ac1

siv2r (commented):
ACK 1ddd76a, tested the `--help` option locally, and it works as expected.

PR #1446, opened 2023-11-23
field: Remove x86_64 asm 

real-or-random:
Widely available versions of GCC and Clang beat our field asm on -O2.
In particular, GCC 10.5.0, which is Bitcoin Core's current compiler
for official x86_64 builds, produces code that is > 20% faster for
fe_mul and > 10% faster for signature verification (see #726).

These are the alternatives to this PR:

We could replace our current asm with the fastest compiler output
that we can find. This is potentially faster, but it has multiple
drawbacks:
 - It's more coding work because it needs detailed benchmarks (e.g.,
   with many compiler/options).
 - It's more review work because we need to deal with inline asm
   (including clobbers etc.) and there's a lack of experts reviewers
   in this area.
 - It's not unlikely that we'll fall behind again in a few compiler
   versions, and then we have to deal with this again, i.e., redo the
   benchmarks. Given our history here, I doubt that we'll revolve
   this timely.

We could change the default of the asm build option to off. But this
will also disable the scalar asm, which is still faster.

We could split the build option into two separate options for field
and scalar asm and only disable the field asm by default. But this
adds complexity to the build and to the test matrix.

My conclusion is that this PR gets the low-hanging fruit in terms of
performance. It simplifies our code significantly. It's clearly an
improvement, and it's very easy to review. Whether re-introducing
better asm (whether from a compiler or from CryptOpt) is worth the
hassle can be evaluated separately, and should not hold up this
improvement.

Solves #726.


theStack (commented):
Concept ACK

Two PR description / commit message nits:
- s/10.0.5/10.5.0/ (GCC version number)
- s/723/726/ (linked issue number)

real-or-random:
> Two PR description / commit message nits:

Force-pushed to fix. (I had already force-pushed multiple times yesterday to fix typos in the commit message... The commit message is by far the most complex part of this PR. :D)

theStack (approved):
ACK f07cead0ca96e26356466b635ce6e7fe3834c949

Out of curiosity, I ran the benchmarks again with gcc-11 (which is the default on Ubuntu 22.04 LTS) and clang-14. As expected, the PR branch is faster for sign/verify and the relevant field operations (mul and sqr) on both:
```
======================
===== gcc 11.4.0 =====
======================

master (with field asm):

$ ./configure CC=gcc-11 && make clean && make && ./bench verify sign && ./bench_internal mul split sqr | tail -n +5

ecdsa_verify                  ,    32.9       ,    33.6       ,    35.7    
ecdsa_sign                    ,    24.8       ,    25.4       ,    28.2    
schnorrsig_sign               ,    18.9       ,    19.3       ,    20.7    
schnorrsig_verify             ,    33.3       ,    34.1       ,    36.1    
field_sqr                     ,     0.0165    ,     0.0185    ,     0.0214 
field_mul                     ,     0.0182    ,     0.0195    ,     0.0228

pr1446 (without field asm):

$ ./configure CC=gcc-11 && make clean && make && ./bench verify sign && ./bench_internal mul split sqr | tail -n +5

ecdsa_verify                  ,    28.9       ,    29.4       ,    30.1    
ecdsa_sign                    ,    23.0       ,    23.3       ,    23.9    
schnorrsig_sign               ,    17.2       ,    17.5       ,    18.1    
schnorrsig_verify             ,    30.0       ,    30.3       ,    31.0    
field_sqr                     ,     0.0130    ,     0.0167    ,     0.0475 
field_mul                     ,     0.0136    ,     0.0147    ,     0.0161


========================
===== clang 14.0.0 =====
========================

master (with field asm):

$ ./configure CC=clang-14 && make clean && make && ./bench verify sign && ./bench_internal mul split sqr | tail -n +5

ecdsa_verify                  ,    32.7       ,    34.7       ,    38.7    
ecdsa_sign                    ,    24.6       ,    25.0       ,    25.6    
schnorrsig_sign               ,    18.3       ,    18.4       ,    19.4    
schnorrsig_verify             ,    32.8       ,    33.1       ,    33.8    
field_sqr                     ,     0.0152    ,     0.0158    ,     0.0165 
field_mul                     ,     0.0175    ,     0.0181    ,     0.0191


pr1446 (without field asm):

$ ./configure CC=clang-14 && make clean && make && ./bench verify sign && ./bench_internal mul split sqr | tail -n +5

ecdsa_verify                  ,    30.6       ,    31.0       ,    32.3    
ecdsa_sign                    ,    24.7       ,    24.9       ,    25.7    
schnorrsig_sign               ,    18.2       ,    18.4       ,    18.6    
schnorrsig_verify             ,    31.2       ,    31.6       ,    32.6    
field_sqr                     ,     0.0117    ,     0.0123    ,     0.0133 
field_mul                     ,     0.0139    ,     0.0148    ,     0.0168
```
(Interesting that clang's generated code for `_fe_sqr` performs so much better than gcc's, being ~35% faster).

real-or-random:
> (Interesting that clang's generated code for `_fe_sqr` performs so much better than gcc's, being ~35% faster).

Perhaps this was some glitch in the benchmark, given the rather high "max" value for gcc? `sqr` is just a specialization of `mul`, so its only purpose is to be faster than `mul`.

jonasnick (approved):
ACK f07cead0ca96e26356466b635ce6e7fe3834c949

I ran the benchmarks with gcc 10.5.0 (before noticing that this had already been done in 726) and confirmed the >10% speedup. I think speedups of that magnitude (and common/default configure settings) justify a mention in the changelog.

deadalnix:
> It's not unlikely that we'll fall behind again in a few compiler versions

I can confirm, I've been working on this on the LLVM front, and there is still a lot more it can do.

russeree:
Concept Ack; 

Nit: It is critical that the output is 1:1 compatible with existing implementation because inducing a bug though the optimizations would lead to a chainsplit if on one side a sig validates and on another it doesn't for any reason.

PR #1450, opened 2023-12-01
Add group.h ge/gej equality functions

sipa:
This pull requests removes the test-only functions `ge_equals_ge` and `ge_equals_gej`, and replaces them with proper group.h functions `secp256k1_ge_eq_var` and `secp256k1_gej_eq_ge_var` (mimicking the existing `secp256k1_gej_eq_var` function).

This drops some of the arbitrary and undocumented magnitude restristrictions these functions have, makes them properly tested on their own, and makes their semantics cleaner (I'm always left checking whether `ge_equals_ge` does a `CHECK` internally or whether it returns a value...).

### DIFF
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -354,6 +354,35 @@ static int secp256k1_gej_eq_var(const secp256k1_gej *a, const secp256k1_gej *b) 
     return secp256k1_gej_is_infinity(&tmp);
 }
 
+static int secp256k1_gej_eq_ge_var(const secp256k1_gej *a, const secp256k1_ge *b) {
+    secp256k1_gej tmp;
+    SECP256K1_GEJ_VERIFY(a);
+    SECP256K1_GE_VERIFY(b);
+
+    secp256k1_gej_neg(&tmp, a);
+    secp256k1_gej_add_ge_var(&tmp, &tmp, b, NULL);
+    return secp256k1_gej_is_infinity(&tmp);
+}
+
+static int secp256k1_ge_eq_var(const secp256k1_ge *a, const secp256k1_ge *b) {
+    secp256k1_fe tmp;
+    SECP256K1_GE_VERIFY(a);
+    SECP256K1_GE_VERIFY(b);
+
+    if (a->infinity != b->infinity) return 0;
+    if (a->infinity) return 1;
+
+    tmp = a->x;
+    secp256k1_fe_normalize_weak(&tmp);
+    if (!secp256k1_fe_equal(&tmp, &b->x)) return 0;
+
+    tmp = a->y;
+    secp256k1_fe_normalize_weak(&tmp);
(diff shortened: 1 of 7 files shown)


src/group_impl.h
```diff
@@ -354,6 +354,35 @@ static int secp256k1_gej_eq_var(const secp256k1_gej *a, const secp256k1_gej *b)
...
+    secp256k1_fe tmp;
+    SECP256K1_GE_VERIFY(a);
+    SECP256K1_GE_VERIFY(b);
+
+    if (a->infinity != b->infinity) return 0;
+    if (a->infinity) return 1;
+
+    tmp = a->x;
+    secp256k1_fe_normalize_weak(&tmp);
+    if (!secp256k1_fe_equal(&tmp, &b->x)) return 0;
```
real-or-random, line 377: We know that the magnitude of `a->x` is at most `SECP256K1_GE_X_MAGNITUDE_MAX`, so normalization shouldn't be necessary. And then `tmp` is not necessary.

Same for y below.

src/tests.c
```diff
@@ -3730,6 +3731,22 @@ static void test_ge(void) {
             random_gej_y_magnitude(&gej[1 + j + 4 * i]);
             random_gej_z_magnitude(&gej[1 + j + 4 * i]);
         }
+
+        for (j = 0; j < 4; ++j) {
+            for (k = 0; k < 4; ++k) {
+                if ((j >> 1) == (k >> 1)) {
+                    CHECK(secp256k1_ge_eq_var(&ge[1 + j + 4 * i], &ge[1 + k + 4 * i]));
```
real-or-random, line 3738: nit: could rewrite to 
```
int expect_equal = ((j >> 1) == (k >> 1));
CHECK(secp256k1_ge_eq_var(&ge[1 + j + 4 * i], &ge[1 + k + 4 * i]) == except_equal);
...
```

real-or-random (commented):
Concept ACK

src/group_impl.h
real-or-random: Nevermind. (Duh, too.)

real-or-random (commented):
Naming nit: the field equality function has a suffix `_equal` instead of `_eq`. It would be good to make this consistent. 

 (Sorry for bringing this up now after you addressed the other review... utACK otherwise)

real-or-random:
> > Naming nit: the field equality function has a suffix _equal instead of _eq. It would be good to make this consistent.
> 
> I noticed that too, but that's an existing inconsistency (there is also a `secp256k1_gej_eq_var` function already).

Oh, indeed. I suggest renaming the fe function to `_eq` but this can be done in a separate PR.

real-or-random (approved):
utACK 04af0ba162b152073455a5ccbb2c5833ae6d1d57

stratospher (commented):
ACK 04af0ba.

PR #1451, opened 2023-12-05
changelog: add entry for "field: Remove x86_64 asm"

jonasnick:


real-or-random (commented):
- I found it a bit confusing to mention that now  `--with-asm` is faster when we say at the same time that we remove asm.
 - Just `--with-asm` doesn't work because that is equivalent to `--with-asm=` (empty string), you need `--with-asm=x86_64`.
 - This should probably also mention the CMake option
 - nit: We use the spelling `x86_64` in other places (e.g., user-facing build system stuff)
 - not sure if we need to include the link to the benchmark.

Suggestion:
Optional handwritten x86_64 assembly for field operations was removed because modern C compilers are able to output more efficient assembly. This change results in a significant speedup of some library functions when handwritten x86_64 assembly is enabled (`--with-asm=x86_64` in GNU Autotools, `-DSECP256K1_ASM=x86_64` in CMake), which is the default on x86_64. Benchmarks with GCC 10.5.0 show a 10% speedup for `secp256k1_ecdsa_verify` and `secp256k1_schnorrsig_verify` (see https://github.com/bitcoin-core/secp256k1/issues/726#issuecomment-1824625653).

jonasnick:
Replaced with your suggestion

real-or-random (commented):
ACK d2e36a2b81c0ba9d28425ea14c1ff6523156ee52

 > (see https://github.com/bitcoin-core/secp256k1/issues/726#issuecomment-1824625653).

Just checking if you omitted this intentionally. If you ask me, it's perfectly fine to omit it, we usually don't link to the related issues/PRs in the Changelog, so it's, in fact, consistent with the current changelog.

jonasnick:
Yes, omitted intentionally.

Issue #1427, opened 2023-10-01
Silent Payments (BIP352) module -- discussion about scope and interface


theStack:
At the last CoreDev meeting the idea of creating a libsecp Silent Payments (BIP352) module for handling the low-level parts [was brought up](https://btctranscripts.com/bitcoin-core-dev-tech/2023-09/libsecp256k1-meeting/). I've spent some time recently studying the BIP and  it's current implementation (https://github.com/bitcoin/bips/pull/1458, see also tracking issue https://github.com/bitcoin/bitcoin/issues/28536) and thought about what parts would make sense to put into this new module. A good amount of code is ready to use, but before opening a concrete PR, I'd like to discuss the related questions 
* _What is the exact scope of a potential libsecp256k1 silent payments module?_ and 
* _What should a sane interface look like?_

Note that when I talk about the caller side I'm not only thinking about Bitcoin Core (though this would obviously be the first one taking use of it); one of the benefits of providing a module would be to help other implementations with the hard low-level math parts about Silent Payments, i.e. the interface should definitely have the goal to be easy and intuitive to use.

I'm listing some basic ingredients from the Silent Payments protocol and  my thoughts about scope/interface to each of them:

### 1. calculate the outpoints hash (both sender and receiver sides) 
(see https://github.com/josibake/bips/blob/silent-payments-bip/bip-0352.mediawiki#user-content-Outpoints_hash)

This step includes sorting a given list of serialized 36-bytes outpoints (=32-bytes txid, 4-bytes vout) and then applying a SHA256 hash to the  concatenated result. Should a SP module living in libsecp256k1 land be involved in any of this at all? Three variants are possible:

1) both sorting and hashing is done by the SP module
2) sorting is done by the caller, hashing is done by the SP module
3) both sorting and hashing has to be done by the caller

For the sake of simplicity, I'm leaning towards variant 3) here. Though it would be ideal if users wouldn't need to take care of any  intermediate computations needed for the protocol at all (not even hashing), it seems a bit odd to pass in specific transaction-internal data to an ECC-focused library. Looking at other libsecp modules, we primarily work with "curve-relevant" inputs like scalars and group elements (and their serializations), and I feel that handling more than that is out of scope and would be better suited for a different, higher-level library. Invoking a sorting algorithm in libsecp256k1 also just feels kind of odd (though technically easily possible, as [qsort(3)](https://man7.org/linux/man-pages/man3/qsort.3.html)
is part of the standard library for C89).

[...]

PR #1471, opened 2023-12-23
Add silentpayments (BIP352) module

theStack:
This PR adds a new Silent Payments ([BIP352](https://github.com/bitcoin/bips/pull/1458)) module to secp256k1. The following routines are provided ($a_i$ are input private keys, $A_i$ are input public keys, $b$ and $B$ denote recipient privkeys/pubkeys that would be encoded in silent payment addresses, $d$ and $P$ the keypair for the actual transaction taproot x-only output):

| Side     | Function                        | Inputs                           | Outputs                                                                   |
|----------|---------------------------------|----------------------------------|---------------------------------------------------------------------------|
|  Sender  | `_create_private_tweak_data`    | $a_1...a_n$, $outpoint_L$        | $a_{sum} = (a_1 + a_2 + ... + a_n)$<br>$inputhash =  hash_I(outpoint_L \|\| (a_{sum} * G))$ |
| Receiver | `_create_public_tweak_data`     | $A_1...A_n$, $outpoint_L$        | $A_{sum} = (A_1 + A_2 + ... + A_n)$<br>$inputhash = hash_I(outpoint_L \|\| A_{sum})$ |
| Receiver | `_create_tweaked_pubkey`       |  $A_{sum}, inputhash$       |  $A_{tweaked} = inputhash * A_{sum}$
|  Both  | `_create_shared_secret`    | $Pub$, $sec$<br>(Sender: $B_{scan}, a_{sum}$<br>Receiver: $A_{sum}, b_{scan}$<br>Lightclient: $A_{tweaked}, b_{scan}$)        | $SS = (inputhash * sec) * Pub$ (ECDH)                                      |
| Receiver | `_create_label_tweak`           | $b_{scan}, m$                    | $labeltweak = hash_L(b_{scan} \|\| m)$<br>$label = labeltweak * G$                                    |
| Receiver | `_create_address_spend_pubkey`  | $B_{spend}, label$          | $B_m = B_{spend} + label$                                        |
| Sender   | `_sender_create_output_pubkey`         | $SS, B_m, k$ | $P_{xonly} = B_m + hash_S(SS \|\| k) * G$        |
| Receiver | `_receiver_scan_output`               | $SS, B_m, k, tx_{output}$        | $t_k = hash_S(SS \|\| k)$<br>$P_{xonly} = B_m + t_k * G$  [not returned]<br>$directmatch = P_{xonly} == tx_{output}$<br>if $directmatch == 0$:<br>$\quad label1 = tx_{output} - P$<br>$\quad label2 = -tx_{output} - P$
| Receiver | `_create_output_seckey`         | $b_{spend}, t_k, (labeltweak)$ | $d = (b_{spend} + labeltweak) + t_k$                        |

where
* $hash_I$ denotes a SHA256 tagged hash with tag "BIP0352/Inputs"
* $hash_L$ denotes a SHA256 tagged hash with tag "BIP0352/Label"
* $hash_S$ denotes a SHA256 tagged hash with tag "BIP0352/SharedSecret"

For ending up with output key material used for sending to / scanning for / spending from, both sides would follow the chain of `tweak_data -> shared_secret -> output key`. The public tweak data can be useful for faster scanning of output transactions by storing them in an index, see e.g. Bitcoin Core PR https://github.com/bitcoin/bitcoin/pull/28241. Private tweak data is arguably less useful, so in theory one could collapse the tweak data and shared secret creation functions into a single one, but IMHO it's nicer if the API is symmetric.

As discussed in https://github.com/bitcoin-core/secp256k1/issues/1427#issuecomment-1757675172, the approach of passing in two separate key pairs for taproot and non-taproot inputs is followed here. This may seem a bit confusing at first, but has the advantage that the caller doesn't have to deal with enforcing even y-parity for key material manually (e.g. negating private keys of taproot outputs if they would end up in an odd point), which seems error-prone.

The last commit contains the BIP352 test vectors, converted to C code with a Python script. An earlier version of the tests, directly written in Python (by calling in to the secp256k1 shared library using ctypes) can still be found here: https://github.com/theStack/secp256k1/tree/silentpayments-module-demo_old5


josibake:
Thanks for starting this! Will review. Just wanted to comment that I've made a round of edits to the BIP and it is now updated with the new hashing method (along with the tests).

> The PR assumes that the outpoint hash is provided by the caller, but this could change in the future if its calculation involves elliptic-curve operations that would be a good fit for being done within the module as well.

The hash is calculated as *hash(outpoint<sub>L</sub> || A<sub>sum</sub>)*. Since it commits to the sum of input public keys used, seems like it might make sense to move it into the module? Otherwise, we'd have to do something like calculate the sum of the public keys, return that to the caller so that they can calculate an input hash and then send it back to us.

josibake:
> but then we would need to pass the input pubkeys and calculate the pubkey sum twice

(I need to actually review your implementation, but..) I don't think we need to do it twice? On the sender side, you pass in the private keys, add those up (call it `priv_key_sum`) and then you can generate the pubkey sum from the private keys (e.g. `priv_key_sum.GetPubKey()`. On the receiving side, you pass in the pubkeys, sum them, use them in the hash and then use them in the shared secret derivation.

include/secp256k1_silentpayments.h
```diff
@@ -89,7 +89,39 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_send_cre
...
+ * Given a list of n public keys A_1...A_n (one for each input to spend)
+ * and a serialized outpoint_lowest, compute the corresponding input
+ * public keys tweak data:
+ *
+ * A_tweaked = (A_1 + A_2 + ... + A_n) * hash(outpoint_lowest || A)
+ *
+ * If necessary, the public keys are negated to enforce the right y-parity.
+ * For that reason, the public keys have to be passed in via two different parameter
+ * pairs, depending on whether they were used for creating taproot outputs or not.
+ * The resulting data is needed to create a shared secret for the receiver's side.
```
josibake, line 103: Do we need this? By definition, taproot public keys have even y-parity. My understanding is that this only an issue when dealing with private keys, since for a given x-only public key there are two possible private keys (*d*, and *n - d* and we need to make sure we pick the correct private key (the one that produces a point with even parity).

We could just have the caller pass in one list of public keys (which means they would have already converted them to compressed keys by prefixing them with `0x02`.

src/modules/silentpayments/main_impl.h
```diff
@@ -141,6 +141,60 @@ int secp256k1_silentpayments_send_create_shared_secret(const secp256k1_context *
...
+        secp256k1_ge addend;
+        secp256k1_pubkey_load(ctx, &addend, &plain_pubkeys[i]);
+        secp256k1_gej_add_ge(&A_sum_gej, &A_sum_gej, &addend);
+    }
+
+    /* X-only public keys have to be converted to regular public keys (assuming even parity) */
+    for (i = 0; i < n_xonly_pubkeys; i++) {
+        secp256k1_ge addend;
+        secp256k1_xonly_pubkey_load(ctx, &addend, &xonly_pubkeys[i]);
+        if (secp256k1_fe_is_odd(&addend.y)) {
```
josibake, line 175: I might be missing something here, but I think this will always be true, right? AFAIU, when we "load" an x-only pubkey, we *choose* the even y point.

include/secp256k1_silentpayments.h
```diff
@@ -167,6 +167,29 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_create_l
...
+ *
+ *  Given a recipient's spend public key B_spend and a label_tweak, calculate
+ *  the corresponding serialized labelled spend public key:
+ *
+ *  B_m = B_spend + label_tweak * G
+ *
+ *  The result is used by the receiver to create a Silent Payment address, consisting
+ *  of the serialized and concatenated scan public key and (labelled) spend public key each.
+ *
+ *  Returns: 1 if labellend spend public key creation was successful. 0 if an error occured.
```
josibake, line 180:
```suggestion 
 *  Returns: 1 if labelled spend public key creation was successful. 0 if an error occurred.
```

josibake (commented):
Looks great! Did a quick pass, mostly spelling and wording nits.

Another thought I had (need to actually re-review more carefully to confirm) is we might be able to minimize the de-serializations of keys (moving from bytes -> point). From my understanding, this is one of the "expensive" operations as it involves calculating the y value. This would mean either combining things into larger routines, or having the functions return points instead of serialized pubkeys. Curious what you think @theStack

include/secp256k1_silentpayments.h
josibake: Thanks for the link to the discussion! Reading through that again, I agree it's probably better to just have two lists: one for public keys and another for x-only public keys. Despite a more complicated API, it does make the module more user friendly in that users can pass arguments directly without needing to do any preprocessing.

josibake:
> EDIT: oh, you probably meant the raw 33-byte results of the shared secret creation and public tweak data. Maybe those should be changed to secp256k1_pubkey instead.

Yep! Sorry, my original comment wasn't very clear. But to be precise, what I'm referring to is anytime we have to de-serialize a public key encoding into a point (be it compressed or x-only), we need to calculate a `sqrt` to get the y-value, which is the "expensive" part. So if we only de-serialize once (taking the inputs from the user) and then for the rest of the process use points (i.e. (x, y) pairs) until the final step where we return the generated x-only pubkeys to the user, this should save us some work!

At a quick glance, it looks like you addressed this in your most recent push!

josibake:
>  It might make sense to have a return code signalling "ignore this tx, it's not suitable for silent payments" to differentiate it from an actual error that happened.

Thinking about this a bit more, the errors that can happen are:

1. De-serialization errors: I pass some bytes that cannot be de-serialized into a valid public key
2. Hash returns something greater than the curve order: extremely unlikely to *ever* happen
3. Pubkey / private key sum is point at infinity / 0: extremely unlikely or malicious

I think in all cases the expected user behavior is to move on. For the sender, if they run into any of these errors they would need to pick a different set of inputs (i.e. make a new transaction). For the receiver, their only recourse is to skip the transaction.

I'm not sure what the added value to the user is here if we return different error codes.

include/secp256k1_silentpayments.h
```diff
@@ -89,7 +89,39 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_send_cre
...
+ *  Out:     public_tweak_data: pointer to the resulting public keys tweak data
+ *  In:          plain_pubkeys: pointer to an array of non-taproot public keys
+ *                              (can be NULL if no non-taproot inputs are used)
+ *             n_plain_pubkeys: the number of non-taproot input public keys
+ *               xonly_pubkeys: pointer to an array of taproot x-only public keys
+ *                              (can be NULL if no taproot input public keys are used)
+ *             n_xonly_pubkeys: the number of taproot input public keys
+ *         outpoint_smallest36: serialized smallest outpoint
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_create_public_tweak_data(
```
josibake, line 116: in https://github.com/bitcoin-core/secp256k1/pull/1471/commits/2290e80067b9127ca7931ed4677f688070ec3c5a:

This function covers a very important use case: preparing *public* transaction input data so that it can be served to a light client (a light client in this scenario is any client that does not have access to the blockchain but does have access to their `b_scan` private key).

However, if used by a full node client, this function would cause the receiver to do two ECC multiplications:

1. `A_sum * input_hash_scalar`
2. `(A_sum * input_hash_Scalar) * b_scan`

In https://github.com/bitcoin-core/secp256k1/pull/1471/commits/d7250503aca9d0cfc0c2df12d3f98c02945bcf33 the sender only does one ECC multiplication by first doing the less expensive scalar multiplication of `a_sum * input_hash_scalar`. 

What if we had a single function that is used by both the sender and receiver, e.g.`silentpayments_create_shared_secret`, which takes the `input_hash_scalar` as an input and a private key (can be `b_scan` or `a_sum`) as inputs along with a public key (can be `A_sum` or `B_scan`). This function would multiply the private key and scalar and then do the ECDH step. We would then need *separate* function for the light client receiver only which takes `b_scan` and `A_tweaked` and does the ECDH step.

Just brainstorming, open to other suggestions! Also might be prematurely optimizing but from what I understand ECC Multiplication is expensive, so keeping it to one for both sender and receiver seems worth it even at this stage.

josibake: > Have to think more about it, open for all suggestions. Glad that the interface discussion is unleashed!

Yeah, what I suggested is pretty half baked! I'll also give this a more thorough think and share my thoughts. Happy to keep the discussion here (easier to reference things concretely), or we can move it to the linked issue if you prefer.

josibake:
> "invalid input data is passed", indicating that the user did something wrong (case 1 above) and "the individual input data passed is fine, but we still can't continue" (case 3 above)

This is a good point. In theory, a user could recover from case 1, which at that point I'd agree we need two error codes: one to indicate that the user needs to do something different _with_ the current transaction in order to proceed, and another to indicated the can't do anything with the current transaction and it needs to be skipped.

In practice, I don't think case 1 is likely to happen often since the inputs already exist in validated transactions, it's just a matter of correctly extracting them. That being said, I am slightly leaning towards two error codes but I'd also prefer to be consistent with the other modules if we need a tie breaker. I'm curious if there is a general pattern in the library that we can take inspiration from.

include/secp256k1_silentpayments.h
```diff
@@ -28,7 +28,41 @@ extern "C" {
...
+ *             n_plain_seckeys: the number of sender's non-taproot input private keys
+ *             taproot_seckeys: pointer to an array of 32-byte private keys of taproot inputs
+ *                              (can be NULL if no private keys of taproot inputs are used)
+ *           n_taproot_seckeys: the number of sender's taproot input private keys
+ *         outpoint_smallest36: serialized smallest outpoint
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_create_private_tweak_data(
+    const secp256k1_context *ctx,
+    unsigned char *private_tweak_data32,
+    const unsigned char *plain_seckeys,
```
josibake, line 61: Started trying to use the module in #28122 and noticed this is an array of concatenated bytes, where `n` determines the number of items in the "list." Any particular reason for choosing this method of passing lists of pubkey/seckeys? Is this a convention in `libsecp`?

josibake: > It seems a list of pointers (ending up in a double-pointer) makes more sense?

This makes more sense to me. If we have to concatenate the bytes for the secret keys, it seems like we would have to copy them out of secure memory just to pass them in to the function, whereas if we have a list of pointers the keys can be read directly from their secure memory.

> Should I change the interface to list of pointers? That seems strictly better than the current interface. Or is there another method of passing in key lists that you could think of?

list of pointers was going to be my suggestion :smile: 

> Oh btw, I found an old Bitcoin Core branch from December

Awesome, I'll take a look!

include/secp256k1_silentpayments.h
josibake: > One drawback could be though that it is likely more confusing for the user and it's a bit easier to use it wrong; since the paramters can't be named exactly after what is expected anymore

True, but as you said, I think we can address this with good documentation. Another for using the same routine for both sender and receiver is we ensure that sender and receiver will arrive at the same shared secret (provided they give correct inputs), since they are using the same routine. If we use separate routines, there is a small chance of introducing a bug in one of the routines that would cause the sender and receiver to arrive at different shared secrets despite giving the correct inputs.

> Even the shared secret creation for light clients (passing ) case could be done by that same routine, by making one parameter optional (i.e. if input_hash is NULL, that signals that the tweak is already included in the passed pubkey)

Also not a bad idea! I'd say we can go with this for now and always revert to multiple routines if there are objections.

> Those two calculations are different in the sense that the shared secret creation one does a full ECDH including the call of the ECDH hash function, while the other one is just a normal point multiplication

Good point, I forgot about that. I think the main difference here is that ECDH is done in constant time whereas point multiplication is not. Regardless, you're correct that these should remain separate routines and I agree we should just return the 33 byte serialized pubkey for the routine creating the light client/index data.

include/secp256k1_silentpayments.h
```diff
@@ -192,6 +192,35 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_create_a
...
+ *  counter k, and an optional label_tweak, calculate the corresponding
+ *  output public key:
+ *
+ *  B_m = B_spend + label_tweak * G
+ *  (if no label tweak is used, then B_m = B_spend)
+ *  P_output = B_m + hash(shared_secret || ser_32(k)) * G
+ *
+ *  Returns: 1 if outputs creation was successful. 0 if an error occured.
+ *  Args:                  ctx: pointer to a context object
+ *  Out:   output_xonly_pubkey: pointer to the resulting output x-only pubkey
```
josibake, line 207: in "silentpayments: implement output pubkey creation (both sender and receiver)" (https://github.com/bitcoin-core/secp256k1/pull/1471/commits/e8cb4e74ff586d332eb023e157b69e38660668ce):

I think we will need sender and receiver specific routines for creating outputs. For the sender, what you have here is perfect. For the receiver, [GetTxOutputTweaks](https://github.com/bitcoin/bitcoin/blob/de7dd3aaf64bc9812ac4c30994ee6fbb5695e953/src/wallet/silentpayments.cpp#L162) (from #28122) returns `t_k` instead of the full output, which allows the wallet to scan without ever needing the private key `b_spend`. The idea here is `t_k` can be stored in the wallet and then added to `b_spend` at signing time. Additionally, the receiver needs to perform the following steps when scanning for labels:

* calculate `tx_output - P_k`
* calculate `- tx_output - P_k`

What if we renamed `_create_output_pubkey` to `_sender_create_output_pubkey` and added a new function for the receiver, something like:

```
typedef struct {
    unsigned char t_k[32];
    secp256k1_xonly_pubkey xonly_output;
    secp256k1_pubkey label;
    secp256k1_pubkey label_negated;
} secp256k1_silentpayments_scanning_data;

int secp256k1_silentpayments_receiver_create_scanning_data(
    const secp256k1_context *ctx,
    secp256k1_silentpayments_scanning_data *scanning_data,
    const unsigned char *shared_secret33,
    const secp256k1_pubkey *receiver_spend_pubkey,
    unsigned int k
)
```

Where `_receiver_create_scanning_data` creates the output same as before and also calculates `label_pubkey = tx_output + P_k_negated` and `label_pubkey_negated = tx_output_negated + P_k_negated`.

This way, if `xonly_output` matches one of the transaction outputs, we add `t_k` to the wallet. If `xonly_output` doesn't match, we check if `label_pubkey` or `label_pubkey_negated` is in our labels cache and if there is a match we add `t_k + label_tweak` to the wallet.

If the scanning client is not using labels, they can just ignore the label data and move on if they don't find a match for `xonly_pubkey`. Creating the label data only involves a few negations and EC additions so I don't think it hurts to always calculate it. The alternative would be *even moar routines:* one for receiving clients that use labels and one for receiving clients that don't.

WDYT?

bip352-testsuite/run_bip352_tests.py
```diff
@@ -0,0 +1,227 @@
...
+
+        # scan through outputs
+        k = 0
+        outputs_to_check = receive_data_given['outputs'].copy()
+        while True:
+            if len(outputs_to_check) == 0:
+                break
+            for output_to_check in outputs_to_check:
+                found_sth = False
+                for label_m in [None] + receive_data_given['labels']:
```
josibake, line 159: in "[DEMO-ONLY] add bip352 test vectors running suite using python and ctypes" (https://github.com/bitcoin-core/secp256k1/pull/1471/commits/3deeebd1b5d7f3684e6425be749c643504f7c471):

This is *kinda* cheating in that you are iterating through all of your labels for every output in the transaction that you're scanning. So lets say I have 100k labels and I'm checking a tx with 100 outputs.. I'd doing `2 * 100k * 100` hashes, along with `100k * 100` EC additions, etc.

Since the receiver already has to generate the label to use it, the idea is they would store the label (something like `{(B_spend + label_tweak * G) : label_tweak}` so that when scanning they can calculate `label = tx_output - P_k` and `label = tx_output_negated - P_k`. This means that regardless of the number of labels used, they only have to do at most two look ups in their label cache when scanning.

josibake (commented):
Working on using this branch in [#28122](https://github.com/bitcoin/bitcoin/pull/28122) and had some thoughts regarding labels

include/secp256k1_silentpayments.h
josibake: I started POC branch to demonstrate: https://github.com/josibake/secp256k1/commit/fa3f887ab689b937aeb2ea28cf50b591d3673960

josibake: > senders can pass in NULL for t_k as they don't have any use for it

Nice, I like this. 

> only receivers with labels support have to process further

Small correction: per the BIP, it is recommended the receiver *always* check for the change label. That being said, I somewhat prefer what you're proposing here with a separate routine for generating the label data. Initially, my thinking was its better to keep it all in one routine because we avoid needing to save `P_output` as a `secp256k1_pubkey` only to load it back into a `secp256k1_ge/gej` group element for label processing, but I suspect this conversion is trivial enough.

josibake: Hm, I completely forgot about the parity of `P_output` being relevant. I'm not crazy about returning two encodings of the same public key, one of which is only relevant if you're the receiver and scanning for labels. This means the function now has an optional `t_k` parameter and an optional `P_output_pubkey` that are only used if you are a receiver, so it seems like it might be better to just have `_sender/_receiver` specific functions.

Apologies for the back and forth on this, but what do you think about the following:

For the sender, same as before:
```
int secp256k1_silentpayments_sender_create_output_pubkey(
    const secp256k1_context *ctx,
    secp256k1_xonly_pubkey *output_xonly_pubkey,
    const unsigned char *shared_secret33,
    const secp256k1_pubkey *receiver_spend_pubkey,
    unsigned int k
);
```
For the receiver:
```
typedef struct {
    unsigned char t_k[32];
    secp256k1_xonly_pubkey xonly_output;
} secp256k1_silentpayments_output_data;

typedef struct {
    secp256k1_pubkey label;
    secp256k1_pubkey label_negated;
} secp256k1_silentpayments_label_data;

int secp256k1_silentpayments_receiver_create_scanning_data(
    const secp256k1_context *ctx,
    secp256k1_silentpayments_output_data *output_data,
    secp256k1_silentpayments_label_data *label_data,
    const unsigned char *shared_secret33,
    const secp256k1_pubkey *receiver_spend_pubkey,
    unsigned int k
)
```

This way, the receiver can pass `NULL` for label_data if they don't support labels, otherwise we can compute the label data while we still have `P_output` as a group element and avoid all of the pubkey encoding woes.

josibake: Yeah, I was thinking about this today. One thought was have the `_create_scanning_data` check if `P_output == tx_output` before calculating the label data (since it has both), and then just skip creating the label data if they match.. but it felt a bit weird to have the routine doing checks for the caller. Curious what you think.

josibake:
I've rewritten [#28122](https://github.com/bitcoin/bitcoin/pull/28122) to use this module here: https://github.com/josibake/bitcoin/tree/implement-bip352-secp256k1-module. So far, so good! I'll take a fresh look here tomorrow with some feedback, and then update #28122 once I've cleaned up the branch a bit and we are ready to move this PR out of draft. @theStack can you remind me what your criteria for considering this PR "not a draft" were?

include/secp256k1_silentpayments.h
josibake: Hah, I don't know if there is a good answer here. Our options are:

* Don't do the check and always compute the label data - this means in the event of finding an output, we are doing a few unnecessary EC negations / additions and one of the labels we return is the point at infinity. Not too big a deal, since this would be unusable by the caller anyways
* Do the check to avoid computing label data when we don't need it - this means the caller will end up doing the check again
* Return a boolean (which is a separate in param, instead of the functions normal `int` return) which is checked by the caller. If false, look at the label data (if using)

Using the boolean approach seems like the most efficient of the options and avoids calculating / returning unnecessary data, at the expense of a slightly awkward API. Personally, I'm leaning toward performance given that scanning is a resource intensive operation.

josibake:
> @josibake: I've add the missing `_create_tweaked_pubkey` function to the API. As far as I'm aware, a user would have all the necessary routines now that are needed for a full silent payments implementation, without the need to do any manual ECC calculations (see also the summary table with formulas in the PR description). If there are no further objections, I'll move the PR out of draft within the next days, after increasing the test coverage a bit more.

Looks great! I've updated [#28122](https://github.com/bitcoin/bitcoin/pull/28122) to use this branch. I think we can take this out of draft whenever you're ready. While I'm sure the API will change with further review, this seems like a pretty solid starting point!

Sjors:
What is the input `k` in `_sender_create_output_pubkey`?

I find the label stuff fairly confusing, both in the BIP and in the description here. But no concrete suggestion on how to improve that...

josibake:
> What is the input `k` in `_sender_create_output_pubkey`?
 
`k` is a counter used by the receiver to create multiple outputs for the sender. Using the variable `k` here to match the notation in the BIP (see [BIP352#creating-outputs](https://github.com/bitcoin/bips/blob/73f1a52aafbc54a6ea2ce9a9e5edb20c24948b87/bip-0352.mediawiki#creating-outputs))

src/modules/silentpayments/Makefile.am.include
```diff
@@ -0,0 +1,2 @@
+include_HEADERS += include/secp256k1_silentpayments.h
+noinst_HEADERS += src/modules/silentpayments/main_impl.h
```
josibake, line 2: in "build: add skeleton for new silentpayments (BIP352) module" (https://github.com/bitcoin-core/secp256k1/pull/1471/commits/a9a5fe8e28f2f4dd53fdfb4ce797b23b4452f0a6):

Needed for Bitcoin Core's CI:
```suggestion
noinst_HEADERS += src/modules/silentpayments/tests_impl.h
noinst_HEADERS += src/modules/silentpayments/main_impl.h
```

include/secp256k1_silentpayments.h
```diff
@@ -104,6 +104,28 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_create_p
...
+ *
+ * The resulting data is useful for light clients and silent payment indexes.
+ *
+ *  Returns: 1 if tweaked public key creation was successful. 0 if an error occured.
+ *  Args:              ctx: pointer to a context object
+ *  Out:         A_tweaked: pointer to the resulting tweaked public key
+ *  In:              A_sum: pointer to the public keys sum
+ *              input_hash: pointer to the 32-byte input hash
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_create_tweaked_pubkey(
```
josibake, line 123: in " silentpayments: add tweaked pubkey creation routine (for light clients / sp index)" (https://github.com/bitcoin-core/secp256k1/pull/1471/commits/842e5bf427d63b86d55aef3c234c05b42c2a6ad7):

If we end up using the `_create_shared_secret_from_public_data` routine, we could also have this be `_create_tweaked_pubkey_from_public_data` where it takes all the same inputs as `_create_shared_secret_from_public_data` and returns `unsigned char *A_tweaked33`. This function would be used by anyone intending to store this data in an index or serve directly to a light client.

include/secp256k1_silentpayments.h
```diff
@@ -126,6 +127,50 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_create_t
...
+ *
+ *  Returns: 1 if shared secret creation was successful. 0 if an error occured.
+ *  Args:                  ctx: pointer to a context object
+ *  Out:       shared_secret33: pointer to the resulting 33-byte shared secret
+ *  In:       public_component: pointer to the public component
+ *           private_component: pointer to 32-byte private component
+ *                  input_hash: pointer to 32-byte input hash (can be NULL if the
+ *                              public component is already tweaked with the input hash)
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_create_shared_secret(
```
josibake, line 166: in "silentpayments: add shared secret creation routine (a*B == A*b)" (https://github.com/bitcoin-core/secp256k1/pull/1471/commits/b0e37968b09b830b422bafb60c0e870d611ef6ed):

Building off the last couple suggestions, this function is now exclusively for light clients and there is no ambiguity about the arguments:

* `public_component` is always `A_tweaked`
* `private_component` is always `b_scan`

we can also drop the `input_hash` argument and change `public_component` to `unsigned char *A_tweaked`

include/secp256k1_silentpayments.h
```diff
@@ -171,6 +171,28 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_create_s
...
+ *
+ *  Returns: 1 if label tweak and label creation was successful. 0 if an error occured.
+ *  Args:                  ctx: pointer to a context object
+ *  Out:           label_tweak: pointer to the resulting label tweak
+ *   In:  receiver_scan_seckey: pointer to the receiver's scan private key
+ *                           m: label integer (0 is used for change outputs)
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_create_label_tweak(
+    const secp256k1_context *ctx,
+    secp256k1_pubkey *label,
```
josibake, line 190: in "silentpayments: add label tweak calculation routine" (https://github.com/bitcoin-core/secp256k1/pull/1471/commits/2a00e12e58b6afd2b9ea9b3999dedde91f8ff82f):

I was trying to see if we could completely get rid of `secp256k1_pubkey` as a return value, but here it makes sense since this output is likely to be used in the next commit to create the labeled spend pubkey.

Alternatively, we could combine this and the next function into one that takes `scan_seckey`, `spend_pubkey`, and `m` as arguments and returns `label_tweak`, `label33`, `labeled_spend_pubkey33` as arguments. Not convinced this is strictly better, but curious to hear your thoughts.

include/secp256k1_silentpayments.h
```diff
@@ -216,6 +216,30 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_create_a
...
+ *                           or receiver's perspective with routines from above
+ *    receiver_spend_pubkey: pointer to the receiver's spend pubkey
+ *                        k: output counter (usually set to 0, should be increased for
+ *                           every additional output to the same recipient)
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_sender_create_output_pubkey(
+    const secp256k1_context *ctx,
+    secp256k1_xonly_pubkey *P_output_xonly,
+    const unsigned char *shared_secret33,
+    const secp256k1_pubkey *receiver_spend_pubkey,
```
josibake, line 239: in "silentpayments: implement output pubkey creation (for sender)" (https://github.com/bitcoin-core/secp256k1/pull/1471/commits/8460be58ccad9ca98e27bc60e838633f8abe2ed3):

I think the sender will always be getting the receiver spend key externally, so we could just have this be `unsigned char *receiver_spend_pubkey` and handle the conversion to `secp256k1_pubkey` for the caller.

include/secp256k1_silentpayments.h
```diff
@@ -289,6 +289,30 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_receiver
...
+ *                         t_k: pointer to the 32-byte output tweak
+ *                 label_tweak: pointer to an optional 32-byte label tweak
+ *                              (can be NULL if no label is used)
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_create_output_seckey(
+    const secp256k1_context *ctx,
+    unsigned char *output_seckey,
+    const unsigned char *receiver_spend_seckey,
+    const unsigned char *t_k,
+    const unsigned char *label_tweak
```
josibake, line 313: in "silentpayments: implement output spending privkey creation (for receiver)" (https://github.com/bitcoin-core/secp256k1/pull/1471/commits/26bdb5f195ea5e9ebcd96919709c0a8f50156f80#):

In [#28122](https://github.com/bitcoin/bitcoin/pull/28122) I ended up abusing this function to first combine `t_k` and `label_tweak` into a single tweak, and then later called it again to combine `receiver_spend_seckey` with my combined tweak for the final key.

One option is keep this function as is and then we store the `t_k` and the `label_tweak` (presumably already stored since we the curve point for scanning) and use them both at signing time. The other option would be to add another routine for combining the tweaks. The only advantage I can think of here is for HWW, where it might be more convenient to pass the combined tweak (32 bytes) instead of `t_k` and `label` (64 bytes) when signing. In all other cases, the receiver would have already calculated `label` or can cheaply recalculate it with `scan_seckey` and `m`, so leaning toward keeping this as is.

josibake (commented):
I know I just said the API looks good, but here's a bunch of API feedback :see_no_evil:

include/secp256k1_silentpayments.h
```diff
@@ -28,7 +28,44 @@ extern "C" {
...
+ * input_hash = hash(outpoint_smallest || (a_sum * G))
+ *
+ * If necessary, the private keys are negated to enforce the right y-parity.
+ * For that reason, the private keys have to be passed in via two different parameter
+ * pairs, depending on whether they were used for creating taproot outputs or not.
+ * The resulting data is needed to create a shared secret for the sender side.
+ *
+ *  Returns: 1 if shared secret creation was successful. 0 if an error occured.
+ *  Args:                  ctx: pointer to a context object
+ *  Out:                 a_sum: pointer to the resulting 32-byte private key sum
```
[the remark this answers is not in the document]
josibake: Great point. This could be a fairly common usecase (e.g. an exchange processing withdrawals to $n$ silent payment addresses), but it is still a bounded problem in that you can't send to more than $N$ silent payment addresses in a single transaction, where $N$ is the number of outputs you can fit in a single block.

I'm leaning toward making this so the caller doesn't need to handle intermediary secret data. The other option is we expose both routines in the API?

include/secp256k1_silentpayments.h
```diff
@@ -67,7 +67,42 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_create_p
...
+ *                  input_hash: pointer to the resulting 32-byte input hash
+ *  In:          plain_pubkeys: pointer to an array of pointers to non-taproot
+ *                              public keys (can be NULL if no non-taproot inputs are used)
+ *             n_plain_pubkeys: the number of non-taproot input public keys
+ *               xonly_pubkeys: pointer to an array of pointers to taproot x-only
+ *                              public keys (can be NULL if no taproot inputs are used)
+ *             n_xonly_pubkeys: the number of taproot input public keys
+ *         outpoint_smallest36: serialized smallest outpoint
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_create_public_tweak_data(
```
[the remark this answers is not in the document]
josibake: Also worth pointing out this is an unbounded problem (unlike the sending case): I could be scanning for 30M scan keys for a single transaction. In this case, we would want to avoid recalculating `A_sum` and the `inputs_hash` 30M times.

include/secp256k1_silentpayments.h
josibake: > Are there use-cases for creating labelled addresses given a labelled pubkey without having access to private key data

None that I can think of, but as I thought about this more I realized:

* Its likely a scanner might want to generate a labels cache when recovering a wallet without needing the `labeled_receiver_pubkey`
* A receiver might want to use a custom label scheme (e.g. hashing a customerId for a static deposit address)

For both of those cases, it seems having the two separate routines is best.

real-or-random:
> I ended up abusing this function [secp256k1_silentpayments_create_output_seckey]

Yeah, this is basically `scalar_add`, with some high-level docs. 

This is something we need to think about. A goal of the library is to provide safe high-level APIs. I guess we often assume that this is equivalent to avoiding access to low-level operations in the API. But as this functions shows, it's not equivalent. This function is part of a high-level API, but it happens to expose a simple low-level operation... Now, there's nothing wrong with this per se, but people will abuse this for all kinds of things.

And let's be honest. If you need a `scalar_add`, and the library gives you a (disguised) `scalar_add`, it's just pragmatic to use it. The API docs won't prevent anyone from doing so. Is it their responsibility? Probably yes.

---

I think the entire module, as proposed, is a bit unusual for libsecp256k1 because it only wraps some ECC operations of a higher-level protocol (as explained in the header). This design makes sense, but it's different from what we usually have in libsecp256k1. 

You could say that BIP340 signatures are also just some ECC parts of a higher-level protocol "Bitcoin", but the difference is that they implement a common interface of a signature scheme, what people call a cryptographic primitive (like public-key encryption, key exchange, hashing, etc.). But this here is merely the "cryptographic core of silent payments", and it's tightly coupled to it.

Again, there's per se nothing wrong with a more low-level module, but I think we'll need to spend some thoughts on how we make it fit the library best. Sometimes the potential for abuse can be reduced by having opaque data types that you can only pass from API function to API function.  We do something similar in the [proposed musig2 API](https://github.com/bitcoin-core/secp256k1/pull/1479) for example. I think this would be closest to the current spirit of the library.

Another design pattern is to have "stage" functions that perform all the different computation that should be done at a specific stage of the protocol. This doesn't seem elegant, but it prevents mistakes where the caller combines the API functions wrongly, e.g., calls them in the wrong order.

---

Off-topic ramblings about the library: 

So if we expose the `silentpayments_create_output_seckey` function, we might as well go ahead and expose `scalar_add`  and possibly some more scalar and group functions as part of a mid-level, or "hazmat API" as it's called in some libraries (ignoring for a moment that this is work and raises more questions). We anyway have [ec_pubkey_combine](https://github.com/bitcoin-core/secp256k1/blob/427e86b9edcaa5cb19e71233ad93c2ec60dff10f/include/secp256k1.h#L864-L878) left as a relic. Then the caller at least won't have to abuse the silent payment API.

And if we had hazmat API, then one could alternatively simply build some libsilentpayments around it, and other applications could do the same. This would certainly help some applications. But it's not obvious that this is the ideal solution. There are multiple related but different questions, e.g., in should stuff be a module, in which repo should it live, who should maintain it, how should it be linked, etc... (And lol, please don't be discouraged by my comments here. I see value in this as a libsecp256k1 module, it's just that the PR makes me think about the bigger picture again...)

josibake:
Thanks for chiming in @real-or-random , this is exactly the kind of review we are looking for at this stage! As much as possible, we'd like to keep this module in the spirit of the library. I'll take a look at the Musig2 API for inspiration and also spend some more time thinking about the "stages" concept. That feels somewhat closer to what we have now.

> And if we had hazmat API, then one could alternatively simply build some libsilentpayments around it

This also makes a lot of sense to me as an approach. For example, I've been chatting with the developers of [rust-silentpayments](https://github.com/cygnet3/rust-silentpayments) about how their library might evolve if there is a `libsecp256k1` silentpayments module. Seems like having a rust silent payments library that consumes a "hazmat" secp256k1 API could be a reasonable solution?

> And lol, please don't be discouraged by my comments here

The only thing that discourages me is no comments at all :smile:

josibake:
@real-or-random spent some time digesting comment and wanted to summarize some thoughts: 

> I think the entire module, as proposed, is a bit unusual for libsecp256k1 because it only wraps some ECC operations of a higher-level protocol (as explained in the header). This design makes sense, but it's different from what we usually have in libsecp256k1.

In hindsight, I think BIP352 could (should) have been three BIPS: 

* BIP-silent-payments-non-interactive-multikey-ecdh (SPNIMKECDH)
* BIP-silent-payments-v0 (the part that defines which inputs to use)
* BIP-silent-payments-address-encoding

This way, if we need change how the inputs are chosen in the future, we'd write a small BIP-silent-payments-v1, and nothing about the SPNIMKECDH part would need to change. Conceptually, when talking about a libsecp256k1 module I'm only talking about SPNIMKECDH (i.e. we wouldn't expect a libsecp256k1module to filter transaction inputs to the silent-payments-v0 eligible inputs before performing the SPNIMKECDH protocol). If we start from this assumption and forget about the labels part of silent payments for a second, we could build a high-level API like so:

### Sender 
```
typedef struct {
    secp256k1_pubkey recipient_scan_pubkey,
    secp256k1_pubkey recipient_spend_pubkey,
    size_t n_desired_outputs
} secp256k1_silentpayments_recipient;

/**
 * Takes plain and taproot seckeys as inputs, and a list of the desired recipients
 * and returns a list of generated xonly outputs.
 * In this example, a recipient is a scan/spend pubkey pair along with an int to indicate
 * how many outputs should be created for this particular recipient
 * /
int secp256k1_silentpayments_sender_create_outputs(
    const secp256k1_context *ctx,
    secp256k1_xonly_pubkeys **outputs,
    const unsigned char * const *plain_seckeys,
    size_t n_plain_seckeys,
    const unsigned char * const *taproot_seckeys,
    size_t n_taproot_seckeys,
    const secp256k1_silentpayments_recipient * const *recipients,
    size_t n_recipients,
    const unsigned char *outpoint_smallest36
)
```
### Receiver
```
typedef struct {
    secp256k1_xonly_pubkey pubkey,
    unsigned char output_tweak[32],
} secp256k1_silentpayments_output;

/**
 * Takes tx input pubkeys, the receivers spend pubkey and scan private key, and tx outputs as inputs
 * and returns a list of found outputs, where the found output consists of the xonly pubkey and the tweak data
 * needed to spend it
 * /
int secp256k1_silentpayments_receiver_scan_outputs(
    const secp256k1_context *ctx,
    secp256k1_silentpayments_output **found_outputs,
    const secp256k1_pubkey **input_pubkeys,
    size_t n_input_pubkeys,
    const secp256k1_pubkey *receiver_spend_pubkey,
    const unsigned char *receiver_scan_seckey,
    const secp256k1_xonly_pubkey * const *tx_outputs,
    size_t n_tx_outputs
)
/** 
 * Takes tx input pubkeys and the smallest outpoint and returns a pubkey
 * that can be stored and distributed to clients who need to scan but dont have
 * access to block data/tx data (index for wallet rescanning, light clients, etc).
 * /
int secp256k1_silentpayments_receiver_create_index_data(
    const secp256k1_context *ctx,
    secp256k1_pubkey *A_sum_tweaked,
    const secp256k1_pubkey **input_pubkeys,
    size_t n_input_pubkeys,
    const unsigned char *outpoint_smallest36
)
/**
 * takes the index data (A_sum*input_hash), tx outputs, receivers spend pubkey and scan priv key
 * and returns found outputs. Intended for light clients/wallet rescanning.
 * /
int secp256k1_silentpayments_scan_from_index_data(
    const secp256k1_context *ctx,
    secp256k1_silentpayments_outputs **found_outputs,
    const secp256k1_pubkey *receiver_spend_pubkey,
    const unsigned char *receiver_scan_seckey,
    const secp256k1_pubkey *A_sum_tweaked,
    const secp256k1_xonly_pubkey *tx_outputs,
)
/** 
 * produces a signature for a silent payments taproot output
 * this function is just a wrapper function around schnorrsign which first
 * tweaks the receivers spend key with the tweak data found during scanning.
 * this way, we avoid needing to expose any `scalar_add` functions to the caller.
 * /
int secp256k1_silentpayments_sign32(
    const secp256k1_context *ctx,
    unsigned char *sig64,
    const unsigned char *msg32,
    const secp256k1_keypair *keypair,
    const unsigned char *aux_rand32
    const unsigned char *tweak32
)
```
This is more of an example than a proposal to demonstrate a few simple function calls and not exposing things that could be abused or used unsafely. The main question here is "would something like this fit as a module for libsecp256k1," considering its not _really_ a generic cryptographic primitive but also not _quite_ the a full silent payments v0 implementation in that the sender and receiver still need to filter the inputs. If the answer is yes, I think we can start from here and figure out how to add labels in sane way (I have a few ideas for labels, both with opaque data types or stages). If the answer is no, then I think looking into other approaches, like a hazmat API is better.

I also spent some time thinking about whether or not this could be a more generic "non-interactive-multikey-ecdh" module that is used by a `libsilentpayments` or something, but this didn't seem super promising. Mainly because it wasn't clear that the hashing with a counter `k` would be generically useful, and without that we'd need to return an unhashed ECDH result. Also similar questions about whether the spend and scan key are generically useful or silent payments specific. Furthermore, we'd still need to do low-level operations even after the ECDH step (adding the spend pubkey, scanning with labels, etc), so it felt like just kicking the can down the road a bit.
<hr/>

@theStack 

> To my understanding, secp256k1 as of now provides all the necessary routines for a full BIP352 implementation (Bitcoin Core PR https://github.com/bitcoin/bitcoin/pull/28122 followed that approach before it used this module). The two main reasons for a secp256k1 module (or library using secp256k1, if the libsilentpayments approach is followed) would be AFAICT 1) efficiency and 2) hiding cryptographic complexity from the user.

I think this was more an example of "could" rather than "should." From my understanding, bitcoin core is also a bit of a "privileged" user of secp256k1 and the goal of a module would be something generally useful for wallets outside of bitcoin core. Regarding efficiency, I'd be very surprised if this approach is _less_ efficient than what we were originally doing in core, although IIRC bitcoin core is faster when it comes to hashing? The main advantages of doing it all here w.r.t efficiency is not need to serialize/deserialize all the time from compressed keys to group elements. Regarding benchmarking, I do still have the old version not using this module, if you want to compare with that! Would be nice to see some numbers!

josibake:
@theStack I started refactoring the API based on https://github.com/bitcoin-core/secp256k1/pull/1471#issuecomment-1984303397. I've only completed the sender side, but will continue working on the receiver side.

The main changes are:

1. Provide a single function for the sender: this means the sender doesn't need to worry about grouping addresses and ensuring they pick the right values for `k`
    * Requires [heapsort (from the musig2 PR)](https://github.com/bitcoin-core/secp256k1/blob/3b6c90a93fb55b2ea5fc629762bf4c3c569211cf/src/hsort_impl.h)
    * Removes the `create_private_tweak_data` routine
2. Remove the `tweaked_pubkey` routine (for light clients): instead, if the caller does not pass a pointer for `input_hash`, then include `input_hash` with the returned `A_sum`

The branch is still a bit rough so some of the comments / commit messages are likely correct. Mostly just looking for conceptual feedback and curious if you have any objections to this approach. The branch is https://github.com/josibake/secp256k1/tree/bip352-api-refactor . I've tried to keep the edits in the relevant commits so that if the changes make sense we can easily incorporate them into your branch.

josibake:
> A single-function API for sending that takes care of everything (even the scan-pubkey grouping, sort ftw!) seems a good idea to me.

yeah, the more I thought about it, we are already requiring the user to pass all of the keys at once so it didn't seem like we are gaining anything with separate function calls. In the future, it might make sense to have functions to enable sending where the sender does not have access to all of the private keys at once, but that will require a lot more thought and seems out of scope for now.

> the user doesn't have a way to determine which of the created outputs belong to which recipient

Good catch, I overlooked this! I took your suggestion of extending the recipient struct and updated the branch. This has the added benefit of simplifying the function signature (i.e. no longer need to pass arguments for `outputs` and `n_outputs`).

---
Thinking about the receiver API a bit more, I realized we can't assume the receiver has access to the `tx_outputs` when scanning. For example, if the client is using BIP158 block filters, they would generate the scriptPubKey and then check if it is in the filter before downloading the block, or in the case of an electrum style client, they might generate the taproot output and then call the electrum backend to see if the scriptPubKey exists in the UTXO set.

Given this, what do you think about using a callback function for checking if the generated output exists / checking if the label exists? It seems `libsecp256k1` does use this pattern already in the case of the ECDH module where the caller can provide their own hash function and using callbacks could allow for a much simpler, easy to use API but I'm not what the cost in performance would be or if this would work with language bindings (e.g. a javascript library using `libsecp256k1`). Curious to hear your thoughts.

jonasnick (commented):
@josibake I had a quick look at the API in your branch, a few minor comments:
- The In/Out parameter `secp256k1_silentpayments_recipient` seems a bit strange. Why not just group by index? `secp256k1_silentpayments_recipient` could just contain the scan and spend pks, and the i-th generated output would correspond to the i-th provided `secp256k1_silentpayments_recipient`?
- Instead of taking a `const unsigned char * const *taproot_seckeys` argument and multiplying each seckey with base point G to see if it needs to be negated, you could accept an array of `secp256k1_xonly_keypair`, which holds both the seckey and the pubkey. Often the caller will already have the pubkey and would, if the function wouldn't accept `xonly_keypair`, do a costly recomputation of the public key unnecessarily. If the caller doesn't have an `xonly_keypair`, they can just create one.
- I can imagine that others disagree, but I wouldn't add another `sign32` function if the spend key can just be tweaked via some `keypair_tweak` function and then given to the regular `schnorrsig_sign32` function. Then this would be the same workflow for a silent payments derived key to sign via `schnorrsig_custom`. And a similar workflow for the musig module (where tweaking happens via `secp256k1_musig_pubkey_tweak_add`.

josibake:
Thanks for the review, @jonasnick !

>     * The In/Out parameter `secp256k1_silentpayments_recipient` seems a bit strange. Why not just group by index? `secp256k1_silentpayments_recipient` could just contain the scan and spend pks, and the i-th generated output would correspond to the i-th provided `secp256k1_silentpayments_recipient`?

IIUC, this means bringing back the out param `taproot_outputs` and `n_taproot_outputs` (or just reuse `n_recipients`), correct? We could use the index in the `taproot_outputs` array to match to the recipient (this is because the heapsort does the sort in place, so the recipient order after sorting would match the order in the `taproot_outputs` array, right?). This seems more complex for the caller in that the need to initialize the taproot output array and match up the indices and I don't see any clear advantage, but perhaps I'm missing something? Also fine to go with the proposed change even if the reason is just that this is a more common pattern

>     * Instead of taking a `const unsigned char * const *taproot_seckeys` argument and multiplying each seckey with base point G to see if it needs to be negated, you could accept an array of `secp256k1_xonly_keypair`, which holds both the seckey and the pubkey. Often the caller will already have the pubkey and would, if the function wouldn't accept `xonly_keypair`, do a costly recomputation of the public key unnecessarily. If the caller doesn't have an `xonly_keypair`, they can just create one.

Great point, this was on my list of things to change but I forgot to include it!

>     * I can imagine that others disagree, but I wouldn't add another `sign32` function if the spend key can just be tweaked via some `keypair_tweak` function and then given to the regular `schnorrsig_sign32` function. Then this would be the same workflow for a silent payments derived key to sign via `schnorrsig_custom`. And a similar workflow for the musig module (where tweaking happens via `secp256k1_musig_pubkey_tweak_add`.

I agree. The silent payments protocol doesn't really have anything to do with signing because once the spend key is tweaked, it is just a regular taproot output. My initial thinking was not wanting to expose any low level tweaking functions in the module, but based on the recent IRC discussion regarding a hazmat API, this seems less of a concern.

josibake:
> So it seems like the purpose of sorting is to make sure that any two tuples (B_scan, B_m), (B_scan', B_m') with B_scan = B_scan' are next to each other in the array. 

Yep! This is because the counter `k` is incremented over a single `B_scan` for each value of `B_m` (included repeated values of `B_m`. As you mentioned, if the sender does not do this grouping correctly the recipient will be unable to find the outputs, so I think we will always need the sorting step. My thinking was if we are going to do the sorting regardless, it seemed better to abstract the grouping step away from the caller completely.

> 1. Change the API as I had suggested (remove generated_output from recipient, change recipient to be an input-only parameter, and add taproot_outputs and n_taproot_outputs as output parameters).
> 2. Sort inside the function to make sure that grouping was done correctly. If it wasn't, return 0.


IIUC, heapsort is unstable which means if the caller groups and passes in the tuples `[(B_scan, B_m), (B_scan, B_m')...]`, we might end up reordering them to `[(B_scan, B_m'), (B_scan, B_m) ...]`. If the caller meant to send 1 BTC to `(B_scan, B_m)` and 2 BTC to `(B_scan, B_m')`, it seems there's a risk the values might get switched using the index approach. So I think we'd have to also reconsider how we are verifying the grouping (e.g. use a different sorting algorithm) if we wanted to go with the approach you're suggesting.

josibake:
@theStack I pushed a refactor of the recipient API to https://github.com/josibake/secp256k1/tree/bip352-api-refactor, let me know what you think. Most notably, it allows for the receiver to scan the transaction in a single call (vs scanning each output individually and managing values of `k`). This simplifies things for the receiver and I think gets us a bit more speed by reducing some overhead related to serializing/deserializing pubkeys.

I also removed the `_seckey` function: using the `label_lookup` callback, I am able to get the label tweak and apply it directly while scanning, so the end result is one single `sp_tweak`. Given that, it seems we can just use the `_ec_seckey_add` function directly before signing.

For clients that do not have access to the transaction outputs (e.g. a BIP158 client), they are able to scan using the `create_public_tweak_data`, `create_shared_secret`, and `create_output` functions. They would do so by creating an output, checking if it exists in the UTXO set, and proceeding to create the next output if a match is found. Once they have all of the outputs, they can get the necessary transaction data via their preferred source (e.g. downloading the whole block, downloading the transaction, downloading the outpoints, etc). Another option would be to create a single output, check if it exits in a block, download the full block if it does and continue scanning with the `_receiver_scan_outputs` function since they would now have all of the transaction outputs.

---

@jonasnick (since you've been looking at the API), one thing that's kinda gross about this approach is the `found_objects` struct. I couldn't really think of a better way to do this since we don't know how many `found_objects` there will be before scanning and I wanted to avoid having the scanner go over each output individually (the old approach). Curious if you have any ideas on how we could improve this.

include/secp256k1_silentpayments.h
```diff
@@ -0,0 +1,320 @@
...
+ *  Given a shared_secret, a recipient's spend public key B_spend, and
+ *  an output counter k, calculate the corresponding output public key:
+ *
+ *  P_output_xonly = B_spend + hash(shared_secret || ser_32(k)) * G
+ *
+ *  Returns: 1 if output creation was successful. 0 if an error occured.
+ *  Args:               ctx: pointer to a context object
+ *  Out:     P_output_xonly: pointer to the resulting output x-only pubkey
+ *  In:     shared_secret33: shared secret, derived from either sender's
+ *                           or receiver's perspective with routines from above
```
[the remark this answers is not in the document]
josibake: I removed `_sender_` from the name in my refactor branch since this function can be used by the sender when creating outputs and also by the receiver when scanning without access to the transaction outputs (e.g. light clients).

jonasnick:
@josibake

> My thinking was if we are going to do the sorting regardless, it seemed better to abstract the grouping step away from the caller completely.

It's not abstracted away completely though because, as far as I understand, the caller needs to search through the In/Out recipients array to find the right `(B_scan, B_m)` tuple.

Anyway, my suggestion is flawed because it would actually require the caller to sort instead of just grouping the inputs recipients correctly.

> @jonasnick (since you've been looking at the API), one thing that's kinda gross about this approach is the found_objects struct.

You mean `found_outputs`? Unfortunately I don't have an idea right now for a cleaner approach.

Speaking of the API, what's the reason for exposing create_shared_secret to the user instead of just calling it internally in `sender_create_outputs` and `receiver_scan_outputs`?

josibake:
> the caller needs to search through the In/Out recipients array to find the right (B_scan, B_m) tuple.

You mean to match up the original `sp1qxxx: amount` address to the correct generated output? Good point, which is where being able to use the index from the original array would be nice. Will give this some thought. When I say "abstracted" from the caller, I was more referring to the caller not needing to worry about ordering the recipients in any particular way before calling the function.

> Speaking of the API, what's the reason for exposing create_shared_secret to the user instead of just calling it internally in sender_create_outputs and receiver_scan_outputs?

For the sender, no reason. `create_shared_secret` and `create_output_pubkey` are called internally in `sender_create_outputs`. For the receiver, there are three clients:

* Index: indexes transactions for wallet rescanning and for serving light clients (uses `create_public_tweak_data`)
* Fullnode: scans the transaction using `receiver_scan_outputs` or is doing a wallet rescan using the index data
  * A Fullnode might also be scanning for multiple scan secret keys, which is why it made sense to me to separate `create_shared_secret` from `receiver_scan_outputs`. This would allow a node to first do the ECDH step (creating multiple shared secrets) and then scan the transaction with each generated shared secret. You could also have the scan seckeys not on the scanning application and instead in an HSM or some separate application
* Light clients: likely does not have access to the transaction, so generates an output using `create_output_pubkey` and checks if it exists in the UTXO set
  * In this model, it makes sense to separate `create_shared_secret` and `create_output_pubkey`, since create shared secret is a fairly expensive operation, so we'd want a light client to be able to reuse the `shared_secret` when creating multiple outputs
 
That being said, open to suggestions on how we could lock down the API even further!

josibake:
> Currently, it seems that label is not set at all, I assume the idea is to set it within the if-branch that follows the label_lookup callback invocation

Oops! That was an oversight. It would only be set if the output found contained a label, not set in the case of a direct match. The idea was to make it easy for the caller to match up the output to the correct label, but something else I considered is making that the responsibility of the callback function.

> From an organizational point of view, do you think it makes sense to open a new PR for the refactored branch soon

I’ll defer to the libsecp maintainers on what they prefer. For me, I don’t mind merging the branch into this PR and updating the description, but I’m also happy to open an entirely new PR. Whichever is easiest!

Will read the rest of your comment more carefully and respond later.

josibake:
> One alternative approach could be to also pass an in-out parameter here with a list of struct instances that contains both the tx-output to scan for and the found results (label + tweak), together with a boolean flag found

I think this makes more sense than what I have now. Currently, the caller already needs to allocate `found_outputs` to be the same size as `tx_outputs` (in the event every output is a payment to the recipient). Seems easy enough to have them iterate through the list again after scanning is finished and grab all the outputs for them (by checking the boolean). They would have the scriptPubKey, the tweak, and the label if used all in one place.

josibake:
@jonasnick 

> EDIT: Actually, I think there's a case to be made that the caller organizing the recipient array correctly before calling the function (i.e., my suggested alternative) is cleaner than having the caller scan through the output of the function to find the correct scanning key and the corresponding generated_outputs.

I'm starting to agree, my only concern is I'd still want the sending function to verify that the addresses are in the correct groupings and fail otherwise. Sorting seemed like the easiest way to do this, but another idea is something like the following:

```
    if (n_recipients < 2) return true;
    for (i = 1; i < n_recipeints; i++) {
        if (secp256k1_ec_pubkey_cmp(ctx, recipients[i-1].scan_pubkey, recipients[i].scan_pubkey) != 0) {
            for (j = i + 1; j < size; j++) {
                if (secp256k1_ec_pubkey_cmp(ctx, recipients[i-1].scan_pubkey, recipients[j].scan_pubkey) == 0) {
                    return 0;
                }
            }
        }
    }
    return 1;
```

Essentially, anytime there is a scan key change, check the remainder of the list to see if the last seen scan key shows up again. This wouldn't change ordering of the list, but the downside is that the check is `O(n^2)` in the worst case. But in the context of recipient addresses we _kinda_ know the upper bound in that you can't really request more outputs than you can fit in a single transaction (roughly 9k outputs, see https://bitcoin.stackexchange.com/questions/116958/what-is-the-maximum-number-of-taproot-transactions-that-can-be-mined-in-a-single), so maybe `O(n^2)` is okay? In the context of a signing device, probably not, but just spitballing some ideas.

jonasnick (commented):
@josibake

> You mean to match up the original sp1qxxx: amount address to the correct generated output?

Yes.

> This wouldn't change ordering of the list, but the downside is that the check is O(n^2) in the worst case.

I would try to avoid adding a function with O(n^2) complexity.

Here's yet another alternative, which adds an index field to the recipient list, sorts the recipients inside the `sender_create_outputs` but returns the `generated_output` at the "right" position:

```
/* In param */
typedef struct {
    secp256k1_pubkey scan_pubkey;
    secp256k1_pubkey spend_pubkey;
    size_t idx;
} secp256k1_silentpayments_recipient;

int secp256k1_silentpayments_sender_create_outputs(secp256k1_xonly_pubkey *generated_output, secp256k1_silentpayments_recipient **recipients) {
  /* sort based on (scan_pubkey, spend_pubkey) */
  recipients = sort(recipients)
  for rec in recipient {
    generated_output[rec->idx] = create_output(...)
  }
}
```

This modifies the recipient array which may still be annoying for the caller. Of course, the caller can just copy the array to avoid that. Or we could sort it back based on the `idx` field, but that seems to be overkill. Would the Bitcoin Core implementation require the order of the recipient array unmodified?

The caller would be responsible for setting idx to the right value, but this could be easily checked in `sender_create_outputs`.

@josibake @theStack

> From an organizational point of view, do you think it makes sense to open a new PR for the refactored branch soon?
> I’ll defer to the libsecp maintainers on what they prefer.

I have a slight preference for opening a new PR since josi's branch is more a  rewrite than a refactor and a lot of the discussion in this PR doesn't apply.

jonasnick:
@josibake

> For the receiver, there are three clients:

> This would allow a node to first do the ECDH step (creating multiple shared secrets) and then scan the transaction with each generated shared secret.

I don't see advantage of this. Wouldn't it be an equal amount of work to scan the transaction with each scan secret key and generate the shared secret internally?

> You could also have the scan seckeys not on the scanning application and instead in an HSM or some separate application

If you can run create_shared_secret on the HSM, why couldn't you run receiver_scan_outputs?

> In this model, it makes sense to separate create_shared_secret and create_output_pubkey

I see. Or have `create_output_pubkeys` function that derives the shared_secret and creates one or more pubkeys.

josibake:
> Wouldn't it be an equal amount of work to scan the transaction with each scan secret key and generate the shared secret internally?

I misspoke, it’s not the shared secret part, but the summing of the input public keys that would duplicated work when scanning with multiple scan keys. So it seems we could have a `receiver_scan_outputs` function that takes the summed input pubkeys as an input, a scan secret key, and an optional `smallest_outpoint` (null if the input hash is already included with the summed input pubkeys) and have the share secret be created internally.

Regarding an HSM, I suppose there’s nothing stopping it from taking everything it needs to scan the transaction since the data needed will always be less than the maximum transaction size.

> Or have create_output_pubkeys function that derives the shared_secret and creates one or more pubkeys.

I think for this to work the client would need to know the total number of outputs in the transaction to avoid needing to redo ECDH in the event they don’t generate enough pubkeys the first time. However, if we assume the client knows the total number of outputs, we could make this function more closely match the `receiver_scan_outputs` function by using a `output_lookup` callback (instead of a label lookup). This way we could completely remove `create_shared_secret` from the API

(apologies for any typos, doing this on a phone)

josibake:
> I have a slight preference for opening a new PR since josi's branch is more a rewrite than a refactor and a lot of the discussion in this PR doesn't apply.

I've opened #1519 as a new PR, which attempts to incorporate all of the outstanding feedback here and includes a few new ideas.

PR #1473, opened 2024-01-02
Fix typos

shuoer86:
Fix some typos caught by spell checker

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,7 +44,7 @@ The Contributor Workflow & Peer Review in libsecp256k1 are similar to Bitcoin Co
 
 In addition, libsecp256k1 tries to maintain the following coding conventions:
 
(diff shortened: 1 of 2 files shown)


include/secp256k1.h
```diff
@@ -553,7 +553,7 @@ SECP256K1_API int secp256k1_ecdsa_signature_serialize_compact(
 /** Verify an ECDSA signature.
  *
  *  Returns: 1: correct signature
- *           0: incorrect or unparseable signature
+ *           0: incorrect or unparsable signature
```
real-or-random, line 556: That one is correct, see https://en.wiktionary.org/wiki/unparseable

shuoer86: @real-or-random thanks for your review. I haved reverted it.

real-or-random (approved):
utACK d77170a88d0d6f27d0b90057fa7cd25ec74e3850

PR #1476, opened 2024-01-04
include: make docs more consistent

jonasnick:


real-or-random:
We often say "a pointer" (instead of just "pointer") in other places, e.g., all other arguments of `secp256k1_ec_pubkey_serialize`. I mean, it doesn't matter too much, but if the point of the PR is to get it consistent, perhaps we should also make this part consistent.

real-or-random:
`git grep -E ":[ ]+a pointer"` shows three more occurrences. (Well, wo of them are in `contrib/lax_der_parsing.h`... Since we anyway converged from the version of this file in Core, I'd replace them too, but either way is good, I guess.) 

ACK mod this nit

real-or-random (approved):
ACK da7bc1b803b14274bc1687514e5da6a3e1cd9765

PR #1481, opened 2024-01-11
tests: Make CHECK enforce 1 instead of just "!= 0"

real-or-random:
This ensures that we don't omit the "== 1" in tests accidentally,
and thus also strengthens existing tests in which it has been omitted.

We want to check "== 1" in particular for the return values of API
functions, but it also makes sense in the case of internal functions.

If you really want to check "!= 0", you can still write it explicitly.

---

The second commit drops redundant "-1". I tried to be careful to drop it only where it makes sense. Now, `git grep "== 1"` shows only instances where the 1 is supposed to be an actual integer and not a boolean.  Still, this touches many lines, so I can drop this commit if you think it's too intrusive.


jonasnick (commented):
concept ACK

jonasnick:
> The second commit drops redundant "-1". 

You mean "== 1"?

theStack:
Concept ACK

Issue #1452, opened 2023-12-05
Add MuSig2 module


jonasnick:
I think a module for MuSig2 would be in the scope of libsecp256k1. Its relevance for the Bitcoin ecosystem is demonstrated by several factors:
- MuSig2 is already being adopted for [Bitcoin payments](https://blog.bitgo.com/save-fees-with-musig2-at-bitgo-3248d690f573).
- I have been inquired by various Lightning and (hardware) wallet developers about if and when MuSig2 will be in libsecp.
- Various developers of LN and (hardware) wallets have inquired about the integration of MuSig2 into libsecp and its timeline.
- There are specifications that depend on MuSig2, such as the ["simple taproot channels" BOLT](https://github.com/lightning/bolts/pull/995) and MuSig2 [PSBT](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2023-October/021988.html) and [descriptor](https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2023-October/021989.html) BIPs.

MuSig2 has a detailed [specification](https://github.com/bitcoin/bips/blob/master/bip-0327.mediawiki) (with reference code and test vectors) and [security proofs](https://eprint.iacr.org/2020/1261).

I suggest to copy the MuSig2 module from [libsecp256k1-zkp](https://github.com/BlockstreamResearch/secp256k1-zkp) which has already undergone significant review. I volunteer to do this. We should, however, remove the functions for MuSig2 adaptor signatures as they lack both a specification and a satisfactory security proof.

real-or-random:
Concept ACK

> We should, however, remove the functions for MuSig2 adaptor signatures as they lack both a specification and a satisfactory security proof. 

Makes sense.

LLFourn:
> We should, however, remove the functions for MuSig2 adaptor signatures as they lack both a specification and a satisfactory security proof.

I read this and assumed it just means not peer reviewed to the same standard as the MuSig2 protocol itself not that you had found a problem with your security proof you proposed for MuSig + adaptor before. Is this right?

t-bast:
Concept ACK, we'd be happy to start integrating this into lightning once that PR is opened!

jonasnick:
[...]

PR #1482, opened 2024-01-16
build: Clean up handling of module dependencies

real-or-random:
This is a cleanup which makes it easier to add further modules with dependencies, e.g., in #1452. The diff looks larger than it is because I also reordered the modules and made the order consistent between CMake and autotools.

(We noticed that the current logic could be improved in https://github.com/BlockstreamResearch/secp256k1-zkp/pull/275.)


configure.ac
```diff
@@ -387,29 +387,31 @@ SECP_CFLAGS="$SECP_CFLAGS $WERROR_CFLAGS"
 ### Handle module options
 ###
 
-if test x"$enable_module_ecdh" = x"yes"; then
-  SECP_CONFIG_DEFINES="$SECP_CONFIG_DEFINES -DENABLE_MODULE_ECDH=1"
-fi
-
-if test x"$enable_module_recovery" = x"yes"; then
-  SECP_CONFIG_DEFINES="$SECP_CONFIG_DEFINES -DENABLE_MODULE_RECOVERY=1"
+# Processing must be done in a reverse topological sorting of the dependency graph.
```
jonasnick, line 390: micronit: Since "reverse" is a matter of defining the direction of the edge, maybe 
```suggestion
# Processing must be done in a topological sorting of the dependency graph (dependency first).
```

configure.ac
```diff
@@ -387,29 +387,31 @@ SECP_CFLAGS="$SECP_CFLAGS $WERROR_CFLAGS"
 ### Handle module options
 ###
 
-if test x"$enable_module_ecdh" = x"yes"; then
-  SECP_CONFIG_DEFINES="$SECP_CONFIG_DEFINES -DENABLE_MODULE_ECDH=1"
-fi
```
jonasnick, line 392: Why change the order in configure.ac?

jonasnick (commented):
ACK eb52535bb87b36839095fc1034882bf9a5e65aa7

summoning @hebasto :)

hebasto:
May I ask a bit more explanation for the second commit please?

```
$ cmake -B build -DSECP256K1_ENABLE_MODULE_EXTRAKEYS=OFF
-- The C compiler identification is GNU 11.4.0
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
CMake Error at CMakeLists.txt:71 (message):
  schnorrsig module requires explicitly disabled extrakeys module.


-- Configuring incomplete, errors occurred!
See also "/home/hebasto/git/secp256k1/secp256k1/build/CMakeFiles/CMakeOutput.log".
```

The error message seems contradicts to `-DSECP256K1_ENABLE_MODULE_EXTRAKEYS=OFF`.

Or did I miss something?

jonasnick:
@hebasto Not sure if I understand your note about the error message. I think it is correct. But I did parse it incorrectly the first time.

>   schnorrsig module requires explicitly disabled extrakeys module.

means that the schnorrsig module requires the extrakeys module. It does *not* mean that the schnorrsig module requires the extrakeys module to be explicitly disabled.

jonasnick:
> The user can build the schnorrsig module either providing -DSECP256K1_ENABLE_MODULE_EXTRAKEYS=ON explicitly or relying on the default behavior when the SECP256K1_ENABLE_MODULE_EXTRAKEYS is set to ON by the CMake script, right?

No, turning on the extrakeys module should not turn on the schnorrsig module. The schnorrsig module requires the extrakeys module, not the other way around.

hebasto:
> > The user can build the schnorrsig module either providing -DSECP256K1_ENABLE_MODULE_EXTRAKEYS=ON explicitly or relying on the default behavior when the SECP256K1_ENABLE_MODULE_EXTRAKEYS is set to ON by the CMake script, right?
> 
> No, turning on the extrakeys module should not turn on the schnorrsig module. The schnorrsig module requires the extrakeys module, not the other way around.

Sorry for bad wording from my side. I mean, the schnorrsig module is enabled by default, and no extra actions, like providing `-DSECP256K1_ENABLE_MODULE_SCHNORRSIG=ON`, are needed, no?

jonasnick:
On master if you provide `--disable-module-extrakeys` to `./configure` or `-DSECP256K1_ENABLE_MODULE_EXTRAKEYS=OFF` it is simply ignored (because schnorrsig turns it on again). So it makes sense to throw an error instead of ignoring the explicit module disabling.

hebasto:
> On master if you provide `--disable-module-extrakeys` to `./configure` or `-DSECP256K1_ENABLE_MODULE_EXTRAKEYS=OFF` to cmake, it is simply ignored (because schnorrsig turns it on again). So it makes sense to throw an error instead of ignoring the explicit module disabling.

Agree.

configure.ac
real-or-random: To have the same order as in CMake, I think this makes maintenance easier. Let me add this rationale to the commit message? Or do you think I should not do this change?

real-or-random:
I changed the error message, which was apparently difficult to parse based on your feedback. (Thanks for being my user study.)

configure.ac
jonasnick: No I think it's fine.

jonasnick (approved):
ACK e6822678ea05c431b4f43be9dfbde54e0f7f645b

hebasto (approved):
ACK e6822678ea05c431b4f43be9dfbde54e0f7f645b.

PR #1483, opened 2024-01-17
cmake: Recommend native CMake commands in README

real-or-random:
Resolves one item in #1235. Closes #1294.

cc @hebasto

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,9 +79,9 @@ To maintain a pristine source tree, CMake encourages to perform an out-of-source
 
     $ mkdir build && cd build
     $ cmake ..
-    $ make
-    $ make check  # run the test suite
-    $ sudo make install  # optional
+    $ cmake --build .
+    $ ctest  # run the test suite
+    $ sudo cmake --build . --target install  # optional
 
 To compile optional modules (such as Schnorr signatures), you need to run `cmake` with additional flags (such as `-DSECP256K1_ENABLE_MODULE_SCHNORRSIG=ON`). Run `cmake .. -LH` to see the full list of available flags.
 


README.md
```diff
@@ -79,9 +79,9 @@ To maintain a pristine source tree, CMake encourages to perform an out-of-source
 
     $ mkdir build && cd build
     $ cmake ..
```
hebasto, line 81: nit: Might be explicit:
```suggestion
    $ cmake -S ..
```
?

README.md
```diff
@@ -79,9 +79,9 @@ To maintain a pristine source tree, CMake encourages to perform an out-of-source
...
-    $ make
-    $ make check  # run the test suite
-    $ sudo make install  # optional
+    $ cmake --build .
+    $ ctest  # run the test suite
+    $ sudo cmake --build . --target install  # optional
```
hebasto, line 84: Is it worth mentioning a modern and more effective alternative, which is available for CMake >= 3.15:
```
$ sudo cmake --install .  # optional
```
?

README.md
real-or-random: Hm, yeah, I think shorter is better, and omitting the -S is also more common. (See the official tutorial https://cmake.org/cmake/help/latest/guide/tutorial/A%20Basic%20Starting%20Point.html for example.)

README.md
real-or-random: Thanks for the suggestion. In fact, I had this first, but I changed to the current form because we still support CMake 3.13.

README.md
hebasto: OK, nm.

README.md
hebasto: I mean, mention it additionally, not instead of the current line.

hebasto: However, it's OK in its current view.

jonasnick (approved):
ACK 3777e3f36a61a05da19d5c6575fdd59288848948

PR #1487, opened 2024-01-29
Add SGX recipe

nolasconapoleao:


real-or-random:
What is an SGX recipe?

You can define `USE_EXTERNAL_CALLBACKS` if you want to avoid the printfs, or wait for https://github.com/bitcoin-core/secp256k1/pull/1461.

nolasconapoleao:
Hey sorry for that, didn't mean to make a PR.
Great feedback tho, will keep an eye on it.
Intel SGX [Software Guard Extensions](https://en.wikipedia.org/wiki/Software_Guard_Extensions) recipe provides a way to compile this code in a secure enclave that is tamper safe and cannot be monitored even by a malicious hardware host, as a requirement it cannot use calls to printf, that's what the PR is about.

PR #1488, opened 2024-01-31
ci: Add native macOS arm64 job

hebasto:
This PR starts using the [new](https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/) M1 macOS runner.

The alternative approach might be using a matrix, but it is not trivial to implement.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -585,10 +585,10 @@ jobs:
         run: env
         if: ${{ always() }}
 
-  macos-native:
-    name: "x86_64: macOS Monterey"
+  x86_64-macos-native:
+    name: "x86_64: macOS Monterey, Valgrind"
     # See: https://github.com/actions/runner-images#available-images.
-    runs-on: macos-12 # Use M1 once available https://github.com/github/roadmap/issues/528
+    runs-on: macos-12
 
     env:
       CC: 'clang'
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -585,10 +585,10 @@ jobs:
         run: env
         if: ${{ always() }}
 
-  macos-native:
-    name: "x86_64: macOS Monterey"
+  x86_64-macos-native:
+    name: "x86_64: macOS Monterey, Valgrind"
```
real-or-random, line 589: Did you remove the "Valgrind" intentionally?

real-or-random (commented):
Concept ACK

It's a bit sad that we can't run Valgrind on the new machines yet.

.github/workflows/ci.yml
hebasto: Do you mean in the https://github.com/bitcoin-core/secp256k1/pull/1412?

real-or-random: No, I mean you have removed the substring ", Valgrind" from the job name. :D

hebasto: Sorry, I don't follow... The diff _appends_ the substring ", Valgrind" to the job name, no?

real-or-random: Lol sure, sorry, nevermind...

real-or-random:
I think it will be good to add an MSan job with ctimetests enabled. We have an arm64 job with Valgrind on Linux (Cirrus CI), but I guess having a ctimetest for macOS (Apple clang) can't hurt.

hebasto:
> I think it will be good to add an MSan job with ctimetests enabled. We have an arm64 job with Valgrind on Linux (Cirrus CI), but I guess having a ctimetest for macOS (Apple clang) can't hurt.

For both Apple Clang and LLVM Clang, I have:
```
clang: error: unsupported option '-fsanitize=memory' for target 'arm64-apple-darwin23.5.0'
```

real-or-random:
Ok, damn, I didn't know that MSan is not supported on macOS... https://github.com/google/sanitizers/issues/588

real-or-random (approved):
ACK 218f0cc93bd77eae926f4c4e259951b92bf00aec

Issue #1472, opened 2023-12-24
fe_sqr tests could be improved


Coding-Enthusiast:
https://github.com/bitcoin-core/secp256k1/blob/efe85c70a2e357e3605a8901a9662295bae1001f/src/tests.c#L3289-L3303  

This method doesn't test anything and after looking through commits it was added in https://github.com/bitcoin-core/secp256k1/commit/59447da38d6fa394ab7b10f32aea55930da732be and got removed in the next commit https://github.com/bitcoin-core/secp256k1/commit/21f81a846957408d4a2fe30cff32370567887a93. I'm not sure what the "problem" mentioned in the first commit is but if it is checking it using the `SECP256K1_FE_VERIFY` method, it is pointless to call `run_sqr` without this build configuration.

real-or-random:
> it is pointless to call `run_sqr` without this build configuration.

Well, there's nothing wrong with it, except that it adds a few milliseconds to the noverify_tests. I claim that adding `#ifdefs` isn't worth the complexity and the development time. So this is not an issue.

The test itself could perhaps be improved. We could compare with the result of another method, e.g., `fe_mul`. We can't use `fe_mul` to compute `x*x` directly (see docs) but we could, e.g., check that `(x+y)*(x-y) = x^2 - y^2` for some random values `y != 0`.

real-or-random:
@Coding-Enthusiast 

> we could, e.g., check that `(x+y)*(x-y) = x^2 - y^2` for some random values `y != 0`.

Do you want to give it a try and work on a PR?

Coding-Enthusiast:
Sadly my C skills are readonly 😊

real-or-random:
We all started small... ;)

real-or-random:
This has been resolved by #1490.

PR #1489, opened 2024-02-01
tests: add missing fe comparison checks for inverse field test cases

theStack:
In the course of looking into #1472, I discovered the function `check_fe_equal` and noticed that at two call-sites (from 18 in total) there is no `CHECK` macro around, i.e. the return value is just thrown away and a failed test would go unnoticed. This seems to be easy to miss, as the name of the function suggests that a check happens, but it's merely a wrapper around `secp256k1_fe_equal` that takes care of normalization.

(Maybe doing the actual CHECK in `check_fe_equal` would be an option? On the other hand, I guess this makes it harder to spot the cause of a failed assertion, as the line number wouldn't tell anything about from where the function was called.)

### DIFF
--- a/src/modules/ellswift/tests_impl.h
+++ b/src/modules/ellswift/tests_impl.h
@@ -188,9 +188,9 @@ void run_ellswift_tests(void) {
             CHECK(ret == ((testcase->enc_bitmap >> c) & 1));
             if (ret) {
                 secp256k1_fe x2;
-                CHECK(check_fe_equal(&t, &testcase->encs[c]));
+                CHECK(fe_equal(&t, &testcase->encs[c]));
                 secp256k1_ellswift_xswiftec_var(&x2, &testcase->u, &testcase->encs[c]);
-                CHECK(check_fe_equal(&testcase->x, &x2));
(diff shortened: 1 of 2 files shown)


real-or-random:
> but it's merely a wrapper around `secp256k1_fe_equal` that takes care of normalization.

Perhaps we should just get rid of it. I don't think it's worth having a test function if it's only used in two call sites, and anyway, people will forget that it exists when writing new tests.

real-or-random:
Then I suggest renaming it just to `fe_equal`. 

> (Maybe doing the actual CHECK in check_fe_equal would be an option? On the other hand, I guess this makes it harder to spot the cause of a failed assertion, as the line number wouldn't tell anything about from where the function was called.)

Yeah, I agree. Our idiom is to use CHECK and not have special check functions.

real-or-random (approved):
utACK e7bdddd9c9c4b9f4f44420f3c7eeaef44f0c82ff

jonasnick (approved):
ACK e7bdddd9c9c4b9f4f44420f3c7eeaef44f0c82ff

PR #1490, opened 2024-02-01
tests: improve fe_sqr test (issue #1472)

theStack:
As pointed out in issue #1472, currently the `run_sqr` test doesn't do anything with the result of the `fe_sqr` call. Improve that by checking that the equation `(x+y)*(x-y) = x^2 - y^2` holds for some random values y, as suggested by real-or-random in https://github.com/bitcoin-core/secp256k1/issues/1472#issuecomment-1868562354. The existing loop for generating the x values is kept as-is, the sqr call on x is still included in each iteration on the course of calulating the RHS of the equation (`secp256k1_fe_sqr(&rhs, &x);`).

The suggestion says "for random values `y != 0`", but I couldn't see a good reason to exclude zero. Happy to apply s/random_fe_test/random_fe_non_zero_test/ if needed.

### DIFF
--- a/src/field.h
+++ b/src/field.h
@@ -255,8 +255,8 @@ static void secp256k1_fe_add(secp256k1_fe *r, const secp256k1_fe *a);
 /** Multiply two field elements.
  *
  * On input, a and b must be valid field elements; r does not need to be initialized.
- * r and a may point to the same object, but neither can be equal to b. The magnitudes
- * of a and b must not exceed 8.
+ * r and a may point to the same object, but neither may point to the object pointed
+ * to by b. The magnitudes of a and b must not exceed 8.
  * Performs {r = a * b}
  * On output, r will have magnitude 1, but won't be normalized.
  */
--- a/src/tests.c
+++ b/src/tests.c
@@ -3285,18 +3285,31 @@ static void run_fe_mul(void) {
 }
 
 static void run_sqr(void) {
-    secp256k1_fe x, s;
+    int i;
+    secp256k1_fe x, y, lhs, rhs, tmp;
 
-    {
-        int i;
-        secp256k1_fe_set_int(&x, 1);
-        secp256k1_fe_negate(&x, &x, 1);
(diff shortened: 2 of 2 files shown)


real-or-random:
Nice!

> The suggestion says "for random values `y != 0`", but I couldn't see a good reason to exclude zero. Happy to apply s/random_fe_test/random_fe_non_zero_test/ if needed.

Ah, this is because `fe_mul(_, a, b)` cannot handle the case when `a` is equal to `b`:
https://github.com/bitcoin-core/secp256k1/blob/master/src/field.h#L255-L263

In fact, this is why I suggested the binomial equation. Otherwise, the most natural test would just be  `x*x = x^2`.

I mean, it doesn't matter for truly random `y`, but we should still call `random_fe_non_zero_test` instead. (This reminds me that we could consider biasing the test random function towards zero and other corner cases, but that's a different project...)

real-or-random:
Oh nevermind, you're right. :D Sorry, this was a brain fart. And indeed, this would be pretty bad.  

Now that you have implemented it, I still think that `(x+y)*(x-y) = x^2 - y^2` is a nice test. It gets us a good test for two sqr operations for the price of 1 mul. ;)

real-or-random (approved):
utACK b702461a8ddd45c6edcf45627200e52e78985947

real-or-random (approved):
utACK 7dcb3a8326507c2cd267cda7d333f8e261568f9b

jonasnick (approved):
ACK 7dcb3a8326507c2cd267cda7d333f8e261568f9b

jonasnick (changes requested):
Needs rebase (due to `check_fe_equal` not existing anymore in master)

real-or-random (approved):
utACK 2028069df2e16a05b332ae24c7ae63791f461063

jonasnick (approved):
ACK 2028069df2e16a05b332ae24c7ae63791f461063

PR #1492, opened 2024-02-04
tests: Add Wycheproof ECDH vectors

RandomLattice:
Adds a test for the ECDH module using the Wycheproof vectors as outlined in https://github.com/bitcoin-core/secp256k1/issues/1106.

This commit adds 479 ECDH test vectors. All test vectors pass. The vectors cover:

* edge cases in the shared secret
* edge cases in the ephemeral public keys
* edge cases in arithmetic operations

We use a python script to convert the JSON-formatted vectors into C code, in the same spirit as https://github.com/bitcoin-core/secp256k1/pull/1245


src/modules/ecdh/tests_impl.h
```diff
@@ -7,6 +7,14 @@
 #ifndef SECP256K1_MODULE_ECDH_TESTS_H
 #define SECP256K1_MODULE_ECDH_TESTS_H
 
+static int ecdh_hash_function_test_xpassthru(unsigned char *output, const unsigned char *x, const unsigned char *y, void *data) {
+    (void)x;
```
real-or-random, line 11:
```suggestion
```
You use `x` below.

real-or-random (commented):
Awesome, Concept ACK

I think most contributors here are currently busy with other projects, but we'll come back to this for sure.

src/modules/ecdh/tests_impl.h
```diff
@@ -142,11 +150,53 @@ static void test_result_basepoint(void) {
...
+        int actual;
+
+        memset(&point, 0, sizeof(point));
+        pk = &wycheproof_ecdh_public_keys[testvectors[t].pk_offset];
+        parsed_ok = secp256k1_ec_pubkey_parse(CTX, &point, pk, testvectors[t].pk_len);
+
+        expected_result = testvectors[t].expected_result;
+
+        /* fail if public key is valid, but doesn't parse */
+        CHECK(parsed_ok || expected_result == 0);
```
real-or-random, line 174: I think this should be tighter:
```suggestion
        CHECK(parsed_ok == expected_result);
```

The missing case is `parsed_ok == 1` and `expected_result == 0`. If I'm not mistaken, then the entire test doesn't fail currently:
 - the CHECK here passes
 - we don't `continue`
 - but then the `CHECK(secp256k1_memcmp_var`)...` passes vacuously because `testvectors[t].shared_len == 0`

src/wycheproof/WYCHEPROOF_COPYING
```diff
@@ -3,9 +3,18 @@
   `b063b4aedae951c69df014cd25fa6d69ae9e8cb9`, see
   https://github.com/google/wycheproof/blob/b063b4aedae951c69df014cd25fa6d69ae9e8cb9/testvectors_v1/ecdsa_secp256k1_sha256_bitcoin_test.json
 
+* The file `ecdh_secp256k1_test.json` in this directory
+  comes from Google's project Wycheproof with git commit
+  `d9f6ec7d8bd8c96da05368999094e4a75ba5cb3d`, see
+  https://github.com/google/wycheproof/blob/d9f6ec7d8bd8c96da05368999094e4a75ba5cb3d/testvectors_v1/ecdh_secp256k1_test.json
```
real-or-random, line 9: Wycheproof ownership was recently moved to C2SP (https://github.com/C2SP/wycheproof community maintenance), so this should be updated to the new URL.) See @FiloSottile's talk https://archive.org/details/oscw-2024-fillippo-valsorda-cryptographic-test-vectors for background.)

You could update the other URLs in a separate commit, and update the ECDSA vectors, see https://github.com/C2SP/wycheproof/pull/91 (if you're willing to care of this in this PR).

src/modules/ecdh/tests_impl.h
```diff
@@ -142,11 +150,53 @@ static void test_result_basepoint(void) {
...
+    int t;
+    for (t = 0; t < SECP256K1_ECDH_WYCHEPROOF_NUMBER_TESTVECTORS; t++) {
+        int parsed_ok;
+        secp256k1_pubkey point;
+        const unsigned char *pk;
+        const unsigned char *sk;
+        const unsigned char *expected_shared_secret;
+        unsigned char output_ecdh[65] = { 0 };
+
+        int expected_result;
```
real-or-random, line 164: Can we rename this to `expected_parses` or similar? `result` is a bit imprecise.

tools/tests_wycheproof_generate_ecdh.py
```diff
@@ -0,0 +1,150 @@
...
+
+import json
+import sys
+
+from binascii import hexlify, unhexlify
+from wycheproof_utils import to_c_array
+
+def should_skip(test_vector_flags):
+    # skip these vectors because they are for ASN.1 encoding issues and other curves
+    flags_to_skip = {"InvalidAsn", "InvalidCurveAttack", "InvalidEncoding", "WrongCurve", "UnnamedCurve"}
```
real-or-random, line 17: I'm not sure about all of these.
 - "InvalidAsn': :heavy_check_mark: 
 - "InvalidCurveAttack". the json says "The point of the public key is not on the curve." -- shouldn't we have these? (How is this different from "InvalidPublic"? I assume the keys in case of "InvalidCurveAttack" are on some other curve.) 
 - What is "InvalidEncoding"?
 - "WrongCurve": I really don't understand the JSON here. For example, test case 492 says: "public key has invalid point of order 2 on secp256r1. The point of the public key is a valid on secp256k1. ", but then says "invalid"?! Do you know what they have in mind? 
 - "UnnamendCurve":  Have you tried these? If we reject correctly, let's just include them? Again, I can't follow the JSON entirely. :/  For example, test case 511 has "public key of order 3" with "WeakPublicKey", "InvalidPublic", "UnnamedCurve". How can it be invalid and at the same time have an order? How can the order be 3 on our curve? (Shouldn't it have InvalidCurveAttack then? ...)

real-or-random (commented):
Concept ACK

tools/tests_wycheproof_generate_ecdh.py
```diff
@@ -0,0 +1,171 @@
...
+offset_sk_running, offset_pk_running, offset_shared = 0, 0, 0
+out = ""
+private_keys = ""
+shared_secrets = ""
+public_keys = ""
+cache_sks = {}
+cache_public_keys = {}
+
+for i in range(num_groups):
+    group = doc['testGroups'][i]
```
real-or-random, line 86:
```suggestion
for group in doc['testGroups']
```

And then we can remove `num_groups = len(doc['testGroups'])`

tools/tests_wycheproof_generate_ecdh.py
```diff
@@ -0,0 +1,171 @@
...
+    num_tests = len(group['tests'])
+    assert group["type"] == "EcdhTest"
+    assert group["curve"] == "secp256k1"
+    for j in range(num_tests):
+        test_vector = group['tests'][j]
```
real-or-random, line 91:
```suggestion
    for test_vector in group['tests']:
```

And then we can remove `num_tests = len(group['tests'])`

tools/tests_wycheproof_generate_ecdh.py
```diff
@@ -0,0 +1,171 @@
...
+filename_input = sys.argv[1]
+
+with open(filename_input) as f:
+    doc = json.load(f)
+
+num_groups = len(doc['testGroups'])
+
+num_vectors = 0
+offset_sk_running, offset_pk_running, offset_shared = 0, 0, 0
+out = ""
```
real-or-random, line 78: `out` is a bit ambigous. `test_vectors_out`?

tools/tests_wycheproof_generate_ecdh.py
```diff
@@ -0,0 +1,171 @@
...
+        return parse_der_pk(rest)
+    if tag == 0x03: # BIT STRING
+        return value
+    if tag == 0x30: # SEQUENCE
+        return parse_der_pk(value)
+    raise ValueError("unknown tag")
+
+def parse_public_key(pk):
+    der_pub_key = parse_der_pk(unhexlify(pk))   # Convert back to str and strip off the `0x`
+    return hexlify(der_pub_key).decode('utf-8')[2:]
```
real-or-random, line 57:
```suggestion
    return hexlify(der_pub_key).decode()[2:]
```

`utf-8` is the default

tools/tests_wycheproof_generate_ecdh.py
```diff
@@ -0,0 +1,171 @@
...
+        if L == 0x81:
+            L = int(s[2])
+            offset = 1
+        elif L == 0x82:
+            L = 256*int(s[2]) + int(s[3])
+            offset = 2
+        else:
+            raise ValueError("invalid L")
+    value = s[(offset+2):(L+2+offset)]
+    rest = s[(L+2+offset):]
```
real-or-random, line 45: just a micronit, but I'd prefer spacing around the operators

tools/wycheproof_utils.py
```diff
@@ -0,0 +1,12 @@
...
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or https://www.opensource.org/licenses/mit-license.php.
+'''
+Utility functions for generating C files for testvectors from the Wycheproof project.
+'''
+
+def to_c_array(x):
+    if x == "":
+        return ""
+    s = ',0x'.join(a+b for a,b in zip(x[::2], x[1::2]))
```
real-or-random, line 11:
```suggestion
    s = ',0x'.join(a + b for a, b in zip(x[::2], x[1::2]))
```

tools/tests_wycheproof_generate_ecdh.py
```diff
@@ -0,0 +1,171 @@
...
+        public_key = parse_public_key(test_vector['public'])
+        private_key = normalize_private_key(test_vector['private'])
+        expected_result = normalize_expected_result(test_vector['result'])
+
+        # // 2 to convert hex to byte length
+        shared_size = len(test_vector['shared']) // 2
+        sk_size = len(private_key) // 2
+        pk_size = len(public_key) // 2
+
+        shared_secrets += ",\n  " if num_vectors and shared_size else ""
```
real-or-random, line 105: move this down just above the other modification of `shared_secrets`?

tools/tests_wycheproof_generate_ecdh.py
```diff
@@ -0,0 +1,171 @@
...
+    raise ValueError("unknown tag")
+
+def parse_public_key(pk):
+    der_pub_key = parse_der_pk(unhexlify(pk))   # Convert back to str and strip off the `0x`
+    return hexlify(der_pub_key).decode('utf-8')[2:]
+
+def normalize_private_key(sk):
+    # Ensure the private key is at most 64 characters long, retaining the last 64 if longer.
+    normalized = sk[-64:].zfill(64)
+    assert len(normalized) == 64, "private key must be exactly 64 characters long."
```
real-or-random, line 62:
```suggestion
    if len(normalized) != 64:
        raise ValueError("private key must be exactly 64 characters long.")
```

(Don't rely on assertions being enabled)

tools/tests_wycheproof_generate_ecdh.py
```diff
@@ -0,0 +1,171 @@
...
+    if tag == 0x30: # SEQUENCE
+        return parse_der_pk(value)
+    raise ValueError("unknown tag")
+
+def parse_public_key(pk):
+    der_pub_key = parse_der_pk(unhexlify(pk))   # Convert back to str and strip off the `0x`
+    return hexlify(der_pub_key).decode('utf-8')[2:]
+
+def normalize_private_key(sk):
+    # Ensure the private key is at most 64 characters long, retaining the last 64 if longer.
```
real-or-random, line 60: How could it ever be longer?

tools/tests_wycheproof_generate_ecdh.py
real-or-random: Nice, thanks a lot for going through all of this!

tools/tests_wycheproof_generate_ecdh.py
```diff
@@ -0,0 +1,171 @@
...
+    return any(flag in test_vector_flags for flag in flags_to_skip)
+
+def should_skip_tcid(test_vector_tcid):
+    '''
+    We skip some test case IDs that have a public key whose custom ASN.1 representation explicitly
+    encodes some curve parameters that are invalid. libsecp256k1 never parses this part so we do
+    not care testing those. See https://github.com/bitcoin-core/secp256k1/pull/1492#discussion_r1572491546
+    '''
+    tcids_to_skip = [496, 497, 502, 503, 504, 505, 507]
+    return test_vector_tcid in tcids_to_skip
```
real-or-random, line 27: consistency: docstring vs comment. (I'd just stick to the comment).

tools/tests_wycheproof_generate_ecdh.py
```diff
@@ -0,0 +1,171 @@
...
+
+import json
+import sys
+
+from binascii import hexlify, unhexlify
+from wycheproof_utils import to_c_array
+
+def should_skip_flags(test_vector_flags):
+    # skip these vectors because they are for ASN.1 encoding issues and other curves
+    flags_to_skip = {"InvalidAsn", "WrongCurve"}
```
real-or-random, line 17: A short description list of all the (unclear) categories would be nice. This has confused us, and I assume it will confuse others in the future. (Or just add another reference to https://github.com/bitcoin-core/secp256k1/pull/1492#discussion_r1572491546 )

real-or-random (commented):
Nice! All I have is some nits, and I believe this is ready. It should perhaps get the eyes of another reviewer, but I don't think another in-depth review is necessary. In the end, this "just" adds tests.

Sorry, that this too so long.

src/wycheproof/WYCHEPROOF_COPYING
real-or-random: If you're still interested, then I think a proper version of the abandoned (?) https://github.com/bitcoin-core/secp256k1/pull/1638 could be included in this PR. It should be trivial.

jonasnick (commented):
Concept ACK

Hey @RandomLattice, thanks for your contribution. This looks very good. Can you squash the last commit? And since you're touching the commit again can you simplify the following code in test_ecdh_wycheproof

```C
actual = secp256k1_ecdh(CTX, output_ecdh, &point, sk, ecdh_hash_function_test_xpassthru, NULL);
expected_shared_secret = &wycheproof_ecdh_shared_secrets[testvectors[t].shared_offset];

CHECK(actual == expected_result);
```
to

```C
CHECK(secp256k1_ecdh(CTX, output_ecdh, &point, sk, ecdh_hash_function_test_xpassthru, NULL) == 1);
expected_shared_secret = &wycheproof_ecdh_shared_secrets[testvectors[t].shared_offset];

```
and remove the declaration of `actual`? Thanks.

jonasnick (approved):
ACK e266ba11ae9fd20f011ac303d82caf8017d8b627

PR #1496, opened 2024-02-21
msan: notate variable assignments from assembly code

theuni:
msan isn't smart enough to see that these are set without some help.

This was pointed out here: https://github.com/bitcoin-core/secp256k1/pull/1169#issuecomment-1370003449

With this commit, msan output is clean even with x86 asm turned on.

### DIFF
--- a/src/checkmem.h
+++ b/src/checkmem.h
@@ -30,6 +30,8 @@
  * - SECP256K1_CHECKMEM_DEFINE(p, len):
  *   - marks the len-byte memory pointed to by p as defined data (public data, in the
  *     context of constant-time checking).
+ * - SECP256K1_CHECKMEM_MSAN_DEFINE(p, len):
+ *   - Like SECP256K1_CHECKMEM_DEFINE, but applies only to memory_sanitizer.
  *
  */
 
@@ -48,9 +50,14 @@
 #    define SECP256K1_CHECKMEM_ENABLED 1
 #    define SECP256K1_CHECKMEM_UNDEFINE(p, len) __msan_allocated_memory((p), (len))
 #    define SECP256K1_CHECKMEM_DEFINE(p, len) __msan_unpoison((p), (len))
+#    define SECP256K1_CHECKMEM_MSAN_DEFINE(p, len) __msan_unpoison((p), (len))
 #    define SECP256K1_CHECKMEM_CHECK(p, len) __msan_check_mem_is_initialized((p), (len))
 #    define SECP256K1_CHECKMEM_RUNNING() (1)
 #  endif
+#endif
+
+#if !defined SECP256K1_CHECKMEM_MSAN_DEFINE
+#  define SECP256K1_CHECKMEM_MSAN_DEFINE(p, len) SECP256K1_CHECKMEM_NOOP((p), (len))
 #endif
 
 /* If valgrind integration is desired (through the VALGRIND define), implement the
--- a/src/scalar_4x64_impl.h
+++ b/src/scalar_4x64_impl.h
@@ -462,6 +462,14 @@ static void secp256k1_scalar_reduce_512(secp256k1_scalar *r, const uint64_t *l) 
     : "S"(l), "i"(SECP256K1_N_C_0), "i"(SECP256K1_N_C_1)
     : "rax", "rdx", "r8", "r9", "r10", "r11", "r12", "r13", "r14", "cc");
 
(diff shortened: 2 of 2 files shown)


real-or-random (approved):
utACK 8f237f6ff13f5654b1ca7d31481ca96e58fc7d02

Oh, that's nice. I run into this regularly, but I didn't think of this simple solution.

fanquake:
Nice. Concept ACK. I'll run this through out MSAN CI while using asm.

hebasto:
Concept ACK.

Is still there any specific reason to keep `ASM=no` in the CI?

FWIW, I've removed it everywhere in the `.github/workflows/ci.yml` and the CI passed.

real-or-random:
> Is still there any specific reason to keep `ASM=no` in the CI?

Yes, I think we want to test the non-ASM code even on x86_64. But you're right in principle. We should probably enable ASM in some more tasks. (Reworking the matrix something we should do anyway. It's one of the few boxes we haven't ticked in https://github.com/bitcoin-core/secp256k1/issues/1392 ... :/ )

hebasto (approved):
ACK 8f237f6ff13f5654b1ca7d31481ca96e58fc7d02, tested on Ubuntu 23.10. Removing any of added macros triggers "MemorySanitizer: use-of-uninitialized-value" warning in the setup mentioned in the PR description.

real-or-random:
@theuni Do you think it's worth making these conditional under `__has_feature(memory_sanitizer)` (or cleaner, a macro `SECP256K1_CHECKMEM_MSAN` to be added to `checkmem.h`) to make sure that this workaround applies only to MSan, and we never suppress positives omitted by Valgrind.

theuni:
@real-or-random I didn't understand your question at first but I think maybe I know what you're asking now. I'll try to rephrase:

Because notation is necessary for msan but not valgrind (I don't know if that's true, I'm assuming based on your question), there should be a macro that is used only to notate for msan without affecting valgrind's detection? If that's right, sure, I can hook up a `SECP256K1_CHECKMEM_MSAN`. I'm unsure if other parts of the code would also benefit from one but not the other, but presumably if so we'd want to switch them over too.

real-or-random:
Yeah, sorry, my comment was not very detailed. :)

> Because notation is necessary for msan but not valgrind (I don't know if that's true, I'm assuming based on your question), 

Indeed. MSan's analysis needs instrumentation from the compiler, and it's not surprising that the compiler is not clever enough to understand our ASM and instrument the binary accordingly. Valgrind's analysis looks at the final (uninstrumented) binary only, and thus doesn't care at all if the code has been generated from C or asm or whatever.

> there should be a macro that is used only to notate for msan without affecting valgrind's detection? If that's right, sure, I can hook up a `SECP256K1_CHECKMEM_MSAN`.

That's what I had in mind, yes.

> I'm unsure if other parts of the code would also benefit from one but not the other, but presumably if so we'd want to switch them over too.

I grepped for `SECP256K1_CHECKMEM_DEFINE`, and it turns out that all current uses of the macro are related to constant-time testing. In these cases, we'll actually need it for both MSan and Valgrind. So no, no need to change any existing uses.  

> Nit: I think `SECP256K1_CHECKMEM_MSAN_DEFINE` would be a better name so the other macros could get the same treatment in the future if necessary.

Yeah, totally. Sorry, what I actually had in mind was that `SECP256K1_CHECKMEM_MSAN` (or perhaps better: ``SECP256K1_CHECKMEM_USES_MSAN`) is macro that indicates that we use MSan, and then we could do this after the asm code.  
```
#if SECP256K1_CHECKMEM_USES_MSAN
    SECP256K1_CHECKMEM_DEFINE(...)
    ...
#endif
```

And there could also be a `SECP256K1_CHECKMEM_USES_VALGRIND` macro.

I think that's a tiny tiny bit simpler in a possible future with more than two backends, but really, either approach is fine in the end. And I agree, if we follow your approach, then `SECP256K1_CHECKMEM_MSAN_DEFINE` is a better name for sure.

real-or-random (approved):
utACK 31ba40494428dcbf2eb5eb6f2328eca91b0b0746

hebasto (approved):
re-ACK 31ba40494428dcbf2eb5eb6f2328eca91b0b0746.

Just leaving here a reminder to reconsider disabling assembly in the CI jobs.

fanquake:
Have a followup to this here: https://github.com/bitcoin/bitcoin/pull/29742.

Issue #1511, opened 2024-03-27
msan: use of uninitialized value in secp256k1_scalar_mul_shift_var


fanquake:
Building master (05bfab69aef3622f77f754cfb01220108a109c91) in the following way (same flags as we use in our MSAN CI), results in the following failure:
```bash
Ubuntu clang version 17.0.6 (5build1) (x86_64)
./autogen.sh
./configure CFLAGS="-fsanitize=memory -fsanitize-memory-track-origins=2 -fno-omit-frame-pointer -g -O1 -fno-optimize-sibling-calls" CC=clang
Build Options:
  with external callbacks = no
  with benchmarks         = yes
  with tests              = yes
  with ctime tests        = yes
  with coverage           = no
  with examples           = no
  module ecdh             = yes
  module recovery         = no
  module extrakeys        = yes
  module schnorrsig       = yes
  module ellswift         = yes

  asm                     = x86_64
  ecmult window size      = 15
  ecmult gen prec. bits   = 4

  valgrind                = yes
  CC                      = clang
  CPPFLAGS                = 
  SECP_CFLAGS             = -O2  -std=c89 -pedantic -Wno-long-long -Wnested-externs -Wshadow -Wstrict-prototypes -Wundef -Wno-overlength-strings -Wall -Wno-unused-function -Wextra -Wcast-align -Wconditional-uninitialized -Wreserved-identifier -fvisibility=hidden 
  CFLAGS                  = -fsanitize=memory -fsanitize-memory-track-origins=2 -fno-omit-frame-pointer -g -O1 -fno-optimize-sibling-calls
  LDFLAGS                 = 
make check -j17
<snip>
cat test-suite.log 
==============================================
   libsecp256k1 0.4.2-dev: ./test-suite.log
==============================================

# TOTAL: 3
# PASS:  2
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0

.. contents:: :depth: 2

FAIL: tests
===========

test count = 64
random seed = 799c333349f126d5ee67e7ac48fa9dc8
==3962699==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x556d1c78cb8d in secp256k1_scalar_mul_shift_var /root/secp256k1/src/scalar_4x64_impl.h:909:5
    #1 0x556d1c7894d8 in secp256k1_scalar_split_lambda /root/secp256k1/src/scalar_impl.h:162:5
[...]

PR #1512, opened 2024-03-27
msan: notate more variable assignments from assembly code

theuni:
This was missed in 31ba40494428dcbf2eb5eb6f2328eca91b0b0746 because older versions of clang did not complain about it. But clang-17, at least, does.

The array-as-a-param makes this annoying because `sizeof(l)` is not helpful. I'd be happy to change the size calculation if there are any better suggestions or strong preferences.

### DIFF
--- a/src/scalar_4x64_impl.h
+++ b/src/scalar_4x64_impl.h
@@ -814,6 +814,9 @@ static void secp256k1_scalar_mul_512(uint64_t *l8, const secp256k1_scalar *a, co
     : "+d"(pb)
     : "S"(l8), "D"(a->d)
     : "rax", "rbx", "rcx", "r8", "r9", "r10", "r11", "r12", "r13", "r14", "r15", "cc", "memory");
+
+    SECP256K1_CHECKMEM_MSAN_DEFINE(l8, sizeof(*l8) * 8);
+
 #else
     /* 160 bit accumulator. */
     uint64_t c0 = 0, c1 = 0;
--- a/src/scalar_4x64_impl.h
+++ b/src/scalar_4x64_impl.h
@@ -681,7 +681,7 @@ static void secp256k1_scalar_reduce_512(secp256k1_scalar *r, const uint64_t *l) 
     secp256k1_scalar_reduce(r, c + secp256k1_scalar_check_overflow(r));
 }
 
-static void secp256k1_scalar_mul_512(uint64_t l[8], const secp256k1_scalar *a, const secp256k1_scalar *b) {
+static void secp256k1_scalar_mul_512(uint64_t *l8, const secp256k1_scalar *a, const secp256k1_scalar *b) {
 #ifdef USE_ASM_X86_64
     const uint64_t *pb = b->d;
     __asm__ __volatile__(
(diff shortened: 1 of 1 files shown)


theuni:
Should fix #1511

fanquake:
Nice. This fixes #1511 for me, under Clang 17 (`Ubuntu clang version 17.0.6 (5build1)`) and 18 (`Ubuntu clang version 18.1.0 (rc2-4)`). I've also cherry-picked it into https://github.com/bitcoin/bitcoin/pull/29742 to run through our CI.

real-or-random (approved):
utACK 241a2145d15f5b9e2978c3a76d4fec9eefbc96f8

real-or-random:
> The array-as-a-param makes this annoying because `sizeof(l)` is not helpful.

Hm, yeah, idk, the `uint64_t l[8]` parameter declaration is a bit strange, or at least unusual for this code base. I'd ACK a patch that changes this a pointer declaration, which we usually do in the codebase. If you do, maybe also rename to `l8`.

For others: The issue is that, despite the declaration, `l` will still decay to a pointer, i.e., `l` won't be of array type, but of type `uint64_t*`. And the `[8]` has no significance.  That means that `sizeof(l) == sizeof(uint64_t*)`, which, for maximum confusion, happens to be 8 coincidentally.

https://github.com/bitcoin-core/secp256k1/blob/05bfab69aef3622f77f754cfb01220108a109c91/src/scalar_4x64_impl.h#L684

src/scalar_4x64_impl.h
```diff
@@ -812,39 +812,42 @@ static void secp256k1_scalar_mul_512(uint64_t l[8], const secp256k1_scalar *a, c
...
     : "rax", "rbx", "rcx", "r8", "r9", "r10", "r11", "r12", "r13", "r14", "r15", "cc", "memory");
+
+    SECP256K1_CHECKMEM_MSAN_DEFINE(l8, sizeof(*l8) * 8);
+
 #else
     /* 160 bit accumulator. */
     uint64_t c0 = 0, c1 = 0;
     uint32_t c2 = 0;
 
     /* l[0..7] = a[0..3] * b[0..3]. */
```
real-or-random, line 825:
```suggestion
    /* l8[0..7] = a[0..3] * b[0..3]. */
```
And there are some other comments that may require updating... E.g., `Extract l7`. I admit that these are confusing now that the variable name is `l8`. But I guess it should anyway be `Extract l8[7]`...

real-or-random:
> ```c
> void func(uint64_t l[static 8])
> ```

Haha okay, I totally wasn't aware of this, even though I happen to look at the standards text from time to time... And yeah, `static` is a bit cursed.

src/scalar_4x64_impl.h
theuni: Ah, I figured comments like `l6` still made sense conceptually, but sure, I'll update them to match.

real-or-random:
For posterity, the reason why clang 15 was happy with just the annotation in https://github.com/bitcoin-core/secp256k1/commit/31ba40494428dcbf2eb5eb6f2328eca91b0b0746 is this: 

The default of what is considered a "use" of uninitialized memory was changed in clang 16.  Returning an uninitialized variables from a function, or passing uninitialized values to a function as a parameter is now considered, and MSan will report it by default. See the [Clang 16.0.09 Release Notes](https://releases.llvm.org/16.0.0/tools/clang/docs/ReleaseNotes.html#sanitizers):

>    -fsanitize-memory-param-retval is turned on by default. With -fsanitize=memory, passing uninitialized variables to functions and returning uninitialized variables from functions is more aggressively reported. -fno-sanitize-memory-param-retval restores the previous behavior.

Issue #1516, opened 2024-04-09
False positives in constant-time tests when using MSan on Clang >= 16


real-or-random:
(Description partly copied from https://github.com/bitcoin-core/secp256k1/pull/1512#issuecomment-2045254282)

The default of what is considered a "use" of uninitialized memory was changed in clang 16.  Returning an uninitialized variables from a function, or passing uninitialized values to a function as a parameter is now considered, and MSan will report it by default. See the [Clang 16.0.09 Release Notes](https://releases.llvm.org/16.0.0/tools/clang/docs/ReleaseNotes.html#sanitizers):

>    `-fsanitize-memory-param-retval` is turned on by default. With `-fsanitize=memory`, passing uninitialized variables to functions and returning uninitialized variables from functions is more aggressively reported. `-fno-sanitize-memory-param-retval` restores the previous behavior.

This makes sense for actual memory checking, in the sense that uninitialized memory at the boundary of a function call should probably be considered a bug or at least a smell. But that's certainly *not* what we want if we (ab)use MSan for constant-time checking, and this gives us false positives. 

We should pass `-fno-sanitize-memory-param-retval` to clang >=16, but probably only for the `ctime_test` target.

<details>
<summary>Example false positives, `clang version 17.0.6` on d8311688bd383d3a923a1b11789cded3cc8e5e03, and `./configure --enable-dev-mode CFLAGS='-fsanitize=memory -fsanitize-recover=memory -g'`:</summary>

```
> ./ctime_tests
==18035==WARNING: MemorySanitizer: use-of-uninitialized-value
    #0 0x77094683e2fa in secp256k1_ecdsa_sign_inner /home/tim/bs/dev/secp256k1/src/secp256k1.c:544:5
    #1 0x77094683c06f in secp256k1_ecdsa_sign /home/tim/bs/dev/secp256k1/src/secp256k1.c:556:11
    #2 0x5cb76630a11f in run_tests /home/tim/bs/dev/secp256k1/src/ctime_tests.c:106:11
    #3 0x5cb76630a11f in main /home/tim/bs/dev/secp256k1/src/ctime_tests.c:57:5
    #4 0x770946518ccf  (/usr/lib/libc.so.6+0x25ccf) (BuildId: c0caa0b7709d3369ee575fcd7d7d0b0fc48733af)
[...]

PR #1517, opened 2024-04-15
autotools: Disable eager MSan in ctime_tests

real-or-random:
This is the autotools solution for #1516. 

Alternatively, we could have a full-blown `--enable-msan` option, but it's more work, and I'm not convinced that it's necessary or at least much better.

hebasto If you're Concept ACK, are you willing to work on an equivalent PR for CMake?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -485,18 +485,24 @@ jobs:
       matrix:
         configuration:
           - env_vars:
+              CTIMETESTS: 'yes'
               CFLAGS: '-fsanitize=memory -fsanitize-recover=memory -g'
           - env_vars:
               ECMULTGENKB: 2
               ECMULTWINDOW: 2
+              CTIMETESTS: 'yes'
               CFLAGS: '-fsanitize=memory -fsanitize-recover=memory -g -O3'
+          - env_vars:
+              # -fsanitize-memory-param-retval is clang's default, but our build system disables it
+              # when ctime_tests when enabled.
+              CFLAGS: '-fsanitize=memory -fsanitize-recover=memory -fsanitize-memory-param-retval -g'
+              CTIMETESTS: 'no'
 
     env:
       ECDH: 'yes'
       RECOVERY: 'yes'
       SCHNORRSIG: 'yes'
       ELLSWIFT: 'yes'
-      CTIMETESTS: 'yes'
       CC: 'clang'
       SECP256K1_TEST_ITERS: 32
       ASM: 'no'
--- a/build-aux/m4/bitcoin_secp.m4
+++ b/build-aux/m4/bitcoin_secp.m4
@@ -45,6 +45,18 @@ fi
 AC_MSG_RESULT($has_valgrind)
 ])
 
+AC_DEFUN([SECP_MSAN_CHECK], [
+AC_MSG_CHECKING(whether MemorySanitizer is enabled)
+AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
+  #if defined(__has_feature)
+  #  if __has_feature(memory_sanitizer)
+  #    error "MemorySanitizer is enabled."
+  #  endif
+  #endif
+  ]])], [msan_enabled=no], [msan_enabled=yes])
+AC_MSG_RESULT([$msan_enabled])
+])
+
 dnl SECP_TRY_APPEND_CFLAGS(flags, VAR)
 dnl Append flags to VAR if CC accepts them.
 AC_DEFUN([SECP_TRY_APPEND_CFLAGS], [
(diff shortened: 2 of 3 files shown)


hebasto:
Concept ACK.

> @hebasto If you're Concept ACK, are you willing to work on an equivalent PR for CMake?

Sure. I am :)

hebasto (commented):
To have any effect, the `ctime_tests_CFLAGS` variable has to `AC_SUBST`ed.

However, to achieve the goal, the `-fno-sanitize-memory-param-retval` option has to be applied to the library code as well. So, I suggest:
```diff
--- a/configure.ac
+++ b/configure.ac
@@ -142,7 +142,7 @@ if test "x$GCC" = "xyes"; then
   # We need to wrap the argument in a --(start/end)-no-unused-arguments block to suppress a
   # -Wunused-command-line-argument warning, which clang emits if MSan is not actually enabled, i.e.,
   # if -fsanitize=memory does not appear on the command line.
-  SECP_TRY_APPEND_CFLAGS([--start-no-unused-arguments -fno-sanitize-memory-param-retval --end-no-unused-arguments], ctime_tests_CFLAGS)
+  SECP_TRY_APPEND_CFLAGS([--start-no-unused-arguments -fno-sanitize-memory-param-retval --end-no-unused-arguments], SECP_CFLAGS)
 fi
 
 ###
```

real-or-random:
> To have any effect, the `ctime_tests_CFLAGS` variable has to `AC_SUBST`ed.

No, this seems to happen automatically. `make V=1` shows me that it's effective.

> However, to achieve the goal, the `-fno-sanitize-memory-param-retval` option has to be applied to the library code as well. 

Okay, true. Sorry, I thought I had tested this before opening a PR.

Then this is a bit annoying. Your suggestion will work for our "abuse" of MSan in the ctime_tests. But it will also disable the new eager checking behind the back of the user in the case of normal use, i.e., checking for actual memory issues. 

So if we want to have both eager checking in "normal" MSan mode, and clean ctime_tests, I see no other way than having two builds (and adding a build to CI is not a problem). I think one of the following approaches will be sensible:
1. disable the ctime_tests if we find `-fsanitize=memory` or similar on the command line (but parsing this is complex in cases such as `-fsanitize=unreachable,memory` this may be fragile and overkill)
2. work with your suggestion to add this to `SECP_CFLAGS`, but only if constant-time tests are enabled. We could additionally print a warning that tells the users to disable the constant-time tests if they want to avoid that we add `-fno-sanitize-memory-param-retval`.

The second looks better to me. What do you think?

hebasto:
> > To have any effect, the `ctime_tests_CFLAGS` variable has to `AC_SUBST`ed.
> 
> No, this seems to happen automatically. `make V=1` shows me that it's effective.

I mean, only `ctime_tests.c` is compiled with `ctime_tests_CFLAGS`. However, all code is needed to be compiled with `-fno-sanitize-memory-param-retval`.

hebasto:
> 1. disable the ctime_tests if we find `-fsanitize=memory` or similar on the command line (but parsing this is complex in cases such as `-fsanitize=unreachable,memory` this may be fragile and overkill)

The following check might be used:
```m4
AC_DEFUN([SECP_MSAN_CHECK], [
AC_MSG_CHECKING(whether MemorySanitizer is enabled)
AC_COMPILE_IFELSE([AC_LANG_SOURCE([[
  #if defined(__has_feature)
  #  if __has_feature(memory_sanitizer)
  #    error "MemorySanitizer is enabled."
  #  endif
  #endif
  ]])], [msan_enabled=no], [msan_enabled=yes])
AC_MSG_RESULT([$msan_enabled])
])
```

real-or-random:
> 2\. work with your suggestion to add this to `SECP_CFLAGS`, but only if constant-time tests are enabled. We could additionally print a warning that tells the users to disable the constant-time tests if they want to avoid that we add `-fno-sanitize-memory-param-retval`.

I did this, using your suggested check for msan. Also rebased. Ready for review from my side.

Note: CI is currently failing on macOS. This should be resolved once the GitHub Actions image is updated to have brew 4.3.1, which has the fix (https://github.com/Homebrew/brew/pull/17336). The images are updated every week, so the issue should just disappear in a few days.

real-or-random:
Pushed another commit that adds a CI job with eager checks enabled explicitly (`-fsanitize-memory-param-retval`).

real-or-random:
Force pushed to fix a logic bug, should be really ready for review now. :)

hebasto (approved):
ACK ebfb82ee2f8c15ae6129932380b1dfb0942ea35a, tested on Ubuntu 24.04 with different clang versions (from 15 to 18) and different build configurations. CI changes look OK as well.

hebasto:
@real-or-random 
> @hebasto If you're Concept ACK, are you willing to work on an equivalent PR for CMake?

Please consider https://github.com/bitcoin-core/secp256k1/pull/1532.

PR #1518, opened 2024-04-16
extrakeys: add secp256k1_pubkey_sort

jonasnick:
This PR adds a  `secp256k1_pubkey_sort` function the the public API which was originally part of the musig PR (#1479). However, I opened a separate PR because it adds internal functions that are also used by the WIP silent payments module.


real-or-random:
Concept ACK. This is useful independently of MuSig and Silent Payments. Also for users with no standard library, e.g., on hardware wallets.

cc @roconnor-blockstream who wrote most of this code

---

Just out of curiosity, I checked what glibc actually does. They have Quicksort, Insertionsort, Mergesort, and Heapsort in their code base. [If malloc fails, they'll always fall back to Heapsort.](https://github.com/bminor/glibc/blob/14e56bd4ce15ac2d1cc43f762eb2e6b83fec1afe/stdlib/qsort.c#L361-L368) I think they had been using Quicksort and Mergesort for decades (perhaps selected depending on the platform?), but they changed their mind quite often lately, and this was indeed related to degenerate and adversarially constructed inputs: 
 -  Oct 2023: They added Heapsort and used it to implement Introsort. Introsort is basically Quicksort two fallbacks: 1) a fallback on some O(n log n) sorting algorithm in degenerate cases, and 2) a fallback on Insertion sort for small arrays (https://github.com/bminor/glibc/commit/274a46c9b25ab733a1fb9fb1497f1beecae30193). They had also removed Mergesort, making Introsort the algorithm used everywhere. (https://github.com/bminor/glibc/commit/03bf8357e8291857a435afcc3048e0b697b6cc04)
 -  Nov 2023: The logic for the fallback to Heapsort was tightened, and a test with a degerate array. (https://github.com/bminor/glibc/commit/64e4acf24da15c11cb83f933947df3b2e8a700cd: "To avoid a denial-of-service condition with adversarial input, it is necessary to fall over to heapsort if tail-recursing deeply, too [...]")
 - Jan 2024: They reverted to Mergesort because it turns out that people rely on the sorting being stable, even though this is not guaranteed by the standard, and even though their Heapsort fallback won't guarantee it anyway. (https://github.com/bminor/glibc/commit/709fbd3ec3595f2d1076b4fec09a739327459288)

This means that the current libc is probably fine. But old versions may be problematic, and other implementations of the standard library may also be problematic. If anything, all of this confirms that Heapsort is a good choice, and that we should not rely on the standard library here.

src/modules/extrakeys/Makefile.am.include
real-or-random, line 1: nit: No need to touch this file. Is this a leftover diff from when the sorting function was still in extrakeys (see https://github.com/bitcoin-core/secp256k1/pull/1479#discussion_r1457727273)? Also, the commit message still mentions "extrakeys".

src/hsort.h
```diff
@@ -0,0 +1,22 @@
...
+
+#include <stddef.h>
+#include <string.h>
+
+/* In-place, iterative heapsort with an interface matching glibc's qsort_r. This
+ * is preferred over standard library implementations because they generally
+ * make no guarantee about being fast for malicious inputs.
+ *
+ * See the qsort_r manpage for a description of the interface.
+ */
```
josibake, line 18: nit: Might be worth leaving a small reminder to the user that `hsort` is unstable. I had forgotten this detail, which caused me quite a bit of confusion. A small comment would either a) remind the user or b) prompt them to look up what that means before using and hopefully save some headaches.

josibake (commented):
Concept ACK. This is particularly useful for the silent payments module in that we are able to hide a lot of complexity from the caller. In addition, I agree with Tim's assessment that this is a generally useful thing to have (not specific to Musig2/Silent payments).

Left a small nit regarding documentation but overall I found this pretty easy to work with as implemented when creating a custom `cmp` for silent payments.

src/hsort.h
real-or-random: > * See the qsort_r manpage for a description of the interface.

By the way, we could also just spell out the interface in the comment. But the current version is also fine, it's an internal function in the end.

src/hsort_impl.h
```diff
@@ -0,0 +1,116 @@
...
+#ifndef SECP256K1_HSORT_IMPL_H
+#define SECP256K1_HSORT_IMPL_H
+
+#include "hsort.h"
+
+/* An array is a heap when, for all non-zero indexes i, the element at index i
+ * compares as less than or equal to the element at index parent(i) = (i-1)/2.
+ */
+
+static SECP256K1_INLINE size_t child1(size_t i) {
```
roconnor-blockstream, line 16: Do we need to prefix all these functions with 'secp256k1_'?

real-or-random: Yep, the child1/child2 functions should probably also get an additional `_heap` prefix, e.g., `secp256k1_heap_child1`

src/modules/extrakeys/Makefile.am.include
jonasnick: removed the file and the mention of extrakeys

src/hsort_impl.h
jonasnick: done

src/hsort_impl.h
```diff
@@ -0,0 +1,116 @@
...
+
+static SECP256K1_INLINE void secp256k1_heap_swap(unsigned char *a, size_t i, size_t j, size_t stride) {
+    while (64 < stride) {
+        secp256k1_heap_swap64(a + (stride - 64), i, j, 64);
+        stride -= 64;
+    }
+    secp256k1_heap_swap64(a, i, j, stride);
+}
+
+static SECP256K1_INLINE void heap_down(unsigned char *a, size_t i, size_t heap_size, size_t stride,
```
roconnor-blockstream, line 42: This needs a `secp256k1_` prefix too.

src/hsort_impl.h
```diff
@@ -0,0 +1,116 @@
...
+    memcpy(a + j*stride, tmp, stride);
+}
+
+static SECP256K1_INLINE void secp256k1_heap_swap(unsigned char *a, size_t i, size_t j, size_t stride) {
+    while (64 < stride) {
+        secp256k1_heap_swap64(a + (stride - 64), i, j, 64);
+        stride -= 64;
+    }
+    secp256k1_heap_swap64(a, i, j, stride);
+}
```
real-or-random, line 40: `secp256k1_heap_swap` is incorrect for `stride > 64`. See https://github.com/real-or-random/secp256k1/commit/95138ca6c1336bcd20ef332dd742f8a5f0f7cf75 for a fix and a test. This was overlooked because we never use such a stride, not even in the tests.

Alternatively, we could drop `secp256k1_heap_swap` entirely. `secp256k1_heap_swap64` appears to be correct. But I'd say let's just keep it if my fix looks good.

elichai:
A wild idea:
What if we used something simple(like shellsort) that is very easy to verify, and has good performance in the best case (when it's already sorted) and then we recommend users that use huge list of pub keys to sort them before passing them in with the stdlib sort function of the platform their using.

Reimplementing and maintaining highly efficient in place sorting every time is very sad

src/hsort_impl.h
roconnor-blockstream: FWIW, below is more like what I intended to write:

```c
static SECP256K1_INLINE void swap64(unsigned char *a, unsigned char *b, size_t len) {
    unsigned char tmp[64];
    VERIFY_CHECK(len <= 64);
    memcpy(tmp, a, len);
    memmove(a, b, len);
    memcpy(b, tmp, len);
}

static SECP256K1_INLINE void swap(unsigned char *arr, unsigned char *a, size_t stride) {
    unsigned char *a = arr + i*stride;
    unsigned char *b = arr + j*stride;
    size_t len = stride;
    while (64 < len) {
        swap64(a + (len - 64), b + (len - 64), 64);
        len -= 64;
    }
    swap64(a, b, len);
}
```

Obviously I've gotten this wrong before, so take it with a grain of salt.

real-or-random: Looks correct (except that `i` and `j` should be in the parameter instead of `a`) to me, and is easier to read than my version, I think.

real-or-random:
> A wild idea: What if we used something simple(like shellsort or even bubblesort) that is very easy to verify, and has good performance in the best case (when it's already sorted) and then we recommend users that use huge list of pub keys to sort them before passing them in with the stdlib sort function of the platform their using.

Hm, I think this gives us the worst of both worlds in the end:
 - We have to maintain the implementation of sorting algorithm.
 - We still tell users to rely on the stdlib.

Also, I doubt that Shellsort or Bubblesort are much easier to implement than Heapsort. (And look at the nature of the bug I found: The swap was wrong, not the Heapsort itself. And you'll need a swap for every algorithm.)

> 
> Needing to reimplement and maintaining highly efficient in place sorting every time is very sad

Yeah, but that's because C is sad (sometimes).

src/hsort_impl.h
jonasnick: Oops. Good catch @real-or-random. I cherry picked your test, copied @roconnor-blockstream's fix and improved the existing test.

src/tests.c
```diff
@@ -3675,6 +3675,77 @@ static void run_inverse_tests(void)
...
+    CHECK(data.counter >= NUM);
+    test_hsort_is_sorted(elements, NUM, element_len);
+
+    /* Test hsort with array of random length n */
+    for (i = 0; i < COUNT; i++) {
+        int n = secp256k1_testrand_int(NUM);
+        secp256k1_testrand_bytes_test(elements, n*element_len);
+        secp256k1_hsort(elements, n, element_len, test_hsort_cmp, &data);
+        test_hsort_is_sorted(elements, n, element_len);
+    }
```
real-or-random, line 3736: suggestion: Also call the C library `qsort_r` and compare the results. This is a very effective test at the cost of just a handful of code lines.

src/tests.c
```diff
@@ -6607,6 +6678,161 @@ static void run_pubkey_comparison(void) {
...
+    memset(&pks[0], 0, sizeof(pks[0]));
+    CHECK_ILLEGAL_VOID(CTX, CHECK(secp256k1_ec_pubkey_sort(CTX, pks_ptr, 2) == 1));
+    memset(&pks[1], 0, sizeof(pks[1]));
+    {
+        int32_t ecount = 0;
+        secp256k1_context_set_illegal_callback(CTX, counting_callback_fn, &ecount);
+        CHECK(secp256k1_ec_pubkey_sort(CTX, pks_ptr, 2) == 1);
+        CHECK(ecount == 2);
+        secp256k1_context_set_illegal_callback(CTX, NULL, NULL);
+    }
```
real-or-random, line 6736: Not exactly this PR, but do you think it's worth noting in the API docs that callbacks may be triggered more than once  per API call? This may be interesting in the case of the illegal callback because you wouldn't need to crash necessarily for this type of callback (e.g., thinking of python bindings that could choose to raise). If yes, we could track this separately.

real-or-random (commented):
ACK mod nits, I've inspected the code and the proofs again.

New improved tests are nice.

src/tests.c
jonasnick: To test that giving exactly the same arguments to `qsort_r` gives the same result? Unfortunately, `qsort_r` is not part of the standard lib.

jonasnick:
I added a description of the interface to the hsort doc. Apparently mac OS [uses a different order of arguments](https://stackoverflow.com/questions/39560773/different-declarations-of-qsort-r-on-mac-and-linux).

src/tests.c
jonasnick: Hm good point. But if we think that calling the callback multiple times causes problems, it may makes more sense to fix/change that because it's unlikely someone will read this and take it into account.

For the error callback the documentation says "After this callback returns, anything may happen, including crashing." and for the illegal callback  "When this callback is triggered, the API function called is guaranteed not to cause a crash, though its return value and output arguments are undefined."

src/tests.c
```diff
@@ -3675,6 +3675,77 @@ static void run_inverse_tests(void)
...
+    VERIFY_CHECK(element_len <= MAX_ELEMENT_LEN);
+    data.counter = 0;
+    data.element_len = element_len;
+
+    secp256k1_hsort(elements, 0, element_len, test_hsort_cmp, &data);
+    CHECK(data.counter == 0);
+    secp256k1_hsort(elements, 1, element_len, test_hsort_cmp, &data);
+    CHECK(data.counter == 0);
+    secp256k1_hsort(elements, NUM, element_len, test_hsort_cmp, &data);
+    CHECK(data.counter >= NUM);
```
roconnor-blockstream, line 3728: Not entirely sure what we are trying to do with this check, but if we are thinking of hsort as a black box sorting function, the minimum number of comparisons is `NUM-1`.

roconnor-blockstream:
> Doing the sorting for the users eliminates a difficult to catch error and we'd need to sort anyways to be able to detect and alert the user that they did not sort (or sorted incorrectly).

You can test that a list is sorted in O(n) time, but sorting (by comparisons) take O(n log n) time

That said, I do agree that we should just sort ourselves, in O(n log n) worst case time, with constant memory overhead.

src/tests.c
real-or-random: > Unfortunately, `qsort_r` is not part of the standard lib.

Yes, okay, I see. I think we could implement a test also with `qsort` if we add an adapter comparison function. But then it's a few lines of code more. It certainly won't hurt, but then it's not entirely clear if it's worth the hassle and gives us much more than the simple check that the output array is sorted.

src/tests.c
real-or-random: > But if we think that calling the callback multiple times causes problems, it may makes more sense to fix/change that because it's unlikely someone will read this and take it into account.

Let's maybe track this in a separate issue. It's certainly not great to call the callback twice, but last time when I tried to "fix" it, I wasn't happy with the added code complexity.

jonasnick:
Rebased on master and added change log entry.

src/tests.c
jonasnick: If we can't test against the API of `qsort_r` then I doubt that's worth it. Note that we now spell out the interface of hsort explicitly instead of referring to `qsort_r`.

src/hsort_impl.h
```diff
@@ -0,0 +1,119 @@
...
+         * If [child1(i)] > [i] and [child2(i)] > [i],
+         *   swap [i] with the larger child to ensure the new parent is larger
+         *   than both children. When [child1(i)] == [child2(i)], swap [i] with
+         *   [child2(i)].
+         * Else if [child1(i)] > [i], swap [i] with [child1(i)].
+         * Else if [child2(i)] > [i], swap [i] with [child2(i)].
+         */
+        if (secp256k1_heap_child2(i) < heap_size
+                && 0 <= cmp(arr + secp256k1_heap_child2(i)*stride, arr + secp256k1_heap_child1(i)*stride, cmp_data)) {
+            if (0 < cmp(arr + secp256k1_heap_child2(i)*stride, arr +         i*stride, cmp_data)) {
```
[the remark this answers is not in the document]
jonasnick: fixed

src/hsort_impl.h
```diff
@@ -0,0 +1,119 @@
...
+    unsigned char *b = arr + j*stride;
+    size_t len = stride;
+    while (64 < len) {
+        secp256k1_heap_swap64(a + (len - 64), b + (len - 64), 64);
+        len -= 64;
+    }
+    secp256k1_heap_swap64(a, b, len);
+}
+
+static SECP256K1_INLINE void secp256k1_heap_down(unsigned char *arr, size_t i, size_t heap_size, size_t stride,
```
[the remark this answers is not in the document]
jonasnick: done

src/secp256k1.c
```diff
@@ -325,6 +326,40 @@ int secp256k1_ec_pubkey_cmp(const secp256k1_context* ctx, const secp256k1_pubkey
     return secp256k1_memcmp_var(out[0], out[1], sizeof(out[0]));
 }
 
+/* This struct wraps a const context pointer to satisfy the secp256k1_hsort api
```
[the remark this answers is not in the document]
jonasnick: Changed it to just cast away const now.

src/secp256k1.c
```diff
@@ -325,6 +326,34 @@ int secp256k1_ec_pubkey_cmp(const secp256k1_context* ctx, const secp256k1_pubkey
     return secp256k1_memcmp_var(out[0], out[1], sizeof(out[0]));
 }
 
+static int secp256k1_ec_pubkey_sort_cmp(const void* pk1, const void* pk2, void *ctx) {
+    return secp256k1_ec_pubkey_cmp((secp256k1_context *)ctx,
+                                     *(secp256k1_pubkey **)pk1,
+                                     *(secp256k1_pubkey **)pk2);
```
josibake, line 332: In #1519, I initially created a `secp256k1_silentpayments_recipient_sort_cmp` function using `(secp256k1_silentpayments_recipient *)r->scan_pubkey` instead of `*(secp256k1_silentpayments_recipient **)r->scan_pubkey`. Everything compiled fine, but the end result was my array was not sorted correctly. Took me awhile to figure out that it was due to using `(typedef *)` instead of `*(typedef **)`.

This could just be a me problem in not having a strong C background, but wanted to mention it in case its worth documenting.

josibake (approved):
ACK https://github.com/bitcoin-core/secp256k1/pull/1518/commits/7d2591ce12d8a9b85f210cf9d678e91cee125ee9

Don't have much to offer on the `hsort` side, but did review the `secp256k1` specific parts and have been using this commit in #1519 without any issues.

src/secp256k1.c
jonasnick: Hm, this is documented. The hsort doc says "comparison function is called with two arguments that point to the objects being compared" (which was copied from the qsort_r manpage). If the array you give to hsort consists of elements of type `secp256k1_silentpayments_recipient *`, the comparison function is called with pointers to the elements (`secp256k1_silentpayments_recipient **`).

real-or-random: I guess seeing stuff like `*(secp256k1_pubkey **)pk1` can be confusing even to seasoned C programmers at first glance, due to the asterisk being used for two related but different things here (first as a dereference operator and then in the cast for indicating a pointer type). But yeah, I don't think we can do anything to improve this.

src/secp256k1.c
```diff
@@ -325,6 +326,34 @@ int secp256k1_ec_pubkey_cmp(const secp256k1_context* ctx, const secp256k1_pubkey
...
+    #endif
+
+    /* Casting away const is fine because neither secp256k1_hsort nor
+     * secp256k1_ec_pubkey_sort_cmp modify the data pointed to by the cmp_data
+     * argument. */
+    secp256k1_hsort(pubkeys, n_pubkeys, sizeof(*pubkeys), secp256k1_ec_pubkey_sort_cmp, (void *)ctx);
+
+    #if defined(_MSC_VER) && (_MSC_VER < 1933)
+    #pragma warning(pop)
+    #endif
```
real-or-random, line 352: I verified that the pragmas are still necessary now that the call to secp256k1_hsort was changed. See https://godbolt.org/z/13MY3acP1 .

real-or-random (approved):
ACK 7d2591ce12d8a9b85f210cf9d678e91cee125ee9

PR #1524, opened 2024-05-06
check-abi: explicitly provide public headers

jonasnick:
Without this commit, the check-abi shell script outputs false positives because it consider some headers public that are actually not public.


real-or-random (approved):
ACK https://github.com/bitcoin-core/secp256k1/pull/1524/commits/dd695563e68d605b837b7813ae7bfafbf866ddff

real-or-random:
cc @hebasto

hebasto (approved):
ACK dd695563e68d605b837b7813ae7bfafbf866ddff, tested on Ubuntu 24.04.

Issue #1527, opened 2024-05-08
tests: `secp256k1_ecmult_multi_var` is called with a `NULL` error callback


niooss-ledger:
Hello,

In the tests, function `test_ecmult_accumulate` calls `secp256k1_ecmult_multi_var` with `error_callback = NULL` (since version 0.2.0, PR https://github.com/bitcoin-core/secp256k1/pull/920): https://github.com/bitcoin-core/secp256k1/blob/7712a53061b1e36ecf47a3a46ea1e67ef31904d9/src/tests.c#L5497-L5498

This function eventually calls `secp256k1_scratch_max_allocation`: https://github.com/bitcoin-core/secp256k1/blob/7712a53061b1e36ecf47a3a46ea1e67ef31904d9/src/scratch_impl.h#L58-L60

... which directly dereferences the callback parameter: https://github.com/bitcoin-core/secp256k1/blob/7712a53061b1e36ecf47a3a46ea1e67ef31904d9/src/util.h#L86-L87

In short, it seems `secp256k1_ecmult_multi_var` does not expect `error_callback` to be `NULL`.

The consequence of `test_ecmult_accumulate` not following this expectation would be a possible crash (by null pointer dereference) if something ever go wrong in the test. While this bug does not directly impact `secp256k1` library (it occurs in the test suite), I believe this issue should be fixed because I think tests should follow the calling convention of the library functions (such as not passing `NULL` where functions expects non-`NULL` parameters).

Moreover, `CHECK()` could probably be added to verify the result of `secp256k1_ecmult_multi_var`. Therefore, I am suggesting this change:

```diff
diff --git a/src/tests.c b/src/tests.c
index 2eb3fbfdcea7..dab47608c2e5 100644
--- a/src/tests.c
+++ b/src/tests.c
@@ -5494,8 +5494,8 @@ static void test_ecmult_accumulate(secp256k1_sha256* acc, const secp256k1_scalar
     secp256k1_ecmult_gen(&CTX->ecmult_gen_ctx, &rj1, x);
     secp256k1_ecmult(&rj2, &gj, x, &secp256k1_scalar_zero);
     secp256k1_ecmult(&rj3, &infj, &secp256k1_scalar_zero, x);
-    secp256k1_ecmult_multi_var(NULL, scratch, &rj4, x, NULL, NULL, 0);
[...]

PR #1528, opened 2024-05-08
tests: call `secp256k1_ecmult_multi_var` with a non-`NULL` error callback

niooss-ledger:
Hello,
This Pull Request fixes the issue reported in https://github.com/bitcoin-core/secp256k1/issues/1527. Function `secp256k1_ecmult_multi_var` expects to be called with a non-`NULL` `error_callback` parameter. Fix the invocation in `test_ecmult_accumulate` to do this. While at it, wrap the call in a `CHECK` macro to ensure it succeeds.


real-or-random (approved):
utACK 9554362b1532d8292ab805cefb9c886cd55ac205

siv2r (commented):
ACK 9554362, I have also verified that other invocations of `ecmult_multi_var` (in tests) don’t use `NULL` for the error callback function argument.

PR #1532, opened 2024-05-26
cmake: Disable eager MSan in ctime_tests

hebasto:
Same as https://github.com/bitcoin-core/secp256k1/pull/1517, but for the CMakle build system.

The second commit improves the configure summary (similar to https://github.com/hebasto/bitcoin/pull/189.


real-or-random (commented):
Concept ACK

cmake/CheckMemorySanitizer.cmake
```diff
@@ -0,0 +1,16 @@
+include_guard(GLOBAL)
+include(CheckCSourceCompiles)
+
+function(check_memory_sanitizer output)
+  set(CMAKE_TRY_COMPILE_TARGET_TYPE STATIC_LIBRARY)
+  check_c_source_compiles("
+    #if defined(__has_feature)
+    #  if __has_feature(memory_sanitizer)
+    //   MemorySanitizer is enabled.
```
real-or-random, line 9: nit: We should probably use the same test program on configure and CMake (and perhaps avoid `//` comments which are invalid in C89, though that really is more of a philosophical convern.)

CMakeLists.txt
```diff
@@ -263,6 +263,17 @@ endif()
 
 set(CMAKE_C_VISIBILITY_PRESET hidden)
 
+set(print_msan_notice)
+if(SECP256K1_BUILD_CTIME_TESTS AND CMAKE_C_COMPILER_ID MATCHES "^(Clang|AppleClang)$")
```
fanquake, line 267: Does Apple Clang makes sense here?

fanquake: Actually, why are you hard coding compiler IDs at all? This just makes things less flexible, and the check should be enough.

real-or-random: Yeah, I had the same thought. It saves a test compilation for GCC, but it's probably simpler to drop the check for clang. In https://github.com/bitcoin-core/secp256k1/pull/1517, I also added a similar check (but for "GCC or Clang", so it's much more useless). Let's just drop it.

hebasto:
Addressed comments:
- https://github.com/bitcoin-core/secp256k1/pull/1532#discussion_r1615986643:
> avoid `//` comments which are invalid in C89

- https://github.com/bitcoin-core/secp256k1/pull/1532#discussion_r1615994037
> Actually, why are you hard coding compiler IDs at all? This just makes things less flexible, and the check should be enough.

cmake/CheckMemorySanitizer.cmake
hebasto: > nit: We should probably use the same test program on configure and CMake

I'd prefer align Autotools' `SECP_MSAN_CHECK` with CMake's `check_memory_sanitizer` to keep the latter straightforward without negated values.

> ... and perhaps avoid `//` comments which are invalid in C89

Thanks. [Fixed](https://github.com/bitcoin-core/secp256k1/pull/1532#issuecomment-2133405610).

CMakeLists.txt
hebasto: > Actually, why are you hard coding compiler IDs at all? This just makes things less flexible, and the check should be enough.

Thanks! The compiler ID check has been [deleted](https://github.com/bitcoin-core/secp256k1/pull/1532#issuecomment-2133405610).

cmake/CheckMemorySanitizer.cmake
real-or-random: > I'd prefer to align Autotools' `SECP_MSAN_CHECK` code with CMake's `check_memory_sanitizer` to keep the latter straightforward without negated values.

Fine with that if you want to add a commit that makes them consistent, but then we'll also need to `#error` if `!defined(__has_feature)`.

CMakeLists.txt
real-or-random: Thanks, do you want to add a commit that also removes the `$GCC` check in configure.ac, just to keep the system consistent?

hebasto:
Addressed comments:
- https://github.com/bitcoin-core/secp256k1/pull/1532#discussion_r1616380720:
> ... we'll also need to `#error` if `!defined(__has_feature)`.

- https://github.com/bitcoin-core/secp256k1/pull/1532#discussion_r1616382346:
> ... to add a commit that also removes the $GCC check in configure.ac, just to keep the system consistent

cmake/CheckMemorySanitizer.cmake
hebasto: > ... but then we'll also need to `#error` if `!defined(__has_feature)`.

Thanks! [Fixed](https://github.com/bitcoin-core/secp256k1/pull/1532#issuecomment-2134674954).

CMakeLists.txt
hebasto: > Thanks, do you want to add a commit that also removes the `$GCC` check in configure.ac, just to keep the system consistent?

Sure. [Added](https://github.com/bitcoin-core/secp256k1/pull/1532#issuecomment-2134674954).

real-or-random (approved):
ACK f55703ba49454fc46226f4846fe292d4a3dfa3ef

Issue #1491, opened 2024-02-01
tests: Tidy the util functions


real-or-random:
When I look at all the helper functions near the top of `tests.c`, there are still some things that could be improved:
 - A lot of these functions could be moved to `testutil` or `testrand`, also functions like `random_fe`, `random_fe_magnitude`. They're currently only in `tests.c` but they're utils in the end.
 - We have functions with `group_element` and `field_element`, but they should just be `ge` and `fe`.
 - The ones in `testutil` should probably also get a `secp256k1_testutil` prefix then. You could say it's good that they stand out currently, but at least that's inconsistent with the ones in `testrand`, which have a prefix. Or perhaps better: we use prefixes `testutil_` and `testrand_`, dropping the additional `secp256k1_` for test functions. Then they still stand out. (The main purpose of `secp256k1_` is to avoid namespace collisions if someone uses the library by just including it, but that doesn't matter in test-only code.)

Noticed when looking into https://github.com/bitcoin-core/secp256k1/pull/1489.

fanquake:
What is left to do here post #1533 ?

PR #1533, opened 2024-05-27
tests: refactor: tidy up util functions (#1491)

theStack:
This PR is an attempt at tidying up test util functions, as suggested in #1491. The following changes are done:
* rename `_group_element...` functions to `_ge...`
* rename `_field_element...` functions to `_fe...`
* move `random_` helpers from tests.c to testutil.h (the alternative would be testrand.h, but to my understanding, this one is meant to contain the actual RNG implementation rather than helpers using it; happy to move the helpers there if that is preferred though)
* prefix testutil.h functions with `testutil_`
* prefix testrand.h functions with `testrand_` (this is currently done in a sloppy way by simply dropping the `secp256k1_` prefix, so some functions don't have the full prefix, like e.g. `testrand256`; naming suggestions welcome)


real-or-random (approved):
utACK e73f6f8fd9158a044c6fb22653105446f3542bb0

real-or-random:
> What is up with the macOS CI?

See https://github.com/bitcoin-core/secp256k1/issues/1536

real-or-random:
@sipa Want to review this?

PR #1535, opened 2024-05-27
build: Replace hardcoded "auto" value with default one

hebasto:
"auto" implies that a value is being chosen based on build system introspection or host system capabilities. However, for the `--with-ecmult-window` and `--with-ecmult-gen-kb` options, the values "auto" are hardcoded, which might lead to confusion.

This PR replaces "auto" with more appropriate default values.

If Concept ACKed, I'll add equivalent commits for CMake.


real-or-random:
Concept ACK from my side 

I think having a default is good enough, we don't need two layers of indirection (default -> auto -> some number).

hebasto:
> If Concept ACKed, I'll add equivalent commits for CMake.

The equivalent commits for CMake have been added.

real-or-random (approved):
utACK 4d9645bee06e732f5d7d9e72b0c26c22c5006eb8 good from my side, but let's see if we can get more (Concept) ACKs

I think we'd still want to have something like profiles (e.g., "desktop", "embedded", "minimal") eventually, but this change makes sense independently of this.

Nichebiche:
> "auto" implies that a value is being chosen based on build system introspection or host system capabilities. However, for the `--with-ecmult-window` and `--with-ecmult-gen-kb` options, the values "auto" are hardcoded, which might lead to confusion.
> 
> 
> 
> This PR replaces "auto" with more appropriate default values.
> 
> 
> 
> If Concept ACKed, I'll add equivalent commits for CMake.

real-or-random:
Does this need a changelog entry? Strictly speaking yes because this removes a valid config option. But pragmatically, I tend to say no: the breakage occurs only for people who passed `=auto` (the default) explicitly, so this should hardly affect anyone.

PR #1546, opened 2024-06-21
cmake: Rename `SECP256K1_LATE_CFLAGS` and switch to Bitcoin Core's approach

hebasto:
This PR address this https://github.com/hebasto/bitcoin/issues/239#issuecomment-2182713690:
> For consistency with libsecp256k1:
> 
> > > Is this code block supposed to achieve the same as our `SECP256K1_LATE_CFLAGS` (implemented by a user-defined function `all_targets_add_compile_options`) in libsecp256k1?
> > 
> > 
> > It is. But this approach guaranties to override even options that are abstracted by CMake, for instance [#157 (comment)](https://github.com/hebasto/bitcoin/pull/157#issuecomment-2090465123).
> 
>    * If we agree that appending to rule variables is superior, should we also do this in libsecp256k1?
> 
>    * And/or should we rename the `SECP256K1_LATE_CFLAGS` variable to `APPEND_CFLAGS`?


hebasto:
@real-or-random 

Your other comments from the discussion in https://github.com/hebasto/bitcoin/issues/239 have also been included.

CMakeLists.txt
```diff
@@ -284,14 +284,16 @@ if(SECP256K1_BUILD_BENCHMARK OR SECP256K1_BUILD_TESTS OR SECP256K1_BUILD_EXHAUST
   enable_testing()
 endif()
 
-set(SECP256K1_LATE_CFLAGS "" CACHE STRING "Compiler flags that are added to the command line after all other flags added by the build system.")
-include(AllTargetsCompileOptions)
+set(SECP256K1_APPEND_CFLAGS "" CACHE STRING "Compiler flags that are appended to the command line after all other flags added by the build system. It is intended for debugging and special builds.")
```
real-or-random, line 287: nit:
```suggestion
set(SECP256K1_APPEND_CFLAGS "" CACHE STRING "Compiler flags that are appended to the command line after all other flags added by the build system. This variable is intended for debugging and special builds.")
```

real-or-random (commented):
ACK mod nit

CMakeLists.txt
hebasto: Reworked. Not using "global override" as it can be confusing in cases when libsecp is integrated by an upstream project.

real-or-random (approved):
utACK 4706be2cd0bcac97da10ed3b98790f1ac6f04efb

hebasto:
Friendly ping @theuni :)

fanquake:
Concept ACK. Aligning the naming is nice and this implementation is simpler. I think "special builds" is ambiguous, and downstreams may likely (mis)use this option in any case, but I can't really think of anything better to call it.

PR #1547, opened 2024-06-22
ci: Switch from "RelWithDebInfo" to "Release" config for MSVC jobs

hebasto:
It is [reasonable](https://github.com/bitcoin-core/secp256k1/pull/1543#issuecomment-2176914205) to test a configuration that is used in the Bitcoin Core build.

For MSVC, the "Release" config (`/O2 /Ob2`) is preferable over "RelWithDebInfo" (`/Zi /O2 /Ob1`) for the following reasons:

1. `/Ob2` can be slightly better than `/Ob1` for performance; see:
- https://learn.microsoft.com/en-us/cpp/build/reference/ob-inline-function-expansion

3. `/Zi`, which produces a separate PDB file that contains all the symbolic debugging information, is incompatible with `ccache`; see:
- https://learn.microsoft.com/en-us/cpp/build/reference/z7-zi-zi-debug-information-format
- https://github.com/ccache/ccache/wiki/MS-Visual-Studio

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -735,20 +735,20 @@ jobs:
         run: cmake -E env CFLAGS="/WX ${{ matrix.configuration.cpp_flags }}" cmake -B build -DSECP256K1_ENABLE_MODULE_RECOVERY=ON -DSECP256K1_BUILD_EXAMPLES=ON ${{ matrix.configuration.cmake_options }}
 
       - name: Build
-        run: cmake --build build --config RelWithDebInfo -- /p:UseMultiToolTask=true /maxCpuCount
+        run: cmake --build build --config Release -- /p:UseMultiToolTask=true /maxCpuCount
 
       - name: Binaries info
         # Use the bash shell included with Git for Windows.
         shell: bash
         run: |
-          cd build/src/RelWithDebInfo && file *tests.exe bench*.exe libsecp256k1-*.dll || true
+          cd build/src/Release && file *tests.exe bench*.exe libsecp256k1-*.dll || true
 
       - name: Check
         run: |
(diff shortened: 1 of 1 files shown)


fanquake:
> It is https://github.com/bitcoin-core/secp256k1/pull/1543#issuecomment-2176914205 to test a configuration that is used in the Bitcoin Core build.

Isn't our build going to be `RelWithDebInfo` once we switch to using the secp256k1 build system? Not sure it's worth it to switch this now, just to switch it back again very shortly after?

> /Zi, which produces a separate PDB file that contains all the symbolic debugging information, is incompatible with ccache; see:

Why is that relevant here (I can't see any ccache usage)?

hebasto:
> > It is [#1543 (comment)](https://github.com/bitcoin-core/secp256k1/pull/1543#issuecomment-2176914205) to test a configuration that is used in the Bitcoin Core build.
> 
> Isn't our build going to be `RelWithDebInfo` once we switch to using the secp256k1 build system? Not sure it's worth it to switch this now, just to switch it back again very shortly after?

I think that Bitcoin Core should use the "Release" config when building with MSVC for the reasons mentioned in the PR description.

> > /Zi, which produces a separate PDB file that contains all the symbolic debugging information, is incompatible with ccache; see:
> 
> Why is that relevant here (I can't see any ccache usage)?

We use ccache in the CI [job](https://github.com/bitcoin/bitcoin/actions/runs/9613953299/job/26517746715), which saves us up to 13 minutes of build time.

fanquake:
> I think that Bitcoin Core should use the "Release" config when building with MSVC for the reasons mentioned in the PR description.

Seems odd we'd make an exception in regard to build types for MSVC. If that flag really is that important, and the only difference, we can just pass it manually?

> We use ccache in the CI [job](https://github.com/bitcoin/bitcoin/actions/runs/9613953299/job/26517746715), which saves us up to 15 minutes of build time.

That isn't in this repo.

PR #1551, opened 2024-06-25
Add ellswift usage example

theStack:
This should hopefully be useful as orientation for users implementing the shared secret derivation part of BIP324. Conceptually the example is not very different to the ECDH one, so a lot of code/comments are just copied (e.g. context creation, secret key generation, shared secret comparison, console output, cleanup with secret key clearing).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ ctime_tests
 ecdh_example
 ecdsa_example
 schnorr_example
+ellswift_example
 *.exe
 *.so
 *.a
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+#### Added
+ - Added usage example for an ElligatorSwift key exchange.
+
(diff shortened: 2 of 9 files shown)


real-or-random (commented):
Concept ACK

.gitignore
```diff
@@ -10,6 +10,7 @@ ctime_tests
 ecdh_example
 ecdsa_example
 schnorr_example
+ellswift_example
```
hebasto, line 13: nit: I understand that this doesn’t adhere to the surrounding code, but, strictly speaking, it should be
```suggestion
/ellswift_example
```

hebasto (commented):
31f84595c4fa8fcc89021790257b85416aae44f6

I have reviewed both build systems changes and they look OK.

real-or-random (approved):
utACK 31f84595c4fa8fcc89021790257b85416aae44f6

.gitignore
real-or-random: Yeah, feel free to adjust this for the binaries, but I guess either variant is fine (we also have the pattern `*.exe`, which may in rare cases a bit too coarse`...)

jonasnick (approved):
ACK 31f84595c4fa8fcc89021790257b85416aae44f6

PR #1554, opened 2024-06-26
cmake: Clean up testing code

hebasto:
1. Delete `CTest` module.

The `CTest` module handles `CDash` integration, which we do not use. It is not required for testing functionality.

2. Clean up cases when to invoke `enable_testing()`

The `enable_testing()` command invocation is required for `add_test()` commands, which are used only for `{noverify_}tests`, `exhaustive_tests` and examples.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -269,9 +269,6 @@ if(SECP256K1_BUILD_CTIME_TESTS)
   unset(msan_enabled)
 endif()
 
-include(CTest)
-# We do not use CTest's BUILD_TESTING because a single toggle for all tests is too coarse for our needs.
-mark_as_advanced(BUILD_TESTING)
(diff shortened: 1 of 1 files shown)


real-or-random:
> The `enable_testing()` command invocation is required for `add_test()` commands, which are used only for `{noverify_}tests`, `exhaustive_tests` and examples.

I don't think it's required. It's just that `add_test()` has no effect without `enable_testing()`.

`enable_testing()` seems cheap. Couldn't we just run it always? Or do you think this has drawbacks?

hebasto:
> > The `enable_testing()` command invocation is required for `add_test()` commands, which are used only for `{noverify_}tests`, `exhaustive_tests` and examples.
> 
> I don't think it's required. It's just that `add_test()` has no effect without `enable_testing()`.
> 
> `enable_testing()` seems cheap.

I agree.

> Couldn't we just run it always?

Sure. The code has been adjusted.

> Or do you think this has drawbacks?

I did some research and I haven't found any drawbacks. When `enable_testing()` is invoked and no tests being added, the created `CTestTestfile.cmake` files in the binary tree are noop/empty.

real-or-random (approved):
utACK 7c987ec89e6cbee5d087683ba29b97012e679484

Issue #1556, opened 2024-06-28
Is default `-Wl,--no-undefined` desired?


hebasto:
Consider the following diff:
```diff
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -765,7 +765,9 @@ int secp256k1_ec_pubkey_tweak_mul(const secp256k1_context* ctx, secp256k1_pubkey
     return ret;
 }
 
+void undefined_function(void);
 int secp256k1_context_randomize(secp256k1_context* ctx, const unsigned char *seed32) {
+    undefined_function();
     VERIFY_CHECK(ctx != NULL);
     ARG_CHECK(secp256k1_context_is_proper(ctx));
 
```

It compiles fine:
```
$ ./autogen.sh
$ ./configure
$ make libsecp256k1.la
```

However, it should fail like that:
```
$ ./autogen.sh
$ ./configure LDFLAGS="-Wl,--no-undefined"
$ make libsecp256k1.la
  CC       src/libsecp256k1_la-secp256k1.lo
  CC       src/libsecp256k1_precomputed_la-precomputed_ecmult.lo
  CC       src/libsecp256k1_precomputed_la-precomputed_ecmult_gen.lo
  CCLD     libsecp256k1_precomputed.la
  CCLD     libsecp256k1.la
/usr/bin/ld: src/.libs/libsecp256k1_la-secp256k1.o: in function `secp256k1_context_randomize':
/home/hebasto/git/secp256k1/secp256k1/src/secp256k1.c:770:(.text+0xee01): undefined reference to `undefined_function'
collect2: error: ld returned 1 exit status
make: *** [Makefile:1020: libsecp256k1.la] Error 1
```

real-or-random:
I think we want to check for undefined symbols, and `-Wl,--no-undefined` seems to be the canonical way of doing so.

I assume one important exception is builds with external default callbacks, so we can't enable it in these cases.

PR #1559, opened 2024-06-28
cmake: Add `--no-undefined` linker option

hebasto:
This PR adds the `--no-undefined` option to a linker, which supports it.

From the GNU ld [docs](https://sourceware.org/binutils/docs/ld.html):
> `--no-undefined`
>
> Report unresolved symbol references from regular object files. This is done even if the linker is creating a non-symbolic shared library.

Closes https://github.com/bitcoin-core/secp256k1/issues/1556.

~Looking for Concept (N)ACKs. Once ACKed, I'll add the same functionality to the Autotools-based build system.~

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -269,6 +269,11 @@ if(SECP256K1_BUILD_CTIME_TESTS)
   unset(msan_enabled)
 endif()
 
+if(NOT SECP256K1_USE_EXTERNAL_DEFAULT_CALLBACKS)
+  include(TryAppendLinkerFlags)
+  try_append_linker_flags(-Wl,--no-undefined)
+endif()
+
 include(CTest)
 # We do not use CTest's BUILD_TESTING because a single toggle for all tests is too coarse for our needs.
 mark_as_advanced(BUILD_TESTING)
--- a/cmake/TryAppendLinkerFlags.cmake
+++ b/cmake/TryAppendLinkerFlags.cmake
@@ -1,0 +1,32 @@
+include_guard(GLOBAL)
+
+include(CheckCSourceCompiles)
+
+function(secp256k1_check_linker_flags_internal flags output)
+  string(MAKE_C_IDENTIFIER "${flags}" result)
+  string(TOUPPER "${result}" result)
+  set(result "LINKER_SUPPORTS_${result}")
+  set(CMAKE_REQUIRED_FLAGS "${flags}")
+  if(MSVC)
+    string(APPEND " /WX")
+  elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    string(APPEND " -Wl,-fatal_warnings")
+  else()
+    string(APPEND " -Wl,--fatal-warnings")
+  endif()
+
+  # This ensures running a linker.
+  set(CMAKE_TRY_COMPILE_TARGET_TYPE EXECUTABLE)
+  check_c_source_compiles("int main() { return 0; }" ${result})
+
+  set(${output} ${${result}} PARENT_SCOPE)
+endfunction()
+
+# Append flags to the LINK_OPTIONS directory property if a linker accepts them.
+macro(try_append_linker_flags)
+  secp256k1_check_linker_flags_internal("${ARGV}" result)
+  if(result)
+    add_link_options(${ARGV})
+  endif()
+  unset(result)
+endmacro()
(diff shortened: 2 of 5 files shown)


theuni:
Concept ACK.

I think we might be after `-Wl,--no-allow-shlib-undefined`, though, and only for the shared lib as that's what we care about. `--no-undefined` is strictly stronger, I guess, and should also be satisfied for the binaries. So I suppose the sledgehammer approach here should be fine.

hebasto:
> I think we might be after `-Wl,--no-allow-shlib-undefined`...

From my experience, it fails to catch an issue like one I demoed in https://github.com/bitcoin-core/secp256k1/issues/1556.

real-or-random:
> I think we might be after `-Wl,--no-allow-shlib-undefined`, though, and only for the shared lib as that's what we care about. `--no-undefined` is strictly stronger, I guess, and should also be satisfied for the binaries. So I suppose the sledgehammer approach here should be fine.

If https://stackoverflow.com/a/2356407 is right, then `--no-undefined` covers objects created by the linker and  `--no-allow-shlib-undefined` covers shared libraries consumed by the linker.

real-or-random:
Concept ACK

hebasto:
> Looking for Concept (N)ACKs. Once ACKed, I'll add the same functionality to the Autotools-based build system.

Added a commit for the Autotools-based build system.

CMakeLists.txt
```diff
@@ -269,6 +269,11 @@ if(SECP256K1_BUILD_CTIME_TESTS)
   unset(msan_enabled)
 endif()
 
+if(NOT SECP256K1_USE_EXTERNAL_DEFAULT_CALLBACKS)
+  include(TryAppendLnkerFlags)
```
real-or-random, line 273:
```suggestion
  include(TryAppendLinkerFlags)
```
and rename the file?

hebasto: Sure! Sorry for a typo :)

hebasto: Thanks! Done.

real-or-random (commented):
I think we need to make add an exception to this if external default callbacks are enabled because the entire point of this option is that the user provides the callbacks at link time.

hebasto:
> I think we need to make add an exception to this if external default callbacks are enabled because the entire point of this option is that the user provides the callbacks at link time.

Right. I did in CMake, but forgot about it in Autotools. Fixed.

cmake/TryAppendLinkerFlags.cmake
```diff
@@ -0,0 +1,32 @@
...
+  string(MAKE_C_IDENTIFIER "${flags}" result)
+  string(TOUPPER "${result}" result)
+  set(result "LINKER_SUPPORTS_${result}")
+  set(CMAKE_REQUIRED_FLAGS "${flags}")
+  if(MSVC)
+    string(APPEND " /WX")
+  elseif(CMAKE_SYSTEM_NAME STREQUAL "Darwin")
+    string(APPEND " -Wl,-fatal_warnings")
+  else()
+    string(APPEND " -Wl,--fatal-warnings")
```
real-or-random, line 15: Shouldn't this depend on the linker and not on the OS?

hebasto: I don't think it can be implemented in a simple way; see https://gitlab.kitware.com/cmake/cmake/-/issues/18209.

However, we can just check each variant.

real-or-random: Okay, I think checking `CMAKE_SYSTEM_NAME` is good enough and pragmatic. GNU ld doesn't even support the creation of macOS binaries, and even clang's macOS linker `ld64.lld` uses `-fatal_warnings`...

But shouldn't we append `-Wl,-fatal_warnings` or `-Wl,--fatal-warnings` also for the autotools build?

hebasto:
> This PR adds the `--no-undefined` option to a linker, which supports it.

Using sanitizers with Clang involves a static sanitizer runtime by default, which eventually fails when using the `--no-undefined` option. As a workaround, a dynamically linked sanitizer runtime can be forced with `-shared-libsan`, which in turn requires setting `RPATH` or `LD_PRELOAD`.

Perhaps we could set `--no-undefined` for non-sanitizer jobs in the CI?

In any case, closing for now.

PR #1563, opened 2024-07-01
doc: Add convention for defaults

### LINKED ISSUE #1549: discussion: Config used by Core vs defaults and config in CI/testing
Quoting myself from https://github.com/bitcoin-core/secp256k1/issues/1543#issuecomment-2176914205:

> I believe we want to build libsecp256k1 with the default `RelWithDebInfo` (`-O2 -g` in our case) even when we build it as part of Core. Because (almost) all our testing uses this config.

But now that this has led @hebasto to make or suggest a few changes (https://github.com/hebasto/bitcoin/pull/192#issuecomment-2184085876, https://github.com/bitcoin-core/secp256k1/pull/1547), I'm not confident enough, and I think we should discuss what our desired policy is. 

To put that in perspective, Core now builds with `--with-ecmult-gen-kb=86` like @sipa suggested, but almost all of our testing/CI the default value of 22.

*If* we come to the conclusion that most of our testing should align with what Core uses, then the difference of 86 vs 22 (as some kind of algorithmic change) is probably much more likely to cause trouble than some `-g` compiler flag. (On the other hand, stuff like `-O3` vs `-O2` could make a real difference again.)

cc @fanquake

---

Independently of that discussion, it would be awesome to have the ability to set something like "--with-ecmult-gen-kb=random" for CI. But this requires someone to implement it. :)

real-or-random:
Resolves #1549.

This does not say anything about compiler options and the like, but I think that's okay for now.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -49,6 +49,7 @@ In addition, libsecp256k1 tries to maintain the following coding conventions:
 * Operations involving secret data should be tested for being constant time with respect to the secrets (see [src/ctime_tests.c](src/ctime_tests.c)).
 * Local variables containing secret data should be cleared explicitly to try to delete secrets from memory.
 * Use `secp256k1_memcmp_var` instead of `memcmp` (see [#823](https://github.com/bitcoin-core/secp256k1/issues/823)).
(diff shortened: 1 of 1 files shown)


hebasto (approved):
ACK 16685649d2a6e52f81659ec629bc4e603128c54c

fanquake:
Concept ACK

jonasnick (approved):
ACK 16685649d2a6e52f81659ec629bc4e603128c54c

PR #1564, opened 2024-07-01
build, ci: Adjust the default size of the precomputed table for signing 

hebasto:
This PR implements the [outcomes](https://github.com/bitcoin-core/secp256k1/issues/1549#issuecomment-2200559257) from today's IRC meeting:

1. The default size of the precomputed table for signing is now aligned with Bitcoin Core's [default](https://github.com/bitcoin/bitcoin/commit/a057869aa3c42457570765966cb66accb2375b13).

2. The default value in CI has been updated to reflect the new default.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ env:
   BUILD: 'check'
   ### secp256k1 config
   ECMULTWINDOW: 15
-  ECMULTGENKB: 22
+  ECMULTGENKB: 86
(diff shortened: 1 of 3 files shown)


hebasto:
> ... I don't think it deserves a 0.6.0 instead of 0.5.1 (assuming it makes it to the next release).

I agree because there are no API changes.

real-or-random (approved):
utACK e2af491263e2f7e086dc7f1925d86660ff265dc6

PR #1565, opened 2024-07-01
cmake: Bump CMake minimum required version up to 3.16

hebasto:
Debian 10 [reached](https://wiki.debian.org/DebianReleases) EOL LTS yesterday, on 2024-06-30.

There no longer seem to be compelling reasons to maintain support for CMake 3.13.

The suggested minimum required version, CMake 3.16, is shipped with Ubuntu 20.04 LTS, which is [supported](https://wiki.ubuntu.com/Releases) until April 2025.

Debian 11 ships with CMake 3.18 (3.25 in backports). In [other](https://repology.org/project/cmake/versions) major distros and package managers, CMake versions are not older.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -864,5 +864,5 @@ jobs:
           CI_BUILD: ${{ runner.temp }}/${{ github.run_id }}${{ github.action }}/build
           CI_INSTALL: ${{ runner.temp }}/${{ github.run_id }}${{ github.action }}/install
         run: |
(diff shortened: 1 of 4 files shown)


real-or-random (approved):
utACK 730f9e1f43947a3e89dc1bbab7089e0f0b09810a seems reasonable to get rid of compatibility code

hebasto:
Fixed the "release" CI job.

fanquake (approved):
ACK 9420eece24117fc55540ab6d8bbfaca6b834b761 - If the only significant change is for some MSVC compat code, this doesn't seem super compelling, but seems ok either way.

real-or-random (approved):
utACK 9420eece24117fc55540ab6d8bbfaca6b834b761

PR #1566, opened 2024-07-02
BIP458: Schnorr (Incremental) Half Aggregation

fjahr:
This implements BIP458: https://github.com/bitcoin/bips/pull/2205

The original version of this code ported over the work from b-wagn in https://github.com/BlockstreamResearch/secp256k1-zkp/pull/261


.cirrus.yml
```diff
@@ -21,6 +21,8 @@ env:
   ECDH: no
   RECOVERY: no
```
jjones3695, line 22: yes

jjones3695, line 22: yes

.cirrus.yml
```diff
@@ -21,6 +21,8 @@ env:
   ECDH: no
```
jjones3695, line 21: yes

jjones3695 (approved):

fjahr:
Rebased on the musig changes and added the compile options for cmake

fjahr:
Rebased to take into account the latest changes to cmake, CI and test framework. Also squashed the commit since maintaining them was annoying and there wasn't really a good reason for them. It was more of a crutch for me while I was taking my first steps here.

include/secp256k1_schnorrsig_halfagg.h
```diff
@@ -0,0 +1,113 @@
...
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
+
+/** Verify a (Half-)aggregate Schnorr signature.
+ *
+ *  Returns:          1: correct signature.
+ *                    0: incorrect signature.
+ *  Args:           ctx: a secp256k1 context object.
+ *  In:         pubkeys: Array of n many x-only public keys. Can only be NULL if n is 0.
+ *               msgs32: Array of n many 32-byte messages. Can only be NULL if n is 0.
+ *                    n: number of signatures to that have been aggregated.
```
Jonatanc05:
```suggestion
 *                    n: number of signatures that have been aggregated.
```

Jonatanc05 (commented):
I think there's a typo

include/secp256k1_schnorrsig_halfagg.h
fjahr: Thanks, I will address it the next time I will push an update!

fjahr:
Rebased and added the test vectors from the BIP draft: https://github.com/bitcoin/bips/pull/2205

include/secp256k1_schnorrsig_halfagg.h
fjahr: Fixed

src/modules/schnorrsig_halfagg/main_impl.h
```diff
@@ -0,0 +1,219 @@
...
+    secp256k1_scalar_set_b32(&s, &aggsig[n*32], &overflow);
+    if (overflow) {
+        return 0;
+    }
+    secp256k1_ecmult(&lhs, &rhs, &secp256k1_scalar_zero, &s);
+
+    /* Check that lhs == rhs */
+    secp256k1_gej_neg(&lhs, &lhs);
+    secp256k1_gej_add_var(&lhs, &lhs, &rhs, NULL);
+    return secp256k1_gej_is_infinity(&lhs);
```
theStack, line 216: could use the `secp256k1_gej_eq_var` helper here which does the some steps internally:
```suggestion
    return secp256k1_gej_eq_var(&lhs, &rhs);
```

src/modules/schnorrsig_halfagg/main_impl.h
```diff
@@ -0,0 +1,219 @@
...
+    sha->s[0] = 0xd11f5532ul;
+    sha->s[1] = 0xfa57f70ful;
+    sha->s[2] = 0x5db0d728ul;
+    sha->s[3] = 0xf806ffe1ul;
+    sha->s[4] = 0x1d4db069ul;
+    sha->s[5] = 0xb4d587e1ul;
+    sha->s[6] = 0x50451c2aul;
+    sha->s[7] = 0x10fb63e9ul;
+
+    sha->bytes = 64;
```
theStack, line 22: could use `secp256k1_sha256_initialize_midstate` here (introduced in PR https://github.com/bitcoin-core/secp256k1/pull/1825, commit https://github.com/bitcoin-core/secp256k1/commit/f48b1bfa5d40a4d7303b196017d2e298520d1066), see other modules that use BIP340 tagged hashes

src/modules/schnorrsig_halfagg/main_impl.h
```diff
@@ -0,0 +1,219 @@
...
+int secp256k1_schnorrsig_inc_aggregate(const secp256k1_context *ctx, unsigned char *aggsig, size_t *aggsig_len, const secp256k1_xonly_pubkey *all_pubkeys, const unsigned char *all_msgs32, const unsigned char *new_sigs64, size_t n_before, size_t n_new) {
+    size_t i;
+    size_t n;
+    secp256k1_sha256 hash;
+    secp256k1_scalar s;
+    int overflow;
+    const secp256k1_hash_ctx *hash_ctx;
+
+    VERIFY_CHECK(ctx != NULL);
+    hash_ctx = secp256k1_get_hash_context(ctx);
```
theStack, line 34: this function was removed in #1908, so there is a (silent) merge conflict

src/CMakeLists.txt
```diff
@@ -12,6 +12,18 @@ if(SECP256K1_ENABLE_MODULE_ELLSWIFT)
   set_property(TARGET secp256k1 APPEND PROPERTY PUBLIC_HEADER ${PROJECT_SOURCE_DIR}/include/secp256k1_ellswift.h)
 endif()
 
+if(SECP256K1_ENABLE_MODULE_SCHNORRSIG_HALFAGG)
+  if(NOT SECP256K1_EXPERIMENTAL)
+    message(FATAL_ERROR "Schnorrsig half-aggregation is experimental. Use -DSECP256K1_EXPERIMENTAL=ON to allow.")
+  endif()
```
theStack, line 18: is the "experimental" category still used for new modules? (cc @real-or-random @sipa) I haven't seen it for the most recent ones (ellswift, musig, silentpayments), but IIUC the schnorrsig had it back then.

fjahr: Yeah, I myself argued that it might be possible to remove the category at some IRC meeting or the in-person meeting probably about a year ago. This was about the batch module specifically that still had it and it made sense for me to remove it there because it was already in BIP340 and the code would only be merged if it was in a usable state anyway, the bar is just that high already. I think @sipa said the category should be kept around if we want to introduce a module where the API may still change, at least that's the only argument I remember from the conversation. I already considered removing the category here and in fullagg for the same reasons as batch but then reconsidered: Maybe we would want to set it for some code that is specifically added for a soft fork before the soft fork is actually activated? (There are other use cases outside of consensus for halfagg and fullagg but realistically those will not get these PRs merged here.) I also saw that the Schnorr module was also merged as experimental but I did not go the extra mile to ask if this was actually because of the consensus angle, but I did the lazy thing and made no change here :)

However, I just now did some more digging and found this comment from @sipa that clarifies the intention for BIP340 at the time: https://github.com/bitcoin-core/secp256k1/issues/817#issuecomment-693033273 The whole issue was actually just closed after we discussed it as a group, btw.

I am currently basically 50/50 because I think the arguments for keeping Schnorr experimental from the comment very much apply to this pull here as is. But on the other hand I think the pull has a very low chance of getting merged while these arguments still apply because of the high standard here and if we agree on that we can remove the status already. Well, there is a chance that an alternative, newly discovered use case becomes strong enough for an earlier merge so maybe 51/49 for keeping it experimental for now and potentially removing it before the merge if doesn't seem needed anymore.

That's a long way of saying, I am happy to remove or keep it around.

src/modules/schnorrsig_halfagg/main_impl.h
fjahr: Ah, good catch, fixed and rebased.

src/modules/schnorrsig_halfagg/main_impl.h
fjahr: Done, thanks

src/modules/schnorrsig_halfagg/main_impl.h
fjahr: Done, thanks

fjahr:
Addressed the feedback from @theStack and rebased, happy to remove the experimental status if reviewers think it doesn't apply here.

src/modules/schnorrsig_halfagg/main_impl.h
```diff
@@ -0,0 +1,211 @@
...
+
+        /* Step 0: Serialize pk_i into pk_ser   */
+        /* We need that in Step 1 and in Step 2 */
+        if (!secp256k1_xonly_pubkey_load(ctx, &pp, &pubkeys[i])) {
+            return 0;
+        }
+        secp256k1_fe_get_b32(pk_ser, &pp.x);
+
+        /* Step 1: z_i = TaggedHash(...) */
+        /* 1.a) Write into hash r_i, pk_i, m_i, r_i */
```
theStack:
```suggestion
        /* 1.a) Write into hash r_i, pk_i, m_i */
```

src/modules/schnorrsig_halfagg/main_impl.h
```diff
@@ -0,0 +1,211 @@
...
+
+    /* Compute lhs = s*G. s is public (part of the aggregate signature), so we
+     * use the variable-time secp256k1_ecmult with the precomputed G table
+     * instead of constant-time secp256k1_ecmult_gen. This is faster and needs
+     * no ecmult_gen context, so aggverify works with secp256k1_context_static. */
+    secp256k1_scalar_set_b32(&s, &aggsig[n*32], &overflow);
+    if (overflow) {
+        return 0;
+    }
+    secp256k1_ecmult(&lhs, &rhs, &secp256k1_scalar_zero, &s);
```
theStack, line 205: not applicable yet, but just leaving a note already that in the future this could be replaced with `ecmult_gen_var_gej`, if #1883 gets in. for the `agg_verify` benchmark (with N=3), I'm seeing a ~5% speedup on my machine (95.5us vs. 90.5us): https://github.com/theStack/secp256k1/commit/cbe15b8f2ceaec0c0262d86c427617e0e65eb01d

fjahr: Nice, I will keep an eye on that!

src/modules/schnorrsig_halfagg/main_impl.h
fjahr: Good catch, fixed here and found the same error in `inc_aggregate`, fixed there as well.

fjahr:
Changed the tag to include the BIP number

PR #1574, opened 2024-07-22
Fix compilation when extrakeys module isn't enabled

### LINKED ISSUE #1573: Tests fail to compile when extrakeys module isn't enabled.
(tests use secp256k1_keypair without checking if the module is enabled)

jonasnick:
Fixes #1573. This wasn't caught by CI because the extrakeys was always enabled. I added a commit that changes that.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -20,6 +20,7 @@ env:
   EXPERIMENTAL: no
   ECDH: no
   RECOVERY: no
+  EXTRAKEYS: no
   SCHNORRSIG: no
   ELLSWIFT: no
   ### test options
@@ -66,6 +67,7 @@ task:
   env:
     ECDH: yes
     RECOVERY: yes
+    EXTRAKEYS: yes
     SCHNORRSIG: yes
     ELLSWIFT: yes
   matrix:
@@ -82,6 +84,7 @@ task:
   env:
     ECDH: yes
     RECOVERY: yes
+    EXTRAKEYS: yes
     SCHNORRSIG: yes
     ELLSWIFT: yes
     WRAPPER_CMD: 'valgrind --error-exitcode=42'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,6 +31,7 @@ env:
   EXPERIMENTAL: 'no'
   ECDH: 'no'
   RECOVERY: 'no'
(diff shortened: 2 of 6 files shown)


jonasnick:
Hm, I have no clue why the `macOS Sonoma (int128_struct, 2, 4)` CI job fails while the same settings work on `macOS Monterey` under valgrind in CI.

Sentini14:
What do you have and how can I withdraw

jonasnick:
The CI job failure doesn't seem to go away by rerunning the job.

jonasnick:
To investigate the CI failure, I played around on my fork but I couldn't reproduce the error anymore. The reported clang versions differed:
- Apple clang version 15.0.0 (clang-1500.0.40.1) in the failed CI job
- Apple clang version 15.0.0 (clang-1500.3.9.4) in my fork's successful CI job

I then restarted the job here, which resulted in the newer clang version being reported and the failure is gone.

hebasto (approved):
ACK 763d938cf0e68ef6dc52fda4f45cc03c5d2e31f0.

In the af551ab9db09a3e4f2d06f68cf3e140fb1acfc4b commit message:
> This fixes a bug introduced in 7d2591ce12d8a9b85f210cf9d678e91cee125ee9 that
prevented compiling the library without enabling the extrakeys module.

it seems that "compiling the tests" suits better than "compiling the library", doesn't it?

Is a combination of the enabled extrakeys module with disabled schnorrsig module not worth testing on the CI at all?

jonasnick:
Thanks for the review.

@hebasto:

>it seems that "compiling the tests" suits better than "compiling the library", doesn't it?

Slightly, yeah. But I don't want to invalidate the ACKs.

> Is a combination of the enabled extrakeys module with disabled schnorrsig module not worth testing on the CI at all?

Could be worth doing. We cannot really exhaustively check the entire configuration space, but for every module we could enable just that single module and all of its dependencies.

real-or-random:
> We cannot really exhaustively check the entire configuration space

We could generate a handful of random configs in every CI run. I think this will be very nice, but it's not entirely trivial. We currently set the config statically in CI, so this will need some thoughts. Perhaps a wrapper script that inspects the config env variables. If a variable is set to `MAGICRANDOM`, it will replace it by a random value.

PR #1575, opened 2024-08-01
release: prepare for 0.5.1

jonasnick:
I ran the sanity checks according to the release process and they passed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,20 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.5.1] - 2024-08-01
 
 #### Added
  - Added usage example for an ElligatorSwift key exchange.
 
+#### Changed
+ - The default size of the precomputed table for signing was changed from 22 KiB to 86 KiB. The size can be changed with the configure option `--ecmult-gen-kb` (`SECP256K1_ECMULT_GEN_KB` for CMake).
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -5,13 +5,20 @@ All notable changes to this project will be documented in this file.
...
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [Unreleased]
+## [0.5.1] - 2024-08-01
 
 #### Added
  - Added usage example for an ElligatorSwift key exchange.
 
+#### Changed
+ - The default size of the precomputed table for signing was changed from 22 KiB to 86 KiB. The size can be changed with the configure option `--ecmult-gen-kb` (`ECMULT_GEN_KB` for CMake).
```
hebasto, line 14:
```suggestion
 - The default size of the precomputed table for signing was changed from 22 KiB to 86 KiB. The size can be changed with the configure option `--ecmult-gen-kb` (`SECP256K1_ECMULT_GEN_KB` for CMake).
```

CHANGELOG.md
```diff
@@ -5,13 +5,20 @@ All notable changes to this project will be documented in this file.
...
+ - "auto" is no longer an accepted value for the `--with-ecmult-window` and `--with-ecmult-gen-kb` configure options (this also applies to  `ECMULT_WINDOW_SIZE` and `ECMULT_GEN_KB` in CMake). To achieve the same configuration as previously provided by the "auto" value, omit setting the configure option explicitly.
```
hebasto, line 15:
```suggestion
 - "auto" is no longer an accepted value for the `--with-ecmult-window` and `--with-ecmult-gen-kb` configure options (this also applies to `SECP256K1_ECMULT_WINDOW_SIZE` and `SECP256K1_ECMULT_GEN_KB` in CMake). To achieve the same configuration as previously provided by the "auto" value, omit setting the configure option explicitly.
```

CHANGELOG.md
jonasnick: Does  that mean the entry "(`ECMULT_GEN_KB` for CMake)" for 0.5.0 is wrong (or at least misleading)?

hebasto: I suppose it is misleading.

jonasnick: Thanks, fixed. Also added a commit to clarify the 0.5.0 entry.

PR #1579, opened 2024-08-06
Clear sensitive memory without getting optimized out (revival of #636)

### LINKED ISSUE #185: Memory zeroization improvements

theStack:
This PR picks up #636 (which in turn picked up #448, so this is take number three) and is essentially a rebase on master.

Some changes to the original PR:
* the clearing function now has the `secp256k1_` prefix again, since the related helper `_memczero` got it as well (see PR #835 / commit e89278f211a526062745c391d48a7baf782b4b2b)
* the original commit b17a7df8145a6a86d49c354c7e7b59a432ea5346 ("Make _set_fe_int( . , 0 ) set magnitude to 0") is not needed anymore, since it was already applied in PR #943 (commit d49011f54c2b31807158bdf06364f331558cccc7)
* clearing of stack memory with `secp256k1_memclear` is now also done on modules that have been newly introduced since then, i.e. schnorr and ellswift (of course, there is still no guarantee that all places where clearing is necessary are covered)

So far I haven't looked at any disassembly and possible performance implications yet (there were some concerns expressed in https://github.com/bitcoin-core/secp256k1/pull/636#issuecomment-620118629), happy to go deeper there if this gets Concept ACKed.

The proposed method of using a memory barrier to prevent optimizating away the memset is still used in BoringSSL (where it was originally picked up from) and in the Linux Kernel, see e.g. https://github.com/google/boringssl/blob/5af122c3dfc163b5d1859f1f450756e8e320a142/crypto/mem.c#L335 and https://github.com/torvalds/linux/blob/d4560686726f7a357922f300fc81f5964be8df04/include/linux/string.h#L348 / https://github.com/torvalds/linux/blob/d4560686726f7a357922f300fc81f5964be8df04/include/linux/compiler.h#L102 

Fixes #185.

### DIFF
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -597,6 +597,7 @@ static int secp256k1_ec_pubkey_create_helper(const secp256k1_ecmult_gen_context 
 
     secp256k1_ecmult_gen(ecmult_gen_ctx, &pj, seckey_scalar);
     secp256k1_ge_set_gej(p, &pj);
+    secp256k1_gej_clear(&pj);
     return ret;
 }
 
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -494,7 +494,7 @@ static int nonce_function_rfc6979(unsigned char *nonce32, const unsigned char *m
        buffer_append(keydata, &offset, algo16, 16);
    }
    secp256k1_rfc6979_hmac_sha256_initialize(&rng, keydata, offset);
-   memset(keydata, 0, sizeof(keydata));
+   secp256k1_memclear(keydata, sizeof(keydata));
    for (i = 0; i <= counter; i++) {
        secp256k1_rfc6979_hmac_sha256_generate(&rng, nonce32, 32);
    }
@@ -548,7 +548,7 @@ static int secp256k1_ecdsa_sign_inner(const secp256k1_context* ctx, secp256k1_sc
      * seckey. As a result is_sec_valid is included in ret only after ret was
      * used as a branching variable. */
     ret &= is_sec_valid;
-    memset(nonce32, 0, 32);
+    secp256k1_memclear(nonce32, sizeof(nonce32));
     secp256k1_scalar_clear(&msg);
     secp256k1_scalar_clear(&non);
     secp256k1_scalar_clear(&sec);
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -494,11 +494,13 @@ static int nonce_function_rfc6979(unsigned char *nonce32, const unsigned char *m
        buffer_append(keydata, &offset, algo16, 16);
    }
    secp256k1_rfc6979_hmac_sha256_initialize(&rng, keydata, offset);
-   secp256k1_memclear(keydata, sizeof(keydata));
    for (i = 0; i <= counter; i++) {
        secp256k1_rfc6979_hmac_sha256_generate(&rng, nonce32, 32);
    }
    secp256k1_rfc6979_hmac_sha256_finalize(&rng);
+
+   secp256k1_memclear(keydata, sizeof(keydata));
+   secp256k1_rfc6979_hmac_sha256_clear(&rng);
    return 1;
 }
 
@@ -799,6 +801,7 @@ int secp256k1_tagged_sha256(const secp256k1_context* ctx, unsigned char *hash32,
     secp256k1_sha256_initialize_tagged(&sha, tag, taglen);
     secp256k1_sha256_write(&sha, msg, msglen);
     secp256k1_sha256_finalize(&sha, hash32);
+    secp256k1_sha256_clear(&sha);
     return 1;
 }
 
--- a/src/util.h
+++ b/src/util.h
@@ -285,7 +285,7 @@ static SECP256K1_INLINE int secp256k1_is_zero_array(const unsigned char *s, size
     }
     ret = (acc == 0);
     /* acc may contain secret values. Try to explicitly clear it. */
-    acc = 0;
+    secp256k1_memclear(&acc, sizeof(acc));
     return ret;
 }
 
--- a/src/util.h
+++ b/src/util.h
@@ -8,11 +8,17 @@
 #define SECP256K1_UTIL_H
 
 #include "../include/secp256k1.h"
+#include "checkmem.h"
 
+#include <string.h>
 #include <stdlib.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <limits.h>
+#if defined(_MSC_VER)
+/* For SecureZeroMemory */
+#include <Windows.h>
+#endif
 
 #define STR_(x) #x
 #define STR(x) STR_(x)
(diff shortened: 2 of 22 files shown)


real-or-random (commented):
Concept ACK (obviously)

Thanks for reviving this, I never had the time/motivation to come back to this PR, but it's important.

We should call `SECP256K1_CHECKMEM_UNDEFINE` (https://github.com/bitcoin-core/secp256k1/blob/b307614401790850b48fb3ba878247290857a975/src/checkmem.h#L27) in `secp256k1_memclear` after clearing the memory. Reading cleared memory would clearly be a bug.

real-or-random:
This one comes to my mind, too: https://github.com/bitcoin-core/secp256k1/blob/a88aa9350633c2d2472bace5c290aa291c7f12c9/src/util.h#L253-L254

src/util.h
```diff
@@ -8,11 +8,17 @@
...
 #include "../include/secp256k1.h"
+#include "checkmem.h"
 
+#include <string.h>
 #include <stdlib.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <limits.h>
+#if defined(_MSC_VER)
+// For SecureZeroMemory
```
real-or-random, line 19: nit: `//`-style comments not allowed in C89. Just for consistency, it appears that MSVC doesn't care...

src/secp256k1.c
```diff
@@ -494,11 +494,13 @@ static int nonce_function_rfc6979(unsigned char *nonce32, const unsigned char *m
...
    }
    secp256k1_rfc6979_hmac_sha256_initialize(&rng, keydata, offset);
-   secp256k1_memclear(keydata, sizeof(keydata));
    for (i = 0; i <= counter; i++) {
        secp256k1_rfc6979_hmac_sha256_generate(&rng, nonce32, 32);
    }
    secp256k1_rfc6979_hmac_sha256_finalize(&rng);
+
+   secp256k1_memclear(keydata, sizeof(keydata));
+   secp256k1_rfc6979_hmac_sha256_clear(&rng);
```
real-or-random, line 503: We should add similar calls in other code locations where we derive secrets via hash functions, e.g., in the schnorrsig and musig modules (and ellswift? now sure). The previous PR predates even the schnorrsig module, see the commit message.

real-or-random (commented):
If you want motivated, you could look at `git grep secp256k1_ge_set_gej.*(`. @sipa's recent comment in the MuSig PR really caught my attention. When I worked on the previous PR, I really didn't consider the possibility that a `gej` could leak secret data.

----

I wonder if it makes sense to have the `_finalize` functions in the hash module clear the state automatically. And then have "unsafe" funtions like `_finalize_keep` if the callers knows that data is public or if the caller wants to reuse the midstate. It sounds like a neat idea, but I'm not really convinced because it special-cases the hash module somewhat: we'll need to manually clear everything else including scalars, etc... So we have to be careful with this anyway when writing and reviewing code, and perhaps having yet another safety mechanism that works only for the hash module. (This could perhaps be a nice follow-up if we add the same mechanism to more modules, e.g., `ge_set_gej` could clear the gejs, unless you use a `_keep` variant.)

real-or-random:
> Taking care of those could be probably go into an own PR, as its trivial to fix and review and hence has a significantly higher chance to land in the next release than this one? (and having a version where the compiler still might optimize it out seems still much better than not doing it at all)


My guess is that what we do currently is useless on any modern compiler. I admit that I haven't looked at the compiler output, but I'd rather spend the time on resolving the problem.

And to be honest, while this is a great for defense in depth, we haven't deeply cared about this so far. It's not the end of the world if we need to wait a few more months. So I think it's good to add these cases to this PR here.

src/util.h
```diff
@@ -221,6 +227,32 @@ static SECP256K1_INLINE void secp256k1_memczero(void *s, size_t len, int flag) {
...
+     * just not remove it entirely. See "Dead Store Elimination (Still) Considered Harmful" by
+     * Yang et al. (USENIX Security 2017) for more background.
+     */
+    memset(ptr, 0, len);
+    __asm__ __volatile__("" : : "r"(ptr) : "memory");
+#else
+    void *(*volatile const volatile_memset)(void *, int, size_t) = memset;
+    volatile_memset(ptr, 0, len);
+#endif
+    SECP256K1_CHECKMEM_UNDEFINE(ptr, len);
```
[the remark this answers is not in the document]
real-or-random: Hm, great point! The reason why `secp256k1_declassify` uses a run-time flag (instead of compile-time flag) is that we want to run the constant-time tests on the real binary.

I don't know what the performance impact of these additional instructions is, but I doubt that having a compile-time flag is a concern in this case. That means that we could just wrap the `SECP256K1_CHECKMEM_UNDEFINE` in an `#ifdef VERIFY`. @sipa What do you think?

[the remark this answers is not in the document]
real-or-random: Oh, sorry, I think we're talking past each other. My suggestion, for which I'd like to have your opinion on, is to wrap `SECP256K1_CHECKMEM_UNDEFINE` in an `#ifdef VERIFY` block **here in this particular case of clearing memory**, i.e., inside `secp256k1_memclear` (and not in `secp256k1_declassify`.  Do you think that's a reasonable thing to do?

---
But regarding what you said:

> I think the reasoning is that we want the release binaries to be as close as possible to what we actually test in the ctime test, but disable the effect at runtime using `SECP256K1_CONTEXT_DECLASSIFY` to avoid a performance impact.

Yes, I agree. Perhaps an additional reason is to avoid any shenanigans that the valgrind syscalls may have. They should be noops, but it's certainly a bit safer not to run them at all. 

> It wouldn't surprise me that that is overkill; the conditional to determine whether or not to declassify operations at runtime may have a higher cost than actually executing the nop instructions that declassification actually correspond to when not running instrumented by valgrind.

I can imagine that the overhead of the noops is negligible, but at least checking the conditional should be negligible as well because we use `EXCEPT` to help the compiler predict the branch:
https://github.com/bitcoin-core/secp256k1/blob/68b55209f1ba3e6c0417789598f5f75649e9c14c/src/secp256k1.c#L236-L238

src/util.h
```diff
@@ -221,6 +227,32 @@ static SECP256K1_INLINE void secp256k1_memczero(void *s, size_t len, int flag) {
     }
 }
 
+/* Cleanses memory to prevent leaking sensitive info. Won't be optimized out. */
+static SECP256K1_INLINE void secp256k1_memclear(void *ptr, size_t len) {
+#if defined(_MSC_VER)
+    /* SecureZeroMemory is guaranteed not to be optimized out by MSVC. */
+    SecureZeroMemory(ptr, len);
```
real-or-random, line 234: I looked at the same code in Core again, and found this PR, and commented there: https://github.com/bitcoin/bitcoin/pull/26950#issuecomment-2429613882

src/util.h
```diff
@@ -221,6 +227,32 @@ static SECP256K1_INLINE void secp256k1_memczero(void *s, size_t len, int flag) {
...
+     * This method used in memzero_explicit() the Linux kernel, too. Its advantage is that it is
+     * pretty efficient, because the compiler can still implement the memset() efficently,
+     * just not remove it entirely. See "Dead Store Elimination (Still) Considered Harmful" by
+     * Yang et al. (USENIX Security 2017) for more background.
+     */
+    memset(ptr, 0, len);
+    __asm__ __volatile__("" : : "r"(ptr) : "memory");
+#else
+    void *(*volatile const volatile_memset)(void *, int, size_t) = memset;
+    volatile_memset(ptr, 0, len);
```
real-or-random, line 251: I've tested this locally, but it would be nice to see a full CI run where this branch is forced (e.g., via `#if 0` / `#elif 0`)

real-or-random (commented):
ACK mod my review comments

src/util.h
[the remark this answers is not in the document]
real-or-random: I see. That's a valid point, which has been brought up before, see https://github.com/bitcoin-core/secp256k1/pull/813#issuecomment-691696004. (This thread is also very helpful to understand why we have run-time flag etc.) The conclusion was that the benefits outweigh the drawback that the build outputs depends on the availability of valgrind, but yeah, there's no perfect solution here. And yes, it's also auto-detected in autotools.

src/util.h
real-or-random: Everything else being equal, keeping the code in sync with Core is nice, but as I said there, I have a slight preference to stick to the current `#if defined(_MSC_VER)` because we have no OS-specific macros anywhere else in the library (ignoring the examples). See also https://github.com/bitcoin-core/secp256k1/pull/1625.

real-or-random:
> OTOH, "ctime-tests should run with the same binary as in the release" then doesn't hold anymore.

I don't think so. If we wrap the new  _CHECKMEM_UNDEFINE **within _memclear** in #ifdef VERIFY, I don't see how this affects the ctimetests at all.  We'd still include the _CHECKMEM_UNDEFINE in declassify in the release builds. And the ctimetests executable will still link against the (release) binary of the library. (Note that the ctimetests test executable uses the public API, it's really just an application that consumes the library by linking against it, as opposed to the other test binaries, which essentially #include all the library code.)

src/util.h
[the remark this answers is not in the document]
real-or-random: Indeed. 

> That would mean `tests_noverify` wouldn't detect violations of use-after-clear

Fwiw, I had this downside in mind, and I don't think it's a big deal because we still have the normal tests.

> nor would applications linking against the production library when running in valgrind.

Okay, I had not considered this one... I still think that's acceptable, but yeah, it's a good point...

real-or-random (approved):
ACK 765ef53335a3e0fafdafe1e757f6fe0789f2797f

fanquake:
This got a changelog in the 0.6.0 release notes: https://github.com/bitcoin-core/secp256k1/blob/master/CHANGELOG.md#060---2024-11-04

PR #1583, opened 2024-08-12
ci: Bump GCC_SNAPSHOT_MAJOR to 15

maflcko:
Follow-up to https://github.com/bitcoin-core/secp256k1/pull/1313

Clang should silently follow the `main` devel branch, but GCC needs to be bumped manually.

### DIFF
--- a/ci/linux-debian.Dockerfile
+++ b/ci/linux-debian.Dockerfile
@@ -40,7 +40,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
         apt-get clean && rm -rf /var/lib/apt/lists/*
 
 # Build and install gcc snapshot
-ARG GCC_SNAPSHOT_MAJOR=14
+ARG GCC_SNAPSHOT_MAJOR=15
 RUN apt-get update && apt-get install --no-install-recommends -y wget libgmp-dev libmpfr-dev libmpc-dev flex && \
     mkdir gcc && cd gcc && \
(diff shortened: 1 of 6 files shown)


real-or-random:
GCC 15 has a new warning, see https://gcc.gnu.org/pipermail/gcc-patches/2024-June/656014.html .

real-or-random:
> ```c++
> static const unsigned char PREFIX[19] = {'s', 'e', 'c', 'p', '2', '5', '6', 'k', '1', ' ', 't', 'e', 's', 't', ' ', 'i', 'n', 'i', 't'};
> ```

This is ultimately a question of taste, but I suggest this because I think it's the most readable:

```
static const unsigned char PREFIX[] = {'s', 'e', 'c', 'p', '2', '5', '6', 'k', '1', ' ', 't', 'e', 's', 't', ' ', 'i', 'n', 'i', 't'};
```

This makes it unambiguously clear that we consider it an array and not a string (i.e., it's not null-terminated), the reader doesn't need to count the chars, and we can use `sizeof(PREFIX)`.


We define such constant arrays in multiple code locations, see 
 https://github.com/bitcoin-core/secp256k1/actions/runs/10356727938/job/28667982412?pr=1583
Whenever I review such code, I need to count the chars manually to check whether the terminating null byte is included or not.

hebasto (approved):
ACK fa67b6752d8ba3e4c41f6c36b1c6b94a21770419, I have reviewed the code and it looks OK.

real-or-random (approved):
utACK fa67b6752d8ba3e4c41f6c36b1c6b94a21770419

PR #1584, opened 2024-08-13
feat(build): add autoreconf dependency check

epiccurious:
Add check for autoreconf dependency to `autogen.sh` and refactor `-if` to long forms of optional arguments.

## Impact

- Add a functional change to improves error handling to provide a more helpful message if autoreconf is not found.
- Add a non-functional refactor to change the autoreconf optional arguments from -if to --install --force.
- These changes align with the format of [autogen.sh on the bitcoin repo](https://github.com/bitcoin/bitcoin/blob/master/autogen.sh#L14:L16)

## Background

**10 years ago:** Created the entire file with no other updates.

- [Link to commit](https://github.com/bitcoin-core/secp256k1/commit/db72c181bbace68072a2545ad8de2d92a3f06e63)

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,3 +1,5 @@
 #!/bin/sh
 set -e
+command -v autoreconf >/dev/null || \
+  (echo "configuration failed, please install autoconf first" >&2 && exit 1)
 autoreconf -if --warnings=all
--- a/autogen.sh
+++ b/autogen.sh
@@ -2,4 +2,4 @@
 set -e
 command -v autoreconf >/dev/null || \
   (echo "configuration failed, please install autoconf first" >&2 && exit 1)
-autoreconf -if --warnings=all
+autoreconf --install --force --warnings=all


hebasto:
>    * Add a functional change to improves error handling to provide a more helpful message if autoreconf is not found.

On my system, the current error message is already quite descriptive:
```
$ ./autogen.sh 
./autogen.sh: 3: autoreconf: not found
```

epiccurious:
> the current error message is already quite descriptive

I agree, but the error message could be made even more helpful for those less familiar with the build process.

Since `autoreconf` is a part of the `autoconf` package, directing users to install `autoconf` -- rather than just indicating that `autoreconf` is missing -- makes which package to install clearer.

`autoconf` is named consistently across popular Unix-based package managers, so this saves a step for users who might look to install `autoreconf`:
```
apt-get install autoconf
dnf install autoconf
pacman -S autoconf
install autoconf
zypper install autoconf
apk add autoconf
nix-env -iA nixpkgs.autoconf
emerge dev-util/autoconf
```

hebasto:
> > the current error message is already quite descriptive
> 
> I agree, but the error message could be made even more helpful for those less familiar with the build process.
> 
> Since `autoreconf` is a part of the `autoconf` package, directing users to install `autoconf` (rather than just indicating that `autoreconf` is missing) makes it clearer which package to install.
> 
> `autoconf` is named consistently across popular Unix-based package managers, so this saves a step for users who try to install `autoreconf`:
> 
> ```
> apt-get install autoconf
> dnf install autoconf
> pacman -S autoconf
> install autoconf
> zypper install autoconf
> apk add autoconf
> nix-env -iA nixpkgs.autoconf
> emerge dev-util/autoconf
> ```

The suggested approach is incomplete. Wouldn't it be better to document build prerequisites, which also include `automake`, `libtool`, etc?

real-or-random:
> The suggested approach is incomplete. Wouldn't it be better to document build prerequisites, which also include `automake`, `libtool`, etc?

Yes, I think so, and this should probably simply go to the README then.

PR #1585, opened 2024-08-13
fixes a few grammatical errors

justinvforvendetta:


src/modinv32_impl.h
```diff
@@ -80,7 +80,7 @@ static void secp256k1_modinv32_normalize_30(secp256k1_modinv32_signed30 *r, int3
     /* In a first step, add the modulus if the input is negative, and then negate if requested.
      * This brings r from range (-2*modulus,modulus) to range (-modulus,modulus). As all input
      * limbs are in range (-2^30,2^30), this cannot overflow an int32_t. Note that the right
-     * shifts below are signed sign-extending shifts (see assumptions.h for tests that that is
+     * shifts below are signed sign-extending shifts (see assumptions.h for tests that verify that is
      * indeed the behavior of the right shift operator). */
```
jonasnick, line 84: Not a native speaker, but the current version seems easier to parse. Maybe "... for tests that this is indeed ...".

real-or-random: NACK on this change. I think the current version on master is perfectly fine and readable, and so there's no reason to bother with it. Neither your first suggestion nor your updated suggestion seems natural to me. I feel that "that" should be omitted here. If your grammar checker complains about a repeated word, then that's a false positive.  (And I don't think Jonas' suggestion is an improvement either.)

PR #1587, opened 2024-08-14
use variable-time group addition in `_ec_pubkey_combine`

theStack:
Unless I'm missing something, it seems that there is no need to use constant-time group addition for `secp256k1_ec_pubkey_combine`, so the faster variable-time addition function `_gej_add_ge_var` can be used instead. Happy to add a benchmark if wanted.

### DIFF
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -791,7 +791,7 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
     for (i = 0; i < n; i++) {
         ARG_CHECK(pubnonces[i] != NULL);
         secp256k1_pubkey_load(ctx, &Q, pubnonces[i]);
-        secp256k1_gej_add_ge(&Qj, &Qj, &Q);
+        secp256k1_gej_add_ge_var(&Qj, &Qj, &Q, NULL);
     }
     if (secp256k1_gej_is_infinity(&Qj)) {
         return 0;


Bigheem (approved):

real-or-random:
> I don't know what the original intention of using constant time addition was. Maybe @sipa remembers. It was added here: [#212 (files)](https://github.com/bitcoin-core/secp256k1/pull/212/files#diff-6f71b0372be086d45b4f2740508c03a21835d87008840032fbb767f419fd988aR552)

The function was added for key aggregation in multi-signatures (predating MuSig(2)). Variable-time group addition was available back then.

Perhaps the idea was that the individual keys to be aggregated are secrets. One can make that point (also in MuSig), but I think our current implicit policy is that only secret keys are secret when it comes to timing (and public keys are, well, public).

> I'm aware that some implementations use `pubkey_combine` to add "secret" group elements in a "blind DH" (see [here](https://github.com/cashubtc/nutshell/blob/77ba35649993d229f5f01c2ab2cf66278884cd30/cashu/core/crypto/b_dhke.py), [here](https://github.com/cashubtc/nutshell/blob/77ba35649993d229f5f01c2ab2cf66278884cd30/cashu/core/crypto/secp.py#L7) and [here](https://github.com/cashubtc/cdk/blob/5f87df2cef71b97fac42419e97849335854743f4/crates/cdk/src/dhke.rs#L75)). I haven't checked in detail if using variable time addition leads to a sidechannel that isn't already there with constant time addition. In any case, constant time addition is not something one can expect from libsecp's API (it's called "**public** key combine").

Hm, yeah. I suggest not bothering with these functions for now. I think our long-term plan is to deprecate them and replace them with a properly designed hazmat API (see for example https://github.com/bitcoin-core/secp256k1/pull/1471#issuecomment-1967953283). No matter, if we'll do this in the future or not, it's not a big deal for performance if this function stays constant-time.

PR #1593, opened 2024-08-20
Remove deprecated `_ec_privkey_{negate,tweak_add,tweak_mul}` aliases from API

theStack:
These function aliases have been described as DEPRECATED in the public API docs already many years ago (see #701, commit 41fc7856), and in addition explicit deprecation warnings are shown by the compiler at least since the first official release 0.2.0 (see PR #1089, commit fc94a2da), so it should be fine to just remove them by now without causing bad surprises to any users.

Note that this PR intentionally doesn't include other deprecated flags/types/aliases by now, as they all have been introduced or marked as deprecated later (though the compiler deprecation warnings have been introduced at the same time). There are no deprecation rules yet and seemingly no removal of deprecated types/functions has ever happened so far, so it might make sense to discuss how to handle deprecation in general and introduce guidelines in the future.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+#### Removed
+- The previously deprecated function aliases `secp256k1_ec_privkey_negate`, `secp256k1_ec_privkey_tweak_add` and
+  `secp256k1_ec_privkey_tweak_mul` were removed. The corresponding replacements `secp256k1_ec_seckey_negate`,
+  `secp256k1_ec_seckey_tweak_add` and `secp256k1_ec_seckey_tweak_mul` have to be used instead.
+
 ## [0.5.1] - 2024-08-01
 
 #### Added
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+#### Removed
+- The previously deprecated function aliases `secp256k1_ec_privkey_negate`, `secp256k1_ec_privkey_tweak_add` and
+  `secp256k1_ec_privkey_tweak_mul` were removed. The corresponding replacements `secp256k1_ec_seckey_negate`,
+  `secp256k1_ec_seckey_tweak_add` and `secp256k1_ec_seckey_tweak_mul` have to be used instead.
```
real-or-random, line 13: nit, slightly shorter without loss of information:
```suggestion
- Removed previously deprecated function aliases `secp256k1_ec_privkey_negate`, `secp256k1_ec_privkey_tweak_add`, and
  `secp256k1_ec_privkey_tweak_mul`. Use `secp256k1_ec_seckey_negate`,
  `secp256k1_ec_seckey_tweak_add`, and `secp256k1_ec_seckey_tweak_mul` instead.
```

real-or-random (approved):
utACK

But needs rebase.

I think it's totally fine to remove these now. Would be nice to merge this for the same release as https://github.com/bitcoin-core/secp256k1/pull/1639 which is also breaking.

real-or-random (approved):
utACK 37d2c60beca394e8f4733c65f5d558eb302cac06

jonasnick (approved):
ACK 37d2c60beca394e8f4733c65f5d558eb302cac06

PR #1595, opened 2024-08-24
build: 45839th attempt to fix symbol visibility on Windows

### LINKED ISSUE #1421: mingw-w64: "visibility attribute not supported in this configuration; ignored" under `-flto`

real-or-random:
Fixes #1421. See code comments for rationale. 

Related meta-bug: #1181.  This reminds me that we should move forward with #1359.

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -147,6 +147,15 @@ typedef int (*secp256k1_nonce_function)(
      * 1. If using Libtool, it defines DLL_EXPORT automatically.
      * 2. In other cases, SECP256K1_DLL_EXPORT must be defined. */
 #   define SECP256K1_API extern __declspec (dllexport)
+#  else
+    /* Building libsecp256k1 as a static library on Windows.
+     * No declspec is needed, and so we would want the non-Windows-specific
+     * logic below take care of this case. However, this may result in setting
+     * __attribute__ ((visibility("default"))), which is supposed to be a noop
+     * on Windows but may trigger warnings when compiling with -flto due to a
+     * bug in GCC, see
+     * https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116478 . */
+#   define SECP256K1_API extern
 #  endif
   /* The user must define SECP256K1_STATIC when consuming libsecp256k1 as a static
    * library on Windows. */
@@ -156,11 +165,12 @@ typedef int (*secp256k1_nonce_function)(
 # endif
 #endif
 #ifndef SECP256K1_API
+/* All cases not captured by the Windows-specific logic. */
 # if defined(__GNUC__) && (__GNUC__ >= 4) && defined(SECP256K1_BUILD)
-   /* Building libsecp256k1 on non-Windows using GCC or compatible. */
+   /* Building libsecp256k1 using GCC or compatible. */
 #  define SECP256K1_API extern __attribute__ ((visibility ("default")))
 # else
-   /* All cases not captured above. */
+   /* Fall back to standard C's extern. */
 #  define SECP256K1_API extern
 # endif
 #endif


fanquake:
I can at least confirm that this "fixes" the output I was seeing in #1421. The most recent comment on the upstream GCC issue is not 100% clear to me.

real-or-random:
> The most recent comment on the upstream GCC issue is not 100% clear to me.

Can you elaborate on which part is unclear? 

In any case, I could rephrase this such that it says it's a GCC bug. When I wrote it, I wasn't sure yet, but the upstream response seems to confirm that this is a bug.

fanquake:
I've re-read, and yes, it seems like the conclusion is that it is a GCC issue. I guess what is actually unclear is if/when anyone might fix it upstream.

real-or-random:
Okay. I've nevertheless rephrased the comment to make it (hopefully) clearer.

fanquake:
ACK 447334cb06de229024161021cc79b3af32ce8b5c

real-or-random:
@hebasto Wanna review this? :)

fanquake:
also cc @theuni

include/secp256k1.h
```diff
@@ -147,6 +147,15 @@ typedef int (*secp256k1_nonce_function)(
      * 1. If using Libtool, it defines DLL_EXPORT automatically.
      * 2. In other cases, SECP256K1_DLL_EXPORT must be defined. */
 #   define SECP256K1_API extern __declspec (dllexport)
+#  else
+    /* Building libsecp256k1 as a static library on Windows.
```
hebasto, line 151: nit: Referencing the use of a static library does not apply to our cases with the `tests` and `noverify_tests` targets, where object files are linked directly.

hebasto (approved):
ACK 447334cb06de229024161021cc79b3af32ce8b5c, tested on Ubuntu 24.04 using the following commands:
```
$ env CFLAGS="-flto" cmake -B build --preset dev-mode --toolchain cmake/x86_64-w64-mingw32.toolchain.cmake
$ cmake --build build -j 16 -t tests noverify_tests
```

include/secp256k1.h
real-or-random: You mean https://github.com/bitcoin-core/secp256k1/blob/master/src/CMakeLists.txt#L89, right? True, and it's also true for `bench_internal` and `bench_ecmult`. 

Would you prefer "Building libsecp256k1 not as a DLL."? (I dropped the Windows part because it's anyway only Windows in the outer #ifdef.)

real-or-random: > Looks good. However, I'm not sure if it worth invalidating all ACKs.

Ok, let's just get it merged.

Issue #1570, opened 2024-07-12
Examples should not retry on bad randomness keys in a loop


real-or-random:
> Well, we should change the other examples and perhaps also the API docs. Creating keys in a while loop is not good practice. If you hit an invalid key, you're most probably not very lucky, but very unlucky because your randomness is broken. But hey, yeah, let's just yolo and try again. :)

_Originally posted by @real-or-random in https://github.com/bitcoin-core/secp256k1/pull/1519#discussion_r1675553262_

real-or-random:
Resolved by #1570 and #1616.

PR #1519, opened 2024-04-19
Add BIP352 `silentpayments` module

josibake:
This PR adds a new Silent Payments ([BIP352](https://github.com/bitcoin/bips/pull/1458)) module to secp256k1. It is a continuation of the work started in https://github.com/bitcoin-core/secp256k1/pull/1471.

The module implements the full protocol, except for transaction input filtering and silent payment address encoding / decoding as those will be the responsibility of the wallet software. It is organized with functions for sending (prefixed with `_sender`) and receiving (prefixed by `_recipient`).

#### For sending

1. Collect private keys into two lists: `taproot_seckeys` and `plain_seckeys`
    Two lists are used since the `taproot_seckeys` may need negation. `taproot_seckeys` are passed as keypairs to avoid the function needing to compute the public key to determine parity. `plain_seckeys` are passed as just secret keys
2. Create the `_silentpayment_recipient` objects
    These structs hold the scan and spend public key and an index for remembering the original ordering. It is expected that a caller will start with a list of silent payment addresses (with the desired amounts), convert these into an array of `recipients` and then match the generated outputs back to the original silent payment addresses. The index is used to return the generated outputs in the original order
3. Call `silentpayments_sender_create_outputs` to generate the xonly public keys for the recipients
    This function can be called with one or more recipients. The same recipient may be repeated to generate multiple outputs for the same recipient
    
#### For scanning

1. Collect the public keys into two lists `taproot_pubkeys` and `plain_pubeys`
    This avoids the caller needing to convert taproot public keys into compressed public keys (and vice versa)
2. Compute the input data needed, i.e. sum the public keys and compute the `input_hash`
    This is done as a separate step to allow the caller to reuse this output if scanning for multiple scan keys. It also allows a caller to use this function for aggregating the transaction inputs and storing them in an index to vend to light clients later (or for faster rescans when recovering a wallet)
4. Call `silentpayments_recipient_scan_outputs` to scan the transaction outputs and return the tweak data (and optionally label information) needed for spending later

In addition, a few utility functions for labels are provided for the recipient for creating a label tweak and tweaked spend public key for their address. Finally, two functions are exposed in the API for supporting light clients, `_recipient_created_shared_secret` and `_recipient_create_output_pubkey`. These functions enable incremental scanning for scenarios where the caller does not have access to the transaction outputs:

1. Calculating a shared secret
    This is done as a separate step to allow the caller to reuse the shared secret result when creating outputs and avoid needing to do a costly ECDH every time they need to check for an additional output
[...]

PR #1599, opened 2024-08-29
#1570 improve examples: remove key generation loop

Cheapshot003:
Hi! As discussed in #1570 I propose this as a change to the examples.c

If the key is either zero or out of range we just return 1 and end the example instead of looping until a good key is found. This is all very unlikely but could indicate a faulty/manipulated RNG.
Anybody knows where I can find the docs for this? First commit to this library, hope I don't break anything!

### DIFF
--- a/examples/ellswift.c
+++ b/examples/ellswift.c
@@ -48,17 +48,16 @@ int main(void) {
 
     /*** Generate secret keys ***/
 
-    /* If the secret key is zero or out of range (bigger than secp256k1's
-     * order), we try to sample a new key. Note that the probability of this
-     * happening is negligible. */
-    while (1) {
-        if (!fill_random(seckey1, sizeof(seckey1)) || !fill_random(seckey2, sizeof(seckey2))) {
-            printf("Failed to generate randomness\n");
-            return 1;
-        }
-        if (secp256k1_ec_seckey_verify(ctx, seckey1) && secp256k1_ec_seckey_verify(ctx, seckey2)) {
-            break;
-        }
+    /* If the secret key is zero or out of range (greater than secp256k1's
+     * order), we return 1. Note that the probability of this occurring
+     * is negligible with a properly functioning random number generator. */
+    if (!fill_random(seckey1, sizeof(seckey1)) || !fill_random(seckey2, sizeof(seckey2))) {
+        printf("Failed to generate randomness\n");
+        return 1;
+    }
+    if (!secp256k1_ec_seckey_verify(ctx, seckey1) || !secp256k1_ec_seckey_verify(ctx, seckey2)) {
+        printf("Generated secret key is invalid. This indicates an issue with the random number generator.\n");
+        return 1;
     }
 
     /* Generate ElligatorSwift public keys. This should never fail with valid context and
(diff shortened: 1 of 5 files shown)


examples/ellswift.c
```diff
@@ -49,16 +49,15 @@ int main(void) {
...
-    while (1) {
-        if (!fill_random(seckey1, sizeof(seckey1)) || !fill_random(seckey2, sizeof(seckey2))) {
-            printf("Failed to generate randomness\n");
-            return 1;
-        }
-        if (secp256k1_ec_seckey_verify(ctx, seckey1) && secp256k1_ec_seckey_verify(ctx, seckey2)) {
-            break;
-        }
+     * order), we return 1. Note that the probability of this
+     * happening is negligible, though it could indicate a faulty RNG */
```
jonasnick, line 53:
```
/* If the secret key is zero or out of range (greater than secp256k1's
 * order), we return 1. Note that the probability of this occurring
 * is negligible with a properly functioning random number generator. */
 ```

jonasnick (commented):
Hey @Cheapshot003, thanks for the contribution. What sort of docs are you looking for exactly?

Cheapshot003:
> Hey @Cheapshot003, thanks for the contribution. What sort of docs are you looking for exactly?

If you look [here](https://github.com/bitcoin-core/secp256k1/pull/1519#discussion_r1675553262) someone talks about API docs that should be changed. I don't find any docs related to the examples.

jonasnick:
Ah. I don't know which API docs @real-or-random meant.

Cheapshot003:
ok @jonasnick should be all fixed now :)

Cheapshot003:
Ah I pushed from another machine that wasn't properly configured. 
Squashed commits, should be alright now. Thx for the help!

real-or-random:
> Ah. I don't know which API docs @real-or-random meant.

My thinking was that the docs of `secp256k1_ec_seckey_verify` could not only say that the probability of getting an invalid key is negligible, but also spell out the truth that if this occurs on a randomly generated key, then the caller should assume that the randomness source is severely broken and not retry.   
https://github.com/bitcoin-core/secp256k1/blob/1988855079fa8161521b86515e77965120fdc734/include/secp256k1.h#L682-L697

Please don't hesitate to add a commit, but we can also do this in a different PR instead. 

(Now that I see this, we should also replace "ECDSA secret key" by "elliptic curve secret key" or just "secret key". This dates back to the early days of the library, when it supported only ECDSA.)

Cheapshot003:
Oh thx, I'll adjust these when I come home in the afternoon. I think it would make sense to do it within this pull request

real-or-random (commented):
Looks good from my side except minor corrections.

Can you squash your commits again after addressing the comments? I recommend either keeping it two logical commits (one for the examples and one for the docs), or just squashing everything in a single commit if that's easier. (It's a bit a matter of taste in the end. The example changes and the doc changes are not exactly of the same nature, but certainly closely related.)

real-or-random:
friendly ping :) @Cheapshot003

Cheapshot003:
Heyo! Back from vacation! The force push looked a bit scary at first but I think I didn't break anything. Should be alright now!
Greetings

real-or-random (approved):
utACK cd4f84f3ba8de37726274f61d5a34477c88433cf

jonasnick (approved):
ACK cd4f84f3ba8de37726274f61d5a34477c88433cf

PR #1600, opened 2024-09-02
cmake: Introduce `SECP256K1_APPEND_LDFLAGS` variable

hebasto:
Downstream projects may include libsecp256k1 test executables in their test suites. In such cases, if a `-fsanitize=...` flag is passed via the `SECP256K1_APPEND_CFLAGS` variable, the same flag must also be passed to the linker.

Required for https://github.com/bitcoin/bitcoin/pull/30791.


fanquake:
Concept ACK - It'd be great if a maintainer could take a look, as we have some changes downstream that are blocked on this.

jonasnick (approved):
@real-or-random sent me a message that he utACKs this and I tested that it works.

ACK 421ed1b46ff018f33a497df913a11d20f8dd1acc

PR #1603, opened 2024-09-06
f can never equal -m

roconnor-blockstream:
In fact, before reaching this particular VERIFY_CHECK, we had already successfully passed through

    VERIFY_CHECK(secp256k1_modinv64_mul_cmp_62(&f, len, &modinfo->modulus, -1) > 0); /* f > -modulus */

ensuring that f is not -m.

### DIFF
--- a/src/modinv32_impl.h
+++ b/src/modinv32_impl.h
@@ -565,13 +565,12 @@ static void secp256k1_modinv32(secp256k1_modinv32_signed30 *x, const secp256k1_m
 
     /* g == 0 */
     VERIFY_CHECK(secp256k1_modinv32_mul_cmp_30(&g, 9, &SECP256K1_SIGNED30_ONE, 0) == 0);
-    /* |f| == 1, or (x == 0 and d == 0 and |f|=modulus) */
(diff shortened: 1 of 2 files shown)


real-or-random (approved):
Okay yes.  All VERIFY_CHECKs are redundant in a sense, but this one is really covered by another VERIFY_CHECK already.
 
utACK e53ed2efebcca6483a511ac0d9f3162973d66ddathis is in

roconnor-blockstream:
Just to be clear, this PR isn't removing a redundant check, it is **strengthening** the existing check by removing a disjunctive clause.

real-or-random:
> Just to be clear, this PR isn't removing a redundant check, it is **strengthening** the existing check by removing a disjunctive clause.

Oh, right, I got this wrong. But I still think the change makes sense.

roconnor-blockstream:
Done.  I've also added the same change to the constant-time versions of these functions.

real-or-random (approved):
utACK ef7ff03407fa79e7eab1b6771b77f72828e63636

(There seems to be some CI hiccup unrelated to this PR.)

PR #1619, opened 2024-10-20
musig: ctimetests: fix _declassify range for generated nonce points

theStack:
As noticed in https://github.com/bitcoin-core/secp256k1/pull/1614#discussion_r1796215582, the area marked as non-secret exceeds the nonce_pts array in the second iteration of the for loop. Fix that by passing the correct size to the _declassify call.


real-or-random (approved):
utACK 57eda3ba300acb2e54f26a8ad00fbdd5750867ed

PR #1311, opened 2023-05-12
Revert "Remove unused scratch space from API"

jonasnick:
This reverts commit 712e7f8722eba5dec2bc6b37d75aadeb6f6e633b.

Removing the scratch space from the API may break bindings to the library.

PR #1620, opened 2024-10-21
Remove unused scratch space from API

jonasnick:
We had already merged this in #1305, but it was reverted before a release (#1311) because this change is not backwards compatible but at the time we only wanted to make a patch release in order to fix an actual issue.

Due to the musig module, the next release will increment the version number from 0.5.x to 0.6.0, so it would be a good time to remove the scratch space from the API.


real-or-random (approved):
utACK 8be3839fb2e1e501ef82c5b341e5a7f88587997a

PR #1622, opened 2024-10-22
[CI test, dontmerge] PR #1579 with secp256k1_memclear using volatile_memset

theStack:
As suggested in https://github.com/bitcoin-core/secp256k1/pull/1579#discussion_r1810990006.


real-or-random:
Okay, awesome CI seems to like this. Let me close this PR, I think this is a sufficient test for this #if branch, even if further small changes will be made to #1579.

PR #1628, opened 2024-10-31
Name public API structs

### LINKED ISSUE #1627: Name public API structs so they can be forward declared
Currently all of the structs in the public API are only typedef'd, rather than both named and typedef'd. As such, the structs cannot be forward declared and instead using the struct name (typedef) in a header requires including the library headers. This can be problematic for projects which want to reduce the inclusion tree as much as possible.

***

For some context behind this request, in my [MuSig2 PR to Bitcoin Core](https://github.com/bitcoin/bitcoin/pull/29675), since the `secp256k1_musig_secnonce` cannot be copied or serialized, I have to pass around a pointer to the object in various functions, so many headers need to include `secp256k1_musig.h` just for the struct's declaration. Including `secp256k1_musig.h` in those headers result in needing to directly link libsecp256k1 in several of build targets that did not previously require it. This issue seems like it can be avoided by forward declaring that (and other) struct, but in order to do so, it needs to be named in the libsecp headers.

My current workaround is to pass `void *` and then cast them as necessary but I don't really like doing that.

achow101:
Closes #1627

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -58,7 +58,7 @@ typedef struct secp256k1_context_struct secp256k1_context;
  *  use secp256k1_ec_pubkey_serialize and secp256k1_ec_pubkey_parse. To
  *  compare keys, use secp256k1_ec_pubkey_cmp.
  */
-typedef struct {
+typedef struct secp256k1_pubkey {
     unsigned char data[64];
 } secp256k1_pubkey;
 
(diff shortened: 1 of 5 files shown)


real-or-random (approved):
utACK 694342fdb71722b31dd4b6bc7241b992fdf6f492

Fwiw, I had discussed this with @jonasnick because we were not sure if reusing the struct tag as the type name is okay also in C++. But while C++ has slightly different rules here, they're made so that compatibility to C is ensured, see https://stackoverflow.com/questions/1675351/typedef-struct-vs-struct-definitions

jonasnick (approved):
ACK 694342fdb71722b31dd4b6bc7241b992fdf6f492

real-or-random:
We may want to add a changelog entry for PR. The change doesn't break anything, but it makes using the API more convenient, so users could be interested.

fanquake:
Changelog was added in #1631. Dropping "needs-changelog".

PR #1634, opened 2024-11-05
Fix some misspellings

niooss-ledger:
Hello,

Some files contained English misspellings or math issues (`lamba` instead of `lambda`), mainly in comments. Fixing them helps readability.

By the way, the misspellings found in the Wycheproof test vector file were also reported upstream: https://github.com/C2SP/wycheproof/issues/124

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,7 +55,7 @@ option(SECP256K1_INSTALL "Enable installation." ${PROJECT_IS_TOP_LEVEL})
 ## Modules
 
 # We declare all options before processing them, to make sure we can express
-# dependendencies while processing.
+# dependencies while processing.
 option(SECP256K1_ENABLE_MODULE_ECDH "Enable ECDH module." ON)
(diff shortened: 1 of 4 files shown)


real-or-random:
Thanks! Concept ACK, obviously.

> By the way, the misspellings found in the Wycheproof test vector file were also reported upstream: [C2SP/wycheproof#124](https://github.com/C2SP/wycheproof/issues/124)

Good, that's probably the right path.  And then I think it will be good then not to touch these files here. If you ask me, avoiding conflicts with the upstream is more important than fixing these typos.  Could you modify the commit?

niooss-ledger:
>  I think it will be good then not to touch these files here. If you ask me, avoiding conflicts with the upstream is more important than fixing these typos. Could you modify the commit?

Ok, this works for me. I removed the changes related to Wycheproof test vectors.

real-or-random (approved):
utACK 39705450ebcfc7a2a8be540c868da9d2c44900cb

PR #1638, opened 2024-11-23
tests: update ecdsa_secp256k1_sha256_bitcoin_test.json

eltociear:
implemenation -> implementation


real-or-random:
This has been fixed upstream in https://github.com/C2SP/wycheproof/pull/91 along with other changes. We should probably just reimport the file from upstream.

real-or-random:
The only other change is https://github.com/C2SP/wycheproof/commit/875f85e2129d5287d94c7f589516b24813af9b61 ... @eltociear Can you include this in the PR here?

real-or-random:
We should also update these lines then: https://github.com/bitcoin-core/secp256k1/blob/8deef00b33ca81202aca80fe0bcd9730f084fbd2/src/wycheproof/WYCHEPROOF_COPYING#L1-L4

PR #1639, opened 2024-11-23
Make static context const

### LINKED ISSUE #1637: `secp256k1_context_static` should be a `const` variable [edited]
Some language bindings have issues with exported variables. The static context is currently exported as a variable:

```c
SECP256K1_API const secp256k1_context *secp256k1_context_static;
```

which is implemented as:

```c
const secp256k1_context *secp256k1_context_static = &secp256k1_context_static_;
```

A more portable way would be to provide it through a function:

```c
SECP256K1_API const secp256k1_context *secp256k1_context_static(void);
```

with implementation:

```c
const secp256k1_context *secp256k1_context_static(void)
{
    return &secp256k1_context_static_;
}
```

PS: It is clear that changing the symbol into a function is a breaking change and therefore off the table. However, please consider adding a new symbol with the proposed implementation.

purpleKarrot:
Fixes #1637

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -230,10 +230,10 @@ typedef int (*secp256k1_nonce_function)(
  *
  *  It is highly recommended to call secp256k1_selftest before using this context.
  */
-SECP256K1_API const secp256k1_context *secp256k1_context_static;
+SECP256K1_API const secp256k1_context * const secp256k1_context_static;
 
 /** Deprecated alias for secp256k1_context_static. */
-SECP256K1_API const secp256k1_context *secp256k1_context_no_precomp
+SECP256K1_API const secp256k1_context * const secp256k1_context_no_precomp
 SECP256K1_DEPRECATED("Use secp256k1_context_static instead");
 
 /** Perform basic self tests (to be used in conjunction with secp256k1_context_static)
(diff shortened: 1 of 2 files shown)


real-or-random:
Looks good to me.

The only caveat is that this is, strictly speaking, a breaking change. If we want to be strict, we'd need a 0.7.0 instead of a 0.6.1. But I don't think we care?! I mean, any user code writing to the variable is most likely buggy... Hm.

real-or-random:
> it is only at runtime that an attempt to modify the variable may (or may not) trigger a memory violation.

The good thing is that a clean crash is probably the worst that can happen in practice...

I think I'm okay with anything here. If you ask me, we can treat this as a breaking or a non-breaking chance, when it comes to both API and ABI.

real-or-random (approved):
In any case, I prefer this proper fix over introducing a second symbol just to keep backwards compatibility.  

utACK bfc3ae0735597b676a0472ab6cbd0f909ee30195

We can debate at release time whether this is breaking or not.

real-or-random:
cc @jonasnick @sipa 

Can you take a look? I think this will be good to have merged because it's addressing a real user issue.

purpleKarrot:
I just rebased and signed the commit.

real-or-random (approved):
ACK 432ac57705b8e2f6a5874095214c443f1f8ea9bf

fanquake:

PR #1642, opened 2024-12-02
Verify `compressed` argument in `secp256k1_eckey_pubkey_serialize`

theStack:
Due to similarity to the public API function `secp256k1_ec_pubkey_serialize`, public API flags like `SECP256K1_EC_COMPRESSED` are sometimes mistakingly passed to `secp256k1_eckey_pubkey_serialize` in newly proposed code (this is currently the case for several modules in secp256k1-zkp, see https://github.com/BlockstreamResearch/secp256k1-zkp/pull/300), which is currently not detected. To avoid this in the future, a VERIFY_CHECK is added to check that the `compressed` argument is either 0 or 1.

### DIFF
--- a/src/eckey_impl.h
+++ b/src/eckey_impl.h
@@ -9,6 +9,7 @@
 
 #include "eckey.h"
 
+#include "util.h"
 #include "scalar.h"
 #include "field.h"
 #include "group.h"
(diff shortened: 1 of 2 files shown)


src/eckey_impl.h
```diff
@@ -35,6 +36,8 @@ static int secp256k1_eckey_pubkey_parse(secp256k1_ge *elem, const unsigned char
 }
 
 static int secp256k1_eckey_pubkey_serialize(secp256k1_ge *elem, unsigned char *pub, size_t *size, int compressed) {
+    VERIFY_CHECK(compressed == 0 || compressed == 1);
```
real-or-random, line 39: In other places, we simply rely on C's int-as-bool convention, so I suggest adding a comment that explains why we enforce 0/1 in this case.

real-or-random (approved):
utACK 1823594761046ae38e002781d65e9538dd211d27

PR #1646, opened 2024-12-11
README: add instructions for verifying GPG signatures

### LINKED ISSUE #1644: GPG signed verification of releases
Given the security-critical nature of this project, I think it would be preferable to have GPG-signed hashes available alongside source releases. Right now (AFAICT) this project is hinging completely on Github/HTTPS trust model when retrieving this repo for build and use.

Obviously the ~~hashes and signatures~~ GPG IDs would have to be posted somewhere aside from Github for full benefit.

I'm happy to help in whatever manner I can.

jamesob:
Fixes #1644.


README.md
```diff
@@ -61,6 +61,48 @@ Implementation details
...
+tag's signature using git.
+
+This can be done with the following steps:
+
+1. Check the latest release on the [Releases
+   page](https://github.com/bitcoin-core/secp256k1/releases). Determine the signing GPG ID
+   by clicking the green icon next to the tag name. For example, in the case of v0.6.0,
+   this would be `4861DBF262123605`.
+1. Cross-reference this key ID with another source controlled by its owner (e.g.
+   https://x.com/n1ckler).
```
jonasnick, line 79: Not sure if this step is security theater. Either this repo is legitimate and the GPG keys are correct or it's not and will have arbitrary instructions here.

One possible justification for including this step is educational: it reminds users how to increase assurance that a GPG key they encounter for the first time is correct. In that case I'd suggest to run the fingerprint through a search engine. Using google, this gives somewhat reasonable results for sipa and myself, but not for @real-or-random.

jonasnick (commented):
Thank you @jamesob. I think this is a significant improvement to our README.

README.md
jamesob: It may be overly paranoid, but I don't think it's security theater. The problem we're trying to mitigate here is a singular trust of what's presented on Github. I think definitionally, we need to consult a source that is not Github (or something obtained from Github) in order to get additional assurance that the keys we're verifying against are indeed your keys.

The thought experiment is "if Github is compromised, and the attacker can modify any of its content, will the instructions result in catching that?" I think consulting a second source for GPG IDs is necessary to fulfill that.

README.md
```diff
@@ -61,6 +61,48 @@ Implementation details
...
+For a fully verified build of this project, it is recommended to obtain this repository
+via git, obtain the GPG keys of the signing maintainer(s), and then verify the release
+tag's signature using git.
+
+This can be done with the following steps:
+
+1. Check the latest release on the [Releases
+   page](https://github.com/bitcoin-core/secp256k1/releases). Determine the signing GPG ID
+   by clicking the green icon next to the tag name. For example, in the case of v0.6.0,
+   this would be `4861DBF262123605`.
```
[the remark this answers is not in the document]
jamesob: Ah yeah I did miss that, thanks. I'll revise this change to point to that file.

jamesob:
@jonasnick I've pushed some changes based on your feedback, hope it helps.

README.md
```diff
@@ -61,6 +61,51 @@ Implementation details
...
+    - Consult [SECURITY.md](./SECURITY.md) for a complete listing of maintainer keys,
+      per Github.
+1. If possible, cross-reference this key ID with another source controlled by its owner (e.g.
+   https://x.com/n1ckler).
+    - This is to guard against the unlikely case that incorrect content is being presented by Github.com.
```
real-or-random, line 82: It's bad enough that we rely on GitHub, so I'd avoid links to other third-party sites. It's up to the user to decide which source they trust, and I don't think it's our job to offer suggestions here. That's bad, but it's a fundamental problem. We cannot do much about it. (For what it's worth, I don't trust x.com too much.) I expect people who obtain the source code of a C library to be able to do their research.

README.md
```diff
@@ -61,6 +61,51 @@ Implementation details
...
+    ```
+1. Clone the repository: 
+    ```
+    git clone https://github.com/bitcoin-core/secp256k1
+    ```
+1. Check out the latest release tag, e.g. 
+    ```
+    git checkout v0.6.0
+    ```
+1. Use git to verify the GPG signature:
```
real-or-random, line 95: nit: I prefer "1. / 2. / 3."  instead of "1. / 1. / 1.", i.e., I prefer readability of the plain text file over maintainability.

README.md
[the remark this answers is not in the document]
real-or-random: Now that you've added the reference to the file, why not drop this step in the GitHub web interface entirely? 

I think the steps should be:
 - Obtain the keys in SECURITY.md
 - Use the command line to verify the tag.
 - If successful, compare the key fingerprint with a trusted source. (I think it's a good idea to compare the full fingerprint, and not just the long key ID.)

None of these require the web (except maybe obtaining a trusted copy of the fingerprint).

README.md
```diff
@@ -61,6 +61,51 @@ Implementation details
...
+via git, obtain the GPG keys of the signing maintainer(s), and then verify the release
+tag's signature using git.
+
+This can be done with the following steps:
+
+1. Check the latest release on the [Releases
+   page](https://github.com/bitcoin-core/secp256k1/releases). Determine the signing GPG ID
+   by clicking the green icon next to the tag name. For example, in the case of v0.6.0,
+   this would be `4861DBF262123605`.
+    - Consult [SECURITY.md](./SECURITY.md) for a complete listing of maintainer keys,
```
real-or-random, line 78: "per Github" assumes that the user has obtained the git repo from GitHub, which is likely but not the only possibility.

README.md
jonasnick: Ok, so iiuc the practical failure mode this can prevent is when instructions and SECURITY.md are obtained from two different sources somehow (or same source, at different times) and only SECURITY.md is compromised. Seems reasonable.

jamesob:
Pushed a more succinct version that I think covers all feedback

README.md
```diff
@@ -61,6 +61,42 @@ Implementation details
...
+For a fully verified build of this project, it is recommended to obtain this repository
+via git, obtain the GPG keys of the signing maintainer(s), and then verify the release
+tag's signature using git.
+
+This can be done with the following steps:
+
+1. Obtain the GPG keys listed in [SECURITY.md](./SECURITY.md).
+2. If possible, cross-reference these key IDs with another source controlled by its owner (e.g.
+   social media, personal website).
+    - This is to mitigate the unlikely case that incorrect content is being presented by this repository.
```
real-or-random, line 77: if you touch this again,  no need to use a list: 
```suggestion
2. If possible, cross-reference these key IDs with another source controlled by its owner (e.g. social media, personal website).
   This is to mitigate the unlikely case that incorrect content is being presented by this repository.
```

real-or-random (approved):
ACK 918072709397da5cbe8d2a04b9defd343539ceeb

README.md
jamesob: Fixed

jonasnick (approved):
ACK b682dbcf845d80f69c4e2d8243c90dd48006a4c1

PR #1647, opened 2024-12-13
cmake: Adjust diagnostic flags for `clang-cl`

hebasto:
When building with `clang-cl` on Windows, the output is cluttered with warning messages because compiler diagnostic flags are not applied correctly:
```
> cmake -B build -G Ninja -DCMAKE_C_COMPILER="C:\Users\hebasto\Downloads\clang+llvm-18.1.8-x86_64-pc-windows-msvc\bin\clang-cl.exe"
> cmake --build build
[1/16] Building C object src\CMakeFiles\bench.dir\bench.c.obj
In file included from C:\Users\hebasto\secp256k1\src\bench.c:11:
C:\Users\hebasto\secp256k1\src\util.h(34,13): warning: unused function 'print_buf_plain' [-Wunused-function]
   34 | static void print_buf_plain(const unsigned char *buf, size_t len) {
      |             ^~~~~~~~~~~~~~~
1 warning generated.
[2/16] Building C object src\CMakeFiles\secp256k1_precomputed.dir\precomputed_ecmult_gen.c.obj
In file included from C:\Users\hebasto\secp256k1\src\precomputed_ecmult_gen.c:3:
In file included from C:\Users\hebasto\secp256k1\src\group.h:10:
In file included from C:\Users\hebasto\secp256k1\src\field.h:10:
C:\Users\hebasto\secp256k1\src\util.h(34,13): warning: unused function 'print_buf_plain' [-Wunused-function]
   34 | static void print_buf_plain(const unsigned char *buf, size_t len) {
      |             ^~~~~~~~~~~~~~~
```

This PR resolves this issue.

---

**Additional note for reviewers:** The VS builtin clang can also be used assuming that the following VS components are installed:

photo_2024-12-12_12-38-17

The user can generate a build system on Windows as follows:
- Using the default "Visual Studio" generator:
```
cmake -B build -T ClangCL
```
- Using the "Ninja" generator:
```
cmake -B build -G Ninja -DCMAKE_C_COMPILER=clang-cl
```

---

Required for downstream projects which aim to build with `clang-cl` (see https://github.com/bitcoin/bitcoin/issues/31456).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -242,17 +242,21 @@ endif()
 
 include(TryAppendCFlags)
 if(MSVC)
-  # Keep the following commands ordered lexicographically.
+  # For both cl and clang-cl compilers.
   try_append_c_flags(/W3) # Production quality warning level.
-  try_append_c_flags(/wd4146) # Disable warning C4146 "unary minus operator applied to unsigned type, result still unsigned".
-  try_append_c_flags(/wd4244) # Disable warning C4244 "'conversion' conversion from 'type1' to 'type2', possible loss of data".
-  try_append_c_flags(/wd4267) # Disable warning C4267 "'var' : conversion from 'size_t' to 'type', possible loss of data".
   # Eliminate deprecation warnings for the older, less secure functions.
   add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
 else()
+  try_append_c_flags(-Wall) # GCC >= 2.95 and probably many other compilers.
+endif()  
+if(CMAKE_C_COMPILER_ID STREQUAL "MSVC")
+  # Keep the following commands ordered lexicographically.
+  try_append_c_flags(/wd4146) # Disable warning C4146 "unary minus operator applied to unsigned type, result still unsigned".
+  try_append_c_flags(/wd4244) # Disable warning C4244 "'conversion' conversion from 'type1' to 'type2', possible loss of data".
+  try_append_c_flags(/wd4267) # Disable warning C4267 "'var' : conversion from 'size_t' to 'type', possible loss of data".
+else()
   # Keep the following commands ordered lexicographically.
   try_append_c_flags(-pedantic)
-  try_append_c_flags(-Wall) # GCC >= 2.95 and probably many other compilers.
   try_append_c_flags(-Wcast-align) # GCC >= 2.95.
   try_append_c_flags(-Wcast-align=strict) # GCC >= 8.0.
   try_append_c_flags(-Wconditional-uninitialized) # Clang >= 3.0 only.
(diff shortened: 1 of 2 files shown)


real-or-random:
I guess the patch is trivial enough not to bother, but I'm not entirely sure if I understand the purpose of this. Couldn't you build using clang on Windows directly? Or is the purpose of this PR to help users who want to use Visual Studio but with clang instead of MSVC? My thinking is that if you use Visual Studio for an application that depends on libsecp256k1, you don't bother how libsecp256k1 itself is built.

real-or-random (approved):
utACK 4c50d73dd98edc0b968525a291cf14b708bbdbf1

CMakeLists.txt
```diff
@@ -242,17 +242,21 @@ endif()
...
-  # Keep the following commands ordered lexicographically.
+  # For both cl and clang-cl compilers.
   try_append_c_flags(/W3) # Production quality warning level.
+  # Eliminate deprecation warnings for the older, less secure functions.
+  add_compile_definitions(_CRT_SECURE_NO_WARNINGS)
+else()
+  try_append_c_flags(-Wall) # GCC >= 2.95 and probably many other compilers.
+endif()  
+if(CMAKE_C_COMPILER_ID STREQUAL "MSVC")
+  # Keep the following commands ordered lexicographically.
```
theuni, line 253: I'd prefer to set `/W3` here and `-Wall` with the other `try_append_c_flags` for consistency, as I don't see why it needs to be special-cased.

theuni (commented):
utACK other than the nit.

CMakeLists.txt
hebasto: According to the [docs](https://clang.llvm.org/docs/UsersManual.html#id11), the `/Wall` option in clang-cl, which is equivalent to `-Wall`, enables Clang's core `-Weverything` option.

We use `/W3`, which enables Clang's core `-Wall` option.

theuni: Wow. Confirmed: https://godbolt.org/z/WEr3qGeox

That's... hugely unexpected. It really doesn't give me much faith in the proper forwarding of other options :(

hebasto:
@real-or-random @jonasnick 

Ready to merge?

CMakeLists.txt
real-or-random: > That's... hugely unexpected. It really doesn't give me much faith in the proper forwarding of other options :(

Hm, I see that it can be unexcepted, but I also understand why they do this, and mapping these generic warning options is only a best-effort thing. (I believe the objectively most confusing thing here is the existence of `-Wall` alongside `-Weverything`... It's just not confusing to us because we're familar with gcc.)

PR #1648, opened 2024-12-13
cmake: Improve handling of `SECP256K1_APPEND_*FLAGS` options

hebasto:
On the master branch @ f79f46c70386c693ff4e7aef0b9e7923ba284e56, setting `SECP256K1_APPEND_*FLAGS` when using [IDE build tool generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#ide-build-tool-generators), such as "Visual Studio" and "Xcode", has no effect because these generators do not utilise low-level rule variables when creating project files.

This PR disables `SECP256K1_APPEND_*FLAGS` options in such scenarios. CMake will issue a warning if any of them are set.

Additionally, this PR fixes a bug with the clang-cl compiler (see https://github.com/bitcoin-core/secp256k1/pull/1647):
```
> cmake -B build -G Ninja -DCMAKE_C_COMPILER=clang-cl -DSECP256K1_APPEND_CFLAGS=/W4
> cmake --build build
...
clang-cl: error: no such file or directory: '/W4'
...
```

The last change is beneficial in its own right, as it makes the compile invocation string more natural by ensuring that flags do not follow source files.


real-or-random:
Can you indicate why this is a draft?

hebasto:
> Can you indicate why this is a draft?

I haven't tested this branch with Xcode on macOS yet.

hebasto:
> > Can you indicate why this is a draft?
> 
> I haven't tested this branch with Xcode on macOS yet.

Reworked to support only [command-line build tool generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#command-line-build-tool-generators).

Undrafted and ready for review :)

gatleas17 (approved):

CMakeLists.txt
```diff
@@ -280,19 +280,21 @@ if(SECP256K1_BUILD_CTIME_TESTS)
   unset(msan_enabled)
 endif()
 
-set(SECP256K1_APPEND_CFLAGS "" CACHE STRING "Compiler flags that are appended to the command line after all other flags added by the build system. This variable is intended for debugging and special builds.")
-if(SECP256K1_APPEND_CFLAGS)
-  # Appending to this low-level rule variable is the only way to
+if(CMAKE_GENERATOR MATCHES "Make|Ninja")
+  # Only command-line build tool generators support low-level rule
+  # variables.
```
real-or-random, line 285: nit:

While this seems to be true, I think it's a coincidence that these happen to be the command-line tools.
```suggestion
# Only some generators use low-level rule variables.
if(CMAKE_GENERATOR MATCHES "Make|Ninja")
```

And I think the command should be above the `if` line because it comments on the `if`.

CMakeLists.txt
```diff
@@ -287,7 +287,7 @@ if(CMAKE_GENERATOR MATCHES "Make|Ninja")
   # guarantee that the flags appear at the end of the command line.
   set(SECP256K1_APPEND_CFLAGS "" CACHE STRING "Compiler flags that are appended to the command line after all other flags added by the build system. This variable is intended for debugging and special builds.")
   if(SECP256K1_APPEND_CFLAGS)
-    string(APPEND CMAKE_C_COMPILE_OBJECT " ${SECP256K1_APPEND_CFLAGS}")
+    string(REPLACE "<FLAGS>" "<FLAGS> ${SECP256K1_APPEND_CFLAGS}" CMAKE_C_COMPILE_OBJECT "${CMAKE_C_COMPILE_OBJECT}")
```
real-or-random, line 290: Do you think this same should be done for `CMAKE_C_CREATE_SHARED_LIBRARY` and/or `CMAKE_C_LINK_EXECUTABLE`?

real-or-random (commented):
> On the master branch @ [f79f46c](https://github.com/bitcoin-core/secp256k1/commit/f79f46c70386c693ff4e7aef0b9e7923ba284e56), setting `SECP256K1_APPEND_*FLAGS` when using [IDE build tool generators](https://cmake.org/cmake/help/latest/manual/cmake-generators.7.html#ide-build-tool-generators), such as "Visual Studio" and "Xcode", has no effect because these generators do not utilise low-level rule variables when creating project files.
> 
> This PR disables `SECP256K1_APPEND_*FLAGS` options in such scenarios. CMake will issue a warning if any of them are set.


Concept ACK

Special casing generators is not elegant, but yeah, it's probably a good idea. Otherwise, anyone who wants to use `SECP256K1_APPEND_CFLAGS` on an unsupported build system will waste hours trying to understand why this is the case.

CMakeLists.txt
hebasto: Thanks! Done.

real-or-random (approved):
utACK c12bc0163b03210e6187906273c6063dd94aa284

Datis77re (approved):

real-or-random:
cc @theuni

Interested in reviewing this?

CMakeLists.txt
```diff
@@ -280,19 +280,20 @@ if(SECP256K1_BUILD_CTIME_TESTS)
...
-  # Appending to this low-level rule variable is the only way to
+# Only some generators use low-level rule variables.
+if(CMAKE_GENERATOR MATCHES "Make|Ninja")
+  # Appending to the low-level rule variables is the only way to
   # guarantee that the flags appear at the end of the command line.
-  string(APPEND CMAKE_C_COMPILE_OBJECT " ${SECP256K1_APPEND_CFLAGS}")
-endif()
+  set(SECP256K1_APPEND_CFLAGS "" CACHE STRING "Compiler flags that are appended to the command line after all other flags added by the build system. This variable is intended for debugging and special builds.")
+  if(SECP256K1_APPEND_CFLAGS)
+    string(APPEND CMAKE_C_COMPILE_OBJECT " ${SECP256K1_APPEND_CFLAGS}")
```
theuni: I think we want an else case here to throw an error if using (for example) xcode + `SECP256K1_APPEND_CFLAGS` so that they're not just silently dropped. If that's an unsupported combo, it should be reported as such.

Edit: throw an error, not a warning.

CMakeLists.txt
```diff
@@ -286,13 +286,13 @@ if(CMAKE_GENERATOR MATCHES "Make|Ninja")
   # guarantee that the flags appear at the end of the command line.
   set(SECP256K1_APPEND_CFLAGS "" CACHE STRING "Compiler flags that are appended to the command line after all other flags added by the build system. This variable is intended for debugging and special builds.")
   if(SECP256K1_APPEND_CFLAGS)
-    string(APPEND CMAKE_C_COMPILE_OBJECT " ${SECP256K1_APPEND_CFLAGS}")
+    string(REPLACE "<FLAGS>" "<FLAGS> ${SECP256K1_APPEND_CFLAGS}" CMAKE_C_COMPILE_OBJECT "${CMAKE_C_COMPILE_OBJECT}")
```
theuni, line 254: Could you please add a comment here explaining what's going on?

Something like:
"`CMAKE_C_COMPILE_OBJECT` internally uses placeholders to maintain the order of certain fields. Rather than appending to `CMAKE_C_COMPILE_OBJECT` itself, append to its `<FLAGS>`. This is an implementation detail and not guaranteed to be stable in future versions of CMake"

theuni (changes requested):
Concept ACK. This is annoying and ugly, but makes sense to me.

Requested a few changes.

hebasto:
The feedback from @theuni has been addressed. Rebased.

CMakeLists.txt
hebasto: Thanks! [Reworked](https://github.com/bitcoin-core/secp256k1/pull/1648#issuecomment-4387260230).

CMakeLists.txt
hebasto: Thanks! The comment has been [added](https://github.com/bitcoin-core/secp256k1/pull/1648#issuecomment-4387260230).

hebasto:

PR #1650, opened 2025-01-29
schnorrsig: clear out masked secret key in BIP-340 nonce function

theStack:
Considering that the secret key can be determined from `masked_key` if the passed auxiliary data is known (which is optional and set to all-zeros if not explicitly provided, see `ZERO_MASK`), it seems reasonable to clear it from the stack.


real-or-random (approved):
utACK a82287fb85746e0f7ea142a68a859e56ee86ee10

jonasnick (approved):
ACK a82287fb85746e0f7ea142a68a859e56ee86ee10

PR #1651, opened 2025-02-03
add DLEQ proof based on BIP374 and use it for silent payments

stratospher:
(EDIT: maybe look at https://github.com/bitcoin-core/secp256k1/pull/1802 first which PRs just the DLEQ part)

built on top of c4942d36.

> BIP352 requires senders to compute output scripts using ECDH shared secrets from the same secret keys used to sign the inputs. Generating an incorrect signature will produce an invalid transaction that will be rejected by consensus. An incorrectly generated output script can still be consensus-valid, meaning funds may be lost if it gets broadcast. By producing a DLEQ proof for the generated ECDH shared secrets, the signing entity can prove to other entities that the output scripts have been generated correctly without revealing the private keys.

(from [BIP 374](https://github.com/bitcoin/bips/blob/master/bip-0374.mediawiki))

This PR:

- adds support for DLEQ proof generation and verification using `secp256k1_dleq_prove`, `secp256k1_dleq_verify` based on implementation in [secp256k1-zkp](https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/src/modules/ecdsa_adaptor/dleq_impl.h)
- introduces a new structure for storing proofs in silent payment module - `secp256k1_silentpayments_dleq_data` which stores the shared secret computed using ECDH, DLEQ proof and also the index of which recipient in the [original unsorted recipient array](https://github.com/bitcoin-core/secp256k1/blob/f42e0dde59943ad0c9247c61f7af521b70838e8f/include/secp256k1_silentpayments.h#L43)  the proof refers to.
- adds 4 new APIs in silent payments module to create, verify, serialize and parse proof
    1. sender can create proof using `secp256k1_silentpayments_sender_create_outputs_with_proof`
        - in the existing `secp256k1_silentpayments_sender_create_outputs` API, output pubkey is created by iterating over all the recipients and computing ECDH for each unique recipient.
        - since creating the proof also requites iterating over all recipients and computing ECDH for each unique recipient, compute both proof(s) and output pubkey(s) in a new function - `secp256k1_silentpayments_sender_create_outputs_with_proof`.
        - `secp256k1_silentpayments_sender_create_outputs` can use `secp256k1_silentpayments_sender_create_outputs_with_proof`internally.
    2.  both sender and receiver can verify proof using `secp256k1_silentpayments_verify_proof`
    3. Serialisation function (`secp256k1_silentpayments_dleq_data_serialize`) serialises  `secp256k1_silentpayments_dleq_data` structure into bytes. `secp256k1_silentpayments_dleq_data_parse` parses the bytes back into `secp256k1_silentpayments_dleq_data` structure.
        1. useful if proof creation is done on 1 sender side device (ex: hardware wallet) and proof verification is done on another sender side device (ex: software wallet). `secp256k1_silentpayments_dleq_data` structure can be serialised into bytes using `secp256k1_silentpayments_dleq_data_serialize` and then the bytes can be sent from 1 device to another.
        2. `secp256k1_silentpayments_dleq_data_parse` can reconstruct the bytes back into `secp256k1_silentpayments_dleq_data` structure which can then be used for proof verification.

open questions:

1. is the API design ok?
    1. see `examples/silentpayments.c` for 2 scenarios in which proof verification could be useful. (Alice sends bitcoins to Bob and Carol in 1 silent payment transaction)
        1. useful situation - On Alice’s side
            1. Alice’s hardware wallet creates proof
            2. Alice’s hardware wallet sends proof to Alice’s software wallet
            3. Alice’s software wallet verifies proof and makes sure ECDH shared secrets for generating SP output is computed correctly. Alice can get additional safety guarantee about computed SP output before sending funds.
        2. not sure if this situation is useful - On Bob’s/Carol’s side
            1. Alice’s software wallet sends proof to Bob’s/Carol’s  software wallet
            2. Bob’s/Carol’s software wallet verifies proof and makes sure ECDH shared secrets for generating SP output is computed correctly.
2. should this be shipped with SP module or separately? only problem with separate release would be redundant APIs like `secp256k1_silentpayments_sender_create_outputs` and `secp256k1_silentpayments_sender_create_outputs_with_proof` for backward compatibility. 

Huge thanks to @ theStack for brainstorming and discussing this with me! ❤️

### DIFF
--- a/examples/silentpayments.c
+++ b/examples/silentpayments.c
@@ -18,6 +18,7 @@
 
 #define N_INPUTS  2
 #define N_OUTPUTS 3
+#define N_RECIPIENTS 2
 
 /* Static data for Bob and Carol's silent payment addresses */
 static unsigned char smallest_outpoint[36] = {
@@ -127,6 +128,7 @@ int main(void) {
     secp256k1_xonly_pubkey *tx_output_ptrs[N_OUTPUTS];
     int ret;
     size_t i;
+    unsigned char dleq_proof[N_RECIPIENTS][64];
 
     /* Before we can call actual API functions, we need to create a "context" */
     secp256k1_context* ctx = secp256k1_context_create(SECP256K1_CONTEXT_NONE);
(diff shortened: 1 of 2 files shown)


jonatack:
Concept ACK

real-or-random:
> > ```
> > 2. should this be shipped with SP module or separately? only problem with separate release would be redundant APIs like `secp256k1_silentpayments_sender_create_outputs` and `secp256k1_silentpayments_sender_create_outputs_with_proof` for backward compatibility.
> > ```
> 
> @josibake Do you have an opinion on this?

friendly ping :)

josibake:
Concept ACK

> > > ```
> > > 2. should this be shipped with SP module or separately? only problem with separate release would be redundant APIs like `secp256k1_silentpayments_sender_create_outputs` and `secp256k1_silentpayments_sender_create_outputs_with_proof` for backward compatibility.
> > > ```
> > 
> > 
> > @josibake Do you have an opinion on this?
> 
> friendly ping :)

Ayo, late to the party here 😅 First off, thanks for picking this up @stratospher ! I only briefly looked at the module, so all commentary here is very high level. Regarding shipping with the silent payments module, I don't see any reason to couple the two. Rather, I would consider the silent payments module a client of the DLEQ module.

> only problem with separate release would be redundant APIs like `secp256k1_silentpayments_sender_create_outputs` and `secp256k1_silentpayments_sender_create_outputs_with_proof`

I don't see this as a problem. An alternative to this would be to extend `create_outputs` with an unused out param in anticipation of the proof, but I much prefer what you have here with `create_outputs_with_proof`. There will also still be use cases where computing the proof is unnecessary (e.g., single party sender where sender has full access to the private keys) and I think two separate functions in the API is clearer than a single function with an optional output parameter.

> not sure if this situation is useful - On Bob’s/Carol’s side
Alice’s software wallet sends proof to Bob’s/Carol’s software wallet
Bob’s/Carol’s software wallet verifies proof and makes sure ECDH shared secrets for generating SP output is computed correctly.

I haven't thought this all the way through, but I believe this could function as a proof of payment. Imagine Alice has paid Bob and wants to prove this to a 3rd party. Alice could compute the DLEQ proof and then pass this with the along with the shared secret to the 3rd party, who can then verify the DLEQ proof and use the shared secret to verify that the the SP output is present in the transaction. All this to say, I think having an example of this would be helpful!

src/modules/silentpayments/dleq_impl.h
```diff
@@ -0,0 +1,235 @@
...
+    if (!secp256k1_eckey_pubkey_serialize(B, B_33, &pubkey_size, 1)) {
+        return 0;
+    }
+    if (!secp256k1_eckey_pubkey_serialize(A, A_33, &pubkey_size, 1)) {
+        return 0;
+    }
+    if (!secp256k1_eckey_pubkey_serialize(C, C_33, &pubkey_size, 1)) {
+        return 0;
+    }
+    ret &= secp256k1_dleq_nonce(&k, a32, A_33, C_33, aux_rand32);
```
theStack: as per the latest version of the BIP (0.2.0), the nonce should also include the message, see https://github.com/bitcoin/bips/blob/3d0bab3cc211be9d40c0029f62c7e4eebd27ea21/bip-0374.mediawiki?plain=1#L77, I guess the test vectors are outdated and would fail if re-generated with the latest .csv file

src/modules/silentpayments/dleq_impl.h
```diff
@@ -0,0 +1,235 @@
...
+        for (i = 0; i < 32; i++) {
+            masked_key[i] = key32[i] ^ ZERO_MASK[i];
+        }
+    }
+
+    secp256k1_nonce_function_bip374_sha256_tagged(&sha);
+    /* Hash masked-key||msg using the tagged hash as per the spec */
+    secp256k1_sha256_write(&sha, masked_key, 32);
+    secp256k1_sha256_write(&sha, msg, msglen);
+    secp256k1_sha256_finalize(&sha, nonce32);
```
theStack, line 94: should memclear the `sha` and `masked_key` instances here (as also done in the [BIP340 nonce function](https://github.com/bitcoin-core/secp256k1/blob/baa265429fa8f1686138380e52a75c25b0344719/src/modules/schnorrsig/main_impl.h#L96))

src/modules/silentpayments/dleq_impl.h
```diff
@@ -0,0 +1,235 @@
...
+    secp256k1_declassify(ctx, &R1, sizeof(R1));
+    secp256k1_declassify(ctx, &R2, sizeof(R2));
+
+    /* e = tagged hash(A, B, C, R1, R2) */
+    /* s = k + e * a */
+    secp256k1_dleq_challenge(e, B, &R1, &R2, A, C, m);
+    secp256k1_scalar_mul(s, e, a);
+    secp256k1_scalar_add(s, s, &k);
+
+    secp256k1_scalar_clear(&k);
```
theStack, line 195: should also clear out the secret key serialization `a32` here

include/secp256k1_silentpayments.h
```diff
@@ -444,6 +457,30 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_silentpayments_recipien
...
 
+/** Serialize a secp256k1_silentpayments_dleq_data object into a 101-byte sequence.
+ *  101-byte sequence = 33 bytes shared secret + 64 bytes proof + 4 bytes index
+ *                      where index is position in an array of pointers to silent payment recipients
+ *
+ *  Out:       output: pointer to a 101-byte array to place the serialized `secp256k1_silentpayments_dleq_data` in
+ *  In:    dleq_data: pointer to an initialized secp256k1_silentpayments_dleq_data object
+ */
+SECP256K1_API void secp256k1_silentpayments_dleq_data_serialize(
+    unsigned char *output33,
```
theStack:
```suggestion
    unsigned char *output101,
```

w0xlt:
Concept ACK

stratospher:
update: https://github.com/bitcoin-core/secp256k1/pull/1802 takes the first 2 commits from this PR and exposes the `dleq_prove` and `dleq_verify` API.

thank you to @macgyver13 for taking this forward! it is being used in [Silent Payments with PSBT (BIP375)](https://github.com/bitcoin/bips/blob/master/bip-0375.mediawiki#computing-the-ecdh-shares-and-dleq-proofs) - https://github.com/macgyver13/rust-dleq

I think the DLEQ functionality can be used in other contexts as well, so it makes sense for it to go in independently of the full Silent Payments implementation. I'm not actively addressing comments in this SP+DLEQ PR since it depends on SP and would like to look at it first.

(sorry for the confusion! will review #1802!)

PR #1618, opened 2024-10-15
Examples: swap magic numbers for EXIT_ codes

Cheapshot003:
As discussed in #1609 this PR will use EXIT_{success, failure} codes instead of magic numbers in the return statements.

If it looks good I will continue to remove the asserts too in another commit.

Issue #1609, opened 2024-09-27
use EXIT_ codes instead of magic numbers for exit(...) and main return values


theStack:
This is really only a minor issue, but I noticed while reviewing #1479 that the return codes of functions in the examples could potentially be confusing. Throughout the API and internal functions we use 0=failure/1=success, while for the main function and (`exit(...)`) it's the other way round, i.e. 0=success/1=failure. We could use `EXIT_{SUCCESS,FAILURE}` (defined in stdlib.h, see https://en.cppreference.com/w/c/program/EXIT_status) for the latter instead for more clarity.

See e.g. https://github.com/bitcoin/bitcoin/pull/9067/commits/4441018d0860fce64ee74fa78da79bbb21114ca9 for a comparable change in Bitcoin Core as orientation. This could be a good first issue.

real-or-random:
Concept ACK

I had the same thought in the past.

This issue is slightly related to my suggestion to remove `assert()` in the examples, which could be addressed in another commit in a single PR.

Cheapshot003:
While working on this I noticed a mistake(?) in the comments in the verification section of ecdsa.c line 95 and 101 (probably other examples too).

  
https://github.com/Cheapshot003/secp256k1/blob/01b5893389efe6615d9ec2ee5f7f0f46c44b1c41/examples/ecdsa.c#L92C1-L102C6

The comment says "This will return 0 if the signature can't be parsed correctly" however the return value in case of failure is 1. Shall I just adjust it together with the other return values and replace it with EXIT_FAILURE?

real-or-random:
> The comment says "This will return 0 if the signature can't be parsed correctly" however the return value in case of failure is 1. Shall I just adjust it together with the other return values and replace it with EXIT_FAILURE?

I think what is meant here is that `secp256k1_ecdsa_signature_parse_compact` (and the other called function) will return 0 in case of failure. (And yes, then the main function in the example will return 1, which is precisely demonstrates the confusion described in this issue.)

But yeah, the fact that you misread this is proof that this is misleading and should be improved.

> Shall I just adjust it together with the other return values and replace it with EXIT_FAILURE?

[...]

PR #1654, opened 2025-02-14
use `EXIT_` constants over magic numbers for indicating program execution status

theStack:
This simple PR addresses #1609 for all example and internal binaries. Alternative to #1618, which is stale (the author confirmed to me that they are not working on that PR anymore). The last commits adds a suggestion to CONTRIBUTING.md, not sure though if we want to go that far.


jonasnick (approved):
ACK 13d389629aba2c686798a856332c00c6934cacd5

PR #1655, opened 2025-03-10
gha: Print all *.log files, in a separate action

real-or-random:
Before this commit, we didn't print *_example.log files and
test_suite.log.

Printing is now handled in a separate action, which avoids code
duplication and makes the ci.yml file more readable. This changes the
folding/grouping of the log output in the GitHub Actions CI, but I
think the new variant is as good as the old one.

Furthermore, the condition for printing the logs is changed from
"always()" to "!cancelled()". This ensures that logs will still be
printed if previous steps such as the CI script failed, but that they
won't be printed if the entire run is cancelled (e.g., by clicking a
button in the UI or through a force-push to the PR). This is in line
with a recommendation in the GHA docs:
https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#always

### DIFF
--- a/.github/actions/print-logs/action.yml
+++ b/.github/actions/print-logs/action.yml
@@ -1,0 +1,34 @@
+name: "Print logs"
+description: "Print the log files produced by ci/ci.sh"
+runs:
+  using: "composite"
+  steps:
+    - shell: bash
+      run: |
+        # Print the log files produced by ci/ci.sh
+
+        # Helper functions
+        group() {
+            title=$1
+            echo "::group::$title"
+        }
+        endgroup() {
+            echo "::endgroup::"
+        }
+        cat_file() {
+            file=$1
+            group "$file"
+            cat "$file"
+            endgroup
+        }
+
+        # Print all *.log files
+        shopt -s nullglob
+        for file in *.log; do
+            cat_file "$file"
+        done
+
+        # Print environment
+        group "CI env"
+        env
+        endgroup
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -107,23 +107,9 @@ jobs:
           dockerfile: ./ci/linux-debian.Dockerfile
           tag: linux-debian-image
 
-      - run: cat tests.log || true
-        if: ${{ always() }}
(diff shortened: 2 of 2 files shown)


real-or-random:
Ready for review. If you prefer, I could split the changes into separate commits, but I don't think it's worth the hassle. It's not entirely clear how to separate these changes, and doing so would increase the sum of the diff because later commits will need to touch lines moved in earlier commits.

real-or-random:
@hebasto Can you take a look at this?

.github/actions/print-logs/action.yml
```diff
@@ -0,0 +1,31 @@
...
+            group "$file"
+            cat "$file"
+            endgroup
+        }
+
+        # Print all *.log files
+        shopt -s nullglob
+        for file in *.log; do
+            cat_file "$file"
+        done
```
hebasto, line 26: The `make.log` file is now printed along with other logs.

Does it make sense to separate logs by groups:
- `config.log` and `make.log`
- tests and benchmarks
- examples
?

.github/actions/print-logs/action.yml
```diff
@@ -0,0 +1,31 @@
+name: "Print logs"
+description: "Print the log files produced by ci/ci.sh"
+runs:
+  using: "composite"
+  steps:
+    - shell: bash
+      run: |
+        group() {
```
hebasto, line 8: The first line of the `run` command is printed as a group header in the Web interface. It could be a meaningful comment, for example: `# Print the log files produced by ci/ci.sh`.

hebasto (commented):
Concept ACK.

.github/actions/print-logs/action.yml
real-or-random: I agree that this would produce better output, but I also think that `*.log` is good enough and much easier to maintain.

.github/actions/print-logs/action.yml
real-or-random: Yeah, it's annoying that it's not possible to get rid of this line, but making it contain a comment is a great idea. Fixed.

PR #1659, opened 2025-03-13
include: remove WARN_UNUSED_RESULT for functions always returning 1

### LINKED ISSUE #1379: Remove `SECP256K1_WARN_UNUSED_RESULT` from `secp256k1_keypair_*` functions
Why do `secp256k1_keypair_*` functions have `SECP256K1_WARN_UNUSED_RESULT` if they always return 1 ?
The only exception is `secp256k1_keypair_create`.

jonasnick:
This makes the usage of the atribute consistent. In the musig and ellswift module, functions that return 1 always already don't have the WARN_UNUSED_RESULT attribute. In secp256k1.h and the extrakeys module, this has only been the case partially.

In all cases where this was removed, the function only returns 0 if the illegal callback has been called.

Fixes #1379


fanquake:

PR #1660, opened 2025-03-13
ci: Fix exiting from ci.sh on error

real-or-random:
Fixes the following bash error when make fails:
    ./ci/ci.sh: line 100: return: can only `return' from a function or
    sourced script


hebasto (approved):
ACK d87c3bc58f369a91fb07d0812eec8b0d77e8968b

PR #1668, opened 2025-03-27
bench_ecmult: add benchmark for ecmult_const_xonly

theStack:
In the course of trying to understand the x-only ECDH benchmark results in the silent payments PR (https://github.com/bitcoin-core/secp256k1/pull/1519#discussion_r2006144436), I noticed that we don't have a benchmark yet for `ecmult_const_xonly`, so I added one.

Results on my machine:
```
$ ./build/bin/bench_ecmult

ecmult_gen                    ,    12.3       ,    12.6       ,    13.2
ecmult_const                  ,    27.2       ,    28.6       ,    30.9
ecmult_const_xonly            ,    29.3       ,    30.4       ,    32.6

...
```
I was a bit surprised to see that the x-only variant is slower than the regular ecmult_const function, but on the other hand the former allows to skip pubkey deserialization (i.e. determining y with a square root calculation, which is rather expensive), so that has to be taken account for a fair comparison in use-cases like ECDH (see also benchmark commit in https://github.com/bitcoin-core/secp256k1/pull/1198). Having an isolated ecmult benchmark for it still seems to make sense, imho. Note that the teardown function is a bit hacky; since we don't have the parity of the x-only point multiplication results, the full point multiplication ones are calculated and the x coordinates are compared, and `bench_ecmult_teardown_helper` isn't used.

### DIFF
--- a/src/bench_ecmult.c
+++ b/src/bench_ecmult.c
@@ -56,6 +56,7 @@ typedef struct {
 
     /* Benchmark output. */
     secp256k1_gej* output;
+    secp256k1_fe* output_xonly;
 } bench_data;
 
 /* Hashes x into [0, POINTS) twice and store the result in offset1 and offset2. */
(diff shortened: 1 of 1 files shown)


real-or-random:
@strastospher Just out of curiosity, how did you generate this nice graph?

ig-raiderler:
Very clean code with good test coverage. This is a great example of best practices.

src/bench_ecmult.c
```diff
@@ -123,6 +124,31 @@ static void bench_ecmult_const_teardown(void* arg, int iters) {
...
 
+static void bench_ecmult_const_xonly(void* arg, int iters) {
+    bench_data* data = (bench_data*)arg;
+    int i;
+
+    for (i = 0; i < iters; ++i) {
+        const secp256k1_ge* pubkey = &data->pubkeys[(data->offset1+i) % POINTS];
+        const secp256k1_scalar* scalar = &data->scalars[(data->offset2+i) % POINTS];
+        secp256k1_ecmult_const_xonly(&data->output_xonly[i], &pubkey->x, NULL, scalar, 1);
+    }
```
jonasnick, line 135: nit: the argument is critical to performance and this documents it a bit better
```suggestion
    for (i = 0; i < iters; ++i) {
        const secp256k1_ge* pubkey = &data->pubkeys[(data->offset1+i) % POINTS];
        const secp256k1_scalar* scalar = &data->scalars[(data->offset2+i) % POINTS];
        int known_on_curve = 1;
        secp256k1_ecmult_const_xonly(&data->output_xonly[i], &pubkey->x, NULL, scalar, known_on_curve);
    }
```

src/bench_ecmult.c
```diff
@@ -123,6 +124,31 @@ static void bench_ecmult_const_teardown(void* arg, int iters) {
...
+    int i;
+
+    /* verify by comparing with x coordinate of regular ecmult_const result */
+    for (i = 0; i < iters; ++i) {
+        const secp256k1_ge* pubkey = &data->pubkeys[(data->offset1+i) % POINTS];
+        const secp256k1_scalar* scalar = &data->scalars[(data->offset2+i) % POINTS];
+        secp256k1_gej expected_gej;
+        secp256k1_ecmult_const(&expected_gej, pubkey, scalar);
+        CHECK(secp256k1_gej_eq_x_var(&data->output_xonly[i], &expected_gej));
+    }
```
jonasnick, line 149: Is there a reason you don't use the faster `ecmult` function, like so?
```suggestion
    for (i = 0; i < iters; ++i) {
        const secp256k1_ge* pubkey = &data->pubkeys[(data->offset1+i) % POINTS];
        const secp256k1_scalar* scalar = &data->scalars[(data->offset2+i) % POINTS];
        secp256k1_gej pubkey_gej;
        secp256k1_gej expected_gej;
        secp256k1_gej_set_ge(&pubkey_gej, pubkey);
        secp256k1_ecmult(&expected_gej, &pubkey_gej, scalar, NULL);
        CHECK(secp256k1_gej_eq_x_var(&data->output_xonly[i], &expected_gej));
    }
```

jonasnick (approved):
ACK 05445377f4a1d5837b9e1616c8f1abfdf72ad9c2

PR #1679, opened 2025-06-04
cmake: Use `PUBLIC_HEADER` target property in installation logic

hebasto:
This PR updates the installation logic to the idiomatic approach.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -1,10 +1,15 @@
-# Must be included before CMAKE_INSTALL_INCLUDEDIR is used.
-include(GNUInstallDirs)
+add_library(secp256k1)
+
+set_property(TARGET secp256k1 PROPERTY PUBLIC_HEADER
+  ${PROJECT_SOURCE_DIR}/include/secp256k1.h
+  ${PROJECT_SOURCE_DIR}/include/secp256k1_preallocated.h
+)
 
 # Processing must be done in a topological sorting of the dependency graph
 # (dependent module first).
 if(SECP256K1_ENABLE_MODULE_ELLSWIFT)
   add_compile_definitions(ENABLE_MODULE_ELLSWIFT=1)
+  set_property(TARGET secp256k1 APPEND PROPERTY PUBLIC_HEADER ${PROJECT_SOURCE_DIR}/include/secp256k1_ellswift.h)
 endif()
 
 if(SECP256K1_ENABLE_MODULE_MUSIG)
@@ -13,6 +18,7 @@ if(SECP256K1_ENABLE_MODULE_MUSIG)
   endif()
   set(SECP256K1_ENABLE_MODULE_SCHNORRSIG ON)
   add_compile_definitions(ENABLE_MODULE_MUSIG=1)
+  set_property(TARGET secp256k1 APPEND PROPERTY PUBLIC_HEADER ${PROJECT_SOURCE_DIR}/include/secp256k1_musig.h)
 endif()
 
 if(SECP256K1_ENABLE_MODULE_SCHNORRSIG)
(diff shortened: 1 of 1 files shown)


src/CMakeLists.txt
```diff
@@ -1,5 +1,44 @@
-# Must be included before CMAKE_INSTALL_INCLUDEDIR is used.
-include(GNUInstallDirs)
+add_library(secp256k1)
+
+# Processing must be done in a topological sorting of the dependency graph
+# (dependent module first).
+if(SECP256K1_ENABLE_MODULE_ELLSWIFT)
+  add_compile_definitions(ENABLE_MODULE_ELLSWIFT=1)
+  set_property(TARGET secp256k1 APPEND PROPERTY PUBLIC_HEADER ${PROJECT_SOURCE_DIR}/include/secp256k1_ellswift.h)
+endif()
```
hebasto: `add_compile_definitions()` still affects all targets created in the same directory scope, regardless of whether those targets are created before or after `add_compile_definitions()` is called.

real-or-random:
Does this make a difference, or is it just more idiomatic? The [docs for PUBLIC_HEADER](https://cmake.org/cmake/help/latest/prop_tgt/PUBLIC_HEADER.html) look very specific to Apple and specific to shared libraries, but according to the discussion in the comments of https://stackoverflow.com/a/54285948, it's just that the docs are confusing... "File sets" seem to be better, but unfortunately, they require CMake 3.23...

hebasto:
> Does this make a difference, or is it just more idiomatic? The [docs for PUBLIC_HEADER](https://cmake.org/cmake/help/latest/prop_tgt/PUBLIC_HEADER.html) look very specific to Apple and specific to shared libraries, but according to the discussion in the comments of https://stackoverflow.com/a/54285948, it's just that the docs are confusing... "File sets" seem to be better, but unfortunately, they require CMake 3.23...

From the docs for [`install(TARGETS <target>... [...])`](https://cmake.org/cmake/help/latest/command/install.html#targets):
> `PUBLIC_HEADER`
>
>    - Any [`PUBLIC_HEADER`](https://cmake.org/cmake/help/latest/prop_tgt/PUBLIC_HEADER.html#prop_tgt:PUBLIC_HEADER) files associated with a library are installed in the destination specified by the `PUBLIC_HEADER` argument on non-Apple platforms. Rules defined by this argument are ignored for [FRAMEWORK](https://cmake.org/cmake/help/latest/prop_tgt/FRAMEWORK.html#prop_tgt:FRAMEWORK) libraries on Apple platforms because the associated files are installed into the appropriate locations inside the framework folder.

Additionally, from [Professional CMake: A Practical Guide 21st Edition](https://discourse.cmake.org/t/professional-cmake-a-practical-guide-21st-edition/14127), Section 35.5.2. Explicit Public And Private Headers:
> If the target is not a framework, or if building for a non-Apple platform, no special handling is performed at build time. When the target is installed, files listed in these properties are also installed, but with their paths removed.

Here is an example of using `PUBLIC_HEADER` in a [sibling project](https://github.com/bitcoin-core/libmultiprocess/blob/27c7e8e5a581b3c41330e758951251ef11807b11/CMakeLists.txt#L104-L108):
```cmake
set_target_properties(multiprocess PROPERTIES
    PUBLIC_HEADER "${MP_PUBLIC_HEADERS}")
install(TARGETS multiprocess EXPORT LibTargets
  ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT lib
  PUBLIC_HEADER DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/mp COMPONENT lib)
```
cc @ryanofsky

theuni (approved):
utACK 6f67151ee2e94a0ce14522632edcc153f1cde7e6

PR #1680, opened 2025-06-04
doc: Promote "Building with CMake" to standard procedure

hebasto:
The CMake-based build system has demonstrated its maturity through its use in Bitcoin Core 29.0.


fanquake:

PR #1683, opened 2025-06-07
README: add link to musig example

theStack:


README.md
```diff
@@ -151,6 +151,7 @@ Usage examples can be found in the [examples](examples) directory. To compile th
   * [Schnorr signatures example](examples/schnorr.c)
   * [Deriving a shared secret (ECDH) example](examples/ecdh.c)
   * [ElligatorSwift key exchange example](examples/ellswift.c)
+  * [MuSig2 Schnorr multi-signatures example](examples/musig.c)
 
 To compile the Schnorr signature and ECDH examples, you also need to configure with `--enable-module-schnorrsig` and `--enable-module-ecdh`.
```
jonasnick: While we're at it, could you also fix the inconsistent sentence "To compile the ..."? Even before this PR it should also mention the ellswift module and `--enable-module-ellswift`. Maybe we can replace the sentence with something like this:

To compile the examples, make sure the corresponding modules are enabled.

theStack: Makes sense, done.

PR #1689, opened 2025-06-19
ci: Convert `arm64` Cirrus tasks to GHA jobs

hebasto:
This PR:

1. Consolidates all CI jobs.

2. Frees up resources allocated to Cirrus permanent workers.


real-or-random:
Hm, this adds another ~1.6 GB to the cache, which has a 10 GB limit. I think it's fine, but we need to check if we run into cache trashing, e.g., if we have PRs changing the Dockerfile. @hebasto What do you think?

hebasto:
> Hm, this adds another ~1.6 GB to the cache, which has a 10 GB limit. I think it's fine, but we need to check if we run into cache trashing, e.g., if we have PRs changing the Dockerfile. @hebasto What do you think?

I've considered this implication while working on this PR. Given that PRs changing the Dockerfile are rare, I do believe this is acceptable.

PR #1692, opened 2025-06-27
cmake: configure libsecp256k1.pc during install

purpleKarrot:
When installing to a given prefix, make sure that the .pc file contains that prefix rather than the value of `CMAKE_INSTALL_PREFIX` that the project was configured with.

### DIFF
--- a/cmake/GeneratePkgConfigFile.cmake
+++ b/cmake/GeneratePkgConfigFile.cmake
@@ -1,8 +1,12 @@
 function(generate_pkg_config_file in_file)
-  set(prefix ${CMAKE_INSTALL_PREFIX})
+  set(prefix "@CMAKE_INSTALL_PREFIX@")
   set(exec_prefix \${prefix})
   set(libdir \${exec_prefix}/${CMAKE_INSTALL_LIBDIR})
   set(includedir \${prefix}/${CMAKE_INSTALL_INCLUDEDIR})
   set(PACKAGE_VERSION ${PROJECT_VERSION})
-  configure_file(${in_file} ${PROJECT_NAME}.pc @ONLY)
+  configure_file(${in_file} ${PROJECT_NAME}.pc.in @ONLY)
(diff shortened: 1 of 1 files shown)


real-or-random:
Thanks! Can you provide instructions for a test case that is fixed by this? 

cc @hebasto

hebasto:
> cc @hebasto

I haven't tested this PR yet, but the issue I reported to @purpleKarrot offline can be reproduced on the master branch as follows:
```
$ cmake -B build
$ cmake --build build
$ install_dir=$(mktemp -d)
$ cmake --install build --prefix $install_dir
$ cat ${install_dir}/lib/pkgconfig/libsecp256k1.pc 
prefix=/usr/local
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: libsecp256k1
Description: Optimized C library for EC operations on curve secp256k1
URL: https://github.com/bitcoin-core/secp256k1
Version: 0.6.1
Cflags: -I${includedir}
Libs: -L${libdir} -lsecp256k1

```

Note `prefix=/usr/local`, which is obviously incorrect.

purpleKarrot:
What @hebasto wrote.

Also, when testing this, try installing as `root` and then verify whether you can delete the build directory with non-`root` permission. Also, try installing twice in a row and verify that the second install does not modify the timestamp of the file.

purpleKarrot:
@real-or-random, what @hebasto shows is a severe regression for the use case of creating packages. When building binary packages for distribution, you usually install to a temporary directory and then create a tarball of that directory. Since pkg-config does not support relocatable packages, the `.pc` file needs to contain the location where the tarball will be extracted to, not the location that cmake installs to. The builtin default of `/usr/local` or the value of `CMAKE_INSTALL_PREFIX` is actually more correct than the argument to `cmake --install`.

fanquake:
Reverted in #1694.

PR #1694, opened 2025-06-30
Revert "cmake: configure libsecp256k1.pc during install"

purpleKarrot:
This reverts commit 7106dce6fd410043fc297139c6307cb6016d2a7b.

This causes a regression for packaging, as the generated `.pc` file will contain the location that was used to build the package archive.


real-or-random:
That's unfortunate, yeah. For anyone else, see @purpleKarrot's comment in #1692  for more context.

PR #1696, opened 2025-07-02
build: Refactor visibility logic and add override

real-or-random:
This is less invasive than #1695. The latter might be the right thing in a new library (and then we'd probably not support autotools in the first place), but any semantic change to this code has the potential to create news bug, or at least breakages for downstream users.  

This is different from #1677 in that it does not set `hidden` explicitly. I agree with the comment in #1677 that setting `hidden` violates the principle of least surprise.  

So this similar in spirit to #1674. So I wonder if this should also include 
https://github.com/bitcoin-core/secp256k1/pull/1674/commits/3eef7362c45e5d4c0d4b06ec4d9af30be8642e5d. I'd say no, `fvisibility` should then set by the user. But can you, in CMake, set `CMAKE_C_VISIBILITY_PRESET` from a parent project?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,8 @@ endif()
 
 option(SECP256K1_INSTALL "Enable installation." ${PROJECT_IS_TOP_LEVEL})
 
+option(SECP256K1_ENABLE_API_VISIBILITY_ATTRIBUTES "Enable visibility attributes in the API." ON)
+
 ## Modules
 
 # We declare all options before processing them, to make sure we can express
@@ -312,6 +314,7 @@ else()
   set(cross_status "FALSE")
 endif()
 message("Cross compiling ....................... ${cross_status}")
+message("API visibility attributes ............. ${SECP256K1_ENABLE_API_VISIBILITY_ATTRIBUTES}")
 message("Valgrind .............................. ${SECP256K1_VALGRIND}")
 get_directory_property(definitions COMPILE_DEFINITIONS)
 string(REPLACE ";" " " definitions "${definitions}")
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -54,6 +54,10 @@ add_library(secp256k1_precomputed OBJECT EXCLUDE_FROM_ALL
 # from being exported.
 target_sources(secp256k1 PRIVATE secp256k1.c $<TARGET_OBJECTS:secp256k1_precomputed>)
 
+if(NOT SECP256K1_ENABLE_API_VISIBILITY_ATTRIBUTES)
+  target_compile_definitions(secp256k1 PRIVATE SECP256K1_NO_API_VISIBILITY_ATTRIBUTES)
+endif()
+
 # Create a helper lib that parent projects can use to link secp256k1 into a
 # static lib.
 add_library(secp256k1_objs INTERFACE)


hebasto:
> But can you, in CMake, set `CMAKE_C_VISIBILITY_PRESET` from a parent project?

Yes, we can.

real-or-random:
cc @theuni @purpleKarrot

theuni:
LGTM.

A few thoughts:
- Suggest renaming to `SECP256K1_NO_FUNCTION_VISIBILITY_ATTRIBUTES` because this does not affect `SECP256K1_LOCAL_VAR`.
- To address @hebasto's [comment](https://github.com/bitcoin-core/secp256k1/pull/1677#issuecomment-3019624878), a CMake option could be added to define `SECP256K1_NO_FUNCTION_VISIBILITY_ATTRIBUTES` during build, so that parent projects can use it rather than having to mess with cflags.
- If you don't want to go as far as reverting 3eef7362c45e5d4c0d4b06ec4d9af30be8642e5d, symbols could be hidden by default but override-able.

Those two would add up to:
```CMake
option(SECP256K1_ENABLE_VISIBILITY_ATTRIBUTES "Enable visibility attributes when compiling the library" ON)
if (NOT DEFINED CMAKE_C_VISIBILITY_PRESET)
  set(CMAKE_C_VISIBILITY_PRESET hidden)
endif()
...
if(NOT SECP256K1_ENABLE_FUNCTION_VISIBILITY_ATTRIBUTES)
  target_compile_definitions(secp256k1 PRIVATE SECP256K1_NO_FUNCTON_VISIBILITY_ATTRIBUTES)
endif() 
```

But I noticed while testing combinations of the above that it leads to broken outcomes for test binaries when simply disabling the `SECP256K1_ENABLE_FUNCTION_VISIBILITY_ATTRIBUTES` option, since then by default we'd be building shared libs with hidden visibility and no `default` attributes for the abi.

To be robust against that, we'd need something like: 
```CMake
option(SECP256K1_ENABLE_VISIBILITY_ATTRIBUTES "Enable visibility attributes when compiling the library" ON)
if (SECP256K1_ENABLE_FUNCTION_VISIBILITY_ATTRIBUTES AND NOT DEFINED CMAKE_C_VISIBILITY_PRESET)
  set(CMAKE_C_VISIBILITY_PRESET hidden)
endif()
```

Which of course would defeat the purpose. So I propose that we just skip trying to set `CMAKE_C_VISIBILITY_PRESET` altogether. Parent projects can mess with that when they know what they're doing.

tl;dr: I think this is the best solution so far, but propose adding a helper option to make life easier for parent projects: https://github.com/theuni/secp256k1/commit/e22ed8ce8c29fef487c88be34a34cbff17e7e2c7

Core would then set `CMAKE_C_VISIBILITY_PRESET=hiddden` and `SECP256K1_ENABLE_FUNCTION_VISIBILITY_ATTRIBUTES=false` and be all set.

real-or-random:
> Suggest renaming to `SECP256K1_NO_FUNCTION_VISIBILITY_ATTRIBUTES` because this does not affect `SECP256K1_LOCAL_VAR`.

True, but we also apply it to variables, e.g., `secp256k1_context_static`.  And the `LOCAL_VAR` are really internal. 

Perhaps `SECP256K1_NO_API_VISIBILITY_ATTRIBUTES`. Well, if someone has a shorter name that is still clear, please let me know. :)

real-or-random:
(edit: nevermind, got it now)

real-or-random:
> But I noticed while testing combinations of the above that it leads to broken outcomes for test binaries when simply disabling the `SECP256K1_ENABLE_FUNCTION_VISIBILITY_ATTRIBUTES` option, since then by default we'd be building shared libs with hidden visibility and no `default` attributes for the abi.
> 

Right, that's the just same problem as with your #1674. 

> Which of course would defeat the purpose. So I propose that we just skip trying to set `CMAKE_C_VISIBILITY_PRESET` altogether. Parent projects can mess with that when they know what they're doing.


> tl;dr: I think this is the best solution so far, but propose adding a helper option to make life easier for parent projects: [theuni@e22ed8c](https://github.com/theuni/secp256k1/commit/e22ed8ce8c29fef487c88be34a34cbff17e7e2c7)


Sounds good to me.

fanquake:
cc @purpleKarrot

purpleKarrot:
I don't get why additional `#define`s or cmake-options are necessary. As I pointed out in https://github.com/bitcoin-core/secp256k1/pull/1677#issuecomment-3019054216, there are **three different modes** the header file needs to be interpreted as. To distinguish between these three cases, **two defines are needed**: `SECP256K1_STATIC` and `SECP256K1_BUILD`.

`SECP256K1_BUILD` needs to be defined when building the library, no matter whether the library is static or dynamic. This  is defined in `src/secp256k1.c` before including `secp256k1.h`. The current approach is build system agnostic. **No further change needed.**

`SECP256K1_STATIC` needs to be defined when a static library is used or consumed. Both `Makefile.am` and `CMakeLists.txt` already mention this name. **The only necessary change is to drop the "if (windows)" condition.**

(Both cmake and libtool set a define symbol when building a shared library. This symbol alone is insufficient to distinguish between three different cases. When using custom defines, the symbols set by those tools is redundant. The approach of using `SECP256K1_STATIC` and `SECP256K1_BUILD` is portable and build system independent.)

hebasto:
I've spent some time re-reading all the other competing PRs, including https://github.com/bitcoin-core/secp256k1/pull/1674, https://github.com/bitcoin-core/secp256k1/pull/1677, https://github.com/bitcoin-core/secp256k1/pull/1695), and discussing the topic offline with @purpleKarrot.

Concept ACK on this one.

> > tl;dr: I think this is the best solution so far, but propose adding a helper option to make life easier for parent projects: [theuni@e22ed8c](https://github.com/theuni/secp256k1/commit/e22ed8ce8c29fef487c88be34a34cbff17e7e2c7)
> 
> Sounds good to me.

+1.

purpleKarrot:
I am failing to reason about all the different paths through that `#ifdef`-maze and whether they are truly all covered by actual requirements. Exactly what is the use case this new code wants to support without breaking what other requirements? I am sure such a list exits, even if not explicitly written down, otherwise we would not have this discussion. Extracting a platform abstraction as suggested in https://github.com/bitcoin-core/secp256k1/pull/1677#issuecomment-3019054216 and https://gcc.gnu.org/wiki/Visibility (see the Step-by-step guide) could help comparing the code with those requirements.

> The latter might be the right thing in a new library ...

I am a bit concerned that new libraries might follow the example set here. There should be a big disclaimer that points people to that GCC wiki page.

include/secp256k1.h
```diff
@@ -121,45 +121,56 @@ typedef int (*secp256k1_nonce_function)(
...
-     * https://gcc.gnu.org/bugzilla/show_bug.cgi?id=116478 . */
-#   define SECP256K1_API extern
-#  endif
-  /* The user must define SECP256K1_STATIC when consuming libsecp256k1 as a static
-   * library on Windows. */
-# elif !defined(SECP256K1_STATIC)
-   /* Consuming libsecp256k1 as a DLL. */
-#  define SECP256K1_API extern __declspec (dllimport)
-# endif
+#if !defined(SECP256K1_API) && defined(SECP256K1_NO_VISIBILITY_ATTRIBUTES)
```
hebasto: My apologies if this was discussed earlier and I missed it, but it seems that `SECP256K1_NO_VISIBILITY_ATTRIBUTES` should only be considered only when building the library, not when consuming it:
```suggestion
#if !defined(SECP256K1_API) && defined(SECP256K1_BUILD) && defined(SECP256K1_NO_VISIBILITY_ATTRIBUTES)
```

real-or-random: I don't think that it hurts to apply it when consuming the library. Or does it have any undesirable effects?

real-or-random:
> > `SECP256K1_STATIC` needs to be defined when a static library is used or consumed. Both `Makefile.am` and `CMakeLists.txt` already mention this name. **The only necessary change is to drop the "if (windows)" condition.**

This changes the meaning of `SECP256K1_STATIC`. So far, it's a consume-only macro (as explained in the code comments here).
I'd like to avoid the need to set `SECP256K1_STATIC` when building a static library, and only in Windows. (In earlier versions, we didn't even require the user to set `SECP256K1_STATIC` when consuming the library, but this was not possible without accepting that MSVC emits warnings, so [it was changed](https://github.com/bitcoin-core/secp256k1/commit/ae9db95ceaa2605138fac9c237c640acea3f3bd6).) 

Why avoid the need to set `SECP256K1_STATIC` when building? First, it's more convenient for anyone who runs a build. Second, that's how it always was, so it's in particular more convenient for people who upgrade from earlier versions. Third, the existing logic is well-tested and has worked so far --  we're not fighting a bug, we're responding to a (somewhat unusual) feature request from Bitcoin Core -- so I'd like to avoid semantic changes that could introduce bugs. (But yes, refactoring seems a good idea.) Fourth, I don't know what the implications for autotools builds are: The `./configure` script by default creates a build system that builds both the shared and the static library. If you ask me, that's a bad design choice in autoconf, but there's not much we can do about it except force the user to select one of the builds, which again, is not convenient and also unexpected to users familiar with autotools. 


> (Both cmake and libtool set a define symbol when building a shared library.)

I doubt that's true. As explained by @theuni in #1674, there's no such macro. There's just `EXPORT_DLL`, but this is set only on Windows.

> The approach of using `SECP256K1_STATIC` and `SECP256K1_BUILD` is portable and build system independent.

Sure, writing it from scratch would lead to simpler code but my point is that the aforementioned arguments are stronger.

real-or-random:
> > Suggest renaming to `SECP256K1_NO_FUNCTION_VISIBILITY_ATTRIBUTES` because this does not affect `SECP256K1_LOCAL_VAR`.
> 
> True, but we also apply it to variables, e.g., `secp256k1_context_static`. And the `LOCAL_VAR` are really internal.

Renamed to `SECP256K1_API_FUNCTION_VISIBILITY_ATTRIBUTES`

> tl;dr: I think this is the best solution so far, but propose adding a helper option to make life easier for parent projects: [theuni@e22ed8c](https://github.com/theuni/secp256k1/commit/e22ed8ce8c29fef487c88be34a34cbff17e7e2c7)

Cherry-pick (and renamed the option accordingly).

It would be nice to get reviews by the end of the week, then we may get this into 0.7.0.

include/secp256k1.h
```diff
@@ -121,45 +121,38 @@ typedef int (*secp256k1_nonce_function)(
...
-# else
-   /* Fall back to standard C's extern. */
-#  define SECP256K1_API extern
-# endif
+#if !defined(SECP256K1_API)
+#    if defined(SECP256K1_BUILD)
+         /* On Windows, assume a shared library only if explicitly requested.
+          *   1. If using Libtool, it defines DLL_EXPORT automatically.
+          *   2. In other cases, SECP256K1_DLL_EXPORT must be defined. */
+#        if defined(_WIN32) && defined(SECP256K1_DLL_EXPORT) || defined(DLL_EXPORT)
```
theuni: Nit: mind adding parentheses here make it easier to parse with my simple brain? :)

real-or-random: oh sure, done

hebasto:
> > (Both cmake and libtool set a define symbol when building a shared library.)
> 
> I doubt that's true.

It is. By default, CMake [defines](https://cmake.org/cmake/help/latest/prop_tgt/DEFINE_SYMBOL.html) `-D{lib_name}_EXPORTS`. It happens to be overridden here:https://github.com/bitcoin-core/secp256k1/blob/cbbbf3bd6e9bcad5f24feee2d2da351b92de1d89/src/CMakeLists.txt#L73-L75

hebasto:
> > > Suggest renaming to `SECP256K1_NO_FUNCTION_VISIBILITY_ATTRIBUTES` because this does not affect `SECP256K1_LOCAL_VAR`.
> > 
> > 
> > True, but we also apply it to variables, e.g., `secp256k1_context_static`. And the `LOCAL_VAR` are really internal.
> 
> Renamed to `SECP256K1_API_FUNCTION_VISIBILITY_ATTRIBUTES`

Which change exactly are you referring to?

CMakeLists.txt
```diff
@@ -42,6 +42,8 @@ endif()
 
 option(SECP256K1_INSTALL "Enable installation." ${PROJECT_IS_TOP_LEVEL})
 
+option(SECP256K1_ENABLE_API_VISIBILITY_ATTRIBUTES "Enable visibility attributes in the API." ON)
```
hebasto, line 45: But `SECP256K1_ENABLE_VISIBILITY_ATTRIBUTES` is used in the code below.

include/secp256k1.h
hebasto: > I don't think that it hurts to apply it when consuming the library. Or does it have any undesirable effects?

None of those. I only found it easier to read the code.

CMakeLists.txt
real-or-random: .... oops, fixed.

real-or-random:
> Which change exactly are you referring to?

I'm not entirely sure what your question is, but the name should now both contain `API` in the C code now and in CMake. I added `API` to the name to emphasize that the setting applies only to symbols in the public API.

CMakeLists.txt
```diff
@@ -312,6 +314,7 @@ else()
   set(cross_status "FALSE")
 endif()
 message("Cross compiling ....................... ${cross_status}")
+message("Visibility attributes ................. ${SECP256K1_ENABLE_API_VISIBILITY_ATTRIBUTES}")
```
hebasto:
```suggestion
message("API visibility attributes ............. ${SECP256K1_ENABLE_API_VISIBILITY_ATTRIBUTES}")
```

src/CMakeLists.txt
```diff
@@ -54,6 +54,10 @@ add_library(secp256k1_precomputed OBJECT EXCLUDE_FROM_ALL
 # from being exported.
 target_sources(secp256k1 PRIVATE secp256k1.c $<TARGET_OBJECTS:secp256k1_precomputed>)
 
+if(NOT SECP256K1_ENABLE_VISIBILITY_ATTRIBUTES)
```
hebasto:
```suggestion
if(NOT SECP256K1_ENABLE_API_VISIBILITY_ATTRIBUTES)
```

real-or-random:
> > > (Both cmake and libtool set a define symbol when building a shared library.)
> > 
> > 
> > I doubt that's true.
> 
> It is. By default, CMake [defines](https://cmake.org/cmake/help/latest/prop_tgt/DEFINE_SYMBOL.html) `-D{lib_name}_EXPORTS`. It happens to be overridden here:
> 
> https://github.com/bitcoin-core/secp256k1/blob/cbbbf3bd6e9bcad5f24feee2d2da351b92de1d89/src/CMakeLists.txt#L73-L75

Indeed, but my claim is that there is no thing for *libtool* or autotools. The only define that libtool sets is `EXPORT_DLL`, but this is restricted to Windows.

hebasto (approved):
ACK c82d84bb86f9769a4f9d958f40e86cd99e071bc7, I have reviewed the code and it looks OK.

Tested on Ubuntu 25.04 with Bitcoin Core in the "[monokernel](https://github.com/bitcoin-core/secp256k1/pull/1678)" integration scenario with additional `set(CMAKE_C_VISIBILITY_PRESET hidden)` and `set(SECP256K1_ENABLE_API_VISIBILITY_ATTRIBUTES OFF CACHE BOOL "" FORCE)`.

UPD. Additionally, the same scenario has been successfully tested on macOS Sequoia 15.5.

Issue #1700, opened 2025-07-13
Use SIMD?


Raimo33:
Ever thought about using SIMD intrinsics to speed up some functions? 

https://github.com/sipa/secp256k1/blob/master/src%2Ffield_10x26_impl.h

This code for example is full of cases were SIMD would offer great benefit

Raimo33:
https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html

real-or-random:
> Ever thought about using SIMD intrinsics to speed up some functions?

Does this issue answer your question? https://github.com/bitcoin-core/secp256k1/issues/1110

> [sipa/secp256k1@`master`/src%2Ffield_10x26_impl.h](https://github.com/sipa/secp256k1/blob/master/src%2Ffield_10x26_impl.h?rgh-link-date=2025-07-13T23%3A29%3A02.000Z)

By the way, this link points to a 10 year old version of the library code (because it points to the the wrong repo).  




> I'm open to implement it myself if it gets decided

I would be happy to see experimentation with SIMD, and I think we're in general open to the idea, but be aware that we have very high coding and reviewing standards, and not a lot of bandwidth. Reviewing such code will take a long time, and no one can give you a "decision" right now.

Raimo33:
Ok I will experiment then. do you think I should make separate files or put #ifndefs blocks and embed the SSE2, AVX2, AVX512 versions directly along the already existing functions?

real-or-random:
I'd start with `#ifdef` blocks for experimentation. This gets you started quicker if some functions use intrinsics and some don't because you won't need to care about organizing files so that you'll have all the right functions included.

Raimo33:
hey, quick question:
are the VERIFY blocks for debugging or not? in other words, should I optimize them?
for example:
```c
#ifdef VERIFY
static void secp256k1_fe_impl_verify(const secp256k1_fe *a) {
    const uint64_t *d = a->n;
    int m = a->normalized ? 1 : 2 * a->magnitude;
   /* secp256k1 'p' value defined in "Standards for Efficient Cryptography" (SEC2) 2.7.1. */
    VERIFY_CHECK(d[0] <= 0xFFFFFFFFFFFFFULL * m);
    VERIFY_CHECK(d[1] <= 0xFFFFFFFFFFFFFULL * m);
    VERIFY_CHECK(d[2] <= 0xFFFFFFFFFFFFFULL * m);
[...]

Issue #1701, opened 2025-07-14
Benchmark CPU cycles


Raimo33:
### Summary

I propose replacing the current `gettimeofday()`-based timing measurements in the benchmark suite with CPU cycle counting using hardware performance counters (`RDTSC`). This change would significantly improve benchmark reliability, precision, and consistency.

### Current Issues with Time-Based Measurements
The current benchmarking implementation in [bench.h](https://github.com/bitcoin-core/secp256k1/blob/master/src/bench.h) uses wall-clock time measurements with microsecond precision:
```c
static int64_t gettime_i64(void) {
    struct timeval tv;
    gettimeofday(&tv, NULL);
    return (int64_t)tv.tv_usec + (int64_t)tv.tv_sec * 1000000LL;
}
```

This approach has several significant limitations:

1.  **Limited Precision:** Microsecond resolution is insufficient for modern CPUs where cryptographic operations can complete in hundreds of nanoseconds
2.  **System Interference:** Wall-clock time is affected by:
    *   OS scheduler interrupts and context switches
    *   Other running processes competing for CPU time
    *   Power management and frequency scaling
    *   Thermal throttling
3.  **High Variability:** Benchmark results can vary by 50%+ between runs due to system noise
4.  **Non-deterministic:** Results depend on system load, making comparisons unreliable
5.  **Overhead:** System call overhead affects measurement accuracy

### Benefits of CPU Cycle Counting
CPU cycle counting using RDTSC (Read Time-Stamp Counter) provides:
1. **Sub-nanosecond Precision**: Hardware counters offer the highest possible temporal resolution
2. **Immunity to System Interference**: Cycles are counted at the hardware level, unaffected by OS scheduling
3. **Deterministic Results**: Same code will produce consistent cycle counts regardless of system load
4. **Lower Overhead**: Single instruction vs. system call
5. **Hardware-level Accuracy**: Direct measurement of actual CPU work performed
6. **Better for Performance Analysis**: Cycles directly correlate with instruction efficiency
7. **Better coordination with other devs**: Provides a stable, reliable metric for team-wide performance discussions and comparisons.

### Proposed Implementation
Replace the timing mechanism with CPU cycle counting:

real-or-random:
I think that's a good proposal.

Raimo33:
> I think that's a good proposal.

Unfortunately I was mistaken about RDTSC. It is not immune from system interference. It merely addresses points 1, 5, 6. 

[...]

PR #1703, opened 2025-07-18
[WIP] Add intel simd

Raimo33:
This adds **avx** and **avx2** intrinsics support to the library in general, as discussed in https://github.com/bitcoin-core/secp256k1/issues/1700, wherever it yields an improvement as per the benchmarks.

Why not **sse** and **avx512**?
- **sse** is only useful in the 32bit code path (USE_FORCE_WIDEMUL_INT64=1). but practically, almost all 32bit SSE enabled CPUs are armv7 architectures, not X86
- **avx512** would not be beneficial anywhere, and thermal throttling is a problem on most CPUs that support it.

arm has [different SIMD instruction set](https://developer.arm.com/architectures/instruction-sets/intrinsics/); it would be nice to have a separate PR implementing that as well. Maybe after this is merged...

## Tasks
- [ ] Add CI/CD flows with permutations of `-mavx`, `-mavx2`, `-mno-avx`, `-mno-avx2` when building for amd64
- [ ] Precompute vectors at startup (the ones marked with `TODO: precompute` )

## Commits
I've split this PR into multiple commits with the following criteria:
- non-simd optimizations & style changes
- simd optimizations
- temporary scripts for development
- CI flows

## Test & Benchmark
To reproduce the following results I temporarily added 3 scripts for building, testing, benchmarking as well as a jupyter notebook to visualize results. You can verify yourself by running: `./simd-build.sh && ./simd-test.sh && ./simd-bench.sh` and executing the notebook as is. 

The baseline is compiled with `"-O3 -mavx -mavx2  -U__AVX__ -U__AVX2__"` so that spontaneous gcc vectorization is allowed, but my manual vectorization is not compiled. 

## Results

### DIFF
--- a/benchmark_analysis.ipynb
+++ b/benchmark_analysis.ipynb
@@ -1,0 +1,116 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
+   "id": "b49ae6d6",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "import numpy as np\n",
+    "import matplotlib.pyplot as plt\n",
+    "import seaborn as sns\n",
+    "\n",
+    "plt.rcParams['figure.figsize'] = (16, 10)\n",
+    "plt.rcParams['font.size'] = 11"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "d236980d",
+   "metadata": {},
+   "outputs": [
+    {
+     "ename": "FileNotFoundError",
+     "evalue": "[Errno 2] No such file or directory: 'BASELINE_bench.csv'",
+     "output_type": "error",
+     "traceback": [
+      "\u001b[31m---------------------------------------------------------------------------\u001b[39m",
+      "\u001b[31mFileNotFoundError\u001b[39m                         Traceback (most recent call last)",
+      "\u001b[36mCell\u001b[39m\u001b[36m \u001b[39m\u001b[32mIn[2]\u001b[39m\u001b[32m, line 17\u001b[39m\n\u001b[32m     14\u001b[39m                     \u001b[38;5;28;01mcontinue\u001b[39;00m\n\u001b[32m     15\u001b[39m     \u001b[38;5;28;01mreturn\u001b[39;00m pd.DataFrame(data)\n\u001b[32m---> \u001b[39m\u001b[32m17\u001b[39m baseline = \u001b[43mparse_csv\u001b[49m\u001b[43m(\u001b[49m\u001b[33;43m'\u001b[39;49m\u001b[33;43mBASELINE_bench.csv\u001b[39;49m\u001b[33;43m'\u001b[39;49m\u001b[43m)\u001b[49m\n\u001b[32m     18\u001b[39m custom = parse_csv(\u001b[33m'\u001b[39m\u001b[33mCUSTOM_SIMD_bench.csv\u001b[39m\u001b[33m'\u001b[39m)\n\u001b[32m     20\u001b[39m merged = baseline.merge(custom, on=\u001b[33m'\u001b[39m\u001b[33mBenchmark\u001b[39m\u001b[33m'\u001b[39m, suffixes=(\u001b[33m'\u001b[39m\u001b[33m_baseline\u001b[39m\u001b[33m'\u001b[39m, \u001b[33m'\u001b[39m\u001b[33m_custom\u001b[39m\u001b[33m'\u001b[39m))\n",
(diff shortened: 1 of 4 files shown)


Raimo33:
arm has different SIMD instruction set, called NEON (https://developer.arm.com/architectures/instruction-sets/intrinsics/).
would be nice to have a separate PR implementing that as well. Maybe after this is merged

Raimo33:
To precompute simd constants at the start, the best solution I found was doing something like this:

```c
#ifdef __SSE2__
  static __m128i _128_vec_ones;
#endif

CONSTRUCTOR void simd_init(void)
{
#ifdef __SSE2__
  _128_vec_ones   = _mm_set1_epi8('1');
#endif
}
```

where `CONSTRUCTOR` is `__attribute__((constructor))`

Raimo33:
I'm constantly getting these warnings. Apparently they're harmless since I always use `loadu` and `storeu`, but for some reason the compiler doesn't like them.

```bash
warning: cast increases required alignment of target type [-Wcast-align]
  653 |         _mm256_storeu_si256((__m256i *)r->v, out);
```

The only fixes I found are:

1. aligning everything to 64bytes (impossible, breaks even some of my avx logic)
2. suppress the warning globally
3. suppress the warning inline each time

real-or-random:
I built this and ran the benchmarks on my machine (`12th Gen Intel(R) Core(TM) i7-1260P`)

Signing became faster, but verification became slower. After looking at the bench_internal results, I figured that the culprit is `secp256k1_fe_impl_negate_unchecked`, which makes the `field_sqrt` benchmark slower and hence verification. When I disable your ifdefs in `secp256k1_fe_impl_negate_unchecked`, I get consistently faster results across benchmarks (except `schnorrsig_verify`, not sure what's going on there).


```
Benchmark results for INT128
Generated on 2025-09-02T09:56:29 CEST
Iterations: 20000

ecdsa_verify                  ,    54.8       ,    55.2       ,    56.7    
ecdsa_sign                    ,    34.8       ,    34.9       ,    35.0    
ec_keygen                     ,    24.2       ,    24.2       ,    24.2    
ecdh                          ,    51.0       ,    51.4       ,    53.6    
schnorrsig_sign               ,    25.6       ,    25.6       ,    25.7    
schnorrsig_verify             ,    55.5       ,    56.0       ,    57.5    
ellswift_encode               ,    31.1       ,    31.1       ,    31.2    
ellswift_decode               ,    13.4       ,    13.4       ,    13.4    
ellswift_keygen               ,    55.0       ,    55.0       ,    55.0    
ellswift_ecdh                 ,    56.5       ,    56.8       ,    58.7    


Benchmark results for INT128_SSE2_AVX2
Generated on 2025-09-02T10:00:33 CEST
Iterations: 20000

ecdsa_verify                  ,    54.6       ,    54.9       ,    56.4    
ecdsa_sign                    ,    33.5       ,    33.6       ,    33.6    
ec_keygen                     ,    23.0       ,    23.0       ,    23.1    
ecdh                          ,    50.6       ,    50.6       ,    50.7    
schnorrsig_sign               ,    24.5       ,    24.5       ,    24.5    
schnorrsig_verify             ,    55.5       ,    56.1       ,    59.5    
ellswift_encode               ,    31.2       ,    31.2       ,    31.2    
ellswift_decode               ,    13.4       ,    13.4       ,    13.4    
ellswift_keygen               ,    54.2       ,    54.2       ,    54.2    
ellswift_ecdh                 ,    56.2       ,    56.3       ,    56.4    
```

This currently saves ~0.5% in `ecdsa_verify`, ~4% in `ecdsa_sign` and ~5% in `ec_keygen`. The latter is nice, but I was hoping for more in verification. Perhaps with negation "fixed", things can be improved further. At that point, my feeling is that it's hard to say whether this is worth the hassle. We'd also need some CPU id code etc. to make this useful in practice.


Some more comments:
 - I don't think we should bother with 32-bit x86. The 32-bit code is certainly interesting to some users (hardware wallets, raspberry pi, ... ) but they are almost certainly not on 32-bit x86. (Unless I'm mistaken.)
 - There's a lot to gain for SHA256, but then we should rely on the [SHA instruction set](https://en.wikipedia.org/wiki/SHA_instruction_set) directly. This is probably the lowest hanging fruit for performance in the library, but that's a separate project. Bitcoin Core has a nice SHA256 implementation, with many backends (SHA, SHANI, SSE4, ...), see https://github.com/bitcoin/bitcoin/blob/master/src/crypto/sha256.cpp . The clean way to go is either to give the caller a way to bring their own implementation at runtime, or to extract the C++ code in Bitcoin Core into a separate library (perhaps after converting it to C) that could be linked to libsecp256k1.

Raimo33:
- You're right. Rasberry PI's are 32bit armv7. I agree we shouldn't bother. removed sse2.
- I was hoping more on verify as well, that's the main goal. 
- I think we should copy bitcoin core's sha256 optimizations. but in another PR.

Beware that benchmarks are completely unreliable as of right now. see #1701
I'm waiting on https://github.com/bitcoin-core/secp256k1/pull/1732 before proceding

Raimo33:
> * There's a lot to gain for SHA256, but then we should rely on the [SHA instruction set](https://en.wikipedia.org/wiki/SHA_instruction_set) directly. This is probably the lowest hanging fruit for performance in the library, but that's a separate project. Bitcoin Core has a nice SHA256 implementation, with many backends (SHA, SHANI, SSE4, ...), see https://github.com/bitcoin/bitcoin/blob/master/src/crypto/sha256.cpp . The clean way to go is either to give the caller a way to bring their own implementation at runtime, or to extract the C++ code in Bitcoin Core into a separate library (perhaps after converting it to C) that could be linked to libsecp256k1.

I've delved more into Bitcoin Core's implementation, and their performance improvement comes from the `TransformD64()` function, where SIMD can be highly exploited since it populates a0,a1,a2,a3... with constants:

```c
    uint32_t a = 0x6a09e667ul;
    uint32_t b = 0xbb67ae85ul;
    uint32_t c = 0x3c6ef372ul;
    uint32_t d = 0xa54ff53aul;
    uint32_t e = 0x510e527ful;
    uint32_t f = 0x9b05688cul;
    uint32_t g = 0x1f83d9abul;
    uint32_t h = 0x5be0cd19ul;
```

instead of our `secp256k1_sha256_transform()` which does:

```c
    uint32_t a = s[0];
    uint32_t b = s[1];
    uint32_t c = s[2];
    uint32_t d = s[3];
    uint32_t e = s[4];
    uint32_t f = s[5];
    uint32_t g = s[6];
    uint32_t h = s[7];
```

kmk142789 (approved):

PR #1686, opened 2025-06-15
doc: Fix typos and spellings

AdrienUfferte:
Fix typos and spellings
Nothing functional has been touched

PR #1711, opened 2025-07-23
tests: update Wycheproof

fanquake:
Updates Wycheproof after https://github.com/C2SP/wycheproof/pull/150.
Cherry-picks the commit from #1686 (let me know if you want this commit message changed).

### DIFF
--- a/src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.json
+++ b/src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.json
@@ -5,7 +5,7 @@
   "numberOfTests" : 463,
   "header" : [
     "Test vectors of type EcdsaBitcoinVerify are meant for the verification",
-    "of a ECDSA variant used for bitcoin, that add signature non-malleability."
+    "of a ECDSA variant used for Bitcoin, that add signature non-malleability."
   ],
   "notes" : {
     "ArithmeticError" : {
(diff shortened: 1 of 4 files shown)


real-or-random:
Can you also adjust the references in https://github.com/bitcoin-core/secp256k1/blob/master/src/wycheproof/WYCHEPROOF_COPYING?

fanquake:
> Can you also adjust the references

Fixed up.

src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.json
```diff
@@ -95,8 +95,8 @@
     },
     "SignatureMalleabilityBitcoin" : {
       "bugType" : "SIGNATURE_MALLEABILITY",
-      "description" : "\"BitCoins\"-curves are curves where signature malleability can be a serious issue. An implementation should only accept a signature s where s < n/2. If an implementation is not meant for uses cases that require signature malleability then this implementation should be tested with another set of test vectors.",
-      "effect" : "In bitcoin exchanges, it may be used to make a double deposits or double withdrawals",
+      "description" : "Signature malleability can be a serious issue in Bitcoin. An implementation should only accept a signature s where s < n/2. If an implementation is not meant for use cases that require signature malleability then this implementation should be tested with another set of test vectors.",
```
josibake, line 98: I think this comment should read something like: 

```suggestion
      "description" : "Signature malleability can be a serious issue in Bitcoin. An implementation should only accept a signature s where s < n/2. If an implementation is meant for use cases that require signature malleability then this implementation should be tested with another set of test vectors.",
```

jonasnick: Only one test vector has the `SignatureMalleabilityBitcoin` flag and it tests a failing signature. So I cannot make sense of the original descriptioin either. Maybe it should say something like:

If an implementation is meant for use cases that tolarete signature malleability then this implementation should not be tested with this set of test vectors.

josibake: Opened https://github.com/C2SP/wycheproof/pull/153 upstream, will pipe through the changes to here if/when the upstream PR is merged.

PR #1698, opened 2025-07-10
Add BIP352 module (take 3)

josibake:
This PR implements [BIP352 - Silent payments](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki). It is recommended to read through the BIP before reviewing this PR.

This is a continuation of the work in https://github.com/bitcoin-core/secp256k1/pull/1519 and only opened as a new PR due to the comment history on #1519 becoming quite long and difficult to sift through. It is recommended reviewers go through #1519 for background context, if interested.

PR #1714, opened 2025-07-30
Add ECDSA pubkey recovery usage example

theStack:
The `recovery` module is probably not super-relevant these days for newer projects (the primary use-case I'm aware of is message signing in Bitcoin Core and Electrum, likely other wallets; something that is hopefully replaced by [BIP-322](https://github.com/bitcoin/bips/blob/master/bip-0322.mediawiki) one day), but it still seems better to have an example than to have none. It contains all of the five API calls, i.e. for signing, recovering, converting, serializing, parsing. As usual with examples, a lot of code and comments are duplicated (e.g. context creation, keypair generation, cleanup with secret key clearing etc.).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -166,6 +166,17 @@ if BUILD_WINDOWS
 ecdsa_example_LDFLAGS += -lbcrypt
 endif
 TESTS += ecdsa_example
+if ENABLE_MODULE_RECOVERY
+noinst_PROGRAMS += recovery_example
+recovery_example_SOURCES = examples/recovery.c
+recovery_example_CPPFLAGS = -I$(top_srcdir)/include -DSECP256K1_STATIC
+recovery_example_LDADD = libsecp256k1.la
+recovery_example_LDFLAGS = -static
+if BUILD_WINDOWS
+recovery_example_LDFLAGS += -lbcrypt
+endif
+TESTS += recovery_example
+endif
 if ENABLE_MODULE_ECDH
 noinst_PROGRAMS += ecdh_example
 ecdh_example_SOURCES = examples/ecdh.c
--- a/examples/recovery.c
+++ b/examples/recovery.c
@@ -1,0 +1,139 @@
+/*************************************************************************
+ * To the extent possible under law, the author(s) have dedicated all    *
+ * copyright and related and neighboring rights to the software in this  *
+ * file to the public domain worldwide. This software is distributed     *
+ * without any warranty. For the CC0 Public Domain Dedication, see       *
+ * EXAMPLES_COPYING or https://creativecommons.org/publicdomain/zero/1.0 *
+ *************************************************************************/
+
+/** This file demonstrates how to use the recovery module to create a
+  * recoverable ECDSA signature and extract the corresponding
+  * public key from it.
+  */
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <assert.h>
+#include <string.h>
+
+#include <secp256k1.h>
+#include <secp256k1_recovery.h>
+
+#include "examples_util.h"
+
+int main(void) {
+    unsigned char msg[32] = "this_could_be_the_hash_of_a_msg";
+    unsigned char seckey[32];
+    unsigned char randomize[32];
+    unsigned char recoverable_sig_ser[64];
+    unsigned char serialized_pubkey[33];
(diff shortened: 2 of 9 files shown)


Makefile.am
```diff
@@ -163,6 +163,17 @@ if BUILD_WINDOWS
 ecdsa_example_LDFLAGS += -lbcrypt
 endif
 TESTS += ecdsa_example
+if ENABLE_MODULE_RECOVERY
+noinst_PROGRAMS += recovery_example
+recovery_example_SOURCES = examples/recovery.c
+recovery_example_CPPFLAGS = -I$(top_srcdir)/include -DSECP256K1_STATIC
```
hebasto, line 172: I understand that this line follows the pattern, but `-DSECP256K1_STATIC` flag is relevant for Windows only:https://github.com/bitcoin-core/secp256k1/blob/20e3b4474639edf8fab6766ac6bda161bd9f52e5/include/secp256k1.h#L165-L168

theStack:
I suppose the CI failures for the Valgrind jobs on MacOS Ventura (https://github.com/bitcoin-core/secp256k1/actions/runs/16623024583/job/47031936047 and https://github.com/bitcoin-core/secp256k1/actions/runs/16623024583/job/47031935996) are caused by the same memcmp-issue as observed in the Silent Payments PR (see https://github.com/bitcoin-core/secp256k1/pull/1519#discussion_r1993755203 ff.). Stole the `my_memcmp_var` reimplementation in #1698 from @josibake (thanks!) and force-pushed. Will consider moving it to `examples_util.h` if it solves the problem.

Makefile.am
theStack: I see, so a separate PR could make sense where this is moved into the `if BUILD_WINDOWS` block for each example? Happy to open one (or review) if that is what you had in mind.

josibake:
The CI failure seems unrelated and probably just needs a restart; its caused by a 503 when installing packages inside the docker container (if I'm reading the logs correctly): https://github.com/bitcoin-core/secp256k1/actions/runs/16635429498/job/47075060509?pr=1714

fanquake:
Kicked the failure.

theStack:
> The CI failure seems unrelated and probably just needs a restart; its caused by a 503 when installing packages inside the docker container (if I'm reading the logs correctly): https://github.com/bitcoin-core/secp256k1/actions/runs/16635429498/job/47075060509?pr=1714

Yeah indeed. Before the latest force-push (i.e. still using `memcmp`) the failure looked like this though:
https://github.com/bitcoin-core/secp256k1/actions/runs/16623024583/job/47031936047#step:7:1141
pointing to the line using memcmp, so I guess I have to keep the workaround.

> Kicked the failure.

Thanks!

josibake:
> Yeah indeed. Before the latest force-push

Sorry, my comment was unclear. What I meant to say is your solution works and the _current_ failure is definitely not related.

Makefile.am
real-or-random: I'm not convinced that `if BUILD_WINDOWS` is better. Sure, the macro is checked only on Windows but it certainly doesn't hurt to set it everywhere.

examples/recovery.c
```diff
@@ -0,0 +1,153 @@
...
+    assert(return_val);
+
+    /* Successful recovery guarantees a correct signature, but we also do an explicit verification
+       do demonstrate how to convert a recoverable to a normal ECDSA signature */
+    return_val = secp256k1_ecdsa_recoverable_signature_convert(ctx, &normal_sig, &recoverable_sig);
+    assert(return_val);
+    if (!secp256k1_ecdsa_verify(ctx, &normal_sig, msg, &recovered_pubkey)) {
+        printf("Signature verification with converted recoverable signature failed\n");
+        return EXIT_FAILURE;
+    }
```
jonasnick, line 110: This is true for signatures created with `secp256k1_ecdsa_sign_recoverable`, but not necessarily for arbitrarily generated ECDSA signatures (see #1718).

theStack: That's interesting, wasn't aware. Added an `secp256k1_ecdsa_signature_normalize` call before the verification (it's a no-op in this example, but for demonstration purposes) and tried to explain that with a comment.

theStack:
Rebased on master (resolving a small conflict in CHANGELOG.md).

PR #1729, opened 2025-09-01
hash: Use size_t instead of int for RFC6979 outlen copy

john-moffett:
If `outlen > INT_MAX` it results in segfault or hang (when `outlen` is a multiple of 2^32) on most implementations due to conversion in: `int now = outlen` producing negative values or zero. Unreachable in current code and highly improbable in future practice, but fits contract better and fixes a couple of compiler warnings.

### DIFF
--- a/src/hash_impl.h
+++ b/src/hash_impl.h
@@ -265,7 +265,7 @@ static void secp256k1_rfc6979_hmac_sha256_generate(secp256k1_rfc6979_hmac_sha256
 
     while (outlen > 0) {
         secp256k1_hmac_sha256 hmac;
-        int now = outlen;
+        size_t now = outlen;
         secp256k1_hmac_sha256_initialize(&hmac, rng->k, 32);
         secp256k1_hmac_sha256_write(&hmac, rng->v, 32);
         secp256k1_hmac_sha256_finalize(&hmac, rng->v);


fanquake:
>  fixes a couple of compiler warnings

Which compiler / warnings?

john-moffett:
Not with default flags. I used `clang -Wimplicit-int-conversion -Wshorten-64-to-32`. It'd probably show up with `gcc -Wconversion`. There are a lot of benign warnings, but this one stood out to me.

src/hash_impl.h
real-or-random, line 1: If you want to improve this further, I'd suggest
 - renaming `now` to `chunk_len` (consistency with the SHA256 implementation above) 
 - moving the entire `if` block just below the initialization of `now` (or `chunk_len`). (The goal is to compute max(outlen, 32). Spreading this over two blocks is a bit silly. It's probably a result of the C89 rule that  declarations need to be at the beginning of a block. *edit: Indeed, see https://github.com/bitcoin-core/secp256k1/commit/792bcdb015b0272928d4dd602557faa685dcd5a1#diff-e91b67aa33d606267c28ce5da48a89cc789f4081c176eb581a9bff780afd1d6cR260-L262* )

john-moffett: Thanks! I'll hold off for the moment. Hopefully it'll inspire someone to replace this nonce generation process with something more lightweight and straightforward like BIP340's approach. :)

theStack (approved):
Code-review ACK 960ba5f9c60c3fd454ad4160ce9131eb77d61d7e

PR #1731, opened 2025-09-02
schnorrsig: Securely clear buf containing k or its negation

john-moffett:
Follow-up to #1579. `buf` still holds the nonce or its negation, so ought to be cleared.


real-or-random:
@theStack  Want to review this? :)

john-moffett:
> we could rename it to reflect that, e.g. to `nonce_buf` (fwiw, in the ECDSA signing routine the name [`nonce32`](https://github.com/bitcoin-core/secp256k1/blob/f36afb8b3dd7daf9edf4bf15c49fcd540f8ce393/src/secp256k1.c#L553) is used)

Good idea, as it makes it clearer that it's sensitive information, too.

PR #1732, opened 2025-09-02
bench: replace wall-clock timer with per-process CPU timer 

Raimo33:
## Goal
This PR refactors the benchmarking functions as per #1701, in order to make benchmarks more deterministic and less influenced by the environvment. 

This is achieved by replacing Wall-Clock Timer with Per-Process CPU Timer when possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,8 @@ Benchmark
 ------------
 If configured with `--enable-benchmark` (which is the default), binaries for benchmarking the libsecp256k1 functions will be present in the root directory after the build.
 
+For reliable benchmarks, it is strongly recommended to pin the process to a single CPU core and to disable CPU frequency scaling.
+
 To print the benchmark result to the command line:
 
     $ ./bench_name


real-or-random:
Just some quick comments:
 
>  [x] remove the number of runs (`count`) in favor of simpler cleaner approach with just number of iterations (`iter`).

I think there's a reason to have this. Some benchmarks take much longer than others, so it probably makes sense to run fewer iters for these.

>  [x]  remove `min` and `max` statistics in favor of simpler approach with just `avg`.

I think `min` and `max` are useful. For constant-time code, you can also compare `min`. And `max` gives you an idea if there were fluctuations or not.

>  [x]  remove needless fixed point conversion in favor of simpler floating point divisions.

Well, okay, that has a history; see https://github.com/bitcoin-core/secp256k1/pull/689. It's debatable if it makes sense to avoid floating point math, but as long as it doesn't get in your way here, it's a cool thing to keep it. :D

real-or-random:
It will be useful to split your changes into meaningful and separate commits, see https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#committing-patches.

Raimo33:
I think `min` and `max` just complicate things. let me explain:
first of all, as it is right now, they don't even measure the min and max, they just measure the min/max of the averages of all runs. aka not the absolute. Furthermore, in order to have them, one would need to run all the iterations 10 times more. benchmarks are already slow, adding this min/max slows them by 10 fold. imho it's completely unnecessary. @real-or-random

sipa:
If we're going to rework this, I'd suggest using the stabilized quartiles approach from https://cr.yp.to/papers/rsrst-20250727.pdf:
* StQ1: the average of all samples between 1st and 3rd octile
* StQ2: the average of all samples between 3rd and 5th octile
* StQ3: the average of all samples between 5th and 7th octile

Raimo33:
> I think there's a reason to have this. Some benchmarks take much longer than others, so it probably makes sense to run fewer iters for these.

right now all benchmarks are run with count=10 and fixed iters (apart from ecmult_multi which adjusts the number of iters, not count).

therefore `count` is only useful to extrapolate `min` and `max`

Raimo33:
> Well, okay, that has a history; see #689. It's debatable if it makes sense to avoid floating point math, but as long as it doesn't get in your way here, it's a cool thing to keep it. :D

I disagree with https://github.com/bitcoin-core/secp256k1/pull/689. It overcomplicate things for the sake of not having floating point math. those divisions aren't even in the hot path, they're outside the benchmarks.

sipa:
Concept NACK on removing any ability to observe variance in timing. The current min/avg/max are far from perfect, but they work fairly well in practice. Improving is welcome, but removing them is a step backwards.

Raimo33:
> Concept NACK on removing any ability to observe variance in timing. The current min/avg/max are far from perfect, but they work fairly well in practice. Improving is welcome, but removing them is a step backwards.

what is the usefulness of measuring min/max when we are removing OS interference & thermal throttling out of the equation? min/max will be extremely close to the avg no matter how bad the benchmarked function is.

Raimo33:
by the way, `gettimeofday()` is officially discouraged since 2008 in favor of `clock_gettime()`. The POSIX standard marks it as **obsolescent** but still provides the API for backward compatibility.

Raimo33:
even though the [manual](https://linux.die.net/man/3/clock_gettime) says that `CLOCK_PROCESS_CPUTIME_ID` is only useful if the process is locked to a core, modern CPUs have largely [addressed this issue](https://chatgpt.com/s/dr_68b7486a05a481919d9f121d182cc0cf). So I think it is fair to compile `CLOCK_PROCESS_CPUTIME_ID` even though we don't have the guarantee that the user has pinned the benchmarking process to a core. The worst case scenario is a unreliable benchmark, which the current repo has anyways.

I added a line in the README.md for best practices to run the benchmarks. 

I also tried adding a function to pin the process to a core directly in C, but there's no standard POSIX compliant way to do so. There is `pthread_setaffinity_np()` on linux, where 'np' stands for 'not portable'

real-or-random:
> > Concept NACK on removing any ability to observe variance in timing. The current min/avg/max are far from perfect, but they work fairly well in practice. Improving is welcome, but removing them is a step backwards.
> 
> what is the usefulness of measuring min/max when we are removing OS interference & thermal throttling out of the equation? min/max will be extremely close to the avg no matter how bad the benchmarked function is.

The point is exactly having a simple way of verifying that there's indeed no interference. Getting rid of sources of variance is hard to get right, and it's impossible to get a perfect solution. (This discussion shows this!) So we better have a way of spotting if something is off.

I like the stabilized quartiles idea.

Raimo33:
> I like the stabilized quartiles idea.

tbh it scares me a bit, will see what I can do. Maybe in a future PR.

CMakeLists.txt
```diff
@@ -30,6 +30,8 @@ set(${PROJECT_NAME}_LIB_VERSION_AGE 0)
 #=============================
 set(CMAKE_C_STANDARD 90)
 set(CMAKE_C_EXTENSIONS OFF)
+# Enable POSIX features while maintaining ISO C compliance
+add_compile_definitions(_POSIX_C_SOURCE=200112L) #needed for `clock_gettime()` in bench.c
```
real-or-random: we'll only need this in the benchmarks. Can you define it at the top of `bench.c`?

README.md
```diff
@@ -159,6 +159,8 @@ Benchmark
 ------------
 If configured with `--enable-benchmark` (which is the default), binaries for benchmarking the libsecp256k1 functions will be present in the root directory after the build.
 
+For an accurate benchmark, it is strongly recommended to pin the process to a dedicated CPU core and to disable CPU frequency scaling.
```
real-or-random: https://github.com/bitcoin/bitcoin/blob/46369583f3a99d2b403349790f8b26dbc28de132/src/bench/bench_bitcoin.cpp#L94 may be useful

src/bench.h
```diff
@@ -7,30 +7,63 @@
 #ifndef SECP256K1_BENCH_H
 #define SECP256K1_BENCH_H
 
+#if defined(_WIN32)
+# include <windows.h>
+#else
+# include <time.h>
+# include <sys/time.h>
```
real-or-random: Do we need both?

src/bench.h
```diff
@@ -7,30 +7,63 @@
...
+    k.LowPart = kernel.dwLowDateTime;
+    k.HighPart = kernel.dwHighDateTime;
+    u.LowPart = user.dwLowDateTime;
+    u.HighPart = user.dwHighDateTime;
+
+    return (int64_t)((k.QuadPart + u.QuadPart) / 10);
+#else /* POSIX */
+
+# if defined(CLOCK_PROCESS_CPUTIME_ID)
+    /* in theory, CLOCK_PROCESS_CPUTIME_ID is only useful if the process is locked to a core. see https://linux.die.net/man/3/clock_gettime */
```
real-or-random: and in practice?

src/bench.h
```diff
@@ -7,30 +7,63 @@
...
+        return 0;
+    }
+    return (int64_t)ts.tv_sec * 1000000 + ts.tv_nsec / 1000;
+# else
+    /* WARN: timer is influenced by environvment (OS, scheduling, interrupts...) */
     struct timeval tv;
-    gettimeofday(&tv, NULL);
-    return (int64_t)tv.tv_usec + (int64_t)tv.tv_sec * 1000000LL;
+    gettimeofday((struct timeval*)&tv, NULL);
+    return (int64_t)tv.tv_sec * 1000000 + tv.tv_usec;
```
real-or-random: How is this diff an improvement?
  - The pointer cast is a no-op.
  - `LL` is not needed.

src/bench.h
```diff
@@ -7,30 +7,63 @@
...
+    }
+    return (int64_t)ts.tv_sec * 1000000 + ts.tv_nsec / 1000;
+# elif defined(CLOCK_REALTIME)
+    /* WARN: timer is influenced by environvment (OS, scheduling, interrupts...) */
+    struct timespec ts;
+    if (clock_gettime(CLOCK_REALTIME, &ts) != 0) {
+        return 0;
+    }
+    return (int64_t)ts.tv_sec * 1000000 + ts.tv_nsec / 1000;
+# else
```
real-or-random: This could be DRYed more. Maybe it's a good idea to print to the user which clock method is used.

src/bench.h
```diff
@@ -7,92 +7,64 @@
...
 
-static int64_t gettime_i64(void) {
-#if (defined(_MSC_VER) && _MSC_VER >= 1900)
-    /* C11 way to get wallclock time */
-    struct timespec tv;
-    if (!timespec_get(&tv, TIME_UTC)) {
-        fputs("timespec_get failed!", stderr);
-        exit(EXIT_FAILURE);
+    if (GetProcessTimes(GetCurrentProcess(), &creation, &exit, &kernel, &user) == 0) {
+        return 0;
```
real-or-random: What's the rationale for the change on Windows?

src/bench.h
Raimo33: yes, one for `gettimeofday()` and one for `clock_gettime()`. they don't conflict.

src/bench.h
Raimo33: see https://github.com/bitcoin-core/secp256k1/pull/1732#issuecomment-3246579226

src/bench.h
Raimo33: yes, I tried everything you mentioned.

removing redundancy here can be done by just using the ifdefs for the constant, but you'd need a different path for the last fallback. the last case uses tv, not ts.

I tried issuing warnings with `#warning` to tell the user about the precision of the clock but it's not compatible with c90.

src/bench.h
Raimo33: previously wall-clock times.
now per-process cpu time.

Raimo33:
> The point is exactly having a simple way of verifying that there's indeed no interference. Getting rid of sources of variance is hard to get right, and it's impossible to get a perfect solution. (This discussion shows this!) So we better have a way of spotting if something is off.

fine, but at least let's make it optional. I don't like my benchmarks being 10 times slower just because min/max need to be computed. if I say 20'000 iterations I want it to be 20`000 iterations.

src/bench.h
Raimo33: you're right. removed.

src/bench.h
Raimo33: Here's the DRY'd version:

```c
# if defined(CLOCK_PROCESS_CPUTIME_ID)
    /* in theory, CLOCK_PROCESS_CPUTIME_ID is only useful if the process is locked to a core. see https://linux.die.net/man/3/clock_gettime */
const clockid_t clk_id = CLOCK_PROCESS_CPUTIME_ID;
# elif defined(CLOCK_MONOTONIC)
    /* WARN: timer is influenced by environvment (OS, scheduling, interrupts...) */
const clockid_t clk_id = CLOCK_MONOTONIC;
# elif defined(CLOCK_REALTIME)
    /* WARN: timer is influenced by environvment (OS, scheduling, interrupts...) */
const clockid_t clk_id = CLOCK_REALTIME;
# else
# define USE_GETTIMEOFDAY
# endif

# ifndef USE_GETTIMEOFDAY
struct timespec ts;
if (clock_gettime(clk_id, &ts) != 0) {
    return 0;
}
return (int64_t)ts.tv_sec * 1000000 + ts.tv_nsec / 1000;
# else
    /* WARN: timer is influenced by environvment (OS, scheduling, interrupts...) */
struct timeval tv;
gettimeofday(&tv, NULL);
return tv.tv_usec + (int64_t)tv.tv_sec * 1000000;
# endif
```

less readable imo

real-or-random: > I tried issuing warnings with `#warning` to tell the user about the precision of the clock but it's not compatible with c90.

What I had in mind is printing the warning out at runtime.

src/bench.h
real-or-random: I think then it could be slightly more portable to include only what we'll use.

src/bench.h
real-or-random: Okay, sorry. I was aware of the comment in the issue, but a future reader of the code comment will not. So what I'm trying to say is that the code comment can be improved.

Raimo33: Added comment and link to chatgpt deepsearch. might not be very professional but it includes the full analysis and lots of citations.

src/bench.h
Raimo33: Agree. Removed.

Raimo33: just tried. unfortunately the macros to detect wether `gettimeofday()` or `clock_gettime()` is needed are defined in the `time.h` header. so the only viable solution is:

```c
#include <time.h>
#if !defined(CLOCK_PROCESS_CPUTIME_ID) && !defined(CLOCK_MONOTONIC) && !defined(CLOCK_REALTIME)
#  include <sys/time.h>
#endif
```

which every LLM tells me is 'inherently unreliable' and can lead to portability issues.

> 'Including sys/time.h on modern POSIX systems that already have time.h and clock_gettime doesn’t hurt, and ensures gettimeofday is available if needed.'

> 'Trying to skip it adds complexity with almost no benefit.'

src/bench.h
Raimo33: good suggestion. done.

CMakeLists.txt
Raimo33: that was my initial setup. but it doesn't work because it has to be before *any* include. I added it to the `src/CMakeLists.txt` targeting only the benchmarks.

README.md
l0rinc: This will be more reliable at best, but not more "accurate", since it doesn't reflect actual usage.

Raimo33: agreed. changed.

src/bench.h
real-or-random: Okay, but if we include time.h anyway, then let's stick to `clock_gettime` on POSIX?  https://pubs.opengroup.org/onlinepubs/9799919799/functions/clock_getres.html says that the proper macro to check is `_POSIX_CPUTIME` and that `CLOCK_MONOTONIC` must be supported, so there's no need to check for it.

real-or-random:
> > The point is exactly having a simple way of verifying that there's indeed no interference. Getting rid of sources of variance is hard to get right, and it's impossible to get a perfect solution. (This discussion shows this!) So we better have a way of spotting if something is off.
> 
> fine, but at least let's make it optional. I don't like my benchmarks being 10 times slower just because min/max need to be computed. if I say 20'000 iterations I want it to be 20`000 iterations.

Not sure. I suggest making this PR focused on a single (uncontroversial) change, which is switching to per process clocks.

src/bench.h
Raimo33: You're right, the latest specification makes `CLOCK_MONOTONIC` support mandatory, this makes it way simpler. no need to include `<sys/time.h>` anymore.

Raimo33: This would be possible if every compiler supported the Std 1003.1-2024 version of POSIX, which is unrealistic. Until then, for backward compatibility, I kept `CLOCK_REALTIME` as the last fallback.

Raimo33: and btw, the `_POSIX_CPUTIME` and `_POSIX_MONOTONIC` flags are to be used when you want to ensure that the library has mandatory support for those features. In my case for example, gcc doesnt expose `_POSIX_MONOTONIC` but still supports it. Therefore it's better to check `CLOCK_PROCESS_CPUTIME_ID` and `CLOCK_MONOTONIC` directly.

src/bench.h
```diff
@@ -7,30 +7,57 @@
...
+
+    return (int64_t)((k.QuadPart + u.QuadPart) / 10);
+
+#else /* POSIX */
+
+# if defined(CLOCK_PROCESS_CPUTIME_ID)
+    /* In theory, CLOCK_PROCESS_CPUTIME_ID is only useful if the process is locked to a core. 
+    * See https://linux.die.net/man/3/clock_gettime. 
+    * In practice, modern CPUs have largely addressed this issue. The worst case scenario is a unreliable benchmark. 
+    * See https://chatgpt.com/s/dr_68b7486a05a481919d9f121d182cc0cf
```
sipa: Can you find a source that's a bit more authoritative than ChatGPT?

src/bench.h
```diff
@@ -176,6 +204,14 @@ static int get_iters(int default_iters) {
     }
 }
 
+static void print_clock_info(void) {
+#if defined(_WIN32) || defined(CLOCK_PROCESS_CPUTIME_ID)
+    printf("INFO: using CPU timer, results are not influenced by other running processes\n\n");
```
sipa: That seems like an exaggeration; hyperthreading, task swapping, and competing memory bandwidth, may all still affect the results.

Raimo33: you're right. memory bandwith, cache pollution, hyperthreading, hot/cold branch predictor...
But task swapping shouldn't influence `CLOCK_PROCESS_CPUTIME_ID`. I changed the message to notify the user that he should probably pin the process.

src/bench.h
Raimo33: yes. Added multiple links.

src/bench.h
```diff
@@ -7,30 +7,54 @@
...
+    u.LowPart = user.dwLowDateTime;
+    u.HighPart = user.dwHighDateTime;
+
+    return (int64_t)((k.QuadPart + u.QuadPart) / 10);
+
+#else /* POSIX */
+
+# if defined(CLOCK_PROCESS_CPUTIME_ID)
+    /* In theory, CLOCK_PROCESS_CPUTIME_ID is only useful if the process is locked to a core. see https://man7.org/linux/man-pages/man2/clock_gettime.2.html *
+     * in practice, modern CPUs have synchronized TSCs which addresses this issue. see https://docs.amd.com/r/en-US/ug1586-onload-user/Timer-TSC-Stability   */
```
real-or-random:
```suggestion
    /* In theory, CLOCK_PROCESS_CPUTIME_ID is only useful if the process is locked to a core,
     * see `man clock_gettime` on Linux. In practice, modern CPUs have synchronized TSCs which
     * address this issue, see https://docs.amd.com/r/en-US/ug1586-onload-user/Timer-TSC-Stability . */
```

src/bench.h
real-or-random: > I changed the message to notify the user that he should probably pin the process. 

If that's not required on modern CPUs, why advise to do this?

We could just say "INFO: Using per-process CPU timer".

Raimo33:
I've applied the recommended changes and split the PR into multiple commits for simpler diffs

l0rinc:
I haven't investigated this in detail but couldn't find any related issue: were there any efforts to use https://nanobench.ankerl.com like we do in Core instead? That will likely need some adjustments on the benchmarks themselves, but at least we wouldn't be rolling our own framework.

Raimo33:
not familiar with nano bench, but I can see it uses `std::chrono`. which uses `CLOCK_MONOTONIC` (our fallback) under the hood, not TSC. 

also the user would need to have a c++ compiler, not a big real but I think this repo needs to strictly adhere to c89.

Raimo33:
I realized the windows version I implemented was using a unprecise clock. drafting. will fix.

Raimo33:
I changed windows clock to `QueryPerformanceCounter()` which has microsecond precision.
I also removed if checks from `clock_gettime()` as it can only fail for programming errors.

real-or-random:
> I haven't investigated this in detail but couldn't find any related issue: were there any efforts to use [nanobench.ankerl.com](https://nanobench.ankerl.com) like we do in Core instead?

nanobench is a framework for benchmarking C++ but this is C. Of course, we could still call our C from C++, but adding C++ to the code base seems a bit overkill to me if if's "just" for the purpose of benchmarks. I agree that using an existing framework may be a good idea, but I'm not aware of any in pure C.

l0rinc:
> but adding C++ to the code base seems a bit overkill to me if if's "just" for the purpose of benchmarks

The purpose would be to standardize the benchmarking to avoid reinventing solutions to problems that aren't strictly related to the purpose of this library. It would also help with trusting the result more, since we already know how that benchmarking library behaves.
The main client of this library is C++ code, it doesn't seem far-fetched to me to test that.

real-or-random:
> > but adding C++ to the code base seems a bit overkill to me if if's "just" for the purpose of benchmarks
> 
> The purpose would be to standardize the benchmarking to avoid reinventing solutions to problems that aren't strictly related to the purpose of this library. It would also help with trusting the result more, since we already know how that benchmarking library behaves. The main client of this library is C++ code, it doesn't seem far-fetched to me to test that.

Sorry, I didn't want to be discouraging. I don't think at all that nanobench is a far-fetched idea. And I truly agree that reinventing the wheel is not great.

In the end, what counts is the maintenance burden. My main concern with C++ is that I'm not sure how fluent our regular contributors are in C++. (I'm not, but luckily I'm not the only one here.) If the bit of C++ is harder to touch (for us) than this code, then nanobench is not a good idea. If, on the other hand, we have people who can maintain C++, or if we get support from the Core contributors who are familiar with nanobench, then this can be a win.

I'd be happy to see a demo/WIP PR but if you feel that this is a lot of work, then it might be a good idea to see what other contributors and maintainers think.

Raimo33:
This PR currently conflicts with https://github.com/bitcoin-core/secp256k1/pull/1734

furszy:
> This PR currently conflicts with #1734

Don't worry about it. #1734 only moves `gettime()` so it can be accessed by the test framework too. Any changes you make here will function correctly there as well, and vice versa.

src/bench.h
```diff
@@ -176,6 +190,14 @@ static int get_iters(int default_iters) {
     }
 }
 
+static void print_clock_info(void) {
+#if defined(CLOCK_PROCESS_CPUTIME_ID)
+    printf("INFO: Using per-process CPU timer\n\n");
+#else
+    printf("WARN: using Wall-Clock timer, results are highly influenced by other running processes\n\n");
```
real-or-random:
```suggestion
    printf("WARN: Using wall-clock timer instead of per-process CPU timer.\n\n");
```

Let's just stick to the facts and omit the interpretation of them, this makes it easier to move forward. ;)

README.md
```diff
@@ -159,6 +159,8 @@ Benchmark
 ------------
 If configured with `--enable-benchmark` (which is the default), binaries for benchmarking the libsecp256k1 functions will be present in the root directory after the build.
 
+For reliable benchmarks, it is strongly recommended to pin the process to a dedicated CPU core and to disable CPU frequency scaling.
```
real-or-random:
```suggestion
For reliable benchmarks, it is strongly recommended to pin the process to a single CPU core and to disable CPU frequency scaling.
```

Dedicated sounds as if the core is only responsible for this process, which would be nice but it's typically not doable.

src/bench.c
```diff
@@ -4,6 +4,8 @@
  * file COPYING or https://www.opensource.org/licenses/mit-license.php.*
  ***********************************************************************/
 
+#define _POSIX_C_SOURCE 200112L /* for clock_gettime() */
```
hebasto: Are there any references explaining why the value `200112L` was chosen?

hebasto (commented):
According to https://www.man7.org/linux/man-pages/man3/clock_gettime.3.html:
> Link with `-lrt` (only for glibc versions before 2.17).

I think this check, and adding the `-lrt` flag if necessary, should be included in both build systems.

src/bench.c
Raimo33: no particular reason besides the fact that it is the bare minimum version to make it work on my machine and CI.
The reported `199309L` version number didn't expose the necessary macros.

Raimo33:
> I think this check, and adding the `-lrt` flag if necessary, should be included in both build systems.

Agreed, good catch! what do you think about something like this?

```.cmake
include(CheckFunctionExists)
check_function_exists(clock_gettime HAVE_CLOCK_GETTIME)
if(NOT HAVE_CLOCK_GETTIME)
    # On some platforms, clock_gettime requires librt
    target_link_libraries(your_target PRIVATE rt)
endif()
```

hebasto:
Friendly ping @martinus, benchmark connoisseur :)

Could you please give a rough estimate of these changes at the concept level?

real-or-random:
> Could you please give a rough estimate of these changes at the concept level?

And what's your take on using nanobench instead, even though this is a pure C library instead of C++?

hebasto:
> ## Goal
> 
> This PR refactors the benchmarking functions as per #1701, in order to make benchmarks more deterministic and less influenced by the environvment.
> 
> This is achieved by replacing Wall-Clock Timer with Per-Process CPU Timer when possible.

I've tested 8925b9588259d27257f42eb4fb8bebd4a7be4d95 by running it alongside `stress --cpu $(nproc)` and observed the same effect on benchmark average values as on the master branch.

Could you please clarify how one would measure or observe the claimed "less influenced by the environment" behaviour?

Raimo33:
`stress --cpu` spawns workers that use 100% of the cpu. your test pegs every core to 100% utilization. and each gets probably pinned to a physical core. therefore you won't see any difference in terms of variance.

a better example would be starting various threads with random usage spikes at random times **on the same core** the benchmark is running on.

basically, if **you have no other processes running on that CPU**, then this PR won't show any improvement.

this PR helps in the more realistic scenario where a user has background processes running, and the scheduler assigns multiple of them on the same CPU that's running the benchmark, putting the thread to sleep, which the wall clock timer doesn't account for.

also, this PR is not merely a replacement of wall clock time with CPU time, but it also modernizes the clock function as per the Unix standard.

src/bench.c
hebasto: > The reported `199309L` version number didn't expose the necessary macros.

I can't reproduce this on my machine.

> no particular reason besides the fact that it is the bare minimum version to make it work on my machine...

Could you share your build environment details so I can try to reproduce the issue?

> ... and CI.

It seems that CI is fine with the documented `199309L` value.
See: https://github.com/hebasto/secp256k1/actions/runs/18008380590

hebasto:
> > I think this check, and adding the `-lrt` flag if necessary, should be included in both build systems.
> 
> Agreed, good catch! what do you think about something like this?
> 
> ```cmake
> include(CheckFunctionExists)
> check_function_exists(clock_gettime HAVE_CLOCK_GETTIME)
> if(NOT HAVE_CLOCK_GETTIME)
>     # On some platforms, clock_gettime requires librt
>     target_link_libraries(your_target PRIVATE rt)
> endif()
> ```

Feel free to grab f59c45f2cef03681074eef00046d3b146daf42e8 from https://github.com/hebasto/secp256k1/commits/pr1732/0925.cmake.

UPD. @real-or-random do we need a CI job to build on some old system with old glibc?

src/bench.h
```diff
@@ -176,6 +190,14 @@ static int get_iters(int default_iters) {
...
 }
 
+static void print_clock_info(void) {
+#if defined(CLOCK_PROCESS_CPUTIME_ID)
+    printf("INFO: Using per-process CPU timer\n\n");
+#else
+    printf("WARN: Using wall-clock timer instead of per-process CPU timer.\n\n");
+#endif
+}
+
```
hebasto, line 166: 5fe7c017682240295d701570f052b6ddfb2d1fb7:

Do these messages make sense on Windows?

Raimo33: while on windows there's no option for per-process cpu clock (or at least not high precision), I still think issuing the warning is fine.

hebasto: It seems a bit of a stretch to call the native Windows [QPC](https://learn.microsoft.com/en-us/windows/win32/sysinfo/acquiring-high-resolution-time-stamps) framework a "wall-clock timer".

Raimo33: what about this?
"WARN: Using global timer instead of per-process CPU timer."

Raimo33:
> Feel free to grab [f59c45f](https://github.com/bitcoin-core/secp256k1/commit/f59c45f2cef03681074eef00046d3b146daf42e8) from https://github.com/hebasto/secp256k1/commits/pr1732/0925.cmake.

This seems overcomplicated and convoluted. but if you manage to simplify it I'll include your commit.

Raimo33:
Fixed by unconditionally adding `-lrt` when available.
```
root@5997b2bd1aca:/workspace# ldd --version | head -1
ldd (Ubuntu EGLIBC 2.15-0ubuntu10.23) 2.15
root@5997b2bd1aca:/workspace# cmake --build build -t bench
[ 33%] Built target secp256k1_precomputed
[ 66%] Built target secp256k1
[100%] Built target bench
```

src/CMakeLists.txt
```diff
@@ -125,12 +125,17 @@ endif()
 unset(${PROJECT_NAME}_soversion)
 
 if(SECP256K1_BUILD_BENCHMARK)
+  find_library(RT_LIBRARY rt)
+  add_library(optional_rt INTERFACE)
+  if(RT_LIBRARY)
+    target_link_libraries(optional_rt INTERFACE rt)
+  endif()
```
purpleKarrot: Could you please extract this logic into a find-module, like `FindRT.cmake`? That module should provide an `IMPORTED` target with a namespace in the name.

Have a look at the [`FindIconv.cmake`](https://github.com/Kitware/CMake/blob/master/Modules/FindIconv.cmake) module as an example. This module also provides an interface library for functionality that may be found in an actual library or built into the C standard library.

README.md
```diff
@@ -157,6 +159,8 @@ Benchmark
 ------------
 If configured with `--enable-benchmark` (which is the default), binaries for benchmarking the libsecp256k1 functions will be present in the root directory after the build.
 
+For reliable benchmarks, it is strongly recommended to pin the process to a single CPU core and to disable CPU frequency scaling.
```
purpleKarrot, line 160: It would be great if the benchmark was configured in a way such that the guideline is followed.

```cmake
add_test(NAME secp256k1_bench COMMAND ...)
set_tests_properties(secp256k1_bench PROPERTIES
  PROCESSORS 1
  PROCESSOR_AFFINITY ON
)
```

I am not aware how to disable scaling, unfortunately.

src/CMakeLists.txt
```diff
@@ -125,12 +125,17 @@ endif()
...
   add_executable(bench bench.c)
```
purpleKarrot, line 138: Note that CMake requires that

* test names are unique *per directory*
* target names are unique.

While it is not strictly required to prefix test names with the project, it is still recommended, because it allows users to control which tests to run: `ctest -R secp256k1_`. This recommendation is already followed, good.

For target names, this requirement is much stronger. Since this project is used as a subproject in bitcoin, neither bitcoin nor any other of its subprojects can define a target with the name `bench`. It would be recommended to rename the target to `secp256k1_bench`.

src/CMakeLists.txt
Raimo33: This looks compilicated, almost overcomplicated, without much benefits, I won't be able to replicate FindIconv.cmake on my own. but you're welcome to provide a commit and I'll see if it should be cherrypicked.

src/CMakeLists.txt
Raimo33: Agreed, done!

README.md
Raimo33: This is a great idea, added.

src/CMakeLists.txt
Raimo33: although CI scripts now need to change to reflect the executable name changes

Raimo33:
As per @purpleKarrot suggestion, benchmarks now can be run via ctest as well (as opposed to ./bench_name). This would be the preferable way to run them as it automatically handles CPU pinning and affinity. This means that, if both `SECP256K1_BUILD_BENCHMARK` and `SECP256K1_BUILD_TESTS` flags are set, the standard `ctest` command will run benchs+tests.
This can be avoided by selectively choosing which "tests" to run with the following ctest syntax:
`ctest -R bench` or `ctest -R test` (which uses regex).

Otherwise we can evaluate the possibility of using labels for groups of tests.

real-or-random:
@hebasto @purpleKarrot Want to re-review this? :)

real-or-random:
This needs to be rebased after the merge of #1734, sorry.

Raimo33:
there appears to be an issue in the CI. @real-or-random can you trigger it again?

also there's one last thing I don't like about this PR: having to define `POSIX_C_SOURCE` in every source file. if anyone finds a way to define it once, perhaps in the `CMakeLists.txt` please submit it.

src/tests_exhaustive.c
```diff
@@ -4,6 +4,8 @@
  * file COPYING or https://www.opensource.org/licenses/mit-license.php.*
  ***********************************************************************/
 
+#define _POSIX_C_SOURCE 199309L /* for clock_gettime() */
+
```
hebasto: This doesn’t seem necessary here.

src/tests.c
```diff
@@ -4,6 +4,8 @@
  * file COPYING or https://www.opensource.org/licenses/mit-license.php.*
  ***********************************************************************/
 
+#define _POSIX_C_SOURCE 199309L /* for clock_gettime() */
+
```
hebasto: That's annoying that test sources have to be modified. See https://github.com/bitcoin-core/secp256k1/pull/1734#issuecomment-3447767148.

hebasto:
https://github.com/bitcoin-core/secp256k1/pull/1732#issuecomment-3335074974:
> This seems overcomplicated and convoluted.

https://github.com/bitcoin-core/secp256k1/pull/1732#discussion_r2394557478:
> This looks compilicated, almost overcomplicated, without much benefits...

The PR author appears to favor simplicity over correctness (see the comment above). I disagree with that approach.

> but you're welcome to provide a commit and I'll see if it should be cherrypicked.

Sure. Here is a branch that addresses some of @purpleKarrot's comments and most of mine: https://github.com/hebasto/secp256k1/commits/251026-pr1732.alt/. However, it doesn't yet include the necessary changes to the Autotools build system.

UPD. Asked a question about handling `_POSIX_C_SOURCE` on IRC.

Raimo33:
I agree, droppped the 3 unrelated commits from this PR, removed the unnecessary `#define`, cherry-picked your solution for the library detection

furszy:
Just started reviewing it; CMake compilation fails locally.
It should set `_POSIX_C_SOURCE=199309L` instead, so the compiler expands correctly to `-D_POSIX_C_SOURCE=199309L`. See
```diff
diff --git a/cmake/FindClockGettime.cmake b/cmake/FindClockGettime.cmake
--- a/cmake/FindClockGettime.cmake	(revision e14981e28e1c1e4b2fb6321cd94342d0b2849be7)
+++ b/cmake/FindClockGettime.cmake	(date 1761574207222)
@@ -20,7 +20,7 @@
 
 cmake_push_check_state(RESET)
 
-set(CMAKE_REQUIRED_DEFINITIONS -D_POSIX_C_SOURCE=199309L)
+set(CMAKE_REQUIRED_DEFINITIONS _POSIX_C_SOURCE=199309L)
 check_symbol_exists(clock_gettime "time.h" CLOCK_GETTIME_IS_BUILT_IN)
 set(${CMAKE_FIND_PACKAGE_NAME}_FOUND ${CLOCK_GETTIME_IS_BUILT_IN})
```

hebasto:
> Just started reviewing it; CMake compilation fails locally. It should set `_POSIX_C_SOURCE=199309L` instead, so the compiler expands correctly to `-D_POSIX_C_SOURCE=199309L` during `check_symbol_exists`.

I can't reproduce it. What CMake version are you using?

From CMake [docs](https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html):
> `CMAKE_REQUIRED_DEFINITIONS`
>
>    A [semicolon-separated list](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-lists) of compiler definitions, each of the form -DFOO or -DFOO=bar.

Raimo33:
> Just started reviewing it; CMake compilation fails locally. It should set `_POSIX_C_SOURCE=199309L` instead, so the compiler expands correctly to `-D_POSIX_C_SOURCE=199309L` during `check_symbol_exists`. See

Can you provide the error output? my CMake doesn't complain

furszy:
> Can you provide the error output? my CMake doesn't complain

Sure.
```
In file included from In file included from <built-in><built-in>:418:
<command line>:In file included from 1<built-in>:418:
:9<command line>:: 1In file included from :<built-in>::418:
4189error: :
<command line>:macro name must be an identifier<command line>:1: 1:9: 
:error: 9error: : #define -D_POSIX_C_SOURCE 199309L
        ^macro name must be an identifier
macro name must be an identifier
error: 
macro name must be an identifier
#define -D_POSIX_C_SOURCE 199309L
```

> I can't reproduce it. What CMake version are you using?

cmake version 3.22.3

hebasto:
> cmake version 3.22.3

Confirming. I'll suggest a fix shortly.

UPD. There was a [change](https://cmake.org/cmake/help/latest/release/3.26.html#other-changes) in CMake 3.26:
> For all `COMPILE_DEFINITIONS` properties, any leading `-D` on an item is removed...

hebasto:
> Just started reviewing it; CMake compilation fails locally. It should set `_POSIX_C_SOURCE=199309L` instead, so the compiler expands correctly to `-D_POSIX_C_SOURCE=199309L`. See
> 
> ```diff
> diff --git a/cmake/FindClockGettime.cmake b/cmake/FindClockGettime.cmake
> --- a/cmake/FindClockGettime.cmake	(revision e14981e28e1c1e4b2fb6321cd94342d0b2849be7)
> +++ b/cmake/FindClockGettime.cmake	(date 1761574207222)
> @@ -20,7 +20,7 @@
>  
>  cmake_push_check_state(RESET)
>  
> -set(CMAKE_REQUIRED_DEFINITIONS -D_POSIX_C_SOURCE=199309L)
> +set(CMAKE_REQUIRED_DEFINITIONS _POSIX_C_SOURCE=199309L)
>  check_symbol_exists(clock_gettime "time.h" CLOCK_GETTIME_IS_BUILT_IN)
>  set(${CMAKE_FIND_PACKAGE_NAME}_FOUND ${CLOCK_GETTIME_IS_BUILT_IN})
> ```

Here is the minimal diff to fix the error for CMake older thank 3.26:
```diff
--- a/cmake/FindClockGettime.cmake
+++ b/cmake/FindClockGettime.cmake
@@ -34,7 +34,7 @@ if(${CMAKE_FIND_PACKAGE_NAME}_FOUND)
   if(NOT TARGET POSIX::clock_gettime)
     add_library(POSIX::clock_gettime INTERFACE IMPORTED)
     set_target_properties(POSIX::clock_gettime PROPERTIES
-      INTERFACE_COMPILE_DEFINITIONS "${CMAKE_REQUIRED_DEFINITIONS}"
+      INTERFACE_COMPILE_DEFINITIONS _POSIX_C_SOURCE=199309L
       INTERFACE_LINK_LIBRARIES "${CMAKE_REQUIRED_LIBRARIES}"
     )
   endif()
```

furszy:
Isn't that going to make `CMAKE_REQUIRED_DEFINITIONS` unused? the `set(CMAKE_REQUIRED_DEFINITIONS -D_POSIX_C_SOURCE=199309L)` line.

hebasto:
> Isn't that going to make `CMAKE_REQUIRED_DEFINITIONS` unused? the `set(CMAKE_REQUIRED_DEFINITIONS -D_POSIX_C_SOURCE=199309L)` line.

It is still used by [`check_symbol_exists()`](https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html).

src/tests_common.h
```diff
@@ -17,25 +17,38 @@
...
-    return (int64_t)tv.tv_nsec / 1000 + (int64_t)tv.tv_sec * 1000000LL;
-#else
-    struct timeval tv;
-    gettimeofday(&tv, NULL);
-    return (int64_t)tv.tv_usec + (int64_t)tv.tv_sec * 1000000LL;
+static int64_t gettime_us(void) {
+#if defined(_WIN32)
+
+    LARGE_INTEGER freq, counter;
+    QueryPerformanceFrequency(&freq);
```
furszy, line 30: Per https://learn.microsoft.com/en-us/windows/win32/api/profileapi/nf-profileapi-queryperformancefrequency, `freq` should be static and initialized only once. The function's description says:

> The frequency of the performance counter is fixed at system boot and is consistent across all processors. Therefore, the frequency need only be queried upon application initialization, and the result can be cached.

Raimo33:
> Have you considered `GetProcessTimes` or `GetThreadTimes` 

I have, but they lack precision.

src/tests_common.h
Raimo33: I know, it's not mandatory but recommended. It is redundant to initialize it every time but I found no other solution since we are not in an OOP environment. and using an if statement to initialize it only the first time doesn't seem optimal.

furszy:
> > Have you considered `GetProcessTimes` or `GetThreadTimes`
> 
> I have, but they lack precision.

That's interesting. What do you think about the semantical difference between the Windows and POSIX approaches?
It seems to me that we should at least mention in the docs that results aren't comparable across platforms.

src/tests_common.h
furszy: I see. Using an `init_time(void)` function could work in that case, but it would require every binary to call this method early in every program’s execution. All good anyway.

Another nit; `QueryPerformanceFrequency` returns 0 if the call fails, so it would be good to handle that as well.

Raimo33: > Another nit; `QueryPerformanceFrequency` returns 0 if the call fails, so it would be good to handle that as well.

my reasoning Is the same here. It would be one extra branch. and a failure in the clock is not a huge deal.

Raimo33:
>  What do you think about the semantical difference between the Windows and POSIX approaches? It seems to me that we should at least mention in the docs that results aren't comparable across platforms.

We added print statements for that exact reason. On windows there simply isn't a reliable way to get per-process time.

hebasto:
Here is a [branch](https://github.com/hebasto/secp256k1/commits/pr1732/1215.rb/) that [passes](https://github.com/hebasto/secp256k1/actions/runs/20249638379) CI.

Raimo33:
rebased and applied @hebasto Makefile fixes

Issue #1621, opened 2024-10-21
Handle invalid inputs consistently with respect to constant-time


jonasnick:
Right now, we have two different ways of handling invalid inputs in constant-time functions:

1. The function is constant-time, even for invalid inputs. For example, `secp256k1_schnorrsig_sign_internal` continues with the signing procedure even if `keypair_load` fails (which happens when the keypair is unitialized):
   ```
   ret &= secp256k1_keypair_load(ctx, &sk, &pk, keypair);
   ...
   return ret;
   ```
   This works because even if `secp256k1_keypair_load` fails, valid values (sk=1, pk=G) are returned.
2. The function is constant-time only for valid inputs. For example, in `secp256k1_musig_partial_sign`, we have
   ```
   if (!secp256k1_keypair_load(ctx, &sk, &keypair_pk, keypair)) {
       secp256k1_musig_partial_sign_clear(&sk, k);
       return 0;
   }
   ```
   This works because the return value of `keypair_load` is declassified.

I think we should make functions only constant-time with respect to valid inputs. This leads to more readable and maintainable code (due to fewer `ret &=`). Calling functions with invalid inputs (such as an unitialized keypair) should never happen outside of development.

Whatever version we're choosing, we should document it in CONTRIBUTING.md.

real-or-random:
> I think we should make functions only constant-time with respect to valid inputs. This leads to more readable and maintainable code

Indeed, and https://github.com/bitcoin-core/secp256k1/pull/1735 shows that insisting on remaining constant-time even for invalid inputs may worsen our defenses against leaving secret values on the stack.

theStack:
> I think we should make functions only constant-time with respect to valid inputs. This leads to more readable and maintainable code (due to fewer ret &=).

I agree, opened #1919 to get the ball rolling on this one.

PR #1735, opened 2025-09-04
musig: Invalidate secnonce in secp256k1_musig_partial_sign

john-moffett:
Replace `memset` which can still be optimized out as `secnonce` isn't read later in this function. The API makes it clear the callee is responsible for it, so we need to assure it's cleared properly.

### DIFF
--- a/src/modules/musig/session_impl.h
+++ b/src/modules/musig/session_impl.h
@@ -385,10 +385,10 @@ static void secp256k1_nonce_function_musig(secp256k1_scalar *k, const unsigned c
         secp256k1_scalar_set_b32(&k[i], buf, NULL);
 
         /* Attempt to erase secret data */
-        secp256k1_memclear(buf, sizeof(buf));
+        secp256k1_memclear_explicit(buf, sizeof(buf));
         secp256k1_sha256_clear(&sha_tmp);
     }
-    secp256k1_memclear(rand, sizeof(rand));
+    secp256k1_memclear_explicit(rand, sizeof(rand));
     secp256k1_sha256_clear(&sha);
 }
 
@@ -518,7 +518,7 @@ int secp256k1_musig_nonce_gen_counter(const secp256k1_context* ctx, secp256k1_mu
     if (!secp256k1_musig_nonce_gen_internal(ctx, secnonce, pubnonce, buf, seckey, &pubkey, msg32, keyagg_cache, extra_input32)) {
         return 0;
     }
-    secp256k1_memclear(seckey, sizeof(seckey));
+    secp256k1_memclear_explicit(seckey, sizeof(seckey));
     return 1;
 }
 
@@ -679,7 +679,7 @@ int secp256k1_musig_partial_sign(const secp256k1_context* ctx, secp256k1_musig_p
     ret = secp256k1_musig_secnonce_load(ctx, k, &pk, secnonce);
     /* Set nonce to zero to avoid nonce reuse. This will cause subsequent calls
      * of this function to fail */
-    memset(secnonce, 0, sizeof(*secnonce));
+    secp256k1_memzero_explicit(secnonce, sizeof(*secnonce));
     if (!ret) {
         secp256k1_musig_partial_sign_clear(&sk, k);
         return 0;
--- a/src/ecmult_gen_impl.h
+++ b/src/ecmult_gen_impl.h
@@ -277,8 +277,8 @@ static void secp256k1_ecmult_gen(const secp256k1_ecmult_gen_context *ctx, secp25
     /* Cleanup. */
     secp256k1_fe_clear(&neg);
     secp256k1_ge_clear(&add);
-    secp256k1_memclear(&adds, sizeof(adds));
-    secp256k1_memclear(&recoded, sizeof(recoded));
+    secp256k1_memclear_explicit(&adds, sizeof(adds));
+    secp256k1_memclear_explicit(&recoded, sizeof(recoded));
 }
 
 /* Setup blinding values for secp256k1_ecmult_gen. */
@@ -310,7 +310,7 @@ static void secp256k1_ecmult_gen_blind(secp256k1_ecmult_gen_context *ctx, const 
     VERIFY_CHECK(seed32 != NULL);
     memcpy(keydata + 32, seed32, 32);
     secp256k1_rfc6979_hmac_sha256_initialize(&rng, keydata, 64);
-    secp256k1_memclear(keydata, sizeof(keydata));
+    secp256k1_memclear_explicit(keydata, sizeof(keydata));
 
     /* Compute projective blinding factor (cannot be 0). */
     secp256k1_rfc6979_hmac_sha256_generate(&rng, nonce32, 32);
@@ -331,7 +331,7 @@ static void secp256k1_ecmult_gen_blind(secp256k1_ecmult_gen_context *ctx, const 
     secp256k1_ge_set_gej(&ctx->ge_offset, &gb);
 
     /* Clean up. */
-    secp256k1_memclear(nonce32, sizeof(nonce32));
+    secp256k1_memclear_explicit(nonce32, sizeof(nonce32));
     secp256k1_scalar_clear(&b);
     secp256k1_gej_clear(&gb);
     secp256k1_fe_clear(&f);
--- a/src/field_impl.h
+++ b/src/field_impl.h
@@ -19,7 +19,7 @@
 #endif
 
 SECP256K1_INLINE static void secp256k1_fe_clear(secp256k1_fe *a) {
-    secp256k1_memclear(a, sizeof(secp256k1_fe));
+    secp256k1_memclear_explicit(a, sizeof(secp256k1_fe));
 }
 
 SECP256K1_INLINE static int secp256k1_fe_equal(const secp256k1_fe *a, const secp256k1_fe *b) {
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -337,11 +337,11 @@ static void secp256k1_ge_set_infinity(secp256k1_ge *r) {
 }
 
 static void secp256k1_gej_clear(secp256k1_gej *r) {
-    secp256k1_memclear(r, sizeof(secp256k1_gej));
+    secp256k1_memclear_explicit(r, sizeof(secp256k1_gej));
 }
 
 static void secp256k1_ge_clear(secp256k1_ge *r) {
-    secp256k1_memclear(r, sizeof(secp256k1_ge));
+    secp256k1_memclear_explicit(r, sizeof(secp256k1_ge));
 }
 
 static int secp256k1_ge_set_xo_var(secp256k1_ge *r, const secp256k1_fe *x, int odd) {
--- a/src/hash_impl.h
+++ b/src/hash_impl.h
@@ -172,7 +172,7 @@ static void secp256k1_sha256_initialize_tagged(secp256k1_sha256 *hash, const uns
 }
 
 static void secp256k1_sha256_clear(secp256k1_sha256 *hash) {
-    secp256k1_memclear(hash, sizeof(*hash));
+    secp256k1_memclear_explicit(hash, sizeof(*hash));
 }
 
 static void secp256k1_hmac_sha256_initialize(secp256k1_hmac_sha256 *hash, const unsigned char *key, size_t keylen) {
@@ -200,7 +200,7 @@ static void secp256k1_hmac_sha256_initialize(secp256k1_hmac_sha256 *hash, const 
         rkey[n] ^= 0x5c ^ 0x36;
     }
     secp256k1_sha256_write(&hash->inner, rkey, sizeof(rkey));
-    secp256k1_memclear(rkey, sizeof(rkey));
+    secp256k1_memclear_explicit(rkey, sizeof(rkey));
 }
 
 static void secp256k1_hmac_sha256_write(secp256k1_hmac_sha256 *hash, const unsigned char *data, size_t size) {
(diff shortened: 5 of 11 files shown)


real-or-random:
> Replace `memset` which can still be optimized out as `secnonce` isn't read later in this function.

Why do you think it can be optimized out? I think it can be optimized out if the compiler can prove that no other code will read it. That seems unlikely, as the call is an API call. But if the compiler can prove this, then it's indeed fine to optimize it out? Or are you saying we should make absolutely sure that it's cleared out, even if the caller won't read it? Perhaps.

src/modules/musig/session_impl.h
```diff
@@ -677,9 +677,9 @@ int secp256k1_musig_partial_sign(const secp256k1_context* ctx, secp256k1_musig_p
     ARG_CHECK(secnonce != NULL);
     /* Fails if the magic doesn't match */
     ret = secp256k1_musig_secnonce_load(ctx, k, &pk, secnonce);
-    /* Set nonce to zero to avoid nonce reuse. This will cause subsequent calls
-     * of this function to fail */
-    memset(secnonce, 0, sizeof(*secnonce));
+    /* Clear and invalidate nonce to avoid nonce reuse. This will cause subsequent 
+     * calls of this function to fail */
+    secp256k1_musig_secnonce_invalidate(ctx, secnonce, 1);
```
real-or-random: This replaces one dead store with another, so if the compiler can optimize out the one, it could also optimize out the other. If we want to avoid that the store is optimized out, we'll need `secp256k1_memclear`.

real-or-random:
cc @jonasnick who wrote this module

jonasnick (commented):
Thanks @john-moffett. I think it's an improvement to use `secp256k1_memclear` in this case. 

> But if the compiler can prove this, then it's indeed fine to optimize it out?

I think in this case it still makes sense to clear it to try to avoid having secret data in memory that we don't use anymore. This is also the argument for clearing local secret variables at the end of functions.

john-moffett:
> Why do you think it can be optimized out? I think it can be optimized out if the compiler can prove that no other code will read it. That seems unlikely, as the call is an API call.

I agree that it's unlikely to be optimized out, and I confirmed on my setup (arm64 macOS Clang 17.0.0) that even with -O3 and full LTO that it wasn't. However, that's not guaranteed since it's possible a different compiler (or later version) can prove no later reads, so I figured it's just best practice to use a wiper much less likely to be DSE'd. 

> But if the compiler can prove this, then it's indeed fine to optimize it out? Or are you saying we should make absolutely sure that it's cleared out, even if the caller won't read it? 

The latter. If they assumed the callee wiped the memory (as is [promised](https://github.com/bitcoin-core/secp256k1/blob/03fb60ad2e38ae9415e64bf5ba523a10100d4c3c/include/secp256k1_musig.h#L484-L486) in the API comment), they might not bother themselves and it could be left in caller memory.

src/modules/musig/session_impl.h
```diff
@@ -678,8 +678,12 @@ int secp256k1_musig_partial_sign(const secp256k1_context* ctx, secp256k1_musig_p
     /* Fails if the magic doesn't match */
     ret = secp256k1_musig_secnonce_load(ctx, k, &pk, secnonce);
     /* Set nonce to zero to avoid nonce reuse. This will cause subsequent calls
-     * of this function to fail */
-    memset(secnonce, 0, sizeof(*secnonce));
+     * of this function to fail. Declassify as memclear marks the bytes
+     * as undefined. */
+    secp256k1_memclear(secnonce, sizeof(*secnonce));
+    secp256k1_declassify(ctx, secnonce->data, sizeof(secp256k1_musig_secnonce_magic));
+    secp256k1_declassify(ctx, &secnonce->data[68], 64);
```
jonasnick: We do not need to declassify because we do not use `secnonce` anymore after this.

jonasnick:
@john-moffett Can you remove the remaining `declassify` call as well?

src/modules/musig/session_impl.h
john-moffett: The tests [reuse](https://github.com/bitcoin-core/secp256k1/blob/03fb60ad2e38ae9415e64bf5ba523a10100d4c3c/src/modules/musig/tests_impl.h#L404-L405) it inentioanlly and valgrind flags it and gives an error on VERIFY builds. 

I can declassify in the tests if that's more natural.

src/modules/musig/tests_impl.h
```diff
@@ -400,7 +400,9 @@ static void musig_api_tests(void) {
 
     memcpy(&secnonce_tmp, &secnonce[0], sizeof(secnonce_tmp));
     CHECK(secp256k1_musig_partial_sign(CTX, &partial_sig[0], &secnonce_tmp, &keypair[0], &keyagg_cache, &session) == 1);
-    /* The secnonce is set to 0 and subsequent signing attempts fail */
+    /* The secnonce is set to 0 and subsequent signing attempts fail.
+     * Redefine memory first as secp256k1_musig_partial_sign securely clears and undefines the bytes.*/
+    SECP256K1_CHECKMEM_DEFINE(&secnonce_tmp, sizeof(secnonce_tmp));
```
jonasnick:
```suggestion
    /* The secnonce is set to 0 and subsequent signing attempts fail.
     * musig_partial_sign calls secp256k1_memclear to clear the secnonce, which
     * marks it as undefined for memory checkers. So we need to explicitly mark
     * secnonce_tmp as defined before we can compare its value. */
    SECP256K1_CHECKMEM_DEFINE(&secnonce_tmp, sizeof(secnonce_tmp));
```

src/modules/musig/session_impl.h
jonasnick: Ok I understand now. I missed that `secp256k1_memclear` uses the UNDEFINED macro (in VERIFY builds).

I think it's indeed preferrable to explicitly "redefine" the macro in the test instead of declassifying in partial_sign. I think the purpose of UNDEFINEing is to catch accidental use of the buffer and there's no legitimate reason to read the secnonce buffer except in the test that checks that it was cleared correctly.

src/modules/musig/tests_impl.h
john-moffett: Thanks! A clearer explanation is indeed warranted.

jonasnick (approved):
ACK d872a08d6dab8157259328365f73b6d7d8dfa5d3

real-or-random:
I agree it's better to prevent the compiler from optimizing out that store. 

I'm not sure if this PR adds another code smell. I know it's me who suggested `secp256k1_memclear` above. In the rest of the code base, we use `secp256k1_memclear` to nuke memory contents. This is true here as well, but here we want the additional guarantee that the memory is overwritten by *zero bytes*. That fits the current implementation but not the documented contract of `secp256k1_memclear`. An implementation that overwrites everything with 0x42 is equally correct. We simply picked 0x00 because it appears to be faster. I doubt that will change in the future, but I still think we should avoid the contract violation.

### What could we do instead?
We could use `secp256k1_memclear` but rename `secp256k1_memclear` to `secp256k1_memzero` and change its docs to guarantee it writes zeros. There's nothing wrong with adding that guarantee, but I don't think the result will be much cleaner if we only do this. We use `SECP256K1_CHECKMEM_UNDEFINE` inside this function for a reason. If the goal is to nuke a buffer, then there should be no legitimate reason to read the buffer again (as @jonasnick pointed out above). So nuking and zeroing are different concerns, and neither is a strict subset of the other. What we want in this specific case of zeroing the secnonce is simply both.

We could instead call `secp256k1_memzero` (after renaming) and then call `SECP256K1_CHECKMEM_DEFINE`. That's totally fine, and that's my suggestion. But we really should call `SECP256K1_CHECKMEM_DEFINE` inside the function and not only in the tests. If we provide the API guarantee that the secnonce is zeroed, then it's totally legitimate that the API user reads (if only in a test). And we don't want the user to get a Valgrind false positive when doing that!

We could also call `secp256k1_memclear` and then `memset(..., 0, ...)`. That's also good. It may be a performance hit, but in any case a tiny one. (On Linux, the compiler should not be able to remove the second `memset` due to the memory barrier. But I wouldn't care too much.)


edit: @john-moffett Sorry that this PR escalated quickly given that the issue is seemingly simple...

jonasnick:
> If we provide the API guarantee that the secnonce is zeroed, then it's totally legitimate that the API user reads (if only in a test). And we don't want the user to get a Valgrind false positive when doing that!

That's a good point that I hadn't considered.

An alternative to your suggestion is to have both a `memclear` and `memzero` function, where the latter differs from the former by guaranteeing that the memory is overwritten with 0s and not calling `CHECKMEM_UNDEFINE`. We'd `clear` memory if we never read it  again (and therefore do not care what it's being overwritten with) and we `zero` memory if we care what it's being overwritten with because it might be read again. I think this is cleaner because we wouldn't have to call `CHECKMEM_DEFINE` in  `musig_partial_sign` (which seems like a code smell in and of itself).

john-moffett:
> An alternative to your suggestion is to have both a `memclear` and `memzero` function, where the latter differs from the former by guaranteeing that the memory is overwritten with 0s and not calling `CHECKMEM_UNDEFINE`

I was going to make the same suggestion. But is there any value in just reusing `memczero` and making it non-elidable in the same way `memclear` is (though `SecureZeroMemory` would have to be replaced with something suitable for conditional erase)? As a separate matter, is it worth making `memczero` non-elidable anyway? Perhaps not given its current usage.

>  @john-moffett Sorry that this PR escalated quickly given that the issue is seemingly simple..

No problem! I should've been aware of the issue from the start, but I've learned a lot, so I'm happy.

At this point, I think the `memclear` and (new) `memzero` separation is the cleanest.

john-moffett:
I suppose [this usage](https://github.com/bitcoin-core/secp256k1/blob/03fb60ad2e38ae9415e64bf5ba523a10100d4c3c/src/modules/musig/session_impl.h#L493) of `memczero` suffers from the same potential issue I initially raised in this PR, so maybe `memczero` ought to be hardened? I agree that it's unlikely that any current compiler would actually DSE any of the calls we've been discussing, but is the possibility enough to warrant a change?

real-or-random:
> An alternative to your suggestion is to have both a `memclear` and `memzero` function, where the latter differs from the former by guaranteeing that the memory is overwritten with 0s and not calling `CHECKMEM_UNDEFINE`

Ok, sure, that's another option. Concept ACK .

I was considering that option too, but I didn't mention it because I had the feeling that the existence of two functions may confuse future devs. But now that I think about it, I guess it's fine. Perhaps then we can make it clear from the naming that both variants protect against optimization, e.g., by adding an `_explicit` (inspired by `memzero_explicit` in the Linux kernel) or `_secure` suffix.

> But is there any value in just reusing `memczero` and making it non-elidable in the same way `memclear` is (though `SecureZeroMemory` would have to be replaced with something suitable for conditional erase)?

> As a separate matter, is it worth making `memczero` non-elidable anyway? Perhaps not given its current usage.

That's an interesting point you raise.

Regarding its usage: I think it was initially introduced to clear the `pubkey` when the secret key is invalid, but without leaking the fact that it is. But protecting the fact that it is already "above and beyond" and just defense-in-depth. And perhaps a bit questionable, though it won't hurt. 

But all its three usages in `musig/session_impl.h` (two of them wrapped by `_musig_secnonce_invalidate`) are for clearing secrets, and this is where it gets interesting. In this case, I believe there's a tension here between protecting against two kinds of side-channels, namely timing/memory access (protecting the secrecy of the `flag`) vs. someone stealing secrets from a memory dump such as a crash dump (protecting the secrecy of the memory region). 

How would a function look like that is both non-elidable and protects the `flag`? I have no idea, exactly due to platform-specific APIs such as `SecureZeroMemory`. That's the right way(TM) to clear memory on Windows, and I'm not at all convinced that it's a good idea to replace it, e.g., by a handrolled version. Or by falling back to the "volatile" function pointer trick, which is supposed to be portable, but who knows what compilers will do in the future and what they will do when they encounter "volatile", which pretty much has implementation-defined semantics. The nice thing about `SecureMemoryZero` is really that we get a guarantee from MSVC not to elide it. So I think `memczero` should stay entirely separate from `memclear` and `memzero`, at least for now.

So if we can't have the best of both worlds, we need to decide what protection we want. Currently, whenever we use `memczero`, we implicitly favor the protection of the flag, and this seems to be a questionable choice for its usages in `musig/session_impl.h` (or in general when clearing out secrets). I'd much rather give the attacker that one bit whether the secret key was invalid vs. taking the risk to keep the entire secnonce (or the entire nonce seed) somewhere on the stack. If we believe that this PR is a good idea, i.e., the clearing of the secnonce should better be non-elidable, then I think we should arrive at the same conclusion for the other three uses. And in particular this use here, where the `flag` is a constant anyway: https://github.com/bitcoin-core/secp256k1/blob/03fb60ad2e38ae9415e64bf5ba523a10100d4c3c/src/modules/musig/session_impl.h#L485

But this may be a larger discussion, and I'd like to hear @jonasnick's opinion on it. So my suggestion is to leave `memczero` for a follow-up PR. This PR here could then just add `memzero` as a variant of `memclear`, perhaps add a suffix to the names, and use it in `_musig_partial_sign`.

jonasnick:
>  we use memczero, we implicitly favor the protection of the flag, and this seems to be a questionable choice for its usages in musig/session_impl.h. [...] And in particular this use here, where the flag is a constant anyway

I agree. We have issue #1621 pointing out the inconsistencies in handling constant-timeness with respect to invalid arguments in the library.

theStack:
post-merge ACK 399b582a5f7422bac3b1707d34ffad80e072f796

PR #1737, opened 2025-09-05
doc: mention ctx requirement for `_ellswift_create` (not secp256k1_context_static)

theStack:
Public functions that require a context for generator point multiplication (i.e. `ctx->ecmult_gen_ctx` is built) usually denote this in the API header by mentioning to not use `secp256k1_context_static`, so add this for `_ellswift_create` as well. This seems the only instance where this is missing currently, see
```
$ git grep ecmult_gen_context_is_built
$ git grep ctx:.*context_static
```
(note that in the musig and schnorr modules, two public functions for nonce generation / signing map to a single internal function where the context requirement is checked).

Noted while reviewing #1698.

### DIFF
--- a/include/secp256k1_ellswift.h
+++ b/include/secp256k1_ellswift.h
@@ -130,7 +130,7 @@ SECP256K1_API int secp256k1_ellswift_decode(
  *
  *  Returns: 1: secret was valid, public key was stored.
  *           0: secret was invalid, try again.
- *  Args:    ctx:        pointer to a context object
+ *  Args:    ctx:        pointer to a context object (not secp256k1_context_static)
  *  Out:     ell64:      pointer to a 64-byte array to receive the ElligatorSwift
  *                       public key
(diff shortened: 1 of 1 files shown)


sipa:
Why "(not secp256k1_context_static)"?

jonasnick:
@sipa Because `ellswift_create` requires the `ecmult_gen_context` to be built, which is not the case for the static context. The wording "(not secp256k1_context_static)" is consistent with the other functions that have the same requirement on the context.

josibake (commented):
ACK https://github.com/bitcoin-core/secp256k1/pull/1737/commits/806de38bfc24b481c5858f1df011eb909208a674

Did a quick grep to verify this wording matches the wording used in other places, and in general agree we should consistently document this for functions that won't work with the static context.

sipa:
ACK 806de38bfc24b481c5858f1df011eb909208a674

PR #1740, opened 2025-09-09
doc: update release notes version to latest v0.7.0

l0rinc:
While reading the docs I noticed the latest isn't shown in the docs.


l0rinc:
The example states that it's the latest

real-or-random:
> The example states that it's the latest

What about just dropping the word "latest" then? We could also add instruction on how to list all tags (`git tag`).

PR #1741, opened 2025-09-09
doc: clarify API doc of `secp256k1_ecdsa_recover` return value

### LINKED ISSUE #1718: Clarify API doc of ecdsa_recover return value
The API doc for `ecdsa_recover` states the following:

```C
/** Recover an ECDSA public key from a signature.
 *
 *  Returns: 1: public key successfully recovered (which guarantees a correct signature).
 *           0: otherwise.
```

After reading this, one may assume that when `ecdsa_recover` successfully recovers a pubkey, `ecdsa_verify` on the same signature, message and public key will never fail. Or more precisely, the assertion in the following code snippet never fails:

```C
secp256k1_ecdsa_signature sig;
secp256k1_ecdsa_recoverable_signature rsig;

/* Set rsig to something */
/* ... */

if (secp256k1_ecdsa_recover(ctx, pubkey, rsig, msg)) {
    secp256k1_ecdsa_recoverable_signature_convert(ctx, &sig, &rsig);
    assert(secp256k1_ecdsa_verify(CTX, &sig, msg, &pubkey));

}
```

However, this is **not** true due to "s-malleability" of ECDSA signatures: `ecdsa_verify` requires low-s while `ecdsa_recover` does not. Thus, if we add
```
    secp256k1_ecdsa_signature_normalize(CTX, &sig, &sig);
```
after the `ecdsa_recoverable_signature_convert` call in above code snipped, then the assertion *never* fails.

In other words, `ecdsa_recover` returning 1 *does* guarantee a correct signature, but `ecdsa_verify` passes only after the signature is normalized. We should document this fact more clearly.

Thanks to @erickcestari for reporting this!

theStack:
This PR aims to fix #1718. I thought it's nice to provide an usage example, but not sure if we want actual code snippets in the API header (it's also incomplete, as the declarations are missing), so an alternative might be to drop it, or add a "(see example)" reference in the future in case #1714 gets in.


include/secp256k1_recovery.h
```diff
@@ -92,7 +92,22 @@ SECP256K1_API int secp256k1_ecdsa_sign_recoverable(
...
 /** Recover an ECDSA public key from a signature.
  *
- *  Returns: 1: public key successfully recovered (which guarantees a correct signature).
+ *  Note that successful public key recovery implies a correct ECDSA signature
+ *  in a mathematical sense, but it is only guaranteed to also pass verification
+ *  with `secp256k1_ecdsa_verify` if it is in lower-S form. For verifying
+ *  recoverable signatures that were not created with libsecp256k1 (or might have
+ *  been malleated), it is thus necessary to normalize the signature using
+ *  `secp256k1_ecdsa_signature_normalize` between conversion and verification.
+ *  A typical usage might look like this:
```
jonasnick: I think it'd be better if the docs pointed out that this is not typical usage. In typical usage you only care whether the signature is valid for the recovered public key. So you just need run `ecdsa_recover` and do _not_ need to call `ecdsa_verify`.

Maybe there are cases where this is necessary (e.g., you have to hand over the signature and public key to a verify that doesn't support `ecdsa_recover`, like the Bitcoin consensus protocol).

include/secp256k1_recovery.h
```diff
@@ -92,7 +92,22 @@ SECP256K1_API int secp256k1_ecdsa_sign_recoverable(
 
 /** Recover an ECDSA public key from a signature.
  *
- *  Returns: 1: public key successfully recovered (which guarantees a correct signature).
+ *  Note that successful public key recovery implies a correct ECDSA signature
+ *  in a mathematical sense, but it is only guaranteed to also pass verification
+ *  with `secp256k1_ecdsa_verify` if it is in lower-S form. For verifying
```
jonasnick: "correct ECDSA signature in a mathematical sense" sounds a bit imprecise. How about the following wording, which avoids that term entirely?

```
Successful public key recovery guarantees that the signature, when normalized,
passes `secp256k1_ecdsa_verify`. Thus, explicit verification is not necessary.

A recoverable signature converted to a normal signature may not pass
`secp256k1_ecdsa_verify` if it is not normalized. If a normalized signature is
required, call `secp256k1_ecdsa_signature_normalize` after
`secp256k1_ecdsa_recover`.
```

theStack: Took that suggestion, and set you as commit author accordingly. As tiny-nit, it's slightly unfortunate that "normal" and "normalized" sound so similar, but not sure how to avoid it (maybe it's not a big deal).

include/secp256k1_recovery.h
theStack: Makes sense. My assumption when writing this was that most users would very likely want to enforce lower-S in order to avoid malleability, but that was more based on gut feeling than actual research (seems BOLT11 now explicitly allows low- and high-S signatures if pubkey recovery is used: https://github.com/lightning/bolts/pull/1284).

include/secp256k1_recovery.h
real-or-random: > Took that suggestion, and set you as commit author accordingly. As tiny-nit, it's slightly unfortunate that "normal" and "normalized" sound so similar, but not sure how to avoid it (maybe it's not a big deal).

Stumbled upon this, too. One fix is s/normal/non-recoverable. Another one is to spell out the types explicitly:


Here's an improved version. This uses a middle way. It says "non-recoverable" but mentions `secp256k1_ecdsa_recoverable_signature_convert` so the reader can look up the exact types.

```
Successful public key recovery guarantees that the signature, after normalization,
passes `secp256k1_ecdsa_verify`. Thus, explicit verification is not necessary.

However, a non-recoverable signature converted from a recoverable signature
(using `secp256k1_ecdsa_recoverable_signature_convert`) is not guaranteed
to be normalized and thus not guaranteed to pass `secp256k1_ecdsa_verify`
(even if it passes `secp256k1_ecdsa_recover`). If a normalized signature is
required, call `secp256k1_ecdsa_signature_normalize` after `secp256k1_ecdsa_recoverable_signature_convert`.
```

theStack: SGTM, taken.

jonasnick: "a non-recoverable signature [...] (even if it passes `secp256k1_ecdsa_recover`)" may be a bit confusing because a non-recoverable signature cannot pass `secp256k1_ecdsa_recover`.

How about this:

```
However, a recoverable signature that successfully passes `secp256k1_ecdsa_recover`,
when converted to a non-recoverable signature (using
`secp256k1_ecdsa_recoverable_signature_convert`), is not guaranteed to be
normalized and thus not guaranteed to pass `secp256k1_ecdsa_verify`.
```

real-or-random: > "a non-recoverable signature [...] (even if it passes `secp256k1_ecdsa_recover`)" may be a bit confusing because a non-recoverable signature cannot pass `secp256k1_ecdsa_recover`.

Good point.

theStack: Updated again with the suggestion.

jonasnick (approved):
ACK 7321bdf27bdc6a68875c61510c1a0d9ec4b0b7e3

Thanks @theStack

PR #1734, opened 2025-09-04
Introduce (mini) unit test framework

furszy:
Context:
Currently, all tests run single-threaded sequentially and the library lacks the ability to specify which test (or group of tests) you would like to run. This is not only inconvenient as more tests are added but also time consuming during development and affects downstream projects that may want to parallelize the workload (such as Bitcoin-Core CI).

PR Goal:
Introduce a lightweight, extensible unit test framework with no dynamic memory allocations, providing a structured way to register, execute, and report tests. The framework supports named command-line arguments in `-key=value` form, parallel test execution across multiple worker processes, granular test selection (selecting tests either by name or by module name), and time accumulation reports.

The introduced framework supports:
* `-jobs=<num>`: distribute tests across multiple worker processes (default: sequential if 0).
* `-target=<name>` or `-t=<name>`: run only specific tests by name; can be repeated to select multiple tests.
* `-seed=<hex>`: set a specific RNG seed (defaults to random if unspecified).
* `-iter=<n>`: specify the number of iterations.

Beyond these features, the idea is to also make future developments smoother, as adding new tests require only a single entry in the central test registry, and new command-line options can be introduced easily by extending the framework’s `parse_arg()` function.

Compatibility Note:
The framework continues accepting the two positional arguments previously supported (iterations and seed), ensuring existing workflows remain intact.

Testing Notes:
Have fun. You can quickly try it through `./tests -j=<workers_num>` for parallel execution  or `./tests -t=<test_name>` to run a specific test (check the test registry on `tests.c` for test names).

Extra Note:
I haven't checked the exhaustive tests file so far, but I will soon. For now, this only runs all tests declared in `tests.c`.

Extra Extra Note:
I’d be happy to hear your experience running this on different systems. I’ve only tested it on a very small subset of environments so far.

PR #1724, opened 2025-08-14
Make files mostly self-contained

real-or-random:
This is a set of commits, each one trivial, that reorganizes code so that all C files are *mostly* self-contained. 

These changes are merely fixes that make the files adhere to our current implicit style, which I'd describe as:
  - Every C file (library, tests.c, ...) is a single translation unit ("unit build"), except that the libsepc256k1_precomputed is separate to speed up compilation.
  - Internal "modules" are organized in headers (`.h`) and implementation files (`_impl.h`). Symbols not present in the header are private to that module.
  - Internal "modules" should include the headers of all other modules they use.  
  - `secp256k1.c` includes all `_impl.h` files. Other C files include simply `secp256k1.c`, or, if that isn't possible for some reason (e.g., in the `precompute_*.c` files), the individual necessary `_impl.h` files.

So, none of these changes here should be controversial. In particular, I refrain from changing the above rules (e.g., by renaming the `_impl.h` to `.c` as suggested in https://github.com/bitcoin-core/secp256k1/issues/1039).

####  Why "mostly" self-contained?

 After this PR, some errors remain when I run `clang src/*.c` and `clang src/*.h` on my machine. But all of them are special cases: 
 - `ctime_tests.c` refuses to be compiled without valgrind or msan enabled.
 - `src/field_10x26_impl.h`, `src/int128_struct_impl.h`, and `src/scalar_low_impl.h` are confused because the autodetection logic in the preprocessor  picks field_5x64, native int 128, and scalar_4x64 on my machine. 

<details>
<summary>Full output of "clang src/*.c" and "clang src/*.h` </summary>

```
> clang src/*.c
src/ctime_tests.c:16:4: error: "This tool cannot be compiled without memory-checking interface (valgrind or msan)"
   16 | #  error "This tool cannot be compiled without memory-checking interface (valgrind or msan)"
      |    ^
1 error generated.
```

```
> clang src/*.h -ferror-limit=100
src/field_10x26_impl.h:265:21: warning: incompatible pointer types initializing 'const uint32_t *' (aka 'const unsigned int *') with an expression of type 'const uint64_t[5]' (aka 'const unsigned long[5]') [-Wincompatible-pointer-types]
  265 |     const uint32_t *t = a->n;
      |                     ^   ~~~~
[...]

PR #1743, opened 2025-09-11
Fix headers in `testrand_impl.h`

hebasto:
While reviewing https://github.com/bitcoin-core/secp256k1/pull/1734 and [suggesting](https://github.com/bitcoin-core/secp256k1/pull/1734#discussion_r2340384653) keeping the new test miniframework as a separate translation unit, I noticed a couple of issues with the headers included in `testrand_impl.h`. In my [approach](https://github.com/hebasto/secp256k1/commits/pr1734/0911.2c.alt/), these lead to compiler warnings and linker errors. This PR fixes them.

These changes are reasonable on their own.

### DIFF
--- a/src/testrand_impl.h
+++ b/src/testrand_impl.h
@@ -10,9 +10,10 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
+#include <time.h>
 
 #include "testrand.h"
-#include "hash.h"
+#include "hash_impl.h"
 #include "util.h"
 
 static uint64_t secp256k1_test_state[4];


src/testrand_impl.h
```diff
@@ -10,9 +10,10 @@
 #include <stdint.h>
 #include <stdio.h>
 #include <string.h>
+#include <time.h>
```
real-or-random, line 13: That one is also in https://github.com/bitcoin-core/secp256k1/pull/1724. (Want to review it? :) )

src/testrand_impl.h
```diff
@@ -10,9 +10,10 @@
...
 
 #include "testrand.h"
-#include "hash.h"
+#include "hash_impl.h"
```
real-or-random, line 16: Why do we need this?

src/testrand_impl.h
hebasto: > Want to review it? :)

It is next one in my list after https://github.com/bitcoin-core/secp256k1/pull/1734 :)

src/testrand_impl.h
hebasto: As I pointed out in the PR description, when building the new test miniframework (from https://github.com/bitcoin-core/secp256k1/pull/1734) as a separate translation unit, it fails:
```
$ cmake --build build
[1/20] Building C object src/CMakeFiles/unit_test.dir/unit_test.c.o
In file included from /home/hebasto/dev/secp256k1/secp256k1/src/testrand_impl.h:17,
                 from /home/hebasto/dev/secp256k1/secp256k1/src/unit_test.c:31:
/home/hebasto/dev/secp256k1/secp256k1/src/hash.h:19:13: warning: ‘secp256k1_sha256_initialize’ used but never defined
   19 | static void secp256k1_sha256_initialize(secp256k1_sha256 *hash);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/hebasto/dev/secp256k1/secp256k1/src/hash.h:20:13: warning: ‘secp256k1_sha256_write’ used but never defined
   20 | static void secp256k1_sha256_write(secp256k1_sha256 *hash, const unsigned char *data, size_t size);
      |             ^~~~~~~~~~~~~~~~~~~~~~
/home/hebasto/dev/secp256k1/secp256k1/src/hash.h:21:13: warning: ‘secp256k1_sha256_finalize’ used but never defined
   21 | static void secp256k1_sha256_finalize(secp256k1_sha256 *hash, unsigned char *out32);
      |             ^~~~~~~~~~~~~~~~~~~~~~~~~
[18/20] Linking C executable bin/noverify_tests
FAILED: bin/noverify_tests 
: && /usr/bin/cc -O2 -g -Wl,--dependency-file=src/CMakeFiles/noverify_tests.dir/link.d src/CMakeFiles/unit_test.dir/unit_test.c.o src/CMakeFiles/secp256k1_precomputed.dir/precomputed_ecmult.c.o src/CMakeFiles/secp256k1_precomputed.dir/precomputed_ecmult_gen.c.o src/CMakeFiles/noverify_tests.dir/tests.c.o -o bin/noverify_tests   && :
/usr/bin/ld: src/CMakeFiles/unit_test.dir/unit_test.c.o: in function `testrand_seed':
/home/hebasto/dev/secp256k1/secp256k1/src/testrand_impl.h:29:(.text+0x6c2): undefined reference to `secp256k1_sha256_initialize'
/usr/bin/ld: /home/hebasto/dev/secp256k1/secp256k1/src/testrand_impl.h:30:(.text+0x6d6): undefined reference to `secp256k1_sha256_write'
/usr/bin/ld: /home/hebasto/dev/secp256k1/secp256k1/src/testrand_impl.h:31:(.text+0x6e6): undefined reference to `secp256k1_sha256_write'
/usr/bin/ld: /home/hebasto/dev/secp256k1/secp256k1/src/testrand_impl.h:32:(.text+0x6f1): undefined reference to `secp256k1_sha256_finalize'
collect2: error: ld returned 1 exit status
[19/20] Building C object src/CMakeFiles/tests.dir/tests.c.o
ninja: build stopped: subcommand failed.
```

furszy: > As I pointed out in the PR description, when building the new test miniframework (from https://github.com/bitcoin-core/secp256k1/pull/1734) as a separate translation unit, it fails

In that case, you should add the `*_impl.h` include in the binary’s top-level file, not here.
I think it helps to think about the `*_impl.h` files as if they were `.cpp` files: you would never include them directly in any file. You would only include the headers with the public functions/structs and then link the object files to get their implementations. But since we have a single translation unit here, we include them only in the top-level binary.

Side note: we could also remove this include from the unit test side if needed. Could just add a pointer to the RNG init function and done.

real-or-random: > As I pointed out in the PR description, when building the new test miniframework (from #1734) as a separate translation unit, it fails:

That may be true, but without #1734, including just `hash.h` is perfectly correct. Exactly because `hash_impl.h` is included in `tests.c`. That's our "unity-build" convention, as correctly described by @furszy.

hebasto: Right. Including `hash_impl.h` in `unit_tests.c` fixes my branch.

PR #1745, opened 2025-09-12
test: introduce group order byte-array constant for deduplication

theStack:
Several tests currently define their own byte arrays containing the secp256k1 group order $n$ (=[0xfffffffffffffffffffffffffffffffebaaedce6af48a03bbfd25e8cd0364141](https://github.com/bitcoin-core/secp256k1/blob/88be4e8d86773dd6882d32ad0a6cd8892b31eacd/src/ecdsa_impl.h#L20)) in 32-byte big endian representation. This PR deduplicates those by introducing a `secp256k1_group_order_bytes` constant in `testutil.h`. Noticed while reviewing the sending tests of #1698, which also adds another instance (called [`ORDERC`](https://github.com/bitcoin-core/secp256k1/pull/1698/commits/759328ae9163a78379653c2d13b75a55ad77bc84#diff-c444b4b5d4d38da75b0b694c1fe238dd9857ee8e8ca8ecc9acaaf5dca475fa3cR23) there), i.e. including this it would be 5 instances in total (maybe there are even more around that I missed due to insufficient `grep`ing).


src/testutil.h
```diff
@@ -11,6 +11,14 @@
 #include "testrand.h"
 #include "util.h"
 
+/* group order of the secp256k1 curve in 32-byte big endian representation */
+static const unsigned char secp256k1_group_order_bytes[32] = {
```
furszy, line 15: tiny q: why not use capital letters for a constant?
It’s usually the convention to distinguish functions vs constants so we don't have to search for its definition.

theStack: Fair question. It seems that we use capital letters mainly for constants and macros defined with the preprocessor (i.e. `#define ...`), where as actual global constant objects are still in lowercase, see e.g.
```
$ git grep "^static const"
```
vs.
```
$ git grep "^#define"
```
Maybe it's worth it to add that to the style conventions section in CONTRIBUTING.md (and fix the few exceptions that don't follow this conventions, e.g. `SECP256K1_SIGNED{30,62}_ONE`).

furszy: Don’t you feel this is one of those things so widely accepted that it’s strange to have to formally state it? It’s like seeing someone write class members in capital letters.

In any case, it was a tiny question anyway. The upside of this change is that we can script any future change, since everything uses the same variable now.

jonasnick (approved):
ACK 0c91c5604130a56bb1628fd12c56edf60de966b0

PR #1747, opened 2025-09-14
fix: ensure matching ECDH shared secret

mdqst:


mdqst:
> I'm pretty sure that the current ECDH example code produces the same shared secret for both parties. This is tested in CI and I just double checked locally.

this change addresses an inconsistency in our fork where the default ECDH behavior was modified, preventing parties from deriving the same shared secret. explicitly using the ecdh_hash function (the identity hash copying the x-coordinate) ensures interoperability by mandating a standard, deterministic algorithm instead of relying on a potentially altered default. this is a common requirement for protocols needing the raw shared secret.

jonasnick:
The default hash function in libsecp256k1 [always hashed the compressed representation of the point](https://github.com/bitcoin-core/secp256k1/pull/252/files#diff-038202ab5b1c3d61a58d26556c798ac0ab5fc636902af1a5dae6fd5298fbcfe4R27). So you want to change libsecp's example to match your code?

Note that outputting the X-coordinate directly can lead to vulnerabilities because the resulting shared secret is easily distinguishable from uniform randomness (and potentially enables ["Cheon's attack"](https://yongsoosong.github.io/files/papers/DLPwAI.pdf)).

jonasnick:
Now you're outputting the X-coordinates of both participants' public keys (and remove a bunch of documentation comments). Why?

PR #1748, opened 2025-09-16
bench: improve context creation in ECDH benchmark

theStack:
Calling `secp256k1_context_create` with `SECP256K1_FLAGS_TYPE_CONTEXT` seems to be confusing and not strictly API-compliant, as the only allowed (non-deprecated) value is `SECP256K1_CONTEXT_NONE`, even if the former happens to map to the latter currently.

Fix this by not dynamically creating a context in the first place and switch to using the static context, as it is sufficient for this benchmark and presumably matches what the "no capabilities" comment intended back then.

Alternatives are:
* keep the signing context and only fix the name, i.e. s/_FLAGS_TYPE_CONTEXT/_CONTEXT_NONE/
* use `secp256k1_context_static` everywhere directly and get rid of the `ctx` field in the `bench_ecdh_data` struct (less flexible for future changes, deviates from other bench structures)

Found while reviewing #1698, see https://github.com/bitcoin-core/secp256k1/pull/1698#discussion_r2350395913.

### DIFF
--- a/src/modules/ecdh/bench_impl.h
+++ b/src/modules/ecdh/bench_impl.h
@@ -10,7 +10,7 @@
 #include "../../../include/secp256k1_ecdh.h"
 
 typedef struct {
-    secp256k1_context *ctx;
+    const secp256k1_context *ctx;
     secp256k1_pubkey point;
     unsigned char scalar[32];
 } bench_ecdh_data;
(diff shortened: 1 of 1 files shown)


sipa:
utACK dfe284ed2d24ab44705ca5ecd8f3f7e1b8e4c278

furszy (commented):
ACK dfe284ed2d24ab44705ca5ecd8f3f7e1b8e4c278

If it wouldn't be for the initial `VERIFY_CHECK`, the context could be null as it is not used internally.

PR #1756, opened 2025-10-03
ci: Fix image caching and apply other improvements

hebasto:
This PR fixes an issue where only the latest image cache was available.

For other minor improvements, see the individual commit messages.

### DIFF
--- a/.github/actions/run-in-docker-action/action.yml
+++ b/.github/actions/run-in-docker-action/action.yml
@@ -3,7 +3,8 @@ description: 'Run a command in a Docker container, while passing explicitly set 
 inputs:
   dockerfile:
     description: 'A Dockerfile that defines an image'
-    required: true
+    required: false
+    default: ./ci/linux-debian.Dockerfile
   command:
     description: 'A command to run in a container'
     required: false
--- a/.github/actions/run-in-docker-action/action.yml
+++ b/.github/actions/run-in-docker-action/action.yml
@@ -14,7 +14,7 @@ runs:
   steps:
     - uses: docker/setup-buildx-action@v3
 
-    - uses: docker/build-push-action@v5
+    - uses: docker/build-push-action@v6
       id: main_builder
       continue-on-error: true
       with:
@@ -23,7 +23,7 @@ runs:
         load: true
         cache-from: type=gha,scope=${{ runner.arch }}
 
-    - uses: docker/build-push-action@v5
+    - uses: docker/build-push-action@v6
       id: retry_builder
       if: steps.main_builder.outcome == 'failure'
       with:
--- a/.github/actions/run-in-docker-action/action.yml
+++ b/.github/actions/run-in-docker-action/action.yml
@@ -24,7 +24,7 @@ runs:
         file: ${{ inputs.dockerfile }}
         tags: ${{ inputs.tag }}
         load: true
-        cache-from: type=gha
+        cache-from: type=gha,scope=${{ runner.arch }}
 
     - uses: docker/build-push-action@v5
       id: retry_builder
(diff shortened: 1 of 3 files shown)


real-or-random:
Relevant docs: https://docs.docker.com/build/cache/backends/gha/

.github/actions/run-in-docker-action/action.yml
```diff
@@ -16,25 +14,23 @@ runs:
...
     - uses: docker/setup-buildx-action@v3
 
-    - uses: docker/build-push-action@v5
+    - uses: docker/build-push-action@v6
       id: main_builder
       continue-on-error: true
       with:
         context: .
         file: ${{ inputs.dockerfile }}
-        tags: ${{ inputs.tag }}
```
real-or-random, line 25: Even after looking at the docs, I'm not sure what the `tags` parameter is actually supposed to do. Did you figure this out?

.github/actions/run-in-docker-action/action.yml
```diff
@@ -16,25 +14,23 @@ runs:
...
         load: true
-        cache-from: type=gha
+        cache-from: type=gha,scope=${{ runner.arch }}
```
real-or-random: Would it be a good idea to add `{{input.tag}}` here (and maybe rename `tag` to `scope`)?

.github/actions/run-in-docker-action/action.yml
```diff
@@ -3,10 +3,8 @@ description: 'Run a command in a Docker container, while passing explicitly set
 inputs:
   dockerfile:
     description: 'A Dockerfile that defines an image'
-    required: true
-  tag:
-    description: 'A tag of an image'
-    required: true
+    required: false
+    default: ./ci/linux-debian.Dockerfile
```
real-or-random: I'm not sure if this is a reasonable default. I mean, yes, this is what we always use. But currently, this action is not specific to libsecp256k1 at all, and I think it would be nice to keep it that way.

ci/linux-debian.Dockerfile
```diff
@@ -1,5 +1,7 @@
 FROM debian:stable-slim
 
+ENV DEBIAN_FRONTEND=noninteractive
```
real-or-random: I think this is not a good idea because the Dockerfile can also be used interactively on a developer's machine. See https://serverfault.com/a/797318

.github/actions/run-in-docker-action/action.yml
hebasto: It's meant to be used when exporting a built image to a registry.

.github/actions/run-in-docker-action/action.yml
hebasto: I don't think so. Tags are not used for caching.

.github/actions/run-in-docker-action/action.yml
hebasto: It seems unlikely that this action would be reused elsewhere.

.github/actions/run-in-docker-action/action.yml
real-or-random: I think we're talking past each other. (But sorry, my message was very unclear.) My suggestion is to add an input parameter to the `run-in-docker-action` that is used as a prefix to `scope`, i.e., the caller sets the `scope`.

In our case, the caller could then set scope=`ci-image-${{ runner.arch }}` or something like this.

ci/linux-debian.Dockerfile
hebasto: Thanks! Reworked.

.github/actions/run-in-docker-action/action.yml
hebasto: I see. Reworked.

.github/actions/run-in-docker-action/action.yml
hebasto: The controversial commit has been dropped.

hebasto:
@real-or-random 

.github/actions/run-in-docker-action/action.yml
real-or-random: > It seems unlikely that this action would be reused elsewhere.

True, but I think it's still a good idea to drop the default. It simply decreases coupling.

PR #1758, opened 2025-10-15
ci: Drop workaround for Valgrind older than 3.20.0

hebasto:
This is no longer needed in the current CI framework.

If someone runs the CI scripts locally, it is reasonable to expect that they are using up-to-date tools, including Valgrind.

### DIFF
--- a/ci/ci.sh
+++ b/ci/ci.sh
@@ -52,22 +52,6 @@ if [ -n "$WRAPPER_CMD" ]; then
     $WRAPPER_CMD --version
 fi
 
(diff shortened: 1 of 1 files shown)


real-or-random (approved):
ACK https://github.com/bitcoin-core/secp256k1/pull/1758/commits/c09519f0e383da3c36e33522ca8708c17103d96a 

The 3.20 release was [3 years ago](https://sourceware.org/git/?p=valgrind.git;a=tag;h=dc06ae5475cb5742a79793c429e4654da26aff35), and [Debian stable has 3.24](https://packages.debian.org/trixie/valgrind).

PR #1760, opened 2025-10-16
cmake: Add dynamic test discovery to improve parallelism

hebasto:
This PR implements the idea suggested in https://github.com/bitcoin-core/secp256k1/pull/1734#pullrequestreview-3284918572 and is based on the work from https://github.com/bitcoin/bitcoin/pull/33483.

Here is an example of the `ctest` output:
```
$ ctest --test-dir build -j $(nproc)
Test project /home/hebasto/dev/secp256k1/secp256k1/build
        Start   1: secp256k1.noverify_tests.selftest_tests
        Start   2: secp256k1.noverify_tests.all_proper_context_tests
        Start   3: secp256k1.noverify_tests.all_static_context_tests
        Start   4: secp256k1.noverify_tests.deprecated_context_flags_test
<snip>
193/196 Test  #31: secp256k1.noverify_tests.ecmult_constants .........................   Passed    5.32 sec
194/196 Test #184: secp256k1.tests.ellswift_xdh_correctness_tests ....................   Passed    5.62 sec
195/196 Test #191: secp256k1.exhaustive_tests ........................................   Passed    6.97 sec
196/196 Test #126: secp256k1.tests.ecmult_constants ..................................   Passed    9.60 sec

100% tests passed, 0 tests failed out of 196

Label Time Summary:
secp256k1_example           =   0.02 sec*proc (5 tests)
secp256k1_exhaustive        =   6.97 sec*proc (1 test)
secp256k1_noverify_tests    =  23.77 sec*proc (95 tests)
secp256k1_tests             =  43.67 sec*proc (95 tests)

Total Test time (real) =  10.21 sec
```

For comparison, here is the output for the master branch on the same machine:
```
$ ctest --test-dir build -j $(nproc)
Test project /home/hebasto/dev/secp256k1/secp256k1/build
    Start 1: secp256k1_noverify_tests
    Start 2: secp256k1_tests
    Start 3: secp256k1_exhaustive_tests
    Start 4: secp256k1_ecdsa_example
    Start 5: secp256k1_ecdh_example
    Start 6: secp256k1_schnorr_example
    Start 7: secp256k1_ellswift_example
    Start 8: secp256k1_musig_example
1/8 Test #4: secp256k1_ecdsa_example ..........   Passed    0.00 sec
2/8 Test #5: secp256k1_ecdh_example ...........   Passed    0.00 sec
3/8 Test #6: secp256k1_schnorr_example ........   Passed    0.00 sec
4/8 Test #7: secp256k1_ellswift_example .......   Passed    0.00 sec
5/8 Test #8: secp256k1_musig_example ..........   Passed    0.00 sec
6/8 Test #3: secp256k1_exhaustive_tests .......   Passed    6.26 sec
7/8 Test #1: secp256k1_noverify_tests .........   Passed   14.31 sec
8/8 Test #2: secp256k1_tests ..................   Passed   31.65 sec

100% tests passed, 0 tests failed out of 8

Total Test time (real) =  31.65 sec
```

---

**New Feature:** As the number of tests has grown, the _labels_ have been introduced to simplify test management. Now, one can run:
```
$ ctest --test-dir build -j $(nproc) -L example
Test project /home/hebasto/dev/secp256k1/secp256k1/build
    Start 192: secp256k1.example.ecdsa
    Start 193: secp256k1.example.ecdh
    Start 194: secp256k1.example.schnorr
    Start 195: secp256k1.example.ellswift
    Start 196: secp256k1.example.musig
1/5 Test #192: secp256k1.example.ecdsa ..........   Passed    0.00 sec
2/5 Test #193: secp256k1.example.ecdh ...........   Passed    0.00 sec
3/5 Test #194: secp256k1.example.schnorr ........   Passed    0.00 sec
4/5 Test #195: secp256k1.example.ellswift .......   Passed    0.00 sec
5/5 Test #196: secp256k1.example.musig ..........   Passed    0.00 sec

100% tests passed, 0 tests failed out of 5

Label Time Summary:
secp256k1_example    =   0.01 sec*proc (5 tests)

Total Test time (real) =   0.01 sec
```
or
```
$ ctest --test-dir build -j $(nproc) -LE tests
Test project /home/hebasto/dev/secp256k1/secp256k1/build
    Start 192: secp256k1.example.ecdsa
    Start 193: secp256k1.example.ecdh
    Start 194: secp256k1.example.schnorr
    Start 195: secp256k1.example.ellswift
    Start 196: secp256k1.example.musig
    Start 191: secp256k1.exhaustive_tests
1/6 Test #192: secp256k1.example.ecdsa ..........   Passed    0.00 sec
2/6 Test #193: secp256k1.example.ecdh ...........   Passed    0.00 sec
3/6 Test #194: secp256k1.example.schnorr ........   Passed    0.00 sec
4/6 Test #195: secp256k1.example.ellswift .......   Passed    0.00 sec
5/6 Test #196: secp256k1.example.musig ..........   Passed    0.00 sec
6/6 Test #191: secp256k1.exhaustive_tests .......   Passed    6.19 sec

100% tests passed, 0 tests failed out of 6

Label Time Summary:
secp256k1_example       =   0.01 sec*proc (5 tests)
secp256k1_exhaustive    =   6.19 sec*proc (1 test)

Total Test time (real) =   6.20 sec
```

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -9,6 +9,9 @@ function(add_example name)
     $<$<PLATFORM_ID:Windows>:bcrypt>
   )
   add_test(NAME secp256k1.example.${name} COMMAND ${target_name})
+  set_tests_properties(secp256k1.example.${name} PROPERTIES
+    LABELS secp256k1_example
+  )
 endfunction()
 
 add_example(ecdsa)
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -154,6 +154,8 @@ if(SECP256K1_BUILD_TESTS)
       DISCOVERY_MATCH       "^\\t\\\\[ *[0-9]+\\\\] ([^ ].*)$"
       TEST_NAME_REPLACEMENT "secp256k1.${exe_name}.\\\\1"
       TEST_ARGS_REPLACEMENT "--target=\\\\1 --log=1"
+      PROPERTIES
+        LABELS "secp256k1_${exe_name}"
     )
   endfunction()
 
@@ -170,6 +172,9 @@ if(SECP256K1_BUILD_EXHAUSTIVE_TESTS)
   target_link_libraries(exhaustive_tests secp256k1_asm)
   target_compile_definitions(exhaustive_tests PRIVATE $<$<NOT:$<CONFIG:Coverage>>:VERIFY>)
   add_test(NAME secp256k1.exhaustive_tests COMMAND exhaustive_tests)
+  set_tests_properties(secp256k1.exhaustive_tests PROPERTIES
+    LABELS secp256k1_exhaustive
+  )
 endif()
 
 if(SECP256K1_BUILD_CTIME_TESTS)


real-or-random:
The output is nice for sure, but wouldn't it be much simpler to just invoke the test binaries with `-j`? 

My concern is about the list of tests here. It will require maintenance work, and it's easy to forget a test.

sipa:
> The output is nice for sure, but wouldn't it be much simpler to just invoke the test binaries with -j?

I don't think that's desirable, because it lacks the ability to share the granularity with other test targets.

Specifically, I'm thinking in the context of the Bitcoin Core build, which has tons of test cases of its own. The ideal scenario is that you have a *single* process pool (controlled by the `-j` argument to `cmake`), and it is used to run both the Bitcoin Core and libsecp256k1 (and other) tests. This will generally result in all processes making progress throughout the entire run, until `cmake` runs out of test tasks to schedule.

In contrast, if you push the task scheduling down to libsecp256k1's own test binary, you at best end up with a situation where you first run all Bitcoin Core tests, wait for those to complete (temporarily running out of tasks in the process), and then run libsecp256k1's tests with `-j`. I don't know if that's actually possible to do in `cmake`'s supported generators, and more likely you'll end up in a situation where the entire libsecp256k1 test run (`-j` and all) is seen as a single process from the perspective of the build system - meaning it'd simultaneously run N-1 Bitcoin Core tests, *plus* 1 libsecp256k1 test (which itself runs in N processes), temporarily raising the parallellism to 2N-1.

src/CMakeLists.txt
```diff
@@ -144,15 +144,23 @@ if(SECP256K1_BUILD_TESTS)
...
+    add_test(NAME secp256k1_${exe_name} COMMAND ${exe_name})
+  endfunction()
+
+  add_executable_and_tests(NO)
   if(NOT CMAKE_BUILD_TYPE STREQUAL "Coverage")
-    add_executable(tests tests.c)
-    target_compile_definitions(tests PRIVATE VERIFY ${TEST_DEFINITIONS})
-    target_link_libraries(tests secp256k1_precomputed secp256k1_asm)
-    add_test(NAME secp256k1_tests COMMAND tests)
+    add_executable_and_tests(YES)
```
furszy: Could write this shorter as:
```
function(add_executable_and_tests bin_name verify_definition)
  add_executable(${bin_name} tests.c)
  target_link_libraries(${bin_name} secp256k1_precomputed secp256k1_asm)
  target_compile_definitions(${bin_name} PRIVATE ${verify_definition} ${TEST_DEFINITIONS})
  add_test(NAME secp256k1_${bin_name} COMMAND ${bin_name})
endfunction()

add_executable_and_tests(noverify_tests "")
if(NOT CMAKE_BUILD_TYPE STREQUAL "Coverage")
  add_executable_and_tests(tests VERIFY)
endif()
```

Also, probably the `target_link_libraries` should be PRIVATE too.

See 00f20f694704079f1f9c05851e7407883763268e (I applied it on top of your first commit -  could directly squash it there)

furszy:
Regarding the maintenance topic, wouldn't be better to call `--list_tests` to fetch the test names instead of manually adding them one by one? We could make `--list_tests=2` output the tests names separated by commas to make your life easier at the parsing side.

I experimented a bit with this idea and was able to dynamically generate a file containing all the tests names. See https://github.com/furszy/secp256k1/tree/2025_ci_concurrent_tests (just need to run the `tests_discover` target). 
The next step would be to connect this to run separate ctest targets in some way, which doesn't seem to be so simple because the binary does not exist during configure time but ctest targets are added there.. (chicken-egg problem). Maybe we could execute a custom script that reads the file and launches per target ctest runs dynamically too.

hebasto:
> > The output is nice for sure, but wouldn't it be much simpler to just invoke the test binaries with -j?
> 
> I don't think that's desirable, because it lacks the ability to share the granularity with other test targets.
> 
> Specifically, I'm thinking in the context of the Bitcoin Core build, which has tons of test cases of its own. The ideal scenario is that you have a _single_ process pool (controlled by the `-j` argument to `cmake`), and it is used to run both the Bitcoin Core and libsecp256k1 (and other) tests. This will generally result in all processes making progress throughout the entire run, until `cmake` runs out of test tasks to schedule.

I share the same perspective.

hebasto:
> Regarding the maintenance topic, wouldn't be better to call `--list_tests` to fetch the test names instead of manually adding them one by one? We could make `--list_tests=2` output the tests names separated by commas to make your life easier at the parsing side.
> 
> I experimented a bit with this idea and was able to dynamically generate a file containing all the tests names. See https://github.com/furszy/secp256k1/tree/2025_ci_concurrent_tests (just need to run the `tests_discover` target). The next step would be to connect this to run separate ctest targets in some way, which doesn't seem to be so simple because the binary does not exist during configure time but ctest targets are added there.. (chicken-egg problem). Maybe we could execute a custom script that reads the file and launches per target ctest runs dynamically too.

@purpleKarrot's [approach](https://github.com/bitcoin/bitcoin/pull/33483) could be applied as a follow-up.

src/CMakeLists.txt
hebasto: > Also, probably the `target_link_libraries` should be PRIVATE too.

All `target_link_libraries` commands should use the same signature. I think this change should be made in a separate PR.

hebasto:
Feedback from @furszy has been addressed.

src/CMakeLists.txt
hebasto: > Could write this shorter as: ...

Thanks! [Reworked](https://github.com/bitcoin-core/secp256k1/pull/1760#issuecomment-3425282663).

src/CMakeLists.txt
```diff
@@ -144,16 +144,42 @@ if(SECP256K1_BUILD_TESTS)
...
+  set(test_names "")
+  # Start with the longest-running tests.
+  list(APPEND test_targets ellswift)
+  list(APPEND test_names   ellswift)
+  list(APPEND test_targets ecmult_constants)
+  list(APPEND test_names   ecmult_constants)
+  list(APPEND test_targets "ecmult_pre_g wnaf point_times_order ecmult_near_split_bound ecmult_chain ecmult_gen_blind ecmult_const_tests ecmult_multi_tests ec_combine")
+  list(APPEND test_names   ecmult)
+  # Continue with the remaining tests.
+  list(APPEND test_targets integer scalar field group ec ecdh ecdsa recovery extrakeys schnorrsig musig)
```
real-or-random: Is there a reason why you have quotation marks in the first line here but not in the last?

hebasto: The first line has quotes to combine the quoted test cases into a single `ctest`'s test "ecmult".

In the last line, all mentioned test cases and modules are wrapped as its own `ctest`'s tests.

purpleKarrot: I take @real-or-random's question as a proof that procedural logic in `CMakeLists.txt` is hard to reason about and should be avoided.

src/CMakeLists.txt
```diff
@@ -144,15 +144,43 @@ if(SECP256K1_BUILD_TESTS)
...
+    endforeach()
+  endfunction()
+
+  add_executable_and_tests(noverify_tests "" secp256k1_noverify)
   if(NOT CMAKE_BUILD_TYPE STREQUAL "Coverage")
-    add_executable(tests tests.c)
-    target_compile_definitions(tests PRIVATE VERIFY ${TEST_DEFINITIONS})
-    target_link_libraries(tests secp256k1_precomputed secp256k1_asm)
-    add_test(NAME secp256k1_tests COMMAND tests)
+    add_executable_and_tests(tests VERIFY secp256k1_verify)
```
purpleKarrot: The target name `tests` is very generic and likely to cause clashes when the project is used as a subproject. This is orthogonal to the PR, but should be kept in mind for a follow up.

src/CMakeLists.txt
purpleKarrot: If we switch to automated test discovery in a follow-up, we will no longer have the ability to combine test cases. Maybe we should keep them separated so that the follow-up will not change the number of tests again?

real-or-random: > Maybe we should keep them separated so that the follow-up will not change the number of tests again?

Sounds good, let's keep it simple. 

For test ordering, we could perhaps simply make sure that automated discovery runs the tests in the order they are listed by `tests -l`.

furszy: > For test ordering, we could perhaps simply make sure that automated discovery runs the tests in the order they are listed by tests -l.

We could reorder them internally based on their execution time, so the slow ones run first.
It is just matter of moving a few lines in the registry anyway, very simple.

purpleKarrot: > We could reorder them internally based on their execution time, so the slow ones run first.
> It is just matter of moving a few lines in the registry anyway, very simple.

The order in which they are registered to CTest is not the order they will be executed. If you want slow tests to run first, you need to set the [`COST`](https://cmake.org/cmake/help/git-master/prop_test/COST.html) property.

real-or-random: Oh, I see, also fine! I suggest we just run the test suite once on some dev's machine and hard-code the resulting numbers.

furszy: Created #1788 so we don’t have to worry about the test costs for now. The longest running test is just an all-in-one test that we can split very easily (the gains of it are very noticeable).

furszy: Can we simplify anything now that https://github.com/bitcoin-core/secp256k1/pull/1788 was merged?

hebasto:
> Regarding the maintenance topic, wouldn't be better to call `--list_tests` to fetch the test names instead of manually adding them one by one? We could make `--list_tests=2` output the tests names separated by commas to make your life easier at the parsing side.
> 
> I experimented a bit with this idea and was able to dynamically generate a file containing all the tests names. See https://github.com/furszy/secp256k1/tree/2025_ci_concurrent_tests (just need to run the `tests_discover` target). The next step would be to connect this to run separate ctest targets in some way, which doesn't seem to be so simple because the binary does not exist during configure time but ctest targets are added there.. (chicken-egg problem). Maybe we could execute a custom script that reads the file and launches per target ctest runs dynamically too.

Reworked. Using the dynamic test discovery now.

furszy:
nice! will review soon.

examples/CMakeLists.txt
```diff
@@ -9,7 +9,7 @@ function(add_example name)
     $<$<PLATFORM_ID:Windows>:bcrypt>
   )
   set(test_name ${name}_example)
-  add_test(NAME secp256k1_${test_name} COMMAND ${target_name})
+  add_test(NAME secp256k1.example.${name} COMMAND ${target_name})
```
furszy, line 11: In 0c8178e86cfddc193429ae398c6d5ff6537ef73c:
with this change, `test_name` is no longer used.

Update:
I see you are removing it in the next commit: 54ffc2137004ea3085d3c5068c1562ca46a6faeb.

src/CMakeLists.txt
```diff
@@ -154,6 +154,8 @@ if(SECP256K1_BUILD_TESTS)
       DISCOVERY_MATCH       "^\\t\\\\[ *[0-9]+\\\\] ([^ ].*)$"
       TEST_NAME_REPLACEMENT "secp256k1.${exe_name}.\\\\1"
       TEST_ARGS_REPLACEMENT "--target=\\\\1 --log=1"
+      PROPERTIES
+        LABELS "secp256k1_${exe_name}"
```
furszy, line 158: it would be nice to explain what LABELS are.

cmake/DiscoverTests.cmake
```diff
@@ -0,0 +1,71 @@
...
+
+  set(properties_content)
+  list(LENGTH arg_PROPERTIES properties_len)
+  if(properties_len GREATER "0")
+    set(properties_content "      set_tests_properties(\"\${test_name}\" PROPERTIES\n")
+    math(EXPR num_properties "${properties_len} / 2")
+    foreach(i RANGE 0 ${num_properties} 2)
+      math(EXPR value_index "${i} + 1")
+      list(GET arg_PROPERTIES ${i} name)
+      list(GET arg_PROPERTIES ${value_index} value)
```
furszy, line 20: This assumes properties come in pairs, but someone could:

```c++
discover_tests(${exe_name}
      DISCOVERY_ARGS        "--list_tests"
      DISCOVERY_MATCH       "^\\t\\\\[ *[0-9]+\\\\] ([^ ].*)$"
      TEST_NAME_REPLACEMENT "secp256k1.${exe_name}.\\\\1"
      TEST_ARGS_REPLACEMENT "--target=\\\\1 --log=1"
      PROPERTIES
        LABELS "secp256k1_${exe_name}"
        NO_TIMEOUT   // <-- property with no value
)
```

It would be good to check that the value exist before accessing it, and fail with a clear error if not: "Property 'name' has no associated value. Must be specified as name/value pair" or something similar.

furszy (commented):
Aside from the minimal comments just left, this works nicely.

For reviewers, just run:
```
ctest -j <threads_num>
```

examples/CMakeLists.txt
hebasto: > Update: I see you are removing it in the next commit: [54ffc21](https://github.com/bitcoin-core/secp256k1/commit/54ffc2137004ea3085d3c5068c1562ca46a6faeb).

It has been moved to previous commit.

src/CMakeLists.txt
hebasto: From [Professional CMake: A Practical Guide 21st Edition](https://discourse.cmake.org/t/professional-cmake-a-practical-guide-21st-edition/14127):
> Selecting tests individually by name or number can become cumbersome if a large set of related tests needs to be executed. Tests can be assigned an arbitrary list of labels using the `LABELS` test property, and then tests can be selected by these labels. The `-L` and `-LE` options are analogous to the `-R` and `-E` options respectively, except they operate on test labels rather than test names.
> ...
> Labels not only enable convenient grouping for test execution, they also provide grouping for basic execution time statistics. As seen in the example output above, the `ctest` command prints a label summary when any test in the set of executed tests has its `LABELS` property set. This allows the developer to get an idea how each label group is contributing to the overall test time.

cmake/DiscoverTests.cmake
hebasto: Every set property has a value. The syntax in the snippet above is simply incorrect. Compare it with the syntax of the [`set_tests_properties`](https://cmake.org/cmake/help/latest/command/set_tests_properties.html) command.

hebasto:
@furszy 

furszy (commented):
Tested ACK 8354618

purpleKarrot:
ACK 8354618e02bcccf3893bcf0ec99639b6e520509f

Reviewed and tested.

PR #1764, opened 2025-10-28
group: Avoid using infinity field directly in other modules

real-or-random:
Minor refactoring to make the abstraction cleaner

### DIFF
--- a/src/testutil.h
+++ b/src/testutil.h
@@ -96,17 +96,13 @@ static void testutil_random_ge_test(secp256k1_ge *ge) {
             break;
         }
     } while(1);
-    ge->infinity = 0;
 }
 
 static void testutil_random_ge_jacobian_test(secp256k1_gej *gej, const secp256k1_ge *ge) {
-    secp256k1_fe z2, z3;
-    testutil_random_fe_non_zero_test(&gej->z);
-    secp256k1_fe_sqr(&z2, &gej->z);
-    secp256k1_fe_mul(&z3, &z2, &gej->z);
-    secp256k1_fe_mul(&gej->x, &ge->x, &z2);
-    secp256k1_fe_mul(&gej->y, &ge->y, &z3);
-    gej->infinity = ge->infinity;
+    secp256k1_fe z;
+    testutil_random_fe_non_zero_test(&z);
+    secp256k1_gej_set_ge(gej, ge);
+    secp256k1_gej_rescale(gej, &z);
 }
 
 static void testutil_random_gej_test(secp256k1_gej *gej) {
--- a/src/ecmult_const_impl.h
+++ b/src/ecmult_const_impl.h
@@ -87,17 +87,15 @@ static void secp256k1_ecmult_const_odd_multiples_table_globalz(secp256k1_ge *pre
     secp256k1_fe neg_y; \
     VERIFY_CHECK((n) < (1U << ECMULT_CONST_GROUP_SIZE)); \
     VERIFY_CHECK(index < (1U << (ECMULT_CONST_GROUP_SIZE - 1))); \
-    /* Unconditionally set r->x = (pre)[m].x. r->y = (pre)[m].y. because it's either the correct one
+    /* Unconditionally set r->x = (pre)[m].x and r->y = (pre)[m].y because it's either the correct one
      * or will get replaced in the later iterations, this is needed to make sure `r` is initialized. */ \
-    (r)->x = (pre)[m].x; \
-    (r)->y = (pre)[m].y; \
+    secp256k1_ge_set_xy((r), &(pre)[m].x, &(pre)[m].y); \
     for (m = 1; m < ECMULT_CONST_TABLE_SIZE; m++) { \
         /* This loop is used to avoid secret data in array indices. See
          * the comment in ecmult_gen_impl.h for rationale. */ \
         secp256k1_fe_cmov(&(r)->x, &(pre)[m].x, m == index); \
         secp256k1_fe_cmov(&(r)->y, &(pre)[m].y, m == index); \
     } \
-    (r)->infinity = 0; \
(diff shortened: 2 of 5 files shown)


theStack:
Concept ACK

real-or-random:
Updated. There are three lines left in the tests, but all of them are justified. Check:
```
git grep -E "(\->|\.)infinity" -- src ':(exclude)src/group_impl.h'
```

By the way, one could think that this PR is a bit arbitrary because we also access the x, y, z fields everywhere. Perhaps this could also be improved. And this would actually have an immediate benefit because we could check that they're only accessed if the point is not infinity. But this should go to a separate PR, I think.

hebasto:
> Updated. There are three lines left in the tests, but all of them are justified. Check:
> 
> ```
> git grep -E "(\->|\.)infinity" -- src ':(exclude)src/group_impl.h'
> ```
> 
> By the way, one could think that this PR is a bit arbitrary because we also access the x, y, z fields everywhere. Perhaps this could also be improved. And this would actually have an immediate benefit because we could check that they're only accessed if the point is not infinity. But this should go to a separate PR, I think.

I still think that `secp256k1_gej_is_infinity` could be used here:https://github.com/bitcoin-core/secp256k1/blob/2f73e5281de44e2c2bcf60ebc7b113440e13b07f/src/tests.c#L3653-L3654

If I'm wrong, maybe add an explanatory comment there?

src/testutil.h
```diff
@@ -96,17 +96,13 @@ static void testutil_random_ge_test(secp256k1_ge *ge) {
             break;
         }
     } while(1);
-    ge->infinity = 0;
```
hebasto, line 99: Why is it correct to drop this line?

src/testutil.h
```diff
@@ -96,17 +96,13 @@ static void testutil_random_ge_test(secp256k1_ge *ge) {
...
-    testutil_random_fe_non_zero_test(&gej->z);
-    secp256k1_fe_sqr(&z2, &gej->z);
-    secp256k1_fe_mul(&z3, &z2, &gej->z);
-    secp256k1_fe_mul(&gej->x, &ge->x, &z2);
-    secp256k1_fe_mul(&gej->y, &ge->y, &z3);
-    gej->infinity = ge->infinity;
+    secp256k1_fe z;
+    testutil_random_fe_non_zero_test(&z);
+    secp256k1_gej_set_ge(gej, ge);
+    secp256k1_gej_rescale(gej, &z);
```
hebasto, line 105: I had a hard time following these changes. Could you shed a bit more light on them?

src/testutil.h
real-or-random: `secp256k1_ge_set_xo_var` sets this already. 
https://github.com/bitcoin-core/secp256k1/blob/c8206b1ce60704bb1030d079ea9b3d99d8906220/src/group_impl.h#L355

src/testutil.h
real-or-random: What this function does is take a `ge` and convert it to a `gej` with a random `z` coordinate. This can be done in two steps:
 1. convert ge to gej
 2. rescale the gej

More details:
1. ge is a "normal" affine point with x and y coordinates. gej ([Jacobian coordinates](https://hyperelliptic.org/EFD/g1p/auto-shortw-jacobian-0.html)) represent `ge.x` and `ge.y` as `gej.x`  `gej.y` `gej.z` with
    ```
    ge.x = gej.x / gej.z^2
    ge.y = gej.y / gej.z^3
    ```
    Thus, the canonical way to `ge` to `gej` is to set `gej.z = 1` and then just `gej.x = ge.x` and `gej.y = ge.y` because `z^2 = 1` and `z^3 = 1`
2. Rescaling a gej means changing the representation without changing the effective x and y. This can be multiplying the existing z coordinate with a given field element. To accommodate, we'll need to multiply x with z^2 and y with z^3.

Or another answer: If you look into `secp256k1_gej_set_ge` and `secp256k1_gej_rescale` you'll be able to check that there are no semantic changes.

real-or-random:
> I still think that `secp256k1_gej_is_infinity` could be used here:
> 
> https://github.com/bitcoin-core/secp256k1/blob/2f73e5281de44e2c2bcf60ebc7b113440e13b07f/src/tests.c#L3653-L3654

Yeah, it could be used. My thinking is that `gej_xyz_equals_gej` is conceptually a "group" function. It would be in `group_impl.h` if it wasn't needed only in the tests. It will need raw access to x, y, z too, so it can also have raw access to the infinity flag. I could rename it and move it to `group_impl.h`. I mean, there's nothing wrong with having functions there that are used only in the tests.

hebasto:
> > I still think that `secp256k1_gej_is_infinity` could be used here:
> > https://github.com/bitcoin-core/secp256k1/blob/2f73e5281de44e2c2bcf60ebc7b113440e13b07f/src/tests.c#L3653-L3654
> 
> Yeah, it could be used. My thinking is that `gej_xyz_equals_gej` is conceptually a "group" function.

Indeed. I was looking for `secp256k1_gej_xyz_equals_gej` in `group.h` :)

> It would be in `group_impl.h` if it wasn't needed only in the tests. It will need raw access to x, y, z too, so it can also have raw access to the infinity flag. I could rename it and move it to `group_impl.h`.

That sounds good.

> I mean, there's nothing wrong with having functions there that are used only in the tests.

Agreed.

theStack (approved):
ACK 2f73e5281de44e2c2bcf60ebc7b113440e13b07f

real-or-random:
> > I mean, there's nothing wrong with having functions there that are used only in the tests.
> 
> Agreed.

I changed my mind here. There's also some benefit in keeping test-only code in the test files. That means moving the function to `group_impl.h` is not clearly better.  

I took the freedom to merge this as-is because it has two ACKs.

Issue #1768, opened 2025-11-07
ci: python -m pip install fails with error: externally-managed-environment


real-or-random:
See for example on master: https://github.com/bitcoin-core/secp256k1/actions/runs/19143357521/job/54714339317#step:6:37 This happens only on MacOS so far, but we should see this on Linux soon, too.

`lief` is needed for `symbol-check.py`. Since `lief` is a library, we can't use `pipx` and the clean way to solve this is to create a virtual environment for running the `symbol-check.py` script.

I suggest using `uv`, a modern Python package manager. It has support for declaring dependencies of single-file scripts in a comment, see https://docs.astral.sh/uv/guides/scripts/#running-a-script-without-dependencies . When you run the script, `uv` automatically creates a virtual environment and everything works out of the box. I think that's the best solution for stuff like `symbol-check.py`.

We use `uv` in secp256k1lab and are happy with it, I've also used it in a few other projects. It has a few other advantages, e.g., it can manage Python installations which is useful if we ever need a specific Python version. Or we could use it to test our Python scripts on multiple Python versions easily. [That's what we do in secp256k1lab, and we use a GitHub Action to install it.](https://github.com/secp256k1lab/secp256k1lab/blob/master/.github/workflows/main.yml#L15) (Note that using `uv` on CI does not force anyone to do the same. If you don't want to use `uv`, you can still install `lief` manually and run the script.)

PR #1771, opened 2025-11-07
ci: Use Python virtual environment in "x86_64-macos-native" job

hebasto:
Fixes https://github.com/bitcoin-core/secp256k1/issues/1768.

This PR explicitly sets up a virtual environment instead of using `uv`, as suggested [here](https://github.com/bitcoin-core/secp256k1/issues/1768#issue-3599176194), for the following reasons:
1. It does not require granting a third-party action access to the repository.
2. This approach is already used in other parts of the CI framework (it’s unclear why it was missed in https://github.com/bitcoin-core/secp256k1/pull/1359 in the first place).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -469,9 +469,14 @@ jobs:
         env: ${{ matrix.env_vars }}
         run: ./ci/ci.sh
 
-      - name: Symbol check
+      - &SYMBOL_CHECK_MACOS
+        name: Symbol check
+        env:
+          VIRTUAL_ENV: '${{ github.workspace }}/venv'
         run: |
           python3 --version
+          python3 -m venv $VIRTUAL_ENV
+          export PATH="$VIRTUAL_ENV/bin:$PATH"
           python3 -m pip install lief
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -469,9 +469,14 @@ jobs:
         env: ${{ matrix.env_vars }}
         run: ./ci/ci.sh
 
-      - name: Symbol check
+      - &SYMBOL_CHECK
```
real-or-random: nit:

```suggestion
      - &SYMBOL_CHECK_MACOS
```

 - because it's specific to macOS
 - avoid naming confusing with the env variable `SYMBOL_CHECK`

real-or-random (approved):
utACK d3ae05447429d66b065ce3982fb0c5864c18bb53 

Let's wait for green.

Thanks for fixing CI so quickly. :)

I agree, that's simpler and does the job. I still think uv is a good idea for devs who want to run the script locally, but that's a different story then.

.github/workflows/ci.yml
hebasto: Thanks! Reworked per your feedback.

Issue #1773, opened 2025-11-15
Add internal helper for serializing a non-infinity group element to a compressed public key


theStack:
In more recent modules (ellswift, musig and silentpayments), the following pattern of serializing a non-infinity group element to a 33-bytes compressed public key appears repeatedly:
```
    size_t ser_size = 33;
    int ser_ret;
    .....
    ser_ret = secp256k1_eckey_pubkey_serialize(ge, out33, &ser_size, 1);
#ifdef VERIFY
    VERIFY_CHECK(ser_ret && ser_size == 33);
#else
    (void)ser_ret;
#endif
```
One could deduplicate these instances by a new function, as suggested by @w0xlt in the latest silentpayments module PR, see https://github.com/bitcoin-core/secp256k1/pull/1765#pullrequestreview-3462461331. This could simplify the code in the modules quite a bit (less LOC, less variables, less pre-processor blocks). Demo branch: https://github.com/theStack/secp256k1/commit/e02365188ac8bb5129dc36c817fb465dfa5f7b76. It seems a reasonable idea to me, but not sure about the right naming of the function and where to put it. Also, one drawback might be that if the VERIFY_CHECK fails, it's a bit harder to find out the cause if it fails in a helper rather than directly in the module code (not sure if we care about that?). Opening this as an issue first to see if there is conceptual support.

real-or-random:
What about going a further step back and splitting `secp256k1_eckey_pubkey_serialize` into `secp256k1_eckey_pubkey_serialize_compressed` and `secp256k1_eckey_pubkey_serialize_uncompressed` (or alternatively, `_serialize33` and `serialize65`)? That seems even cleaner to me.  

Then we'll need to change the implementation of the API function `secp256k1_ec_pubkey_serialize` (which still has `compressed` flag) and do the branching there. But that's not a big deal: the current implementation needs to branch on this flag anyway for an `ARG_CHECK`, and it also needs the literals `33` and `65`, i.e., it doesn't really make great use of the fact that the public function and the internal function both have a `compressed` flag.

theStack:
@real-or-random: Sounds like a good idea, opened #1774 that does this.

PR #1765, opened 2025-10-31
Add "silentpayments" module implementing BIP352 (take 4, limited to full-node scanning)

theStack:
### Description

This PR implements [BIP352](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki) with scanning limited to full-nodes. Light-client scanning is planned to be added in a separate PR in the future. The following 7 API functions are currently introduced:

Sender side [[BIP description](https://github.com/bitcoin/bips/blob/c9a6ca6297eb8de850f6b64dafb8e60ee9b64d66/bip-0352.mediawiki?plain=1#L297-L312)]: 
* `secp256k1_silentpayments_sender_create_outputs`: given a list of $n$ secret keys $a_1 ... a_n$, a serialized outpoint, and a list of recipients (each consisting of silent payments scan pubkey and spend pubkey), create the corresponding transaction outputs (x-only public keys) for the sending transaction 

Receiver side, label creation [[BIP description](https://github.com/bitcoin/bips/blob/c9a6ca6297eb8de850f6b64dafb8e60ee9b64d66/bip-0352.mediawiki?plain=1#L200)]:
* `secp256k1_silentpayments_recipient_label_create`: given a scan secret key and label integer, calculate the corresponding label tweak and label object
* `secp256k1_silentpayments_recipient_label_serialize`: given a label object, create the corresponding 33-byte serialization
* `secp256k1_silentpayments_recipient_label_parse`: given a 33-byte label representation, create the corresponding label object
* `secp256k1_silentpayments_recipient_create_labeled_spend_pubkey`: given a spend public key and a label object, create the corresponding labeled spend public key

Receiver side, scanning [[BIP description](https://github.com/bitcoin/bips/blob/c9a6ca6297eb8de850f6b64dafb8e60ee9b64d66/bip-0352.mediawiki?plain=1#L334)]: 
* `secp256k1_silentpayments_recipient_prevouts_summary_create`: given a list of $n$ public keys $A_1 ... A_n$ and a serialized outpoint, create a `prevouts_summary` object needed for scanning
* `secp256k1_silentpayments_recipient_scan_outputs`: given a `prevouts_summary` object, a recipients scan secret key and spend public key, and the relevant transaction outputs (x-only public keys), scan for outputs belonging to the recipients and and return the tweak(s) needed for spending the output(s). Optionally, a label_lookup callback function can be provided to also scan for labels.

[...]

PR #1774, opened 2025-11-17
refactor: split up internal pubkey serialization function into compressed/uncompressed variants

theStack:
This PR splits up the pubkey serialization function `secp256k1_eckey_pubkey_serialize` into two variants for the compressed (33 bytes) and uncompressed (65 bytes) public key output format each, where only non-infinity group elements as input are allowed. The motivation is to simplify call-sites significantly, as they currently need to introduce two variables and a VERIFY_CHECKs on the return value and the in/out size parameter within a pre-processor block, typically leading to 8 lines of code. By using the new functions, the code is reduced to a single line of code that just calls the function (see #1773). This is helpful for already existing modules on master (ellswift, musig) and upcoming ones (silentpayments, see #1765).

One drawback is that the public API function `secp256k1_ec_pubkey_serialize` is now slightly more complex (we now call one of two functions instead of a single one, depending on whether the compressed flag is set or not), but that should hopefully not be a problem.

The commits are intentionally kept small to ease review, happy to squash them if that is preferred.

(Kudos to w0xlt for the initial idea (https://github.com/bitcoin-core/secp256k1/pull/1765#pullrequestreview-3462461331) and to real-or-random for the suggestion to split the already existing function (https://github.com/bitcoin-core/secp256k1/issues/1773#issuecomment-3540461718).)

### DIFF
--- a/src/eckey.h
+++ b/src/eckey.h
@@ -15,7 +15,6 @@
 #include "ecmult_gen.h"
 
 static int secp256k1_eckey_pubkey_parse(secp256k1_ge *elem, const unsigned char *pub, size_t size);
-static int secp256k1_eckey_pubkey_serialize(secp256k1_ge *elem, unsigned char *pub, size_t *size, int compressed);
 /** Serialize a group element (that is not allowed to be infinity) to a compressed public key (33 bytes). */
 static void secp256k1_eckey_pubkey_serialize33(secp256k1_ge *elem, unsigned char *pub33);
 /** Serialize a group element (that is not allowed to be infinity) to an uncompressed public key (65 bytes). */
--- a/src/eckey_impl.h
+++ b/src/eckey_impl.h
@@ -35,26 +35,6 @@ static int secp256k1_eckey_pubkey_parse(secp256k1_ge *elem, const unsigned char 
     }
 }
 
-static int secp256k1_eckey_pubkey_serialize(secp256k1_ge *elem, unsigned char *pub, size_t *size, int compressed) {
-    VERIFY_CHECK(compressed == 0 || compressed == 1);
-
-    if (secp256k1_ge_is_infinity(elem)) {
-        return 0;
-    }
-    secp256k1_fe_normalize_var(&elem->x);
-    secp256k1_fe_normalize_var(&elem->y);
-    secp256k1_fe_get_b32(&pub[1], &elem->x);
-    if (compressed) {
-        *size = 33;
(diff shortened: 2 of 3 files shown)


w0xlt (commented):
Approach ACK

src/secp256k1.c
```diff
@@ -280,10 +280,14 @@ int secp256k1_ec_pubkey_serialize(const secp256k1_context* ctx, unsigned char *o
...
+        if (flags & SECP256K1_FLAGS_BIT_COMPRESSION) {
+            secp256k1_eckey_pubkey_serialize33(&Q, output);
+            *outputlen = 33;
+        } else {
+            secp256k1_eckey_pubkey_serialize65(&Q, output);
+            *outputlen = 65;
         }
+        ret = 1;
     }
     return ret;
```
furszy, line 292: In adb76f82eadadf4f9df4cd958f7c266b551b50c6:

nit: could remove `ret` if you write it as:
```c
if (!secp256k1_pubkey_load(ctx, &Q, pubkey)) return 0;

if (flags & SECP256K1_FLAGS_BIT_COMPRESSION) {
    secp256k1_eckey_pubkey_serialize33(&Q, output);
    *outputlen = 33;
} else {
    secp256k1_eckey_pubkey_serialize65(&Q, output);
    *outputlen = 65;
}
return 1;
```

theStack: Makes sense yeah, will do if I have to retouch

fanquake:
Note that our downstream fuzzing of secp256k1 is currently broken until we land changes to cryptofuzz to adapt to the API changes here. We are using https://github.com/fanquake/cryptofuzz, however there is some activity in https://github.com/MozillaSecurity/cryptofuzz, which could be(come?) the new upstream (https://github.com/MozillaSecurity/cryptofuzz/issues/3).

real-or-random:
> Note that our downstream fuzzing of secp256k1 is currently broken until we land changes to cryptofuzz to adapt to the API changes here. We are using [fanquake/cryptofuzz](https://github.com/fanquake/cryptofuzz?rgh-link-date=2025-12-10T10%3A27%3A05Z), however there is some activity in [MozillaSecurity/cryptofuzz](https://github.com/MozillaSecurity/cryptofuzz?rgh-link-date=2025-12-10T10%3A27%3A05Z), which could be(come?) the new upstream ([MozillaSecurity/cryptofuzz#3](https://github.com/MozillaSecurity/cryptofuzz/issues/3)).

working on a PR to your repo

real-or-random:
https://github.com/fanquake/cryptofuzz/pull/1

PR #1775, opened 2025-11-18
Add CMake build directory patterns to `.gitignore`

hebasto:
Whenever I work on changes that require comparison, such as benchmarking, I end up with two or more build directories that provide different binary variants simultaneously. Adding these build directories to `.gitignore` makes the workflow a bit easier.

Additionally, a trivial refactoring is included to reduce the code.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -52,8 +52,8 @@ libsecp256k1.pc
 
 ### CMake
 /CMakeUserPresets.json
-# Default CMake build directory.
-/build
+# CMake build directories.
+/*build*
 
 ### Python
 __pycache__/


.gitignore
```diff
@@ -45,26 +45,15 @@ coverage.*.html
...
+!/build-aux/m4/bitcoin_secp.m4
+
 libsecp256k1.pc
 
 ### CMake
 /CMakeUserPresets.json
-# Default CMake build directory.
-/build
+# CMake build directories.
+/*build*
```
hebasto, line 56: The same pattern is used in [Bitcoin Core](https://github.com/bitcoin/bitcoin/blob/2444488f6ad32dcbbed51a73cd4f59ff3a239e32/.gitignore#L7).

real-or-random:
Concept ACK

We could also get rid of `build-aux` entirely by moving the contents of the only file into `configure.ac`. I think that would be nice. `build-aux` can be confusing to people familiar only with CMake, and it hinders tab-completion with CMake dirs etc...  What do you think?

hebasto:
> We could also get rid of `build-aux` entirely by moving the contents of the only file into `configure.ac`. I think that would be nice.

I think the main purpose of:https://github.com/bitcoin-core/secp256k1/blob/b6c2a3cd7752085d46434921238401766c964c01/configure.ac#L22 is to keep the top level of the repo uncluttered. Without it, it would end up filled as follows:
```
$ ./autogen.sh | grep -v libtoolize
configure.ac:38: installing './ar-lib'
configure.ac:36: installing './compile'
configure.ac:23: installing './config.guess'
configure.ac:23: installing './config.sub'
configure.ac:26: installing './install-sh'
configure.ac:26: installing './missing'
Makefile.am: installing './depcomp'
parallel-tests: installing './test-driver'
```

> `build-aux` can be confusing to people familiar only with CMake, and it hinders tab-completion with CMake dirs etc... What do you think?

The `build-aux` directory could be renamed. Perhaps to something like `autotools-aux`?

real-or-random:
Oh sure! Even I forgot the main purpose of the directory. 

> The `build-aux` directory could be renamed. Perhaps to something like `autotools-aux`?

I like that suggestion.

hebasto:
> > The `build-aux` directory could be renamed. Perhaps to something like `autotools-aux`?
> 
> I like that suggestion.

Thanks! Done.

PR #1778, opened 2025-12-04
doc/bench: Added cmake build options to bench error messages

kevkevinpal:
## Motivation
I wanted to try and run the benchmarking scripts and I noticed the recovery benchmark in `bench.c`. I wanted to run but I was using `cmake` and the error message telling me to use `./configure -enable-module-recovery` wasn't sufficient.

I figure rather than forcing users to look into the `CMakeLists.txt` file or anywhere else we should add this to the output

## Solution
I appended to the message to include the `-DSECP256K1_ENABLE_MODULE_...=ON` in the message.

### DIFF
--- a/src/bench.c
+++ b/src/bench.c
@@ -205,7 +205,7 @@ int main(int argc, char** argv) {
 #ifndef ENABLE_MODULE_ECDH
     if (have_flag(argc, argv, "ecdh")) {
         fprintf(stderr, "./bench: ECDH module not enabled.\n");
-        fprintf(stderr, "Use ./configure --enable-module-ecdh.\n\n");
+        fprintf(stderr, "See README.md for configuration instructions.\n\n");
         return EXIT_FAILURE;
     }
 #endif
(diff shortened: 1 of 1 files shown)


real-or-random:
Thanks for spotting this. Maybe a simpler fix (which will be easier to maintain in the future) is to remove this line (or simply refer to the README).

kevkevinpal:
> Thanks for spotting this. Maybe a simpler fix (which will be easier to maintain in the future) is to remove this line (or simply refer to the README).

Sounds good I pushed [3b5b03f](https://github.com/bitcoin-core/secp256k1/pull/1778/commits/3b5b03f30169903baae7f365cff3137a371a9a8d) to replace the message with this instead:

"See README.md for configuration instructions.\n\n"


let me know if this looks good, I can reword or simply delete the print if that is wanted

real-or-random (approved):
utACK 3b5b03f30169903baae7f365cff3137a371a9a8d

PR #1779, opened 2025-12-06
Add ARG_CHECKs to ensure "array of pointers" elements are non-NULL

theStack:
We currently have five public API functions that take an "array of pointers" as input parameter:
* `secp256k1_ec_pubkey_combine` (`ins`: array of pointers to public keys to add)
* `secp256k1_ec_pubkey_sort` (`pubkeys`: array of pointers to public keys to sort)
* `secp256k1_musig_pubkey_agg` (`pubkeys`: array of pointers to public keys to aggregate)
* `secp256k1_musig_nonce_agg` (`pubnonces`: array of pointers to public nonces to aggregate)
* `secp256k1_musig_partial_sig_agg` (`partial_sigs`: array of pointers to partial signatures to aggregate)

Out of these, only `_ec_pubkey_combine` verifies that the individual pointer elements in the array are non-NULL each:
https://github.com/bitcoin-core/secp256k1/blob/e7f7083b530a55c83ce9089a7244d2d9d67ac8b2/src/secp256k1.c#L774-L775

This PR adds corresponding `ARG_CHECKS` for the other API functions as well, in order to avoid running into potential UB due to NULL pointer dereference. It seems to me that the tiny run-time overhead is worth it doing this for consistency and to help users in case the arrays are set up incorrectly (I'm thinking e.g. of language binding writers where getting this right might be a bit more involved).

Looking into this was motivated by a [review of furszy](https://github.com/bitcoin-core/secp256k1/pull/1765#discussion_r2590644412) (thanks!), who pointed out that the non-NULL checks are missing in at least one API function in the silentpayments module PR as well. Happy to add some `CHECK_ILLEGAL` tests if there is conceptual support for this PR.

### DIFF
--- a/src/modules/musig/tests_impl.h
+++ b/src/modules/musig/tests_impl.h
@@ -201,6 +201,13 @@ static void musig_api_tests(void) {
     CHECK(secp256k1_musig_pubkey_agg(CTX, &agg_pk, &keyagg_cache, pk_ptr, 2) == 1);
     CHECK(secp256k1_musig_pubkey_agg(CTX, NULL, &keyagg_cache, pk_ptr, 2) == 1);
     CHECK(secp256k1_musig_pubkey_agg(CTX, &agg_pk, NULL, pk_ptr, 2) == 1);
+    /* check that NULL in array of public key pointers is not allowed */
+    for (i = 0; i < 2; i++) {
+        const secp256k1_pubkey *original_ptr = pk_ptr[i];
+        pk_ptr[i] = NULL;
+        CHECK_ILLEGAL(CTX, secp256k1_musig_pubkey_agg(CTX, &agg_pk, NULL, pk_ptr, 2));
+        pk_ptr[i] = original_ptr;
+    }
     CHECK_ILLEGAL(CTX, secp256k1_musig_pubkey_agg(CTX, &agg_pk, &keyagg_cache, NULL, 2));
     CHECK(memcmp_and_randomize(agg_pk.data, zeros132, sizeof(agg_pk.data)) == 0);
     CHECK_ILLEGAL(CTX, secp256k1_musig_pubkey_agg(CTX, &agg_pk, &keyagg_cache, invalid_pk_ptr2, 2));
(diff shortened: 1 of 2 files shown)


furszy:
Cool, the review paid off so fast :). Would be good to add test coverage for it.

src/secp256k1.c
```diff
@@ -325,8 +325,13 @@ static int secp256k1_ec_pubkey_sort_cmp(const void* pk1, const void* pk2, void *
 }
 
 int secp256k1_ec_pubkey_sort(const secp256k1_context* ctx, const secp256k1_pubkey **pubkeys, size_t n_pubkeys) {
+    size_t i;
+
     VERIFY_CHECK(ctx != NULL);
     ARG_CHECK(pubkeys != NULL);
+    for (i = 0; i < n_pubkeys; i++) {
```
kevkevinpal, line 332: Does it make sense to check `n_pubkeys` is greater than 0?
```suggestion
    ARG_CHECK(n_pubkeys > 0);
    for (i = 0; i < n_pubkeys; i++) {
```

real-or-random: I think it's okay to sort the empty array.

theStack:
Added `CHECK_ILLEGAL` tests to ensure NULL pointers in "array of pointer" arguments for the mentioned five functions are detected. They are placed right after (or at least as close to) successful executions with `n_{pubkeys,pubnonces,sigs} >= 2` each and test with a single NULL pointer at each position.

kevkevinpal:
utACK [8bcda18](https://github.com/bitcoin-core/secp256k1/pull/1779/commits/8bcda186d20e11ba9a04e6917928f062513b534f)

john-moffett:
utACK 8bcda186d20e11ba9a04e6917928f062513b534f

I like the macro idea, but I'd make the non-empty part explicit, like `ARG_CHECK_ARRAY_NONEMPTY` or something.

w0xlt (approved):
Looks good to me.
ACK https://github.com/bitcoin-core/secp256k1/pull/1779/commits/8bcda186d20e11ba9a04e6917928f062513b534f

real-or-random:
> What about introducing a new macro?

Concept ~0: It's more DRY, and it will make it harder for people to forget one of the checks, e.g., that the members are non-NULL, in new code. On the other hand, I believe the code will be more difficult to read with a macro. Having the three checks explicitly is not super verbose, and noone will need to look up or remember the definition of the macro. (Note that we'd use the macro only in a handful of functions, so people may not recall the definiton.)

PR #1783, opened 2025-12-09
Add VERIFY_CHECKs and documentation that flags must be 0 or 1 

john-moffett:
Flags for constant-time masking rely on the values being exactly `0` or `1` rather than `0` or true (any nonzero). One function, `secp256k1_fe_cmov` [documents](https://github.com/bitcoin-core/secp256k1/blob/e7f7083b530a55c83ce9089a7244d2d9d67ac8b2/src/field.h#L315) and [`VERIFY_CHECK`s](https://github.com/bitcoin-core/secp256k1/blob/e7f7083b530a55c83ce9089a7244d2d9d67ac8b2/src/field_impl.h#L365) this, but most don't. 

This updates the documentation and adds `VERIFY_CHECK`s enforcing `flag == 0 || flag == 1` for:

`secp256k1_fe_storage_cmov`
`secp256k1_gej_cmov`
`secp256k1_ge_storage_cmov`
`secp256k1_scalar_cadd_bit`
`secp256k1_scalar_cond_negate`
`secp256k1_scalar_cmov`
`secp256k1_int_cmov`

### DIFF
--- a/src/field.h
+++ b/src/field.h
@@ -307,7 +307,8 @@ static void secp256k1_fe_to_storage(secp256k1_fe_storage *r, const secp256k1_fe 
  */
 static void secp256k1_fe_from_storage(secp256k1_fe *r, const secp256k1_fe_storage *a);
 
-/** If flag is true, set *r equal to *a; otherwise leave it. Constant-time.  Both *r and *a must be initialized.*/
+/** If flag is 1, set *r equal to *a; if flag is 0, leave it. Constant-time.
+ * Both *r and *a must be initialized. Flag must be 0 or 1. */
 static void secp256k1_fe_storage_cmov(secp256k1_fe_storage *r, const secp256k1_fe_storage *a, int flag);
 
 /** Conditionally move a field element in constant time.
--- a/src/field_10x26_impl.h
+++ b/src/field_10x26_impl.h
@@ -1014,6 +1014,7 @@ SECP256K1_INLINE static void secp256k1_fe_impl_sqr(secp256k1_fe *r, const secp25
 SECP256K1_INLINE static void secp256k1_fe_impl_cmov(secp256k1_fe *r, const secp256k1_fe *a, int flag) {
     uint32_t mask0, mask1;
     volatile int vflag = flag;
+    VERIFY_CHECK(flag == 0 || flag == 1);
     SECP256K1_CHECKMEM_CHECK_VERIFY(r->n, sizeof(r->n));
     mask0 = vflag + ~((uint32_t)0);
     mask1 = ~mask0;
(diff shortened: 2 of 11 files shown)


furszy (commented):
Wondering if it worth adding a macro for this; something like `VERIFY_BOOL_ARG`.

real-or-random (approved):
utACK 0a6ebe0b24c247aebfa63ccb0017d4de93e2e5a8

If you want to touch this again: I like "flag must be 0 or 1." because it's so clear. It would be good to add it even to the comments in "If flag is 1..." style.

real-or-random:
> Wondering if it worth adding a macro for this; something like `VERIFY_BOOL_ARG`.

Weak Concept NACK for the reasons laid out in #1779. There, it was a Concept ~0, but here I deem the value below 0 because we don't even save lines. `VERIFY_BOOL_ARG(flag)` is not simpler to read than `VERIFY_CHECK(flag == 0 || flag == 1)`. If anything, it's harder to read because it's less explicit (unless you look up the macro).

If we really want to improve this, then let's switch to C99 and use `bool`. But I'm not sure if I really want to have this discussion. :D 

edit: Sorry, Concept **N**ACK. I had typed "Concept ACK" first.

real-or-random (approved):
utACK 9fa11a76eacbe4774ade813f9c25c2c348cb9086

real-or-random:
> That's probably because `VERIFY_BOOL_ARG` is not explicit/readable enough?

It's just that spelling out the condition, as in `VERIFY_CHECK(flag == 0 || flag == 1)` is, by the meaning of the word, more explicit, and it's still short enough to be read with a single glimpse.  This gives `VERIFY_BOOL_ARG` a tiny disadvantage because readers may need to look it up. But this is not about the name of the macro. I like the name. The name is rather explicit and probably as good as it can get. 

> My broader thought is that argument checks should come from well-named, established helpers at this point, so we don’t have to keep worrying about the raw checks in every function. We should understand what the function does just by reading its name.
>
> Furthermore, if you push me a bit, I’d even say we should aim for a more generic mechanism for argument validation altogether (some sort of decorator), so validating args becomes a no-brainer entirely.

This is a different direction, of course. If we had macros for almost every type (or even some more clever thing, but that seems difficult in C89), then the tiny disadvantage of `VERIFY_BOOL_ARG` would most likely be outweighed by the advantage of having a consistent framework. Curious to see proposals if you have something in mind!

Still, the raw checks are mostly easy to read, and they keep the advantage of being very explicit. So I think *in the current state of the code base*, they make sense.

furszy (commented):
ACK 9fa11a76eacbe4774ade813f9c25c2c348cb9086

It seems you could also do it for `secp256k1_musig_secnonce_invalidate` (at least the comment update).

john-moffett:
Thank you for the reviews. @furszy thanks for the catch; I updated the docs for that function. @hebasto, my impression is that most functions don't have their constant-time-ness included in the function name, eg - `secp256k1_scalar_add`, `secp256k1_scalar_mul`, `secp256k1_scalar_cond_negate`. It seems like only variable-time ones are marked as such with a `_var` suffix. Let me know if you had something else in mind, though.

PR #1784, opened 2025-12-09
refactor: remove ret from secp256k1_ec_pubkey_serialize

kevkevinpal:
This is a follow-up to https://github.com/bitcoin-core/secp256k1/pull/1774#discussion_r2539737079

It is pretty straightforward to remove `ret` and to just return either `0` or `1`


real-or-random (approved):
utACK 3daab83a6050805763da5ac9ebe8217d1965eb0c

Issue #1723, opened 2025-08-12
(RFC) Have CMake generate `vector.h` files for test vectors


josibake:
Opening this as a follow-up to https://github.com/bitcoin-core/secp256k1/pull/1698#discussion_r2268280579

---

Currently, test vector headers are generated from python files automatically as part of the `make` build system. It's particularly nice when developing, as you can just modify the python script and the vector.h file gets regenerated automatically.

I was curious how hard this would be in CMake, so tried to hack something together here: https://github.com/josibake/secp256k1/commit/953953d1f44b08b8ce75525c86eaa7091c3af15b

This works-ish, in that I was able to generate the test vector files in the build directory and verified that changing the python scripts automatically regenerates the files. The main issue is CMake generates the header files out of tree, whereas the makefile generates them in tree. I don't see a clear way to support both? The natural thing in mind would be to have these files be stubs and have generated header files in the build directory (the way that CMake is doing it).

I do think it would be nice to have the build systems consistent, but curious if others thing this is wanted/needed before I spend more time refining it. As mentioned in https://github.com/bitcoin-core/secp256k1/pull/1698#discussion_r2269261738, another perhaps even simpler approach would be to decouple the test vector generation from the build system entirely.

real-or-random:
Concept NACK

> I do think it would be nice to have the build systems consistent,

Hm yes, in a world with infinite resources, yes. But I think if we're pragmatic, then this just isn't worth the effort. I claim it's fine if you are required to use autotools if you are a developer changing the test vectors. (And it's not even true: You can still use CMake and run the scripts manually.)

josibake:
> But I think if we're pragmatic, then this just isn't worth the effort.

Agree, especially since I don't see a clean way to support both when autotools wants to modify the files in place, but CMake would want to have the generated files out of the source tree. Thanks for chiming in!

real-or-random:
> > But I think if we're pragmatic, then this just isn't worth the effort.
> 
> Agree, especially since I don't see a clean way to support both when autotools wants to modify the files in place, but CMake would want to have the generated files out of the source tree. Thanks for chiming in!

Sorry for the off-topic comment, but you may be surprised to learn that autotools support out-of-tree builds as well: https://www.gnu.org/software/automake/manual/html_node/VPATH-Builds.html

PR #1786, opened 2025-12-12
musig: add input JSON files for generating test vectors

theStack:
While looking up past review history of the previous silentpayments PR (take 3), I've found that adding the input JSON files for generating the musig test vectors was suggested once there: https://github.com/bitcoin-core/secp256k1/pull/1698#issuecomment-3284849165

Can be tested by first comparing that the .json files match the ones from the bips repository's [`bip-0327/vectors`](https://github.com/bitcoin/bips/tree/master/bip-0327/vectors) folder and then running e.g.
```
$ rm src/modules/musig/vectors.h
$ make src/modules/musig/vectors.h
$ git diff
<should not show any diff>
```


real-or-random (commented):
Concept ACK

kevkevinpal:
Concept ACK 

---

I was also able to run these commands and get the same results
```
$ rm src/modules/musig/vectors.h
$ make src/modules/musig/vectors.h
$ git diff
<should not show any diff>
```

real-or-random (approved):
ACK 12bba6386913619be81df1dac3f7faa910186510

siv2r (commented):
tACK 12bba6386913619be81df1dac3f7faa910186510, verified that the python script regenrates the same vectors.

hebasto:
> ```
> $ rm src/modules/musig/vectors.h
> $ make src/modules/musig/vectors.h
> $ git diff
> <should not show any diff>
> ```

Same for CMake?

hebasto:
> ```
> $ rm src/modules/musig/vectors.h
> $ make src/modules/musig/vectors.h
> $ git diff
> <should not show any diff>
> ```

It would be helpful to align this workflow with existing patterns.

For `src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.h`, we currently run:
```
$ make clean-testvectors
$ make testvectors
```

We should probably use the same approach for the `musig` module.

hebasto:
> While looking up past review history of the previous silentpayments PR (take 3), I've found that adding the input JSON files for generating the musig test vectors was suggested once there: [#1698 (comment)](https://github.com/bitcoin-core/secp256k1/pull/1698#issuecomment-3284849165)

The comment mentions doing this "without further dependencies". Note that `python3` is not technically a _new_ dependency, as it is already used in the build system: https://github.com/bitcoin-core/secp256k1/blob/2d9137ce9d7e576bfada2859292ecaa0887a6b8c/Makefile.am#L260-L266

However, directly invoking `python3` is not portable and may fail on some systems. The Autotools-based build system should use the configured `$(PYTHON)` variable (or similar) instead.

theStack:
@hebasto: Thanks for reviewing!

> Same for CMake?

Adding support for test vectors generation with CMake has been proposed before, but was rejected, see #1723.

> It would be helpful to align this workflow with existing patterns.
> 
> For src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.h, we currently run:
> 
> $ make clean-testvectors
> $ make testvectors
>
> We should probably use the same approach for the musig module.

Currently, the `{clean-,}testvectors` targets build all the available test vectors, include the musig one in this PR branch (if the module is enabled). I guess that's fine and separate targets only for musig are not needed?

> However, directly invoking python3 is not portable and may fail on some systems. The Autotools-based build system should use the configured $(PYTHON) variable (or similar) instead.

Since that affects all existing test vectors generation targets and not only the musig one, that seems to be a topic for a different PR. (Of course, also happy to include a commit directly in this PR, if its seen as uncontroversial and reviewers prefer that.)

hebasto (commented):
While testing this PR by running
```
$ make clean-testvectors; make testvectors
```
I expected to get a non-zero diff after arbitrarily modifying the JSON files. However, that's not the case. For example, the following diff does not result in changes to the generated header:
```diff
--- a/src/modules/musig/vectors/det_sign_vectors.json
+++ b/src/modules/musig/vectors/det_sign_vectors.json
@@ -26,7 +26,7 @@
         },
         {
             "rand": null,
-            "aggothernonce": "0337C87821AFD50A8644D820A8F3E02E499C931865C2360FB43D0A0D20DAFE07EA0287BF891D2A6DEAEBADC909352AA9405D1428C15F4B75F04DAE642A95C2548480",
+            "aggothernonce": "1337C87821AFD50A8644D820A8F3E02E499C931865C2360FB43D0A0D20DAFE07EA0287BF891D2A6DEAEBADC909352AA9405D1428C15F4B75F04DAE642A95C2548480",
             "key_indices": [1, 0, 2],
             "tweaks": [],
             "is_xonly": [],
```

Are those JSON files redundant?

theStack:
@hebasto: Good catch, looks like currently only six out of the eight .json files from the bips repository are used for test vectors generation: 
```
$ git grep "\.json" ./tools/test_vectors_musig2_generate.py
tools/test_vectors_musig2_generate.py:with open(sys.argv[1] + "/key_agg_vectors.json", "r") as f:
tools/test_vectors_musig2_generate.py:with open(sys.argv[1] + "/nonce_gen_vectors.json", "r") as f:
tools/test_vectors_musig2_generate.py:with open(sys.argv[1] + "/nonce_agg_vectors.json", "r") as f:
tools/test_vectors_musig2_generate.py:with open(sys.argv[1] + "/sign_verify_vectors.json", "r") as f:
tools/test_vectors_musig2_generate.py:with open(sys.argv[1] + "/tweak_vectors.json", "r") as f:
tools/test_vectors_musig2_generate.py:with open(sys.argv[1] + "/sig_agg_vectors.json", "r") as f:
```
Hence, the other two (`det_sign_vectors.json` and `key_sort_vectors.json`) are indeed not needed and could be removed. On the other hand, keeping the folder in sync with https://github.com/bitcoin/bips/tree/master/bip-0327/vectors seems to be more consistent. Happy to go either way, curious to hear what maintainers and reviewers prefer.

theStack:
Rebased on master (resolving the conflict in Makefile.am since #1765 was merged).

PR #1788, opened 2025-12-14
test: split monolithic ellswift test into independent cases

furszy:
No behavior changes.

Refactors the previously monolithic ElligatorSwift test into isolated,
independent test cases. Doing so allows the test suite to execute
these cases in parallel rather than sequentially.

Overall, I'm seeing 35-40% tests time reduction locally.

This is quite useful for Debug builds with no optimizations,
which are noticeably slow.

####  Local Debug-build Results (7 jobs):

- master: 138.0 seconds.
- this PR: 89.3 seconds.
   (~1.55× speedup, ~35% reduction)

####  Local Release-build Results (7 jobs):

- master: 9.5 seconds.
- this PR: 5.9 seconds.
   (~1.61× speedup, ~38% reduction)

####  rp5 Release-build results (5 jobs):
- master: 39 seconds.
- this PR: 24 seconds.
  (~1.62× speedup, ~38% reduction)

### DIFF
--- a/src/modules/ellswift/tests_impl.h
+++ b/src/modules/ellswift/tests_impl.h
@@ -177,9 +177,9 @@ static int ellswift_xdh_hash_x32(unsigned char *output, const unsigned char *x32
     return 1;
 }
 
-void run_ellswift_tests(void) {
-    int i = 0;
-    /* Test vectors. */
+/* Run the test vectors for ellswift encoding */
+void ellswift_encoding_test_vectors_tests(void) {
+    int i;
     for (i = 0; (unsigned)i < sizeof(ellswift_xswiftec_inv_tests) / sizeof(ellswift_xswiftec_inv_tests[0]); ++i) {
         const struct ellswift_xswiftec_inv_test *testcase = &ellswift_xswiftec_inv_tests[i];
         int c;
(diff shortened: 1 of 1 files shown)


src/modules/ellswift/tests_impl.h
```diff
@@ -293,6 +315,10 @@ void run_ellswift_tests(void) {
         /* Compare. */
         CHECK(fe_equal(&res.x, &share_x));
     }
+}
+
+void test_ellswift_xdh_correctness(void) {
```
real-or-random: consistency nit if you touch again:
```suggestion
void ellswift_xdh_correctness_tests(void) {
```

furszy:
> Could we re-arrange test cases to run the longest ones first?

We could do it, but it wouldn’t make much of a difference atm. The ellswift module sits deep in the test registry ([see](https://github.com/bitcoin-core/secp256k1/blob/5c75183344585baee5de74c771099d13613bf7dc/src/tests.c#L7802)). We should move the module up too. Which is really a small change.

Yet, it might be worth considering adding `cost` score functionality to the internal framework, so test execution can be automatically arranged based on it.

src/modules/ellswift/tests_impl.h
furszy: upps, fixed. Thanks!

PR #1789, opened 2025-12-15
ecmult_multi: Replace scratch space with malloc, use abcd cost model

fjahr:
This is an implementation of the discussed changes from an in-person meeting in October. It removes usage of scratch space in batch validation and replaces it with internal malloc usage. It also adds an ABCD cost model for algorithm selection.

The API and internals follow the drafted spec from the meeting very closely: https://gist.github.com/fjahr/c2a009487dffe7a1fbf17ca1821976ca There are few minor changes that should not change the intended behavior. The test coverage may be currently a bit lower than it was previously. I am guessing an adapted form of the `test_ecmult_multi` test should be added back and there are some TODOs left in the test code which I am planning to address after a first round of conceptual feedback.

The second commit demonstrates the calibration tooling that I have been using though it's not exactly what has given me the results that are in the PR. I have still been struggling with the calibration code and seem to never really get a result that just works without manual tweaking. The calibration itself as well as the code added there thus is rather a work in progress. I am assuming some version of the calibration code should be added to the repo and I haven't thought much about what the best place to add it is. Putting it into the benchmark and combining it with the python script was just a convenient way for experimentation. I am very open to suggestions on how to change this.

### DIFF
--- a/src/bench_ecmult.c
+++ b/src/bench_ecmult.c
@@ -16,6 +16,7 @@
 #include "scalar_impl.h"
 #include "ecmult_impl.h"
 #include "bench.h"
+#include "tests_common.h"
 
 #define POINTS 32768
 
@@ -68,6 +69,142 @@ typedef struct {
     secp256k1_gej* output;
     secp256k1_fe* output_xonly;
 } bench_data;
+
+/*
+ * ABCD Calibration Benchmarks
+ *
+ * Measures the performance of each algorithm at various batch sizes and
+ * outputs. Use tools/ecmult_multi_calib.py to calculate optimal C and D
+ * values from the output.
+ *
+ * Each algorithm is only calibrated within its optimal batch size range
+ * to avoid skewing results with uninteresting results far away from that
+ * range.
+ */
+static void run_ecmult_multi_calib(bench_data* data) {
+    static const size_t batch_sizes[] = {
+        /* Small numbers should help stabilize Strauss intercept */
+        2, 3, 5, 7, 10, 15, 20, 30, 50, 70,
+        /* Crossover region between Strauss and Pippenger */
+        85, 88, 90, 100, 120, 150, 175,
+        /* Pippenger windows, getting progressively larger */
+        200, 300, 500, 750, 1000, 1200, 1500, 2000, 3000, 5000, 7500, 10000, 15000, 20000, 30000
+    };
+    static const size_t n_batch_sizes = sizeof(batch_sizes) / sizeof(batch_sizes[0]);
+
+    static const char* algo_names[] = {
+        "TRIVIAL", "STRAUSS", "PIPPENGER_1", "PIPPENGER_2", "PIPPENGER_3", "PIPPENGER_4", "PIPPENGER_5", "PIPPENGER_6", "PIPPENGER_7", "PIPPENGER_8", "PIPPENGER_9", "PIPPENGER_10", "PIPPENGER_11", "PIPPENGER_12"
+    };
+
+    /* Maximum batch size for Strauss calibration. */
+    static const size_t STRAUSS_MAX_CALIB_BATCH = 500;
+
+    /* Per-window min/max batch sizes for Pippenger calibration. */
+    static const size_t pippenger_min_calib_batch[12] = {
+    /*w=1   2   3   4   5   6    7    8    9     10    11    12 */
+        5,  5,  10, 30, 70, 150, 300, 750, 1500, 3000, 7500, 15000
+    };
+    static const size_t pippenger_max_calib_batch[12] = {
+    /*w=1    2    3    4     5     6     7      8      9      10     11     12 */
+        100, 200, 500, 1000, 2000, 5000, 10000, 20000, 30000, 30000, 30000, 30000
+    };
+
+    secp256k1_ge *points = NULL;
+    secp256k1_scalar *scalars = NULL;
+    secp256k1_gej result;
+    size_t max_points = batch_sizes[n_batch_sizes - 1];
+    int algo;
+    size_t i, j;
+    int base_iters = 1000;
+
+    points = (secp256k1_ge *)malloc(max_points * sizeof(secp256k1_ge));
+    scalars = (secp256k1_scalar *)malloc(max_points * sizeof(secp256k1_scalar));
+    CHECK(points != NULL);
+    CHECK(scalars != NULL);
+
+    for (i = 0; i < max_points; i++) {
+        points[i] = data->pubkeys[i % POINTS];
+        scalars[i] = data->scalars[i % POINTS];
+    }
+
+    printf("# ECMULT_MULTI Calibration Data\n");
+    printf("# Format: ALGO,N,TIME_US (microseconds per batch)\n");
+    printf("# Copy the DATA section below into the Python script\n");
+    printf("#\n");
+    printf("# BEGIN DATA\n");
+
+    /* Measure STRAUSS */
+    algo = SECP256K1_ECMULT_MULTI_ALGO_STRAUSS;
+    for (i = 0; i < n_batch_sizes; i++) {
+        size_t n = batch_sizes[i];
+        int64_t t_start, t_end;
+        double time_us;
+        int iters = base_iters;
+        int iter;
+
+        /* Only run up to the max to not skew result */
+        if (n > STRAUSS_MAX_CALIB_BATCH) continue;
+
+        /* Using many iterations in Strauss since batch sizes are small */
+        if (n >= 300) iters = base_iters / 2;
+        if (iters < 100) iters = 100;
+
+        t_start = gettime_i64();
+        for (iter = 0; iter < iters; iter++) {
+            CHECK(secp256k1_ecmult_multi_internal(&data->ctx->error_callback, algo,
+                                                  &result, n, points, scalars, NULL));
+        }
+        t_end = gettime_i64();
+
+        time_us = (double)(t_end - t_start) / iters;
+        printf("%s,%lu,%.3f\n", algo_names[algo], (unsigned long)n, time_us);
+    }
+
+    /* Measure PIPPENGER variants */
+    for (algo = SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_1;
+         algo <= SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_12;
+         algo++) {
+        int window = algo - SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_1;
+        size_t min_batch = pippenger_min_calib_batch[window];
+        size_t max_batch = pippenger_max_calib_batch[window];
+
+        for (j = 0; j < n_batch_sizes; j++) {
+            size_t n = batch_sizes[j];
+            int64_t t_start, t_end;
+            double time_us;
+            int iters = base_iters;
+            int iter;
+
+            /* Only run for the selected range of each algo */
+            if (n < min_batch || n > max_batch) continue;
+
+            /* Limiting iterations to keep run-time managable */
+            if (n >= 1000) iters = base_iters / 10;
+            if (n >= 5000) iters = base_iters / 50;
+            if (n >= 15000) iters = base_iters / 100;
+            if (iters < 3) iters = 3;
+
+            t_start = gettime_i64();
+            for (iter = 0; iter < iters; iter++) {
+                CHECK(secp256k1_ecmult_multi_internal(&data->ctx->error_callback, algo,
+                                                      &result, n, points, scalars, NULL));
+            }
+            t_end = gettime_i64();
+
+            time_us = (double)(t_end - t_start) / iters;
+            printf("%s,%lu,%.3f\n", algo_names[algo], (unsigned long)n, time_us);
+        }
+    }
+
+    printf("# END DATA\n");
+    printf("#\n");
+    printf("# To calculate ABCD constants, run:\n");
+    printf("#   ./bench_ecmult calib 2>&1 | python3 tools/ecmult_multi_calib.py\n");
+    printf("#\n");
+
+    free(points);
+    free(scalars);
+}
 
 /* Hashes x into [0, POINTS) twice and store the result in offset1 and offset2. */
 static void hash_into_offset(bench_data* data, size_t x) {
@@ -338,6 +475,7 @@ static void run_ecmult_multi_bench(bench_data* data, size_t count, int includes_
 int main(int argc, char **argv) {
     bench_data data;
     int i, p;
+    int run_calib = 0;
 
     int default_iters = 10000;
     int iters = get_iters(default_iters);
@@ -368,6 +506,8 @@ int main(int argc, char **argv) {
         } else if(have_flag(argc, argv, "auto")) {
             printf("Using automatic algorithm selection:\n");
             data.forced_algo = BENCH_ALGO_AUTO;
+        } else if(have_flag(argc, argv, "calib")) {
+            run_calib = 1;
         } else {
             fprintf(stderr, "%s: unrecognized argument '%s'.\n\n", argv[0], argv[1]);
             help(argv[0], default_iters);
@@ -398,27 +538,30 @@ int main(int argc, char **argv) {
     }
     secp256k1_ge_set_all_gej_var(data.pubkeys, data.pubkeys_gej, POINTS);
 
-
-    print_output_table_header_row();
-    /* Initialize offset1 and offset2 */
-    hash_into_offset(&data, 0);
-    run_ecmult_bench(&data, iters);
-
-    for (i = 1; i <= 8; ++i) {
-        run_ecmult_multi_bench(&data, i, 1, iters);
-    }
-
-    /* This is disabled with low count of iterations because the loop runs 77 times even with iters=1
-    * and the higher it goes the longer the computation takes(more points)
-    * So we don't run this benchmark with low iterations to prevent slow down */
-     if (iters > 2) {
-        for (p = 0; p <= 11; ++p) {
-            for (i = 9; i <= 16; ++i) {
-                run_ecmult_multi_bench(&data, i << p, 1, iters);
-            }
-        }
+    if (run_calib) {
+        run_ecmult_multi_calib(&data);
     } else {
-        printf("Skipping some benchmarks due to SECP256K1_BENCH_ITERS <= 2\n");
+        print_output_table_header_row();
+        /* Initialize offset1 and offset2 */
+        hash_into_offset(&data, 0);
+        run_ecmult_bench(&data, iters);
+
+        for (i = 1; i <= 8; ++i) {
+            run_ecmult_multi_bench(&data, i, 1, iters);
+        }
+
+        /* This is disabled with low count of iterations because the loop runs 77 times even with iters=1
+        * and the higher it goes the longer the computation takes(more points)
+        * So we don't run this benchmark with low iterations to prevent slow down */
+        if (iters > 2) {
+            for (p = 0; p <= 11; ++p) {
+                for (i = 9; i <= 16; ++i) {
+                    run_ecmult_multi_bench(&data, i << p, 1, iters);
+                }
+            }
+        } else {
+            printf("Skipping some benchmarks due to SECP256K1_BENCH_ITERS <= 2\n");
+        }
     }
 
     secp256k1_context_destroy(data.ctx);
--- a/src/bench_ecmult.c
+++ b/src/bench_ecmult.c
@@ -19,6 +19,11 @@
 
 #define POINTS 32768
 
+/* Default memory limit (64 MB) */
+#define DEFAULT_MEM_LIMIT (64 * 1024 * 1024)
+/* Select bench algorithm automatically */
+#define BENCH_ALGO_AUTO (-1)
+
 static void help(const char *executable_path, int default_iters) {
     printf("Benchmark EC multiplication algorithms\n");
     printf("\n");
@@ -30,23 +35,25 @@ static void help(const char *executable_path, int default_iters) {
     printf("function name. The letter 'g' indicates that one of the points is the generator.\n");
     printf("The benchmarks are divided by the number of points.\n");
     printf("\n");
-    printf("default (ecmult_multi): picks pippenger_wnaf or strauss_wnaf depending on the\n");
-    printf("                        batch size\n");
-    printf("pippenger_wnaf:         for all batch sizes\n");
-    printf("strauss_wnaf:           for all batch sizes\n");
-    printf("simple:                 multiply and sum each point individually\n");
+    printf("default (auto):   automatically select best algorithm\n");
+    printf("pippenger_wnaf:   for all batch sizes\n");
+    printf("strauss_wnaf:     for all batch sizes\n");
+    printf("simple:           multiply and sum each point individually\n");
+    printf("\n");
 }
 
 typedef struct {
     /* Setup once in advance */
     secp256k1_context* ctx;
-    secp256k1_scratch_space* scratch;
     secp256k1_scalar* scalars;
     secp256k1_ge* pubkeys;
     secp256k1_gej* pubkeys_gej;
     secp256k1_scalar* seckeys;
     secp256k1_gej* expected_output;
-    secp256k1_ecmult_multi_func ecmult_multi;
+
+    /* Algorithm selection */
+    int forced_algo;
+    size_t mem_limit;
 
     /* Changes per benchmark */
     size_t count;
@@ -214,32 +221,54 @@ static void run_ecmult_bench(bench_data* data, int iters) {
     run_benchmark(str, bench_ecmult_1p_g, bench_ecmult_setup, bench_ecmult_1p_g_teardown, data, 10, 2*iters);
 }
 
-static int bench_ecmult_multi_callback(secp256k1_scalar* sc, secp256k1_ge* ge, size_t idx, void* arg) {
-    bench_data* data = (bench_data*)arg;
-    if (data->includes_g) ++idx;
-    if (idx == 0) {
-        *sc = data->scalars[data->offset1];
-        *ge = secp256k1_ge_const_g;
-    } else {
-        *sc = data->scalars[(data->offset1 + idx) % POINTS];
-        *ge = data->pubkeys[(data->offset2 + idx - 1) % POINTS];
-    }
-    return 1;
-}
-
 static void bench_ecmult_multi(void* arg, int iters) {
     bench_data* data = (bench_data*)arg;
 
     int includes_g = data->includes_g;
     int iter;
     int count = data->count;
+    size_t n_points = count - includes_g;
+    secp256k1_ecmult_multi_algo algo;
+    secp256k1_ge *points = NULL;
+    secp256k1_scalar *scalars = NULL;
+    size_t i;
     iters = iters / data->count;
 
+    if (n_points > 0) {
+        points = (secp256k1_ge *)malloc(n_points * sizeof(secp256k1_ge));
+        scalars = (secp256k1_scalar *)malloc(n_points * sizeof(secp256k1_scalar));
+        CHECK(points != NULL);
+        CHECK(scalars != NULL);
+    }
+
     for (iter = 0; iter < iters; ++iter) {
-        data->ecmult_multi(&data->ctx->error_callback, data->scratch, &data->output[iter], data->includes_g ? &data->scalars[data->offset1] : NULL, bench_ecmult_multi_callback, arg, count - includes_g);
+        const secp256k1_scalar *g_scalar_ptr = NULL;
+
+        if (includes_g) {
+            g_scalar_ptr = &data->scalars[data->offset1];
+        }
+
+        for (i = 0; i < n_points; ++i) {
+            size_t idx = includes_g ? i + 1 : i;
+            scalars[i] = data->scalars[(data->offset1 + idx) % POINTS];
+            points[i] = data->pubkeys[(data->offset2 + i) % POINTS];
+        }
+
+        if (data->forced_algo >= 0) {
+            algo = data->forced_algo;
+        } else {
+            algo = secp256k1_ecmult_multi_select(data->mem_limit, n_points);
+        }
+
+        CHECK(secp256k1_ecmult_multi_internal(&data->ctx->error_callback, algo, &data->output[iter],
+                                              n_points, points, scalars, g_scalar_ptr));
+
         data->offset1 = (data->offset1 + count) % POINTS;
         data->offset2 = (data->offset2 + count - 1) % POINTS;
     }
+
+    free(points);
+    free(scalars);
 }
 
 static void bench_ecmult_multi_setup(void* arg) {
@@ -309,7 +338,6 @@ static void run_ecmult_multi_bench(bench_data* data, size_t count, int includes_
 int main(int argc, char **argv) {
     bench_data data;
     int i, p;
-    size_t scratch_size;
 
     int default_iters = 10000;
     int iters = get_iters(default_iters);
@@ -318,7 +346,8 @@ int main(int argc, char **argv) {
         return EXIT_FAILURE;
     }
 
-    data.ecmult_multi = secp256k1_ecmult_multi_var;
+    data.forced_algo = BENCH_ALGO_AUTO;
+    data.mem_limit = DEFAULT_MEM_LIMIT;
 
     if (argc > 1) {
         if(have_flag(argc, argv, "-h")
@@ -328,12 +357,17 @@ int main(int argc, char **argv) {
             return EXIT_SUCCESS;
         } else if(have_flag(argc, argv, "pippenger_wnaf")) {
             printf("Using pippenger_wnaf:\n");
-            data.ecmult_multi = secp256k1_ecmult_pippenger_batch_single;
+            /* TODO: Make this a dynamic selection again */
+            data.forced_algo = SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_4;
         } else if(have_flag(argc, argv, "strauss_wnaf")) {
             printf("Using strauss_wnaf:\n");
-            data.ecmult_multi = secp256k1_ecmult_strauss_batch_single;
+            data.forced_algo = SECP256K1_ECMULT_MULTI_ALGO_STRAUSS;
         } else if(have_flag(argc, argv, "simple")) {
             printf("Using simple algorithm:\n");
+            data.forced_algo = SECP256K1_ECMULT_MULTI_ALGO_TRIVIAL;
+        } else if(have_flag(argc, argv, "auto")) {
+            printf("Using automatic algorithm selection:\n");
+            data.forced_algo = BENCH_ALGO_AUTO;
         } else {
             fprintf(stderr, "%s: unrecognized argument '%s'.\n\n", argv[0], argv[1]);
             help(argv[0], default_iters);
@@ -342,12 +376,6 @@ int main(int argc, char **argv) {
     }
 
     data.ctx = secp256k1_context_create(SECP256K1_CONTEXT_NONE);
-    scratch_size = secp256k1_strauss_scratch_size(POINTS) + STRAUSS_SCRATCH_OBJECTS*ALIGNMENT;
-    if (!have_flag(argc, argv, "simple")) {
-        data.scratch = secp256k1_scratch_space_create(data.ctx, scratch_size);
-    } else {
-        data.scratch = NULL;
-    }
 
     /* Allocate stuff */
     data.scalars = malloc(sizeof(secp256k1_scalar) * POINTS);
@@ -393,9 +421,6 @@ int main(int argc, char **argv) {
         printf("Skipping some benchmarks due to SECP256K1_BENCH_ITERS <= 2\n");
     }
 
-    if (data.scratch != NULL) {
-        secp256k1_scratch_space_destroy(data.ctx, data.scratch);
-    }
     secp256k1_context_destroy(data.ctx);
     free(data.scalars);
     free(data.pubkeys);
--- a/src/ecmult.h
+++ b/src/ecmult.h
@@ -9,7 +9,6 @@
 
 #include "group.h"
 #include "scalar.h"
-#include "scratch.h"
 
 #ifndef ECMULT_WINDOW_SIZE
 #  define ECMULT_WINDOW_SIZE 15
@@ -46,19 +45,93 @@
  */
 static void secp256k1_ecmult(secp256k1_gej *r, const secp256k1_gej *a, const secp256k1_scalar *na, const secp256k1_scalar *ng);
 
-typedef int (secp256k1_ecmult_multi_callback)(secp256k1_scalar *sc, secp256k1_ge *pt, size_t idx, void *data);
+/**
+ * Algorithm identifiers for multi-scalar multiplication.
+ *
+ * TRIVIAL:      Simple algorithm, no extra memory needed
+ * STRAUSS:      Strauss algorithm (efficient for small batches)
+ * PIPPENGER_n:  Pippenger algorithm with bucket window size n
+ */
+typedef enum {
+    SECP256K1_ECMULT_MULTI_ALGO_TRIVIAL = 0,
+    SECP256K1_ECMULT_MULTI_ALGO_STRAUSS = 1,
+    SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_1 = 2,
+    SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_2 = 3,
+    SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_3 = 4,
+    SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_4 = 5,
+    SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_5 = 6,
+    SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_6 = 7,
+    SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_7 = 8,
+    SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_8 = 9,
+    SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_9 = 10,
+    SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_10 = 11,
+    SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_11 = 12,
+    SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_12 = 13
+} secp256k1_ecmult_multi_algo;
+
+#define SECP256K1_ECMULT_MULTI_NUM_ALGOS 14
 
 /**
- * Multi-multiply: R = inp_g_sc * G + sum_i ni * Ai.
- * Chooses the right algorithm for a given number of points and scratch space
- * size. Resets and overwrites the given scratch space. If the points do not
- * fit in the scratch space the algorithm is repeatedly run with batches of
- * points. If no scratch space is given then a simple algorithm is used that
- * simply multiplies the points with the corresponding scalars and adds them up.
- * Returns: 1 on success (including when inp_g_sc is NULL and n is 0)
- *          0 if there is not enough scratch space for a single point or
- *          callback returns 0
+ * Calculate max batch size for a given memory limit.
+ *
+ * For each algorithm, memory usage is modeled as m(x) = A*x + B and
+ * running time as c(x) = C*x + D, where x is the batch size. This
+ * function finds the algorithm that minimizes time per operation
+ * C + D/x at the maximum batch size x = (mem_limit - B) / A.
+ *
+ * Note: The algorithm chosen here may differ from what
+ * secp256k1_ecmult_multi_select returns for the same mem_limit and
+ * batch_size returned for this mem_limit here. This can happend
+ * because _select evaluates all algorithms at the given batch_size
+ * only.
+ *
+ * Returns: The optimal batch size. Falls back to ECMULT_MAX_POINTS_PER_BATCH
+ * using the trivial algorithm when no Strauss/Pippenger algorithm fits.
  */
-static int secp256k1_ecmult_multi_var(const secp256k1_callback* error_callback, secp256k1_scratch *scratch, secp256k1_gej *r, const secp256k1_scalar *inp_g_sc, secp256k1_ecmult_multi_callback cb, void *cbdata, size_t n);
+static size_t secp256k1_ecmult_multi_batch_size(size_t mem_limit);
+
+/**
+ * Select the best algorithm for a given batch size within the memory
+ * limit.
+ *
+ * Among algorithms that fit within mem_limit for the given batch_size,
+ * selects the one that minimizes time per operation C + D/batch_size.
+ *
+ * Returns: The optimal algorithm identifier.
+ */
+static secp256k1_ecmult_multi_algo secp256k1_ecmult_multi_select(
+    size_t mem_limit,
+    size_t batch_size
+);
+
+/**
+ * Multi-multiply: R = scalar_g * G + sum_i scalars[i] * points[i].
+ *
+ * Chooses the best algorithm for the given number of points and
+ * mem_limit. If mem_limit is too small to process all points at once,
+ * splits into batches and accumulates results.
+ *
+ * Returns: 1 on success, 0 on memory allocation failure.
+ */
+static int secp256k1_ecmult_multi(
+    const secp256k1_callback *error_callback,
+    secp256k1_gej *r,
+    size_t n_points,
+    const secp256k1_ge *points,
+    const secp256k1_scalar *scalars,
+    const secp256k1_scalar *scalar_g,
+    size_t mem_limit
+);
+
+/* Only for benchmarks and testing */
+static int secp256k1_ecmult_multi_internal(
+    const secp256k1_callback *error_callback,
+    secp256k1_ecmult_multi_algo algo,
+    secp256k1_gej *r,
+    size_t n_points,
+    const secp256k1_ge *points,
+    const secp256k1_scalar *scalars,
+    const secp256k1_scalar *scalar_g
+);
 
 #endif /* SECP256K1_ECMULT_H */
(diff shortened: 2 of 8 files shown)


src/modules/musig/keyagg_impl.h
```diff
@@ -149,26 +149,13 @@ typedef struct {
...
-     * `musig_compute_pks_hash` (and we test this). */
-    VERIFY_CHECK(ret);
-#else
-    (void) ret;
-#endif
-    secp256k1_musig_keyaggcoef_internal(sc, ctx->pks_hash, pt, &ctx->second_pk);
-    return 1;
-}
-
 int secp256k1_musig_pubkey_agg(const secp256k1_context* ctx, secp256k1_xonly_pubkey *agg_pk, secp256k1_musig_keyagg_cache *keyagg_cache, const secp256k1_pubkey * const* pubkeys, size_t n_pubkeys) {
```
siv2r: It might be better to pass `mem_limit` as an input argument rather than defining it internally

src/modules/musig/keyagg_impl.h
```diff
@@ -199,14 +183,36 @@ int secp256k1_musig_pubkey_agg(const secp256k1_context* ctx, secp256k1_xonly_pub
...
+        VERIFY_CHECK(secp256k1_pubkey_load(ctx, &points[i], pubkeys[i]));
+#else
+        (void) secp256k1_pubkey_load(ctx, &points[i], pubkeys[i]);
+#endif
+        secp256k1_musig_keyaggcoef_internal(&scalars[i], ecmult_data.pks_hash, &points[i], &ecmult_data.second_pk);
+    }
+
+    /* TODO: Previously this just used TRIVIAL algorithm by not providing
+     * scratch space. To be discussed if this should be changed. */
+    mem_limit = n_pubkeys * (sizeof(secp256k1_ge) + sizeof(secp256k1_scalar)) * 4;
```
siv2r: nit: Why the constant 4? I assume it covers internal overhead. I'm wondering whether we can define a ratio `k` such that `mem_limit = n_points * (sizeof(secp256k1_ge) + sizeof(secp256k1_scalar)) * k` yields a batch size of `n_points` when given to `_ecmult_multi_batch_size()`

siv2r: Hmm, we could probably do `mem_limit = max(STRAUSS_POINT_SIZE * n, PIPPENGER_POINT_SIZE * n + PIPPENGER_FIXED_SIZE)`. Just thinking out loud, not suggesting a change here.

src/tests_exhaustive.c
```diff
@@ -186,17 +186,13 @@ typedef struct {
     secp256k1_ge pt[2];
 } ecmult_multi_data;
```
siv2r: We can probably remove this data structure. In the old code, I only see this acting as `cbdata` (initialized using local scalars and points arrays) to the `ecmult_multi_var`

src/bench_ecmult.c
```diff
@@ -306,11 +333,11 @@ static void run_ecmult_multi_bench(bench_data* data, size_t count, int includes_
 int main(int argc, char **argv) {
     bench_data data;
     int i, p;
-    size_t scratch_size;
 
     int iters = get_iters(10000);
 
-    data.ecmult_multi = secp256k1_ecmult_multi_var;
+    data.forced_algo = -1;
```
siv2r: nit: We could define `#define BENCH_ALGO_AUTO (-1)` to make this more readable. Then, the check on line 252 would be: `if (data->forced_algo != BENCH_ALGO_AUTO)`

src/ecmult_impl.h
```diff
@@ -657,212 +545,376 @@ SECP256K1_INLINE static void secp256k1_ecmult_endo_split(secp256k1_scalar *s1, s
...
-            secp256k1_scratch_apply_checkpoint(error_callback, scratch, scratch_checkpoint);
-            return 0;
+/* Pippenger per-point memory: 2 entries due to endomorphism */
+#define SECP256K1_PIPPENGER_POINT_SIZE(w) (2 * SECP256K1_PIPPENGER_ENTRY_SIZE(w))
+
+/* Pippenger fixed overhead: buckets + state + 2 extra entries */
+#define SECP256K1_PIPPENGER_FIXED_SIZE(w) \
+    ((sizeof(secp256k1_gej) << (w)) \
+     + sizeof(struct secp256k1_pippenger_state) \
+     + 2 * SECP256K1_PIPPENGER_ENTRY_SIZE(w))
```
siv2r, line 586:
```suggestion
#define SECP256K1_PIPPENGER_FIXED_SIZE(w) \
    ((sizeof(secp256k1_gej) << (w)) \
     + 2 * SECP256K1_PIPPENGER_ENTRY_SIZE(w))
```

We can remove the `pippenger_state` since it's a stack variable. IIUC, the old code allocated this variable on the heap, hence it included it in the size calculation.

src/ecmult_impl.h
```diff
@@ -657,212 +545,376 @@ SECP256K1_INLINE static void secp256k1_ecmult_endo_split(secp256k1_scalar *s1, s
...
+    if (n > 0) {
+        state.aux = (secp256k1_fe *)checked_malloc(error_callback,
+            n * ECMULT_TABLE_SIZE(WINDOW_A) * sizeof(secp256k1_fe));
+        state.pre_a = (secp256k1_ge *)checked_malloc(error_callback,
+            n * ECMULT_TABLE_SIZE(WINDOW_A) * sizeof(secp256k1_ge));
+        state.ps = (struct secp256k1_strauss_point_state *)checked_malloc(error_callback,
+            n * sizeof(struct secp256k1_strauss_point_state));
+        points_gej = (secp256k1_gej *)checked_malloc(error_callback,
+            n * sizeof(secp256k1_gej));
+        scalars_copy = (secp256k1_scalar *)checked_malloc(error_callback,
```
siv2r: Is there a specific reason for copying the scalars here? Since they are managed by the upper layer (like batch verification or MuSig), could we just pass the pointer to `_strauss_wnaf` directly?

src/ecmult_impl.h
```diff
@@ -657,212 +545,376 @@ SECP256K1_INLINE static void secp256k1_ecmult_endo_split(secp256k1_scalar *s1, s
...
+            secp256k1_gej_set_ge(&points_gej[i], &points[i]);
+            scalars_copy[i] = scalars[i];
+        }
+
+        secp256k1_ecmult_strauss_wnaf(&state, r, n, points_gej, scalars_copy, scalar_g);
+    } else {
+        secp256k1_gej tmpj;
+        secp256k1_gej_set_infinity(&tmpj);
+        secp256k1_ecmult(r, &tmpj, &secp256k1_scalar_zero, scalar_g);
+    }
```
siv2r: We can probably remove this else block. The `n=0` case is already handled by the call to `ecmult_strauss_wnaf` in the preceding if block. Plus, `secp256k1_ecmult` calls `strauss_wnaf` internally, so this seems redundant.

src/ecmult_impl.h
```diff
@@ -657,212 +545,376 @@ SECP256K1_INLINE static void secp256k1_ecmult_endo_split(secp256k1_scalar *s1, s
...
+ * ABCD constants for all batch algorithms.
+ *
+ * Index 0    = TRIVIAL (no memory, very slow)
+ * Index 1    = STRAUSS (efficient for small batche sizes)
+ * Index 2-13 = PIPPENGER with window sizes 1-12
+ */
+static const struct secp256k1_ecmult_multi_abcd secp256k1_ecmult_multi_abcds[SECP256K1_ECMULT_MULTI_NUM_ALGOS] = {
+/*  A (per-point bytes)                     B (fixed bytes)                        C        D           */
+    {0,                                     0,                                     1000,    0      },
+    {SECP256K1_STRAUSS_POINT_SIZE,          0,                                     100,     0      },
```
siv2r: `D` shouldn't be 0 here right?

src/bench_ecmult.c
```diff
@@ -59,6 +65,110 @@ typedef struct {
...
 
+/*
+ * ABCD Calibration Benchmarks
+ *
+ * Measures the performance of each algorithm at various batch sizes and
+ * outputs. Use tools/ecmult_multi_calib.py to calculate optimal C and D
+ * values from the output.
+ */
+static void run_ecmult_multi_calib(bench_data* data) {
+    static const size_t batch_sizes[] = {10, 20, 30, 50, 75, 100, 150, 200, 300, 500, 750, 1000, 1500, 2000, 3000, 5000, 7500, 10000, 15000, 20000, 30000};
```
siv2r: I think it woulbe nice to have more batch sizes in the crossover region of strauss and pippenger (~80-150 points). So, something like this:
```
    static const size_t batch_sizes[] = {
        /* Small (Strauss region) */
        5, 10, 15, 20, 30, 50, 70,
        /* Crossover region */
        85, 88, 90, 100, 120, 150, 175,
        /* Medium (Pippenger small windows, w=6..8) */
        200, 300, 500, 750, 1000, 1200,
        /* Large (Pippenger large windows, w=9..12) */
        1500, 2000, 3000, 5000, 7500,
        10000, 15000, 20000, 30000
    };
```

siv2r: We should probably cap the batch size for Strauss calibration (7,500 or less?). At higher sizes, the benchmark results curve and become outliers, which will skew the linear regression model. Since `ecmult_multi_select` won't choose Strauss for large batches anyway, we can avoid these sizes. I've attached a visualization below.

STRAUSS

siv2r:
I ran the calibration tool, and I'm getting the following results. It's very close the exisitng results, uptil `PIPPENGER_8` after which the C value increases for me (this shouldn't happen right?). I'll run this a few more times.
<details>
<summary>My ABCD values</summary>

```
static const struct secp256k1_ecmult_multi_abcd secp256k1_ecmult_multi_abcds[SECP256K1_ECMULT_MULTI_NUM_ALGOS] = {
    {0,                                     0,                                     1000,  0     },
    {SECP256K1_STRAUSS_POINT_SIZE,          0,                                     113,   143  },
    {SECP256K1_PIPPENGER_POINT_SIZE(1),  SECP256K1_PIPPENGER_FIXED_SIZE(1),  199,   303  },
    {SECP256K1_PIPPENGER_POINT_SIZE(2),  SECP256K1_PIPPENGER_FIXED_SIZE(2),  152,   460  },
    {SECP256K1_PIPPENGER_POINT_SIZE(3),  SECP256K1_PIPPENGER_FIXED_SIZE(3),  117,   782  },
    {SECP256K1_PIPPENGER_POINT_SIZE(4),  SECP256K1_PIPPENGER_FIXED_SIZE(4),  100,   1158 },
    {SECP256K1_PIPPENGER_POINT_SIZE(5),  SECP256K1_PIPPENGER_FIXED_SIZE(5),  86,    1837 },
    {SECP256K1_PIPPENGER_POINT_SIZE(6),  SECP256K1_PIPPENGER_FIXED_SIZE(6),  77,    3013 },
    {SECP256K1_PIPPENGER_POINT_SIZE(7),  SECP256K1_PIPPENGER_FIXED_SIZE(7),  72,    4845 },
    {SECP256K1_PIPPENGER_POINT_SIZE(8),  SECP256K1_PIPPENGER_FIXED_SIZE(8),  69,    8775 },
    {SECP256K1_PIPPENGER_POINT_SIZE(9),  SECP256K1_PIPPENGER_FIXED_SIZE(9),  73,    14373},
    {SECP256K1_PIPPENGER_POINT_SIZE(10), SECP256K1_PIPPENGER_FIXED_SIZE(10), 78,    26442},
    {SECP256K1_PIPPENGER_POINT_SIZE(11), SECP256K1_PIPPENGER_FIXED_SIZE(11), 80,    48783},
    {SECP256K1_PIPPENGER_POINT_SIZE(12), SECP256K1_PIPPENGER_FIXED_SIZE(12), 106,   88289},
};
```

</details>

I’ve visualized the benchmarks for all algorithms (see https://github.com/siv2r/secp256k1/commit/311938612268866bd3dd0cf6c1d006e418518b1d for the diagrams). I’m planning to do a deeper dive into the regression model fit to ensure they align with these results. Will share any useful findings here

siv2r:
I did some analysis on the linear regression model for the CD values (see https://github.com/siv2r/secp256k1/commit/b5985a60454e68c7f88bfe7a9d41b89a47f2b530). I basically ran `./bench_ecmult calib` twice. Used the first run to compute the linear regression model, and the second run to test it against new benchmark values.

<details>
<summary>ABCD Values</summary>

```c
static const struct secp256k1_ecmult_multi_abcd secp256k1_ecmult_multi_abcds[SECP256K1_ECMULT_MULTI_NUM_ALGOS] = {
    {0,                                     0,                                     1000,  0     },
    {SECP256K1_STRAUSS_POINT_SIZE,          0,                                     112,   173  },
    {SECP256K1_PIPPENGER_POINT_SIZE(1),  SECP256K1_PIPPENGER_FIXED_SIZE(1),  197,   285  },
    {SECP256K1_PIPPENGER_POINT_SIZE(2),  SECP256K1_PIPPENGER_FIXED_SIZE(2),  152,   480  },
    {SECP256K1_PIPPENGER_POINT_SIZE(3),  SECP256K1_PIPPENGER_FIXED_SIZE(3),  117,   767  },
    {SECP256K1_PIPPENGER_POINT_SIZE(4),  SECP256K1_PIPPENGER_FIXED_SIZE(4),  100,   1167 },
    {SECP256K1_PIPPENGER_POINT_SIZE(5),  SECP256K1_PIPPENGER_FIXED_SIZE(5),  86,    1887 },
    {SECP256K1_PIPPENGER_POINT_SIZE(6),  SECP256K1_PIPPENGER_FIXED_SIZE(6),  78,    3023 },
    {SECP256K1_PIPPENGER_POINT_SIZE(7),  SECP256K1_PIPPENGER_FIXED_SIZE(7),  73,    4906 },
    {SECP256K1_PIPPENGER_POINT_SIZE(8),  SECP256K1_PIPPENGER_FIXED_SIZE(8),  70,    8889 },
    {SECP256K1_PIPPENGER_POINT_SIZE(9),  SECP256K1_PIPPENGER_FIXED_SIZE(9),  74,    14544},
    {SECP256K1_PIPPENGER_POINT_SIZE(10), SECP256K1_PIPPENGER_FIXED_SIZE(10), 79,    26764},
    {SECP256K1_PIPPENGER_POINT_SIZE(11), SECP256K1_PIPPENGER_FIXED_SIZE(11), 84,    49179},
    {SECP256K1_PIPPENGER_POINT_SIZE(12), SECP256K1_PIPPENGER_FIXED_SIZE(12), 106,   89518},
};
```

</details>


<details>
<summary>Statistical analysis </summary>

Metric Definitions

| Metric | Meaning | Good Value for Benchmarking |
|--------|---------|-----------------------------|
| R² | How well the linear model fits (0-1) | > 0.95 (strong linear relationship) |
| Std Error | Uncertainty in slope estimate | Low relative to slope |
| p-value | Probability slope = 0 by chance | < 0.05 (statistically significant) |
| Slope | Time increase per additional n (μs) | Positive, algorithm-dependent |
| Intercept | Fixed overhead time (μs) | Algorithm-dependent |

Per-Algorithm Metrics

| Algorithm | R² | Std Error | p-value | Slope | Intercept |
|-----------|------|-----------|---------|-------|----------|
| STRAUSS | 0.983536 | 0.1660 | 1.29e-25 | 6.2956 | -1917.23 |
| PIPPENGER_1 | 0.999995 | 0.0039 | 6.29e-73 | 8.5860 | -133.20 |
| PIPPENGER_2 | 0.999992 | 0.0036 | 9.98e-71 | 6.7066 | -115.04 |
| PIPPENGER_3 | 0.999993 | 0.0026 | 8.15e-71 | 4.9746 | -34.11 |
| PIPPENGER_4 | 0.999996 | 0.0015 | 5.34e-75 | 4.1772 | -13.27 |
| PIPPENGER_5 | 0.999995 | 0.0015 | 1.16e-73 | 3.5473 | 56.56 |
| PIPPENGER_6 | 0.999991 | 0.0019 | 2.30e-69 | 3.1035 | 105.03 |
| PIPPENGER_7 | 0.999994 | 0.0013 | 4.69e-72 | 2.7083 | 219.62 |
| PIPPENGER_8 | 0.999989 | 0.0016 | 2.64e-68 | 2.4879 | 414.58 |
| PIPPENGER_9 | 0.999977 | 0.0021 | 3.76e-64 | 2.2872 | 719.46 |
| PIPPENGER_10 | 0.999953 | 0.0028 | 5.02e-60 | 2.1705 | 1287.92 |
| PIPPENGER_11 | 0.999847 | 0.0049 | 4.73e-53 | 2.0592 | 2321.53 |
| PIPPENGER_12 | 0.999583 | 0.0075 | 3.62e-47 | 1.9331 | 4153.43 |

</details>

The linear fit ($R^2$) and standard error for `STRAUSS` is relatively worse. We can improve this by using smaller batch sizes when calibrating Strauss. I see negative intercepts (i.e., D values) for `PIPPENGER_1..4`, this should be positive. I think this could also be address by capping these algorithms to medium batch sizes during calibration. Otherwise, the calibrated values are excellent!

src/modules/musig/keyagg_impl.h
fjahr: Hm, I see the point that this would align with the batch validation API that has been discussed but I am not sure that this justifies changing the existing API of musig. It would probably better to have this in a separate/follow-up PR but I will wait a bit to see what other reviewers think about this.

src/modules/musig/keyagg_impl.h
fjahr: This was just a rough heuristic that I put not much thought in because of the TODO above which I am still unsure about. Your suggestions are interesting and I think we could actually just do `mem_limit = SECP256K1_STRAUSS_POINT_SIZE * n_pubkeys;` because in 99.9% of musig cases Strauss should be the optimal algorithm anyway and this makes the reasoning of the code easier as well. Let me know what you think, I have updated the code with this for now.

src/bench_ecmult.c
fjahr: Sounds good, taken.

src/tests_exhaustive.c
fjahr: Oh, true, good catch!

src/ecmult_impl.h
fjahr: That looks correct to me, taken!

src/ecmult_impl.h
fjahr: You are correct! Removing it.

src/ecmult_impl.h
fjahr: I think you are correct, I have removed it.

fjahr:
Addressed the comments on the implementation code and rebased, thanks a lot for the review @siv2r ! I need a little more time to think about the calibration again, but will comment on that asap.

fjahr:
@siv2r Thanks again for the review! The hint about using more narrow ranges for each of the algos in the calibration was awesome, I didn't think of this before and it pretty much fixed all the problems I ran into with the calibration results. I took these suggestions with some smaller additional tweaks, specifically with Strauss which still had quite a bit of variance. I see pretty stable calibration results between multiple runs now and the measurements between the different benchmarks are running smoother than any of my previous iterations.

These improvements have given me a lot more confidence in these changes, so taking this out of draft status too.

src/ecmult_impl.h
fjahr: Yeah, I saw some wild jumps in the `D` value and just set it to zero manually because it didn't seem to matter much. I am pretty sure I wrote a comment at the time but then probably it got lost when I updated the values again at some point later.

fjahr:
Randomly found #638 when looking at older PRs and found a small bug in my pippenger size calculation from comparing the overlapping code.

src/ecmult.h
```diff
@@ -43,19 +42,84 @@
...
- *          0 if there is not enough scratch space for a single point or
- *          callback returns 0
+ * Calculate max batch size for a given memory limit.
+ *
+ * For each algorithm, memory usage is modeled as m(x) = A*x + B and
+ * running time as c(x) = C*x + D, where x is the batch size. This
+ * function finds the algorithm that minimizes time per operation
+ * C + D/x at the maximum batch size x = (mem_limit - B) / A.
+ *
+ * Returns: The optimal batch size, or 0 if memory is insufficient.
```
siv2r: This function doesn't return 0 when memory is insufficient. It returns `ECMULT_MAX_POINTS_PER_BATCH`. We need to update the docs here. Something like:
```
The optimal batch size. Falls back to ECMULT_MAX_POINTS_PER_BATCH 
using the trivial algorithm when no Strauss/Pippenger algorithm fits.
```

fjahr: Thanks, I fixed it and took your suggestion.

fjahr:
Addressed @siv2r 's feedback and rebased, I had kind of lost track of that, sorry.

src/ecmult_impl.h
```diff
@@ -658,212 +546,365 @@ SECP256K1_INLINE static void secp256k1_ecmult_endo_split(secp256k1_scalar *s1, s
...
+    const secp256k1_callback *error_callback,
+    secp256k1_gej *r,
+    size_t n_points,
+    const secp256k1_ge *points,
+    const secp256k1_scalar *scalars,
+    const secp256k1_scalar *scalar_g,
+    size_t mem_limit
+) {
+    secp256k1_ecmult_multi_algo algo = secp256k1_ecmult_multi_select(mem_limit, n_points);
+    return secp256k1_ecmult_multi_internal(error_callback, algo, r, n_points, points, scalars, scalar_g);
```
siv2r: The old `secp256k1_ecmult_multi_var` had an internal batch loop: it split `n` points into `ceil(n / max_points_per_batch)` chunks based on scratch capacity, ran Strauss or Pippenger on each chunk, and accumulated results. The new `secp256k1_ecmult_multi` passes `n_points` directly as `batch_size` to `secp256k1_ecmult_multi_select` with no splitting. All points are processed in one shot.

This means when `n_points * A + B > mem_limit` for every algorithm, `select` falls back to TRIVIAL. On a 64-bit system, Strauss needs ~1,414 bytes/point and even the cheapest Pippenger (window 1) needs ~760 bytes/point. So for `test_ecmult_multi_batching` with `n_points = 200`:

| mem_limit | Strauss (needs 283 KB for 200 pts) | Pippenger w1 (needs 153 KB for 200 pts) | Selected |
|-----------|------------------------------------|-----------------------------------------|----------|
| 1 KB      | exceeds          | exceeds               | TRIVIAL  |
| 4 KB      | exceeds          | exceeds               | TRIVIAL  |
| 16 KB     | exceeds          | exceeds               | TRIVIAL  |
| 64 KB     | exceeds          | exceeds               | TRIVIAL  |
| 256 KB    | exceeds          | fits                  | Pippenger|
| 1 MB+     | fits             | fits                  | best fit |

The old test (176 points, scratch sized from 1 to 176) used TRIVIAL only 2 out of 178 calls (the two explicit edge cases), and every loop iteration used Strauss or Pippenger via batching. The new test hits TRIVIAL for 4 of 7 memory limits, it still passes (TRIVIAL is correct), but no longer exercises the efficient algorithms under constrained memory.

One fix: re-introduce the batch loop in `secp256k1_ecmult_multi` using the existing `secp256k1_ecmult_multi_batch_size` to compute chunk size, then call `secp256k1_ecmult_multi_internal` per chunk. This also means modules that want to manage their own batching (e.g., batch verification) could call `_internal` directly.

Alternatively, if callers are always expected to provide sufficient `mem_limit`, the current design works. Since `ecmult_multi` isn't a public API, that's not unreasonable. But most callers probably won't know what "sufficient memory" means for a given `n_points`, they'd expect `ecmult_multi` to handle that for them (see also [this discussion](https://github.com/bitcoin-core/secp256k1/pull/1789#discussion_r2648251146) about exposing `mem_limit` to musig callers). I'm slightly leaning toward letting `ecmult_multi` do the batching like the old code, but will probably know more once caller requirements become clearer.

src/ecmult_impl.h
```diff
@@ -658,212 +546,365 @@ SECP256K1_INLINE static void secp256k1_ecmult_endo_split(secp256k1_scalar *s1, s
...
-    buckets = (secp256k1_gej *) secp256k1_scratch_alloc(error_callback, scratch, ((size_t)1 << bucket_window) * sizeof(*buckets));
-    if (state_space->ps == NULL || state_space->wnaf_na == NULL || buckets == NULL) {
-        secp256k1_scratch_apply_checkpoint(error_callback, scratch, scratch_checkpoint);
-        return 0;
-    }
+/* Strauss per-point memory */
+#define SECP256K1_STRAUSS_POINT_SIZE \
+    ((sizeof(secp256k1_ge) + sizeof(secp256k1_fe)) * ECMULT_TABLE_SIZE(WINDOW_A) \
+     + sizeof(struct secp256k1_strauss_point_state) \
+     + sizeof(secp256k1_gej) + sizeof(secp256k1_scalar))
```
siv2r:
```suggestion
#define SECP256K1_STRAUSS_POINT_SIZE \
    ((sizeof(secp256k1_ge) + sizeof(secp256k1_fe)) * ECMULT_TABLE_SIZE(WINDOW_A) \
     + sizeof(struct secp256k1_strauss_point_state) \
     + sizeof(secp256k1_gej))
```

We should remove `sizeof(secp256k1_scalar)`, as `ecmult_multi_strauss` now passes the `scalars` pointer directly to `ecmult_multi_strauss_wnaf` without allocating it on the heap (the old code used to do this)

src/bench_ecmult.c
```diff
@@ -62,6 +70,142 @@ typedef struct {
...
+        /* Only run up to the max to not skew result */
+        if (n > STRAUSS_MAX_CALIB_BATCH) continue;
+
+        /* Using many iterations in Strauss since batch sizes are small */
+        if (n >= 300) iters = base_iters / 2;
+        if (iters < 100) iters = 100;
+
+        t_start = gettime_i64();
+        for (iter = 0; iter < iters; iter++) {
+            secp256k1_ecmult_multi_internal(&data->ctx->error_callback, algo,
```
siv2r: Should we wrap this in `CHECK()`?

src/bench_ecmult.c
```diff
@@ -62,6 +70,142 @@ typedef struct {
...
+
+            /* Limiting iterations to keep run-time managable */
+            if (n >= 1000) iters = base_iters / 10;
+            if (n >= 5000) iters = base_iters / 50;
+            if (n >= 15000) iters = base_iters / 100;
+            if (iters < 3) iters = 3;
+
+            t_start = gettime_i64();
+            for (iter = 0; iter < iters; iter++) {
+                secp256k1_ecmult_multi_internal(&data->ctx->error_callback, algo,
```
siv2r: Should we wrap this in `CHECK()`?

src/ecmult_impl.h
siv2r: One downside of having `ecmult_multi` do internal batching (like the old code) is that callers lose visibility into which algorithm is actually being used. In particular, it's hard to detect a silent fallback to TRIVIAL. The current design avoids this since callers can call `_select` beforehand to inspect the choice.

That said, I think this isn't a problem. If `ecmult_multi` handled batching internally, callers who just want the result would use it directly. Callers who want more control (inspect the algorithm, force a specific one, handle batching themselves) would use `_select` + `_multi_internal` instead.

Right now, `ecmult_multi` is basically a thin wrapper over `_select` + `_multi_internal`, so adding batching to it would give the two functions more distinct roles. What do you think?

src/ecmult.h
```diff
@@ -46,19 +45,85 @@
...
+/**
+ * Select the best algorithm for a given batch size within the memory
+ * limit.
+ *
+ * Among algorithms that fit within mem_limit for the given batch_size,
+ * selects the one that minimizes time per operation C + D/batch_size.
+ *
+ * Returns: The optimal algorithm identifier.
+ */
+static secp256k1_ecmult_multi_algo secp256k1_ecmult_multi_select(
```
siv2r, line 102: nit: I think it'd be nice to document that `_batch_size` and `_select` may disagree on their choice of optimial algorithm.

src/ecmult_impl.h
```diff
@@ -658,212 +546,365 @@ SECP256K1_INLINE static void secp256k1_ecmult_endo_split(secp256k1_scalar *s1, s
...
+            return secp256k1_ecmult_multi_pippenger(error_callback, r, n_points,
+                                                    points, scalars, scalar_g, 10);
+        case SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_11:
+            return secp256k1_ecmult_multi_pippenger(error_callback, r, n_points,
+                                                    points, scalars, scalar_g, 11);
+        case SECP256K1_ECMULT_MULTI_ALGO_PIPPENGER_12:
+            return secp256k1_ecmult_multi_pippenger(error_callback, r, n_points,
+                                                    points, scalars, scalar_g, 12);
+        default:
+            return secp256k1_ecmult_multi_trivial(r, n_points, points, scalars, scalar_g);
```
siv2r, line 938:
```suggestion
        default:
            VERIFY_CHECK(0); /* unreachable with valid algo values */
            return secp256k1_ecmult_multi_trivial(r, n_points, points, scalars, scalar_g);
```

Since the enum covers exactly 14 values (0-13) and `_select` only returns valid values, this branch should be unreachable. If it ever runs, it likely means a bug, and silently falling back to TRIVIAL would make that very hard to notice. This would catch bugs in debug builds while keeping the fallback in the default build.

src/tests.c
siv2r, line 1: The PR description acknowledges that test coverage is currently lower than before. Here's a list that might be helpful when adding tests back:

- Test each algorithm individually via `_multi_internal` (TRIVIAL, STRAUSS, PIPPENGER 1-12), verifying results against `secp256k1_ecmult` for small inputs (n=0, 1, 2)
- Structured edge case tests: all-infinity points, all-zero scalars, cancelling pairs (negated scalar with same point, same scalar with negated point), sum-to-zero across multiple points
- Boundary tests for `_batch_size`: `mem_limit=0` should return `ECMULT_MAX_POINTS_PER_BATCH` (TRIVIAL fallback), `mem_limit=1` similarly. For larger values, verify the returned batch_size leads to a successful `_multi` call
- Property tests for `_select`: with large `mem_limit`, `_select(mem_limit, 10)` should pick STRAUSS (fastest for tiny batches), `_select(mem_limit, 100000)` should pick a PIPPENGER variant (fastest for large batches), `_select(0, n)` should return TRIVIAL for any `n`. 
    - I'm wondering if we can also define a range of points where each algorithm transitions to the next, and assert those transitions

siv2r:
I re-ran my old analysis on the new ABCD values. The results are very positive, and there's clear improvement in `STRAUSS` values, the linear fit is better now (0.983 -> 0.998).

<details>
<summary>ABCD Values</summary>

```c
static const struct secp256k1_ecmult_multi_abcd secp256k1_ecmult_multi_abcds[SECP256K1_ECMULT_MULTI_NUM_ALGOS] = {
    {0,                                     0,                                     1000,  0     },
    {SECP256K1_STRAUSS_POINT_SIZE,          0,                                     103,   274  },
    {SECP256K1_PIPPENGER_POINT_SIZE(1),  SECP256K1_PIPPENGER_FIXED_SIZE(1),  195,   417  },
    {SECP256K1_PIPPENGER_POINT_SIZE(2),  SECP256K1_PIPPENGER_FIXED_SIZE(2),  147,   602  },
    {SECP256K1_PIPPENGER_POINT_SIZE(3),  SECP256K1_PIPPENGER_FIXED_SIZE(3),  117,   888  },
    {SECP256K1_PIPPENGER_POINT_SIZE(4),  SECP256K1_PIPPENGER_FIXED_SIZE(4),  100,   1390 },
    {SECP256K1_PIPPENGER_POINT_SIZE(5),  SECP256K1_PIPPENGER_FIXED_SIZE(5),  86,    2227 },
    {SECP256K1_PIPPENGER_POINT_SIZE(6),  SECP256K1_PIPPENGER_FIXED_SIZE(6),  76,    3755 },
    {SECP256K1_PIPPENGER_POINT_SIZE(7),  SECP256K1_PIPPENGER_FIXED_SIZE(7),  67,    6626 },
    {SECP256K1_PIPPENGER_POINT_SIZE(8),  SECP256K1_PIPPENGER_FIXED_SIZE(8),  62,    10976},
    {SECP256K1_PIPPENGER_POINT_SIZE(9),  SECP256K1_PIPPENGER_FIXED_SIZE(9),  57,    19909},
    {SECP256K1_PIPPENGER_POINT_SIZE(10), SECP256K1_PIPPENGER_FIXED_SIZE(10), 54,    36574},
    {SECP256K1_PIPPENGER_POINT_SIZE(11), SECP256K1_PIPPENGER_FIXED_SIZE(11), 50,    72630},
    {SECP256K1_PIPPENGER_POINT_SIZE(12), SECP256K1_PIPPENGER_FIXED_SIZE(12), 46,    135404},
};
```
</details>

<details>
<summary>Statistical analysis</summary>

## Metric Definitions

| Metric | Meaning | Good Value for Benchmarking |
|--------|---------|-----------------------------|
| R² | How well the linear model fits (0-1) | > 0.95 (strong linear relationship) |
| Std Error | Uncertainty in slope estimate | Low relative to slope |
| p-value | Probability slope = 0 by chance | < 0.05 (statistically significant) |
| Slope | Time increase per additional n (μs) | Positive, algorithm-dependent |
| Intercept | Fixed overhead time (μs) | Algorithm-dependent |

## Per-Algorithm Metrics

| Algorithm | R² | Std Error | p-value | Slope | Intercept |
|-----------|------|-----------|---------|-------|----------|
| STRAUSS | 0.998256 | 0.0468 | 2.77e-26 | 4.5477 | -15.70 |
| PIPPENGER_1 | 0.999981 | 0.0110 | 6.54e-25 | 7.6313 | 16.23 |
| PIPPENGER_2 | 0.999932 | 0.0131 | 1.38e-30 | 5.7699 | 21.74 |
| PIPPENGER_3 | 0.999973 | 0.0066 | 1.98e-33 | 4.6724 | 27.37 |
| PIPPENGER_4 | 0.999987 | 0.0040 | 3.30e-33 | 3.9572 | 45.39 |
| PIPPENGER_5 | 0.999988 | 0.0033 | 9.93e-36 | 3.4059 | 80.97 |
| PIPPENGER_6 | 0.999979 | 0.0044 | 1.02e-24 | 2.9971 | 123.83 |
| PIPPENGER_7 | 0.999960 | 0.0057 | 4.08e-21 | 2.6388 | 220.59 |
| PIPPENGER_8 | 0.999989 | 0.0027 | 1.18e-23 | 2.4487 | 353.68 |
| PIPPENGER_9 | 0.999995 | 0.0019 | 8.61e-20 | 2.2403 | 700.71 |
| PIPPENGER_10 | 0.999992 | 0.0028 | 7.05e-14 | 2.1095 | 1444.03 |
| PIPPENGER_11 | 0.999914 | 0.0109 | 3.41e-07 | 1.9767 | 2804.01 |
| PIPPENGER_12 | 0.999997 | 0.0033 | 1.14e-03 | 1.8235 | 5310.41 |

</details>

src/bench_ecmult.c
siv2r: STRAUSS

The Strauss fit improved. More info: https://github.com/bitcoin-core/secp256k1/pull/1789#issuecomment-4137354884

src/ecmult_impl.h
```diff
@@ -658,212 +546,365 @@ SECP256K1_INLINE static void secp256k1_ecmult_endo_split(secp256k1_scalar *s1, s
...
+ * ABCD constants for all batch algorithms.
+ *
+ * Index 0    = TRIVIAL (no memory, very slow)
+ * Index 1    = STRAUSS (efficient for small batche sizes)
+ * Index 2-13 = PIPPENGER with window sizes 1-12
+ */
+ static const struct secp256k1_ecmult_multi_abcd secp256k1_ecmult_multi_abcds[SECP256K1_ECMULT_MULTI_NUM_ALGOS] = {
+/*  A (per-point bytes)                  B (fixed bytes)                     C      D     */
+    {0,                                  0,                                  1000,  0     },
+    {SECP256K1_STRAUSS_POINT_SIZE,       0,                                  109,   120   },
```
siv2r, line 598: I ran benchmark a few times, and my Strauss's D value is twice this. It's in 250ish region. The CD values of the remaining algorithms match pretty well.

siv2r:
I was curious whether we could squeeze out better CD values by calibrating each algorithm on its most optimal batch range, instead of the broader ranges we currently use. So I ran all 14 `ecmult_multi` algorithms across [2, 50000] and plotted the per-point times to figure out which algorithm actually wins at each batch size (see the two toggles below for the full picture).

<details>
<summary>graph comparing all 14 algorithms for speedup</summary>
speedup_vs_individual

</details>

<details>
<summary>optimal algorithm for each batch size</summary>
winners_speedup

### Optimal Algorithm Ranges

| Optimal Range | Best Algorithm | Per-Point Time (us) | Speedup vs Individual |
|---------------|----------------|---------------------|-----------------------|
| [2, 90]       | STRAUSS        | 4.40 - 6.97         | 1.40x - 2.27x         |
| [95, 130]     | PIPPENGER_5    | 4.12 - 4.37         | 2.30x - 2.44x         |
| [150, 250]    | PIPPENGER_6    | 3.67 - 4.01         | 2.51x - 2.76x         |
| [300, 750]    | PIPPENGER_7    | 3.02 - 3.56         | 2.84x - 3.37x         |
| [1000, 1500]  | PIPPENGER_8    | 2.74 - 2.88         | 3.59x - 3.77x         |
| [2000, 7500]  | PIPPENGER_9    | 2.34 - 2.63         | 3.91x - 4.42x         |
| [10000, 12500] | PIPPENGER_11   | 2.24 - 2.29         | 4.53x - 4.62x         |
| [15000, 50000] | PIPPENGER_12   | 1.96 - 2.20         | 4.71x - 5.29x         |

### Key Findings

- **Never optimal:** PIPPENGER_1, PIPPENGER_2, PIPPENGER_3, PIPPENGER_4, PIPPENGER_10 — dominated by neighbors at every batch size.
- **Crossover points:** STRAUSS -> PIPPENGER_5 at n~92; PIPPENGER_5 -> PIPPENGER_6 at n=140; PIPPENGER_6 -> PIPPENGER_7 at n=275; PIPPENGER_7 -> PIPPENGER_8 at n=875; PIPPENGER_8 -> PIPPENGER_9 at n=1750; PIPPENGER_9 -> PIPPENGER_11 at n=8750; PIPPENGER_11 -> PIPPENGER_12 at n=13750.

### Winner at each batch size

| N | Best Algorithm | Per-Point Time (us) | Speedup |
|---|----------------|---------------------|---------|
| 2 | STRAUSS | 6.97 | 1.40x |
| 3 | STRAUSS | 6.12 | 1.60x |
| 4 | STRAUSS | 5.75 | 1.71x |
| 5 | STRAUSS | 5.46 | 1.80x |
| 7 | STRAUSS | 5.15 | 1.93x |
| 10 | STRAUSS | 4.85 | 2.06x |
| 13 | STRAUSS | 4.79 | 2.09x |
| 16 | STRAUSS | 4.61 | 2.15x |
| 20 | STRAUSS | 4.58 | 2.16x |
| 25 | STRAUSS | 4.58 | 2.18x |
| 30 | STRAUSS | 4.46 | 2.24x |
| 35 | STRAUSS | 4.46 | 2.23x |
| 40 | STRAUSS | 4.48 | 2.23x |
| 45 | STRAUSS | 4.44 | 2.25x |
| 50 | STRAUSS | 4.48 | 2.24x |
| 55 | STRAUSS | 4.40 | 2.27x |
| 60 | STRAUSS | 4.42 | 2.26x |
| 65 | STRAUSS | 4.42 | 2.27x |
| 70 | STRAUSS | 4.42 | 2.27x |
| 75 | STRAUSS | 4.41 | 2.27x |
| 80 | STRAUSS | 4.46 | 2.26x |
| 85 | STRAUSS | 4.45 | 2.26x |
| 88 | PIPPENGER_5 | 4.45 | 2.26x |
| 90 | STRAUSS | 4.43 | 2.27x |
| 95 | PIPPENGER_5 | 4.37 | 2.30x |
| 100 | PIPPENGER_5 | 4.33 | 2.33x |
| 110 | PIPPENGER_5 | 4.25 | 2.37x |
| 120 | PIPPENGER_5 | 4.18 | 2.41x |
| 130 | PIPPENGER_5 | 4.12 | 2.44x |
| 150 | PIPPENGER_6 | 4.01 | 2.51x |
| 175 | PIPPENGER_6 | 3.86 | 2.60x |
| 200 | PIPPENGER_6 | 3.77 | 2.68x |
| 250 | PIPPENGER_6 | 3.67 | 2.76x |
| 300 | PIPPENGER_7 | 3.56 | 2.84x |
| 350 | PIPPENGER_7 | 3.46 | 2.93x |
| 400 | PIPPENGER_7 | 3.38 | 2.99x |
| 500 | PIPPENGER_7 | 3.25 | 3.11x |
| 600 | PIPPENGER_7 | 3.09 | 3.29x |
| 750 | PIPPENGER_7 | 3.02 | 3.37x |
| 1000 | PIPPENGER_8 | 2.88 | 3.59x |
| 1200 | PIPPENGER_8 | 2.81 | 3.69x |
| 1500 | PIPPENGER_8 | 2.74 | 3.77x |
| 2000 | PIPPENGER_9 | 2.63 | 3.91x |
| 2500 | PIPPENGER_9 | 2.56 | 4.05x |
| 3000 | PIPPENGER_9 | 2.51 | 4.12x |
| 4000 | PIPPENGER_9 | 2.46 | 4.20x |
| 5000 | PIPPENGER_9 | 2.42 | 4.28x |
| 7500 | PIPPENGER_10 | 2.34 | 4.42x |
| 10000 | PIPPENGER_11 | 2.29 | 4.53x |
| 12500 | PIPPENGER_11 | 2.24 | 4.62x |
| 15000 | PIPPENGER_12 | 2.20 | 4.71x |
| 20000 | PIPPENGER_12 | 2.12 | 4.90x |
| 25000 | PIPPENGER_12 | 2.07 | 5.01x |
| 30000 | PIPPENGER_12 | 2.03 | 5.09x |
| 40000 | PIPPENGER_12 | 2.00 | 5.20x |
| 50000 | PIPPENGER_12 | 1.96 | 5.29x |

</details>

A few things stood out. First, pippenger w=1 through w=4 are never optimal at any batch size — after strauss, `ecmult_multi` jumps straight to pippenger w=5. The pippenger CD values in general are already well calibrated, which makes sense since the linear fit for those is near perfect (R^2 > 0.999).

The interesting part was strauss. Strauss flatlines after about n=90. So when we calibrate using batches from [2, 500] as we currently do, we're fitting a line to data that stops being linear after 90. Including these flatline points pulls the regression intercept C upward (104 -> 109) and the slope D downward (184 -> 120) compared to what we'd get from fitting just the linear region [2, 90]. In other words, the non-linear tail inflates the C value.

Surprisingly, that inflation is actually helpful. The `ecmult_multi_select` picks the method that minimizes `C + D/n`. With the current values (C=109, D=120 for strauss, C=86, D=2187 for pippenger w=5), the crossover `n` works out to `(2187 - 120) / (109 - 86) = 89.9`. Which lines up almost perfectly with the empirical data showing pippenger w=5 becomes faster around n=92 (in the second graph above). Also, the old code switched to pippenger at n = 88.

When I tried to "fix" the linear fit by calibrating strauss only on [2, 90] where it's truly linear, I got a much better R^2 (0.9998 vs 0.983) but the crossover moved to n=110. That means `ecmult_multi` would keep picking strauss all the way through n=110, even though pippenger w=5 is clearly faster by n=95. Better statistical fit, worse algorithm selection.

TL;DR the current CD values are near perfect. The slightly "bad" linear fit for strauss isn't a problem. It is exactly what pushes the crossover to pippenger at the correct batch size (at n=89).

src/ecmult_impl.h
siv2r: I take back this comment. The current CD values are good (see https://github.com/bitcoin-core/secp256k1/pull/1789#issuecomment-4157003939). Feel free to mark this as resolved.

src/ecmult_impl.h
fjahr: You are making a good argument and I agree, I oversimplified things not keeping memory constraint environments in mind here. I have introduced the batch loop. If possible it still tries to one-shot but if not it basically matches the behavior from `secp256k1_ecmult_multi_var` in master.

src/ecmult_impl.h
fjahr: Right, fixed.

src/bench_ecmult.c
fjahr: Makes sense, done.

src/bench_ecmult.c
fjahr: Makes sense, done.

src/ecmult.h
fjahr: Ok, added a comment

src/ecmult_impl.h
fjahr: Makes sense, added!

src/tests.c
fjahr: Thanks, yeah, I had planned to get back to this and add some more tests. I think I was able to cover all your suggestions but I have skipped exploring coverage for the exact transitions since they seemed brittle and we might still recalibrate. But open to revisit this if other reviewers think this would be valuable to add once we had a few more eyes on the current calibration values.

fjahr:
Huge thanks for the helpful review comments @siv2r , I think I have addressed them all now.

w0xlt:
`tools/ecmult_multi_calib.py` can be added to `EXTRA_DIST` in `Makefile.am`. 
That makes Automake include the calibration helper in release tarballs.

```diff
diff --git a/Makefile.am b/Makefile.am
index 07d7a2b..81a2d69 100644
--- a/Makefile.am
+++ b/Makefile.am
@@ -288,6 +288,7 @@ EXTRA_DIST += sage/weierstrass_prover.sage
 EXTRA_DIST += src/wycheproof/WYCHEPROOF_COPYING
 EXTRA_DIST += src/wycheproof/ecdsa_secp256k1_sha256_bitcoin_test.json
 EXTRA_DIST += src/wycheproof/ecdh_secp256k1_test.json
+EXTRA_DIST += tools/ecmult_multi_calib.py
 EXTRA_DIST += tools/tests_wycheproof_generate_ecdsa.py
 EXTRA_DIST += tools/tests_wycheproof_generate_ecdh.py
 ```

fjahr:
> tools/ecmult_multi_calib.py can be added to EXTRA_DIST in Makefile.am.

Done, thanks! Initially I wasn't even sure if this script should be part of the repo but it looks to me like other, similar files are treated similarly as well.

apoelstra:
In this PR what is the plan for users who do not have access to `malloc`?

fjahr:
> In this PR what is the plan for users who do not have access to `malloc`?

Good question, I didn't have this top of mind when working on this and as far as I can remember there wasn't a specific handling of this case discussed at the meeting, other than the BYOM comment in the notes.

For the new `secp256k1_ecmult_multi` the answer would be to pass `mem_limit` 0, which means no malloc is used and the trivial algo selected. So the path should already exist as far as that code is concerned if you think that works well enough.

But for the musig `pubkey_agg` I was using the strauss sized  `mem_limit` as a performance optimization but this removed the current allocation free path that exists on master (with a null scratch space). So I walked this back for now and simply keep using the trivial algo for musig, then revisit a potential performance update in a follow-up. This is probably the better decision in terms of scope for this pull anyway.

apoelstra:
> For the new `secp256k1_ecmult_multi` the answer would be to pass `mem_limit` 0, which means no malloc is used

For rust-secp256k1 it doesn't matter whether it's used or not -- if we attempt to link `malloc` into the library then our wasm build won't work.

The original purpose of the scratch space API was to enable algorithms that use variable amounts of memory without needing to add a `malloc` dependency.

real-or-random:
> For rust-secp256k1 it doesn't matter whether it's used or not -- if we attempt to link `malloc` into the library then our wasm build won't work.
> 
> The original purpose of the scratch space API was to enable algorithms that use variable amounts of memory without needing to add a `malloc` dependency.

I was the one who brought up bring-your-own-malloc idea at the meeting, so let me give some background. Our thinking was that the primary application here is batch verification, and you probably won't run this on a machine without some form of `malloc` available.

I know, wasm is always a bit special, but can't you redirect `malloc` to (a wrapper of) [`std::alloc`](https://doc.rust-lang.org/std/alloc/fn.alloc.html) in Rust? As far as I understand, it uses a [dlmalloc backend](https://docs.rs/dlmalloc/latest/dlmalloc/) for `wasm32-unknown-unknown`.

We should give users an easy way to override `malloc`, e.g., to enable such redirects. For example, we could make a `SECP256K1_MALLOC` macro that we define to be `malloc` unless it's predefined by the user. (This is related to https://github.com/bitcoin-core/secp256k1/issues/1095.) 

This should certainly be sorted out. This could be a road map:
 - We add some way for the user to override `malloc`, e.g., `SECP256K1_MALLOC`.
 - We ship this PR with batch verification as the target in mind.
 - For embedded targets which need `ecmult_multi` (e.g., computing a MuSig aggregate key on a hardware wallet), we'd need to come up with something else. For example, if we stick to the rule that `ecmult_multi` code `free`s all allocations in a row, then a simple bump allocator would do. We could then repurpose the implementation in the scratch space implementation and make it available to the user as another thing `SECP256K1_MALLOC` could be set to.

fjahr:
> For rust-secp256k1 it doesn't matter whether it's used or not -- if we attempt to link malloc into the library then our wasm build won't work.

I see, thanks, I wasn't aware of the exact limitations until now. I will spend some time looking into it more in depth.

> We should give users an easy way to override malloc

The road map makes sense to me. I gave this a shot with https://github.com/bitcoin-core/secp256k1/pull/1917, not sure but maybe this could help revive https://github.com/bitcoin-core/secp256k1/pull/1461 to go along with it?

apoelstra:
I can override `malloc` with a Rust equivalent but I cannot override `free`. The Rust deallocation function needs to know how many bytes are being deallocated, which `free` does not take.

I think we have some existing hackery to handle this situation for context objects, where we hide a length prefix, so maybe it's okay.

But if you can provide a way to override `free` where you provide the number of bytes that are being deallocated, that would make my life much easier. (In fact, it may allow me to remove some existing hacks related to the context object!)

real-or-random:
> But if you can provide a way to override `free` where you provide the number of bytes that are being deallocated

I think a `SECP256K1_FREE` macro could totally also except the number of bytes. The default implementation that uses `free` under the hood could just ignore it.

fjahr:
> I think a `SECP256K1_FREE` macro could totally also except the number of bytes. The default implementation that uses `free` under the hood could just ignore it.

Implemented as part of #1917.

fjahr:
Rebased

PR #1791, opened 2025-12-20
cmake: Add option to generate precomputed files

hebasto:
Using the CMake-based build system enables smoother integration with parent projects by providing both CMake package configuration files and the traditional `libsecp256k1.pc`.

However, packagers on some systems, such as [Gentoo](https://github.com/bitcoin-core/secp256k1/blob/f9a944ff2dbcd7ff1829880636f861caf0a91392/Makefile.am#L83) and [Guix](https://codeberg.org/guix/guix/pulls/2566#issuecomment-7360201), insist on recreating the precomputed files, which is not currently supported by the CMake-based build system.

This PR addresses the issue by introducing a new `SECP256K1_BUILD_PRECOMPUTED` option (`OFF` by default) that generates `precomputed_ecmult.c` and `precomputed_ecmult_gen.c` in the build tree. Those files are then compiled instead of the ones checked into the source tree.

Here is an example:
```
$ cmake -B build -G Ninja -DSECP256K1_BUILD_PRECOMPUTED=ON
$ ninja -C build src/precomputed_ecmult.c src/precomputed_ecmult_gen.c | cat
ninja: Entering directory `build'
[1/6] Building C object src/CMakeFiles/secp256k1_precompute_ecmult.dir/precompute_ecmult.c.o
[2/6] Building C object src/CMakeFiles/secp256k1_precompute_ecmult_gen.dir/precompute_ecmult_gen.c.o
[3/6] Linking C executable secp256k1_precompute_ecmult_gen
[4/6] Linking C executable secp256k1_precompute_ecmult
[5/6] Generating precomputed_ecmult_gen.c
[6/6] Generating precomputed_ecmult.c
```

When configuring with CMake 3.31 or newer, the built-in `codegen` target can be used to generate those files:
```
$ cmake -B build -G Ninja -DSECP256K1_BUILD_PRECOMPUTED=ON
$ cmake --build build -t codegen | cat
[1/6] Building C object src/CMakeFiles/secp256k1_precompute_ecmult.dir/precompute_ecmult.c.o
[2/6] Building C object src/CMakeFiles/secp256k1_precompute_ecmult_gen.dir/precompute_ecmult_gen.c.o
[3/6] Linking C executable secp256k1_precompute_ecmult_gen
[4/6] Linking C executable secp256k1_precompute_ecmult
[5/6] Generating precomputed_ecmult_gen.c
[6/6] Generating precomputed_ecmult.c
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -538,6 +538,10 @@ jobs:
     name: ${{ matrix.configuration.job_name }}
     # See: https://github.com/actions/runner-images#available-images.
     runs-on: windows-2022
+    defaults:
+      run:
+        # Enforce fail-fast behavior for PowerShell.
+        shell: pwsh -Command "$PSNativeCommandUseErrorActionPreference = $true; $ErrorActionPreference = 'Stop'; & '{0}'"
 
     strategy:
       fail-fast: false
@@ -570,7 +574,13 @@ jobs:
       - *CHECKOUT
 
       - name: Generate buildsystem
-        run: cmake -E env CFLAGS="/WX ${{ matrix.configuration.cpp_flags }}" cmake -B build -DSECP256K1_ENABLE_MODULE_RECOVERY=ON -DSECP256K1_BUILD_EXAMPLES=ON ${{ matrix.configuration.cmake_options }}
+        run: |
+          cmake -E env CFLAGS="/WX ${{ matrix.configuration.cpp_flags }}" `
+          cmake -B build `
+            -DSECP256K1_BUILD_PRECOMPUTED=ON `
+            -DSECP256K1_ENABLE_MODULE_RECOVERY=ON `
+            -DSECP256K1_BUILD_EXAMPLES=ON `
+            ${{ matrix.configuration.cmake_options }}
 
       - name: Build
         run: cmake --build build --config RelWithDebInfo -- /p:UseMultiToolTask=true /maxCpuCount
@@ -595,6 +605,11 @@ jobs:
           build\bin\RelWithDebInfo\bench_ecmult.exe
           build\bin\RelWithDebInfo\bench_internal.exe
           build\bin\RelWithDebInfo\bench.exe
+
+      - name: Check precomputed files
+        run: |
+          cmake -E copy build/src/precomputed_ecmult.c build/src/precomputed_ecmult_gen.c -t src
+          git diff --exit-code
 
   win64-native-headers:
     name: "x64 (MSVC): C++ (public headers)"
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,11 @@
 cmake_minimum_required(VERSION 3.22)
 
+if(POLICY CMP0171)
+  # `codegen` is a reserved target name.
+  # See: https://cmake.org/cmake/help/latest/policy/CMP0171.html
+  cmake_policy(SET CMP0171 NEW)
+endif()
+
 #=============================
 # Project / Package metadata
 #=============================
@@ -140,6 +146,11 @@ option(SECP256K1_BUILD_TESTS "Build tests." ON)
 option(SECP256K1_BUILD_EXHAUSTIVE_TESTS "Build exhaustive tests." ON)
 option(SECP256K1_BUILD_CTIME_TESTS "Build constant-time tests." ${SECP256K1_VALGRIND})
 option(SECP256K1_BUILD_EXAMPLES "Build examples." OFF)
+
+option(SECP256K1_BUILD_PRECOMPUTED "Generate the precomputed files in the build tree." OFF)
+if(SECP256K1_BUILD_PRECOMPUTED AND CMAKE_CROSSCOMPILING)
+  message(FATAL_ERROR "SECP256K1_BUILD_PRECOMPUTED requires executing precompute tools, which is not supported when cross compiling.")
+endif()
 
 # Redefine configuration flags.
 # We leave assertions on, because they are only used in the examples, and we want them always on there.
--- a/cmake/Precompute.cmake
+++ b/cmake/Precompute.cmake
@@ -1,0 +1,21 @@
+function(precompute base_name)
+  add_executable(secp256k1_precompute_${base_name} EXCLUDE_FROM_ALL precompute_${base_name}.c)
+  set_target_properties(secp256k1_precompute_${base_name} PROPERTIES
+    COMPILE_DEFINITIONS VERIFY
+    RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}
+  )
+  set(CODEGEN_OPT "")
+  if(POLICY CMP0171)
+    cmake_policy(GET CMP0171 _cmp0171_status)
+    if(_cmp0171_status STREQUAL "NEW")
+      set(CODEGEN_OPT CODEGEN)
+    endif()
+  endif()
+  add_custom_command(
+    OUTPUT ${PROJECT_BINARY_DIR}/src/precomputed_${base_name}.c
+    COMMAND secp256k1_precompute_${base_name}
+    DEPENDS secp256k1_precompute_${base_name}
+    WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
+    ${CODEGEN_OPT}
+  )
+endfunction()
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -45,10 +45,22 @@ if(SECP256K1_ENABLE_MODULE_ECDH)
   set_property(TARGET secp256k1 APPEND PROPERTY PUBLIC_HEADER ${PROJECT_SOURCE_DIR}/include/secp256k1_ecdh.h)
 endif()
 
-add_library(secp256k1_precomputed OBJECT EXCLUDE_FROM_ALL
-  precomputed_ecmult.c
-  precomputed_ecmult_gen.c
-)
+add_library(secp256k1_precomputed OBJECT)
+if(SECP256K1_BUILD_PRECOMPUTED)
+  include(Precompute)
+  precompute(ecmult)
+  precompute(ecmult_gen)
+  target_sources(secp256k1_precomputed PRIVATE
+    ${PROJECT_BINARY_DIR}/src/precomputed_ecmult.c
+    ${PROJECT_BINARY_DIR}/src/precomputed_ecmult_gen.c
+  )
+  target_include_directories(secp256k1_precomputed PRIVATE ${CMAKE_CURRENT_LIST_DIR})
+else()
+  target_sources(secp256k1_precomputed PRIVATE
+    ${CMAKE_CURRENT_LIST_DIR}/precomputed_ecmult.c
+    ${CMAKE_CURRENT_LIST_DIR}/precomputed_ecmult_gen.c
+  )
+endif()
 
 # Add objects explicitly rather than linking to the object libs to keep them
 # from being exported.
--- a/src/precompute_ecmult.c
+++ b/src/precompute_ecmult.c
@@ -75,7 +75,7 @@ int main(void) {
     fprintf(fp, "#include \"precomputed_ecmult.h\"\n");
     fprintf(fp, "#define S(a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p) SECP256K1_GE_STORAGE_CONST(0x##a##u,0x##b##u,0x##c##u,0x##d##u,0x##e##u,0x##f##u,0x##g##u,0x##h##u,0x##i##u,0x##j##u,0x##k##u,0x##l##u,0x##m##u,0x##n##u,0x##o##u,0x##p##u)\n");
     fprintf(fp, "#if ECMULT_WINDOW_SIZE > %d\n", window_g);
-    fprintf(fp, "   #error configuration mismatch, invalid ECMULT_WINDOW_SIZE. Try deleting precomputed_ecmult.c before the build.\n");
+    fprintf(fp, "   #error configuration mismatch, invalid ECMULT_WINDOW_SIZE. If this is an Autotools build, try deleting precomputed_ecmult.c before the build. If this is a CMake build, pass -DSECP256K1_BUILD_PRECOMPUTED=ON to cmake.\n");
     fprintf(fp, "#endif\n");
     fprintf(fp, "#ifdef EXHAUSTIVE_TEST_ORDER\n");
     fprintf(fp, "#    error Cannot compile precomputed_ecmult.c in exhaustive test mode\n");
--- a/src/precompute_ecmult_gen.c
+++ b/src/precompute_ecmult_gen.c
@@ -90,7 +90,7 @@ int main(int argc, char **argv) {
         print_table(fp, COMB_BLOCKS, COMB_TEETH);
     }
     fprintf(fp, "#else\n");
-    fprintf(fp, "#    error Configuration mismatch, invalid COMB_* parameters. Try deleting precomputed_ecmult_gen.c before the build.\n");
+    fprintf(fp, "#    error Configuration mismatch, invalid COMB_* parameters. If this is an Autotools build, try deleting precomputed_ecmult_gen.c before the build. If this is a CMake build, pass -DSECP256K1_BUILD_PRECOMPUTED=ON to cmake.\n");
     fprintf(fp, "#endif\n");
 
     fprintf(fp, "};\n");
--- a/src/precomputed_ecmult.c
+++ b/src/precomputed_ecmult.c
@@ -7,7 +7,7 @@
 #include "precomputed_ecmult.h"
 #define S(a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p) SECP256K1_GE_STORAGE_CONST(0x##a##u,0x##b##u,0x##c##u,0x##d##u,0x##e##u,0x##f##u,0x##g##u,0x##h##u,0x##i##u,0x##j##u,0x##k##u,0x##l##u,0x##m##u,0x##n##u,0x##o##u,0x##p##u)
 #if ECMULT_WINDOW_SIZE > 15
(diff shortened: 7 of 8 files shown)


kevkevinpal:
are we able to do this via `./configure`? 

or is this meant to be only done for CMake?

hebasto:
> are we able to do this via `./configure`?

Yes. Run `make clean-precomp && make precomp`.

src/CMakeLists.txt
```diff
@@ -45,10 +45,41 @@ if(SECP256K1_ENABLE_MODULE_ECDH)
...
+
+  add_custom_command(
+    OUTPUT
+      ${CMAKE_CURRENT_BINARY_DIR}/precomputed_ecmult.c
+      ${CMAKE_CURRENT_BINARY_DIR}/precomputed_ecmult_gen.c
+    COMMAND secp256k1_precompute_ecmult
+    COMMAND secp256k1_precompute_ecmult_gen
+    DEPENDS secp256k1_precompute_ecmult secp256k1_precompute_ecmult_gen
+    WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
+  )
```
real-or-random: Shouldn't these conceptually be two separate commands?

src/CMakeLists.txt
```diff
@@ -45,10 +45,41 @@ if(SECP256K1_ENABLE_MODULE_ECDH)
...
-  precomputed_ecmult_gen.c
-)
+add_library(secp256k1_precomputed OBJECT EXCLUDE_FROM_ALL)
+if(SECP256K1_BUILD_PRECOMPUTED)
+  add_executable(secp256k1_precompute_ecmult EXCLUDE_FROM_ALL precompute_ecmult.c)
+  target_compile_definitions(secp256k1_precompute_ecmult PRIVATE VERIFY)
+  if(CMAKE_C_COMPILER_ID STREQUAL "MSVC")
+    # Disable warning C4334 "'operator': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)".
+    target_compile_options(secp256k1_precompute_ecmult PRIVATE /wd4334)
+  endif()
```
real-or-random: In which line does this warning appear? Could we fix it in the code instead?

real-or-random (commented):
Concept ACK

src/CMakeLists.txt
hebasto: > In which line does this warning appear?

https://github.com/bitcoin-core/secp256k1/blob/2d9137ce9d7e576bfada2859292ecaa0887a6b8c/src/ecmult.h#L41

> Could we fix it in the code instead?

Certainly. Fixed in 19529e70680a5edaa0210b74420693d8e689c7ac.

src/CMakeLists.txt
hebasto: Thanks! Reworked.

hebasto:
@real-or-random 

src/CMakeLists.txt
real-or-random: I think conceptually the result of `ECMULT_TABLE_SIZE` should be a `size_t` because it is an array size. What do you think of this refactor instead? https://github.com/bitcoin-core/secp256k1/pull/1794

hebasto: Rebased on the top of the first commit from https://github.com/bitcoin-core/secp256k1/pull/1794.

hebasto:
> Rebased on the top of the first commit from #1794.

Rebased on top of the merged bitcoin-core/secp256k1#1794.

theStack:
Concept ACK

Tested quickly that 

```
$ cmake -B build -DSECP256K1_ECMULT_WINDOW_SIZE=16 -DSECP256K1_BUILD_PRECOMPUTED=ON && cmake --build build
```
succeeds on the PR branch as expected, while on master
```
$ cmake -B build -DSECP256K1_ECMULT_WINDOW_SIZE=16 && cmake --build build
```
fails (`error: #error configuration mismatch, invalid ECMULT_WINDOW_SIZE`).

The generated files `precomputed_ecmult{,gen}.c` files are created within the build directory. Given the out-of-tree build nature of CMake, I guess this is intended and there are no plans to support directly modifying the source tree? (Which is no big deal, developers can simply do a `$ cp ./build/src/precomputed_ecmult*.c ./src/` after manually or just re-run the precompute binaries).

hebasto:
> Given the out-of-tree build nature of CMake, I guess this is intended and there are no plans to support directly modifying the source tree?

That's correct. We have to assume the source tree might be strictly read-only.

theStack (approved):
Light ACK 8b58c55e3d4b5e18a6bb626c5b55548b832965eb

I'm not too familiar with the ins and outs of CMake, but the changes look good and I tested the new build setting both with higher ecmult window sizes (as [shown above](https://github.com/bitcoin-core/secp256k1/pull/1791#issuecomment-4632407609)) and with different ecmult_gen `CONFIGS` table values (taken from https://github.com/theStack/secp256k1/commit/3421153be6ae8d8d88406d518c9e66ef4d49fb0a), and everything seems to work as expected.

// EDIT: I guess this is worth a release note

CMakeLists.txt
```diff
@@ -141,6 +141,8 @@ option(SECP256K1_BUILD_EXHAUSTIVE_TESTS "Build exhaustive tests." ON)
 option(SECP256K1_BUILD_CTIME_TESTS "Build constant-time tests." ${SECP256K1_VALGRIND})
 option(SECP256K1_BUILD_EXAMPLES "Build examples." OFF)
 
+option(SECP256K1_BUILD_PRECOMPUTED "Recreate the precomputed files." OFF)
```
real-or-random: nit: Do you think it will add clarity to expand this to "Recreate the precomputed files in the build tree"?

It's kind of implied that a CMake build creates stuff only in the build tree. But @theStack felt the need to point this out, so this may be surprising to users who don't really think about carefully (which includes almost everyone when it comes to build systems, I guess).

hebasto: Thanks! [Amended](https://github.com/bitcoin-core/secp256k1/pull/1791#issuecomment-4659102568).

hebasto:
The most recent feedback has been addressed.

theStack (approved):
re-ACK 8460c5902cf83d3b83e22021c7d73d6991347464

Verified that the only change since my last review was the more detailed build option description (+"in the build tree"), which I agree makes sense.

real-or-random:
> fails (`error: #error configuration mismatch, invalid ECMULT_WINDOW_SIZE`).

The truth is that it also outputs `Try deleting precomputed_ecmult.c before the build.` And this is not true for the approach in this PR. With Automake, we construct real dependencies, i.e., the main artifact depends on the pregenerated files. Here, this is just a build option.

I think either approach is fine. I guess even mixing them is fine. Maybe we should just change the error message for now? "If this is an autotools build, try deleting precomputed_ecmult.c before the build. If this is a CMake build, pass -DSECP256K1_BUILD_PRECOMPUTED=ON to cmake."

---

Further thoughts:

I'm wondering if I like the "build option" approach as in this PR more. It's easier for those users who need it, and the code is simpler. And even for contributors working on changes to the generation binaries, they can just set -DSECP256K1_BUILD_PRECOMPUTED=ON in their cache.

If others agree, this raises the question of whether we want to get rid of that stuff in autotools. It never worked 100%. And it seems that all users who want this feature use CMake. (@kevkevinpal Why did you ask about autotools? Do you need this feature in autotools?)

CMakePresets.json
```diff
@@ -7,7 +7,8 @@
       "cacheVariables": {
         "SECP256K1_EXPERIMENTAL": "ON",
         "SECP256K1_ENABLE_MODULE_RECOVERY": "ON",
-        "SECP256K1_BUILD_EXAMPLES": "ON"
+        "SECP256K1_BUILD_EXAMPLES": "ON",
+        "SECP256K1_BUILD_PRECOMPUTED": "ON"
```
real-or-random: I'm not entirely sure if we want this in dev mode. In autotools, we don't have it in dev mode. 

It's something a dev needs when working on the source for the precomputation binaries. But on the other hand, this is rare and this thing increases compile times. Maybe leave it off for now?

real-or-random (commented):
Consider rebasing this. The branch is way behind (for our standards ;)).

hebasto:
> Consider rebasing this. The branch is way behind (for our standards ;)).

Rebased. The recent feedback has been addressed.

> Maybe we should just change the error message for now? "If this is an autotools build, try deleting precomputed_ecmult.c before the build. If this is a CMake build, pass -DSECP256K1_BUILD_PRECOMPUTED=ON to cmake."

Done.

CMakePresets.json
hebasto: Thanks! [Amended](https://github.com/bitcoin-core/secp256k1/pull/1791#issuecomment-4661608463).

cmake/Precompute.cmake
```diff
@@ -0,0 +1,13 @@
...
+  set_target_properties(secp256k1_precompute_${base_name} PROPERTIES
+    COMPILE_DEFINITIONS VERIFY
+    RUNTIME_OUTPUT_DIRECTORY ${PROJECT_BINARY_DIR}
+  )
+  add_custom_command(
+    OUTPUT ${PROJECT_BINARY_DIR}/src/precomputed_${base_name}.c
+    COMMAND secp256k1_precompute_${base_name}
+    DEPENDS secp256k1_precompute_${base_name}
+    WORKING_DIRECTORY ${PROJECT_BINARY_DIR}
+  )
```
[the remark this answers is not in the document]
hebasto: `secp256k1_precompute_${base_name}` is an executable target name.

From CMake [docs](https://cmake.org/cmake/help/latest/command/add_custom_command.html):
> If `COMMAND` specifies an executable target name (created by the [`add_executable()`](https://cmake.org/cmake/help/latest/command/add_executable.html) command), it will automatically be replaced by the location of the executable created at build time...

src/CMakeLists.txt
```diff
@@ -45,10 +45,22 @@ if(SECP256K1_ENABLE_MODULE_ECDH)
...
 
-add_library(secp256k1_precomputed OBJECT EXCLUDE_FROM_ALL
-  precomputed_ecmult.c
-  precomputed_ecmult_gen.c
-)
+add_library(secp256k1_precomputed OBJECT EXCLUDE_FROM_ALL)
+if(SECP256K1_BUILD_PRECOMPUTED)
+  include(Precompute)
+  precompute(ecmult)
+  precompute(ecmult_gen)
```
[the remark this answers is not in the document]
hebasto: Yes. The current CMake implementation doesn't support cross-compiling. And neither does Autotools.

FWIW, I have a branch that introduces this support, but I'm not sure whether the extra complexity is justified.

CMakePresets.json
hebasto: > Maybe leave it off for now?

That raises the question: which CI job should use `-DSECP256K1_BUILD_PRECOMPUTED=ON`?

kevkevinpal:
> If others agree, this raises the question of whether we want to get rid of that stuff in autotools. It never worked 100%. And it seems that all users who want this feature use CMake. (@kevkevinpal Why did you ask about autotools? Do you need this feature in autotools?)  

Trying to remember why but I think was just trying to understand why it wasn't available. If anything I agree with you and also prefer CMake.

src/CMakeLists.txt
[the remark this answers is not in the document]
real-or-random: > FWIW, I have a branch that introduces this support, but I'm not sure whether the extra complexity is justified.

Ah please don't. :D Years ago, when we always ran the precomputation during the build (no precomputed files in repo), we had support for this in Autotools. But handling the different compilers (one for host and one for target) was a nightmare for users who wanted to cross-compile and for us to maintain. It's one of the reasons why added the precomputed files to the repo. 

I think if we really want to support this during cross-compilation, we'd rather rewrite the precomputation helpers in python or sage.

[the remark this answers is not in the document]
hebasto: > Ah please don't. :D Years ago, when we always ran the precomputation during the build (no precomputed files in repo), we had support for this in Autotools.

TIL. Not arguing, but CMake has built-in ways to achieve this.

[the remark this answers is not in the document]
real-or-random: > CMake has built-in ways to achieve this.

autoconf has this here: https://www.gnu.org/software/autoconf-archive/ax_prog_cc_for_build.html (not exactly built-in but that's the standard way on autoconf)

I totally believe that the CMake way of doing this is much cleaner, but I think part of the problem is on a conceptual level. Now you get effectively two build configs with all their complexity. You have separate CFLAGS, CPPFLAGS, LDFLAGS, which raises all kinds of questions, e.g., which warning options do you want to set on the compiler for the build machine? Maybe the compiler is entirely different, e.g., you have a new clang for the build machine but an old gcc for the target, etc... And some of this complexity is necessarily pushed down to the user.

Just assuming the existence of Python or similar is probably simpler. Back then we had already considered using Python instead, but approach to add the precomputed files to the repo was just less work, so we went with this instead.

real-or-random:
> > Maybe we should just change the error message for now? "If this is an autotools build, try deleting precomputed_ecmult.c before the build. If this is a CMake build, pass -DSECP256K1_BUILD_PRECOMPUTED=ON to cmake."
> 
> Done.

We have the same message in precomputed_ecmult_gen.c. Sorry, I assumed that's implied. I should have mentioned this explicitly.

CMakePresets.json
real-or-random: My first thought is "Why not all?"

src/CMakeLists.txt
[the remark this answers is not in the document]
real-or-random: But do you think some error message like suggested by the LLM makes sense for now?

real-or-random:
For autotools builds on CI, we have `git diff` that checks the files in the repo are not outdated. We could do the same on CI for CMake, i.e., copy the files back to the source tree and run `git diff --exit-code`.

https://github.com/bitcoin-core/secp256k1/blob/0f4a7e6bf9d971addb2b851df7cd2777fc62b212/ci/ci.sh#L138-L147

hebasto:
> > > Maybe we should just change the error message for now? "If this is an autotools build, try deleting precomputed_ecmult.c before the build. If this is a CMake build, pass -DSECP256K1_BUILD_PRECOMPUTED=ON to cmake."
> > 
> > 
> > Done.
> 
> We have the same message in precomputed_ecmult_gen.c. Sorry, I assumed that's implied. I should have mentioned this explicitly.

Of course, I've regenerated `precomputed_ecmult_gen.c` in the source tree :)

real-or-random:
> > > > Maybe we should just change the error message for now? "If this is an autotools build, try deleting precomputed_ecmult.c before the build. If this is a CMake build, pass -DSECP256K1_BUILD_PRECOMPUTED=ON to cmake."
> > > 
> > > 
> > > Done.
> > 
> > 
> > We have the same message in precomputed_ecmult_gen.c. Sorry, I assumed that's implied. I should have mentioned this explicitly.
> 
> Of course, I've regenerated `precomputed_ecmult_gen.c` in the source tree :)

Oops, we're talking past each other. Sorry, I meant to write `precompute_ecmult_gen.c` (without the `d`). 

So you've changed `precompute_ecmult.c` and `precomputed_ecmult.c` but not `precompute_ecmult_gen.c` and `precomputed_ecmult_gen.c`.

src/CMakeLists.txt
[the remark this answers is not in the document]
hebasto: > But do you think some error message like suggested by the LLM makes sense for now?

Done.

hebasto:
Reworked per recent feedback. Added [`CODEGEN`](https://cmake.org/cmake/help/latest/policy/CMP0171.html) support for modern CMake.

The PR description has been updated.

> So you've changed `precompute_ecmult.c` and `precomputed_ecmult.c` but not `precompute_ecmult_gen.c` and `precomputed_ecmult_gen.c`.

My bad. Fixed.

> For autotools builds on CI, we have `git diff` that checks the files in the repo are not outdated. We could do the same on CI for CMake, i.e., copy the files back to the source tree and run `git diff --exit-code`.

Done.

CMakePresets.json
```diff
@@ -7,7 +7,8 @@
       "cacheVariables": {
         "SECP256K1_EXPERIMENTAL": "ON",
         "SECP256K1_ENABLE_MODULE_RECOVERY": "ON",
-        "SECP256K1_BUILD_EXAMPLES": "ON"
+        "SECP256K1_BUILD_EXAMPLES": "ON",
+        "SECP256K1_BUILD_PRECOMPUTED": "OFF"
```
real-or-random: nit: Why not omit this line entirely?

.github/workflows/ci.yml
```diff
@@ -538,6 +538,10 @@ jobs:
     name: ${{ matrix.configuration.job_name }}
     # See: https://github.com/actions/runner-images#available-images.
     runs-on: windows-2022
+    defaults:
+      run:
+        # Enforce fail-fast behavior for PowerShell.
+        shell: pwsh -Command "$PSNativeCommandUseErrorActionPreference = $true; $ErrorActionPreference = 'Stop'; & '{0}'"
```
real-or-random, line 544: Shouldn't this be implied according to [the docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#exit-codes-and-error-action-preference)?

hebasto: > Shouldn't this be implied according to [the docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#exit-codes-and-error-action-preference)?

The docs state: "Fail-fast behavior _when possible_." Effectively, it works for cmdlets only. Non-zero exit codes from native commands are ignored unless `$PSNativeCommandUseErrorActionPreference` is also set.

CMakePresets.json
hebasto: Thanks! Dropped.

.github/workflows/ci.yml
real-or-random: Sigh, why can't we just have nice things...

hebasto:

PR #1793, opened 2025-12-30
doc/bench: added help text for SECP256K1_BENCH_ITERS env var for bench_ecmult

kevkevinpal:
## Description
I noticed that in both `bench` and `bench_internal` there was a help text mentioning to the users that you could use `SECP256K1_BENCH_ITERS` to customize the amount of iterations for each benchmark, but `bench_ecmult` did not have this text so I added it in.

There is a caveat that if `SECP256K1_BENCH_ITERS` is less than 3, then `run_ecmult_multi_bench` will not run. Should I add that to the help text?

---

### Before
```
$ ./build/bin/bench_ecmult -h
Benchmark EC multiplication algorithms

Usage: ./build/bin/bench_ecmult <help|pippenger_wnaf|strauss_wnaf|simple>
The output shows the number of multiplied and summed points right after the
function name. The letter 'g' indicates that one of the points is the generator.
The benchmarks are divided by the number of points.

default (ecmult_multi): picks pippenger_wnaf or strauss_wnaf depending on the
                        batch size
pippenger_wnaf:         for all batch sizes
strauss_wnaf:           for all batch sizes
simple:                 multiply and sum each point individually
```
### After
```
$ ./build/bin/bench_ecmult -h
Benchmark EC multiplication algorithms

The default number of iterations for each benchmark is 10000. This can be
customized using the SECP256K1_BENCH_ITERS environment variable.

Usage: ./build/bin/bench_ecmult <help|pippenger_wnaf|strauss_wnaf|simple>
The output shows the number of multiplied and summed points right after the
function name. The letter 'g' indicates that one of the points is the generator.
The benchmarks are divided by the number of points.

default (ecmult_multi): picks pippenger_wnaf or strauss_wnaf depending on the
                        batch size
pippenger_wnaf:         for all batch sizes
strauss_wnaf:           for all batch sizes
simple:                 multiply and sum each point individually
```

### DIFF
--- a/src/bench_ecmult.c
+++ b/src/bench_ecmult.c
@@ -19,8 +19,11 @@
 
 #define POINTS 32768
 
-static void help(char **argv) {
+static void help(char **argv, int default_iters) {
     printf("Benchmark EC multiplication algorithms\n");
+    printf("\n");
+    printf("The default number of iterations for each benchmark is %d. This can be\n", default_iters);
+    printf("customized using the SECP256K1_BENCH_ITERS environment variable.\n");
     printf("\n");
     printf("Usage: %s <help|pippenger_wnaf|strauss_wnaf|simple>\n", argv[0]);
     printf("The output shows the number of multiplied and summed points right after the\n");
@@ -308,7 +311,8 @@ int main(int argc, char **argv) {
     int i, p;
     size_t scratch_size;
 
-    int iters = get_iters(10000);
+    int default_iters = 10000;
+    int iters = get_iters(default_iters);
 
     data.ecmult_multi = secp256k1_ecmult_multi_var;
 
@@ -316,7 +320,7 @@ int main(int argc, char **argv) {
         if(have_flag(argc, argv, "-h")
            || have_flag(argc, argv, "--help")
            || have_flag(argc, argv, "help")) {
-            help(argv);
+            help(argv, default_iters);
             return EXIT_SUCCESS;
         } else if(have_flag(argc, argv, "pippenger_wnaf")) {
             printf("Using pippenger_wnaf:\n");
@@ -328,7 +332,7 @@ int main(int argc, char **argv) {
             printf("Using simple algorithm:\n");
         } else {
             fprintf(stderr, "%s: unrecognized argument '%s'.\n\n", argv[0], argv[1]);
-            help(argv);
+            help(argv, default_iters);
             return EXIT_FAILURE;
         }
     }
@@ -381,6 +385,8 @@ int main(int argc, char **argv) {
                 run_ecmult_multi_bench(&data, i << p, 1, iters);
             }
         }
+    } else {
+        printf("Skipping some benchmarks due to SECP256K1_BENCH_ITERS <= 2\n");
     }
 
     if (data.scratch != NULL) {


real-or-random:
> There is a caveat that if `SECP256K1_BENCH_ITERS` is less than 3, then `run_ecmult_multi_bench` will not run. Should I add that to the help text?

I don't see why this is. Can you explain?

If that's the case, we could just check this condition and bail out (instead of documenting it).

kevkevinpal:
> I don't see why this is. Can you explain?

Yea if you take a look at [this block of code](https://github.com/bitcoin-core/secp256k1/blob/master/src/bench_ecmult.c#L375-L384) if iters is less than 3 we don't run `run_ecmult_multi_bench`

> If that's the case, we could just check this condition and bail out (instead of documenting it).

We already bail out in [this block](https://github.com/bitcoin-core/secp256k1/blob/master/src/bench_ecmult.c#L375-L384), I was just suggesting if we add documentation in the help text for `SECP256K1_BENCH_ITERS` we can note that it will bail out. But we don't need to if that is too verbose.

real-or-random:
Oh yes, I missed that `if`. What about adding an `else` block that prints a message like "Skipping some benchmarks due to SECP256K1_BENCH_ITERS <= 2"?

kevkevinpal:
> Oh yes, I missed that if. What about adding an else block that prints a message like "Skipping some benchmarks due to SECP256K1_BENCH_ITERS <= 2"?

Sounds good I updated it in [8f331c9](https://github.com/bitcoin-core/secp256k1/pull/1793/commits/8f331c9061517444a24233858320ddd951a21fba)

kevkevinpal:
This might be good for a follow-up but I noticed that if we put any value for SECP256K1_BENCH_ITERS not greater than 0

It will print `Floating point exception (core dumped)`

A simple check would fix that 

Example
```
(addBenchItersEnvVarHelp) secp256k1 $ SECP256K1_BENCH_ITERS=0 ./build/bin/bench

Floating point exception (core dumped)
(addBenchItersEnvVarHelp) secp256k1 $ SECP256K1_BENCH_ITERS=abc ./build/bin/bench

Floating point exception (core dumped)
(addBenchItersEnvVarHelp) secp256k1 $ SECP256K1_BENCH_ITERS= ./build/bin/bench

Floating point exception (core dumped)
```

real-or-random (approved):
utACK bd5ced1fe174e4568261e66025b93c922d8e4be3

real-or-random:
> This might be good for a follow-up but I noticed that if we put any value for SECP256K1_BENCH_ITERS not greater than 0
> 
> It will print `Floating point exception (core dumped)`

Indeed, I had noticed the same when testing this earlier. Happy to see a fix for this.

hebasto (approved):
ACK bd5ced1fe174e4568261e66025b93c922d8e4be3, I have reviewed the code and it looks OK. Tested on Ubuntu 25.10.

While aligning implementation across all benchmarks, `argv` could be passed to the `help()` in `bench.c` and `bench_internal.c`.

@real-or-random 
A side question. What does 77 refer to in: https://github.com/bitcoin-core/secp256k1/blob/2d9137ce9d7e576bfada2859292ecaa0887a6b8c/src/bench_ecmult.c#L375 ?

kevkevinpal:
> @real-or-random 
> A side question. What does 77 refer to in: https://github.com/bitcoin-core/secp256k1/blob/2d9137ce9d7e576bfada2859292ecaa0887a6b8c/src/bench_ecmult.c#L375 ?

I see it was added in this PR https://github.com/bitcoin-core/secp256k1/pull/722. 

It might have been added so Travis CI doesn't run this benchmark so many times, thus slowing down the CI.

kevkevinpal:
> While aligning implementation across all benchmarks, argv could be passed to the help() in bench.c and bench_internal.c.

I can work on this and open in a separate refactor/bench PR to keep the scope of this one minimal.

jonasnick (approved):
ACK bd5ced1fe174e4568261e66025b93c922d8e4be3

PR #1794, opened 2026-01-07
ecmult: Use size_t for array indices

real-or-random:
I don't think the current code is incorrect, but using `size_t` improves readability because the type makes it clear that we're dealing with array indices.

Also, making the result of the `ECMULT_TABLE_SIZE` macro (hopefully) a `size_t` fixes a compiler warning on MSVC, see #1791.

### DIFF
--- a/src/ecmult_impl.h
+++ b/src/ecmult_impl.h
@@ -70,10 +70,10 @@
  *  Lastly the zr[0] value, which isn't used above, is set so that:
  *  - a.z = z(pre_a[0]) / zr[0]
  */
-static void secp256k1_ecmult_odd_multiples_table(int n, secp256k1_ge *pre_a, secp256k1_fe *zr, secp256k1_fe *z, const secp256k1_gej *a) {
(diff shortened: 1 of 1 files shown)


hebasto:
Concept ACK.

I came to the same idea while working on https://github.com/bitcoin-core/secp256k1/pull/1791, but left it for later :)

real-or-random:
> * `size_t i` in: https://github.com/bitcoin-core/secp256k1/blob/2d9137ce9d7e576bfada2859292ecaa0887a6b8c/src/ecmult_impl.h#L339-L348

I think there's no perfect solution here, but fixing this one will open up the entire business of cleaning up types in `secp256k1_ecmult_wnaf` (see #1795), and I think this should be a separate PR. 

The same is true for the stuff in ecmult_gen; I think this could be a separate PR.

hebasto (approved):
re-ACK 47eb70959a91bbfa10dc209e8ccbe7350f5b9b04.

real-or-random:
@theStack Want to review this?

PR #1795, opened 2026-01-07
ecmult: Clean up integer types in _ecmult_wnaf

### LINKED ISSUE #1769: `_ecmult_wnaf` relies on `int` having at least 32 value bits
The current `secp256k1_ecmult_wnaf` needs the unstated and unchecked assumption that `int` has at least 32 value bits when it VERIFY_CHECKs that  `w <= 31`. In practice, we call it only with `WINDOW_A == 5` and `WINDOW_G == ECMULT_WINDOW_SIZE` where the latter is configurable in the range `2..24`. 

A consequence of this "bug" is that the code fails on a 16-bit platform if you set `ECMULT_WINDOW_SIZE > 16`. I don't think we need to support this, but code without unchecked assumptions is bad. So I suggest that we rewrite the function to use `int32_t` instead of `int` even if we don't use my macro approach. (We can keep the "bit position" arguments and variables `int`.)

Alternatively, we could add the assumption that `INT_MAX >= INT32_MAX` but this forbids 16-bit platforms, and the code seems to work on them in principle; see https://github.com/bitcoin-core/secp256k1/issues/792#issuecomment-674488586.

_Originally posted by @real-or-random in https://github.com/bitcoin-core/secp256k1/pull/1761#pullrequestreview-3432695127_

real-or-random:
Fixes #1769.

On top of #1794.

### DIFF
--- a/src/ecmult_impl.h
+++ b/src/ecmult_impl.h
@@ -169,7 +169,7 @@ static size_t secp256k1_ecmult_wnaf(int32_t *wnaf, size_t len, const secp256k1_s
     VERIFY_CHECK(wnaf != NULL);
     VERIFY_CHECK(len <= 256);
     VERIFY_CHECK(a != NULL);
-    VERIFY_CHECK(2 <= w && w <= 31);
+    VERIFY_CHECK(2 <= w && w <= 30);
 
     for (bit = 0; bit < len; bit++) {
         wnaf[bit] = 0;
--- a/src/ecmult_impl.h
+++ b/src/ecmult_impl.h
@@ -114,7 +114,7 @@ static void secp256k1_ecmult_odd_multiples_table(size_t n, secp256k1_ge *pre_a, 
     secp256k1_fe_mul(z, &ai.z, &d.z);
 }
 
-SECP256K1_INLINE static void secp256k1_ecmult_table_verify(int n, int w) {
+SECP256K1_INLINE static void secp256k1_ecmult_table_verify(int32_t n, size_t w) {
     (void)n;
     (void)w;
     VERIFY_CHECK(((n) & 1) == 1);
@@ -122,7 +122,7 @@ SECP256K1_INLINE static void secp256k1_ecmult_table_verify(int n, int w) {
     VERIFY_CHECK((n) <=  ((1 << ((w)-1)) - 1));
 }
 
-SECP256K1_INLINE static void secp256k1_ecmult_table_get_ge(secp256k1_ge *r, const secp256k1_ge *pre, int n, int w) {
+SECP256K1_INLINE static void secp256k1_ecmult_table_get_ge(secp256k1_ge *r, const secp256k1_ge *pre, int32_t n, size_t w) {
     secp256k1_ecmult_table_verify(n,w);
     if (n > 0) {
         *r = pre[(n-1)/2];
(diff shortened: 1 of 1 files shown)


hebasto:
Concept ACK.

src/ecmult_impl.h
```diff
@@ -159,15 +159,15 @@ SECP256K1_INLINE static void secp256k1_ecmult_table_get_ge_storage(secp256k1_ge
...
  *    than the number of bits in the (absolute value) of the input.
  */
-static int secp256k1_ecmult_wnaf(int *wnaf, int len, const secp256k1_scalar *a, int w) {
+static size_t secp256k1_ecmult_wnaf(int32_t *wnaf, size_t len, const secp256k1_scalar *a, size_t w) {
     secp256k1_scalar s;
-    int last_set_bit = -1;
-    int bit = 0;
-    int sign = 1;
-    int carry = 0;
+    size_t last_set_bit = -1;
```
theStack, line 164: nit: maybe
```suggestion
    size_t last_set_bit = SIZE_MAX;
```
(that's what we use in [`secp256k1_ge_set_all_gej_var`](https://github.com/bitcoin-core/secp256k1/blob/ebf594320dc838b9de1abb54d5ba98cef84f4297/src/group_impl.h#L239), fwiw)

theStack (commented):
Concept ACK

mnfadel:
Concept ACK on moving these to explicit widths — the `size_t` / `int32_t` split matches the array-index convention established in #1794, and dropping `0 <= len` is right now that `len` is unsigned.

Two observations, one of which I think deserves a mention in the commit message.

### The `w` bound change looks like a correctness fix, not a type cleanup

```diff
-    VERIFY_CHECK(2 <= w && w <= 31);
+    VERIFY_CHECK(2 <= w && w <= 30);
```

Unless I'm misreading this, it isn't cosmetic. In the loop body:

```c
carry = (word >> (w-1)) & 1;
word -= carry << w;
```

`carry` is `int32_t`, so at `w == 31` the expression `carry << w` shifts a `1` into the sign bit of a 32-bit signed type, which is UB. If that's right, the previous `w <= 31` bound already admitted UB for a caller passing the maximum permitted window, and this PR quietly closes it.

In practice `w` is only ever `WINDOW_A` / `WINDOW_G`, so nothing is affected today — but `VERIFY_CHECK` documents the contract, and it seems worth stating that `30` is *required* rather than merely tidier. A reader scanning a commit titled "clean up integer types" wouldn't expect the accepted parameter range to narrow, so calling it out in the message (or splitting it into its own commit) would make the history easier to follow.

### `size_t last_set_bit = -1;`

This is well-defined, and the wrap is load-bearing: with no set bits it stays `SIZE_MAX`, and `last_set_bit + 1` wraps back to `0`, which is the intended return. But the correctness of the return value now rests on a wraparound that nothing in the function mentions.

Would `size_t last_set_bit = SIZE_MAX;` be preferable? It denotes the same value explicitly and avoids an implicit `-1` → unsigned conversion — which is the class of diagnostic that motivated the `ECMULT_TABLE_SIZE` change in #1794. Failing that, a one-line comment noting the wrap is deliberate would cover it.

---

Happy to build and run the test suite on this branch and follow up with a tested ACK.

PR #1797, opened 2026-01-10
ci: fix shell injection in docker action

RinZ27:
Refactored the `docker run` command in the `run-in-docker-action` to use environment variables instead of direct GitHub context interpolation. 

Directly embedding context data like `github.workspace` or `inputs.command` into shell scripts can lead to injection if the data contains unexpected characters. Mapping these to environment variables and referencing them within the script is more robust.

Tested this change by ensuring the variable mapping correctly passes data to the docker container.

### DIFF
--- a/.github/actions/run-in-docker-action/action.yml
+++ b/.github/actions/run-in-docker-action/action.yml
@@ -40,14 +40,20 @@ runs:
       shell: bash
 
     - # Tell Docker to pass environment variables in `env` into the container.
-      run: >
-        docker run \
-          $(echo '${{ toJSON(env) }}' | jq -r 'keys[] | "--env \(.) "') \
-          --volume ${{ github.workspace }}:${{ github.workspace }} \
-          --workdir ${{ github.workspace }} \
-          $(docker images -q | head -n1) \
-          bash -c "
-            git config --global --add safe.directory ${{ github.workspace }}
-            ${{ inputs.command }}
-          "
       shell: bash
+      env:
+        CMD_TO_RUN: ${{ inputs.command }}
+        ENV_JSON: ${{ toJSON(env) }}
+      run: |
+        # Use a temporary file to store the environment variables JSON
+        # and pass it via an environment variable to avoid shell injection.
+        echo "$ENV_JSON" > "$RUNNER_TEMP/env.json"
+        docker run \
+          $(jq -r 'keys[] | select(. != "CMD_TO_RUN" and . != "ENV_JSON") | "--env \(.) "' "$RUNNER_TEMP/env.json") \
+          --volume "$GITHUB_WORKSPACE":"$GITHUB_WORKSPACE" \
+          --workdir "$GITHUB_WORKSPACE" \
+          "${{ steps.main_builder.outputs.imageid || steps.retry_builder.outputs.imageid }}" \
+          bash -c '
+            git config --global --add safe.directory "$1"
+            eval "$2"
+          ' -- "$GITHUB_WORKSPACE" "$CMD_TO_RUN"


RinZ27:
@hebasto Thanks for the review. I've updated the action to use the built-in `GITHUB_WORKSPACE` and removed the redundant variable. 

Regarding the injection hardening: while these contexts aren't directly user-controlled here, mapping them to environment variables is a more robust way to handle multiline inputs and prevents potential quote-mangling in the `bash -c` wrapping. 

@real-or-random Thanks for the ACK. Glad to see support for establishing safer patterns in the CI scripts.

hebasto:
@RinZ27 

Could you please squash your commits?

RinZ27:
@hebasto Done, squashed into a single commit.

.github/actions/run-in-docker-action/action.yml
```diff
@@ -40,14 +40,16 @@ runs:
       shell: bash
 
     - # Tell Docker to pass environment variables in `env` into the container.
-      run: >
+      shell: bash
+      env:
+        CMD_TO_RUN: ${{ inputs.command }}
```
hebasto, line 45: Not a big deal, but `CMD_TO_RUN` now leaks into the container.

RinZ27: Right, it does end up in the container env. It's the trade-off for safely passing the command without interpolation risks. Given it's an ephemeral CI job, this seems acceptable.

real-or-random: I think unsetting it within the command substitution would avoid it (if we want to):
```
$(unset CMD_TO_RUN; echo '${{ toJSON(env) }}' | jq -r 'keys[] | "--env \(.) "') \
```

RinZ27:
Good catch. I've updated the `jq` filter to specifically exclude `CMD_TO_RUN` so it doesn't leak into the container environment. Fixed in the latest commit.

.github/actions/run-in-docker-action/action.yml
```diff
@@ -40,14 +40,16 @@ runs:
...
+      run: |
         docker run \
-          $(echo '${{ toJSON(env) }}' | jq -r 'keys[] | "--env \(.) "') \
-          --volume ${{ github.workspace }}:${{ github.workspace }} \
-          --workdir ${{ github.workspace }} \
+          $(echo '${{ toJSON(env) }}' | jq -r 'keys[] | select(. != "CMD_TO_RUN") | "--env \(.) "') \
```
hebasto: This approach does not protect from the injection via `${{ toJSON(env) }}`.

RinZ27:
@hebasto Fair point. Using a temporary file to store the JSON env instead of piping it through `echo` should close that gap, as it avoids shell interpretation of the JSON string entirely. I've updated the action to use this more isolated approach and squashed the commits.

real-or-random:
I don't think the latest change does anything meaningful. If you can break out of `$(echo '${{ toJSON(env) }}' ...)` (because `env` contains a `'`), then you can also break out of `echo '${{ toJSON(env) }}'` not inside a command substitution.

RinZ27:
@real-or-random Valid point. Switched to mapping the JSON blob to a dedicated environment variable before writing it to the temporary file. This keeps the data entirely out of the shell's command line and avoids interpretation issues if `env` contains single quotes.

Also excluded the new `ENV_JSON` variable from the container's `--env` flags to keep things clean.

real-or-random:
I see you're still working on this, but I think you haven't addressed this comment:

> If you think I'm missing something and these changes have some value, then I'd ask you to provide a clear model of an attacker that they (should) protect against. Then we can clearly evaluate whether it's worth protecting against this kind of attacker.

RinZ27:
@real-or-random Command injection from environment variables containing shell-sensitive characters like single quotes or newlines is what I'm primarily trying to block here. 

Someone who can influence an environment variable used in the workflow—perhaps through a PR from a fork or a compromised upstream job—could use a single quote to break out of the `run` script's wrapping. If I interpolate `${{ toJSON(env) }}` directly into a shell command, the runner might end up executing arbitrary code if the data is malformed.

Using a dedicated environment variable to hold the JSON blob and then reading it from a file keeps the data entirely out of the shell's command line. Moving the data into a file-based workflow ensures the script is robust against multiline or untrusted inputs, so I don't have to worry about the shell misinterpreting the data.

real-or-random:
> Someone who can influence an environment variable used in the workflow—perhaps through a PR from a fork or a compromised upstream job

That's not a "clear attacker model" (if you ask me).

Regarding the points you mention explicitly: PRs can change the entire workflow, including steps and commands to run. Compromised upstream jobs can contain arbitrary steps, too.

RinZ27:
@real-or-random While I agree that a compromised workflow file provides broader attack vectors, my main focus here is following GitHub's recommended security best practices regarding context interpolation.

Interpolating data directly into a shell script forces the interpreter to parse it as code, which often makes the CI brittle. Legitimate characters like single quotes in branch names or newlines in inputs can easily mangle the `docker run` command, regardless of any malicious intent.

Mapping these contexts to environment variables and passing them as arguments ensures that the data is treated as raw data, not as executable code. This makes the action more robust and eliminates a whole class of parsing errors caused by unexpected strings.

If the maintainers feel this added complexity isn't worth the robustness gain for this project, I'm happy to close the PR. I'll leave the final call to your judgment.

fanquake:
~0. This seems to have already taken up too much time, for very unclear/little gain.

real-or-random:
> If the maintainers feel this added complexity isn't worth the robustness gain for this project, I'm happy to close the PR. I'll leave the final call to your judgment.

Let's close it. There's no clear support for this PR.

PR #1803, opened 2026-01-22
doc: add ctime_tests usage instructions to README

### LINKED ISSUE #1697: test: `ctime_tests` usage always suggests using libtool
It will probably be suprising for someone building binaries using CMake to be recommened to run them under libtool:
```bash
cmake -B build
cmake --build build
./build/bin/ctime_tests 
This test can only usefully be run inside valgrind because it was not compiled under msan.
Usage: libtool --mode=execute valgrind ./ctime_tests
```

8144225309:
## Motivation

When building with CMake and running `ctime_tests`, users see a message suggesting `libtool`, which is Autotools-specific:

### Before
```
$ ./build/bin/ctime_tests
This test can only usefully be run inside valgrind because it was not compiled under msan.
Usage: libtool --mode=execute valgrind ./ctime_tests
```

This is confusing for CMake users since they don't have libtool. Reported in #1697.

## Solution

Following the approach in #1778, replace the build-system-specific command with a pointer to README.md and document both methods there.

### After
```
$ ./build/bin/ctime_tests
This test can only usefully be run inside valgrind because it was not compiled under msan.
See README.md for usage instructions.
```

The README now includes instructions for both build systems.

Fixes #1697

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,19 @@ In "Developer Command Prompt for VS 2022":
     >cmake -B build -T ClangCL
     >cmake --build build --config RelWithDebInfo
 
+Constant-time tests
+-------------------
+
+This library comes with constant-time tests to verify that secret data does not leak through timing side channels. These tests require Valgrind.
+
+Autotools:
+
+    $ libtool --mode=execute valgrind ./ctime_tests
+
+CMake:
+
+    $ valgrind ./build/bin/ctime_tests
+
 Usage examples
 -----------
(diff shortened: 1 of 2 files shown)


hebasto:
Concept ACK.

8144225309:
do you like this style better or https://github.com/bitcoin-core/secp256k1/pull/1804 ??

real-or-random:
I prefer #1804.

PR #1805, opened 2026-01-23
ci: add scheduled workflow to clean caches

### LINKED ISSUE #1691: ci: Clean the cache regularly to test on new compiler snapshots
We can use a [GHA cron job](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#schedule) to [delete the cache](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows#force-deleting-cache-entries) every few weeks automatically.

_Originally posted by @real-or-random in https://github.com/bitcoin-core/secp256k1/issues/1690#issuecomment-2990002076_

8144225309:
Add a cron job that deletes all GitHub Actions caches on the 1st and 15th of each month. This ensures the Docker images are rebuilt periodically with newer compiler snapshots (gcc-snapshot, clang-snapshot).

The workflow:
- Runs on the 1st and 15th of each month at 00:00 UTC
- Can be triggered manually via `workflow_dispatch`
- Deletes all caches using `gh cache delete`

Fixes #1691

### DIFF
--- a/.github/workflows/clean-caches.yml
+++ b/.github/workflows/clean-caches.yml
@@ -1,0 +1,31 @@
+name: Clean Caches
+
+on:
+  # Run on the 1st and 15th of each month at 00:00 UTC (approximately every 2 weeks).
+  # This ensures compiler snapshots (gcc-snapshot, clang-snapshot) are refreshed
+  # regularly, as the Docker images will be rebuilt with newer versions.
+  schedule:
+    - cron: '0 0 1,15 * *'
+  # Allow manual trigger.
+  workflow_dispatch:
+
+permissions:
+  actions: write
+
+jobs:
+  clean:
+    name: "Delete all caches"
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Delete caches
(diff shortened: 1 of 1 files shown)


real-or-random:
This couls be a good change. Why are you closing the PR?

8144225309:
I was practicing writing the PR professionally while I tested the solution and I accidently shipped it early.  Feel free to use it or wait till monday at the latest for a replacement.  I only want to push things I've personally tested and I made a mistake here with my setup.  

edit: Can reopen for git cleanliness.

real-or-random:
> I only want to push things I've personally tested and I made a mistake here with my setup.

Sure, that makes sense! I think my main point is that I'm suggesting to always state the rationale for closing a PR (because otherwise it is obvious to anyone else).

8144225309:
Closed in favor of #1807, which uses `gh cache delete --all` instead of a manual loop.

PR #1807, opened 2026-01-25
ci: add workflow to clean caches periodically

### LINKED ISSUE #1691: ci: Clean the cache regularly to test on new compiler snapshots

8144225309:
Deletes all caches on the 1st and 15th of each month.

Fixes #1691

### DIFF
--- a/.github/workflows/caches.yml
+++ b/.github/workflows/caches.yml
@@ -1,0 +1,18 @@
+name: Delete all GitHub Actions caches
+on:
+  schedule:
+    - cron: '0 0 1,15 * *'
+  workflow_dispatch:
+
+permissions:
+  actions: write
+
+jobs:
+  clean:
+    name: Delete all GitHub Actions caches
+    runs-on: ubuntu-latest
+    steps:
+      - run: gh cache delete --all --succeed-on-no-caches
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}


.github/workflows/caches.yml
```diff
@@ -0,0 +1,18 @@
+name: Caches
```
real-or-random: I think the description can be clearer here. Something like "Delete all GitHub Actions caches"

.github/workflows/caches.yml
```diff
@@ -0,0 +1,18 @@
...
+  schedule:
+    - cron: '0 0 1,15 * *'
+  workflow_dispatch:
+
+permissions:
+  actions: write
+
+jobs:
+  clean:
+    name: Delete all caches
```
real-or-random: I'd use the same name here because there's really only a single job in this workflow

real-or-random:
cc @hebasto

real-or-random:
> I think we could avoid introducing a new workflow with the write permission by simply rotating the cache keys regularly.

That seems cleaner indeed.


---

edit: This means for this PR:

Approach NACK

hebasto:
Unrelated to the changes in this PR, the current GCC snapshot has been failing self-tests since [ed2924c7628343ff8b699f320a05c85bdb9c2837](https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=ed2924c7628343ff8b699f320a05c85bdb9c2837).

UPDATE: The issue should be fixed in [a6c3c48148dd60a09ffab57019f163180c159d65](https://gcc.gnu.org/git/?p=gcc.git;a=commit;h=a6c3c48148dd60a09ffab57019f163180c159d65).

hebasto:
> > I think we could avoid introducing a new workflow with the write permission by simply rotating the cache keys regularly.
> 
> That seems cleaner indeed.

Done in https://github.com/bitcoin-core/secp256k1/pull/1816.

real-or-random:
Let me close this in favor of #1816.

PR #1810, opened 2026-01-26
Improve release process: one PR for both steps

sipa:


real-or-random:
I know it was me who suggested this, but I see now why we always used two PRs: If we have just a single PR, then CI won't run on the release commit, and that's a big drawback.

sipa:
Yeah, good point.

PR #1811, opened 2026-01-27
bench: Update help functions in bench and bench_internal

kevkevinpal:
### Motivation
This change is motivated by https://github.com/bitcoin-core/secp256k1/pull/1793#pullrequestreview-3644885897

> While aligning implementation across all benchmarks, argv could be passed to the help() in bench.c and bench_internal.c.

### Description

In the `bench` and `bench_internal` `help` functions `argv` was not being passed. In this change, we pass in argv and use it in the help text.

### DIFF
--- a/src/bench.c
+++ b/src/bench.c
@@ -12,7 +12,7 @@
 #include "util.h"
 #include "bench.h"
 
-static void help(int default_iters) {
+static void help(const char *executable_path, int default_iters) {
     printf("Benchmarks the following algorithms:\n");
     printf("    - ECDSA signing/verification\n");
 
@@ -36,7 +36,7 @@ static void help(int default_iters) {
     printf("The default number of iterations for each benchmark is %d. This can be\n", default_iters);
     printf("customized using the SECP256K1_BENCH_ITERS environment variable.\n");
     printf("\n");
-    printf("Usage: ./bench [args]\n");
+    printf("Usage: %s [args]\n", executable_path);
     printf("By default, all benchmarks will be run.\n");
     printf("args:\n");
     printf("    help              : display this help and exit\n");
@@ -189,7 +189,7 @@ int main(int argc, char** argv) {
     int default_iters = 20000;
     int iters = get_iters(default_iters);
     if (iters == 0) {
-        help(default_iters);
+        help(argv[0], default_iters);
         return EXIT_FAILURE;
     }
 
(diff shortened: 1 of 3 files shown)


src/bench.c
```diff
@@ -12,7 +12,7 @@
 #include "util.h"
 #include "bench.h"
 
-static void help(int default_iters) {
+static void help(char **argv, int default_iters) {
```
real-or-random: nit: I know this uses only `argv[0]`, but still, this interface seems a bit cursed because you can't really use `argv` without `argc`.  Either you pass both `argc` and `argv` (like in `have_flag`), or you simply only `argv[0]`.

kevkevinpal: Thanks for the suggestion

I went ahead and did that in [5fb24a3](https://github.com/bitcoin-core/secp256k1/pull/1811/commits/5fb24a3e26ffc31dd06d87802a188cc65b22dd98)

This modifies the `help()` functions in `bench.c`, `bench_ecmult.c` and `bench_internal.c`

real-or-random: Thanks for the update. Now that I see this variant, I tend to think that the variant where only `argv[0]` is passed to `help()` is less cluttered. I mean all of these `help` functions only need `argv[0]` (and I see no reason why they should require the remaining args in the future).  Sorry for being specific only now...

Also, can you squash after you make this change?

kevkevinpal: No problem and sorry about that, I've updated and squashed in [e1598c7](https://github.com/bitcoin-core/secp256k1/pull/1811/commits/e1598c7ff2ec0ed938c62da96e5aa4a666dd129c)

Now we just have a `const char *executable_path` that we use. Let me know if that looks good!

Thanks for the review!

real-or-random:
cc @hebasto

src/bench.c
```diff
@@ -36,7 +36,7 @@ static void help(int default_iters) {
     printf("The default number of iterations for each benchmark is %d. This can be\n", default_iters);
     printf("customized using the SECP256K1_BENCH_ITERS environment variable.\n");
     printf("\n");
-    printf("Usage: ./bench [args]\n");
+    printf("Usage: %s <help|ecdsa|ecdsa_sign|ecdsa_verify|ec|ec_keygen>\n", executable_path);
```
hebasto: With optional modules enabled, the list of arguments becomes inconsistent with the descriptions below, which creates confusion. I think we should leave "[args]" as a placeholder accompanied by the explanatory list. Regardless of the approach, the help output should be consistent across all benchmark binaries.

kevkevinpal: Thanks for the review, that is a fair point. I've updated the 3 benches to use `[args]` instead in [c49c9be](https://github.com/bitcoin-core/secp256k1/pull/1811/commits/c49c9be5045cfade519446d6fd707b2c36e18387)

real-or-random (approved):
ACK c49c9be5045cfade519446d6fd707b2c36e18387

PR #1820, opened 2026-02-05
test: deduplicate scalar and field constants for n-1 and p-1

therohityadav:
Following up on feedback from #1819, this PR deduplicates the hardcoded constants for the group order minus one (n-1) and the field prime minus one (p-1) within `src/tests.c`.

### Changes:
* Defined `scalar_n_m1` and `fe_p_m1` as global static constants at the top of the file.
* Replaced all hardcoded hex occurrences of these values in:
  * `test_scalar_check_overflow`
  * `run_scalar_tests` (HALF_TESTS)
  * `scalar_cmov_test`
  * `scalar_cases` and `fe_cases` arrays
  * Local `scalar_minus_one` and `fe_minus_one` definitions.
* Verified that `make check` passes all tests.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -42,6 +42,20 @@
 static secp256k1_context *CTX = NULL;
 static secp256k1_context *STATIC_CTX = NULL;
 
+/* The group order n minus 1. This is the maximum valid scalar value. */
+#define EX_SCALAR_N_M1 SECP256K1_SCALAR_CONST( \
+    0xFFFFFFFFUL, 0xFFFFFFFFUL, 0xFFFFFFFFUL, 0xFFFFFFFEUL, \
+    0xBAAEDCE6UL, 0xAF48A03BUL, 0xBFD25E8CUL, 0xD0364140UL \
+)
+
+/* The field order p minus 1. */
+#define EX_FE_P_M1 SECP256K1_FE_CONST( \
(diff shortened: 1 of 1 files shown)


real-or-random (commented):
Concept ~0 

I'm not convinced this really improves readability or maintainability

therohityadav:
> Concept ~0
> 
> I'm not convinced this really improves readability or maintainability

Thanks for the review @real-or-random.

I understand the concern about readability . having explicit values in tests is often preferred.

However, my idea here was to treat n-1 and p-1 as constants rather than arbitrary test vectors. By defining them in one place , we eliminate the risk of typos that could occur when these hex strings are repeated across the file. It ensures that every test checking these boundaries is mathematically guaranteed to be checking the same boundary.

real-or-random:
I'm still not convinced that I want to spend time on this, but let me leave as a note here that a new test relevant to this PR has been added in b99a94c3827e1b8e8505648758512eb3cf4aae03.

theStack:
I tend to agree with @real-or-random that this change is not an improvement w.r.t. readability/maintainability (sorry for the back and forth, it was me who actually suggested this back then: https://github.com/bitcoin-core/secp256k1/pull/1819#pullrequestreview-3752526279).

PR #1842, opened 2026-04-01
WIP Implement post-quantum security

real-or-random:
In light of recent advancements in the area of quantum computing that reduce the number of qubits[^1][^2] and Toffoli gates[^2] necessary to solve the discrete logarithm problem on secp256k1, I believe it's time to take action and secure this library against quantum attacks. 

Implementation plan:
 - [x] 60892c97f660fbb4835661ecc75e9e823e162c33 ecmult_(gen/const): Implement post-quantum mode
 - [x] 9b9eb401fc9b2b9400f56c8fe4b0dbe1a1f9488a configure: Add --enable-post-quantum option 
 - [ ] Port configure option to CMake
 - [x] 9e2e6e7515896e2504ca04214475d5cf6dddd912 Double major version number (BIG breaking change) 
 - [ ] Reevaluate functionality
 - [ ] Fix CI
 - [ ] Add Changelog entry

[^1]:  _Clémence Chevignard, Pierre-Alain Fouque, André Schrottenloher_. Reducing the Number of Qubits in Quantum Discrete Logarithms on Elliptic Curves, EUROCRYPT 2026, https://eprint.iacr.org/2026/280, 
[^2]: _Ryan Babbush, Adam Zalcman, Craig Gidney, Michael Broughton, Tanuj Khattar, Hartmut Neven, Thiago Bergamaschi, Justin Drake, Dan Boneh_. Securing Elliptic Curve Cryptocurrencies against Quantum Vulnerabilities: Resource Estimates and Mitigations, preprint, https://arxiv.org/pdf/2603.28846

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -159,6 +159,11 @@ static const secp256k1_callback default_error_callback = {
 #define VERIFY_CHECK(cond)
 #endif
 
+#define POST_QUANTUM_CHECK(ctx) do { \
+    fprintf(stderr, "Forbes a day before 01/Apr/2026 Google Finds Quantum Computers Could Break Bitcoin Sooner Than Expected\n"); \
+    abort(); \
+} while(0)
+
 static SECP256K1_INLINE void *checked_malloc(const secp256k1_callback* cb, size_t size) {
     void *ret = malloc(size);
     if (ret == NULL) {
--- a/configure.ac
+++ b/configure.ac
@@ -197,6 +197,10 @@ AC_ARG_ENABLE(external_default_callbacks,
     AS_HELP_STRING([--enable-external-default-callbacks],[enable external default callback functions [default=no]]), [],
     [SECP_SET_DEFAULT([enable_external_default_callbacks], [no], [no])])
 
+AC_ARG_ENABLE(post_quantum,
+    AS_HELP_STRING([--enable-post-quantum],[enable post-quantum security [default=yes]]), [],
+    [SECP_SET_DEFAULT([enable_post_quantum], [yes], [yes])])
+
 # Test-only override of the (autodetected by the C code) "widemul" setting.
 # Legal values are:
 #  * int64 (for [u]int64_t),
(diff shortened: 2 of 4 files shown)


sipa:
Hmm, I would expect PQC mode to just be a "return false;" in all signature verification functions, but they seem unaffected here?

hebasto:
> I believe it's time to take action and secure this library against quantum attacks.

Indeed.

theStack:
For better marketing, I'd recommend quadrupling the major version number instead of only doubling (implying an EVEN BIGGER breaking change). Happy to re-review the diff thoroughly if you want to follow that suggestion.

src/util.h
```diff
@@ -159,6 +159,11 @@ static const secp256k1_callback default_error_callback = {
 #define VERIFY_CHECK(cond)
 #endif
 
+#define POST_QUANTUM_CHECK(ctx) do { \
+    fprintf(stderr, "Forbes a day before 01/Apr/2026 Google Finds Quantum Computers Could Break Bitcoin Sooner Than Expected\n"); \
+    abort(); \
```
furszy, line 164: tiny nit for the future:

Execute a GPT/Claude/Gemini prompt here.

furszy (commented):
Change is too complex. Should be divided across multiple PRs. The more, the better.

real-or-random:
> Hmm, I would expect PQC mode to just be a "return false;" in all signature verification functions, but they seem unaffected here?

That's a fair point, and I've considered this as well, but it's orthogonal to this PR. Changing the signature verification functions would entail a softfork in Bitcoin, and so this will need more discussion. What we could do is add new API calls such as `secp256k1_schorrsig_verify_pq` which just return 0 when post-quantum mode is enabled and fall back to ordinary verification otherwise. However, based on @furszy's feedback, I suggest that this is done in another PR.

I should have made this clearer in the initial comment, but the purpose of this PR to ensure that any keys created in post-quantum mode are safe (and the same for any ECDH secrets derived in post-quantum mode). This will also make the library secure for new applications adopting libsecp256k1 in the future, e.g., in 50 years from now.

real-or-random:
> For better marketing, I'd recommend quadrupling the major version number instead of only doubling (implying an EVEN BIGGER breaking change). Happy to re-review the diff thoroughly if you want to follow that suggestion.

I think that's overkill for now, but we could do if @sipa's suggestion is implemented.

real-or-random:
> Change is too complex. Should be divided across multiple PRs. The more, the better.

Agreed. I had a hard time splitting 9e2e6e7515896e2504ca04214475d5cf6dddd912, but I'm happy to move it to a follow-up PR.

sipa:
It is quite impressive how this PR manages to create public keys that reveal nothing about the corresponding private keys, even in a CRQC setting!

PR #1845, opened 2026-04-12
Improve checks for scalar _get_bits methods

peterdettman:
Improves the `VERIFY_CHECK`s in all `_scalar_get_bits_limb32` and `_scalar_get_bits_var` methods.

The initial prompt was noticing that scalar_4x64_impl/`secp256k1_scalar_get_bits_limb32` was not restricting to 32-bit limbs correctly. Then missing range checks for `offset` were added and all such checks rewritten to avoid overflow.

With these changes, the _low and _4x64 impls of `_get_bits_var` can no longer forward to `_get_bits_limb32`, so those calls were inlined instead.

### DIFF
--- a/src/scalar_4x64_impl.h
+++ b/src/scalar_4x64_impl.h
@@ -41,7 +41,8 @@ SECP256K1_INLINE static void secp256k1_scalar_set_int(secp256k1_scalar *r, unsig
 SECP256K1_INLINE static uint32_t secp256k1_scalar_get_bits_limb32(const secp256k1_scalar *a, unsigned int offset, unsigned int count) {
     SECP256K1_SCALAR_VERIFY(a);
     VERIFY_CHECK(count > 0 && count <= 32);
-    VERIFY_CHECK((offset + count - 1) >> 6 == offset >> 6);
+    VERIFY_CHECK(offset <= 256 - count);
+    VERIFY_CHECK((offset + count - 1) >> 5 == offset >> 5);
 
     return (a->d[offset >> 6] >> (offset & 0x3F)) & (0xFFFFFFFF >> (32 - count));
 }
@@ -49,12 +50,13 @@ SECP256K1_INLINE static uint32_t secp256k1_scalar_get_bits_limb32(const secp256k
 SECP256K1_INLINE static uint32_t secp256k1_scalar_get_bits_var(const secp256k1_scalar *a, unsigned int offset, unsigned int count) {
     SECP256K1_SCALAR_VERIFY(a);
     VERIFY_CHECK(count > 0 && count <= 32);
-    VERIFY_CHECK(offset + count <= 256);
+    VERIFY_CHECK(offset <= 256 - count);
 
     if ((offset + count - 1) >> 6 == offset >> 6) {
-        return secp256k1_scalar_get_bits_limb32(a, offset, count);
+        return (a->d[offset >> 6] >> (offset & 0x3F)) & (0xFFFFFFFF >> (32 - count));
     } else {
         VERIFY_CHECK((offset >> 6) + 1 < 4);
+        VERIFY_CHECK((offset & 0x3F) > 0);
         return ((a->d[offset >> 6] >> (offset & 0x3F)) | (a->d[(offset >> 6) + 1] << (64 - (offset & 0x3F)))) & (0xFFFFFFFF >> (32 - count));
     }
 }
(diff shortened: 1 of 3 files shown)


src/scalar_4x64_impl.h
```diff
@@ -41,20 +41,22 @@ SECP256K1_INLINE static void secp256k1_scalar_set_int(secp256k1_scalar *r, unsig
 SECP256K1_INLINE static uint32_t secp256k1_scalar_get_bits_limb32(const secp256k1_scalar *a, unsigned int offset, unsigned int count) {
     SECP256K1_SCALAR_VERIFY(a);
     VERIFY_CHECK(count > 0 && count <= 32);
-    VERIFY_CHECK((offset + count - 1) >> 6 == offset >> 6);
+    VERIFY_CHECK(offset <= 256 - count);
+    VERIFY_CHECK((offset + count - 1) >> 5 == offset >> 5);
```
edilmedeiros, line 45: I'm convinced this will not break things and I like this more after fiddling with the bits.

But feels like the original check combined with line 43 was already sufficient for checking we are not able to cross limbs when trying to collect bits from the scalar.

One such case could be: `offset = 31`.

With the original check:
RHS: `offset >> 6 = 0`
LHS: `(31 + count - 1) >> 6` => `count < 34` so that RHS == LHS.

But line 43 already guarantee that `count <= 32`. Similar for the other limbs.

src/scalar_4x64_impl.h
```diff
@@ -41,20 +41,22 @@ SECP256K1_INLINE static void secp256k1_scalar_set_int(secp256k1_scalar *r, unsig
...
 
     return (a->d[offset >> 6] >> (offset & 0x3F)) & (0xFFFFFFFF >> (32 - count));
 }
 
 SECP256K1_INLINE static uint32_t secp256k1_scalar_get_bits_var(const secp256k1_scalar *a, unsigned int offset, unsigned int count) {
     SECP256K1_SCALAR_VERIFY(a);
     VERIFY_CHECK(count > 0 && count <= 32);
-    VERIFY_CHECK(offset + count <= 256);
+    VERIFY_CHECK(offset <= 256 - count);
```
edilmedeiros, line 53: I wonder if those aren't the same check.

I mean, `offset + count` could overflow while `256 - count` can't because of the first check.
Is that the intended fix?

src/scalar_4x64_impl.h
```diff
@@ -41,20 +41,22 @@ SECP256K1_INLINE static void secp256k1_scalar_set_int(secp256k1_scalar *r, unsig
...
 
     if ((offset + count - 1) >> 6 == offset >> 6) {
```
edilmedeiros, line 55: Here, we are checking if we are not crossing limbs when collecting bits. I think it's ok to inline the function in the following, but not so much if we are testing a different condition compared to the called function, which is what this PR proposes by changing line R45.

Why not
```suggestion
    if ((offset + count - 1) >> 5 == offset >> 5) {
```

as well?

edilmedeiros (commented):
Overall positive, this PR is indeed improving parameter checking. 

E.g. the invalid combination `offset = 300; count = 10;` would pass the original checks in `secp256k1_scalar_get_bits_limb32`.

Do you think it's worth adding some unit tests to avoid future regression?

src/scalar_4x64_impl.h
peterdettman: Despite this implementation being 4 x 64-bit limbs, I assume it must still conform to the generic scalar API, and this method (per name and documentation) limits the get to 32-bit limbs; the original check limits only to 64-bit limbs.

src/scalar_4x64_impl.h
peterdettman: Yes, overflow.

src/scalar_4x64_impl.h
peterdettman: This check is implementation-specific and can take advantage of the 64-bit limbs, unlike the API-32-bit one.

sipa (commented):
ACK 0cad3df503b8426a4304f0f433562f201759514e

src/scalar_4x64_impl.h
theStack, line 45: nit: we could add a comment to this check in each `_get_bits_limb32` implementation, explaining what it ensures (e.g. "requested bits must belong to the same 32-bit limb")

theStack (approved):
ACK 0cad3df503b8426a4304f0f433562f201759514e

real-or-random (approved):
utACK 0cad3df503b8426a4304f0f433562f201759514e

PR #1848, opened 2026-04-27
ci: Bump GCC snapshot major version to 17

hebasto:
See https://gcc.gnu.org/pipermail/gcc/2026-April/248048.html.


real-or-random (approved):
utACK 3cca6451a2521e3cfe353b3787ea9f86d6caee2f

PR #1850, opened 2026-04-29
musig: clear `pubnonce` output for invalid `seckey`

l0rinc:
This is a follow-up to PR #1849.

### Problem
`secp256k1_musig_nonce_gen` returns failure if its [arguments are invalid](https://github.com/bitcoin-core/secp256k1/blob/b11340b3ce2afac1b6ffda4ce5828c30621d2917/include/secp256k1_musig.h#L337), for example when a non-NULL [`seckey`](https://github.com/bitcoin-core/secp256k1/blob/b11340b3ce2afac1b6ffda4ce5828c30621d2917/include/secp256k1_musig.h#L347-L348) fails [secret-key validation](https://github.com/bitcoin-core/secp256k1/blob/b11340b3ce2afac1b6ffda4ce5828c30621d2917/src/modules/musig/session_impl.h#L391).

On that path, `secp256k1_musig_nonce_gen_internal` invalidates the [secret nonce](https://github.com/bitcoin-core/secp256k1/blob/b11340b3ce2afac1b6ffda4ce5828c30621d2917/src/modules/musig/session_impl.h#L414) on failure, but still unconditionally saves the derived [public nonce output](https://github.com/bitcoin-core/secp256k1/blob/b11340b3ce2afac1b6ffda4ce5828c30621d2917/src/modules/musig/session_impl.h#L433) before returning failure.

### Fix
Clear `pubnonce` before returning the invalid-`seckey` failure, so a failed nonce-generation call does not leave a newly written `pubnonce` behind.

The implementation still saves the public nonce before clearing it on failure (instead of branching around the save), to preserve constant-time control flow with respect to the secret-key validation result.

The second commit only extends existing MuSig API failure-path coverage for paths where `secp256k1_musig_nonce_gen_internal` already clears `pubnonce` at function entry before any of its own validation steps.

### DIFF
--- a/src/modules/musig/tests_impl.h
+++ b/src/modules/musig/tests_impl.h
@@ -287,6 +287,7 @@ static void musig_api_tests(void) {
     /* invalid seckey */
     CHECK(secp256k1_musig_nonce_gen(CTX, &secnonce[0], &pubnonce[0], session_secrand[0], max64, &pk[0], msg, &keyagg_cache, max64) == 0);
     CHECK(memcmp_and_randomize(secnonce[0].data, zeros132, sizeof(secnonce[0].data)) == 0);
+    CHECK(memcmp_and_randomize(pubnonce[0].data, zeros132, sizeof(pubnonce[0].data)) == 0);
 
     CHECK_ILLEGAL(CTX, secp256k1_musig_nonce_gen(CTX, &secnonce[0], &pubnonce[0], session_secrand[0], sk[0], NULL, msg, &keyagg_cache, max64));
     CHECK(memcmp_and_randomize(secnonce[0].data, zeros132, sizeof(secnonce[0].data)) == 0);
--- a/src/modules/musig/tests_impl.h
+++ b/src/modules/musig/tests_impl.h
@@ -265,6 +265,7 @@ static void musig_api_tests(void) {
 
     CHECK_ILLEGAL(STATIC_CTX, secp256k1_musig_nonce_gen(STATIC_CTX, &secnonce[0], &pubnonce[0], session_secrand[0], sk[0], &pk[0], msg, &keyagg_cache, max64));
     CHECK(memcmp_and_randomize(secnonce[0].data, zeros132, sizeof(secnonce[0].data)) == 0);
+    CHECK(memcmp_and_randomize(pubnonce[0].data, zeros132, sizeof(pubnonce[0].data)) == 0);
 
     CHECK_ILLEGAL(CTX, secp256k1_musig_nonce_gen(CTX, NULL, &pubnonce[0], session_secrand[0], sk[0], &pk[0], msg, &keyagg_cache, max64));
 
@@ -289,9 +290,11 @@ static void musig_api_tests(void) {
 
     CHECK_ILLEGAL(CTX, secp256k1_musig_nonce_gen(CTX, &secnonce[0], &pubnonce[0], session_secrand[0], sk[0], NULL, msg, &keyagg_cache, max64));
     CHECK(memcmp_and_randomize(secnonce[0].data, zeros132, sizeof(secnonce[0].data)) == 0);
+    CHECK(memcmp_and_randomize(pubnonce[0].data, zeros132, sizeof(pubnonce[0].data)) == 0);
 
     CHECK_ILLEGAL(CTX, secp256k1_musig_nonce_gen(CTX, &secnonce[0], &pubnonce[0], session_secrand[0], sk[0], &invalid_pk, msg, &keyagg_cache, max64));
     CHECK(memcmp_and_randomize(secnonce[0].data, zeros132, sizeof(secnonce[0].data)) == 0);
+    CHECK(memcmp_and_randomize(pubnonce[0].data, zeros132, sizeof(pubnonce[0].data)) == 0);
 
     CHECK(secp256k1_musig_nonce_gen(CTX, &secnonce[0], &pubnonce[0], session_secrand[0], sk[0], &pk[0], NULL, &keyagg_cache, max64) == 1);
     CHECK(memcmp_and_randomize(session_secrand[0], zeros132, sizeof(session_secrand[0])) == 0);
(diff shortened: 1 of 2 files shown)


src/modules/musig/tests_impl.h
```diff
@@ -286,6 +286,7 @@ static void musig_api_tests(void) {
     /* invalid seckey */
     CHECK(secp256k1_musig_nonce_gen(CTX, &secnonce[0], &pubnonce[0], session_secrand[0], max64, &pk[0], msg, &keyagg_cache, max64) == 0);
     CHECK(memcmp_and_randomize(secnonce[0].data, zeros132, sizeof(secnonce[0].data)) == 0);
+    CHECK(memcmp_and_randomize(pubnonce[0].data, zeros132, sizeof(pubnonce[0].data)) == 0);
```
theStack, line 290: in fde940f38a0ddaa3c4b556eeff42bcd3ed89c319: nit: I suspect this change was meant to be part of the next commit?

theStack (commented):
Concept ACK

AFAICT this pattern matches what we do in similar API functions (e.g. [`_schnorrsig_sign32`](https://github.com/bitcoin-core/secp256k1/blob/b11340b3ce2afac1b6ffda4ce5828c30621d2917/src/modules/schnorrsig/main_impl.h#L181), [`_keypair_create`](https://github.com/bitcoin-core/secp256k1/blob/b11340b3ce2afac1b6ffda4ce5828c30621d2917/src/modules/extrakeys/main_impl.h#L208), [`_ellswift_create`](https://github.com/bitcoin-core/secp256k1/blob/b11340b3ce2afac1b6ffda4ce5828c30621d2917/src/modules/ellswift/main_impl.h#L464)).

src/modules/musig/tests_impl.h
l0rinc: No, this is the one that reproduces the fix (it fails without the fix), the next commit adds these checks that were already correct. Would it help if I moved that commit first?

theStack: Ah, I see now. Swapping commits would indeed help to clarify that I think, but happy to (re-)ACK in either order.

theStack:
ACK 7ed2abc96871d78baaeafe2b664381f01068f49

src/modules/musig/tests_impl.h
l0rinc: Swapped the commits, thanks!

theStack (approved):
re-ACK d679d2cf91b9129d8f8394d58f1a54f6a82ede54

real-or-random:
> On that path, `secp256k1_musig_nonce_gen_internal` invalidates the [secret nonce](https://github.com/bitcoin-core/secp256k1/blob/b11340b3ce2afac1b6ffda4ce5828c30621d2917/src/modules/musig/session_impl.h#L414) on failure, but still unconditionally saves the derived [public nonce output](https://github.com/bitcoin-core/secp256k1/blob/b11340b3ce2afac1b6ffda4ce5828c30621d2917/src/modules/musig/session_impl.h#L433) before returning failure.

Why is that a problem?

theStack:
> > On that path, `secp256k1_musig_nonce_gen_internal` invalidates the [secret nonce](https://github.com/bitcoin-core/secp256k1/blob/b11340b3ce2afac1b6ffda4ce5828c30621d2917/src/modules/musig/session_impl.h#L414) on failure, but still unconditionally saves the derived [public nonce output](https://github.com/bitcoin-core/secp256k1/blob/b11340b3ce2afac1b6ffda4ce5828c30621d2917/src/modules/musig/session_impl.h#L433) before returning failure.
> 
> Why is that a problem?

From a reviewer's perspective, I see this PR as a small consistency improvement, fully enforcing the (unwritten?) rule "if any of the input parameters are invalid or the function fails for any other reason, all of the out parameters are zeroed/invalidated". Probably not an actual problem in practice though, considering that users should always check for return values before evaluating out-parameters in the first place.

real-or-random:
Sorry for the very short comment earlier.

> (unwritten?) rule "if any of the input parameters are invalid or the function fails for any other reason, all of the out parameters are zeroed/invalidated".

I'm not sure if we there is such a rule. I can't find any discussion on this, what I remember is that we want to make sure that output parameters are written to at all, simply to avoid uninitialized memory (because that can trigger UB).

> Probably not an actual problem in practice though, considering that users should always check for return values before evaluating out-parameters in the first place.

In this case, it's not a problem at all because the `pubnonce` is simply a random group element for which nobody knows the discrete logarithm (not even the code itself because it was deleted). Also, the `pubnonce` is (computationally) independent of whether the secret key was invalid or not. 

Of course, you can ask if there *should* be a general rule of the kind you mentioned above. Maybe it's good practice, but the drawbacks can be an increase in code complexity and (potentially) an impact on performance. Another spot where we don't follow this rule is `secp256k1_ecdh`.

If we can get consensus on this rule, then yes, let's do this, but then let's also write the rule down.

---

In some of these functions (including the one touched here), I think we're overdoing it with insisting on constant-time behavior. If the secret key is invalid (zero or overflow), it's totally fine to return early. We want to caller to branch on the return value, so not branching inside our code is a bit pointless. Any attacker who is able to measure time will be able to figure that out eventually because the caller will branch anyway. In this case, you can even observe on the network whether a pubnonce will be sent or not. 

I'd even go as far as saying that `secp256k1_scalar_set_b32_seckey` should internally declassify, or at least there should be a variant that does. What we have now encourages a constant-time coding style around that function. Then execution continues with an invalid intermediate values, and that can be problematic if we pass that intermediate value to other functions which don't expect it. That makes it hard to reason about the code. (Concrete examples: https://github.com/bitcoin-core/secp256k1/pull/1316 and https://github.com/BlockstreamResearch/secp256k1-zkp/pull/282).

So if we want to enforce this rule here, then I'd prefer an early return instead of a late memczero.

l0rinc:
Thanks for the detailed explanation, that makes sense.
Closing this since the current framing overstates the issue, and the broader output-clearing rule is not established.

PR #1861, opened 2026-06-03
refactor: introduce `_ecmult_gen_ge` helper (preventing accidental gej leaks)

theStack:
Scalar multiplication with the generator point frequently involves a conversion to affine coordinates and clearing out the temporary Jacobian group element object after to avoid leaking secret key material (see 765ef53335a3e0fafdafe1e757f6fe0789f2797f / #1579 for that last part), i.e. executing the following three functions:
* `secp256k1_ecmult_gen(ctx, &rj, ...)`
* `secp256k1_ge_set_gej(&r, &rj)`
* `secp256k1_gej_clear(&rj)`

This PR introduces a corresponding helper to deduplicate code and mitigate the risk that the last step is forgotten (which can easily happen, as it would not be detected by tests). It is applied in the code paths for ECDSA signing, Schnorr signing, public key creation and ecmult_gen blinding setup. The only remaining instance where we directly call `_ecmult_gen` is for [musig nonce generation](https://github.com/bitcoin-core/secp256k1/blob/a39093de15345a330fc8acbd09515fd150853bc8/src/modules/musig/session_impl.h#L416), as we apply batch inversion there for the two points.

The idea came up during a conversation with furszy, who caught that the gej clearing was missing in the silentpayments module (sending function) as well (see https://github.com/bitcoin-core/secp256k1/pull/1765#issuecomment-4482838033, [b10c mirror link](https://mirror.b10c.me/bitcoin-core-secp256k1/1765/#issuecomment-4482838033)).

If this gets conceptual support, I'd be curious to hear naming suggestions, as I'm not sure if the current one is fits well to the existing terminology (maybe `ecmult_gen_ge` or `ecmult_gen_to_affine`?).

### DIFF
--- a/src/bench_ecmult.c
+++ b/src/bench_ecmult.c
@@ -88,7 +88,7 @@ static void bench_ecmult_teardown_helper(bench_data* data, size_t* seckey_offset
             secp256k1_scalar_add(&sum_scalars, &sum_scalars, &s);
         }
     }
-    secp256k1_ecmult_gen(&data->ctx->ecmult_gen_ctx, &tmp, &sum_scalars);
+    secp256k1_ecmult_gen_gej(&data->ctx->ecmult_gen_ctx, &tmp, &sum_scalars);
     CHECK(secp256k1_gej_eq_var(&tmp, &sum_output));
 }
 
@@ -104,7 +104,7 @@ static void bench_ecmult_gen(void* arg, int iters) {
     int i;
 
     for (i = 0; i < iters; ++i) {
-        secp256k1_ecmult_gen(&data->ctx->ecmult_gen_ctx, &data->output[i], &data->scalars[(data->offset1+i) % POINTS]);
+        secp256k1_ecmult_gen_gej(&data->ctx->ecmult_gen_ctx, &data->output[i], &data->scalars[(data->offset1+i) % POINTS]);
     }
 }
 
--- a/src/ecdsa_impl.h
+++ b/src/ecdsa_impl.h
@@ -273,14 +273,12 @@ static int secp256k1_ecdsa_sig_verify(const secp256k1_scalar *sigr, const secp25
 
 static int secp256k1_ecdsa_sig_sign(const secp256k1_ecmult_gen_context *ctx, secp256k1_scalar *sigr, secp256k1_scalar *sigs, const secp256k1_scalar *seckey, const secp256k1_scalar *message, const secp256k1_scalar *nonce, int *recid) {
     unsigned char b[32];
-    secp256k1_gej rp;
     secp256k1_ge r;
     secp256k1_scalar n;
     int overflow = 0;
     int high;
 
-    secp256k1_ecmult_gen(ctx, &rp, nonce);
-    secp256k1_ge_set_gej(&r, &rp);
+    secp256k1_ecmult_gen_ge(ctx, &r, nonce);
     secp256k1_fe_normalize(&r.x);
     secp256k1_fe_normalize(&r.y);
     secp256k1_fe_get_b32(b, &r.x);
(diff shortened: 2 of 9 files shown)


real-or-random:
Concept ACK


> If this gets conceptual support, I'd be curious to hear naming suggestions, as I'm not sure if the current one is fits well to the existing terminology (maybe `ecmult_gen_ge` or `ecmult_gen_to_affine`?).

I think `ge_ecmult_gen` is in line with the naming of the other group functions. (We could also consider renaming the existing one to `gej_ecmult_gen`.)

theStack:
> > If this gets conceptual support, I'd be curious to hear naming suggestions, as I'm not sure if the current one is fits well to the existing terminology (maybe `ecmult_gen_ge` or `ecmult_gen_to_affine`?).
> 
> I think `ge_ecmult_gen` is in line with the naming of the other group functions. (We could also consider renaming the existing one to `gej_ecmult_gen`.)

Thanks, done. I went ahead and did the proposed `_ecmult_gen` -> `_gej_ecmult_gen` rename as well in a second commit (it can easily be dropped if considered out-of-scope; not sure what's a good stopping point for this PR, one could in theory even go further and rename the other ecmult_ functions to include the gej prefix as well...).

moriarti0281:
Thanks for the feedback.

I agree that naming consistency with existing group functions is important.
I’m fine with using `ge_ecmult_gen` if that better matches the current
terminology, and the renaming commit can be dropped if preferred.

I tried to keep the scope limited and avoided broader renaming for now,
but I’m happy to adjust based on maintainers’ preference.

real-or-random:
> I think `ge_ecmult_gen` is in line with the naming of the other group functions. (We could also consider renaming the existing one to `gej_ecmult_gen`.)

Now that I'm looking at this again, I start to see the drawback of this: These functions live in the ecmult module but now they have a `ge_` or `gej_` prefix which is usually reserved for the group module. 

What about a) moving the `ge_ecmult_gen` helper to group and b) keeping `ecmult_gen` as is in ecmult but re-exporting it in group as `gej_ecmult_gen_`, e.g., using a `#define`? This would keep the identifier<->filename naming convention but adds a bit of complexity due to the re-export. And it makes ecmult a submodule of group (philosophically), and it I think this makes sense.

Sorry, all of this is bike shedding. But at least I'd like to hear more opinions here given that we're renaming these fundamental functions.

theStack:
> Now that I'm looking at this again, I start to see the drawback of this: These functions live in the ecmult module but now they have a ge_ or gej_ prefix which is usually reserved for the group module.

Ah that's a good point indeed!

> What about a) moving the ge_ecmult_gen helper to group and b) keeping ecmult_gen as is in ecmult but re-exporting it in group as gej_ecmult_gen_, e.g., using a #define? This would keep the identifier<->filename naming convention but adds a bit of complexity due to the re-export. And it makes ecmult a submodule of group (philosophically), and it I think this makes sense.

I've tried the a) part now and it seems that this would introduce a circular dependency between the group and ecmult_gen modules. Due to needing the `secp256k1_ecmult_gen_context` struct type, `group.h` would need to include `ecmult_gen.h`, which in turn includes `group.h` again. (Also, we would need to include `scalar.h` in `group.h` as well which wasn't needed before, but maybe that part would be fine?).

> Sorry, all of this is bike shedding. But at least I'd like to hear more opinions here given that we're renaming these fundamental functions.

No worries, I agree that this shouldn't be rushed. For #1765, the helper would be a nice-to-have for making the sending function a tiny bit smaller (-3 LOC for calculating [A = a * G](https://github.com/bitcoin/bips/blob/2ffcd9a4a1b834434f3c02718242235f9e270e94/bip-0352.mediawiki?plain=1#L302)) and thus potentially easier to review, but it's certainly not urgent. I'm changing the PR to draft state until we figure out where to best place the new helper and how to handle (re)naming.

real-or-random:
> I've tried the a) part now and it seems that this would introduce a circular dependency between the group and ecmult_gen modules

Okay okay, I don't think that naming issue should hold up the essence of this PR.

I'd say just call the current function `ecmult_gen_gej`—that rename is minor enough not to bother—and the wrapper `ecmult_gen_ge`. Then they can both live in ecmult. That is consistent with the current convention and clear enough.

theStack:
Thanks, switched back to `_ecmult_gen_{ge,gej}` :ok_hand:

real-or-random (approved):
utACK 9e017e5062bfe306eb5724fe39debe94b6223592

furszy (commented):
ACK 9e017e5062bfe306eb5724fe39debe94b6223592

PR #1863, opened 2026-06-06
ellswift: don't declassify or leave sk in sha256 buffer

furszy:
`secp256k1_ellswift_create` assumes sha256 clears the data in its buffer
after hashing it, which is not the case. So we shouldn't declassify the whole
struct, only the hash result. We should also clear it at the end, so the sk
doesn't linger on the stack when no aux rnd is given.

On master, can add the following diff and run the tests to see sk sitting
in the buffer.
```diff
diff --git a/src/modules/ellswift/main_impl.h b/src/modules/ellswift/main_impl.h
--- a/src/modules/ellswift/main_impl.h	
+++ b/src/modules/ellswift/main_impl.h	
@@ -461,6 +461,9 @@
     secp256k1_ellswift_elligatorswift_var(ctx, ell64, &t, &p, &hash); /* puts u in ell64[0..32] */
     secp256k1_fe_get_b32(ell64 + 32, &t); /* puts t in ell64[32..64] */
 
+    /* DEMO: fail because sk sits in the buffer */
+    if (!auxrnd32) VERIFY_CHECK(memcmp(hash.buf, seckey32, 32) != 0);
     secp256k1_memczero(ell64, 64, !ret);
```

### DIFF
--- a/src/modules/ellswift/main_impl.h
+++ b/src/modules/ellswift/main_impl.h
@@ -449,12 +449,13 @@ int secp256k1_ellswift_create(const secp256k1_context *ctx, unsigned char *ell64
     secp256k1_fe_normalize_var(&p.x);
     secp256k1_fe_normalize_var(&p.y);
 
-    /* Set up hasher state. The used RNG is H(privkey || "\x00"*32 [|| auxrnd32] || cnt++),
+    /* Set up hasher state. The used RNG is H(seckey32 || "\x00"*32 [|| auxrnd32] || cnt++),
      * using BIP340 tagged hash with tag "secp256k1_ellswift_create". */
     secp256k1_ellswift_sha256_init_create(&hash);
     secp256k1_sha256_write(secp256k1_get_hash_context(ctx), &hash, seckey32, 32);
     secp256k1_sha256_write(secp256k1_get_hash_context(ctx), &hash, zero32, sizeof(zero32));
-    secp256k1_declassify(ctx, &hash, sizeof(hash)); /* private key is hashed now */
+    /* Declassify only hash state. seckey32 has been hashed, but copy remains in the hash buffer */
+    secp256k1_declassify(ctx, &hash.s, sizeof(hash.s));
     if (auxrnd32) secp256k1_sha256_write(secp256k1_get_hash_context(ctx), &hash, auxrnd32, 32);
 
     /* Compute ElligatorSwift encoding and construct output. */
@@ -463,6 +464,7 @@ int secp256k1_ellswift_create(const secp256k1_context *ctx, unsigned char *ell64
 
     secp256k1_memczero(ell64, 64, !ret);
     secp256k1_scalar_clear(&seckey_scalar);
+    secp256k1_sha256_clear(&hash);
 
     return ret;
 }


theStack:
Concept ACK

Interesting find. I was very confused at first why there wouldn't be further writes on this passed hashing object within `secp256k1_ellswift_elligatorswift_var`, but it's just copied over to a new instance for each `secp256k1_ellswift_prng` call (which makse sense for reusing the state with different counters):
https://github.com/bitcoin-core/secp256k1/blob/13db747f2b558659b26f8ca3f5ea6e6d34e2054c/src/modules/ellswift/main_impl.h#L310-L311

Fwiw Bitcoin Core's BIP324 implementation is not affected by this, as we [pass in the randomness](https://github.com/bitcoin/bitcoin/blob/1aafd49077041e874d6b5f114927bc37e5b85d71/src/key.cpp#L320) (`auxrnd32` parameter), which IIUC overwrites the secret key in the `_sha256` object buffer.

src/modules/ellswift/main_impl.h
real-or-random: nit:
```suggestion
    /* Set up hasher state. The used RNG is H(seckey23 || "\x00"*32 [|| auxrnd32] || cnt++),
```
could be fixed on the side here

src/modules/ellswift/main_impl.h
```diff
@@ -454,7 +454,7 @@ int secp256k1_ellswift_create(const secp256k1_context *ctx, unsigned char *ell64
     secp256k1_ellswift_sha256_init_create(&hash);
     secp256k1_sha256_write(secp256k1_get_hash_context(ctx), &hash, seckey32, 32);
     secp256k1_sha256_write(secp256k1_get_hash_context(ctx), &hash, zero32, sizeof(zero32));
-    secp256k1_declassify(ctx, &hash, sizeof(hash)); /* private key is hashed now */
+    secp256k1_declassify(ctx, &hash.s, sizeof(hash.s)); /* The private key has been hashed, but sk remains in the buffer */
```
real-or-random: I think the comment is still confusing to me. Isn't `sk` the same as the "private key" (and actually called `seckey32`)?

real-or-random (commented):
Concept ACK

src/modules/ellswift/main_impl.h
furszy: Sure. Done as suggested.

src/modules/ellswift/main_impl.h
furszy: yeah, pushed an improvement.

real-or-random (approved):
utACK 0ae17e304e17a29c277b563139c4810ee7a96319

theStack (approved):
ACK 0ae17e304e17a29c277b563139c4810ee7a96319

Checked that
```diff
diff --git a/src/modules/ellswift/main_impl.h b/src/modules/ellswift/main_impl.h
index 817e7788..44bc6868 100644
--- a/src/modules/ellswift/main_impl.h
+++ b/src/modules/ellswift/main_impl.h
@@ -462,10 +462,16 @@ int secp256k1_ellswift_create(const secp256k1_context *ctx, unsigned char *ell64
     secp256k1_ellswift_elligatorswift_var(ctx, ell64, &t, &p, &hash); /* puts u in ell64[0..32] */
     secp256k1_fe_get_b32(ell64 + 32, &t); /* puts t in ell64[32..64] */
 
+    /* DEMO: succeed because sk sits in the buffer */
+    if (!auxrnd32) VERIFY_CHECK(memcmp(hash.buf, seckey32, 32) == 0);
+
     secp256k1_memczero(ell64, 64, !ret);
     secp256k1_scalar_clear(&seckey_scalar);
     secp256k1_sha256_clear(&hash);
 
+    /* DEMO: succeed because sk doesn't sit in buffer anymore */
+    if (!auxrnd32) VERIFY_CHECK(memcmp(hash.buf, seckey32, 32) != 0);
+
     return ret;
 }
```
succeeds in the tests and also ran the ctime-tests locally.

real-or-random:
Oh I should have waited for CI before merging this. Well, let's wait for it on master...

furszy:
> Oh I should have waited for CI before merging this. Well, let's wait for it on master...

Don't worry much. Only comments changed on the last push and CI was green before.

PR #1865, opened 2026-06-08
test: enable -Wunused-function in test suite (Fix #1831)

kallal79:
This PR addresses issue #1831 by enabling the `-Wunused-function` compiler warning within the test suite. 

Currently, `-Wno-unused-function` is passed globally to disable warnings about unused functions, making it too easy to write a test case but forget to actually call it. To catch untested helper functions safely, this PR uses GCC/Clang pragmas scoped strictly to the body of the test files.

### Changes Made:
- Added `#pragma GCC diagnostic warning "-Wunused-function"` directly after the `#include` statements in `src/tests.c`, `src/tests_exhaustive.c`, `src/ctime_tests.c`, and `src/unit_test.c`. 
Fixes #1831

### DIFF
--- a/src/ctime_tests.c
+++ b/src/ctime_tests.c
@@ -40,6 +40,11 @@
 #include "../include/secp256k1_ellswift.h"
 #endif
 
+#if defined(__GNUC__)
+# pragma GCC diagnostic push
+# pragma GCC diagnostic warning "-Wunused-function"
+#endif
+
 static void run_tests(secp256k1_context *ctx, unsigned char *key);
 
 int main(void) {
@@ -265,3 +270,7 @@ static void run_tests(secp256k1_context *ctx, unsigned char *key) {
 
 #endif
 }
+
+#if defined(__GNUC__)
+# pragma GCC diagnostic pop
+#endif
--- a/src/tests.c
+++ b/src/tests.c
@@ -37,6 +37,11 @@
 #include "int128_impl.h"
 #endif
 
+#if defined(__GNUC__)
+# pragma GCC diagnostic push
+# pragma GCC diagnostic warning "-Wunused-function"
+#endif
+
 #define CONDITIONAL_TEST(cnt, nam) if (COUNT < (cnt)) { printf("Skipping %s (iteration count too low)\n", nam); } else
 
 static secp256k1_context *CTX = NULL;
@@ -8091,3 +8096,7 @@ int main(int argc, char **argv) {
     if (tf_init(&tf, argc, argv) != 0) return EXIT_FAILURE;
     return tf_run(&tf);
 }
+
+#if defined(__GNUC__)
+# pragma GCC diagnostic pop
+#endif
--- a/src/tests_exhaustive.c
+++ b/src/tests_exhaustive.c
@@ -31,6 +31,11 @@
 #include "testutil.h"
 #include "util.h"
 
+#if defined(__GNUC__)
+# pragma GCC diagnostic push
+# pragma GCC diagnostic warning "-Wunused-function"
+#endif
+
 static int count = 2;
 
 static uint32_t num_cores = 1;
@@ -467,3 +472,7 @@ int main(int argc, char** argv) {
     printf("no problems found\n");
     return EXIT_SUCCESS;
 }
+
+#if defined(__GNUC__)
+# pragma GCC diagnostic pop
+#endif
--- a/src/unit_test.c
+++ b/src/unit_test.c
@@ -17,6 +17,11 @@
 #include "testrand.h"
 #include "tests_common.h"
 
+#if defined(__GNUC__)
+# pragma GCC diagnostic push
+# pragma GCC diagnostic warning "-Wunused-function"
+#endif
+
 #define UNUSED(x) (void)(x)
 
 /* Number of times certain tests will run */
@@ -477,3 +482,7 @@ static int tf_run(struct tf_framework* tf) {
 
     return status;
 }
+
+#if defined(__GNUC__)
+# pragma GCC diagnostic pop
+#endif


src/ctime_tests.c
```diff
@@ -40,6 +40,11 @@
 #include "../include/secp256k1_ellswift.h"
 #endif
 
+#if defined(__GNUC__)
+# pragma GCC diagnostic push
```
real-or-random, line 44: A `push` without a `pop` doesn't makes a lot of sense. I think the cleanest thing is to add a `pop` at the end of each test file.

real-or-random (commented):
Thanks for the PR. :)

Concept ACK

src/ctime_tests.c
real-or-random: Also our macro `SECP256K1_GNUC_PREREQ(4, 2)` would be preferable over `defined(__GNUC__)`. 

(This is maybe questionable when it comes to these old compiler versions but as long as we have the macro, we should use it.)

real-or-random (commented):
Thanks for the quick update.

ACK mod that this should be squashed

Could you squash the 2nd and 3rd commit into the 1st one and force-push?

We follow the same workflow as Bitcoin Core here, see https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#squashing-commits for background.

real-or-random:
@hebasto Want to review this?

hebasto (commented):
~From https://github.com/bitcoin-core/secp256k1/issues/1831#issue-4010176205:~
~> ... we should enable -Wunused-function at least in the test suite.~

~Could this be enabled as well?~

src/ctime_tests.c
hebasto: While the `-Wno-unused-function` diagnostics has been available since GCC 3.0, the support for `#pragma GCC diagnostic` was [introduced](https://gcc.gnu.org/onlinedocs/gcc-4.2.4/gcc/Diagnostic-Pragmas.html) only in GCC 4.2, and the `push`/`pop` variants in [GCC 4.6](https://gcc.gnu.org/gcc-4.6/changes.html). Since this PR uses `diagnostic push`, the guard should be `SECP256K1_GNUC_PREREQ(4, 6)`. Note that this check evaluates to false under Clang, which masquerades as GCC 4.2.x, so an additional `defined(__clang__)` check is required.

hebasto (commented):
Concept ACK.

An alternative approach would be appending `-Wunused-function` to the compiler flags when building test targets. However, that would require changing and then maintaining for both build systems.

From https://github.com/bitcoin-core/secp256k1/issues/1831#issue-4010176205:
>  It's too easy to add a test but forget to call it.

Tested this scenario. Works as expected.

real-or-random:
@kallal79 

> Could you squash the 2nd and 3rd commit into the 1st one and force-push?

Just checking: Did you see this request? (I saw that you gave a few thumbs-ups in later comment.)

hebasto:
> @kallal79 
> 
> 
> 
> > Could you squash the 2nd and 3rd commit into the 1st one and force-push?
> 
> 
> 
> Just checking: Did you see this request? (I saw that you gave a few thumbs-ups in later comment.)


And https://github.com/bitcoin-core/secp256k1/pull/1865#discussion_r3389257862?

src/ctime_tests.c
real-or-random: (Oh, I had marked this conversation resolved already, so I missed your comment @hebasto.)

You're right. I suggested the wrong version here. (I trusted ChatGPT 5.4 but apparently you can't even trust it for these simple things... -.-) I think we should really get rid of the `SECP256K1_GNUC_PREREQ` macro. 

@kallal79 
For now, I suggest to change this back to what you had proposed initially (`#if defined(__GNUC__)`) because this will work both for GCC and Clang. Sorry for the back and forth...

real-or-random:
@kallal79 Thanks for the update! I believe you introduced a mistake when squashing the commits. Now there's only a single commits beyond the desired changes, it also reverts some recent commits on master.

hebasto (commented):
Tested a77dacad9ad8a2bff51b51c6a000c7dae45dad78:
```
$ gcc --version
gcc (Debian 4.3.2-1.1) 4.3.2
Copyright (C) 2008 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.

$ gcc -O2 -c src/precomputed_*.c
$ gcc -O2 -o noverify_tests src/tests.c precomputed_ecmult.o precomputed_ecmult_gen.o
In file included from src/tests.c:29:
src/unit_test.c:21: warning: expected [error|warning|ignored] after '#pragma GCC diagnostic'
src/unit_test.c:487: warning: expected [error|warning|ignored] after '#pragma GCC diagnostic'
src/tests.c:41: warning: expected [error|warning|ignored] after '#pragma GCC diagnostic'
src/tests.c:8101: warning: expected [error|warning|ignored] after '#pragma GCC diagnostic'
src/util.h:34: warning: 'print_buf_plain' defined but not used
src/assumptions.h:27: warning: 'secp256k1_assumption_checker' defined but not used
src/hash_impl.h:251: warning: 'secp256k1_hmac_sha256_clear' defined but not used
src/testrand_impl.h:121: warning: 'testrand_flip' defined but not used
```

real-or-random:
ACK a77dacad9ad8a2bff51b51c6a000c7dae45dad78

> Tested [a77daca](https://github.com/bitcoin-core/secp256k1/commit/a77dacad9ad8a2bff51b51c6a000c7dae45dad78):
>

I took the liberty to reset this to this commit, removing https://github.com/bitcoin-core/secp256k1/pull/1865/commits/2999a607f2539dd182c6f50ac15a43847ec23659 and its revert https://github.com/bitcoin-core/secp256k1/pull/1865/commits/6f77415f101425f85be67b3ef57f81ad85eeb497.

I believe the test results with gcc 4.3.2 are entirely fine. Warnings are acceptable if you use such an old compiler. @hebasto Are you saying they're not fine?

hebasto:
> ACK [a77daca](https://github.com/bitcoin-core/secp256k1/commit/a77dacad9ad8a2bff51b51c6a000c7dae45dad78)
> 
> > Tested [a77daca](https://github.com/bitcoin-core/secp256k1/commit/a77dacad9ad8a2bff51b51c6a000c7dae45dad78):
> 
> I took the liberty to reset this to this commit, removing [2999a60](https://github.com/bitcoin-core/secp256k1/commit/2999a607f2539dd182c6f50ac15a43847ec23659) and its revert [6f77415](https://github.com/bitcoin-core/secp256k1/commit/6f77415f101425f85be67b3ef57f81ad85eeb497).
> 
> I believe the test results with gcc 4.3.2 are entirely fine. Warnings are acceptable if you use such an old compiler. @hebasto Are you saying they're not fine?

Currently, when building with GCC 4.2 through 4.5.x, the `-Wunused-function` diagnostic isn't limited to the designated sources (since `push`/`pop` wasn't introduced until 4.6). 

GCC versions older than 4.2 simply drop unrecognized `#pragma`s without failing:
```
$ gcc-4.1 -O2 src/tests.c src/precomputed_*.c -o noverify_tests
```

However, if explicitly enabled with `-Wunknown-pragmas`, they surface as expected:
```
$ gcc-4.1 -O2 -Wunknown-pragmas src/tests.c src/precomputed_*.c -o noverify_tests
In file included from src/tests.c:29:
src/unit_test.c:21: warning: ignoring #pragma GCC diagnostic
src/unit_test.c:22: warning: ignoring #pragma GCC diagnostic
src/unit_test.c:487: warning: ignoring #pragma GCC diagnostic
src/tests.c:41: warning: ignoring #pragma GCC diagnostic
src/tests.c:42: warning: ignoring #pragma GCC diagnostic
src/tests.c:8101: warning: ignoring #pragma GCC diagnostic
```

hebasto (approved):
ACK a77dacad9ad8a2bff51b51c6a000c7dae45dad78.

These changes are useful for our current development. Just note that a bit more care is needed when using GCC older than 4.6, but I think that's fine.

hebasto:
Apparently, the `push`/`pop` machinery [does not work as expected](https://github.com/bitcoin-core/secp256k1/actions/runs/27630881895/job/81705239375) when using GCC 4.7.2:
```
src/tests.c: At top level:
src/util.h:34:13: warning: 'print_buf_plain' defined but not used [-Wunused-function]
```

real-or-random:
Hm, and on GCC 5, it won't warn at all (which is also fine). It seems to work properly from GCC 6. Here's a godbolt link to quickly check this on different compilers. https://godbolt.org/z/co97Pbbe6

Anyway, I think this is still fine. If someone really wants to use GCC 4, then they need to accept that things may be not optimal.

hebasto:
> Hm, and on GCC 5, it won't warn at all (which is also fine). It seems to work properly from GCC 6.

Documenting GCC 10.2, shipped in Debian "oldoldstable", as the minimum recommended compiler should address this issue.

kallal79:
Thanks for the reviews and merge — I’ll keep the GCC diagnostic push/pop nuances in mind for future contributions.

PR #1866, opened 2026-06-09
musig: Add hash context tests

mllwchrry:
Adds missing hash context tests for the musig module, covering the`pubkey_agg`, `nonce_gen`, and `nonce_process` functions. Follows the same pattern as the hash context tests for other modules added in #1777.

### DIFF
--- a/src/modules/musig/tests_impl.h
+++ b/src/modules/musig/tests_impl.h
@@ -20,6 +20,7 @@
 #include "../../group.h"
 #include "../../hash.h"
 #include "../../util.h"
+#include "../../testutil.h"
 #include "../../unit_test.h"
 
 #include "vectors.h"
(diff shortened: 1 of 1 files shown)


real-or-random:
@furszy Want to review this?

src/modules/musig/tests_impl.h
```diff
@@ -608,6 +609,97 @@ static void sha256_tag_test(void) {
...
+    secp256k1_musig_keyagg_cache cache_default, cache_custom;
+    const unsigned char sk[32] = {1};
+    secp256k1_pubkey pk;
+    const secp256k1_pubkey *pks[1];
+
+    CHECK(secp256k1_ec_pubkey_create(ctx, &pk, sk));
+    pks[0] = &pk;
+
+    /* Default behavior. No ctx-provided SHA256 compression */
+    CHECK(secp256k1_musig_pubkey_agg(ctx, NULL, &cache_default, pks, 1));
```
theStack: nit, for a slightly more precise check, here and for other `CHECK` instances:
```suggestion
    CHECK(secp256k1_musig_pubkey_agg(ctx, NULL, &cache_default, pks, 1) == 1);
```
(see also PR #1481 which enforces that)

theStack (approved):
ACK e0b1f34ca0203905779ecf2b518bf9666c40c787

Note that there are three more musig API functions that use the hash context, so adding tests for them would make sense, either here or in a separate PR: `_musig_nonce_gen_counter` (could probably tested alongside of the existing `_musig_nonce_gen`) and `_musig_partial_{sign,sig_verify}`.

furszy (commented):
ACK e0b1f34ca0203905779ecf2b518bf9666c40c787

mllwchrry:
Thanks for the reviews! I applied a `== 1` check to other `CHECK`s and added tests for `nonce_gen_counter`, `partial_sign`, and `partial_sig_verify`.

src/modules/musig/tests_impl.h
```diff
@@ -608,6 +609,206 @@ static void sha256_tag_test(void) {
...
+    CHECK(!sha256_musig_partial_sig_verify_called);
+
+    /* Override SHA256 compression directly, bypassing the ctx setter sanity checks.
+     * The KeyAgg coefficient now differs, so verifying the same partial sig fails. */
+    ctx->hash_ctx.fn_sha256_compression = sha256_musig_partial_sig_verify;
+    CHECK(secp256k1_musig_partial_sig_verify(ctx, &partial_sig, &pubnonce, &pk, &keyagg_cache, &session) == 0);
+    CHECK(sha256_musig_partial_sig_verify_called);
+
+    secp256k1_context_destroy(ctx);
+}
```
furszy: These two tests can be merged, they share the same setup. You can just call the first sign function, then set the `called` flag to false and call the verify one.

And you could apply the same concept for the others as well. Merging `sha256_musig_nonce_process` within this shouldn't be hard either.

mllwchrry:
Thanks @furszy! That's a great idea. Consolidated all the flow tests into one `musig_ctx_sha256` that builds the session once, keeping `nonce_gen_counter` separate as it isn't part of the sign/verify flow.

theStack (approved):
ACK ab6925b8ebbced112e30bf009f5958c0173eebb1

nit: if you need/want to retouch, could still fold the `_nonce_gen_counter` test inside the large one (so the "check that ... override takes effect in every musig function" comment fully applies)

mllwchrry:
Retouched, `_nonce_gen_counter` is now part of the big test.

theStack (approved):
ACK fac66a1b4e4cdc65ad635cbdd1dc2acf31ece35b

PR #1867, opened 2026-06-09
test: musig: fix dead "aggnonce encodes two points at infinity" check

theStack:
Due to the missing `CHECK` around, the return values were discarded and nothing was actually checked here.

(Fwiw I prompted two AI models (MiniMax M3 and Opus 4.8) to find more similar instances in tests with bare statements that miss a surrounding `CHECK` in tests, and both didn't find any.)


real-or-random (approved):
utACK d7125e517d45507df4a3f19c8ca90393a8290480

Maybe we should use `SECP256K1_WARN_UNUSED_RESULT` also internally. Currently we only have this in public API functions.

hebasto (approved):
ACK d7125e517d45507df4a3f19c8ca90393a8290480, I have reviewed the code and it looks OK.

Issue #1387, opened 2023-07-27
ci: Add a "compile the headers" job


real-or-random:
We have a C++ job that checks that all headers compile on their own, but of course, we should test this on C also...

The example binaries cover some headers, but not by all (e.g., there's no example for ellswift or prealloc).

purpleKarrot:
There is a common guideline, that for any header file, there should be an *associated* source file, which `#include`s that header, even if the source file otherwise is empty.

This not only make sure that header files are self contained, but also static code analysis tools like clang-tidy or iwyu only report warnings from the current source file and the *associated* header. For those tools, an *associated* source file needs to exist, or diagnostics in headers are silently ignored.

For a file `stem.h`, both `stem.c` and `stem_test.c` are considered *associated* source files. Since some compilers dislike source files that don't provide any symbols, the easiest approach is to use `stem_test.c` and actually put a test in it.

real-or-random:
This issue is mean to cover the *public* headers.


> This not only make sure that header files are self contained,

For self-contained public headers, see https://github.com/bitcoin-core/secp256k1/issues/1039.

purpleKarrot:
@real-or-random I created a test to compile the public headers in https://github.com/bitcoin-core/secp256k1/pull/1693.
Build jobs that run tests with `ctest` will cover this. A dedicated "compile the public headers" job might not be needed.

PR #1870, opened 2026-06-12
ci: Add a compile the public headers job (#1387)

kallal79:
This PR ensures that all public C headers in `secp256k1` are strictly self-contained and independently compilable without requiring downstream projects to manually include missing dependencies (like `<stddef.h>` or `<stdint.h>`).

### Changes included:
1. **GitHub Actions (`ci.yml`)**: Added a dedicated `c_headers_debian` job that runs `gcc` and `clang` in `-std=c89` and `-pedantic-errors` modes to precompile every public header, mirroring the existing C++ checks. Also updated the Windows MSVC job to test headers using `-TC` (C-mode).
2. **Offline Tests (`test_public_headers.c`)**: Created a dedicated C test file that explicitly `#include`s every public header conditionally based on enabled modules (e.g., `ENABLE_MODULE_ECDH`). This test is fully integrated into both `CMakeLists.txt` and `Makefile.am` so it will automatically be verified during local `make check` and `ctest` runs.


real-or-random:
- CI fails here, see for example https://github.com/bitcoin-core/secp256k1/actions/runs/27440362791/job/81112805983?pr=1870#step:3:1600
 - I don't see what the purpose of the `src/test_public_headers.c` (and its CMake integration) is. Is the goal to make sure that the headers compile even when they're all included in a single compilation unit?

PR #1874, opened 2026-06-17
Minimum recommended toolchain - document and test it

hebasto:
This PR:

1. Establishes clear boundaries for the project's recommended build and testing environments.

2. Adds a CI job that tests the project's minimum recommended build environment.

This is useful for reviewing changes like https://github.com/bitcoin-core/secp256k1/pull/1865 or https://github.com/bitcoin-core/secp256k1/pull/1869.

---

For the context:

- from https://github.com/bitcoin-core/secp256k1/pull/1871#issuecomment-4729692574:
  > ... Wheezy is just ancient. Do you think a Docker image that depends on a generic Debian oldoldstable image could do the job? (Does such an image exist?) This would at least save us from bumping the version regularly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,6 @@ Implementation details
   * No runtime heap allocation.
   * Extensive testing infrastructure.
   * Structured to facilitate review and analysis.
-  * Intended to be portable to any system with a C89 compiler and uint64_t support.
   * No use of floating types.
   * Expose only higher level interfaces to minimize the API surface and improve application security. ("Be difficult to use insecurely.")
 * Field operations
@@ -97,8 +96,15 @@ This can be done with the following steps:
         Subkey fingerprint: 4BBB 845A 6F5A 65A6 9DFA  EC23 4861 DBF2 6212 3605
    ```
 
-Building with Autotools
------------------------
+## Building from source
+
+The project is intended to be portable to any system with a C89/C90-compatible toolchain.
+Additionally, the C standard library must provide the `<stdint.h>` header.
+
+Toolchains older than the default one shipped in the current Debian "oldoldstable" release
+are not tested and are not recommended for building the project.
+
+### Building with Autotools
 
     $ ./autogen.sh       # Generate a ./configure script
     $ ./configure        # Generate a build system
@@ -108,12 +114,11 @@ Building with Autotools
 
 To compile optional modules (such as Schnorr signatures), you need to run `./configure` with additional flags (such as `--enable-module-schnorrsig`). Run `./configure --help` to see the full list of available flags.
 
-Building with CMake
--------------------
+### Building with CMake
 
 To maintain a pristine source tree, CMake encourages to perform an out-of-source build by using a separate dedicated build tree.
 
-### Building on POSIX systems
+#### Building on POSIX systems
 
     $ cmake -B build              # Generate a build system in subdirectory "build"
     $ cmake --build build         # Run the actual build process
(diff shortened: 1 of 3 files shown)


hebasto:

README.md
```diff
@@ -97,8 +96,15 @@ This can be done with the following steps:
...
 
-Building with Autotools
------------------------
+## Building from source
+
+The project is intended to be portable to any system with a C89/C90-compatible toolchain.
+Additionally, the C standard library must provide the `<stdint.h>` header.
+
+Toolchains older than the default one shipped in the current Debian "oldoldstable" release
+are not tested and are not recommended for building the project.
```
real-or-random, line 105: I think we should remove this sentence (and change the PR title accordingly). 
- I don't think we should recommend a mininum compiler. People should use modern compilers, but this is nothing we need to state. What compiler we test with is an internal thing, and this should not become a "user-facing" guarantee (The wording does not imply a guarantee, but if it's not guarantee, it doesn't belong in the README. We could add it to CONTRIBUTING but I don't think it's necessary. The CI check is enough documentation.)
- That older than oldoldstable is not recommended could be interpreted to mean that oldoldstable and newer is recommend. But I wouldn't recommend oldoldstable to anyone.

README.md
```diff
@@ -97,8 +96,15 @@ This can be done with the following steps:
         Subkey fingerprint: 4BBB 845A 6F5A 65A6 9DFA  EC23 4861 DBF2 6212 3605
    ```
 
-Building with Autotools
------------------------
+## Building from source
```
real-or-random, line 99: nit: 
```suggestion
## Building
```
Maybe it's just me, but "building from source" sounds like this is a special thing and we also ship binaries.

ci/linux-debian-oldold.Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+    git \
+    autoconf automake libtool make \
+    gcc \
+    python3-full && \
+    apt-get clean && rm -rf /var/lib/apt/lists/*
+
+ENV VIRTUAL_ENV=/root/venv
+RUN python3 -m venv $VIRTUAL_ENV
+ENV PATH="$VIRTUAL_ENV/bin:$PATH"
+RUN pip install lief
```
real-or-random, line 17: I think we should omit python from this job to simplify maintenance. We don't want to be restricted to whatever Python version oldoldstable ships (and who knows if lief will runs on that version in the future).

real-or-random (commented):
Weak Concept ACK (if we remove the "recommendation part") -- I think it's a useful test and the maintenance burden will be okay

@fanquake I wonder what you think from with your maintainer hat on.

real-or-random:
@hebasto Have you checked if the new Docker image fits in the GHA cache (together with the existing cache artifacts)?

ci/linux-debian-oldold.Dockerfile
hebasto: Disabling Python would require more changes in this PR to handle the `testvectors` build target, similar to 071523b2ceec9d10d53915ebbe942ccaf194d086.

README.md
real-or-random: see my "code" suggestion: Just "Building"

ci/linux-debian-oldold.Dockerfile
real-or-random: Hm, yeah. I was thinking we can just set `SYMBOL_CHECK=no`, but I forgot about the test vectors. Though I can we could also have `TEST_VECTORS` variable that is enabled by default. That would just change a single line in the script.

README.md
fanquake: I agree. I'm not sure there's much value (for this project) in trying to establish a minimum supported compiler version, and it seems possible that tests for this are going to add more friction than value. As long as they are low-effort, and can be deleted/changed easily, probably fine.

I'd also say that if we were going to document something, just state the particular compiler, rather than make people look up what compiler/version pair happened to ship on an old version of some Linux distro.

real-or-random: > I'd also say that if we were going to document something, just state the particular compiler, rather than make people look up what compiler/version pair happened to ship on an old version of some Linux distro.

No strong opinion here but I still I tend to simply say "oldoldstable". Then we won't run into the need to bump the version regularly (e.g., because something breaks on an old compiler/docker image) and worse, use our brain cycles to make a decision whether the bump is justified without clear guidelines.

Issue #1610, opened 2024-09-30
Add extra API for offline/online phases in ECDSA signing


RandomLattice:
We'd love to split the ECDSA signing operation into two steps:
- an "offline" step that is independent on the message,
- an "online" step that depends on the message.

The advantage here is that the offline step can be precomputed arbitrarily early and do "most of the computational work". This is very useful for signers that have a lot of idle time before the signing request comes in, but little time to compute and return the signature. For example, hardware wallets.

In ECDSA, the offline step can essentially compute nonce generation + scalar multiplication + nonce modular inversion. These values ("precomputed material") are input to the online part. The online part just finishes the ECDSA computation. The online part is extremely fast (we're seeing around 5000x faster than the offline phase in a non-libsecp256k1 proof-of-concept).

The precomputed material is secret, can only be used once and needs to be wiped after usage.

API-wise this would probably mean adding two functions: 
 * `secp256k1_ecdsa_sign_split_phase_precompute()`
 * `secp256k1_ecdsa_sign_split_phase_online()`

The composition of the two functions compute `secp256k1_ecdsa_sign()`. The input/output behavior does not change. 

Is there any appetite for this? Happy to write the code for this.

real-or-random:
> The precomputed material is secret, can only be used once and needs to be wiped after usage.

Indeed, and this is the drawback of this. It's a major footgun, and I'm sure this is the reason why it's currently implemented this way.

Personally, I'm a bit on the fence. We're about to add a [MuSig2 module](https://github.com/bitcoin-core/secp256k1/pull/1479), which is also prone to nonce-reuse failures -- but in that case, it is unavoidable to split nonce generation and the actual signing into two separate steps. In the end, I don't think I'm against giving users advanced and dangerous features, but they should be clearly marked as such.

I'd love to hear your use case and scenario here. Noone has asked for this before, and I doubt many users will benefit. (And anyway, ECDSA can be considered somewhat deprecated in Bitcoin, but okay, the same could be done for Schnorr signing). Do you personally need/want this, or is this is just a suggestion for improvement?

It would be interesting to know what other contributors and maintainers think.

jmecom:
> @jmecom Sorry for the slight tangent here, but:
> 
> > On our chip, ECDSA takes ~60ms and Schnorr takes ~55ms. We can't use large precomputation tables due to limited RAM.
> 
[...]

PR #1876, opened 2026-06-18
Add SECP256K1_WARN_UNUSED_RESULT to all relevant static int functions

### LINKED ISSUE #1872: Use `SECP256K1_WARN_UNUSED_RESULT` internally
Maybe we should use `SECP256K1_WARN_UNUSED_RESULT` also internally. Currently we only have this in public API functions.

_Originally posted by @real-or-random in https://github.com/bitcoin-core/secp256k1/pull/1867#pullrequestreview-4466893763_

shivv23:
Annotate all static int functions in the library (non-test code) with `SECP256K1_WARN_UNUSED_RESULT` so that compilers can warn when return values are accidentally discarded. This covers ~140 functions across all modules.

Changes:
- Add `SECP256K1_WARN_UNUSED_RESULT` to field, scalar, group, eckey, ecmult, ecdsa, int128 utility functions and all module header/impl files
- Fix call sites where return values were being discarded:
  - `secp256k1_ec_pubkey_combine`: check `pubkey_load` return
  - `secp256k1_ecdh`: check `pubkey_load` return
  - `secp256k1_ellswift_swiftec_var`, `ellswift_xdh`: use assignment pattern
  - `bench_internal.c`, `bench_ecmult.c`, test files: wrap in `CHECK()`
- Exclude `scalar_add` and `scalar_cond_negate` whose returns are intentionally informational/redundant at all existing call sites

See #1610 for prior discussion.

### DIFF
--- a/src/bench_ecmult.c
+++ b/src/bench_ecmult.c
@@ -135,7 +135,7 @@ static void bench_ecmult_const_xonly(void* arg, int iters) {
         const secp256k1_ge* pubkey = &data->pubkeys[(data->offset1+i) % POINTS];
         const secp256k1_scalar* scalar = &data->scalars[(data->offset2+i) % POINTS];
         int known_on_curve = 1;
-        secp256k1_ecmult_const_xonly(&data->output_xonly[i], &pubkey->x, NULL, scalar, known_on_curve);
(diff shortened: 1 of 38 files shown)


gatleas17 (commented):
Ñla

PR #1880, opened 2026-06-24
test: avoid scratch free warning

l0rinc:
**Problem:** latest GCC 17 snapshot warns in `run_scratch_tests` that a zeroed stack scratch object may be passed to `free()`.
**Fix:** Use the existing heap-allocated scratch object for the invalid scratch-space checks, zero it before the checks, and free it manually afterwards.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -372,7 +372,6 @@ static void run_scratch_tests(void) {
     size_t checkpoint;
     size_t checkpoint_2;
     secp256k1_scratch_space *scratch;
-    secp256k1_scratch_space local_scratch;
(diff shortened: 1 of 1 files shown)


theStack:
One possible alternative for keeping the current test behavior might be to just disable the warning for the problematic code line for GCC 17+: https://github.com/theStack/secp256k1/commit/aa0dbc40dbb1462e5c9978e8fc2e1c99433e6410 (trivial to review, but somewhat ugly admittedly)

real-or-random:
See my suggestion in #1881

PR #1881, opened 2026-06-25
tests: Fix GCC 17 snapshot warning

real-or-random:
Passing a non-malloc pointer to free() would be UB. In this case, the
free() line is never actually reached (and GCC 17 fails to prove this)
in a correct implementation of secp256k1_scratch_space_destroy(), but
the test shouldn't rely on the correctness of the tested function.

Alternative to #1880. I think this is cleaner 
 - it doesn't recreate the scratch space in the middle of some other tests
 - it has an obvious matching (malloc, free) pair 
 - it additionally checks that only `magic` is accessed

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -372,7 +372,6 @@ static void run_scratch_tests(void) {
     size_t checkpoint;
     size_t checkpoint_2;
     secp256k1_scratch_space *scratch;
-    secp256k1_scratch_space local_scratch;
 
     /* Test public API */
     scratch = secp256k1_scratch_space_create(CTX, 1000);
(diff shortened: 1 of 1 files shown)


theStack:
ACK 715e954abe77ece9f16955f5ae83191896582aef

(I wonder if strictly speaking the current code is really UB though? The problematic `free` call is never reached, as `_scratch_destroy` with the non-malloc pointer returns early due to invalid magic)

src/tests.c
```diff
@@ -412,16 +411,7 @@ static void run_scratch_tests(void) {
...
-    /* try to use badly initialized scratch space */
-    secp256k1_scratch_space_destroy(CTX, scratch);
-    memset(&local_scratch, 0, sizeof(local_scratch));
-    scratch = &local_scratch;
-    CHECK_ERROR(CTX, secp256k1_scratch_max_allocation(&CTX->error_callback, scratch, 0));
-    CHECK_ERROR(CTX, secp256k1_scratch_alloc(&CTX->error_callback, scratch, 500));
-    CHECK_ERROR_VOID(CTX, secp256k1_scratch_space_destroy(CTX, scratch));
-
     /* Test that large integers do not wrap around in a bad way */
-    scratch = secp256k1_scratch_space_create(CTX, 1000);
```
theStack, line 424: was this line removed intentionally?

real-or-random: Yes, we don't need to recreate the scratch because the call `secp256k1_scratch_space_destroy()` in line 416 is also removed.

real-or-random:
> (I wonder if strictly speaking the current code is really UB though? The problematic `free` call is never reached, as `_scratch_destroy` with the non-malloc pointer returns early due to invalid magic)


Ok, sure, you're right. I missed this.

But this argument also only holds if the tested code is correct, and the test shouldn't assume this. (Of course, the test may assume that the tested code doesn't invoke UB, because if it does, there's nothing that the can meaningfully do. But at least the test itself shouldn't introduce additional UB.)

real-or-random:
> > (I wonder if strictly speaking the current code is really UB though? The problematic `free` call is never reached, as `_scratch_destroy` with the non-malloc pointer returns early due to invalid magic)
> 
> Ok, sure, you're right. I missed this.

I force-pushed to update the commit message to reflect this. I also changed the PR description accordingly.

real-or-random:
And force-pushed again to rebase on master (to avoid the macOS failure in CI).

hebasto (approved):
ACK 9d75769dec4fcb40ee7e41f674f22f3f5e5cfc7e, I have reviewed the code and it looks OK.

theStack (approved):
re-ACK 9d75769dec4fcb40ee7e41f674f22f3f5e5cfc7e

PR #1882, opened 2026-06-26
scalar: correct `_scalar_get_bits_{limb32,var}` input condition docs

theStack:
This PR is a small correction of documented (off-by-one) input conditions for the `scalar_get_bits_{limb32,var}`, that came up during reviewing #1845.


real-or-random (approved):
ACK 6a599a44289597c2a3061b5ac6afddbf5c3ff293

PR #1895, opened 2026-07-22
tests: Cover all rejection paths of ec_privkey_import_der()

cavdarahmet:
`ec_privkey_import_der()` in `contrib/lax_der_privatekey_parsing.c` rejects
malformed encodings through eight structural checks. None of them were
reachable from the test suite: every `return 0` in the function was
uncovered, and only the successful round-trip path was exercised.

Measured with `gcc --coverage` + `gcov -b` on `./tests 2`:

| | lines | branches (taken at least once) |
|---|---|---|
| master | 82.69% | 57.14% |
| this PR | **100%** | **100%** |

What the test does:

* Mutates every field of an encoding produced by `ec_privkey_export_der()`,
  for both the compressed (1-byte sequence length) and the uncompressed
  (2-byte) layout, and asserts the layout it mutates before mutating it.
* Adds malformed encodings that are *not* truncations of a valid key. A
  truncated valid encoding is always caught by the sequence length check, so
  hand-crafted short encodings are the only way to reach the length checks
  behind it. This includes an indefinite-length sequence (`30 80 ...`) that is
  otherwise well-formed, and a sequence length that exceeds the input.
* Checks that `out32` is cleared whenever no key is extracted.
* Checks the lax behaviour the parser exists for: a key shorter than 32 bytes
  is accepted and left-padded with zeroes.
* Passes the input in a heap buffer of exactly the advertised length, so a
  read past the end is caught by ASan instead of silently succeeding on
  adjacent stack bytes. Four of the bounds checks are only detectable this way.

I sanity-checked that the test is load-bearing by removing each check in the
parser one at a time: 14 of 16 mutations are detected. The two that are not
are the empty-input check (ASan does not instrument a zero-size allocation)
and the `if (privkey[1])` guard added in #879 (a zero-length `memcpy` from an
out-of-bounds pointer is UB but invisible to every sanitizer).

While writing this I also confirmed that the existing bounds checks form
pointers such as `privkey + len` outside the caller's object before comparing
them, which is UB in a strict reading of the standard and is not caught by
ASan/UBSan (including `-fsanitize=pointer-overflow`). That is a separate
concern and is addressed by @l0rinc; the `{0x30, 0x82, 0xff, 0xff}` vector
here is a regression test for it. This PR does not change the parser.

I am aware of #781. If the lax DER helpers are eventually removed these tests
go with them; until then `contrib/` is inside the coverage target described in
CONTRIBUTING.md, and this file had the largest gap in it.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -7382,6 +7382,169 @@ static void run_ecdsa_der_parse(void) {
     }
 }
 
+/* Runs ec_privkey_import_der on a heap copy of the input of exactly privkeylen
+ * bytes, so that a read past the end of the input is caught by ASan or
+ * valgrind instead of silently succeeding on adjacent stack bytes. */
+static int privkey_import_der_exact(unsigned char *out32, const unsigned char *privkey, size_t privkeylen) {
+    /* Allocating zero bytes is not portable, so ask for one byte in that case. */
+    unsigned char *copy = checked_malloc(&CTX->error_callback, privkeylen + (privkeylen == 0));
+    int ret;
+    memcpy(copy, privkey, privkeylen);
+    ret = ec_privkey_import_der(CTX, out32, copy, privkeylen);
+    free(copy);
+    return ret;
+}
+
+/* ec_privkey_import_der accepts a BER SEQUENCE whose length is encoded in long
+ * form, and rejects everything else:
+ *
+ *   30 (80|lenb) <lenb length bytes> 02 01 01 04 <keylen> <key bytes> ...
+ *
+ * ec_privkey_export_der emits lenb = 1 for compressed and lenb = 2 for
+ * uncompressed keys, so the offsets behind the length shift accordingly. This
+ * mutates every field of an exported key and checks that it is rejected. */
+static void test_privkey_import_der_mutations(int compressed) {
+    unsigned char seckey[32];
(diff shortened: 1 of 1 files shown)


cavdarahmet:
Rebased on current master (post-0.8.0) and reworked the tests since the original push: both the compressed and the uncompressed layout are now mutated, rejected inputs are checked to clear the output, and every input is passed in a heap buffer of the exact advertised length so ASan can catch reads past the end. The file is now at 100% line and 100% branch coverage (`gcov -b`), see the updated description for details.

This is test-only and touches no library code, so it is independent of the 0.8.0 release work. @real-or-random you merged #879, the last substantive change to this file, and mentioned keeping it in sync with Bitcoin Core's copy — would you mind taking a look? cc @theStack

cavdarahmet:
Rebased on master. #1924 landed in the same two spots in `src/tests.c` — the block after `run_ecdsa_der_parse()` and the `tests_ecdsa[]` entry — so the merge conflicted. The resolution keeps both test blocks in registration order (`ecdsa_der_parse_long_form`, then `privkey_import_der`); the tests themselves are unchanged and the diff is still the same 164 added lines. Re-ran `tests 64` and an ASan/UBSan build, both clean.

PR #1897, opened 2026-07-26
tests: check results before using outputs

l0rinc:
**Problem:** Several tests inspect outputs from deterministic operations without checking the result.
The later output comparisons usually still exercise behavior, but a failed setup call can leave an output that accidentally satisfies a later assertion.
In `pubnonce_summing_to_inf`, `secp256k1_musig_sum_pubnonces` initializes both sums to infinity before a pubnonce load can fail, so the test could pass without constructing a valid opposing-nonce pair.

**Fix:** Check the result of each deterministic operation before reading its output in the affected exhaustive, EllSwift, recovery, extrakeys, MuSig, silent-payments, and public-key sort tests.
Leave deliberately ignored results alone when the test permits either result or validates a mutated value rather than an auxiliary carry, overflow, or sign flag.

### DIFF
--- a/src/modules/ellswift/tests_impl.h
+++ b/src/modules/ellswift/tests_impl.h
@@ -249,9 +249,9 @@ void ellswift_encode_decode_roundtrip_tests(void) {
         secp256k1_pubkey_save(&pubkey, &g);
         testrand256(rnd32);
         /* Convert the public key to ElligatorSwift and back. */
-        secp256k1_ellswift_encode(CTX, ell64, &pubkey, rnd32);
-        secp256k1_ellswift_decode(CTX, &pubkey2, ell64);
-        secp256k1_pubkey_load(CTX, &g2, &pubkey2);
+        CHECK(secp256k1_ellswift_encode(CTX, ell64, &pubkey, rnd32) == 1);
+        CHECK(secp256k1_ellswift_decode(CTX, &pubkey2, ell64) == 1);
+        CHECK(secp256k1_pubkey_load(CTX, &g2, &pubkey2) == 1);
         /* Compare with original. */
         CHECK(secp256k1_ge_eq_var(&g, &g2));
     }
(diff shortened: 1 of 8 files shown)


real-or-random (approved):
Weak Concept ACK -- I think sometimes there's also something to having the CHECKs only in the "primary" targets of a test case because this makes the code self-documenting and thus more readable. But when it comes to semantics, CHECKing more can't hurt and the earlier an abort, the easier it is to identify the cause of a failing test.

utACK a2ad68cd81787ba70b2caff0a57998ae8a5b4834

src/modules/ellswift/tests_impl.h
```diff
@@ -300,15 +300,15 @@ void ellswift_compute_shared_secret_tests(void) {
...
+        CHECK(secp256k1_ellswift_decode(CTX, &pub, ell64) == 1);
+        CHECK(secp256k1_pubkey_load(CTX, &dec, &pub) == 1);
         secp256k1_gej_set_ge(&decj, &dec);
         /* Compute the X coordinate of seckey*pubkey using ellswift_xdh. Note that we
          * pass ell64 as claimed (but incorrect) encoding for sec32 here; this works
          * because the "hasher" function we use here ignores the ell64 arguments. */
         ret = secp256k1_ellswift_xdh(CTX, share32, ell64, ell64, sec32, i & 1, &ellswift_xdh_hash_x32, NULL);
         CHECK(ret);
-        (void)secp256k1_fe_set_b32_limit(&share_x, share32); /* no overflow is possible */
+        CHECK(secp256k1_fe_set_b32_limit(&share_x, share32)); /* no overflow is possible */
```
real-or-random, line 311: If you touch this again:

```suggestion
        CHECK(secp256k1_fe_set_b32_limit(&share_x, share32) == 1); /* no overflow is possible */
```

If this is done everywhere, we should also do it here.

151henry151:
Tested ACK a2ad68cd81787ba70b2caff0a57998ae8a5b4834

Walked all seven commits. Musig and exhaustive ecmult_multi_var are real initialized-infinity false-passes; the other CHECKs are expected-success setup (failure would not satisfy the later asserts the same way). No obvious sibling miss in the touched files.

Tested with:
`./configure --enable-experimental --enable-module-ellswift --enable-module-extrakeys --enable-module-recovery --enable-module-schnorrsig --enable-module-musig --enable-module-silentpayments`
`make -j$(nproc) check` — passed

theStack (approved):
ACK a2ad68cd81787ba70b2caff0a57998ae8a5b4834

PR #1898, opened 2026-07-27
tests: cover Silent Payments label scanning

l0rinc:
**Problem:** The constant-time test generates an output using one scan key, then mutates the key before scanning, so the scan cannot find it.
It also passes no label callback, leaving label-candidate serialization and batching unexecuted.
The test still passes because it only covers the initial shared-secret computation with the undefined scan key.

**Fix:** Regenerate the output for the actual scan key and prepend a nonmatching output.
Use a callback to check each serialized 33-byte candidate is defined and count misses, and require one direct, unlabeled output match.


theStack:
Concept ACK

PR #1899, opened 2026-07-28
changelog: add missing entries for #1777 and #1860

theStack:
Adds changelog entries for merged PRs that still have the "needs-changelog" label, preparing for the upcoming release (scheduled for next week, August 3rd):
- #1777: "Make SHA256 compression runtime pluggable"
- #1860: "cmake: Emulate Libtool's behavior on NetBSD and OpenBSD"

The release note for #1860 is based on an earlier note on FreeBSD:
https://github.com/bitcoin-core/secp256k1/blob/0f6baf319fcae0d7f11a44fc9b4d4899b3f8082a/CHANGELOG.md?plain=1#L39

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,10 +8,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 #### Added
+ - New function `secp256k1_context_set_sha256_compression` for overriding the internal SHA256 compression function used by the library at runtime (e.g., to route SHA256 through a hardware-accelerated implementation).
  - New module `silentpayments` implements sending and receiving of Silent Payments according to [BIP 352](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki). See:
    - Header file `include/secp256k1_silentpayments.h` which defines the new API.
    - Usage example `examples/silentpayments.c`.
  - The `silentpayments` API currently requires full access to the transaction data (light client scanning is not implemented).
+
+#### Changed
+ - CMake: Shared libraries built with CMake on OpenBSD and NetBSD now create the full versioned filename (e.g. `libsecp256k1.so.6.2` instead of `libsecp256k1.so.6`) and symlink chain, matching the behavior of GNU Autotools builds.
 
 ## [0.7.1] - 2026-01-26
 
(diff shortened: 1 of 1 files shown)


real-or-random (approved):
ACK b77f728a61b9ed9fb86eb668259074b4f15095da

real-or-random:
cc @furszy @hebasto

hebasto (approved):
ACK b77f728a61b9ed9fb86eb668259074b4f15095da.

CHANGELOG.md
```diff
@@ -8,11 +8,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
 #### Added
+ - New function `secp256k1_context_set_sha256_compression` for overriding the internal SHA256 compression function used by the library at runtime (e.g., to route SHA256 through a hardware-accelerated implementation).
  - New module `silentpayments` implements sending and receiving of Silent Payments according to [BIP 352](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki). See:
    - Header file `include/secp256k1_silentpayments.h` which defines the new API.
    - Usage example `examples/silentpayments.c`.
  - The `silentpayments` API currently requires full access to the transaction data (light client scanning is not implemented).
 
+#### Changed
+ - CMake: Shared libraries built with CMake on OpenBSD and NetBSD now create the full versioned filename and symlink chain, matching the behavior of autotools builds.
```
hebasto: nit: I understand that this note mimics the previous one for FreeBSD. However, I'd prefer the tool's proper name here::
```suggestion
 - CMake: Shared libraries built with CMake on OpenBSD and NetBSD now create the full versioned filename and symlink chain, matching the behavior of GNU Autotools builds.
```

real-or-random: Indeed. We could also change it in the old entry (the changelog is not written in stone).

real-or-random: Would it make sense to give (for each BSD) an example file name of the shared library? That will make it much more tangible.

hebasto: It's `libsecp256k1.so.6.2` on OpenBSD and NetBSD.

For more details, see https://github.com/hebasto/secp256k1/actions/runs/27205923502.

real-or-random: And the example for FreeBSD (in 0.7.0 entry) could be `libsecp256k1.so.5.0.1`, see https://github.com/bitcoin-core/secp256k1/pull/1685.

hebasto: Here is the updated CI: https://github.com/hebasto/secp256k1/actions/runs/30361632628.

On FreeBSD, the lib's name's `libsecp256k1.so.6.0.2`.

theStack: Thanks for the suggestions, I agree with all of them (proper tool naming, providing a before/after .so filename example each, and applying the same retroactively for the 0.7.0 entry) and changed accordingly.

furszy (commented):
ACK f52eb393c4b58a229c56a0bc6fa1aed510d83348

real-or-random (approved):
ACK f52eb393c4b58a229c56a0bc6fa1aed510d83348

hebasto (approved):
ACK f52eb393c4b58a229c56a0bc6fa1aed510d83348.

PR #1900, opened 2026-07-29
Use SHA256 override for pointers to known aux functions

real-or-random:
Resolves one item in #1835.

### DIFF
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -417,7 +417,7 @@ typedef void (*secp256k1_sha256_compression_function)(
 /**
  * Set a callback function to override the internal SHA256 compression function.
  *
- * This installs a function to replace the built-in block-compression
+ * This installs a callback to replace the built-in block-compression
  * step used by the library's internal SHA256 implementation.
  * The provided callback must exactly implement the effect of n_blocks
  * repeated applications of the SHA256 compression function.
@@ -426,6 +426,14 @@ typedef void (*secp256k1_sha256_compression_function)(
  * SHA256 compression step through a hardware-accelerated or otherwise
  * specialized implementation. It is NOT meant for replacing SHA256
  * with a different hash function.
+ *
+ * Since auxiliary functions exposed by the library via a function
+ * pointer such as secp256k1_nonce_function_default do not take a
+ * context object, they will not use the callback when called directly
+ * from user code. (But they will use the callback when called from
+ * other library functions that do take a context object, e.g., when
+ * noncefp==NULL or noncefp==secp256k1_nonce_function_default is passed
+ * as an argument to secp256k1_ecdsa_sign.)
  *
  * Args:    ctx:             pointer to a context object.
  * In:      fn_compression:  pointer to a function implementing the compression function;
--- a/src/modules/ecdh/main_impl.h
+++ b/src/modules/ecdh/main_impl.h
@@ -60,7 +60,7 @@ int secp256k1_ecdh(const secp256k1_context* ctx, unsigned char *output, const se
     secp256k1_fe_get_b32(x, &pt.x);
     secp256k1_fe_get_b32(y, &pt.y);
 
-    if (hashfp == NULL) {
+    if (hashfp == NULL || hashfp == secp256k1_ecdh_hash_function_sha256) {
         /* Use ctx-aware function by default */
         ret = ecdh_hash_function_sha256_impl(secp256k1_get_hash_context(ctx), output, x, y, data);
     } else {
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -559,7 +559,7 @@ static int secp256k1_ecdsa_sign_inner(const secp256k1_context* ctx, secp256k1_sc
     while (1) {
         int is_nonce_valid;
 
-        if (noncefp == NULL) {
+        if (noncefp == NULL || noncefp == secp256k1_nonce_function_rfc6979) {
             /* Use ctx-aware function by default */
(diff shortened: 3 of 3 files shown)


real-or-random:
cc @furszy

include/secp256k1.h
```diff
@@ -427,6 +427,13 @@ typedef void (*secp256k1_sha256_compression_function)(
  * specialized implementation. It is NOT meant for replacing SHA256
  * with a different hash function.
  *
+ * Since auxiliary functions exposed by the library via a function
+ * pointer such as secp256k1_nonce_function_default not take a context
```
theStack: missing verb
```suggestion
 * pointer such as secp256k1_nonce_function_default don't take a context
```

include/secp256k1.h
```diff
@@ -427,6 +427,13 @@ typedef void (*secp256k1_sha256_compression_function)(
...
+ * object, they will not use the callback when called directly from user
+ * code. (But they will use the callback when called from other library
+ * functions that do take a context object, e.g., when
+ * secp256k1_nonce_function_default is passed to secp256k1_ecdsa_sign.)
```
theStack: seems that for `secp256k1_ecdsa_sign` (and also for `secp256k1_ecdh`), _only_ passing NULL as function pointer currently leads to usage of the context-aware nonce function call, so the recommendation to pass `_nonce_function_default` explicitly wouldn't work:
https://github.com/bitcoin-core/secp256k1/blob/528863e61f55ba3502f680643c55491bca04c150/src/secp256k1.c#L562-L567

(maybe something we could still fix? // EDIT: thinking of something like https://github.com/theStack/secp256k1/commit/6847ed4475cca7602efe950ed9139010b3e0ff22)

real-or-random: Ah yes, I thought that's the case because we did the same for schnorrsig (as you noticed) and also for ellswift. I will include it in this PR.

> so the recommendation to pass _nonce_function_default explicitly

Okay, yes, and I didn't intent to add this recommendation. If users want the default, passing NULL directly is more natural (though it doesn't make a difference in the end). I will try to rephrase this.

include/secp256k1.h
real-or-random: fixed

include/secp256k1.h
real-or-random: added a commit

theStack:
ACK modulo s/245/256/ nit in 2f38d0cb27854d8b7c87dc0b958dfe77ce13c2b1 commit subject

real-or-random:
> ACK modulo s/245/256/ nit in [2f38d0c](https://github.com/bitcoin-core/secp256k1/commit/2f38d0cb27854d8b7c87dc0b958dfe77ce13c2b1) commit subject

lol, fixed :)

theStack (approved):
ACK 4147f8bdf628fa95b42adc1d7ce55436eb63861e

real-or-random:
cc @hebasto want to review this too, so it can safely make it into the release?

hebasto (approved):
ACK 4147f8bdf628fa95b42adc1d7ce55436eb63861e, I have reviewed the code and it looks OK.

include/secp256k1.h
```diff
@@ -427,6 +427,14 @@ typedef void (*secp256k1_sha256_compression_function)(
  * specialized implementation. It is NOT meant for replacing SHA256
  * with a different hash function.
  *
+ * Since auxiliary functions exposed by the library via a function
+ * pointer such as secp256k1_nonce_function_default do not take a
+ * context object, they will not use the callback when called directly
+ * from user code. (But they will use the callback when called from
+ * other library functions that do take a context object, e.g., when
+ * noncefp==NULL or noncefp==secp256k1_nonce_function_default is passed
```
furszy, line 435: small note; this comment kinda assumes `secp256k1_nonce_function_default` will always be equal to the used `secp256k1_nonce_function_rfc6979` (we are using that instead of the default in the first commit).

As there are no plans to change the default value ever, it is not a problem. Just something that itches me a bit.

furszy (commented):
ACK 4147f8bdf628fa95b42adc1d7ce55436eb63861e

include/secp256k1.h
theStack: I had similar thoughts, but also think it's fine. Even if we ever changed the default value (very unlikely indeed), we would then want to also adapt/extend the if clause(s) accordingly to also use the ctx-aware function for that e.g.
```
const secp256k1_nonce_function secp256k1_nonce_function_default = nonce_function_new;

{
    ...
    if (noncefp == NULL || noncefp == secp256k1_nonce_function_new) {
        ret = nonce_function_new_impl(...);
    } else if (noncefp == secp256k1_nonce_function_rfc6979) {
        ret = nonce_function_rfc6979_impl(...);
    } else {
        ret = !!noncefp(...);
    }
    ...
}
```
and the comment would still apply.

real-or-random: If I understand correctly, what @furszy wanted to say here is that the *docs* would get out-of-date should we ever switch to a default function which does not use SHA256.

 - Switching the default nonce function to a more efficient one was discussed earlier. It wouldn't be entirely crazy but at this age of the repo, we'd probably rather do this when (if!) we add a [modern ECDSA module](https://github.com/bitcoin-core/secp256k1/issues/974). 
 - But switchting the default nonce function to one not based on SHA256 would be crazy. This should address @furszy's concern.

theStack: Oh, I see now, sorry for misinterpreting then if that was the case. Yeah, not using SHA256 seems pretty wild and far-fetched :)

PR #1859, opened 2026-05-30
field: force-inline 5x52 mul and sqr

l0rinc:
**Problem:** The 5x52 field multiplication and squaring routines are hot in group arithmetic and scalar multiplication. Some compilers leave the thin wrappers and int128 inner helpers out of line, which keeps a call boundary in this hot path and limits scheduling of the 64x64->128 arithmetic.

**Fix:** Define `SECP256K1_FORCE_INLINE` next to the existing inline helper and use it for the 5x52 multiplication and squaring wrappers and `int128` inner helpers.

For default optimized builds, this expands to `__forceinline` on MSVC-compatible compilers and to `__attribute__((always_inline))` on GCC-compatible compilers. It falls back to the existing inline spelling when inlining is disabled, when optimization is disabled, when optimizing for size on GCC/Clang, or when `_DEBUG` is defined.

**Benchmarks:** Values are relative changes in `Min(us)`, lower is better.

| Source | Host / CPU | Compiler | ecdsa_verify | ecdh | schnorrsig_verify | field_sqr | field_mul |
|---|---|---|---:|---:|---:|---:|---:|
| local | M4-Max.local | gcc-14 14.3.0 | -9.1% | -9.0% | -9.6% | -7.0% | -4.0% |
| local | i9-ssd | GCC 16.1.0 | -5.3% | -4.1% | -5.5% | -15.7% | -11.6% |
| local | WIN-A2EHOAU4JET / Xeon E5-2637 v2 | MSVC 19.50.35728 | -2.6% | -9.3% | -2.4% | -7.4% | -7.4% |
| local | i7-hdd | GCC 14.2.0 | -10.9% | -11.1% | -10.5% | -9.4% | -21.6% |
| local | umbrel / Intel N150 | GCC 12.2.0 | -4.9% | -4.3% | -4.6% | +0.6% | -1.1% |
| local | rpi5-16-3 | GCC 14.2.0 | -0.6% | -0.7% | -0.6% | -5.5% | -1.0% |
| local | rpi4-2-1 | GCC 14.2.0 | -2.7% | -2.3% | -2.7% | -5.6% | -4.0% |
| local | nodl / Cortex-A53 | GCC 11.4.0 | -3.3% | -7.6% | -5.7% | -9.9% | -1.8% |
| andrewtoth | i9-14900HX | GCC 12.3 | -5.3% | -4.2% | -5.6% | -1.5% | -6.1% |
[...]

PR #1902, opened 2026-07-30
changelog: add entry for #1859

theStack:
As discussed with @real-or-random: an entry for this PR wouldn't be strictly necessary, but the performance improvement is worthy to mention and incentivizes users to upgrade, for getting security fixes. Intentionally kept short to match typical changelog size entries; I guess we could be more detailed in some points if we wanted to (e.g. avoid the potentially misleading "up to" phrasing and/or give also concrete numbers for the size increase).
cc @l0rinc

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
  - The `silentpayments` API currently requires full access to the transaction data (light client scanning is not implemented).
 
 #### Changed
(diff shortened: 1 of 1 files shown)


real-or-random (approved):
ACK https://github.com/bitcoin-core/secp256k1/pull/1902/changes/51fc633e3ec3d09c4f71415fb74bfaba8010353e

l0rinc (approved):
ACK 51fc633e3ec3d09c4f71415fb74bfaba8010353e

Thanks for covering this.

PR #1709, opened 2025-07-22
release-process: Add signing of tarball plus minor improvements

real-or-random:

PR #1903, opened 2026-07-31
changelog: add entry for #1821

theStack:
Even though #1821 has the "needs-changelog" label, it doesn't show up in the search: https://github.com/bitcoin-core/secp256k1/pulls?q=is%3Apr+label%3Aneeds-changelog+is%3Aclosed (probably a bug?). Thanks to @real-or-random for finding this candidate by looking at the commit list manually (which is now also suggested in the release process since #1709).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
    - Usage example `examples/silentpayments.c`.
  - The `silentpayments` API currently requires full access to the transaction data (light client scanning is not implemented).
 
+#### Fixed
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
    - Usage example `examples/silentpayments.c`.
  - The `silentpayments` API currently requires full access to the transaction data (light client scanning is not implemented).
 
+#### Fixed
+ - Module `ellswift`: `secp256k1_ellswift_xdh` now treats secret keys greater than or equal to the curve order as invalid and returns 0. Previously, keys greater than the curve order were silently reduced modulo the order and accepted.
```
real-or-random:
```suggestion
 - Module `ellswift`: `secp256k1_ellswift_xdh` now treats secret keys greater than or equal to the curve order as invalid and returns 0 as documented. Previously, keys greater than the curve order were silently reduced modulo the order and accepted. The probability that a securely generated key is greater than the curver is negligible, and thus the old behavior does not constitute a security issue.
```

theStack: Thanks, applied.

real-or-random (approved):
ACK 2076b06a42b096315ead528acc7fe78da64d8d9c

PR #1904, opened 2026-07-31
sha256: cross-check caller supplied compression function

furszy:
This is something we left for a follow-up in the original PR #1777.

Essentially, the existing selftest hashes one 63 byte string against a known
digest. Which catches a compression function that is wrong everywhere,
but not one that is wrong on multi-block calls, unaligned input, or a state
that is not the IV.

This introduces `secp256k1_sha256_compression_equiv`, which hashes
messages of various lengths and starting offsets with both the supplied
function and the built-in one, and fails if the digest differs.

The check runs once during startup, so a faulty compression function
is caught early rather than silently producing incorrect output later.

Measured locally, this takes 0.05ms, which should be negligible for any
application.

See the introduced test for a clear view of the bugs this catches.

### DIFF
--- a/src/hash_impl.h
+++ b/src/hash_impl.h
@@ -137,6 +137,71 @@ static void secp256k1_sha256_transform(uint32_t *state, const unsigned char *blo
     }
 }
 
+/* Perform a smoke test on a supplied SHA256 compression function. */
+static int secp256k1_sha256_smoke_test(const secp256k1_sha256_compression_function fn_compression) {
+    secp256k1_hash_ctx ctx;
+    secp256k1_sha256 sha_msg, sha_accum;
+    unsigned char out[32];
+    size_t i, j;
+
+    /* SHA256 works on 64 byte blocks, secp256k1_sha256_write gives as many blocks
+     * at once to compression, so the count is what varies here. A SIMD implementation
+     * typically hashes four or eight at a time, then any left over one by one.
+     * These lengths cover every number from 1 to 9, which includes counts that
+     * divide evenly and counts leaving one, two or three over. */
+    static const size_t msg_lens[] = {
+        0, 1, 28,               /* Shorter than a block, so padding makes up the rest */
+        55, 56,                 /* Final 0x80 and 8 byte length fit in the last block, and don't */
+        64, 128, 192, 256, 320, /* 1 to 5 blocks */
+        384, 448, 512, 576      /* 6 to 9 blocks */
+    };
+    unsigned char msg[640]; /* Longest message, plus 64 for the shifted start */
+
+    /* Accumulated digest of every message, hashed with the built-in secp256k1_sha256_transform.
+     * Note: To regenerate set 'ctx.fn_sha256_compression = secp256k1_sha256_transform' below
+     * and print the sha_accum digest. */
+    static const unsigned char accum_expected[32] = {
+        0x22, 0x8E, 0x6A, 0x1F, 0x78, 0x02, 0x1B, 0xCF,
+        0x4E, 0xF4, 0xEA, 0xB3, 0x8A, 0x40, 0x69, 0xBF,
+        0x8D, 0xF3, 0x72, 0xAB, 0xE2, 0x11, 0x93, 0xA6,
+        0xE6, 0x46, 0x98, 0xBF, 0xD4, 0x3D, 0x19, 0x84,
+    };
+    /* The purpose of this VERIFY_CHECK is to make anyone aware that they
+     * should also change the size of msg_buf when changing the length of the
+     * longest message. */
+    VERIFY_CHECK(msg_lens[ARRAY_SIZE(msg_lens) - 1] == 576);
+
+    VERIFY_CHECK(fn_compression != NULL);
+    secp256k1_hash_ctx_init(&ctx);
+    ctx.fn_sha256_compression = fn_compression;
+    secp256k1_sha256_initialize(&sha_accum);
+
+    /* No two blocks of a message are equal, so a function that doesn't advance the block
+     * pointer gives a different digest. 251 is prime, so its repeats only line up with
+     * block starts every 251 blocks; wrapping at 256 would do so every 4. */
+    for (i = 0; i < sizeof(msg); i++) {
+        msg[i] = (unsigned char)(i % 251);
+    }
+
+    /* Each round starts one byte further along to check different alignments,
+     * secp256k1_sha256_write invokes compression directly on input >= 64 bytes */
+    for (i = 0; i < 64; i++) {
+        unsigned char *m = msg + i;
+        m[0] ^= 0xff; /* Changes the first byte, so every state after it changes too */
+        for (j = 0; j < ARRAY_SIZE(msg_lens); j++) {
+            secp256k1_sha256_initialize(&sha_msg);
+            secp256k1_sha256_write(&ctx, &sha_msg, m, msg_lens[j]);
+            secp256k1_sha256_finalize(&ctx, &sha_msg, out);
+            secp256k1_sha256_write(&ctx, &sha_accum, out, 32);
+        }
+        m[0] ^= 0xff; /* Reset first byte */
+    }
+
+    /* Compare against pre-computed accumulated digest */
+    secp256k1_sha256_finalize(&ctx, &sha_accum, out);
+    return secp256k1_memcmp_var(accum_expected, out, 32) == 0;
+}
+
 static void secp256k1_hash_ctx_init(secp256k1_hash_ctx *hash_ctx) {
     VERIFY_CHECK(hash_ctx != NULL);
     hash_ctx->fn_sha256_compression = secp256k1_sha256_transform;
--- a/include/secp256k1.h
+++ b/include/secp256k1.h
@@ -431,6 +431,13 @@ typedef void (*secp256k1_sha256_compression_function)(
  * noncefp==NULL or noncefp==secp256k1_nonce_function_default is passed
  * as an argument to secp256k1_ecdsa_sign.)
  *
+ * Note: The provided function is tested against a set of known SHA256
+ * digests; invokes the context's illegal callback on any mismatch
+ * (which aborts by default), in order to catch basic misbehavior early.
+ * It takes well under 2.5 ms on a desktop machine.
+ * This is NOT a substitute for having proper test coverage of the
+ * supplied function outside this library.
+ *
  * Args:    ctx:             pointer to a context object.
  * In:      fn_compression:  pointer to a function implementing the compression function;
  *                           passing NULL restores the default implementation.
--- a/src/selftest.h
+++ b/src/selftest.h
@@ -24,7 +24,13 @@ static int secp256k1_selftest_sha256(secp256k1_sha256_compression_function fn_co
     hash_ctx.fn_sha256_compression = fn_compression;
     secp256k1_sha256_write(&hash_ctx, &hasher, (const unsigned char*)input63, 63);
     secp256k1_sha256_finalize(&hash_ctx, &hasher, out);
-    return secp256k1_memcmp_var(out, output32, 32) == 0;
+    if (secp256k1_memcmp_var(out, output32, 32) != 0) {
+        return 0;
+    }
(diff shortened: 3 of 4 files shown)


src/hash_impl.h
```diff
@@ -137,6 +137,66 @@ static void secp256k1_sha256_transform(uint32_t *state, const unsigned char *blo
...
+        64, 128, 192, 256, 320, /* 1 to 5 blocks */
+        384, 448, 512, 576      /* 6 to 9 blocks */
+    };
+    unsigned char msg[640]; /* Longest message, plus 64 for the shifted starts */
+
+    VERIFY_CHECK(fn_transform_supplied != NULL);
+    /* Nothing to learn from comparing the built-in against itself. */
+    if (fn_transform_supplied == secp256k1_sha256_transform) {
+        return 1;
+    }
```
real-or-random: nit: I think moving this check to the caller will enhance readability.

src/hash_impl.h
```diff
@@ -137,6 +137,66 @@ static void secp256k1_sha256_transform(uint32_t *state, const unsigned char *blo
...
+    /* No two blocks of a message are equal, so a function that doesn't advance the block
+     * pointer gives a different digest. 251 is prime, so its repeats only line up with
+     * block starts every 251 blocks; wrapping at 256 would do so every 4. */
+    for (i = 0; i < sizeof(msg); i++) {
+        msg[i] = (unsigned char)(i % 251);
+    }
+
+    /* Each round starts one byte further along to check different alignments,
+     * secp256k1_sha256_write invokes compression directly on input >= 64 bytes */
+    for (i = 0; i < 64; i++) {
```
real-or-random, line 188: Checking alignment requirements is a great idea. 

But I think running the entire thing 64 times is overkill. Checking i=0 and i=1 should suffice, no? Any requirement that relies on data being aligned will be violated if the pointer is moved by one. That is, if it works for i=0 and has some alignment requirement, then it certainly won't work for i=1 (and the other way around). 

```suggestion
    /* Pass pointers `msg` and `msg + 1` to compression to catch alignment issues, 
     * secp256k1_sha256_write invokes compression directly on input >= 64 bytes */
    for (i = 0; i < 2; i++) {
```

src/hash_impl.h
```diff
@@ -137,6 +137,66 @@ static void secp256k1_sha256_transform(uint32_t *state, const unsigned char *blo
...
+    VERIFY_CHECK(fn_transform_supplied != NULL);
+    /* Nothing to learn from comparing the built-in against itself. */
+    if (fn_transform_supplied == secp256k1_sha256_transform) {
+        return 1;
+    }
+
+    secp256k1_hash_ctx_init(&ctxs[0]);
+    secp256k1_hash_ctx_init(&ctxs[1]);
+    ctxs[0].fn_sha256_compression = secp256k1_sha256_transform; /* Built-in */
+    ctxs[1].fn_sha256_compression = fn_transform_supplied;      /* User provided */
```
real-or-random: Instead of running the built-in function, we could also do this:
 - Just run the user-provided function
 - Accumulate all resulting SHA256 outputs into another SHA256 `accum_hash` (again using the built-in function)
 - Compare the output `accum_hash` against a constan, i.e., the precomputed correct value (use debugging macro `PRINT_BUF` to get the value easily)

This would make it even faster with basically no losses in coverage.

real-or-random (commented):
Concept ACK

Hm, it would be great to have this in before the release on Monday evening (UTC).

We should probably also add a comment to the docs of `secp256k1_context_set_sha256_compression` sayting that this API function will run a cursory test to check that callback behaves as it should, and maybe that it costs <1 ms on a desktop machine (or whatever the cost is). 

If the "precomputed" optimization that I suggest below is too much of a hassle for now, we could also do this later.

theStack:
Concept ACK

src/hash_impl.h
furszy: > But I think running the entire thing 64 times is overkill. Checking i=0 and i=1 should suffice, no? Any requirement that relies on data being aligned will be violated if the pointer is moved by one. That is, if it works for i=0 and has some alignment requirement, then it certainly won't work for i=1 (and the other way around).

What if it is not about expecting data to be aligned, but a fast path that is taken only if a specific alignment is provided. E.g.

    if (ptr is 16-byte aligned)
        super_fast_path(ptr)   // but this has a bug
    else
        regular_path(ptr)

This would not get caught by msg + 1, which is odd, and whether msg catches it is out of our control, as we don't specify any alignment?

Still, I agree it may be way too far fetched. Let me apply the {0, 1} simplification.

src/hash_impl.h
furszy: Sure, done as suggested

src/hash_impl.h
furszy: As the check now takes under 0.05 ms, we can run this for the built-in compression too. Which doesn't hurt and ensures we did not screw it up. So I just dropped the line.

furszy:
Updated per feedback, thanks real-or-random!

The check now accumulates the msg digests and compares their accumulated result against a precomputed constant. Moved alignment loop down to {0, 1} only, and added docs to `secp256k1_context_set_sha256_compression`.

Measured locally, now it takes under ~0.05 ms.

src/hash_impl.h
```diff
@@ -137,6 +137,72 @@ static void secp256k1_sha256_transform(uint32_t *state, const unsigned char *blo
...
+    for (i = 0; i < sizeof(msg); i++) {
+        msg[i] = (unsigned char)(i % 251);
+    }
+
+    /* Pass pointers `msg` and `msg + 1` to compression to catch alignment issues,
+    * secp256k1_sha256_write invokes compression directly on input >= 64 bytes */
+    for (i = 0; i < 2; i++) {
+        unsigned char *m = msg + i;
+        m[0] ^= 0xff; /* Changes the first byte, so every state after it changes too */
+        for (j = 0; j < sizeof(msg_lens) / sizeof(msg_lens[0]); j++) {
```
theStack: nit:
```suggestion
        for (j = 0; j < ARRAY_SIZE(msg_lens); j++) {
```

src/hash_impl.h
```diff
@@ -137,6 +137,72 @@ static void secp256k1_sha256_transform(uint32_t *state, const unsigned char *blo
...
+     * and print the sha_accum digest. */
+    static const unsigned char accum_out[32] = {
+        0x76, 0xA7, 0x7C, 0x95, 0xE0, 0x96, 0x0E, 0xE5,
+        0x21, 0xCE, 0xDC, 0x95, 0x81, 0x0B, 0xDD, 0xD1,
+        0x38, 0xCB, 0xB1, 0xA2, 0xED, 0xB7, 0x03, 0xF3,
+        0xD2, 0xAE, 0xB3, 0x3E, 0x82, 0x34, 0xB1, 0x14
+    };
+
+    VERIFY_CHECK(fn_compression != NULL);
+    secp256k1_hash_ctx_init(&ctx);
```
theStack, line 175: nit: could remove this line, as it only sets the internal compression function, which is overwritten in the next line (it's also not called in `secp256k1_selftest_sha256`)
```suggestion
```

src/hash_impl.h
```diff
@@ -137,6 +137,72 @@ static void secp256k1_sha256_transform(uint32_t *state, const unsigned char *blo
...
+            secp256k1_sha256_finalize(&ctx, &sha_msg, out);
+            secp256k1_sha256_write(&ctx, &sha_accum, out, 32);
+        }
+        m[0] ^= 0xff; /* Reset first byte */
+    }
+
+    /* Compare against pre-computed accumulated digest */
+    secp256k1_sha256_finalize(&ctx, &sha_accum, out);
+    secp256k1_sha256_clear(&sha_msg);
+    secp256k1_sha256_clear(&sha_accum);
```
theStack: nit: could remove these two lines, as there is no sensitive data involved
```suggestion
```

theStack (commented):
Left some simplification nits below, nothing blocking

src/hash_impl.h
real-or-random: > What if it is not about expecting data to be aligned, but a fast path that is taken only if a specific alignment is provided. 

Oh I didn't consider this case, and I tend to say that this convinces me... What do you think about this? 
https://github.com/bitcoin-core/secp256k1/compare/master...real-or-random:secp256k1:2026_sha_comp_compression (compiles, but didn't have the time to test, sorry :/)

theStack: > > What if it is not about expecting data to be aligned, but a fast path that is taken only if a specific alignment is provided.
> 
> Oh I didn't consider this case, and I tend to say that this convinces me... What do you think about this? [master...real-or-random:secp256k1:2026_sha_comp_compression](https://github.com/bitcoin-core/secp256k1/compare/master...real-or-random:secp256k1:2026_sha_comp_compression) (compiles, but didn't have the time to test, sorry :/)

Wouldn't this solution lead to different `accum_out` values depending on MAX_ALIGNMENT, meaning we would need to store a lot of different constants? (If yes, maybe a middle ground could be to set MAX_ALIGNMENT to a fixed value instead of depending on `__BIGGEST_ALIGNMENT__`?)

real-or-random: Oh right, I overdid it there. Fixed, and also fixed another bug.

theStack (approved):
ACK 4f91975777e3332bb27ec2fa0f71e31c8ab4f448

In case we can't convince ourselves / get enough review for #1905, we should get this simpler variant in (which I'd be much more comfortable merging even without additional review due to significantly lower complexity). @real-or-random made a good point about the performance regression of context creation though (1.5us vs. 75us, see https://github.com/bitcoin-core/secp256k1/pull/1905#issuecomment-5166498238), so we might want to open a simple follow-up PR that skips the check for the built-in compression function to avoid that.

theStack (approved):
re-ACK ccbdd4925380735fe61ac2520a5c6fe7e679f334

Verified that compared to my previous ACK, the check for the built-in compression function is skipped (avoiding a performance regression in `_context_create`), the function names have been aligned to #1905 (making a potential rebase simpler) and most nits above have been tackled. Thanks!

src/hash_impl.h
furszy: done

src/hash_impl.h
furszy: done

src/hash_impl.h
furszy: I left this one in case we ever add other stuff to `secp256k1_hash_ctx_init`. It doesn't hurt to have it.

theStack:
As discussed off-band with @real-or-random earlier today, I'm merging this now as it is a much simpler (though also less thorough in alignment-testing) alternative to #1905. If we are okay with a longer initialization time of `_context_set_sha256_compression` function (~2.2ms instead of ~0.07ms), we could also still bump the loop range from 2 to 64 in a follow-up PR (cc @furszy).

real-or-random (approved):
ACK c84ea46561d9a224d39ff6f0f9d044613f402d27

theStack (approved):
ACK c84ea46561d9a224d39ff6f0f9d044613f402d27

src/hash_impl.h
```diff
@@ -137,6 +137,71 @@ static void secp256k1_sha256_transform(uint32_t *state, const unsigned char *blo
...
+     * and print the sha_accum digest. */
+    static const unsigned char accum_expected[32] = {
+        0x22, 0x8E, 0x6A, 0x1F, 0x78, 0x02, 0x1B, 0xCF,
+        0x4E, 0xF4, 0xEA, 0xB3, 0x8A, 0x40, 0x69, 0xBF,
+        0x8D, 0xF3, 0x72, 0xAB, 0xE2, 0x11, 0x93, 0xA6,
+        0xE6, 0x46, 0x98, 0xBF, 0xD4, 0x3D, 0x19, 0x84,
+    };
+    /* The purpose of this VERIFY_CHECK is to make anyone aware that they
+     * should also change the size of msg_buf when changing the length of the
+     * longest message. */
```
real-or-random, line 171: post-merge comment nit, so we don't forget:
```suggestion
    /* The purpose of this VERIFY_CHECK is to make anyone aware that they
     * should also change the size of msg when changing the length of the
     * longest message. */
```

PR #1905, opened 2026-08-03
sha256: smoke-test caller-supplied compression function

real-or-random:
Extended version of #1904. I'd say if we get enough reviews/confidence before the release, let's merge this here. Otherwise, let's stick with the simpler #1904 and leave this here for after the release.

### DIFF
--- a/src/hash_impl.h
+++ b/src/hash_impl.h
@@ -137,6 +137,94 @@ static void secp256k1_sha256_transform(uint32_t *state, const unsigned char *blo
     }
 }
 
+/* Perform a smoke test on a supplied SHA256 compression function. */
+static int secp256k1_sha256_smoke_test(const secp256k1_sha256_compression_function fn_compression) {
+    secp256k1_hash_ctx ctx;
+    secp256k1_sha256 sha_msg, sha_accum;
+    unsigned char out[32];
+    size_t i, j;
+    size_t alignment, offset;
+    unsigned char* msg;
+
+    /* SHA256 works on 64 byte blocks, secp256k1_sha256_write gives as many blocks
+     * at once to compression, so the count is what varies here. A SIMD implementation
+     * typically hashes four or eight at a time, then any left over one by one.
+     * These lengths cover every number from 1 to 9, which includes counts that
+     * divide evenly and counts leaving one, two or three over. */
+    #define LONGEST 576  /* local macro, undef'd below */
+    static const size_t msg_lens[] = {
+        0, 1, 28,               /* Shorter than a block, so padding makes up the rest */
+        55, 56,                 /* Final 0x80 and 8 byte length fit in the last block, and don't */
+        64, 128, 192, 256, 320, /* 1 to 5 blocks */
+        384, 448, 512, LONGEST  /* 6 to 9 blocks */
+    };
+
+    /* Max alignment to test; even AVX512 needs at most 64 byte alignment. */
+    #define MAX_ALIGNMENT 64 /* local macro, undef'd below */
+    /* sizeof(msg) = length of longest message
+     *               + 2 * MAX_ALIGNMENT - 1 for the offset in the initial value of m
+     *               + 2 * MAX_ALIGNMENT - 1 for the offset in the final value of m */
+    unsigned char msg_buf[LONGEST + 2 * (2 * MAX_ALIGNMENT - 1)];
+
+    /* Accumulated digest of every message, hashed with the built-in secp256k1_sha256_transform.
+     * Note: To regenerate set 'ctx.fn_sha256_compression = secp256k1_sha256_transform' below
+     * and print the sha_accum digest. */
+    static const unsigned char accum_expected[32] = {
+        0x1E, 0x2A, 0xD6, 0x29, 0x16, 0x9D, 0x46, 0xF0,
+        0xD2, 0x31, 0x94, 0xF0, 0xBC, 0x8C, 0x4B, 0x87,
+        0x37, 0x68, 0xA0, 0xFB, 0xD2, 0x9E, 0xF6, 0xB9,
+        0x3E, 0x0C, 0xFA, 0xCE, 0xDE, 0x5F, 0xD3, 0x72,
+    };
+    /* The purpose of this VERIFY_CHECK is to make anyone aware that they
+     * should also change the size of msg_buf when changing the length of the
+     * longest message. */
+    VERIFY_CHECK(msg_lens[ARRAY_SIZE(msg_lens) - 1] == 576);
+
+    VERIFY_CHECK(fn_compression != NULL);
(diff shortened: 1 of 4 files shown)


real-or-random:
Differences: 
 - This tests multiple alignments of the message, namely aligned to 64 bytes (but not to 128), aligned to 32 bytes (but not to 64), ...., down to no alignment at all. (Since alignment is always a power of 2, this should be equal to @furszy's first proposal in coverage, but more efficient because we need the loop only log(64)+1=7 times, instead of 64 times).  
 - This skips the test if we know that we're dealing with the internal function (as @furszy's initial proposal). This keeps context creation to ~1.5 us on my machine (instead of ~75 us).
 - A few cleanups suggested by @theStack. (I kept the `hash_ctx_init` call because it's in principle correct. A hash context may have more stuff to init in the future. Anyway, the compiler will figure out the dead store and optimize it away.)
 - Renamed the function to `smoke_test` (`equiv` isn't a great any longer since it doesn't run the built-in anymore).
 - Adjust the commit message.

theStack (approved):
ACK c1c6365d99dea5d6e68f4bd71136424c7e5e5223

Being thorough with alignment checks seems the right approach, even though it adds quite a bit of complexity to the code (it LGTM, but it's very easy to get something wrong there or overlook something as reviewer).

The following Copilot suggestions seem to make sense to tackle:
* https://github.com/bitcoin-core/secp256k1/pull/1905#discussion_r3704028895
* https://github.com/bitcoin-core/secp256k1/pull/1905#discussion_r3704028987

src/hash_impl.h
```diff
@@ -137,6 +137,94 @@ static void secp256k1_sha256_transform(uint32_t *state, const unsigned char *blo
...
+    size_t i, j;
+    size_t alignment, offset;
+    unsigned char* msg;
+
+    /* SHA256 works on 64 byte blocks, secp256k1_sha256_write gives as many blocks
+     * at once to compression, so the count is what varies here. A SIMD implementation
+     * typically hashes four or eight at a time, then any left over one by one.
+     * These lengths cover every number from 1 to 9, which includes counts that
+     * divide evenly and counts leaving one, two or three over. */
+    #define LONGEST 576  /* local macro, undef'd below */
```
theStack, line 154: nit: an alternative here and for `MAX_ALIGNMENT` below would be
```suggestion
    enum { LONGEST = 576 };
```
for not having to `#undef` below, leading to a smaller patch; but not sure if that's strictly better or if we even have a preference in the project (we seem to use both)

src/hash_impl.h
```diff
@@ -137,6 +137,94 @@ static void secp256k1_sha256_transform(uint32_t *state, const unsigned char *blo
...
+     * longest message. */
+    VERIFY_CHECK(msg_lens[ARRAY_SIZE(msg_lens) - 1] == 576);
+
+    VERIFY_CHECK(fn_compression != NULL);
+    secp256k1_hash_ctx_init(&ctx);
+    ctx.fn_sha256_compression = fn_compression;
+    secp256k1_sha256_initialize(&sha_accum);
+
+    /* Make msg a pointer into msg_buf aligned to 2 * MAX_ALIGNMENT boundary. */
+    offset = ((2 * MAX_ALIGNMENT) - ((uintptr_t)msg_buf % (2 * MAX_ALIGNMENT))) % (2 * MAX_ALIGNMENT);
```
theStack, line 189: nit-like (?): `uintptr_t` does not seem to be part of C89; it was introduced in C99 and is optional even there. Not sure how strict we want to be about this level of portability, though, given that we already rely on `stdint.h` being available.

furszy (commented):
Simple coverage for the introduced changes:

```patch
diff --git a/src/tests.c b/src/tests.c
--- a/src/tests.c
+++ b/src/tests.c
@@ -517,12 +517,28 @@
     s[0] ^= 1;
 }
 
+/* Wrong when input is 64-byte aligned, like a broken SIMD fast path. */
+static void sha256_transform_aligned_fail(uint32_t *s, const unsigned char *chunk, size_t blocks) {
+    int aligned = ((uintptr_t)chunk % 64) == 0;
+    secp256k1_sha256_transform(s, chunk, blocks);
+    if (aligned) s[0] ^= 1;
+}
+
+/* Wrong on any unaligned input. */
+static void sha256_transform_unaligned_fail(uint32_t *s, const unsigned char *chunk, size_t blocks) {
+    int aligned = ((uintptr_t)chunk % 64) == 0;
+    secp256k1_sha256_transform(s, chunk, blocks);
+    if (!aligned) s[0] ^= 1;
+}
+
 static void run_sha256_compression_smoke_test_tests(void) {
     CHECK(secp256k1_sha256_smoke_test(sha256_transform_noadvance) == 0);
     CHECK(secp256k1_sha256_smoke_test(sha256_transform_short) == 0);
     CHECK(secp256k1_sha256_smoke_test(sha256_transform_ivreset) == 0);
     CHECK(secp256k1_sha256_smoke_test(sha256_transform_batch4) == 0);
     CHECK(secp256k1_sha256_smoke_test(sha256_transform_corrupt) == 0);
+    CHECK(secp256k1_sha256_smoke_test(sha256_transform_aligned_fail) == 0);
+    CHECK(secp256k1_sha256_smoke_test(sha256_transform_unaligned_fail) == 0);
     CHECK(secp256k1_sha256_smoke_test(good_sha256_compression) == 1);
 }
 

```

src/hash_impl.h
real-or-random: Hm, yes. Almost all architectures have an `uintptr_t`, but some don't.

The reason not to provide one is that there is no integer type large enough to hold a pointer value so that it can be converted back without loss. We don't need to conversion back, but we still can't just cast to any unsigned integer type because C99 says: If the result cannot be represented in the integer type, the behavior is undefined. 

We could ignore the problem and simply require `uintptr_t`. We could wrap the `offset` calculation into a preprocessor check that checks for the presence of the type (`#ifdef UINTPTR_MAX`) and fall back if it's not present (e.g., simply set `offset = 0`, effectively not doing alignment checks) for these rare architectures. 

[1] ~CHERI archs can't provide it it according to https://www.ralfj.de/blog/2022/04/11/provenance-exposed.html.~ edit: I was wrong here. [CHERI-C](https://www.cl.cam.ac.uk/techreports/UCAM-CL-TR-947.pdf) has it, and it's 64 bits long. According to https://www.open-std.org/JTC1/SC22/WG14/www/docs/n2873.htm, there are at least two other architectures, but they're not mentioned explicitly in the document.

src/hash_impl.h
real-or-random: I tend to think that `enum` is better in this case. I always forget that it's a compile-time constant that can be used in array sizes.

PR #1909, opened 2026-08-05
musig: fix pseudocode in `_musig_pubkey_ec_tweak_add` API docs

theStack:
Reported by [Project Loupe](https://github.com/project-loupe)

### DIFF
--- a/include/secp256k1_musig.h
+++ b/include/secp256k1_musig.h
@@ -236,7 +236,7 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_pubkey_get(
  *
  *  secp256k1_musig_pubkey_agg(..., keyagg_cache, pubkeys, ...)
  *  secp256k1_musig_pubkey_get(..., agg_pk, keyagg_cache)
- *  secp256k1_musig_pubkey_ec_tweak_add(..., output_pk, tweak32, keyagg_cache)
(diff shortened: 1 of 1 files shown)


include/secp256k1_musig.h
```diff
@@ -236,7 +236,7 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_pubkey_get(
  *
  *  secp256k1_musig_pubkey_agg(..., keyagg_cache, pubkeys, ...)
  *  secp256k1_musig_pubkey_get(..., agg_pk, keyagg_cache)
- *  secp256k1_musig_pubkey_ec_tweak_add(..., output_pk, tweak32, keyagg_cache)
```
l0rinc, line 239: I also found many others a while ago: https://github.com/l0rinc/secp256k1/commit/e1cd68c630fe34bf1620c10a6e0cebf2757ef58b

theStack: Nice, either that was not on one of the audit branches you showed me in the past, or (more likely) I missed it. Want to open an alternative PR with that commit?

The exact wording of the tweak32 parameter doc change might need a closer look by reviewers (is it okay to only state "would be invalid", or do we want to keep being explicit how an invalid result is reached), but all the changes seem to make sense in principle.

PR #1910, opened 2026-08-05
scratch: reject sizes that overflow when added to header

theStack:
We only use scratch space internally, so this is not an issue for the user (scratch API functions have been removed since 0.6.0, see #1620), but fixing this might still make sense to prepare for the unlikely case that we expose the scratch API again in the future. An alternative could be to simply delete the scratch space functionality already, as e.g. done in PR #1789.

Reported by [Project Loupe](https://github.com/project-loupe)

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -420,6 +420,12 @@ static void run_scratch_tests(void) {
     CHECK(secp256k1_scratch_alloc(&CTX->error_callback, scratch, SIZE_MAX) == NULL);
     secp256k1_scratch_space_destroy(CTX, scratch);
 
+    /* Creating a scratch space whose size would wrap around when the aligned
+     * header size is added to it fails, both for SIZE_MAX and for the smallest
+     * size that still wraps. */
+    CHECK(secp256k1_scratch_space_create(CTX, SIZE_MAX) == NULL);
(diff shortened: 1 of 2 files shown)


src/tests.c
```diff
@@ -420,6 +420,12 @@ static void run_scratch_tests(void) {
     CHECK(secp256k1_scratch_alloc(&CTX->error_callback, scratch, SIZE_MAX) == NULL);
     secp256k1_scratch_space_destroy(CTX, scratch);
 
+    /* Creating a scratch space whose size would wrap around when the aligned
+     * header size is added to it fails, both for SIZE_MAX and for the smallest
+     * size that still wraps. */
+    CHECK(secp256k1_scratch_space_create(CTX, SIZE_MAX) == NULL);
```
l0rinc, line 426: Had a very similar finding a while back: https://github.com/l0rinc/secp256k1/commit/d7e3b4925c9a97aca6ffb4e7e55e0591f1f8d470

apoelstra:
utACK 3d4340d17325a6732108423f9d3e54de99539a60

real-or-random (approved):
utACK 3d4340d17325a6732108423f9d3e54de99539a60

PR #1911, opened 2026-08-07
refactor: rename `ctx` param to `ecmult_gen_ctx` where applicable

theStack:
To be more consistent and avoid confusion with regular `secp256k1_context` instances, rename the `ctx` parameter to `ecmult_gen_ctx` in all functions that take `secp256k1_ecmult_gen_context` pointers (see discussion https://github.com/bitcoin-core/secp256k1/issues/1835#issuecomment-4057203336 ff.).

While at it, move the * in pointer declarations of the touched functions to the right side for consistency across the code base, see PR #1252.

### DIFF
--- a/src/ecmult_gen_impl.h
+++ b/src/ecmult_gen_impl.h
@@ -14,20 +14,20 @@
 #include "hash_impl.h"
 #include "precomputed_ecmult_gen.h"
 
-static void secp256k1_ecmult_gen_context_build(secp256k1_ecmult_gen_context *ctx, const secp256k1_hash_ctx *hash_ctx) {
-    secp256k1_ecmult_gen_blind(ctx, hash_ctx, NULL);
-    ctx->built = 1;
+static void secp256k1_ecmult_gen_context_build(secp256k1_ecmult_gen_context *ecmult_gen_ctx, const secp256k1_hash_ctx *hash_ctx) {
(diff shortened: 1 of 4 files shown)


src/ecmult_gen_impl.h
```diff
@@ -107,7 +107,7 @@ static void secp256k1_ecmult_gen_gej(const secp256k1_ecmult_gen_context *ctx, se
      */
 
     /* Compute the scalar d = (gn + ctx->scalar_offset). */
-    secp256k1_scalar_add(&d, &ctx->scalar_offset, gn);
+    secp256k1_scalar_add(&d, &ecmult_gen_ctx->scalar_offset, gn);
```
real-or-random: The comment should also be changed. (And there are more comments that need to change.)

real-or-random (commented):
Concept ACK

src/ecmult_gen_impl.h
theStack: Thanks, updated (hope I caught them all).

real-or-random (approved):
utACK 81a5a7567d10b358e6e41ef6085adf2ab2abc879

apoelstra (approved):
ACK 81a5a7567d10b358e6e41ef6085adf2ab2abc879; successfully ran local tests

PR #1913, opened 2026-08-12
context: prevent clones from copying allocator residue

l0rinc:
**Problem:** Context creation initializes every member but leaves padding bytes untouched.
Cloning a dynamically allocated context into caller-provided storage may copy undefined allocator bytes with the other members.
[The caller remains responsible for that storage after destroying the context](https://github.com/bitcoin-core/secp256k1/blob/687155df6b76f1da1639a6f0923b69e6beaa3a09/include/secp256k1_preallocated.h#L105-L119), so memory checkers report the undefined bytes when the block is compared or written out.
This is a local initialization defect, no remote or Bitcoin Core path is known.

**Fix:** Zero the complete context representation before member initialization and clone assignment.
Extend the existing context test to require every byte to be defined after both creation modes and both clone modes.

### DIFF
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -128,6 +128,8 @@ secp256k1_context* secp256k1_context_preallocated_create(void* prealloc, unsigne
     }
     VERIFY_CHECK(prealloc != NULL);
     ret = (secp256k1_context*)prealloc;
+    /* Initialize padding before member writes leave it untouched. */
+    memset(ret, 0, sizeof(*ret));
     ret->illegal_callback = default_illegal_callback;
     ret->error_callback = default_error_callback;
     secp256k1_hash_ctx_init(&ret->hash_ctx);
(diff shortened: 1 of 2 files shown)


theStack:
It's unclear to me if this is a problem in the first place, as users are not meant to inspect the raw bytes of context objects (or any other opaque objects for that matter), so the mentioned "block is compared or written out" scenario which would presumably trigger a report by memory checkers seems one that is outside of the API contract. After destroying the context, that contract arguably ends, but I can't see why users would want to perform any reads from the memory either before freeing (I'd imagine they just do nothing if it's stack/global storage, or calling a `free` equivalent if it was dynamically allocated by the user), so it seems a bit far-fetched. If we are really worried about that latter scenario, maybe do the memset only at the end of `secp256k1_context_preallocated_destroy`, and leave the other functions unchanged? Curious what other reviewers think.

During the lifetime of a context object, uninitialized padding bytes seem to be just fine in any case.

l0rinc:
Thank you, closing, if other reviewers disagree we can reopen

PR #1844, opened 2026-04-11
optimize pubkey tweaking by using vartime jacobian->affine conversion (~10% speedup)

theStack:
This PR has a very similar theme and motivation as #1843, but is significantly simpler.

Tweaks are not considered sensitive data, and the point multiplications are not performed in constant-time either, so we can use the variable-time variant for converting to affine coordinates (ge_set_gej_var) as well to improve the performance. On my arm64 machine, this shows a ~10% speedup for both the additive and multiplicative public key tweaking API functions:

master branch
```
$ ./build/bin/bench tweak
Benchmark                     ,    Min(us)    ,    Avg(us)    ,    Max(us)

ec_pk_tweak_add               ,    16.3       ,    16.4       ,    16.7
ec_pk_tweak_mul               ,    19.7       ,    19.8       ,    19.8
```

PR branch
```
$ ./build/bin/bench tweak
Benchmark                     ,    Min(us)    ,    Avg(us)    ,    Max(us)

ec_pk_tweak_add               ,    14.6       ,    14.7       ,    14.9
ec_pk_tweak_mul               ,    18.1       ,    18.1       ,    18.2
```

The reduced execution time roughly matches the difference between `fe_inv` and `fe_inv_var` (see internal benchmarks), ~1.6 us on my machine.

Note that the improved code path for additive tweaking (function `secp256k1_eckey_pubkey_tweak_add`) is also used in the following API functions, so they should all benefit from it:
* [`secp256k1_xonly_pubkey_tweak_add`](https://github.com/bitcoin-core/secp256k1/blob/7262adb4b40074201fb30847035a82b8d742f350/src/modules/extrakeys/main_impl.h#L128)
* [`secp256k1_xonly_pubkey_tweak_add_check`](https://github.com/bitcoin-core/secp256k1/blob/7262adb4b40074201fb30847035a82b8d742f350/src/modules/extrakeys/main_impl.h#L145) (this one is used for [verifying P2TR script path spends in Bitcoin Core](https://github.com/bitcoin/bitcoin/blob/a7c30da1f6fc2eebd7514dd032fbb54940d467de/src/pubkey.cpp#L257-L263) and thus consensus-critical, see BIP341)
* [`secp256k1_keypair_xonly_tweak_add`](https://github.com/bitcoin-core/secp256k1/blob/7262adb4b40074201fb30847035a82b8d742f350/src/modules/extrakeys/main_impl.h#L274)
[...]

PR #1914, opened 2026-08-14
silentpayments: tweak spend pubkey using ecmult_gen_var (~25% scanning speedup for small N)

theStack:
Based on the #1883 branch, this PR takes advantage of the variable-time generator point multiplication function `ecmult_gen_var` for  spend public key tweaking in the output public key calculation $P_k = B_{spend} + t_k \cdot G$ (used both for [sending](https://github.com/bitcoin/bips/blob/60f5b33b0a7be3cf09b933d97b78071d684db7d1/bip-0352.mediawiki?plain=1#L313) and [scanning](https://github.com/bitcoin/bips/blob/60f5b33b0a7be3cf09b933d97b78071d684db7d1/bip-0352.mediawiki?plain=1#L349)). Note that $t_k$ is not considered a long-term secret (in contrast to e.g. the scan secret key), so using variable-time functions should be fine. This leads to a ~25% scanning speedup if the number of transaction outputs is small (N <= 10).

Results on my arm64 machine (as per `$ ./build/bin/bench silentpayments_scan_nomatch`, see also the commit body):

| N    | master    | PR branch | speedup |
|------|-----------|-----------|---------|
| 2    | 39.3 us   | 30.3 us   | 28.85%  |
| 5    | 40.4 us   | 31.7 us   | 27.44%  |
| 10   | 43.3 us   | 34.6 us   | 25.14%  |
| 100  | 89.6 us   | 81.2 us   | 10.34%  |
| 1000 | 1259.0 us | 1244.0 us | 1.99%   |

This PR is similar to the previous ones #1843, #1844, but it only changes the behavior of the silentpayments module and leaves other API functions unchanged.

### DIFF
--- a/src/bench_ecmult.c
+++ b/src/bench_ecmult.c
@@ -108,6 +108,15 @@ static void bench_ecmult_gen(void* arg, int iters) {
     }
 }
 
+static void bench_ecmult_gen_var(void* arg, int iters) {
+    bench_data* data = (bench_data*)arg;
+    int i;
+
+    for (i = 0; i < iters; ++i) {
+        secp256k1_ecmult_gen_var_gej(&data->output[i], &data->scalars[(data->offset1+i) % POINTS]);
+    }
+}
+
 static void bench_ecmult_gen_teardown(void* arg, int iters) {
     bench_data* data = (bench_data*)arg;
     bench_ecmult_teardown_helper(data, NULL, NULL, &data->offset1, iters);
@@ -199,6 +208,8 @@ static void run_ecmult_bench(bench_data* data, int iters) {
     char str[32];
     sprintf(str, "ecmult_gen");
     run_benchmark(str, bench_ecmult_gen, bench_ecmult_setup, bench_ecmult_gen_teardown, data, 10, iters);
+    sprintf(str, "ecmult_gen_var");
+    run_benchmark(str, bench_ecmult_gen_var, bench_ecmult_setup, bench_ecmult_gen_teardown, data, 10, iters);
     sprintf(str, "ecmult_const");
     run_benchmark(str, bench_ecmult_const, bench_ecmult_setup, bench_ecmult_const_teardown, data, 10, iters);
     sprintf(str, "ecmult_const_xonly");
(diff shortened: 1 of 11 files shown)


real-or-random:
Concept ACK

This was discussed in [yesterday's IRC meeting](https://gnusha.org/secp256k1/2026-08-24.log):

> 08:30 < real_or_random> I tend to think that's a bit of a hopeless case. When josie brought this up first, my initial reaction was "uh, let's stay safe here". But now I'm leaning towards the opposite. Any meaningful caller will branch on whether a match was found or not, with much more timing differences than we can ever generate in the raw crypto code. So getting a 25% speedup seems worth the deal
> [...] 
>08:36 < sipa> real_or_random: seems reasonable

theStack:
> Concept ACK
> 
> This was discussed in [yesterday's IRC meeting](https://gnusha.org/secp256k1/2026-08-24.log):
> 
> > 08:30 < real_or_random> I tend to think that's a bit of a hopeless case. When josie brought this up first, my initial reaction was "uh, let's stay safe here". But now I'm leaning towards the opposite. Any meaningful caller will branch on whether a match was found or not, with much more timing differences than we can ever generate in the raw crypto code. So getting a 25% speedup seems worth the deal
> > [...]
> > 08:36 < sipa> real_or_random: seems reasonable

I only notice now that the PR in its current state applies the same optimization also for the sending side, without giving much thought if that's also okay. The argument from the IRC meeting above is based solely on the scanning scenario ("caller will branch on whether a match was found or not"), for the sending side I assume there could theoretically still be (more of) a point treating shared_secret and t_k as actual secret?

Another slightly related follow-up question would be if at places where we treat these values not as secrets, we could (or should) then consequently also remove the memory clearing efforts, like e.g. the following:
https://github.com/bitcoin-core/secp256k1/blob/a9a61831bdcd97475e200a4bdd1ebed641f7268d/src/modules/silentpayments/main_impl.h#L705-L707

I'd say yes for consistency, though in this case there is no performance upside (I haven't verified, but I'd be surprised if this makes a noticeable difference), it just make the code overall a bit simpler by reducing LOC.

PR #1916, opened 2026-08-17
ecdh/ellswift: simplify seckey loading with `_scalar_set_b32_seckey`

theStack:
Instead of checking for overflow and zero manually, use the existing `_scalar_set_b32_seckey` helper which does both and returns zero if either of these two conditions apply. This can be seen as a very late follow-up to PR #701, commit 3fec9826086aa45ebbac1ff6fc3bb7b25ca78b1d, where the helper has been introduced and applied to the functions `_ecdsa_sign`, `_pubkey_create` and `_seckey_verify`. The variable name `is_sec_valid` has been chosen as it is currently also used in [`secp256k1_ecdsa_sign_inner`](https://github.com/bitcoin-core/secp256k1/blob/439278a649d3099d62dde966a76dc04aaca7ccb3/src/secp256k1.c#L555).

Note that introducing parantheses around `!!ret` was necessary to avoid warnings in the following form:
```
/home/thestack/secp256k1_master/src/modules/ecdh/main_impl.h: In function ‘secp256k1_ecdh’:
/home/thestack/secp256k1_master/src/modules/ecdh/main_impl.h:74:12: warning: suggest parentheses around operand of ‘!’ or change ‘&’ to ‘&&’ or ‘
!’ to ‘~’ [-Wparentheses]                                               
   74 |     return !!ret & is_sec_valid;                                                                                                         
      |            ^~~~~                                                
```

### DIFF
--- a/src/modules/ecdh/main_impl.h
+++ b/src/modules/ecdh/main_impl.h
@@ -33,7 +33,7 @@ const secp256k1_ecdh_hash_function secp256k1_ecdh_hash_function_default = ecdh_h
 
 int secp256k1_ecdh(const secp256k1_context* ctx, unsigned char *output, const secp256k1_pubkey *point, const unsigned char *scalar, secp256k1_ecdh_hash_function hashfp, void *data) {
     int ret = 0;
-    int overflow = 0;
+    int is_sec_valid;
     secp256k1_gej res;
     secp256k1_ge pt;
     secp256k1_scalar s;
@@ -46,10 +46,8 @@ int secp256k1_ecdh(const secp256k1_context* ctx, unsigned char *output, const se
     ARG_CHECK(scalar != NULL);
 
     secp256k1_pubkey_load(ctx, &pt, point);
-    secp256k1_scalar_set_b32(&s, scalar, &overflow);
-
-    overflow |= secp256k1_scalar_is_zero(&s);
-    secp256k1_scalar_cmov(&s, &secp256k1_scalar_one, overflow);
+    is_sec_valid = secp256k1_scalar_set_b32_seckey(&s, scalar);
+    secp256k1_scalar_cmov(&s, &secp256k1_scalar_one, !is_sec_valid);
 
     secp256k1_ecmult_const(&res, &pt, &s);
     secp256k1_ge_set_gej(&pt, &res);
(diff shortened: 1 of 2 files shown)


real-or-random (approved):
utACK c00dc8181054fb31a0149ce03ef8413300a3a8bf

real-or-random:
> Note that introducing parantheses around `!!ret` was necessary to avoid warnings in the following form:

Hm, indeed, but not sure why the compiler warns here. `!` still binds stronger than `&` according to https://en.cppreference.com/c/language/operator_precedence. Perhaps `is_sec_valid & !!ret` would work without parantheses (but I'm not suggesting a change).

It's probably good to keep it for consistency, but this entire business of using `&` instead of `&&` may be a little cargo cult. At least in this case, it shouldn't make a difference when it comes to constant-time compilation. We anyway don't know what the compiler will do.

theStack:
> > Note that introducing parantheses around `!!ret` was necessary to avoid warnings in the following form:
> 
> Hm, indeed, but not sure why the compiler warns here. `!` still binds stronger than `&` according to https://en.cppreference.com/c/language/operator_precedence. Perhaps `is_sec_valid & !!ret` would work without parantheses (but I'm not suggesting a change).

Was also a bit surprised about the warning. `is_sec_valid & !!ret` would indeed work without parantheses as I just verified :shrug: 

> It's probably good to keep it for consistency, but this entire business of using `&` instead of `&&` may be a little cargo cult. At least in this case, it shouldn't make a difference when it comes to constant-time compilation. We anyway don't know what the compiler will do.

Ah interesting, I honestly wasn't even aware that the idea behind using `&` instead of `&&` was constant-time reasons. Agree that it shouldn't make a difference. If we apply https://github.com/bitcoin-core/secp256k1/issues/1621 and follow the suggestion "function is constant-time only for valid inputs", these constructs will be gone anyways.

---

There are by the way two other instances where `_scalar_set_b32_seckey` could be used that I missed in this PR (due to only focusing on public API functions), in the internal silentpayment functions for calculating the `input_hash` and `t_k scalars`:
https://github.com/bitcoin-core/secp256k1/blob/a37d7cc064af8ac5be994b026f100dc51ab98765/src/modules/silentpayments/main_impl.h#L84-L85
https://github.com/bitcoin-core/secp256k1/blob/a37d7cc064af8ac5be994b026f100dc51ab98765/src/modules/silentpayments/main_impl.h#L144-L148

Would be nice to simplify there as well, on the other hand these values are not exactly "secret keys" (they just need to be in the same range, i.e. 0 < scalar < N), so not sure if that would be an improvement w.r.t. readability.

apoelstra (approved):
ACK c00dc8181054fb31a0149ce03ef8413300a3a8bf; successfully ran local tests

PR #1918, opened 2026-08-19
refactor: split `ge_parse` into explicit variants (compressed, uncompressed, uncompressed+hybrid)

theStack:
This PR is a follow-up to #1915, see the suggestion https://github.com/bitcoin-core/secp256k1/pull/1915#pullrequestreview-4962713401.

Splits the `ge_parse` function into three explicit variants for parsing from:
    - compressed public keys (33-bytes, prefixes 0x02 or 0x03)
    - uncompressed public keys (65-bytes, prefix 0x04)
    - uncompressed and hybrid public keys (65-bytes, prefixes 0x04, 0x06 or 0x07)

The first two are the counter-parts for the existing serialization functions `ge_serialize{33,65}`. The latter one is only needed for legacy reasons and likely not needed in the future. Having it named explicitly helps to avoid accepting hybrid pubkeys unintentionally. `ge_parse65` is currently unused outside of tests, but one potential use-case would be the serialization of the prevouts_summary in the silentpayments light-client API PR #1912 (see commit https://github.com/bitcoin-core/secp256k1/pull/1912/changes/82b95fde6fe2c75ec940ade339013939b8ecc5ea), unless we want to define a custom serialization there (where we e.g. remove the constant prefix byte).

### DIFF
--- a/src/group.h
+++ b/src/group.h
@@ -205,8 +205,15 @@ static void secp256k1_ge_to_bytes_ext(unsigned char *data, const secp256k1_ge *g
  *  provided buffer is the output of secp256k1_ge_to_bytes_ext. */
 static void secp256k1_ge_from_bytes_ext(secp256k1_ge *ge, const unsigned char *data);
 
-/** Parse a group element from a 33-byte compressed or 65-byte uncompressed public key. */
-static int secp256k1_ge_parse(secp256k1_ge *elem, const unsigned char *pub, size_t size);
+/** Parse a group element from a 33-byte compressed public key (prefix byte is 0x02 or 0x03). */
+static int secp256k1_ge_parse33(secp256k1_ge *elem, const unsigned char *pub);
+
+/** Parse a group element from a 65-byte uncompressed public key (prefix byte is 0x04). */
+static int secp256k1_ge_parse65(secp256k1_ge *elem, const unsigned char *pub);
+
+/** Parse a group element from a 65-byte uncompressed or hybrid public key (prefix byte is 0x04, 0x06 or 0x07).
+ *  Only intended for legacy public APIs that must accept hybrid keys. */
+static int secp256k1_ge_parse_with_hybrid65(secp256k1_ge *elem, const unsigned char *pub);
 
 /** Serialize a group element (that is not allowed to be infinity) to a compressed public key (33 bytes). */
 static void secp256k1_ge_serialize33(secp256k1_ge *elem, unsigned char *pub33);
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -263,12 +263,18 @@ static void secp256k1_pubkey_save(secp256k1_pubkey* pubkey, secp256k1_ge* ge) {
 
 int secp256k1_ec_pubkey_parse(const secp256k1_context* ctx, secp256k1_pubkey* pubkey, const unsigned char *input, size_t inputlen) {
     secp256k1_ge Q;
+    int is_pubkey_valid;
 
     VERIFY_CHECK(ctx != NULL);
     ARG_CHECK(pubkey != NULL);
     memset(pubkey, 0, sizeof(*pubkey));
     ARG_CHECK(input != NULL);
-    if (!secp256k1_ge_parse(&Q, input, inputlen)) {
+    switch (inputlen) {
+        case 33: is_pubkey_valid = secp256k1_ge_parse33(&Q, input); break;
+        case 65: is_pubkey_valid = secp256k1_ge_parse_with_hybrid65(&Q, input); break;
+        default: is_pubkey_valid = 0;
+    }
+    if (!is_pubkey_valid) {
         return 0;
     }
     if (!secp256k1_ge_is_in_correct_subgroup(&Q)) {
--- a/src/tests.c
+++ b/src/tests.c
@@ -6258,6 +6258,12 @@ static void run_ec_pubkey_parse_test(void) {
         0xA8, 0xFD, 0x17, 0xB4, 0x48, 0xA6, 0x85, 0x54, 0x19, 0x9C, 0x47, 0xD0, 0x8F, 0xFB, 0x10, 0xD4,
         0xB8, 0x00
     };
+    const unsigned char pubkeyc_comp[34] = {
+        /* Compressed serialization of G (y is even, so prefix is 0x02). */
+        0x02, 0x79, 0xBE, 0x66, 0x7E, 0xF9, 0xDC, 0xBB, 0xAC, 0x55, 0xA0, 0x62, 0x95, 0xCE, 0x87, 0x0B,
+        0x07, 0x02, 0x9B, 0xFC, 0xDB, 0x2D, 0xCE, 0x28, 0xD9, 0x59, 0xF2, 0x81, 0x5B, 0x16, 0xF8, 0x17,
+        0x98, 0x00
+    };
     unsigned char sout[65];
     unsigned char shortkey[2] = { 0 };
     secp256k1_ge ge;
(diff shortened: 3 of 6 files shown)


src/secp256k1.c
```diff
@@ -267,12 +267,18 @@ static void secp256k1_pubkey_save(secp256k1_pubkey* pubkey, secp256k1_ge* ge) {
 
 int secp256k1_ec_pubkey_parse(const secp256k1_context* ctx, secp256k1_pubkey* pubkey, const unsigned char *input, size_t inputlen) {
     secp256k1_ge Q;
+    int res;
```
fjahr: super nit: I think it's usually `ret`?

src/secp256k1.c
```diff
@@ -267,12 +267,18 @@ static void secp256k1_pubkey_save(secp256k1_pubkey* pubkey, secp256k1_ge* ge) {
...
 
     VERIFY_CHECK(ctx != NULL);
     ARG_CHECK(pubkey != NULL);
     memset(pubkey, 0, sizeof(*pubkey));
     ARG_CHECK(input != NULL);
-    if (!secp256k1_ge_parse(&Q, input, inputlen)) {
+    switch (inputlen) {
+        case 33: res = secp256k1_ge_parse33(&Q, input); break;
```
fjahr: I like this concise style, just wondering why it's never been used anywhere outside of test code. I guess there is no particular reason?

src/tests.c
```diff
@@ -6840,6 +6840,7 @@ static void test_random_pubkeys(void) {
     secp256k1_ge elem;
     secp256k1_ge elem2;
     unsigned char in[65];
+    int res;
     /* Generate some randomly sized pubkeys. */
     size_t len = testrand_bits(2) == 0 ? 65 : 33;
     if (testrand_bits(2) == 0) {
```
fjahr: With the switch statement added further below I think this randomness here doesn't make sense anymore. If `len` isn't 33 or 65 we just end up hitting the default case there and we are not even calling any library function in the test then.

fjahr (commented):
Concept ACK

src/secp256k1.c
theStack: Seems that in API functions we use `ret` almost exclusively for the to-be-returned value (i.e. paired with a `return ret;` at the bottom), which doesn't apply here. `res` is not a great name either, decided to be explicit and name it `is_pubkey_valid` (inspired by existing `is_sec_valid` instances in the codebase).

src/secp256k1.c
theStack: Fair question, maybe there was no good use-case for it so far, but not sure. Seems there is one switch/case statement in `secp256k1_pippenger_bucket_window_inv`, so there is at least no (unwritten) rule to completely avoid it in (potential) production code. No strong opinion, also happy to change to if/else if other reviewers feel strongly.

In the test I've changed to using a ternary expression, since the default branch is not needed anymore after your suggestion below to limit to 33/65 byte-sized pubkey serialization, so the switch/case became a bit less compelling.

src/tests.c
theStack: Good point, pretty pointless now to generate differently-sized serializations and not do anything with them indeed. Adapted and simplified the test accordingly.

theStack:
Thanks for the detailed review @fjahr, addressed your suggestion of simplifying/adapting the test and renamed the res variable in the touched pubkey parsing API function.

src/group.h
```diff
@@ -196,8 +196,14 @@ static void secp256k1_ge_to_bytes_ext(unsigned char *data, const secp256k1_ge *g
...
 
-/** Parse a group element from a 33-byte compressed or 65-byte uncompressed public key. */
-static int secp256k1_ge_parse(secp256k1_ge *elem, const unsigned char *pub, size_t size);
+/** Parse a group element from a 33-byte compressed public key (prefix byte is 0x02 or 0x03). */
+static int secp256k1_ge_parse33(secp256k1_ge *elem, const unsigned char *pub);
+
+/** Parse a group element from a 65-byte uncompressed public key (prefix byte is 0x04). */
+static int secp256k1_ge_parse65(secp256k1_ge *elem, const unsigned char *pub);
+
+/** Parse a group element from a 65-byte uncompressed or hybrid public key (prefix byte is 0x04, 0x06 or 0x07). */
```
fjahr, line 205: nit: Could have mentioned here in the comment as well what's in the commit description, that this still exists for legacy reasons and further usage is discouraged. But feel free to ignore unless you have to retouch.

fjahr (commented):
ACK 42dd54e46e4223660fa0f6dedb0d4e3f6b0ea152

src/group.h
theStack: Good idea, will tackle that on the next push (I expect #1923 to go in first, so I'll have to rebase anyways).

theStack:
Rebased on master and addressed https://github.com/bitcoin-core/secp256k1/pull/1918#discussion_r3819523047.

src/group.h
```diff
@@ -205,8 +205,15 @@ static void secp256k1_ge_to_bytes_ext(unsigned char *data, const secp256k1_ge *g
...
-/** Parse a group element from a 33-byte compressed or 65-byte uncompressed public key. */
-static int secp256k1_ge_parse(secp256k1_ge *elem, const unsigned char *pub, size_t size);
+/** Parse a group element from a 33-byte compressed public key (prefix byte is 0x02 or 0x03). */
+static int secp256k1_ge_parse33(secp256k1_ge *elem, const unsigned char *pub);
+
+/** Parse a group element from a 65-byte uncompressed public key (prefix byte is 0x04). */
+static int secp256k1_ge_parse65(secp256k1_ge *elem, const unsigned char *pub);
+
+/** Parse a group element from a 65-byte uncompressed or hybrid public key (prefix byte is 0x04, 0x06 or 0x07).
+ *  Only intended for legacy public APIs that must accept hybrid keys; new code should use secp256k1_ge_parse65. */
```
real-or-random:
```suggestion
 *  Only intended for legacy public APIs that must accept hybrid keys. */
```
New code should use ge_parse66, ge_parse33, or xonly even. It depends on the use case, and I don't think we should comment on that here.

src/tests.c
```diff
@@ -6852,11 +6852,9 @@ static void test_random_pubkeys(void) {
...
     secp256k1_ge elem2;
     unsigned char in[65];
-    /* Generate some randomly sized pubkeys. */
-    size_t len = testrand_bits(2) == 0 ? 65 : 33;
-    if (testrand_bits(2) == 0) {
-        len = testrand_bits(6);
-    }
+    int res;
+    /* Generate some random pubkeys with the two supported serialization sizes. */
+    size_t len = testrand_bits(1) == 0 ? 65 : 33;
```
real-or-random, line 6904: IIUC `run_ec_pubkey_parse_test` currently checks 64 and 66 as edge cases but not 32 and 34. I think these should be added.

real-or-random (commented):
ACK mod nits

src/group.h
theStack: Makes sense yeah, removed that part.

src/tests.c
theStack: Added a commit on top that adds coverage for compressed pubkey parsing/serialization, including the 32 and 34 bytes edge cases, in a similar style as how it's already done for uncompressed pubkeys. It could be even more thorough I guess (e.g. by also testing with the 0x03 prefix, not only 0x02), maybe something for a follow-up.

fjahr: nit: Feels like a helper could have deduplicated these edge case tests quite a bit, but not a blocker for me. Basically this (untested):

```
static void ec_pubkey_parse_expect_fail(const unsigned char *input, size_t inputlen) {
    secp256k1_pubkey pubkey;
    secp256k1_ge ge;
    memset(&pubkey, 0xfe, sizeof(pubkey));
    SECP256K1_CHECKMEM_UNDEFINE(&pubkey, sizeof(pubkey));
    CHECK(secp256k1_ec_pubkey_parse(CTX, &pubkey, input, inputlen) == 0);
    SECP256K1_CHECKMEM_CHECK(&pubkey, sizeof(pubkey));
    CHECK_ILLEGAL(CTX, secp256k1_pubkey_load(CTX, &ge, &pubkey));
}
```

fjahr:
Code review ACK de02108c51ca052245c7ba154b02a890e6815fdb

real-or-random (approved):
utACK de02108c51ca052245c7ba154b02a890e6815fdb

PR #1919, opened 2026-08-20
ecdh/ellswift/schnorrsig: return early for invalid inputs (consistent handling w.r.t. constant-time)

theStack:
This PR partly addresses issue #1621, handling invalid inputs consistently w.r.t. constant-time, by applying the suggested "functions are constant-time only for valid inputs" approach (number 2. in the issue), for the following modules and API functions:
* ecdh: `secp256k1_ecdh`
* ellswift: `secp256k1_ellswift_create` and `secp256k1_ellswift_xdh`
* schnorrsig: `secp256k1_schnorrsig_sign{32,_custom}`

As stated in #1621, this leads to more readable and maintainable code, getting rid of constructs like `return (!!ret) & is_sec_valid;` or intermediate `ret &= ...;` statements. Conditional assignments like `secp256k1_scalar_cmov` can be removed as well.

Tests are added for the newly introduced branches if there were none. There are still further instances to tackle (mostly the main module API functions in secp256k1.c, and potential cleanups like e.g. removing the dummy assignments in `_keypair_load`), happy to add these here or in another PR, mostly chasing Conecpt ACKs for now if this is we want to handle constant-time behavior for current and future modules.

### DIFF
--- a/src/modules/ecdh/main_impl.h
+++ b/src/modules/ecdh/main_impl.h
@@ -45,9 +45,15 @@ int secp256k1_ecdh(const secp256k1_context* ctx, unsigned char *output, const se
     ARG_CHECK(point != NULL);
     ARG_CHECK(scalar != NULL);
 
-    secp256k1_pubkey_load(ctx, &pt, point);
+    if (!secp256k1_pubkey_load(ctx, &pt, point)) {
+        return 0;
+    }
     is_sec_valid = secp256k1_scalar_set_b32_seckey(&s, scalar);
-    secp256k1_scalar_cmov(&s, &secp256k1_scalar_one, !is_sec_valid);
+    secp256k1_declassify(ctx, &is_sec_valid, sizeof(is_sec_valid));
+    if (!is_sec_valid) {
+        secp256k1_ge_clear(&pt);
+        return 0;
+    }
 
     secp256k1_ecmult_const(&res, &pt, &s);
     secp256k1_ge_set_gej(&pt, &res);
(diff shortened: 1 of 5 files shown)


src/modules/ecdh/main_impl.h
```diff
@@ -45,9 +45,15 @@ int secp256k1_ecdh(const secp256k1_context* ctx, unsigned char *output, const se
...
+    if (!secp256k1_pubkey_load(ctx, &pt, point)) {
+        return 0;
+    }
     is_sec_valid = secp256k1_scalar_set_b32_seckey(&s, scalar);
-    secp256k1_scalar_cmov(&s, &secp256k1_scalar_one, !is_sec_valid);
+    secp256k1_declassify(ctx, &is_sec_valid, sizeof(is_sec_valid));
+    if (!is_sec_valid) {
+        secp256k1_ge_clear(&pt);
+        return 0;
+    }
```
IsaqueFranklin, line 56: Should we also clear `s` here? I understand that since we already know that `s` is not a valid secret key, it really doesn't matter that much. However, originally the invalid input path would reach the end of the function and `s` would be cleared anyway.

theStack: Good question. I'd argue that there is not much value in protecting invalid key material from leaking (see also recent review comment https://github.com/bitcoin-core/secp256k1/pull/1802#discussion_r3853257606), but being conservative and still clearing it is of course also an option. Curious what other reviewers think.

PR #1921, opened 2026-08-24
tests: correct scalar split bound vectors

### LINKED ISSUE #1920: scalars_near_split_bounds in tests.c don't match their documented construction
---

The comment above `scalars_near_split_bounds[20]` in `src/tests.c` (array at line 4648,
comment ending line 4646 on master) says the values are generated as

    (a*LAMBDA + (ORDER+b)/2) % ORDER,  a in {-2,-1,0,1,2},  b in {-3,-1,1,3}

The array below it doesn't match. None of the 20 entries equal any value of that formula.

Example, a=0 b=-3, i.e. (ORDER-3)/2:

    (ORDER-3)/2 = 0x7fffffffffffffffffffffffffffffff5d576e7357a4501ddfe92f46681b209f
    array entry = 0x7fffffffffffffffffffffffffffffffd576e73557a4501ddfe92f46681b209f

They differ in bits 96-127, and every entry is off in that same 32-bit window.

Because of this the scalars don't reach the split bounds they're named for. Through
secp256k1_scalar_split_lambda they top out around 2^126.2 (|k1|) and 2^126.5 (|k2|),
whereas the actual maxima are 2^127.35 and 2^127.11. test_scalar_split still passes and
there's no security impact, but the 128-bit boundary these vectors exist to cover is never
exercised.

Repro, needs only the order and lambda:

    n   = 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFEBAAEDCE6AF48A03BBFD25E8CD0364141
    lam = 0x5363AD4CC05C30E0A5261C028812645A122E22EA20816678DF02967C1B23BD72
    arr = [
     0xd938a5667f479e3eb5b3c7faefdb37493aa0585cc5ea2367e1b660db0209e6fc,
     0xd938a5667f479e3eb5b3c7faefdb37493aa0585cc5ea2367e1b660db0209e6fd,
     0xd938a5667f479e3eb5b3c7faefdb37493aa0585cc5ea2367e1b660db0209e6fe,
     0xd938a5667f479e3eb5b3c7faefdb37493aa0585cc5ea2367e1b660db0209e6ff,
     0x2c9c52b33fa3cf1f5ad9e3fd77ed9ba5b294b8933722e9a500e698ca4cf7632d,
     0x2c9c52b33fa3cf1f5ad9e3fd77ed9ba5b294b8933722e9a500e698ca4cf7632e,
     0x2c9c52b33fa3cf1f5ad9e3fd77ed9ba5b294b8933722e9a500e698ca4cf7632f,
     0x2c9c52b33fa3cf1f5ad9e3fd77ed9ba5b294b8933722e9a500e698ca4cf76330,
     0x7fffffffffffffffffffffffffffffffd576e73557a4501ddfe92f46681b209f,
     0x7fffffffffffffffffffffffffffffffd576e73557a4501ddfe92f46681b20a0,
     0x7fffffffffffffffffffffffffffffffd576e73557a4501ddfe92f46681b20a1,
     0x7fffffffffffffffffffffffffffffffd576e73557a4501ddfe92f46681b20a2,
     0xd363ad4cc05c30e0a5261c0288126459f85915d77825b696beebc5c2833ede11,
     0xd363ad4cc05c30e0a5261c0288126459f85915d77825b696beebc5c2833ede12,
     0xd363ad4cc05c30e0a5261c0288126459f85915d77825b696beebc5c2833ede13,
     0xd363ad4cc05c30e0a5261c0288126459f85915d77825b696beebc5c2833ede14,
     0x26c75a9980b861c14a4c38051024c8b4704d760ee95e7cd3de1bfdb1ce2c5a42,
     0x26c75a9980b861c14a4c38051024c8b4704d760ee95e7cd3de1bfdb1ce2c5a43,
     0x26c75a9980b861c14a4c38051024c8b4704d760ee95e7cd3de1bfdb1ce2c5a44,
     0x26c75a9980b861c14a4c38051024c8b4704d760ee95e7cd3de1bfdb1ce2c5a45,
    ]
    formula = {(a*lam + (n+b)//2) % n for a in (-2,-1,0,1,2) for b in (-3,-1,1,3)}
    print(sum(v in formula for v in arr), "/ 20 match")   # prints 0 / 20

Fix: regenerate the array from the formula. The b range {-3,-1,1,3} also never hits the
|k2| maximum on its own; that scalar is -LAMBDA + (ORDER-11)/2 (b = -11), so the range
needs widening to cover both bounds. The two exact extremes:

    max|k1|:  2*LAMBDA + (ORDER-1)/2  = 0x26c75a9980b861c14a4c38051024c8b4c704d760e95e7cd3de1bfdb1ce2c5a43
    max|k2|:  -LAMBDA + (ORDER-11)/2  = 0x2c9c52b33fa3cf1f5ad9e3fd77ed9ba54b294b893722e9a500e698ca4cf76329

[repro_testvectors.py](https://github.com/user-attachments/files/31329477/repro_testvectors.py)

l0rinc:
The 20 scalars used to exercise `secp256k1_scalar_split_lambda` near its 128-bit bounds do not match their documented construction. One 32-bit limb is shifted in every entry, so the resulting split components remain around 126 bits and miss the cases this table is intended to cover.

This regenerates the constants from `(a*LAMBDA + (ORDER+b)/2) mod ORDER` and checks every literal against that formula at runtime. The corrected existing input set reaches the maximum absolute values of both split components, so no additional inputs are needed.

Fixes #1920.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -4646,27 +4646,52 @@ static void test_point_times_order(const secp256k1_gej *point) {
  *     - Output (a*LAMBDA + (ORDER+b)/2) % ORDER
  */
 static const secp256k1_scalar scalars_near_split_bounds[20] = {
-    SECP256K1_SCALAR_CONST(0xd938a566, 0x7f479e3e, 0xb5b3c7fa, 0xefdb3749, 0x3aa0585c, 0xc5ea2367, 0xe1b660db, 0x0209e6fc),
-    SECP256K1_SCALAR_CONST(0xd938a566, 0x7f479e3e, 0xb5b3c7fa, 0xefdb3749, 0x3aa0585c, 0xc5ea2367, 0xe1b660db, 0x0209e6fd),
-    SECP256K1_SCALAR_CONST(0xd938a566, 0x7f479e3e, 0xb5b3c7fa, 0xefdb3749, 0x3aa0585c, 0xc5ea2367, 0xe1b660db, 0x0209e6fe),
-    SECP256K1_SCALAR_CONST(0xd938a566, 0x7f479e3e, 0xb5b3c7fa, 0xefdb3749, 0x3aa0585c, 0xc5ea2367, 0xe1b660db, 0x0209e6ff),
-    SECP256K1_SCALAR_CONST(0x2c9c52b3, 0x3fa3cf1f, 0x5ad9e3fd, 0x77ed9ba5, 0xb294b893, 0x3722e9a5, 0x00e698ca, 0x4cf7632d),
-    SECP256K1_SCALAR_CONST(0x2c9c52b3, 0x3fa3cf1f, 0x5ad9e3fd, 0x77ed9ba5, 0xb294b893, 0x3722e9a5, 0x00e698ca, 0x4cf7632e),
-    SECP256K1_SCALAR_CONST(0x2c9c52b3, 0x3fa3cf1f, 0x5ad9e3fd, 0x77ed9ba5, 0xb294b893, 0x3722e9a5, 0x00e698ca, 0x4cf7632f),
-    SECP256K1_SCALAR_CONST(0x2c9c52b3, 0x3fa3cf1f, 0x5ad9e3fd, 0x77ed9ba5, 0xb294b893, 0x3722e9a5, 0x00e698ca, 0x4cf76330),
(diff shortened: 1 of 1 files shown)


l0rinc:
This isn't ready yet, meant to push it elsewhere

real-or-random:
This is fine in its essence, but if we now have a function that generates the vectors, then we could just keep that one and drop the hardcoded values.

PR #1345, opened 2023-06-13
field: Static-assert that int args affecting magnitude are constant

real-or-random:
See #1001.

PR #1922, opened 2026-08-24
field: Check that argument of _fe_set_int() is a constant

real-or-random:
Solves part of https://github.com/bitcoin-core/secp256k1/issues/1001. Overlooked in https://github.com/bitcoin-core/secp256k1/pull/1345.


theStack (approved):
ACK f6d821db032d494daf3bcfe56c0b3e711c06973f
(under the assumption that the remaining CI jobs will also be green)

PR #1923, opened 2026-08-25
group: VERIFY input/output ge/gej/fe exhaustively

real-or-random:
This PR splits the functions in group_impl.h in a wrapper that only performs VERIFY and an `_impl` function that has the actual code. This ensures that the post VERIFY calls are not skipped in case of early returns.

Also, this PR adds VERIFY calls wherever they were missing (inside group_impl.h)

The new structure is similar to field_impl.h but a bit simpler because we don't need to deal with two different implementations. A real difference is that, in non-VERIFY mode, field_impl.h delegates via #defines (ensure there's no overhead due to a function call) and here I decided to delegate via function calls. It keeps the code a bit simpler to read (and maybe also simpler to parser for tools such as language servers). The _impl functions all have SECP256K1_INLINE. I think every sane compiler will inline the function calls in non-VERIFY mode (even without SECP256K1_INLINE) because the body of the wrapper is really just a single function call then with the same signature.

Follow-up PR can cover modifications of ges and fes outside the group and field modules, e.g., ecmult modifies ges/gejs directly. Maybe it will be good that it does this only through group functions but we'll need to see; in C++ this module could legitimately be considered a "friend" of group.

### DIFF
--- a/src/group.h
+++ b/src/group.h
@@ -87,6 +87,15 @@ static void secp256k1_ge_set_all_gej(secp256k1_ge *r, const secp256k1_gej *a, si
 /** Set group elements r[0:len] (affine) equal to group elements a[0:len] (jacobian). */
 static void secp256k1_ge_set_all_gej_var(secp256k1_ge *r, const secp256k1_gej *a, size_t len);
 
+
+/** Set r to the affine coordinates of the Jacobian point (a.x, a.y, 1/zi).
+ *  a must not be infinity. */
+static void secp256k1_ge_set_ge_zinv(secp256k1_ge *r, const secp256k1_ge *a, const secp256k1_fe *zi);
+
+/** Set r to the affine coordinates of the Jacobian point (a.x, a.y, 1/zi), ignoring a.z.
+ *  a must not be infinity. */
+static void secp256k1_ge_set_gej_zinv(secp256k1_ge *r, const secp256k1_gej *a, const secp256k1_fe *zi);
+
 /** Bring a batch of inputs to the same global z "denominator", based on ratios between
  *  (omitted) z coordinates of adjacent elements.
  *
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -702,7 +702,6 @@ SECP256K1_INLINE static void secp256k1_gej_impl_add_zinv_var(secp256k1_gej *r, c
         secp256k1_fe_mul(&r->x, &b->x, &bzinv2);
         secp256k1_fe_mul(&r->y, &b->y, &bzinv3);
         secp256k1_fe_set_int(&r->z, 1);
-        SECP256K1_GEJ_VERIFY(r);
         return;
     }
     if (b->infinity) {
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -104,7 +104,7 @@ SECP256K1_INLINE static void secp256k1_ge_impl_set_gej_zinv(secp256k1_ge *r, con
     secp256k1_fe_mul(&zi3, &zi2, zi);
     secp256k1_fe_mul(&r->x, &a->x, &zi2);
     secp256k1_fe_mul(&r->y, &a->y, &zi3);
-    r->infinity = a->infinity;
+    r->infinity = 0;
 }
 static void secp256k1_ge_set_gej_zinv(secp256k1_ge *r, const secp256k1_gej *a, const secp256k1_fe *zi) {
     SECP256K1_GEJ_VERIFY(a); SECP256K1_FE_VERIFY(zi);
@@ -121,7 +121,7 @@ SECP256K1_INLINE static void secp256k1_ge_impl_set_ge_zinv(secp256k1_ge *r, cons
     secp256k1_fe_mul(&zi3, &zi2, zi);
     secp256k1_fe_mul(&r->x, &a->x, &zi2);
     secp256k1_fe_mul(&r->y, &a->y, &zi3);
-    r->infinity = a->infinity;
+    r->infinity = 0;
 }
 static void secp256k1_ge_set_ge_zinv(secp256k1_ge *r, const secp256k1_ge *a, const secp256k1_fe *zi) {
     SECP256K1_GE_VERIFY(a); SECP256K1_FE_VERIFY(zi);
--- a/src/group_impl.h
+++ b/src/group_impl.h
@@ -95,7 +95,6 @@ static void secp256k1_gej_verify(const secp256k1_gej *a) {
     (void)a;
 }
 
-/* Set r to the affine coordinates of Jacobian point (a.x, a.y, 1/zi). */
 SECP256K1_INLINE static void secp256k1_ge_impl_set_gej_zinv(secp256k1_ge *r, const secp256k1_gej *a, const secp256k1_fe *zi) {
     secp256k1_fe zi2;
     secp256k1_fe zi3;
@@ -113,7 +112,6 @@ static void secp256k1_ge_set_gej_zinv(secp256k1_ge *r, const secp256k1_gej *a, c
     SECP256K1_GE_VERIFY(r);
 }
 
-/* Set r to the affine coordinates of Jacobian point (a.x, a.y, 1/zi). */
 SECP256K1_INLINE static void secp256k1_ge_impl_set_ge_zinv(secp256k1_ge *r, const secp256k1_ge *a, const secp256k1_fe *zi) {
     secp256k1_fe zi2;
     secp256k1_fe zi3;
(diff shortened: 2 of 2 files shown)


theStack:
Concept ACK

Out of curiosity I tried out if following a "no early returns" policy in the group module instead could be a viable alternative to splitting up functions, but as expected it's quite ugly overall: https://github.com/theStack/secp256k1/commit/2f818928543f9dd8c78cd2c2a3af3f0ee589fb02 (and introducing `goto` would probably be very controversial in general, even though I personally think usage in some "common cleanup" scenarios it's okay).

> The _impl functions all have SECP256K1_INLINE. I think every sane compiler will inline the function calls in non-VERIFY mode (even without SECP256K1_INLINE) because the body of the wrapper is really just a single function call then with the same signature.

That would also be my guess. If we are worried about that, using the recently introduced SECP256K1_FORCE_INLINE macro could maybe be an option.

real-or-random:
> Out of curiosity I tried out if following a "no early returns" policy in the group module instead could be a viable alternative to splitting up functions, but as expected it's quite ugly overall: [theStack@2f81892](https://github.com/theStack/secp256k1/commit/2f818928543f9dd8c78cd2c2a3af3f0ee589fb02) (and introducing `goto` would probably be very controversial in general, even though I personally think usage in some "common cleanup" scenarios it's okay).

Yeah, `goto` was my first attempt, I personally don't think `goto` is evil. If used carefully, it certainly cleaner code than using `if`/`else` to eliminate early returns.

But I think the wrapping approach in this PR is conceptually clean, avoids the `goto` debate and it matches what we do in the field module, so I picked that one.

src/group_impl.h
```diff
@@ -1093,15 +1093,22 @@ static int secp256k1_ge_is_in_correct_subgroup(const secp256k1_ge* ge) {
     return secp256k1_ge_impl_is_in_correct_subgroup(ge);
 }
 
-static int secp256k1_ge_x_on_curve_var(const secp256k1_fe *x) {
+static int secp256k1_ge_impl_x_on_curve_var(const secp256k1_fe *x) {
```
theStack: looks like this is the only instance of a `_{ge,gej}_impl_...` function that doesn't have the `_INLINE` keyword, is that intentional?

src/group.h
```diff
@@ -55,6 +55,9 @@ typedef struct {
 /** Set a group element equal to the point with given X and Y coordinates */
 static void secp256k1_ge_set_xy(secp256k1_ge *r, const secp256k1_fe *x, const secp256k1_fe *y);
 
+/** Set r to the affine coordinates of the Jacobian point (a.x, a.y, 1/zi). */
+static void secp256k1_ge_set_ge_zinv(secp256k1_ge *r, const secp256k1_ge *a, const secp256k1_fe *zi);
```
theStack: the same could be done for `secp256k1_ge_set_gej_zinv` (probably even more relevant as that one is called outside of `group_impl.h` [[in `ecmult_impl.h`](https://github.com/bitcoin-core/secp256k1/blob/1c8babcd6c76dbea50bf4d65468dd8088d1b27a6/src/ecmult_impl.h#L96)], while `secp256k1_ge_set_ge_zinv` is currently not)

theStack (commented):
LGTM, will check tomorrow if there is any size difference of the library binary between master and this PR (which would indicate whether the _impl functions are actually inlined or not).

src/group.h
real-or-random: fixed

src/group_impl.h
real-or-random: fixed

src/group.h
```diff
@@ -87,6 +87,15 @@ static void secp256k1_ge_set_all_gej(secp256k1_ge *r, const secp256k1_gej *a, si
 /** Set group elements r[0:len] (affine) equal to group elements a[0:len] (jacobian). */
 static void secp256k1_ge_set_all_gej_var(secp256k1_ge *r, const secp256k1_gej *a, size_t len);
 
+
+/** Set r to the affine coordinates of the Jacobian point (a.x, a.y, 1/zi).
+ *  If a is infinity, then r will be infinity. */
+static void secp256k1_ge_set_ge_zinv(secp256k1_ge *r, const secp256k1_ge *a, const secp256k1_fe *zi);
+
+/** Set r to the affine coordinates of the Jacobian point (a.x, a.y, 1/zi), ignoring a.z.
+ *  If a is infinity, then r will be infinity. */
```
theStack, line 96: for both `_ge_set_{ge,gej}_zinv`: according to the precondition checks `a` is not allowed to be infinity (we could replace `r->infinity = a->infinity` with `r->infinity = 0` accordingly I guess)

src/group_impl.h
```diff
@@ -1120,35 +1127,60 @@ static int secp256k1_ge_x_frac_on_curve_var(const secp256k1_fe *xn, const secp25
      return secp256k1_fe_is_square_var(&r);
 }
 
-static void secp256k1_ge_to_bytes(unsigned char *buf, const secp256k1_ge *a) {
+static int secp256k1_ge_x_frac_on_curve_var(const secp256k1_fe *xn, const secp256k1_fe *xd) {
+    SECP256K1_FE_VERIFY(xn);
+    SECP256K1_FE_VERIFY(xd);
```
theStack, line 1132: could move the `VERIFY_CHECK(!secp256k1_fe_normalizes_to_zero_var(xd));` line from the _impl function to here

real-or-random: Yeah, this was the style I was trying to follow (like in field the same). But now I'm wondering if it's the best one. 

Wouldn't it make more readable to have these "semantic" conditions in the `_impl` function because this is where our brains need them for understanding? _edit: Me overlooking the VERIFY_CHECK in `_ge_set_{ge,gej}_zinv` (your previous comment) somewhat proves this._

(The setting in `field` is a bit different: there are two different implementations but they share the preconditions, so it makes sense to check them in the wrapper.)  

Also thinking about saving some lines:
 - Removing the blank line between `_impl` and wrapper to make them visually appear together.
 - Removing the blank lines within `_impl`.
 - Keeping all inputs on one line and all outputs on one line.

Like this:

```c
SECP256K1_INLINE static void secp256k1_gej_impl_cmov(secp256k1_gej *r, const secp256k1_gej *a, int flag) {
    VERIFY_CHECK(flag == 0 || flag == 1);
    secp256k1_fe_cmov(&r->x, &a->x, flag);
    secp256k1_fe_cmov(&r->y, &a->y, flag);
    secp256k1_fe_cmov(&r->z, &a->z, flag);
    r->infinity ^= (r->infinity ^ a->infinity) & flag;
}
SECP256K1_INLINE static void secp256k1_gej_cmov(secp256k1_gej *r, const secp256k1_gej *a, int flag) {
    SECP256K1_GEJ_VERIFY(r); SECP256K1_GEJ_VERIFY(a);
    secp256k1_gej_impl_cmov(r, a, flag);
    SECP256K1_GEJ_VERIFY(r);
}
```

What do you think?

src/group.h
real-or-random: Oh indeed, will fix.

src/group_impl.h
theStack: > Wouldn't it make more readable to have these "semantic" conditions in the `_impl` function because this is where our brains need them for understanding? _edit: Me overlooking the VERIFY_CHECK in `_ge_set_{ge,gej}_zinv` (your previous comment) somewhat proves this._

Good point. So the idea then is to only use the `_{FE,GE,GEJ}_VERIFY` (depending on the types used in the in and out parameters) checks in the wrappers which mostly checks magnitude conditions and the "semantic" conditions would be in the `_impl` functions, right?

> Also thinking about saving some lines:
> ...
> What do you think?

Nice idea, looks good to me; I think I wouldn't mind still having blank lines in the _impl functions (at least from a ratio point of view they are less wasteful than in the wrappers :p), but either way seems fine.

real-or-random: > I think I wouldn't mind still having blank lines in the _impl functions

ah sorry, yes, I meant removing the blank lines in the wrappers (as in my example). will implement.

apoelstra (approved):
ACK af1c96918f7355ae267f5f997378dd063505a3a4; successfully ran local tests

src/group.h
real-or-random: added a commit

src/group_impl.h
real-or-random: done

real-or-random:
forced-pushed, diff: https://github.com/bitcoin-core/secp256k1/compare/af1c96918f7355ae267f5f997378dd063505a3a4..c36269ef95da99498b3b62736f09dc42ffe5d094

ready from my side

theStack (approved):
ACK c36269ef95da99498b3b62736f09dc42ffe5d094

PR #1924, opened 2026-08-26
tests: add coverage for the DER long form length encoding

brunoerg:
random_ber_signature only emits long form lengths in signatures it marks as certainly_not_der, so run_ecdsa_der_parse asserts that such inputs are rejected but never that a well-formed long form length is accepted. As a result, inverting any of the four conditions in secp256k1_der_read_len that govern the long form goes unnoticed by the test suite.

Add the accepting cases, together with the two rejecting ones that share their boundaries: a length below 128, which must use the short form, and length octets with a leading zero. Since DER permits the long form only for lengths of at least 128, which a signature with two in-range scalars never reaches, the R integers in the accepting cases necessarily overflow to zero; the test checks the parsed values so this stays pinned down.

I noticed this lack in the tests while working on mutation testing for the ecdsa impl.  So these tests kill the following mutants: https://secp256k1.space/src/ecdsa_impl.h#863, https://secp256k1.space/src/ecdsa_impl.h#878 and https://secp256k1.space/src/ecdsa_impl.h#896

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -7388,6 +7388,122 @@ static void run_ecdsa_der_parse(void) {
     }
 }
 
+/* Appends the body of a signature holding a 122-byte R integer, which exceeds
+ * 32 bytes and therefore overflows to zero, and the 2-byte S integer 0x0123.
+ * The body is exactly 128 bytes long, the smallest length whose encoding
+ * requires the long form. */
+static size_t der_long_form_body(unsigned char *buf) {
+    size_t len = 0;
+    size_t i;
+    buf[len++] = 0x02;
+    buf[len++] = 0x7A;
+    for (i = 0; i < 0x7A; i++) {
+        buf[len++] = 0x01;
+    }
+    buf[len++] = 0x02;
+    buf[len++] = 0x02;
+    buf[len++] = 0x01;
+    buf[len++] = 0x23;
+    CHECK(len == 128);
+    return len;
+}
+
+/* Appends the 35-byte encoding of an INTEGER holding a zero pad followed by 32
+ * bytes whose top bit is set. */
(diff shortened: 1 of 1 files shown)


src/tests.c
```diff
@@ -7388,6 +7388,120 @@ static void run_ecdsa_der_parse(void) {
...
+    CHECK(compact[63] == 0x23);
+}
+
+/* Tests the long form length encoding (X.690-0207 8.1.3.5).
+ *
+ * random_ber_signature only emits long form lengths in signatures it marks as
+ * certainly_not_der, so run_ecdsa_der_parse never asserts that a long form
+ * length is accepted. Note that the long form is only valid in DER for lengths
+ * of at least 128, which is more than a signature with two in-range scalars
+ * needs, so the R integers below necessarily overflow to zero. */
```
real-or-random: I get that these values overflow, but can you explain why overflow "to zero"?

real-or-random (commented):
Concept ACK

Nice!

src/tests.c
brunoerg: Afaik, it explicitly zeroes the scalar on any overflow. But anyway, the wording here is not good. I will change it.

brunoerg: Done. Let me know whether it's better.

real-or-random (approved):
utACK ddb1dfa77038442dff86ba90e6d0a30adfddef17

src/tests.c
real-or-random: Okay, my question was kinda stupid, but yes, the improved comment is perfectly clear.

theStack (approved):
ACK ddb1dfa77038442dff86ba90e6d0a30adfddef17

PR #1932, opened 2026-09-07
silentpayments: drop "empty key arrays must be NULL" requirement

### LINKED ISSUE #1930: silentpayments: NULL requirement for empty key arrays (inconsistency between API docs and ARG_CHECKs)
In the silentpayments API, the following functions take two lists of keys as input (each represented as "pointer to array of pointers" and "size" parameters, i.e. four parameters in total):
* `_silentpayments_sender_create_outputs` (secret keys: parameters `keypairs`/`n_keypairs`, `seckeys`/`n_seckeys`)
* `_silentpayments_recipient_prevouts_summary_create` (public keys: `xonly_pubkeys`/`n_xonly_pubkeys`, `pubkeys`/`n_pubkeys`)

The public header docs state that the "pointer to array of pointers" parameters each _"can be NULL"_ if no such inputs are provided. However, the ARG_CHECKs currently are more strict and require the pointer parameter to be NULL if the size is zero, i.e. _"must be NULL"_ would be more appropriate.

One obvious way to fix this inconsistency is to just adapt the docs to match the code accordingly, e.g.: https://github.com/theStack/secp256k1/commit/1f0a8bfe9bab4e4a268757a66e0f4ee8c1d601f0

Alternatively, I wonder if dropping this "must be NULL" requirement would also be an option (which AFAICT would be backwards-compatible) , as I think enforcing it doesn't have much value and is in the end just making things slightly more inconvenient for the user, forcing them to use conditional constructs like `keys_size == 0 ? NULL : keys_ptrs` for the call-sites. This would be a bit of a larger patch as both the ARG_CHECKs within the functions and the tests have to be adapted though, see https://github.com/bitcoin-core/secp256k1/compare/master...theStack:secp256k1:sp-keylists-drop-NULL-requirement.

Noticed while reviewing Bitcoin Core [PR #35301](https://github.com/bitcoin/bitcoin/pull/35301), where the requirement is currently not met, even though the CI is green (background: the C++ standard doesn't guarantee that `.data()` on an empty `std::vector` instance returns `nullptr`, see https://en.cppreference.com/cpp/container/vector/data: _"If `size()` is 0, `data()` may or may not return a null pointer."_).

theStack:
Resolves #1930.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+#### Fixed
+ - Module `silentpayments`: `secp256k1_silentpayments_sender_create_outputs` and `secp256k1_silentpayments_recipient_prevouts_summary_create` no longer require empty key arrays to be passed as `NULL`. If the corresponding size argument is 0, the array pointer is ignored. This matches the API documentation, which only states that unused arrays *can* be `NULL`, and spares callers from special-casing empty arrays.
+
 ## [0.8.0] - 2026-08-03
 
 #### Added
--- a/src/modules/silentpayments/main_impl.h
+++ b/src/modules/silentpayments/main_impl.h
@@ -210,22 +210,18 @@ int secp256k1_silentpayments_sender_create_outputs(
     ARG_CHECK(recipients != NULL);
     ARG_CHECK(n_recipients > 0);
     ARG_CHECK(outpoint_smallest36 != NULL);
-    ARG_CHECK((seckeys != NULL) || (keypairs != NULL));
-    if (keypairs != NULL) {
-        ARG_CHECK(n_keypairs > 0);
+    ARG_CHECK((n_seckeys > 0) || (n_keypairs > 0));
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+#### Changed
```
real-or-random: nit: I think it's a bug, strictly speaking, because the implementation doesn't match the docs. 
 
```suggestion
#### Fixed
```

real-or-random:
@nymius want to review this? :)

CHANGELOG.md
theStack: Agree, done.

real-or-random (approved):
utACK ad3688d1e5a1b2afdbc613b2489e2fdb2ec4121d

nymius (commented):
ACK ad3688d1e5a1b2afdbc613b2489e2fdb2ec4121d

I've to implement the bindings on top to be sure, as there are chances the code requires the checks anyway because of type inference.

PR #1934, opened 2026-09-09
Clear secret-dependent variables in `_ecmult_const_xonly`

theStack:
While working on the refactoring PR #1933, I noticed that there is no `_gej` clearing of the point multiplication result in `_ecmult_const_xonly` (apparently this was missed back then when looking for instances in https://github.com/bitcoin-core/secp256k1/pull/1579#issuecomment-2405693607, likely because in this case there is no `_ge_set_gej` call involved).

Note that for `_ecmult_const` itself there is already a separate PR: https://github.com/bitcoin-core/secp256k1/pull/1885


real-or-random (approved):
utACK 9e65f6ac9b714b85bf03338ffb900a6aeebed800

PR #1935, opened 2026-09-09
tests: generate scalars_near_split_bounds instead of hardcoding

### LINKED ISSUE #1920: scalars_near_split_bounds in tests.c don't match their documented construction

Anshikakalpana:
tests: generate scalars_near_split_bounds instead of hardcoding

The hardcoded scalars_near_split_bounds array didn't match its documented formula (a*LAMBDA + (ORDER+b)/2) % ORDER. All 20 entries were off in the same 32-bit window, likely a transcription error.
As a result the vectors only reached ~126-bit outputs through secp256k1_scalar_split_lambda instead of the intended ~127-128-bit boundary, so the case they exist to test was never exercised.

Generate the values at runtime from the formula via fill_scalars_near_split_bounds(), which writes into a caller-provided
array rather than a file-scope static. It is called at each of the three sites that use the array, so the values can no longer drift
from their formula.

Verified the generated values reach within 1-2 units of the true k1_bound/k2_bound from the split_lambda_verify proof, confirming theexisting a/b ranges are sufficient.

Fixes #1920.

### DIFF
--- a/src/tests.c
+++ b/src/tests.c
@@ -4645,28 +4645,27 @@ static void test_point_times_order(const secp256k1_gej *point) {
  *   - For b in [-3, -1, 1, 3]:
  *     - Output (a*LAMBDA + (ORDER+b)/2) % ORDER
  */
-static const secp256k1_scalar scalars_near_split_bounds[20] = {
-    SECP256K1_SCALAR_CONST(0xd938a566, 0x7f479e3e, 0xb5b3c7fa, 0xefdb3749, 0x3aa0585c, 0xc5ea2367, 0xe1b660db, 0x0209e6fc),
-    SECP256K1_SCALAR_CONST(0xd938a566, 0x7f479e3e, 0xb5b3c7fa, 0xefdb3749, 0x3aa0585c, 0xc5ea2367, 0xe1b660db, 0x0209e6fd),
-    SECP256K1_SCALAR_CONST(0xd938a566, 0x7f479e3e, 0xb5b3c7fa, 0xefdb3749, 0x3aa0585c, 0xc5ea2367, 0xe1b660db, 0x0209e6fe),
-    SECP256K1_SCALAR_CONST(0xd938a566, 0x7f479e3e, 0xb5b3c7fa, 0xefdb3749, 0x3aa0585c, 0xc5ea2367, 0xe1b660db, 0x0209e6ff),
-    SECP256K1_SCALAR_CONST(0x2c9c52b3, 0x3fa3cf1f, 0x5ad9e3fd, 0x77ed9ba5, 0xb294b893, 0x3722e9a5, 0x00e698ca, 0x4cf7632d),
-    SECP256K1_SCALAR_CONST(0x2c9c52b3, 0x3fa3cf1f, 0x5ad9e3fd, 0x77ed9ba5, 0xb294b893, 0x3722e9a5, 0x00e698ca, 0x4cf7632e),
-    SECP256K1_SCALAR_CONST(0x2c9c52b3, 0x3fa3cf1f, 0x5ad9e3fd, 0x77ed9ba5, 0xb294b893, 0x3722e9a5, 0x00e698ca, 0x4cf7632f),
-    SECP256K1_SCALAR_CONST(0x2c9c52b3, 0x3fa3cf1f, 0x5ad9e3fd, 0x77ed9ba5, 0xb294b893, 0x3722e9a5, 0x00e698ca, 0x4cf76330),
-    SECP256K1_SCALAR_CONST(0x7fffffff, 0xffffffff, 0xffffffff, 0xffffffff, 0xd576e735, 0x57a4501d, 0xdfe92f46, 0x681b209f),
-    SECP256K1_SCALAR_CONST(0x7fffffff, 0xffffffff, 0xffffffff, 0xffffffff, 0xd576e735, 0x57a4501d, 0xdfe92f46, 0x681b20a0),
-    SECP256K1_SCALAR_CONST(0x7fffffff, 0xffffffff, 0xffffffff, 0xffffffff, 0xd576e735, 0x57a4501d, 0xdfe92f46, 0x681b20a1),
-    SECP256K1_SCALAR_CONST(0x7fffffff, 0xffffffff, 0xffffffff, 0xffffffff, 0xd576e735, 0x57a4501d, 0xdfe92f46, 0x681b20a2),
-    SECP256K1_SCALAR_CONST(0xd363ad4c, 0xc05c30e0, 0xa5261c02, 0x88126459, 0xf85915d7, 0x7825b696, 0xbeebc5c2, 0x833ede11),
-    SECP256K1_SCALAR_CONST(0xd363ad4c, 0xc05c30e0, 0xa5261c02, 0x88126459, 0xf85915d7, 0x7825b696, 0xbeebc5c2, 0x833ede12),
-    SECP256K1_SCALAR_CONST(0xd363ad4c, 0xc05c30e0, 0xa5261c02, 0x88126459, 0xf85915d7, 0x7825b696, 0xbeebc5c2, 0x833ede13),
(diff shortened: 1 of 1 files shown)


src/tests.c
```diff
@@ -4644,29 +4644,32 @@ static void test_point_times_order(const secp256k1_gej *point) {
...
-    SECP256K1_SCALAR_CONST(0xd363ad4c, 0xc05c30e0, 0xa5261c02, 0x88126459, 0xf85915d7, 0x7825b696, 0xbeebc5c2, 0x833ede11),
-    SECP256K1_SCALAR_CONST(0xd363ad4c, 0xc05c30e0, 0xa5261c02, 0x88126459, 0xf85915d7, 0x7825b696, 0xbeebc5c2, 0x833ede12),
-    SECP256K1_SCALAR_CONST(0xd363ad4c, 0xc05c30e0, 0xa5261c02, 0x88126459, 0xf85915d7, 0x7825b696, 0xbeebc5c2, 0x833ede13),
-    SECP256K1_SCALAR_CONST(0xd363ad4c, 0xc05c30e0, 0xa5261c02, 0x88126459, 0xf85915d7, 0x7825b696, 0xbeebc5c2, 0x833ede14),
-    SECP256K1_SCALAR_CONST(0x26c75a99, 0x80b861c1, 0x4a4c3805, 0x1024c8b4, 0x704d760e, 0xe95e7cd3, 0xde1bfdb1, 0xce2c5a42),
-    SECP256K1_SCALAR_CONST(0x26c75a99, 0x80b861c1, 0x4a4c3805, 0x1024c8b4, 0x704d760e, 0xe95e7cd3, 0xde1bfdb1, 0xce2c5a43),
-    SECP256K1_SCALAR_CONST(0x26c75a99, 0x80b861c1, 0x4a4c3805, 0x1024c8b4, 0x704d760e, 0xe95e7cd3, 0xde1bfdb1, 0xce2c5a44),
-    SECP256K1_SCALAR_CONST(0x26c75a99, 0x80b861c1, 0x4a4c3805, 0x1024c8b4, 0x704d760e, 0xe95e7cd3, 0xde1bfdb1, 0xce2c5a45)
-};
+static secp256k1_scalar scalars_near_split_bounds[20];
```
real-or-random: This should not be static at the file level. The function should take an out param `scalars20`

src/tests.c
```diff
@@ -4644,29 +4644,32 @@ static void test_point_times_order(const secp256k1_gej *point) {
  * - For a in [-2, -1, 0, 1, 2]:
  *   - For b in [-3, -1, 1, 3]:
  *     - Output (a*LAMBDA + (ORDER+b)/2) % ORDER
+ *
+ * Generated at runtime by fill_scalars_near_split_bounds() rather than hardcoded.
```
real-or-random: This comment is a typical AI comment that should be dropped. It will confuse any reader who isn't aware of the history will wonder what it's supposed to convey.

real-or-random (commented):
Thanks for picking that work up!

src/tests.c
Anshikakalpana: Fixed both ,switched to an out param and dropped the comment.

real-or-random (commented):
ACK but can you squash this into a single commit? Ideally, the commit message would be the same as PR summary in your first comment here (and possibly updated)

Anshikakalpana:
updated!

real-or-random (approved):
utACK 014825f0a30f3968efe51bf782ac2d4662841a41

PR #1931, opened 2026-09-07
tests: cover rejection of invalid plain seckey alongside a valid one

brunoerg:
The existing silentpayment sender tests only pass a single invalid plain seckey, which parses to a zero scalar and is therefore also rejected by the later zero-sum check. Add a test with a valid key followed by an invalid one so that the early return in the seckey loop itself is exercised.

It kills the following mutant, the only one remaining within 229 ones :) - https://secp256k1.space/src/modules/silentpayments/main_impl.h

```diff
diff --git a/src/modules/silentpayments/main_impl.h b/src/modules/silentpayments/main_impl.h
index 415f2d2..489ec91 100644
--- a/src/modules/silentpayments/main_impl.h
+++ b/src/modules/silentpayments/main_impl.h
@@ -237,7 +237,7 @@ int secp256k1_silentpayments_sender_create_outputs(
     for (i = 0; i < n_seckeys; i++) {
         ret = secp256k1_scalar_set_b32_seckey(&addend, seckeys[i]);
         secp256k1_declassify(ctx, &ret, sizeof(ret));
-        if (!ret) {
+        if (1==0) {
             secp256k1_scalar_clear(&addend);
             secp256k1_scalar_clear(&seckey_sum_scalar);
             return 0;
```

PR #1937, opened 2026-09-11
tests: silentpayments: cover rejection of invalid keypair alongside a valid one

Yudis-bit:
Add coverage for an invalid Taproot keypair passed alongside a valid one, following the plain secret key tests in #1931.

Both orderings are tested. The invalid-first case catches a missing early return in the keypair loop: the valid keypair leaves a nonzero sum, so the subsequent zero-sum check would not catch the failure.


real-or-random:
cc @brunoerg

brunoerg:
Concept ACK

I'm adding more mutation operators to my tool and nested early-return deletion is one of them, the following mutant isn't killed on master but killed with the test case added here (nice!):

```diff
diff --git a/src/modules/silentpayments/main_impl.h b/src/modules/silentpayments/main_impl.h
index e42dc1a..8d7be21 100644
--- a/src/modules/silentpayments/main_impl.h
+++ b/src/modules/silentpayments/main_impl.h
@@ -249,7 +249,6 @@ int secp256k1_silentpayments_sender_create_outputs(
         if (!ret) {
             secp256k1_scalar_clear(&addend);
             secp256k1_scalar_clear(&seckey_sum_scalar);
-            return 0;
         }
         if (secp256k1_fe_is_odd(&addend_point.y)) {
             secp256k1_scalar_negate(&addend, &addend);
```

src/modules/silentpayments/tests_impl.h
```diff
@@ -295,6 +295,20 @@ static void test_send_api(void) {
         p2[0] = secp256k1_group_order_bytes;
         CHECK(secp256k1_silentpayments_sender_create_outputs(CTX, op, rp, 2, SMALLEST_OUTPOINT, NULL, 0, p2, 2) == 0);
     }
+    /* Check that an invalid keypair is caught even when it is passed alongside a valid one.
```
brunoerg, line 298: The "first nor last keypair skipped" rationale is already covered by the existing checks. What this block actually adds is the invalid-first ordering: if the early return 0; were dropped, a valid keypair followed by an invalid one still fails via the zero-sum check, but an invalid keypair followed by a valid one leaves a nonzero sum and wrongly succeeds.

Could you reword the comment to say that, e.g. "pass the invalid keypair first, followed by a valid one, so that silently skipping it is not masked by the subsequent zero-sum check"?

src/modules/silentpayments/tests_impl.h
```diff
@@ -295,6 +295,20 @@ static void test_send_api(void) {
...
+     * The invalid keypair is tested in both positions so that neither the first nor the
+     * last keypair is skipped by the check. */
+    {
+        secp256k1_keypair valid_keypair;
+        const secp256k1_keypair *t2[2];
```
brunoerg: nit:

```suggestion
        secp256k1_keypair const *t2[2];
```

src/modules/silentpayments/tests_impl.h
Yudis-bit: Reworded the comment in a044ade to explain the invalid-first case, and updated the PR description to match. Also applied the const placement suggestion in ca6b4a2.

theStack:
Concept ACK

@Yudis-bit: Could you squash the commits, please?

Yudis-bit:
Squashed into e5b4d58. No changes to the final diff.

PR #1940, opened 2026-09-21
silentpayments: add hash ctx (SHA256 compression override) test

theStack:
This PR adds a test to check that the ctx-provided SHA256 compression override takes effect for the following `silentpayments` module API functions:
* `secp256k1_silentpayments_label_create` (hash result is used for calculating the [label tweak](https://github.com/bitcoin-core/secp256k1/blob/46db787112beabdb5e17e0dc35680716f1057e7b/src/modules/silentpayments/main_impl.h#L407-L412))
* `secp256k1_silentpayments_sender_create_outputs` (hash results are used for calculating the [input hash](https://github.com/bitcoin-core/secp256k1/blob/46db787112beabdb5e17e0dc35680716f1057e7b/src/modules/silentpayments/main_impl.h#L72-L76) and [output tweak](https://github.com/bitcoin-core/secp256k1/blob/46db787112beabdb5e17e0dc35680716f1057e7b/src/modules/silentpayments/main_impl.h#L125-L135))
* `secp256k1_silentpayments_recipient_prevouts_summary_create` (hash result is used for calculating the [input hash](https://github.com/bitcoin-core/secp256k1/blob/46db787112beabdb5e17e0dc35680716f1057e7b/src/modules/silentpayments/main_impl.h#L72-L76))
* `secp256k1_silentpayments_recipient_scan_outputs` (hash result is used for calculating the [output tweak](https://github.com/bitcoin-core/secp256k1/blob/46db787112beabdb5e17e0dc35680716f1057e7b/src/modules/silentpayments/main_impl.h#L125-L135))

With #1866 for musig and this PR, all public functions that take use of the SHA256 compression override should be covered by tests.


src/modules/silentpayments/tests_impl.h
```diff
@@ -931,6 +932,82 @@ static void silentpayments_sha256_tag_test(void) {
     }
 }
 
+DEFINE_SHA256_TRANSFORM_PROBE(sha256_sp_label_create)
+DEFINE_SHA256_TRANSFORM_PROBE(sha256_sp_create_outputs)
+DEFINE_SHA256_TRANSFORM_PROBE(sha256_sp_prevouts_summary_create)
+DEFINE_SHA256_TRANSFORM_PROBE(sha256_sp_scan_outputs)
```
theStack, line 938: self-nit: I was on the fence between adding multiple probes or a single one (that is set back repeatedly, leading to approximately the same LOC), but decided to reflect the style currently used in #1866. Happy with either style if reviewers have a preference.

